PR #483, opened 2014-12-21
Add deferred example

serpis:
Add deferred shading example.
It demonstrates the use of render targets and uniform buffers.


kvark:
Thanks @serpis, looking forward to it!

PR #497, opened 2015-01-11
Updated macro syntax for gfx-macros

leonkunert:


leonkunert:
never mind. is already in #496

csherratt:
actually this is nicer done then #496. If you don't mind I might roll this change in with it.

leonkunert:
Great. Go ahead

csherratt:
Closed via #496

PR #498, opened 2015-01-12
Clean-up warnings from alpha changes

aepsil0n:
This includes mainly slice notation and renaming to usize/isize. The vertex format macro no longer advises against the use of usize/isize specifically, as the new name alone should deter people from using it for numerical purposes.

To suppress the instability warnings, the "unstable" lint is allowed in all crates.

Oh, and I also restructured `Cargo.toml` to get an overview of the dependencies. I could send a separate pull request for this if that is desired.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,22 +13,22 @@
 # limitations under the License.
 
 [package]
-
 name = "gfx"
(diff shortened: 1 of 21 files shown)


csherratt:
Thank you!

PR #502, opened 2015-01-20
Derive Copy for RefBatch

bvssvni:
This blocks Hematite from working.


kvark:
Looks harmless enough. Waiting for Travis...

toqueteos:
Thank god to this patch, hematite is almost back to life.

PR #503, opened 2015-01-24
Fix build with latest rustc

csherratt:
Fix the gfx_macros for latest ast changes.
Fix the fact that unwrap() has been removed.


kvark:
Nice, thanks!

PR #504, opened 2015-01-25
Make CommandBuffer and associated type on Device

bjz:
This compiles, but is broken by servo/rust-core-foundation#51

### DIFF
--- a/src/device/lib.rs
+++ b/src/device/lib.rs
@@ -68,13 +68,13 @@ pub enum MapAccess {
 }
 
 /// A handle to a readable map, which can be sliced.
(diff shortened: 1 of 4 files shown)


src/device/lib.rs
```diff
@@ -356,17 +356,18 @@ pub enum Command {
     Blit(target::Rect, target::Rect, target::Mask),
 }
 
-// CommandBuffer is really an associated type, so will look much better when
```
csherratt: \o/

csherratt:
Looks good to me.

kvark:
Awesome, the code looks much better now!

Issue #400, opened 2014-09-22
Refactor ShaderParam using associated types


kvark:
We could hide the link type entirely, simplifying the interface and streamlining the batch usage.

bjz:
\o/

bvssvni:
Hey, what happened?

bvssvni:
Oh, nice!

kvark:
Actually, with this in we no longer need to generate a typedef in the macros. One can just say `RefBatch<YourStruct>` and be good with it.

sectopod:
@kvark yeah, that's the end goal. I'm on it ;)

bjz:
Merged now!

PR #509, opened 2015-01-30
Removed batch typedef from #[shader_param]

sectopod:
Completes #400


bjz:
This is a huge ergonomics win. Thanks!

PR #512, opened 2015-02-01
Changes for publish to Crates.io

### LINKED ISSUE #460: Put gfx-rs on crates.io
I think this library is awesome, and should probably get registered on https://crates.io/
That is all.

csherratt:
Resolves #460 

There is some renaming and moving around to make this all work.

`gfx` now contains `render` until we choose to split it.
`device` and `gl_device` are merged for now as `gfx_device_gl`. This will get split later.
`gfx_macros` is publish separately.

There are some things that should be considered before merging this. Mostly to do with the ergonomics of publishing. The relative paths had to be removed to publish to crates.io, but this makes it pretty annoying to do any local work. It probably make sense for this repo to just use relative paths, and a script (or manually) have to modify the paths prior to publishing.


bjz:
> The relative paths had to be removed to publish to crates.io, but this makes it pretty annoying to do any local work. It probably make sense for this repo to just use relative paths, and a script (or manually) have to modify the paths prior to publishing.

Yeah - I think @bvssvni uses a system of commenting/uncommenting out the local path vs the crates.io url. It's not very nice though. I would also be concerned about a brittle deployment process - it would be easy to mess it up.

/cc @wycats

kvark:
yay! thanks @csherratt

csherratt:
I think this is ready.

kvark:
Looks good to me! :+1: 
@bjz, how about you?

PR #514, opened 2015-02-02
ProgramError export

kvark:


kvark:
Self-merging due to the harmless nature of the change.

PR #519, opened 2015-02-04
Updated gfx_macros to the latest Rust

kvark:


kvark:
Self-merging as harmless.

PR #520, opened 2015-02-04
Fix for gl_device errors with latest Rust

kvark:
This will get us ready to the point when gl-rs gets fixed.


kvark:
Self-merging as harmless.

PR #533, opened 2015-02-05
Better comments for CommandBuffer

kvark:


kvark:
Self-merging as harmless.

bjz:
:+1:

PR #535, opened 2015-02-05
Implemented blit mirror flags

### LINKED ISSUE #522: Support mirroring when blitting
Currently, our [blit](https://github.com/gfx-rs/gfx-rs/blob/a753f73b708b835daf95726f4b76ee3693c053d8/src/render/lib.rs#L203) receives rectangle ares with positive coordinates. GL specification allows negative values to produce mirroring:

> If either the source or destination rectangle specifies a negative width or height (`X1 < X0` or `Y1 < Y0`), the image is reversed in the corresponding direction. If both the source and destination rectangles specify a negative width or height for the same direction, no reversal is performed.

kvark:
Fixes #522

### DIFF
--- a/src/device/gl_device/lib.rs
+++ b/src/device/gl_device/lib.rs
@@ -517,8 +517,19 @@ impl GlDevice {
                     },
                 }
             },
-            Command::Blit(s_rect, d_rect, mask) => {
+            Command::Blit(mut s_rect, d_rect, mirror, mask) => {
                 type GLint = gl::types::GLint;
+                // mirror
+                let mut s_end_x = s_rect.x + s_rect.w;
+                let mut s_end_y = s_rect.y + s_rect.h;
+                if mirror.intersects(::target::MIRROR_X) {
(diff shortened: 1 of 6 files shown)


src/device/gl_device/lib.rs
```diff
@@ -517,8 +517,19 @@ impl GlDevice {
...
+                let mut s_end_x = s_rect.x + s_rect.w;
+                let mut s_end_y = s_rect.y + s_rect.h;
+                if mirror.intersects(::target::MIRROR_X) {
+                    s_end_x = s_rect.x;
+                    s_rect.x += s_rect.w;
+                }
+                if mirror.intersects(::target::MIRROR_Y) {
+                    s_end_y = s_rect.y;
+                    s_rect.y += s_rect.h;
+                }
```
csherratt: Won't you have to do this for both `s_rect` and `d_rect`?

csherratt: Ah, nvm. I see understand what you are doing now.

PR #536, opened 2015-02-05
Version bump to 0.1.2

kvark:
... so that we can re-publish after ~~todays~~ yesterdays changes. Hematite would enjoy that ;)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.1.1"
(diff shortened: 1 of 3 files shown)


kvark:
Please publish it.

-Dzmitry

bjz:
Do you not have publisher rights? You seem to here: https://crates.io/crates/gfx

kvark:
@bjz I do, but I didn't want it to wait till I get home.

PR #538, opened 2015-02-06
BGRA8 texture support

### LINKED ISSUE #523: Support for BGRA textures

kvark:
Closes #523 

I decided to make BGRA an exception because it has a very limited use case - interoperation with BMP, which is unlikely anything but 8 bit unsigned normalized data.


csherratt:
lgtm. r+

PR #540, opened 2015-02-06
Added draw call errors for missing targets

### LINKED ISSUE #525: Detect when the depth/stencil test is used with a missing attachment
We should return an error for these calls.

kvark:
Closes #525 

Note: we also check for missing colors

### DIFF
--- a/src/gfx/lib.rs
+++ b/src/gfx/lib.rs
@@ -24,7 +24,7 @@ extern crate libc;
 extern crate "gfx_device_gl" as device;
 
 // public re-exports
-pub use render::{DeviceHelper, ProgramError, Renderer};
+pub use render::{DeviceHelper, ProgramError, Renderer, DrawError};
 pub use render::batch;
 pub use render::mesh::{Attribute, Mesh, VertexFormat};
 pub use render::mesh::{Slice, ToSlice};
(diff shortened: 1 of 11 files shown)


src/gfx/lib.rs
```diff
@@ -89,20 +89,11 @@ impl<D: device::Device> Graphics<D> {
...
-    /// Draw a ref batch with instancing
-    pub fn draw_instanced<'a, T: shade::ShaderParam>(&'a mut self,
-                          batch: &'a batch::RefBatch<T>,
-                          data: &'a T,
-                          count: u32,
-                          vertex_offset: u32,
-                          frame: &Frame) {
-        self.renderer.draw_instanced((batch, data, &self.context), count, vertex_offset, frame)
-    }
-
```
bjz: Why was this removed?

kvark: It's in the `Graphics` struct, which is a simple wrapper around `Renderer` and `Device` for single-threaded mode. It's only a convenience shortcut for simple applications. Once you want anything more advanced, you get rid of `Graphics` or just use `Renderer` directly (it's a public field). Sounds good?

bjz: Sounds good. I would really like to figure out a better way to do this...

bjz: Ie. not need the `Graphics` thing.

bjz:
Re-running travis

bjz:
Broken on closure syntax

PR #543, opened 2015-02-07
Updated verion for publish

csherratt:


kvark:
I already bumped the version in #540 , but I missed the sub-crates. I guess it doesn't matter if we skip some :)

PR #550, opened 2015-02-09
Re-export some missing types

csherratt:


kvark:
How about just `pub use device::target::*`?

csherratt:
done.

kvark:
Thanks :)

Issue #529, opened 2015-02-05
Implement GLFW/Glutin bridges as separate crates


kvark:
@dylanede:
> But you could provide a trait and then split integration into separate crates?
> Just seems a bit painful for everyone to reinvent the wheel when they first use gfx with any windowing system

Every gfx-rs app has several lines of code dedicated to window creation. We could simplify this by introducing "bridge" crates (aka `gfx_bridge_glfw`) under `gfx-rs`  organization that would allow:
  - one-line window creation
  - return the default frame buffer

See also: #515

Note: historically, we had glfw and glutin (later) integration from the start. At some point, we decided to isolate away from it. Looks like it's time to bring those back as separate crates.

bjz:
What wheel are we talking about?

kvark:
@bjz better integration with glfw and Glutin. We need to bring back those bridges for convenient window creation (as separate crates under gfx-rs organization). This would also make the default framebuffer creation safer (#530)

bjz:
Yeah this would be good, and make sense in separate crates. :+1:

bjz:
Could you clearly specify a list of the things we need in the issue description, based off #530?

kvark:
@bjz better now?

bjz:
Sounds great. Not sure about the naming, but we can bikeshed that some other time.

kvark:
This is especially important now that we have `device.after_frame()`.

PR #554, opened 2015-02-12
Removal of shaders! macro

### LINKED ISSUE #552: Fix or refactor shader! macro
Apparently, it's causing build issues now, according to https://github.com/PistonDevelopers/hematite/pull/193

As a suggestion, we could remove the macro entirely. Instead, one can just write this instead:
```rust
gfx::ShaderSource {
    glsl_120: Some(<shader_code>),
    .. Default::default(),
}
```
This is a bit wordier but has less magic in it. We've got to revamp all the shader ergonomics anyway.

kvark:
Closes #552


examples/terrain/main.rs
```diff
@@ -151,10 +118,6 @@ fn main() {
     let mut glfw = glfw::init(glfw::FAIL_ON_ERRORS)
         .ok().expect("Failed to initialize glfs-rs");
 
-    glfw.window_hint(glfw::WindowHint::ContextVersion(3, 2));
```
csherratt: @bjz I recall there was at least one platform that behaved strangely if you did not provide these. Do you remember?

csherratt:
Even if the window_hint removal was wrong, we can easily add it back in as a regression. It only effects the examples. I am going to merge this.

examples/terrain/main.rs
bjz: OS X needs the window hints.

kvark:
Thanks guys!

I was thinking, the context issues will be largely addressed by #529, but then noticed #555.

PR #557, opened 2015-02-13
Batch implementation for the components

### LINKED ISSUE #556: Implement `Batch` for a tuple of its components
Batch concept has proven to be a strong solution for safe and performant code. However, it is slightly more verbose that simply passing the components: `Mesh`, `Slice`, `Program`, and `DrawState`, which is what glium does.

The good thing is - we don't have to sacrifice ergonomics here. We can simply have the `Batch` implemented for the components tuple. It is not intended for serious applications, because it requires the validation of shader inputs at each call. However, it makes gfx more compatible with glium, allows less code for simple demos, and we are not loosing anything.

kvark:
Fixes #556 

Note: this PR does not change the API.

Please have a good look at it before submission: `Batch` interface is changed. I feel that it is more sound now, since the program and its parameters are handled by the `fill_params` method, while `get_data` is mostly taking care of the mesh.

The batch errors appeared to be useful for the `RefBatch` as well as the new component-based implementation, because `RefBatch` can technically be used with the wrong context.

Also, the `AttributeIndex` type seems to improve read-ability of the code.

### DIFF
--- a/examples/triangle/main.rs
+++ b/examples/triangle/main.rs
@@ -18,7 +18,7 @@
 extern crate gfx;
 extern crate glfw;
 
-use gfx::{DeviceExt, ToSlice};
+use gfx::{Device, DeviceExt, ToSlice};
 use glfw::Context;
 
 #[vertex_format]
@@ -82,15 +82,14 @@ fn main() {
     let program = device.link_program(VERTEX_SRC, FRAGMENT_SRC)
                         .ok().expect("Failed to link program");
 
-    let mut graphics = gfx::Graphics::new(device);
-    let batch: gfx::batch::RefBatch<()> = graphics.make_batch(
-        &program, &mesh, slice, &gfx::DrawState::new()).ok().expect("Failed to make batch");
+    let mut renderer = device.create_renderer();
 
     let clear_data = gfx::ClearData {
         color: [0.3, 0.3, 0.3, 1.0],
         depth: 1.0,
         stencil: 0,
     };
+    let state = gfx::DrawState::new();
 
     while !window.should_close() {
         glfw.poll_events();
@@ -102,9 +101,10 @@ fn main() {
             }
         }
 
-        graphics.clear(clear_data, gfx::COLOR, &frame);
-        graphics.draw(&batch, &(), &frame).unwrap();
-        graphics.end_frame();
+        renderer.reset();
+        renderer.clear(clear_data, gfx::COLOR, &frame);
+        renderer.draw(&(&mesh, slice, &program, &(), &state), &frame).unwrap();
+        device.submit(renderer.as_buffer());
 
         window.swap_buffers();
     }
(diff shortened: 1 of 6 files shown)


kvark:
I've made the triangle example be the test for component-based batches. Also switched it from `Graphics` to `Device` + `Renderer`, since `Graphics` is meant for `RefBatch`. I believe this is double win, because the triangle example is now slightly lower level, thus easier to grasp. It has no `Batch` concept or `Graphics` wrapper in it.

examples/triangle/main.rs
```diff
@@ -82,15 +82,14 @@ fn main() {
     let program = device.link_program(VERTEX_SRC, FRAGMENT_SRC)
                         .ok().expect("Failed to link program");
 
-    let mut graphics = gfx::Graphics::new(device);
-    let batch: gfx::batch::RefBatch<()> = graphics.make_batch(
```
csherratt: This is a nice improvement in readability.

bjz: Agreed!

examples/triangle/main.rs
```diff
@@ -102,9 +101,10 @@ fn main() {
             }
         }
 
-        graphics.clear(clear_data, gfx::COLOR, &frame);
-        graphics.draw(&batch, &(), &frame).unwrap();
-        graphics.end_frame();
+        renderer.reset();
+        renderer.clear(clear_data, gfx::COLOR, &frame);
+        renderer.draw(&(&mesh, slice, &program, &(), &state), &frame).unwrap();
```
bjz: Just wondering if it would be tricky to remember this list of params

kvark: Certainly, the error message will not be too helpful. It will just say "Can't find implementation of `Batch` for (your tuple here)". However, adding a special method for it doesn't seem to be a good solution. Perhaps, we could have a named struct instead of the tuple:
```rust
renderer.draw(&gfx::batch::Implicit(&mesh, slice, &program, &(), &state), &frame).unwrap();
```
This will be more wordy but will produce better error messages. Also please note that `RefBatch` has the same problem (with tuple errors), but it would be highly uncomfortable to convert the tuple to another struct there.

All in all, I believe we can add this (the original PR) and decide later, thinking about both `RefBatch` and this calls as parts of the same problem.

bjz: Yup, agreed. Can we merge this?

kvark: I'm all for it. Not sure if @csherratt has finished reviewing it or not.
How about giving him another hour or two and then merge?

csherratt: I am away this weekend. So I'd bjz is happy merge away.

kvark: @csherratt excuses! it's not weekend yet :P

bjz: It is for me!

PR #567, opened 2015-02-17
Added SamplerHandle public export

kvark:


kvark:
Self-merging as harmless.

PR #576, opened 2015-02-19
Fixed gfx_macros for the latest Rust

kvark:


kvark:
Merging as harmless. The tests are broken because of the `rand_macros` crate, but `gfx_macros` is good with this one.

Issue #416, opened 2014-11-03
Make Device completely self-contained with associated types


cmr:
This will be required for supporting multiple-backends. Blocked on https://github.com/rust-lang/rust/issues/17307

PR #589, opened 2015-02-22
Decouple render::mesh from GlDevice

bjz:
Working towards #416


kvark:
I trust you on this. Just one question:
why do generalize some stuff completely (with `R: Resources` bound), while blantly putting `GlResources` into other spots?

bjz:
I've tried doing the whole lot at once and it becomes unmanageable. So starting at the leaves of the API, I gradually work my way up, pushing the `GlResources` dependency higher an higher until I've reached the root of the API.

PR #601, opened 2015-02-26
fix shader_param / vertex_format for the current nightly

MaikKlein:


kvark:
Thanks @MaikKlein !

Don't want to wait for Travis. Taking responsibility to merge this (it was broken anyway).

kvark:
@csherratt apparently, genmesh is also broken (see travis log)

PR #602, opened 2015-02-26
temporary docs fix

MaikKlein:


MaikKlein:
Might not compile because my gfx_gl fix is not on crates.io yet.

kvark:
Travis is still broken by genmesh, I'm merging this.
Thanks @MaikKlein !

PR #605, opened 2015-02-28
Move render and device into gfx directory

bjz:


kvark:
Would it be better to have `src/lib.rs`, `src/render` and `src/device` instead? It seems like this PR introduces another depth layer for no reason. The only thing stopping us from having `src/lib.rs` and stuff is gfx_macros, which we can just move into another crate.

kvark:
Heck, you could even do that before moving `gfx_macros` out.

bjz:
I disagree - this makes it easier for folks to understand, at least in the interim. I worry about moving everything out into separate repos - makes things super hard to maintain.

kvark:
@bjz I don't like the idea of having just `gfx` and `gfx_macros` in `src`. If you don't want to move `gfx_macros` out into a separate library, let's just leave the structure as is.
Also, @csherratt and @cmr  - where do you stand?

bjz:
It's just confusing - from the directory structure it would seem as if `render` and `device` are separate crates.

kvark:
but.. they might as well be separate crates. It's just a technical details that they are modules.

bjz:
I'm talking about conventions here - using `#[path = "..."]` is pretty non-standard.

bjz:
I'll drop this now though - just think about it.

kvark:
How about that - we move `lib.rs` into `src`, thus removing the need for `path=`, but leaving `gfx_macros` where it is (for now)?

bjz:
But then it seems like `gfx_macros` is a sub module...

It would be really nice to keep `gfx_macros` in the same repo as `gfx`...

csherratt:
I agree with this organization change. It does not stop us from splinting this change up into a separate crate in the future. But for how we are using it now, this will make the code easier to navigate and work on.

kvark:
@csherratt While do I accept the vote (meaning it's 2:1 at the moment), I can't do it silently.

I don't see how this change makes it easier to navigate and work with. It's adding another level of depth for the stuff we care. It moves most of the whole code base for reasons that can disappear in the blink of an eye (by `gfx_macros` out, which I'd like to have, instead of this solution).

bvssvni:
Just my two cents: I prefer one crate per repo, all modules under `src/`, no sub folders, and breaking up modules if they contain generic names (if one is looking for X, there should be an x.rs). The cost of navigation outweighs the cost of maintaining separate repos. When working cross-repos, use "~/.cargo/config" to override.

kvark:
thanks @bvssvni ! Your opinion makes it 2:2. What does @cmr think?

cmr:
I agree with @kvark and @bvssvni.

kvark:
Closing by vote. Sorry @bjz ! I assure you this is for the better.

PR #608, opened 2015-02-28
Rewrote shader_param expansion using pure AST

### LINKED ISSUE #606: Using shader parameters with generic code
See https://github.com/PistonDevelopers/gfx_graphics/issues/139

The problem is that one has to pass an object implementing `Resources` to the `#[shader_param(R)]`. In the cube example, this is done by `use gfx_device_gl::GlResources as R;`.

To use it with generic code, the parameter would have to be generic or allowed to be concrete (not depending on `Resources`).

kvark:
Fixes #606


kvark:
Argh, I can't do that, since my fork is outdated. Will try from a local branch.

Issue #607, opened 2015-02-28
Rewrite `ShaderParam` definition via AST instead of `TraitDef` helper


kvark:
https://github.com/gfx-rs/gfx-rs/blob/master/src/gfx_macros/shader_param.rs#L312

We are using `TraitDef` to generate the implementation for `ShaderParam`. It's not really a derivation of the implementation, so we are misusing `TraitDef`. As a result, we can't properly work with generic parameters on the struct: `struct TestParam<R: Resources>`, which is required for #606 .

What we can do is generate the AST directly upon macro expansion. Not sure how difficult it's going to be but certainly doable.

cc @bjz @bvssvni @Potpourri

kvark:
Fixed by #609

PR #609, opened 2015-02-28
Rewrote #[shader_param] expansion using the AST rather than using generic deriving types

### LINKED ISSUE #606: Using shader parameters with generic code

kvark:
Fixes #606, #607 

Allows using generic `gfx::Resources` bounded type for `ShaderParam`, which finalizes our device abstraction. Examples:

```rust
struct MyParam<R: gfx::Resources> {
    value: i32,
    texture: gfx::TextureParam<R>,
}
struct MyParam2<R2: gfx::Resources> {
    value: f32,
    _marker: std::marker::PhantomData<R2>,
}
```

### DIFF
--- a/src/gfx_macros/shader_param.rs
+++ b/src/gfx_macros/shader_param.rs
@@ -1,4 +1,4 @@
-// Copyright 2014 The Gfx-rs Developers.
+// Copyright 2015 The Gfx-rs Developers.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


kvark:
Examples:
```rust
struct MyParam<R: gfx::Resources> {
    value: i32,
    texture: gfx::TextureParam<R>,
}
struct MyParam2<R2: gfx::Resources> {
    value: f32,
    _dummy: std::marker::PhantomData<R2>,
}
```

PR #618, opened 2015-03-02
Replaced type aliases with structs

bvssvni:
Closes https://github.com/gfx-rs/gfx-rs/issues/616

### DIFF
--- a/src/device/lib.rs
+++ b/src/device/lib.rs
@@ -167,7 +167,7 @@ impl<R: Resources, T> BufferHandle<R, T> {
 
     /// Get the underlying name for this BufferHandle
     pub fn get_name(&self) -> <R as Resources>::Buffer {
-        self.raw.name
+        self.raw.get_name()
     }
 
     /// Get the underlying raw Handle
@@ -177,7 +177,7 @@ impl<R: Resources, T> BufferHandle<R, T> {
 
     /// Get the associated information about the buffer
     pub fn get_info(&self) -> &BufferInfo {
-        &self.raw.info
+        self.raw.get_info()
     }
 
     /// Get the number of elements in the buffer.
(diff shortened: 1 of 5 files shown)


src/device/lib.rs
```diff
@@ -210,21 +190,65 @@ impl<R: Resources, T> BufferHandle<R, T> {
 }
 
 /// Raw (untyped) Buffer Handle
-pub type RawBufferHandle<R: Resources> = Handle<<R as Resources>::Buffer, BufferInfo>;
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub struct RawBufferHandle<R: Resources> {
+    /// The name of the raw buffer.
+    pub name: <R as Resources>::Buffer,
```
kvark: We can't have this `pub`, because it would allow 3rd parties to construct bogus handles. `info` can still be public, this would make `get_info()` irrelevant.

kvark:
This is a tricky PR - more code in exchange for better error messages about handles.

Named fields are no-no here, because they serve no goal: handles are constructed once and accessed via getters anyway (`get_link`). If we change all the handle type aliases by their own structs, this would make errors cleaner but would require `get_info()` and `get_name` to be implemented for each and every handle...

@bjz @csherratt @cmr - any ideas?

csherratt:
@kvark I think better error messages are worth the extra code. The library is for the users after all.

kvark:
With @csherratt vote, and mine undecided, we have 2.5 : 0.5 now, pretty much decided positive.
@bvssvni - please proceed with stripping the named fields from these structs and resolving the merge conflicts (sorry!).

src/render/batch.rs
```diff
@@ -81,7 +81,7 @@ impl<'a, T: ShaderParam> Batch for ImplicitBatch<'a, T> {
 
     fn get_data(&self) -> Result<BatchData<T::Resources>, Error> {
         let (mesh, ref slice, program, _, state) = *self;
-        match mesh::Link::new(mesh, program.get_info()) {
+        match mesh::Link::new(mesh, &program.get_info()) {
```
kvark: I don't understand. Why did you need an `&` here? `get_info` should return a reference.

kvark:
@bvssvni the unnecessary borrows are all over `batch.rs` and one in `render/lib.rs`. You could leave these files untouched.

kvark:
It looks complete now. Leaving for others to merge, since I abstained from the vote.

Issue #620, opened 2015-03-02
Rename macro variables to something nicer in "render/shade.rs"


bvssvni:
https://github.com/gfx-rs/gfx-rs/blob/master/src/render/shade.rs#L34

Suggestion: `src_ty` and `dst_ty`

Edit: And arrow `impl_ToUniform!(i32 => UniformValue::I32);` might be nice!

kvark:
:+1:

Issue #619, opened 2015-03-02
Move `MeshError` from "render/batch.rs" to "render/mesh.rs"


bvssvni:
Perhaps name it to `mesh::Error`?

`BatchError` could be named `batch::Error`.

bvssvni:
- [ ] `CreateShadeError` => `shade:Error`
- [ ] `BatchError` => `batch::Error`
- [ ] `MeshError` => `mesh::Error`

kvark:
Sounds good!

kvark:
Fixed by #624. `CreateShaderError` will live a bit longer, I suppose.

Issue #614, opened 2015-03-01
Add lifetime parameter to `Resources` for associated types


bvssvni:
https://github.com/PistonDevelopers/gfx_graphics/blob/master/src/back_end.rs#L198-L208

```
pub struct GfxGraphics<'a, R, C, O>
    where R: gfx::Resources + 'a,
          C: gfx::CommandBuffer<R> + 'a,
          O: gfx::Output<R> + 'a,
          R::Buffer: 'a,
          R::ArrayBuffer: 'a,
          R::Shader: 'a,
          R::Program: 'a,
          R::FrameBuffer: 'a,
          R::Surface: 'a,
          R::Texture: 'a,
          R::Sampler: 'a
{
   ...
}
```

The compiler doesn't know that the associated types outlives `Resources` in every context they are used, and these constraints must be added even if they are never referred to anywhere. By adding a lifetime input parameter to `Resources<'a>` this can be annotated directly on the trait.

bvssvni:
Hmm... this might propagate upward to add a lifetype parameter to `Device`.

bvssvni:
This could be:

```Rust
pub struct GfxGraphics<'a, D>
    where
        D: 'a + gfx::Device<'a>
{
   ...
}
```

kvark:
@bvssvni I certainly agree this looks bizarre. What it not clear to me though is why `GfxGraphics` is the only place I've seen that had to constraint all the lifetimes. Perhaps, there is an easy way to refactor `GfxGraphics` that solves it?

kvark:
Not happened in pre-ll

PR #624, opened 2015-03-03
Batch improvements

### LINKED ISSUE #622: Better batch binding
We are over-using tuples a bit too much:
```rust
impl<'a, T: ShaderParam> Batch for (&'a mesh::Mesh<T::Resources>, mesh::Slice<T::Resources>,
&'a ProgramHandle<T::Resources>, &'a T, &'a DrawState)  {...}

impl<'a, T: ShaderParam> Batch for (&'a RefBatch<T>, &'a T, &'a Context<T::Resources>) {}
```

No, it doesn't look good on the user side either. And the error messages don't help you to figure out what's wrong. It would just say that your tuple doesn't implement `Batch` but will not answer, why.

First we need to use name-less structs here:
```rust
RefBatchFull<'a, T: ShaderParam>(&'a RefBatch<T>, &'a T, &'a Context<T::Resources>);
```

Second, we need to provide helper functions to construct them on site:
```rust
renderer.draw(context.bind(&batch, &data), &frame).unwrap();
```
instead of:
```rust
renderer.draw(&(&batch, &data, &context), &frame).unwrap();
```

kvark:
Closes #622, #620, #619

This PR doesn't break any API yet, because I was unable to use structures (lifetimes issues #614).
Also, I couldn't make `Context::bind()` to return a reference to the temporal batch. Perhaps, a task for future PRs.

### DIFF
--- a/src/render/batch.rs
+++ b/src/render/batch.rs
@@ -27,44 +27,15 @@ use render::mesh::ToSlice;
 use shade::{ParameterError, ShaderParam};
 use render::state::DrawState;
 
-/// An error with a defined Mesh.
-#[derive(Clone, Debug, PartialEq)]
-pub enum MeshError {
-    /// A required attribute was missing.
-    AttributeMissing(String),
-    /// An attribute's type from the vertex format differed from the type used in the shader.
-    AttributeType,
-    /// Internal error due to mesh link limitations
-    MeshLink(mesh::LinkError),
-}
-
 /// An error occurring at batch creation
 #[derive(Clone, Debug, PartialEq)]
-pub enum BatchError {
+pub enum Error {
     /// Error connecting mesh attributes
-    Mesh(MeshError),
+    Mesh(mesh::Error),
     /// Error connecting shader parameters
     Parameters(ParameterError),
     /// Error context is full
     ContextFull,
-}
-
-/// Match mesh attributes against shader inputs, produce a mesh link.
-/// Exposed to public to allow external `Batch` implementations to use it.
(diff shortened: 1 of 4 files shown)


src/render/batch.rs
```diff
@@ -21,52 +21,22 @@ use std::num::from_uint;
...
-    MeshLink(mesh::LinkError),
-}
-
 /// An error occurring at batch creation
 #[derive(Clone, Debug, PartialEq)]
-pub enum BatchError {
+pub enum Error {
     /// Error connecting mesh attributes
-    Mesh(MeshError),
+    Mesh(mesh::Error),
```
csherratt: I like this.

src/render/mesh.rs
```diff
@@ -226,15 +230,33 @@ pub struct Link {
 }
 
 impl Link {
+    /// Match mesh attributes against shader inputs, produce a mesh link.
+    /// Exposed to public to allow external `Batch` implementations to use it.
+    pub fn new<R: Resources>(mesh: &Mesh<R>, pinfo: &device::shade::ProgramInfo)
+                             -> Result<Link, Error> {
+        let mut indices = Vec::new();
+        for sat in pinfo.attributes.iter() {
+            match mesh.attributes.iter().enumerate()
```
csherratt: I realize this was moving of code. But is this no N^2?

kvark: It is `N^2`, yes, and we are doing it when constructing a batch, so it doesn't happen during rendering.

csherratt:
It's harder to judge how the change effects code since the examples are not in `gfx-rs` any more. But otherwise this looks fine.

kvark:
See the examples in https://github.com/gfx-rs/gfx_examples/pull/4

PR #628, opened 2015-03-04
Split of referrence batches into Core/non-Core

kvark:
(This is a breaking change, please review carefully)

I noticed a couple of things that weren't sound:
  1. `RefBatch` carried a publicly accessible `slice`, while not carrying the parameters. The reason for the former was "convenience" (IIRC), the reason for the latter was "just because we can". Hence, this type was neither complete nor minimal.
  2. A lot of applications tend to keep a parameter structure together with a `RefBatch`. They are often seen together, and that wasn't convenient to do.

As a solution, I made a `RefCore` batch that only carries minimal (for safety) sealed data, nothing is available to public. The `RefBatch` now is rather open - it carries the core as well as a slice (like it used to) and a parameter structure, all mutable by public.

The idea is - `RefCore` to be used for maximum flexibility, like creating a batch when not yet having a parameter structure at hands, or using it with multiple parameters/slices. While `RefBatch` is just a convenient thing to have. The maintenance cost is minimal, since they are re-using lots of code.

`RefBatch` is also nice for ergonomic reasons - you can create a batch with type inference.

### DIFF
--- a/src/device/lib.rs
+++ b/src/device/lib.rs
@@ -191,11 +191,19 @@ impl<R: Resources, T> BufferHandle<R, T> {
 }
 
 /// Raw (untyped) Buffer Handle
-#[derive(Copy, Clone, PartialEq, Debug)]
+#[derive(PartialEq, Debug)]
 pub struct RawBufferHandle<R: Resources>(
     <R as Resources>::Buffer,
     BufferInfo
 );
+
+impl<R: Resources> Copy for RawBufferHandle<R> {}
+
+impl<R: Resources> Clone for RawBufferHandle<R> {
+    fn clone(&self) -> RawBufferHandle<R> {
+        RawBufferHandle(self.0, self.1.clone())
+    }
+}
 
 impl<R: Resources> RawBufferHandle<R> {
     /// Creates a new raw buffer handle (used by device)
@@ -292,11 +300,19 @@ impl<R: Resources> SurfaceHandle<R> {
 }
 
 /// Texture Handle
-#[derive(Copy, Clone, PartialEq, Debug)]
+#[derive(PartialEq, Debug)]
 pub struct TextureHandle<R: Resources>(
     <R as Resources>::Texture,
     tex::TextureInfo,
 );
+
+impl<R: Resources> Copy for TextureHandle<R> {}
+
+impl<R: Resources> Clone for TextureHandle<R> {
+    fn clone(&self) -> TextureHandle<R> {
+        TextureHandle(self.0, self.1.clone())
+    }
+}
 
 impl<R: Resources> TextureHandle<R> {
     /// Creates a new texture (used by device)
@@ -311,11 +327,19 @@ impl<R: Resources> TextureHandle<R> {
 }
 
 /// Sampler Handle
-#[derive(Copy, Clone, PartialEq, Debug)]
+#[derive(PartialEq, Debug)]
 pub struct SamplerHandle<R: Resources>(
     <R as Resources>::Sampler,
     tex::SamplerInfo,
 );
+
+impl<R: Resources> Copy for SamplerHandle<R> {}
+
+impl<R: Resources> Clone for SamplerHandle<R> {
+    fn clone(&self) -> SamplerHandle<R> {
+        SamplerHandle(self.0, self.1.clone())
+    }
+}
 
 impl<R: Resources> SamplerHandle<R> {
     /// Creates a new sampler (used by device)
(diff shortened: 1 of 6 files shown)


bvssvni:
Looks good.

src/device/lib.rs
```diff
@@ -562,6 +575,17 @@ mod test {
...
+        type Buffer = ();
+        type ArrayBuffer = ();
+        type Shader = ();
+        type Program = ();
+        type FrameBuffer = ();
+        type Surface = ();
+        type Texture = ();
+        type Sampler = ();
+    }
+
```
bjz: Good stuff. Maybe impl on a `struct TestResources;`?

src/gfx/lib.rs
```diff
@@ -62,6 +62,20 @@ pub struct Graphics<D: device::Device> {
...
+        &self.context
+    }
+}
+
+impl<D: device::Device> std::ops::DerefMut for Graphics<D> {
+    fn deref_mut(&mut self) -> &mut batch::Context<D::Resources> {
+        &mut self.context
+    }
+}
+
```
bjz: Does this need to be in this PR?

src/render/batch.rs
```diff
@@ -236,66 +237,113 @@ impl<T: Clone + PartialEq> Array<T> {
 }
 
 
-/// Ref batch - copyable and smaller, but depends on the `Context`.
+/// Referrenced core - a minimal sealed batch that depends on `Context`.
```
bjz: `Referenced`

src/render/batch.rs
```diff
@@ -236,66 +237,113 @@ impl<T: Clone + PartialEq> Array<T> {
...
 /// It has references to the resources (mesh, program, state), that are held
 /// by the context that created the batch, so these have to be used together.
-pub struct RefBatch<T: ShaderParam> {
+pub struct RefCore<T: ShaderParam> {
```
bjz: What other names have you thought of? What about `RefBatchData`?

src/device/lib.rs
kvark: Agreed

src/gfx/lib.rs
kvark: This allows `make_batch`, `make_core`, and `bind` to be called from `Graphics`, instead of wrapping them up into pass-through methods.

src/render/batch.rs
kvark: I haven't really thought about names, no. `RefBatchData` doesn't sound too good because it bears the "data" term, while in fact `RefCore` can be seen as a batch type. Yes it is incomplete (requires not only the context, but also a slice and a parameter data), but neither is `RefBatch` (needs the context).

I don't insist on `RefCore` name though. My understanding is that since we don't re-export everything in `batch` module, this type is accessed as `gfx::batch::RefCore`, which makes it less ambiguous. Perhaps, you'd prefer `CoreBatch` more?

src/render/batch.rs
kvark: will fix

kvark:
One of the reasons behind the PR is also an attempt to standardize handle operations. I don't like the fact that some handles are copyable (`BufferHandle`) and others are only clonable (`ProgramHandle`). I think we should be consistent with this, and `Clone` is the common denominator. That's why `Slice` is no longer copyable, for instance.

kvark:
@stjahns this PR would improve this code in skeletal_animation:
```rust
pub struct SkinnedRenderer {
    animation_clip: AnimationClip,
    skinning_transforms_buffer: BufferHandle<GlResources, [[f32; 4]; 4]>,
    shader_params: SkinnedShaderParams<GlResources>,
    batch: RefBatch<SkinnedShaderParams<GlResources>>,
}
```
Also, I believe you really need this to be device-agnostic:
```rust
pub struct SkinnedRenderer<R: Resources> {
    animation_clip: AnimationClip,
    skinning_transforms_buffer: BufferHandle<R, [[f32; 4]; 4]>,
    batch: RefBatch<SkinnedShaderParams<R>>,
}
```

src/render/batch.rs
bjz: We can iterate on naming later, I think.

src/gfx/lib.rs
bjz: Hmm, ok - just think it might be an abuse of `Deref`, which is a debate I would rather not have on this PR.

stjahns:
Works for me  / Am I doin' it rite?

src/gfx/lib.rs
kvark: Feel free to open this debate anywhere please. I'm not too experienced with `Deref` practices and would be happy to know more.

kvark:
Self-merging as everything is fixed and the concept has been approved.

PR #631, opened 2015-03-05
Moved all the handles into a separate module

kvark:
Our `device/lib.rs` became a bit overloaded, so I decided to move the handles out.


bvssvni:
Maybe these could be reexported in "device/lib.rs"?

bvssvni:
I think we should merge this and deal with naming and reexport in separate issues.

kvark:
@bvssvni everything is re-exported in `gfx/lib.rs`. It is compatible with all the user code, just `gfx_device_gl` needs to be updated (https://github.com/gfx-rs/gfx_device_gl/pull/8)

bvssvni:
Can I merge this?

kvark:
I think so ;)

bvssvni:
Nice!

Issue #570, opened 2015-02-17
Asynchronous resource creation


kvark:
Currently the device trait has all the methods for creating and updating resources. Next-gen APIs as well as DX11 allow creating resources asynchronously:

https://msdn.microsoft.com/en-us/library/windows/desktop/ff476889%28v=vs.85%29.aspx

This is crucial for us to support. The way I see it is having a `Factory` trait that has all these methods. Then, some back-ends will just implement `Factory` for their devices, some will provide `create_factory` methods to support multiple factories operating simultaneously.

bvssvni:
This might solve the problems with glyph caching, where textures needs to be created by demand.

kvark:
@bvssvni This issue is more about DX11/next-gen, since I don't know how this is gonna work with OpenGL. The issue with glyph caching will be addressed in a more straightforward way, I just didn't happen to look into it yet (sorry!).

kvark:
I'm thinking that we could transform `Resources` to fill the factory role:
```rust
trait Device {
   type Resourcer: Resourcer + Mapper;
   type CommandBuffer: CommandBuffer<Self::Resourcer>;

   fn get_capabilities<'a>(&'a self) -> &'a Capabilities;
   fn spawn_resourcer(&self) -> Self::Resourcer;
   fn reset_state(&mut self);
   fn submit(&mut self, buffer: (&Self::CommandBuffer<Self::Resourcer>, &draw::DataBuffer));
}

trait CommandBuffer<R: Resourcer> {
   ...
}

trait Resourcer {
   type X; ...
   fn create_x() ...
   fn delete_x() ...
   fn map_x()...
}
```

kvark:
We could actually go further by splitting different resource types to be handled by different factories:
```rust
trait BufferFactory {
    type Buffer;
    fn create_buffer_raw();
    fn create_buffer<T>();
    fn create_buffer_static_raw();
    fn create_buffer_static<T: Copy>();
}
```
Aside from the architectural (decoupling!) benefit, I can see the following problems with it:

1: `BufferHandle` should be parametrized by what? We don't want users to even think `BufferFactory` exists in the first place.

Possible solution - return tuples, then have typedefs at the top level:
```rust
type BufferHandle<D: Device> = (D::Resourcer::Buffer, BufferInfo);
```
2: How to hide all the traits from the user? Solutions:
[...]

PR #632, opened 2015-03-05
Resource Factory

kvark:
(ready now)

~~Includes #631~~
Paves the way to #570 

Moves all resource-related methods into `Factory`.
Cleans up `device/lib.rs` to be perfectly readable now!


kvark:
The problem appears to come from `DeviceExt`, which is declared as follows:
```rust
pub trait DeviceExt<D: device::Device>: device::Factory<D::Resources> {..}
```
And then implemented as:
```rust
impl<D: device::Device, F: device::Factory<D::Resources>> DeviceExt<D> for F {..}
```

So, basically, I want it to be automatically implemented for any resource factory. The caveat is that it also needs a `CommandBuffer` type for `create_renderer`. One thing to try would be to implement it for any device that is also a factory. While this is more restricting, it would at least allow us to proceed with the change while not breaking everything.

Idea? Or maybe someone is willing to fix it?
cc @bjz @csherratt @bvssvni

bvssvni:
The `Ext` convention is usually used for traits that inherit the trait, so `DeviceExt` would lead users to believe it inherits `Device`.

kvark:
The inheritance itself is not important here. The fact is - `SomethingExt` gets auto-implemented for `Something` and extends the functionality using methods from `Something`. That's what we do with `DeviceExt`. The change only makes it to extend a `Factory` instead of `Device`, but given that `GlDevice` implements both, this should work for a short term.

bvssvni:
I think we all agree we want to move in this direction, so I'll merge. Looks good.

kvark:
Thanks @bvssvni ! I hope everyone had a chance to look at it.

Issue #634, opened 2015-03-07
gfx-rs build break


csherratt:
Holy errors Batman: https://gist.github.com/csherratt/a2b67022767ad132d9c0

I'm investigating.

kvark:
Fixed by #635 
Thanks!

PR #635, opened 2015-03-07
Remove ParamValue for &mut ParamStorage

csherratt:
This fixes the build break in #634

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -72,7 +72,7 @@ impl<R: Resources> RenderState<R> {
 }
 
 /// Temporary parameter storage, used for shader activation.
-struct ParamStorage<R: Resources> {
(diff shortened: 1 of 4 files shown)


src/render/mod.rs
```diff
@@ -88,15 +88,11 @@ impl<R: Resources> ParamStorage<R> {
         }
     }
 
-    fn get_mut(&mut self) -> shade::ParamValues<R> {
+    fn get_mut(&mut self) -> &mut ParamStorage<R> {
         self.uniforms.truncate(0);
```
kvark: since you are at it, you can replace these by `clear()`

kvark:
Looks pretty neat! I wonder if you need to update macros as well.

kvark:
:+1:

PR #639, opened 2015-03-10
Removed Bare trait

### LINKED ISSUE #637: Reference resources in device operations
`device::handle::Bare` trait should not be needed. Every time the device needs to access a bare handle, it should do it via `self.frame_handles.ref_something()` method, thus making sure the resource lives up to the end of the frame.

This extends #633

kvark:
Closes #637 
Requires https://github.com/gfx-rs/gfx_device_gl/pull/11


bvssvni:
Nice!

kvark:
Less hacks, more profit ;)

PR #646, opened 2015-03-12
Fixed shader values allocation hazard

### LINKED ISSUE #642: Macros are broken for WebGL
Reported by @AerialX for #641:
> `gfx_macros` is broken. So, uh, if you can maybe show me how to not use it that'd be cool. 	The terrain demo has 3 matrix uniforms. By the time it got to the `process()` loop, it would only send two of the three, and they were corrupt / bad data.

kvark:
Fixed #642


kvark:
Merging as urgent, it can't be worse.

bvssvni:
Nice!

PR #653, opened 2015-03-17
Added SubmitInfo

kvark:
Allows user applications to capture `Renderer::as_buffer` if needed.

### DIFF
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -218,6 +218,13 @@ pub trait Factory<R: Resources> {
     fn cleanup(&mut self);
 }
 
+/// All the data needed simultaneously for submitting a command buffer for
+/// execution on a device.
+pub type SubmitInfo<'a, D: Device> = (
+    &'a D::CommandBuffer,
+    &'a draw::DataBuffer,
+    &'a handle::Manager<D::Resources>
+);
 
 /// An interface for performing draw calls using a specific graphics API
 pub trait Device {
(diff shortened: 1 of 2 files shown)


kvark:
Merging as harmless

src/device/mod.rs
```diff
@@ -218,6 +218,13 @@ pub trait Factory<R: Resources> {
     fn cleanup(&mut self);
 }
 
+/// All the data needed simultaneously for submitting a command buffer for
+/// execution on a device.
+pub type SubmitInfo<'a, D: Device> = (
+    &'a D::CommandBuffer,
+    &'a draw::DataBuffer,
+    &'a handle::Manager<D::Resources>
+);
```
bjz: Why the type alias?

src/device/mod.rs
```diff
@@ -233,11 +240,7 @@ pub trait Device {
     fn reset_state(&mut self);
 
     /// Submit a command buffer for execution
-    fn submit(&mut self, (
-        &Self::CommandBuffer,
-        &draw::DataBuffer,
-        &handle::Manager<Self::Resources>
-    ));
+    fn submit(&mut self, SubmitInfo<Self>);
```
bjz: Couldn't this just be a list of args?

kvark: It would be inconvenient, since you wouldn't be able to do `device.submit(renderer.as_buffer())` any more.

src/device/mod.rs
kvark: I don't mind having a `struct` there, just wanted some type quickly.

src/device/mod.rs
bjz: Ah. Then why aren't you using the typedef in the return type of `as_buffer`?

kvark: I simply missed it, and it compiled just fine.

Issue #656, opened 2015-03-26
Callback for the shader/program info log


kvark:
An optional callback would allow to save on querying the log if the user doesn't care, and to use the log at run-time in case the user needs it.

kvark:
After a discussion with Layl on IRC, we came to a solution:
  - keep shader handles alive inside the program
  - don't query the logs during compilation/linkage
  - have separate methods for querying the logs

PR #657, opened 2015-03-27
Add log parameter to link_program

mninja:
As per the discussion on gitter about #656, here's a solution using Option<&mut String>.

### DIFF
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -184,7 +184,7 @@ pub trait Factory<R: Resources> {
     fn create_array_buffer(&mut self) -> Result<handle::ArrayBuffer<R>, ()>;
     fn create_shader(&mut self, stage: shade::Stage, code: &[u8]) ->
                      Result<handle::Shader<R>, shade::CreateShaderError>;
-    fn create_program(&mut self, shaders: &[handle::Shader<R>], targets: Option<&[&str]>)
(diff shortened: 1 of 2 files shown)


kvark:
Thanks for the PR!

Ok, `create_program` and `link_program` look good. But what about shaders? Often, the error you'd really want to see is the one from shader compilation, not the linkage.

Sorry for making this concern so late, it only hit me now.

mninja:
That's a good point, but then how should it work? The same buffer can be extended and reused but then there's no easy way to differentiate where the errors come from.

kvark:
Sorry, still don't have the answer. The simplest way would be to pass `Options<&mut [String]>`, where the first N strings will match give N shaders, and the last one is for the program.

kvark:
Actually, there is a better way. We can abstract it like this:
```rust
trait ShaderErrorSink {
   fn compile(&mut self, &str);
   fn link(&mut self, &str);
}

impl ShaderErrorSink for () {...} //do nothing
impl ShaderErrorSink for String {...} //accumulate
impl ShaderErrorSink for &mut [String] {...} //keep separate logs
```
This is flexible, allows less noisy user syntax (`()` instead of `None`, and just something instead of `Option<something>`).

mninja:
But then we would always query the log, or is there to match on the impl of ShaderErrorSink?

kvark:
Ah, bummer :(

Regards,
Dzmitry

bvssvni:
Closing because outdated.

PR #658, opened 2015-03-27
Update to latest rust

Potpourri:
Add necessary lifetime bound and remove unnecessary cast.

### DIFF
--- a/src/device/mapping.rs
+++ b/src/device/mapping.rs
@@ -31,7 +31,9 @@ pub trait Raw {
 }
 
 /// A handle to a readable map, which can be sliced.
-pub struct Readable<'a, T: Copy, R: 'a + Resources, F: 'a + Factory<R>> {
+pub struct Readable<'a, T: Copy, R: 'a + Resources, F: 'a + Factory<R>> where
+    F::Mapper: 'a
+{
     raw: F::Mapper,
     len: usize,
     factory: &'a mut F,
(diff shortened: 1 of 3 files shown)


src/device/mapping.rs
```diff
@@ -122,22 +130,24 @@ impl<'a, T: Copy, R: Resources, F: Factory<R>> Drop for RW<'a, T, R, F> where
 /// A service trait with methods for mapping already implemented.
 /// To be used by device back ends.
 #[allow(missing_docs)]
-pub trait Builder<R: Resources> {
+pub trait Builder<'a, R: Resources> {
     type RawMapping: Raw;
 
-    fn map_readable<T: Copy>(&mut self, Self::RawMapping, usize)
+    fn map_readable<T: Copy>(&'a mut self, Self::RawMapping, usize)
                     -> Readable<T, R, Self>;
```
kvark: shouldn't this be using `'a` as well?

kvark:
Thanks @Potpourri ! Just one small thing to address.

src/device/mapping.rs
Potpourri: You mean `Readable<'a, T, R, Self>`? If yes then shouldn't, rust inference lifetime from self.

kvark: Right. But. I don't know what rust guidelines tell, but I'd expect the lifetimes to either be fully specified, or fully inferred. Doing it half-way seems to be a bit confusing.

Anyway, not a blocked by any means. Merging this.

PR #659, opened 2015-03-28
Switched to use github version of draw_state

kvark:
This is consistent with how we use `gfx-rs` from `gfx_device_gl` and other projects.


kvark:
Merging as this is blocking me and no one is here.

PR #660, opened 2015-03-29
Remove unused arg from VertexFormat::generate

bjz:
This was needed before we had the `<T as Trait>::method` syntax.


kvark:
Oh neat, thanks!

PR #662, opened 2015-04-01
from_raw_parts -> from_raw_parts_mut

stjahns:


kvark:
Thanks! I was just doing the same ;)

kvark:
Actually, this crashes on `DataBuffer::add_vec()`. I'll make a patch shortly.

PR #664, opened 2015-04-02
Fixed add_vec() crash. Derived Eq for a bunch of stuff.

kvark:


kvark:
Merging as harmless and essential.

Issue #666, opened 2015-04-03
Break the link between Device and Factory


kvark:
These two are conceptually different things on the abstract device level, however they are merged in:
  1. `gfx_device_gl::GlDevice` implements both, which seemed reasonable at first to get less porting headache
  2. `create_renderer` needs both, for some reason I can't see right now
  3. growing amount of user code, including my own `gfx_scene`, that makes me worry

I propose to have a separate `GlFactory` thing, and also (since we are at it...) remove the `Gl` prefix from `GlDevice` (and `GlFactory`, of course). It needs to implement `Factory`, obviously. 
We might have a single `gfx_device_gl::create()` function returning a `(Device, Factory)` tuple, instead of the current `GlDevice::new`. Both need to be non-sendable for GL.

Having them as separate objects, one can write functions that accept both as mutable references (cc @stjahns). The added inconvenience damage should be partially covered by `Graphics` being able to hold these things:
- `window` - an abstract handle to the window that has the context (part of #529). Holding it would allow automatic `swap_buffers` call at the end of the frame.
- `device` - the execution unit for the command buffers
- `factory` - the resource factory, with `cleanup` also called automatically at the end of the frame
- `renderer` - command buffer constructor
- `main_frame` - default framebuffer (related to #530). Unfortunately, it'd had to be cloned before usage, since the methods we call have `&mut self`, so we wouldn't be able to pass `&graphics.main_frame` in them, unless we make wrappers to work with the main frame only (which is not a bad idea for ergonomics!)
- `context` - the batch context - for the convenience

As a middle ground, we could *just* have these wrappers that draw to the main frame in `Graphics`, and expose the `renderer` publicly for the non-simple calls. This way, the simple cases are simple, and there is no abuse of `DerefMut` implementations (cc @bjz).

I hope I described this clear enough. We will be cursed eventually anyway (hell, it's #666 after all) for breaking this interface, but better now than later. Objections? cc @csherratt, @bvssvni

bvssvni:
I think separating `Device` and `Factory` is a good idea!

I have some concerns about the changes to `Graphics`: When using Piston you have to do some glue code because of the separation between window and graphics. Example https://github.com/PistonDevelopers/gfx_graphics/blob/master/examples/image_test.rs#L37-L67

Swapping the buffer within the `Graphics` object would lead to swapping it twice and dealing with two window abstractions, one inside the event iterator and another inside the game loop, making one usage incompatible with another.

[...]

PR #668, opened 2015-04-03
Factory

kvark:
First step towards #666 
This PR splits the `Factory` on the abstract gfx level. Another fix is needed on the device implementation side (https://github.com/gfx-rs/gfx_device_gl/pull/23). And I can't fix `gfx_examples` due to both `glfw-rs` and `glutin` being broken at this moment.

Please review carefully!

There is also an abstract window now, required for proper `Graphics` initialization. We'd need to implement it for `glutin`/`glfw`/`sdl` in separate crates (#529) in order to allow people to transition.

Use API changes:
  1. `into_graphics` is now called on the `(Device, Factory)` tuple.
  2. `Graphics` now explicitly contains `Factory`.
  3. `create_renderer` is implemented for the `Factory`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -69,3 +69,14 @@ pub use device::handle::Sampler as SamplerHandle;
 
 pub mod render;
 pub mod device;
+
+
+/// An abstract window that has the graphics context.
+pub trait Window {
+    /// Get width and height of the render area.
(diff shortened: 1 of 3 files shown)


kvark:
A separate PR will try to integrate the windowing stuff, if any.

PR #670, opened 2015-04-04
Compilation fixes

nicholasbishop:
`cargo test` succeeds for me with these changes, tested with `rustc 1.0.0-nightly (d17d6e7f1 2015-04-02) (built 2015-04-03)`

### DIFF
--- a/src/render/batch.rs
+++ b/src/render/batch.rs
@@ -160,9 +160,10 @@ impl<T: ShaderParam> Batch for OwnedBatch<T> {
 type Index = u16;
 
 //#[derive(PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[derive(Clone)]
 struct Id<T>(Index, PhantomData<T>);
 
-impl<T> Copy for Id<T> {}
+impl<T: Clone> Copy for Id<T> {}
 
 impl<T> Id<T> {
     fn unwrap(&self) -> Index {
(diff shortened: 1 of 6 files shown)


src/render/batch.rs
```diff
@@ -160,9 +160,10 @@ impl<T: ShaderParam> Batch for OwnedBatch<T> {
 type Index = u16;
 
 //#[derive(PartialEq, Eq, PartialOrd, Ord, Debug)]
+#[derive(Clone)]
 struct Id<T>(Index, PhantomData<T>);
 
-impl<T> Copy for Id<T> {}
+impl<T: Clone> Copy for Id<T> {}
```
bjz: Could you instead do:
~~~rust
impl<T> Clone for Id<T> {
    fn clone(&self) -> Id<T> { *self }
}
~~~

src/render/batch.rs
```diff
@@ -252,7 +254,7 @@ pub struct CoreBatch<T: ShaderParam> {
     state_id: Id<DrawState>,
 }
 
-impl<T: ShaderParam> Copy for CoreBatch<T> where T::Link: Copy {}
+impl<T: Clone + ShaderParam> Copy for CoreBatch<T> where T::Link: Copy {}
```
bjz: I assume the need for the `Clone` bound would go away after my previous comment.

bjz:
Thanks! A few comments. I would like to get this merged asap.

bjz:
Actually, I'll merge now and make the changes after.

nicholasbishop:
Thanks for the quick merge. Just for my own edification, could you explain a bit about why the explicit `Clone` impl is right way to go here? I'm still mostly at the stage of making blind edits when the compiler yells at me, so I'd be interested to hear a few words about the bigger picture.

bjz:
No worries! Syntax extensions like `#[derive(..)]` have no knowledge of the types. So `derive` just blindly adds `Clone` bounds to all the type parameters. In our case, `Clone` does not need to be added to `T`, because it does not need to be cloned. Dunno if that makes sense...

kvark:
Thanks @nicholasbishop !

nicholasbishop:
Yep, that indeed makes sense. Thanks for the explanation :)

PR #672, opened 2015-04-04
adding set_info_kind function to TextureHandle

lynks--:
Intended for use with the `TextureCube` TextureKind, when populating all 6 sub-textures for example:
```
		for side in [CubeFace::PosY, CubeFace::NegY, CubeFace::NegX,
					 CubeFace::NegZ, CubeFace::PosX, CubeFace::PosZ].iter() {

			let env_data = cubemap_file.read_exact(env_resolution * env_resolution * 3).unwrap();

			env_tex.set_info_kind(gfx::tex::TextureKind::TextureCube(*side));

			self.rs.device.update_texture(&env_tex, &env_ii, &env_data);
		}
```

### DIFF
--- a/src/device/handle.rs
+++ b/src/device/handle.rs
@@ -149,11 +149,6 @@ pub struct Texture<R: Resources>(Arc<R::Texture>, tex::TextureInfo);
 impl<R: Resources> Texture<R> {
     /// Get texture info
     pub fn get_info(&self) -> &tex::TextureInfo { &self.1 }
-	/// Update texture info kind
-	pub fn set_info_kind(&mut self, kind: tex::TextureKind) {
-		// TODO mutable handles seems wrong.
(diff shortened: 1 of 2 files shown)


bvssvni:
Fix the indention level!

bjz:
Thanks! Please fix the indentation level as @bvssvni said. I will defer to @kvark as to whether he thinks this is a good idea. Perhaps there might be a better way to support what you want to do?

kvark:
@lynks-- Please follow my original recommendation with `Option<TextureKind>` parameter. I don't think having mutable handles is a good idea: we are bind-less, no need to introduce more states where there are simpler ways.

lynks--:
@kvark agreed and your preferred solution in the works. Sorry about the indentation.

PR #673, opened 2015-04-04
Fix compilation with fresh build of rustc.

vickenty:
Fixes compilation errors on rustc 1.0.0-nightly (d17d6e7f1 2015-04-02) (built 2015-04-03).

* Copy trait requires Clone to be implemented or derived.
* -1 as a u8 value seems to be deprecated, can be replaced with a call on a type.


vickenty:
Seems to be already fixed in master, too late. :)

PR #674, opened 2015-04-04
Optional TextureKind Paramter to Texture Update

lynks--:


kvark:
Looks good! Please squash it into a single commit.

kvark:
Why did you use `Option<&tex::TextureKind>` instead of `Option<tex::TextureKind>`? I believe the indirection is not needed there, the type implements `Copy`.

lynks--:
Yeah as you can see I got into a mess re commits after losing track. I'll rebase and squash. Also, yep not sure why I used a reference like that, I'll remove it. Come back in 20 mins! :)

kvark:
Thanks!

PR #680, opened 2015-04-10
sRGB support

### LINKED ISSUE #595: Add helper methods to texture/surface creation
Texture (or render target) creation may be a little ergonomic hell in itself.

We have `RGBA8` format pre-defined, we need to add more (obviously), but that's not enough. We could have `create_texture_rgba8` or `create_texture_depth` method that would really hide all the complexity for newcomers until they need more power.

We can have these methods as a part of `DeviceExt`.

### LINKED ISSUE #679: sRGB support
Required for gamma-correct processing. We need sRGB internal formats for the textures and render surfaces.

kvark:
Closes #679 
Closes #595


bvssvni:
Looks good! Merging.

PR #682, opened 2015-04-11
Frame abstraction and safety

### LINKED ISSUE #681: Make the main frame to be safe
1. Disallow creation by the user
  2. Disallow extraction/cloning of the attached surfaces
  3. Inherit the properties of the windowing system: 
    - size
    - MSAA
    - depth/stencil
    - sRGB

Since the current `Frame` is just a blank state, we can't just attach some (dummy) planes to it and call it a day, because that would fail p2.

Hence, polymorphism can't be achieved by mutating the struct. The only logical solution would be to use the `Frame` as a trait, thus leaving all mutation methods to be a part of the particular struct implementing it. The main frame will be absolutely opaque then. Going this route, we could also have `Frame` implemented for `Texture`, `Surface`, or even their combinations. This still leaves a lot of unknowns, but would be better than the void we have now anyway ;)

kvark:
Closes #681 
GL device: https://github.com/gfx-rs/gfx-rs/pull/682
Examples: https://github.com/gfx-rs/gfx_examples/pull/15
Window: https://github.com/gfx-rs/gfx_window_glutin/pull/1

### DIFF
--- a/src/render/target.rs
+++ b/src/render/target.rs
@@ -133,9 +133,8 @@ pub struct Frame<R: Resources> {
 }
 
 impl<R: Resources> Frame<R> {
-    /// Create an empty `Frame`, which corresponds to the 'default framebuffer',
-    /// which renders directly to the window that was created with the OpenGL context.
-    pub fn new(width: Size, height: Size) -> Frame<R> {
+    /// Create an empty `Frame`.
+    pub fn empty(width: Size, height: Size) -> Frame<R> {
         Frame {
             width: width,
             height: height,
--- a/src/render/target.rs
+++ b/src/render/target.rs
@@ -17,7 +17,9 @@
 use std::fmt::Debug;
 use device;
 use device::Resources;
+use device::tex::Size;
 use draw_state::target::{Layer, Level, Mask};
+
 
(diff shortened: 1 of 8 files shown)


src/render/target.rs
```diff
@@ -29,22 +31,96 @@ pub enum Plane<R: Resources> {
...
+    /// Get canvas dimensions.
+    fn get_size(&self) -> (Size, Size);
+    /// Get array of color planes.
+    fn get_colors(&self) -> &[Plane<R>] { &[] }
+    /// Get depth plane, if any.
+    fn get_depth(&self) -> Option<&Plane<R>> { None }
+    /// Get stencil plane, if any.
+    fn get_stencil(&self) -> Option<&Plane<R>> { None }
+    /// Check if it converts gamma of the output colors.
+    fn does_convert_gamma(&self) -> bool { false }
```
bvssvni: Maybe we could use `Gamma` here?

kvark: Done

bvssvni:
Looks good! Merging.

kvark:
Thanks! I'll fill up the PR description to post an update nicely.

PR #684, opened 2015-04-12
Canvas

kvark:
Canvas is a convenient wrapper around a window, device, factory, and a renderer. It allows you to draw stuff with sufficiently less parameters, while still treating `Batch` generically (unlike `Graphics`). It is meant to work with `gfx_window_*` crates for minimal boilerplate initialization. The `present` method will take care of the buffer swapping, renderer reset, and resource cleanups, thus leaving no chance to miss anything.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.2.6"
+version = "0.2.7"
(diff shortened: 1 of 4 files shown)


kvark:
Actually, I shouldn't have switched `IntoCanvas` to work with `Result`. Doing another PR now.

PR #685, opened 2015-04-12
Canvas - take 2

### LINKED ISSUE #529: Implement GLFW/Glutin bridges as separate crates

### LINKED ISSUE #678: Window abstraction
We need this, because:
  - `swap_buffers()`  can be conveniently paired with `device.after_frame()` and `factory.cleanup()`. It can be done by `Graphics`, which needs to know about the window (see #666)
  - basis for the main frame (see #681)
  - basis for glutin/glfw/sdl integration (see #529)

There is an argument over Piston compatibility, which doesn't like our `Graphics` holding a window. Further discussion may be here.

kvark:
Closes #529
Closes #678 
Window Glutin: https://github.com/gfx-rs/gfx_window_glutin/pull/4
Window GLFW: https://github.com/gfx-rs/gfx_window_glfw/pull/2
Examples: https://github.com/gfx-rs/gfx_examples/pull/16

Canvas is a convenient wrapper around a window, device, factory, and a renderer. It allows you to draw stuff with sufficiently less parameters, while still treating `Batch` generically (unlike `Graphics`). It is meant to work with `gfx_window_*` crates for minimal boilerplate initialization. The `present` method will take care of the buffer swapping, renderer reset, and resource cleanups, thus leaving no chance to miss anything.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.2.6"
+version = "0.2.7"
 description = "A high-performance, bindless graphics API"
 homepage = "https://github.com/gfx-rs/gfx-rs"
 repository = "https://github.com/gfx-rs/gfx-rs"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@ extern crate draw_state;
 /// public re-exported traits
 pub mod traits {
     pub use device::{Device, Factory};
+    pub use render::ext::canvas::IntoCanvas;
     pub use render::ext::device::DeviceExt;
     pub use render::ext::factory::{RenderFactory, FactoryExt};
     pub use render::mesh::ToSlice;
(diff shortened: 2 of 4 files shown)


bvssvni:
Looks good! Merging.

kvark:
Thanks @bvssvni ! We can (or even have to?) start looking into the integration with Piston now.

Issue #676, opened 2015-04-07
Rust 1.0 compatibility


kvark:
Currently unresolved problems:
- [x] syntax extensions (`gfx_macros`)
    - find a way to program it with simple macros
- [x] `Arc` needs an `is_unique` of sorts
    - change to `Rc`, at least this one should work
- [x] `from_u32`
- [x] `ref_slice`

KingOfThePirates:
Is this related or should I make a new issue?

Output:
```bash
me@me-desktop ~/projects/mypiston2 $ cargo run -v
   Compiling gfx_macros v0.1.10 (https://github.com/gfx-rs/gfx_macros#22981219)
     Running `rustc /home/me/.cargo/git/checkouts/gfx_macros-20b8916c81d6b4ed/master/src/lib.rs --crate-name gfx_macros --crate-type dylib -C prefer-dynamic -g -C metadata=996769dbcfbdc64a -C extra-filename=-996769dbcfbdc64a --out-dir /home/me/projects/mypiston2/target/debug/deps --emit=dep-info,link -L dependency=/home/me/projects/mypiston2/target/debug/deps -L dependency=/home/me/projects/mypiston2/target/debug/deps -Awarnings`
       Fresh rustc-serialize v0.3.12
       Fresh libc v0.1.5
       Fresh shader_version v0.0.6 (https://github.com/pistondevelopers/shader_version#14aa73be)
       Fresh khronos_api v0.0.5
       Fresh bitflags v0.1.1
   Compiling gl_generator v0.0.23
     Running `rustc /home/me/.cargo/registry/src/github.com-1ecc6299db9ec823/gl_generator-0.0.23/lib.rs --crate-name gl_generator --crate-type lib -g -C metadata=82ab4c4fe082fae1 -C extra-filename=-82ab4c4fe082fae1 --out-dir /home/me/projects/mypiston2/target/debug/deps --emit=dep-info,link -L dependency=/home/me/projects/mypiston2/target/debug/deps -L dependency=/home/me/projects/mypiston2/target/debug/deps --extern log=/home/me/projects/mypiston2/target/debug/deps/liblog-54cf393d3c69686f.rlib --extern xml=/home/me/projects/mypiston2/target/debug/deps/libxml-3e674a70152e28f6.rlib --extern khronos_api=/home/me/projects/mypiston2/target/debug/deps/libkhronos_api-4b6881d34c5ba88e.rlib -Awarnings`
[...]

PR #689, opened 2015-04-15
Rust beta/1.0 compatibility

kvark:
Makes it actually compile on rust-beta, fixing some of #676 

Note: we could use some of these fixes on master as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 language: rust
+rust:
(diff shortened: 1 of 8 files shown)


kvark:
The failed build is inevitable - resource cleanup is simply disabled for the lack of better solution, so the test fails.

kvark:
I wonder how we can have the docs for both master and rust-1.0 branches...

PR #692, opened 2015-04-18
Improved texture creation helpers

kvark:
Also, fixing this error with the previous helper:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Texture(Invalid data size provided to update the texture, expected size 0)', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libcore/result.rs:750

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.2.7"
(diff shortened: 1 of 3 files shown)


kvark:
Merging since it is fairly straightforward, and my work depends on it now.

PR #695, opened 2015-04-22
Shader compile/link logging

### LINKED ISSUE #656: Callback for the shader/program info log

kvark:
Closes #656 
GL part: https://github.com/gfx-rs/gfx_device_gl/pull/30

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,15 +25,6 @@ notifications:
     on_success: always  # options: [always|never|change] default: always
     on_failure: always  # options: [always|never|change] default: always
     on_start: false     # default: false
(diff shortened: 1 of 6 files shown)


kvark:
There is a problem with this idea... Once a shader is failed to create, we don't have the handle to pass into `get_shader_log` and such. Therefore, the error should carry the message `String` by default. Once can still get the log on success, if he needs to. Alternatively, we can just write into `warn!` for the success, and return in an `Err` for the error.

PR #702, opened 2015-04-25
Render Stream

### LINKED ISSUE #693: Get Renderer and Output together
Such an abstraction could have convenient drawing methods that `Canvas` currently has.

kvark:
Closes #693 

I can't modify the examples now, since cgmath-rs and noise-rs are broken at the moment, but the change should be backwards compatible, so we can do it any time later.

Edit: https://github.com/bjz/cgmath-rs/pull/211

Noise-rs is used from crates.io, where it's not updated yet. Not sure if it will be.
Also, gfx_gl doesn't seem to be building any more on latest nightly.

Anyway, I believe this PR is complete. Calling for reviewers - @csherratt @bvssvni @cmr @bjz


src/extra/stream.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Command buffer type to constraint the `Renderer`.
+    type CommandBuffer: CommandBuffer<R>;
+    /// Constrained `Output` type.
+    type Output: Output<R>;
+
+    /// Get the output only.
+    fn get_output(&self) -> &Self::Output;
+
+    /// Access both of the stream components.
+    fn access(&mut self) -> (&mut Renderer<R, Self::CommandBuffer>, &Self::Output);
```
bvssvni: `get_access_mut` might be better here, but I'm not sure. Discussion on get/set patterns for Piston is here https://github.com/PistonDevelopers/piston/issues/882

bvssvni:
Looks good.

I do not like the name `Stream`, but I have no better suggestion for now.

src/extra/stream.rs
kvark: Problem is, it still returns the read-only `Output`, so adding mut to the function name may be confusing. If `mut` is still relevant, how about `get_all_mut`?

kvark:
@bvssvni thanks for the review!

src/extra/stream.rs
bvssvni: You're right.

bvssvni:
Merging.

bvssvni:
@kvark Are you on IRC?

kvark:
No, sorry, occupied with life at the moment.

PR #708, opened 2015-04-28
Static bounds on resource types

### LINKED ISSUE #707: Static lifetimes for resource types
I'm getting seriously fed up with all the need to specify lifetime bounds on everything once you want to constrain anything related to `Resources`. I believe we can safely force all the resource types to be `'static` bound.

kvark:
Fixes #707

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.3.0"
+version = "0.3.1"
 description = "A high-performance, bindless graphics API"
 homepage = "https://github.com/gfx-rs/gfx-rs"
 repository = "https://github.com/gfx-rs/gfx-rs"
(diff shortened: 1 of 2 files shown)


kvark:
Should be harmless, since `'static` is the least limiting, and our back-end implementations are not expected to have anything dynamic there (need confirmation here). It would be really nice to have one of those other back-ends at least mocked up before deciding this.

Waiting for the review

kvark:
This is too restrictive, even if likely harmless.

bvssvni:
Arghh, no good way to do this!

kvark:
@bvssvni it turned out a bit of a hell for me [here](https://github.com/kvark/gfx_scene/blob/632c6e24d035e3b516dbdac0f05d122ad9d201f4/src/scene/cull.rs), but later I realized two things:
  1. My static bound has nothing to do with gfx in particular. I opened a [forum topic](https://users.rust-lang.org/t/zero-lifetime-for-references-that-never-exist/1168) on it.
  2. We *might* want to have this flexibility preserved for D3D backends, where there is lots of pointers to COM objects. I don't want a major headache later on.

Thus, we are back to the original problem, where `'a` needs to be specified for a lot of stuff. I wish there was `R::* :'a` bound or something...

cc @cmr @bjz

PR #720, opened 2015-05-09
This adds a custom Arc type for stable reason

csherratt:
This includes: https://github.com/gfx-rs/gfx-rs/pull/719 (Feel free to cherry-pick this change if you want to include it what that changeset).

This creates a custom Arc type so that we can use `is_unique` in our codebase.

This could be a breaking change as I do not have the full `Arc` for example `Display` is missing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: rust
 rust:
 - nightly
-#- beta
+- 1.0.0-beta.4
 before_script:
 - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
 notifications:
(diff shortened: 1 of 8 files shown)


kvark:
Superb! Thanks @csherratt !

I'm wondering, does `cfg(unstable)` set automatically by rust? I'm happy to see Travis green on both rust-beta and nightly :)

bvssvni:
Might be a good idea to put the custom Arc in its own library, in case this is needed elsewhere.

kvark:
@bvssvni somehow, I feel that we are the sole major users of `is_unique`, so everyone else can just use the stock `Arc`.

PR #725, opened 2015-05-10
Macros return!

### LINKED ISSUE #676: Rust 1.0 compatibility

kvark:
~~Hold on. I figured the `ShaderParam` implementation needs to handle unused parameters gracefully.~~
Ready to ship!

Closes #676 
Examples: https://github.com/gfx-rs/gfx_examples/pull/23
Macros: https://github.com/gfx-rs/gfx_macros/pull/18

Practically deprecates `gfx_macros`. I've lost a lot of blood writing these syntax extensions, but I'd better have a cleaner (and smaller) code base to deal with, so my vote is to let `gfx_macros` rest in peace.

Current problems (non-critical):
  - I had to add `pub` to all the generated structs and members, which prevents these to be declared inside functions
  - We have `_r: PhantomData<R>` in each shader parameter structure

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.3.1"
+version = "0.4.0"
 description = "A high-performance, bindless graphics API"
 homepage = "https://github.com/gfx-rs/gfx-rs"
 repository = "https://github.com/gfx-rs/gfx-rs"
--- a/src/device/attrib.rs
+++ b/src/device/attrib.rs
@@ -125,9 +125,50 @@ pub struct Format {
     pub instance_rate: InstanceRate,
 }
 
+
+/// Fixed-point version of integer attributes.
+#[derive(Clone, Copy, Debug, PartialEq, Hash)]
+pub struct FixedPoint<T>(pub T);
+
+impl<T: Copy> FixedPoint<T> {
+    /// Cast a fixed-size2 array to fixed-point.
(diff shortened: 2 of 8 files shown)


kvark:
I believe this needs a version bump to 0.4.0 (with publishing), but I'm going to wait for the code review and will include the version bump in the upcoming changes.

csherratt:
Looks good to me. I would like a 3rd to look over it.

cmr:
I'm impressed that the macro is almost as nice to use as the original syntax extension. LGTM.

bvssvni:
Looks good. I noticed the examples uses underscore in front for the shader parameter `x@ _x`. Don't know if this is required, but needs some documentation.

kvark:
Thanks everyone! I'm merging then?

@bvssvni about underscores. 1) they are in tests, not examples. and 2) they are not required, the variables are just unused, so I put an underscore.

kvark:
I've updated `gfx_macros` (the bye-bye patch), to avoid forcing everyone to use the new macros just yet. Merging.

kvark:
Did no one really notice the X-Com reference with `gfx::attrib::Floater`? :earth_africa:

csherratt:
I was wondering where you got the name from. Shamefully, I have never actually played xcom.

PR #729, opened 2015-05-13
Fbo-less support

### LINKED ISSUE #728: Support execution without FBOs
VirtualBox supports only core GL-2.1 features, no FBOs there. We should make sure we run on it.

cc @eddyb @mattico

kvark:
Fixes #728 
Fixes blitting from a main FBO.
Improves `FactoryExt`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.4.0"
+version = "0.4.1"
 description = "A high-performance, bindless graphics API"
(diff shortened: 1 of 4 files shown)


mattico:
@kvark when I get home, I'll see if I can get hematite working in Virtualbox.

kvark:
@mattico thanks! that would be a good test. I didn't verify that it actually works.

mattico:
@kvark hematite still doesn't run in VBox after updating gfx-rs and gfx_device_gl. Same panic for no FBO.  I can look into it more next week.

Issue #737, opened 2015-05-15
Version 0.5


kvark:
Changes:
- [x] add multiple factories in #690
- [x] deprecate/remove `Canvas` in #704
- [x] `link_program_source` semantics
- [x] `create_buffer_*`, `update_buffer*` - some may need to be moved to `FactoryExt`
- [x] refactor shader parameter links in #738 
- [x] deprecate/remove `Graphics`
- [x] remove `IndexBuffer`

PR #741, opened 2015-05-17
Removed Graphics, deprecated Canvas

kvark:
(this goes to v0.5 branch)
related to #737


src/extra/canvas.rs
```diff
@@ -24,7 +24,7 @@ pub trait Window<R: Resources>: Output<R> {
     fn swap_buffers(&mut self);
 }
 
-/// A canvas with everything you need to draw on it.
+/// DEPRECATED
 pub struct Canvas<W, D: Device, F> {
```
csherratt: does `#[deprecated]` work?

kvark: I was under impression that this attribute is only reserved for standard libraries. Will try now.

kvark: `error: stability attributes may not be used outside of the standard library`

csherratt: :( oh well.

csherratt:
lgtm

PR #742, opened 2015-05-18
Remaining v0.5 breaking changes

kvark:
On a second thought, maybe `OwnedStream` as better than re-using `Canvas` name. At least it goes in line with `OwnedBatch` and such...


src/extra/factory.rs
```diff
@@ -26,8 +26,8 @@ pub trait FactoryExt<R: device::Resources>: device::Factory<R> {
     /// Convenience function around `create_buffer` and `Mesh::from_format`.
     fn create_mesh<T: VertexFormat>(&mut self, data: &[T]) -> Mesh<R> {
         let nv = data.len();
-        //debug_assert!(nv < self.max_vertex_count); //TODO
-        let buf = self.create_buffer_static(data);
+        debug_assert!(nv <= self.get_capabilities().max_vertex_count);
```
csherratt: Is an assert better then using a Result here?

kvark: No. I'll fix it.

csherratt:
The rest looks good to me.

kvark:
Thanks @csherratt ! I'll add a bit of glue for index creation and merge to v0.5 branch. Then there's going to be a bunch of PRs to actually move master to v0.5.

Issue #723, opened 2015-05-09
Cargo dependency policy


kvark:
Currently, we have all the gfx-related dependencies as Git paths. However, since we are aiming to be rust-beta compatible, maybe we should start using `crates.io` again? I'm not sure what the proper policy is... The benefit is - better discover-ability, less dependency on Git. The downside is - our headache to update all the crates when a breaking change sneaks in.

cc @csherratt @bvssvni @bjz @cmr

bvssvni:
Using crates.io is a bit cumbersome because you don't know precisely what's there. The more crates you have the bigger problem. In Piston we work against git dependencies whenever possible, and use commented versions that tells what is on crates.io. I think the issue is how stable the library is, if it it is almost stable then this becomes a less problem.

mitchmindtree:
I'd love it if everyone started to stay on top of crates.io publishing more - the most frustrating thing I find about git dependencies is that I can't host my own crate on crates.io.

This might sound crazy, but I'd really appreciate some sort of travis script that could automatically increment the end integer (if the version hasn't changed since the last publish) and attempt to `cargo publish` whenever new code builds successfully and is merged into master. Even if it resulted in crazy versions like 0.1.4567, at least it would be available.

csherratt:
This is actually an important problem to solve right. Currently, `gfx` ecosystem is developing a nice little stack where you need to use all the pieces you are interested in to get anything done. The pieces themselves will grab the correct dependencies that they specify in there Cargo.toml.

The problem arises when these two are in disagreement of which gfx to use. Cargo does not actually see this as a problem and will compile both. You then get a useless error message like ```Trying to use `gfx::FooBar` expected `gfx::FooBar` ```. This happens since the `gfx` is different for the types.

Cargo.io is actually a pretty good solution to this problem if we use the `version` correctly. Since everything is bound to come from Cargo.io, you won't have to deal with something including git randomly. That being said, if one component uses a more strict `version` then another. I'm pretty sure we are in the same boat.

@mitchmindtree I don't have a problem with automation in general, but I will point out that this may causes semver abuse. Since the minor needs to be incremented when publishing adding apis.

kvark:
Ok, after the gfx-0.5 breakage of user libraries and some stabilization of Rust, I'm leaning towards using crates.io for dependencies. We can use `version = "0.y.*"` to keep the last patched non-breaking dependency both between gfx libraries and in the user code.

kvark:
Fixed with #759 landing

Issue #745, opened 2015-05-20
Version 0.6


kvark:
- [x] use `crates.io` for dependencies (#723)
- [x] remove global imports from examples (#744), cleanup `gfx::traits`
- [x] make `Renderer::update_*` to return a `Result` (#139)
- [x] remove `batch::Context` (#751)
- [x] make `Factory` clonable
- [ ] ~~change `clear()/blit()` semantics~~ (#630, possible to delay to 0.7)

PR #746, opened 2015-05-21
Use draw_state from crates.io

csherratt:
part of #723

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,8 @@ authors = ["The Gfx-rs Developers"]
 name = "gfx"
 path = "src/lib.rs"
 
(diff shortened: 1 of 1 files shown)


kvark:
@csherratt this may turn out to be a breaking change (that's why #723 is a part of #745). Imagine someone using draw_state alongside gfx, they will stop being compatible for as long as the person still refers to draw_state git.

With draw_state in particular the risk to break things is low, but spreading the same on gfx and the device will surely break everything. I think we should do this in a branch until we are ready to release version 0.6. Objections are welcome ;)

csherratt:
You are correct. I will direct this to 0.6,

PR #747, opened 2015-05-21
Use draw_state from crates.io

### LINKED ISSUE #723: Cargo dependency policy

csherratt:
Resolves #723 for gfx-rs, Continuation of #746

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,8 @@ authors = ["The Gfx-rs Developers"]
 name = "gfx"
 path = "src/lib.rs"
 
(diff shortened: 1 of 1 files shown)


kvark:
thanks! please self-merge when ready

PR #748, opened 2015-05-21
Fixed update_buffer(size=0) case

kvark:


csherratt:
I'm guessing that was an annoying bug to find.

kvark:
@csherratt our `DataBuffer::add_vec` is not capable to handle null slices. `gfx_text` appears to request that from time to time. Not too difficult to find/repro, fortunately ;)

PR #753, opened 2015-05-23
Make Batch trait object-safe

aepsil0n:
Breaking change!

By removing the associated error type the batch trait becomes object-safe. The advantage is, that it allows run-time dispatch of batches, as long as the backing Resources implementation is known at compile-time.

The downside is, of course, that the error type is less generic. This is deemed less useful, as it only conveys information in case of errors and the additional Error::Other enum variant allows for any kind of String to be passed as an error.

As an alternative one could implement an additional trait `DynamicBatch<R>` for all `B: Batch<R>`. This can also be done in user code. However, I think that the possibility to dynamically dispatch batches warrants this change.

### DIFF
--- a/src/extra/stream.rs
+++ b/src/extra/stream.rs
@@ -20,7 +20,7 @@ use device::{Device, InstanceCount, Resources, VertexCount};
 use device::draw::CommandBuffer;
 use device::target::{ClearData, Mask, Mirror, Rect};
 use render::{DrawError, Renderer, RenderFactory};
-use render::batch::Batch;
+use render::batch::{Batch, Error};
 use render::target::Output;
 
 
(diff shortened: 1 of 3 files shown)


aepsil0n:
Please view this as an RFC. It may also require some bikeshedding about what exactly the error type should look like.

kvark:
@aepsil0n thank you for the contribution!

Is it really a breaking change? I suspect almost no one to notice it.

I think we can merge this. @csherratt, how about you?

csherratt:
If we are worried about breaking things we can also commit it against the `0.6` branch. Otherwise I don't have a problem with this change.

aepsil0n:
Well, strictly speaking it is. And any code that actually uses the associated type as such will break. Personally I wouldn't mind. I merely wanted to make it clear.

csherratt:
It's minor. I am merging it. Most people are probably using `unwrap()` anyhow.

kvark:
Right. I'm going to be the only one to suffer with gfx_scene, I can live with that ;)

kvark:
Looks like @aepsil0n is onto something interesting: https://github.com/aepsil0n/gfx-func

aepsil0n:
Uhm… tentatively yes. I'd like to have something like elmesque for 3D eventually.

PR #755, opened 2015-05-26
[v0.6] Texture/Buffer update errors

### LINKED ISSUE #139: Error codes for `update_*` routines
Since we have `SurfaceInfo` and `TextureInfo` stored on the renderer side, we can check this right away and return a `Result`:
- format size matches `T` size
- number of elements matches the vec len

kvark:
Fixes #139

### DIFF
--- a/src/device/tex.rs
+++ b/src/device/tex.rs
@@ -116,6 +116,18 @@ pub enum Components {
     RGBA,
 }
 
+impl Components {
+    /// Get the number of components.
+    pub fn get_count(&self) -> u8 {
+        match *self {
+            Components::R    => 1,
+            Components::RG   => 2,
+            Components::RGB  => 3,
+            Components::RGBA => 4,
+        }
+    }
+}
+
 /// Codec used to compress image data.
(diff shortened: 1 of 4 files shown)


src/device/tex.rs
```diff
@@ -253,6 +265,33 @@ impl Format {
...
         }
     }
+
+    /// Get size of the texel in bytes.
+    pub fn get_size(&self) -> Option<u8> {
+        Some(match *self {
+            Format::Float(c, FloatSize::F16) => c.get_count() * 2,
+            Format::Float(c, FloatSize::F32) => c.get_count() * 4,
+            Format::Float(c, FloatSize::F64) => c.get_count() * 8,
+            Format::Integer(c, bits, _) => (c.get_count() * bits) >> 3,
```
csherratt: The LLVM is smart enough to turn `x / 8` into `x >> 3`.

csherratt:
LGTM

kvark:
@csherratt Thanks for having a look

PR #756, opened 2015-05-26
[v0.6] Refactored texture kinds

kvark:
Partially contains #756

### DIFF
--- a/src/device/tex.rs
+++ b/src/device/tex.rs
@@ -464,32 +464,6 @@ impl TextureInfo {
         Default::default()
     }
 
-    /// Convert to a default ImageInfo that could be used
-    /// to update the contents of the whole texture.
-    pub fn to_image_info(&self) -> ImageInfo {
-        ImageInfo {
-            xoffset: 0,
-            yoffset: 0,
-            zoffset: 0,
-            width: self.width,
-            height: self.height,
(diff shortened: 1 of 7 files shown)


src/device/tex.rs
```diff
@@ -300,55 +339,55 @@ pub enum FilterMethod {
...
 /// restricted by what Metal exposes, though this could conceivably be
 /// extended in the future. Note that a single texture can *only* ever be of
 /// one kind. A texture created as `Texture2D` will forever be `Texture2D`.
 // TODO: "Texture views" let you get around that limitation.
 #[derive(Eq, Ord, PartialEq, PartialOrd, Hash, Copy, Clone, Debug)]
-pub enum TextureKind {
+pub enum Kind {
     /// A single row of texels.
-    Texture1D,
+    D1,
```
csherratt: YAY \o/. This is much less verbose.

kvark: I was just summing you in gitter to ask what you think about it. `Dx` pattern may be a little confusing at first, but definitely very short to type.

csherratt: I'm ok with it. I can't think of a better name that is short, and legal.

kvark: \o/

PR #757, opened 2015-05-26
[v0.6] Removed batch context

### LINKED ISSUE #751: Burn the batch context with fire
It can be used to save on mesh/program cloning (which have allocations!) when creating lots of batches. However, I'm convinced it brings more trouble than the benefit (#510, #464, #351). Besides, it can just be moved outside of gfx-rs crate.

The problem with the `Context`, outside of it just being another thing to consider or think about, is that it doesn't have a very safe interface. Thus, it could be fine to use inside another library as an implementation detail, but not fine to expose it to the user. I may end up moving it out and using it for gfx_scene, but I'll hesitate to do it until mesh/program cloning shows up in the profiler.

Let's burn this thing! We only really need a `Batch` trait and two kinds of batches: one with everything in it (`batch::Full`?), and another one with just the core components that have to be bound together (`batch::Core`). This is a breaking change supposedly for gfx-0.6

Objections? @csherratt @cmr @bjz

kvark:
Closes #751 
The batch context will be moved into a separate crate (presumably, `gfx_batch`) inside `gfx_scene` project, which needs lightweight batches and can isolate the context unsafety from the user.

### DIFF
--- a/src/render/batch.rs
+++ b/src/render/batch.rs
@@ -36,8 +36,6 @@ pub enum Error {
     Mesh(mesh::Error),
     /// Error connecting shader parameters
     Parameters(ParameterError),
-    /// Error context is full
-    ContextFull,
 }
 
 /// Return type for `Batch::get_data()``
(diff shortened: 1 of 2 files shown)


kvark:
Someone is going to have a sweet time merging this with master...

PR #760, opened 2015-05-27
Untested work around for tuple issue

csherratt:
Compiles, just wanted your 2 cents before I went to far with it.

Should fix https://github.com/gfx-rs/gfx-rs/pull/759#issuecomment-105729381


kvark:
Please check if it works with two or more `gfx_parameters!` blocks. I suppose `__gfx_internal` will conflict.

csherratt:
it does. Closing.

PR #762, opened 2015-05-28
Add cmp_ref and access functions for a Core

csherratt:


kvark:
This would look nicer if we had the same generic type for all the resources, like we used to. It's an internal detail though, so we can proceed as is now.

kvark:
Also, this PR needs to have the patch version bumped, though I don't want to wait for another travis pass, which takes a while.

PR #779, opened 2015-06-21
Closes #739 - Implemented Dummy Device, not sure if it's ok?

### LINKED ISSUE #739: Dummy device
We could have it implemented inside gfx-rs crate. It would do minimal work, just to allow testing gfx-rs apps for compilation.

jeremija:
**DO NOT MERGE YET**

Hi guys, I'm new to Rust and here is my attempt to implement the Dummy Device described in issue #739. I'm not sure if I took the right steps and I would appreciate comments on what I did wrong / how to improve the code.

There are new warnings about unused variables because I was not sure how to do it - I used the implementation of [gl/src/lib.rs#Device](https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/lib.rs#L601) and [gl/src/draw.rs#CommandBuffer](https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/draw.rs#L68) as examples.

Thanks


src/core/src/device/dummy.rs
```diff
@@ -0,0 +1,185 @@
...
+use device::draw::{CommandBuffer, Access, Gamma, Target};
+use device::shade;
+use super::{tex};
+use draw_state::target::{Rect, Mirror, Mask, ClearData, Layer, Level};
+
+pub struct DummyDevice {
+    capabilities: Capabilities
+}
+pub struct DummyCommandBuffer {
+    buf: Vec<String>
```
csherratt: Rather then make this an buffer of strings, it should probably be a buffer of `Commands`. The `Commands` can easily be turned back into a string using `std::fmt::Debug`

We also need some way of reading this Vec back.

jeremija:
Hi @csherratt, thank you for your comment - it makes sense. Right now I have an implementation almost identical to the one in [src/backend/gl/src/draw.rs](https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/draw.rs#L26). At first I wanted to extract the `gfx_device_gl`'s `Command` and `CommandBuffer` implementations somewhere it can be shared, but then I realized it was not that simple because `gfx_device_gl` has it's own type definitions such as `Buffer` and  `ArrayBuffer`, which all point to  `gl::types::GLuint`, and are used in the method signatures. Is there a way to follow the [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) principle in this case?

kvark:
@jeremija if we make `pub enum Command<R: Resources> {}`, it could be shared between GL, DX11, and the dummy device. Not 100% sure, but worth a try ;)

jeremija:
@kvark thanks - I was thinking about something like this, but I'm not sure if this would break existing API as the current gl's `pub enum Command` is not generic. I'll try when I'm back at my PC

kvark:
@jeremija `Command` is public in the module `draw`, but this module is private in the `gfx_device_gl` crate, so it's still hidden from the outside.

jeremija:
@kvark I just made it work with with generics, but doesn't the line `pub use self::draw::{Command, CommandBuffer}` in [src/backend/gl/src/lib.rs#35](https://github.com/gfx-rs/gfx/blob/c6cb23386fd4c4056b18aeccff8c1494df44b3a5/src/backend/gl/src/lib.rs#L35) reexport `Command`, so it's not actually hidden from the outside?

kvark:
@jeremija you are right, it is re-exported, either by mistake or in an invitation to re-use it. In any way, a command is just an implementation detail for the command buffer. Should be no harm in sharing it between devices and making it more generic.

src/core/src/device/mod.rs
```diff
@@ -28,6 +28,8 @@ pub mod handle;
 pub mod mapping;
 pub mod shade;
 pub mod tex;
+pub mod dummy;
+mod command;
```
kvark: Please arrange alphabetically (separately for public and private)

src/core/src/device/dummy.rs
```diff
@@ -0,0 +1,50 @@
...
+use device::command::{GenericCommandBuffer};
+
+pub struct DummyDevice {
+    capabilities: Capabilities
+}
+
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum DummyResources{}
+
+pub type Buffer         = u32;
```
kvark: No need for these types, really. Everything should be accessed by interested parties as `Resources::XXX` associated types.

jeremija: I must have missed the Buffer, ArrayBuffer, and other types after refactoring as I copied it from the gl implementation. Is there a better type than `u32` for dummy device's resources?

kvark: `()` would work too. Note that they are public in `gl` implementation for some reason I don't recall, but definitely not to expose it to the rest of the world. If you remove `pub` from them and everything still compiles - that would be great.

jeremija:
I pushed the generic implementation of `CommandBuffer` and made use of it in `gfx_device_gl`. The only thing left is I'm not sure how to implement the following methods of `Device` trait for `DummyDevice`:

```rust
    /// Reset all the states to disabled/default.
    fn reset_state(&mut self);

    /// Submit a command buffer for execution.
    fn submit(&mut self, SubmitInfo<Self>);

    /// Cleanup unused resources, to be called between frames.
    fn cleanup(&mut self);
```

Would just creating counters for `cleanup()` and `reset_state()`, and `submit()` suffice?

src/core/src/device/command.rs
```diff
@@ -1,41 +1,32 @@
-// Copyright 2014 The Gfx-rs Developers.
```
kvark: Why did you remove the license?

src/core/src/device/dummy.rs
```diff
@@ -0,0 +1,41 @@
+#![allow(missing_docs)]
```
kvark: Please add a license header here.

kvark:
@jeremija thanks! Let's keep these methods implementations empty for now. Just a couple of licenses missing before we can proceed.

src/core/src/device/command.rs
jeremija: Ah git must've recognized the file as being moved because of the similarities between the files, whereas I actually copied some things from `draw.rs` and I didn't think about the licenses. Will add the the license headers.

jeremija:
@kvark you're welcome! Thank you for your help - I learned a lot and I hope I can contribute more in the future!

kvark:
@homu r+

PR #780, opened 2015-06-21
Fix links in Readme.md

csherratt:


csherratt:
@homu r+

kvark:
@csherratt thanks! the documentation seems to work now

PR #781, opened 2015-06-26
Fix typo on extension checking

fkaa:
Doesn't actually check if there is such an extension at all, so typos aren't caught.


kvark:
@homu r+

PR #782, opened 2015-07-02
Improve instanced functions documentation

Bastacyclop:
I was a bit confused by those "counts"


kvark:
Please add the license headers following other examples.

kvark:
Note that GLSL 120 shaders here are useless until we support the EXT version of instancing GL extensions. It's just FYI - no need to remove them, I suppose, we should instead just get the proper EXT instancing support.

kvark:
@homu r+

kvark:
@Bastacyclop well done, thanks for the PR!

Bastacyclop:
Glad I could help.

PR #787, opened 2015-07-18
Bump the version number for 0.7

csherratt:


csherratt:
@kvark @bvssvni any objections?

kvark:
I don't think it's worth it if piston_window is the only thing broken and it's fixed already.

bvssvni:
Yeah, damage is already done. Incrementing to 0.7 won't fix it, so it will be less work to fix code for 0.6.5. If we bump it to 0.7 we need another round of dependency upgrades.

PR #790, opened 2015-08-04
Add UBO TileMap example

olsonjeffery:
Here's the first graf from the README:

> This example shows a shader-driven, Uniform Buffer Object (UBO)-backed scrolling tilemap, akin to what is seen in 2D, tile-based games. The tiles are rendered on a single OpenGL plane mesh, with each vertex referencing data in a UBO. The contents of the UBO are updated, as the user "scrolls" around the tilemap, with fragment shader logic determining it's current tile and appropriate offset, which is used to discover the fragment color to be displayed. As a bonus, basic lighting formation is also contained in the UBO data.

I have a [youtube video](https://www.youtube.com/watch?v=aFjn4TMKZco&feature=youtu.be) of this example in action.

There are probably a few things I did differently, stylistically, from the rest of the `gfx` examples. Let me know what you'd like me to clean up so this can land.

Thanks!


olsonjeffery:
build failure seems unrelated:

```
Build failed, waiting for other jobs to finish...
Could not document `lazy_static`.
Caused by:
  Process didn't exit successfully: `rustdoc /home/travis/.cargo/registry/src/github.com-0a35038f75765ae4/lazy_static-0.1.14/src/lib.rs --crate-name lazy_static -o /home/travis/build/gfx-rs/gfx/target/doc -L dependency=/home/travis/build/gfx-rs/gfx/target/debug/deps -L dependency=/home/travis/build/gfx-rs/gfx/target/debug/deps` (exit code: 101)
--- stderr
thread '<unnamed>' panicked at 'failed to generate documentation: File exists (os error 17)', ../src/librustdoc/lib.rs:293
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: Any', ../src/libcore/result.rs:731
```

csherratt:
Rebuilding it seemed to clear the error. So maybe there is a race in how we are building gfx on travis.

examples/ubo_tilemap/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The plane can be zoomed in/out with `+` and `-`. The camera can be moved in space via the arrow keys. The contents of the tilemap can be "scrolled" with the `W` (up), `S` (down), `A` (left) and `D` (right) keys.
+
+## Tilesheet Information
+
+The provided `scifitiles-sheet_0.png` was created by [Michele Bucelli (aka Buch)](https://www.patreon.com/buch?ty=h) and is [Public Domain](http://opengameart.org/content/sci-fi-interior-tiles).
+
+## Screenshot
+
+![Shadow Example](screenshot.png)
```
csherratt: Might want to have the mouse over be something other then `Shadow Example`.

examples/ubo_tilemap/main.rs
```diff
@@ -0,0 +1,525 @@
...
+        // to the frag shader as a varying, used to determine the "current tile" and the frag's offset,
+        // which is used to calculate the displayed frag color)
+        let vertex_data: Vec<VertexData> = plane.shared_vertex_iter()
+            .map(|(raw_x, raw_y)| {
+                let vertex_x = half_width as f32 * raw_x;
+                let vertex_y = half_height as f32 * raw_y;
+
+                let u_pos = (1.0 + raw_x) / 2.0;
+                let v_pos = (1.0 + raw_y) / 2.0;
+                let tilemap_x = (u_pos * width as f32) as i32;
```
csherratt: You might want to use [floor()](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.floor) here instead of converting it to an int and back.

examples/ubo_tilemap/main.rs
```diff
@@ -0,0 +1,525 @@
...
+            batch: gfx::batch::Full::new(mesh, program, data).unwrap()
+        };
+        item.batch.slice = slice;
+        item.batch.state = gfx::DrawState::new().depth(gfx::state::Comparison::LessEqual, true);
+
+        return item;
+    }
+    
+    pub fn update_data<TFactory>(&mut self, factory: &mut TFactory) where TFactory: Factory<R> {
+        if factory.update_buffer(&self.batch.params.tilemap, &self.data, 0).is_ok() == false { panic!("Failed to update glsl buffer"); }
```
csherratt: You can just use an `unwrap` or `except` here if you just want to abort on error.

csherratt:
Overall this is a pretty neat demo. Other then a few nits, it LGTM.

olsonjeffery:
@csherratt thanks! I will address the feedback and push up changes later today.

fkaa:
:+1:! I'm not able to look at the code from where I am right now but I saw the video and it looks great!

I think we need More "pratical" examples like this, maybe even a game soon :)

examples/ubo_tilemap/main.rs
```diff
@@ -0,0 +1,525 @@
...
+            .map(|i| i as u32)
+            .collect();
+        let slice = index_data.to_slice(factory, gfx::PrimitiveType::TriangleList);
+
+        // set up texture
+        let mesh = factory.create_mesh(&vertex_data);
+        // shader
+        let program = {
+            let vs = gfx::ShaderSource {
+                glsl_150: Some(include_bytes!("tilemap_150.glslv")),
```
kvark: Not much sense using `gfx::ShaderSource` if you only have shaders for a single context version. Using `factory.link_program()` you could just pass the code directly.

examples/ubo_tilemap/main.rs
```diff
@@ -0,0 +1,525 @@
...
+    // paper in with dummy data
+    for ypos in (0..tilemap_size[1]) {
+        for xpos in (0..tilemap_size[0]) {
+            let idx = (ypos * tilemap_size[0]) + xpos;
+            tilemap.tiles[idx] = TileMapData::new([1.0, 7.0, 0.0, 0.0]);
+        }
+    }
+    let calc_idx = {|x, y| (y * tilemap_size[0]) + x};
+
+    tilemap.tiles[calc_idx(1,3)] = TileMapData::new([5.0, 0.0, 0.0, 0.0]);
```
kvark: Asks for a helper like this: `tilemap.set_tile(1, 3, [5.0, 0.0, 0.0, 0.0]);`

kvark:
Thanks @olsonjeffery ! I skimmed through the code, put some notes, but overall it looks good.
One thing being not clear to me is why using UBO here in the first place? You are essentially using it as a 2D texture, so why not expressing it explicitly as such?

olsonjeffery:
> One thing being not clear to me is why using UBO here in the first place? You are essentially using it as a 2D texture, so why not expressing it explicitly as such?

@kvark I think dynamically generating a texture in application code/memory and then copying a subset of the texture into video memory every time the tile offset changes is going to be very slow in comparison? Also, you'd be placing a much lower ceiling on how large of a tilemap you could generate if you're generating and keeping a huge texture in application memory (also with this UBO approach you can store arbitrary, per-tile metadata like lighting-level-and-tint, visibility, etc to drive shader logic)

In this case, the `a_BufPos` passed into the vertex program and varied through to the frag shader is a fixed pointer to a cell in the UBO, whose contents are better suited to be updated frequently by the applicaton, IMO. I'll yield that you know _much more_ about the semantics and implications of this stuff than I do, but this is my impression and experience.

Also, as a small bit of appeal-to-authority: I actually discussed the "generate a huge texture and copy a subset to the GPU as needed" approach with an acquaintance of mine who does a lot of AAA gamedev consulting, and he explained the approached used in this example to me (this technique is actually called a "character map" and it goes back to console programming on constrained systems like GBA, NDS, SNES, etc). After reasoning through and implementing, I _personally_ feel like its a much more efficient way to approach the problem, and it gives you a lot more control.

examples/ubo_tilemap/main.rs
olsonjeffery: yeah I was lazy here and only wanted to work on a single set of shader programs.. this boilerplate code is left intact from the example code that I adapted this from. I believe I should be able to copy/rename to a 120 version and be ok, correct? I don't think I'm using any non 1.2 stuff in my shader programs..

kvark: Sure! Keep in mind that UBO is only in GL core since 3.1 though.

olsonjeffery: Ah! i see.. so that'd be 1.4 . I guess I should just leave it as 1.5 only, then? In that case, would you prefer that I switch to just use `factory.link_program()` directly as in your original comment?

kvark:
@olsonjeffery I didn't have a large texture in mind. I was thinking about using a small texture for exactly the same purpose as your UBO but with 2D indexing. Assuming you'd rewrite it every frame and seeing that the data you store is just `vec4` anyways. Really though, there is nothing wrong with your approach ;) We'll be happy to merge once you address @csherratt's suggestions.

olsonjeffery:
@kvark hah, yeah! other examples I've seen have taken that approach (particularly [this example using SFML/C++](http://www.the2dgame.com/index?page=articles&ref=ART9) ), but the programatic texture API stuff looked pretty hairy, so I ended up just doing in this direction :) . Maybe I can follow-up with that sort of approach to get more comfortable with that part of the API.

examples/ubo_tilemap/main.rs
kvark: Nah, let's have the room open for 1.40 and leave this code as is for now.

olsonjeffery:
updates have been made. should be gtg.

kvark:
@homu r+

Issue #688, opened 2015-04-14
Fencing the command buffers


kvark:
Suggested by @tomaka :

> what I think should be done for vulkan is that you create a fence each time you submit a command buffer
> this fence is returned in a `Fence` object that holds smart pointers to the resources used by the command buffer
> and Fence's destructor waits for the fence, or you can use it when submitting another command buffer behind the first one
> `submit_command_buffer(after: Option<Fence>, commands: &CommandsBuffer)`
> in which case it doesn't wait
>  fact it would be `submit_command_buffer(after: Option<Fence>, commands: &CommandsBuffer) Fence`
> and the smart pointers in the Fence that you pass as parameter are moved to the Fence that is returned
> at the end of the frame, you wait on the fence
> and it destroys the resources

csherratt:
I have been working on a library that could be useful to build a `Fence` with.

https://github.com/csherratt/pulse

kvark:
This is pretty cool!

csherratt:
I have been thinking about this some more, mostly because I need some support for fencing of commands to reduce the latency for `gfx_vr`. The recommended method suggested by Carmack is to draw a small triangle after the frame swap and wait on a fence immediately after it. This stops the GPU from being to aggressive with driver side queueing.

I think overall the approach of adding an optional fence is a good approach. This allows `gfx` to intelligently use `glWaitSync​`, and if the user wants to stall the CPU (to say avoid accessing a memory mapped buffer, or to reduce latency in VR) they can.

The only issue I see with cpu stalling is that the `glClientWaitSync` is not a very nice library for dealing with multi-threaded work. We can't move the OpenGL context around as needed, so we will somehow need to manager multiple threads waiting on multiple fences. Which sounds incredibility messy.

Maybe v1 can just do the fencing and possibly a `glClientWaitSync` using the `device` directly. That solves the multi-threaded issue by making it impossible (for now).

kvark:
This is actually required for the Vulkan backend #629 . A fence per command buffer not only signifies when the used resources are no longer needed, but it also tells us when that command buffer can be re-used.

kvark:
Not happened in pre-ll

PR #791, opened 2015-08-05
Add fence suppot

csherratt:
Solves #688

This exposes the command fences to the developer. Via the `DeviceFence` trait. Fences are managed much like other resources and are added as an associated type. This api adds two abilities, the fence can block the device server by using `fenced_submit`. Or the user may synchronize in the client cpu by using  `fence_wait`.

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -647,6 +647,7 @@ impl gfx::Device for Device {
             |gl, v| unsafe { gl.DeleteRenderbuffers(1, v) },
             |gl, v| unsafe { gl.DeleteTextures(1, v) },
             |gl, v| unsafe { gl.DeleteSamplers(1, v) },
+            |gl, v| unsafe { gl.DeleteSync(v.0) },
         );
     }
 }
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -651,3 +651,31 @@ impl gfx::Device for Device {
         );
     }
 }
+
+impl gfx::traits::DeviceFence<Resources> for Device {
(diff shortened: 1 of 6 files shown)


src/backend/gl/src/lib.rs
```diff
@@ -640,6 +647,35 @@ impl gfx::Device for Device {
...
+                let f = self.frame_handles.ref_fence(&fence);
+                self.share.context.WaitSync(f.0, 0, gl::TIMEOUT_IGNORED);
+            }
+        }
+
+        self.submit(info);
+
+        unsafe {
+            let fence = self.share.context.FenceSync(gl::SYNC_GPU_COMMANDS_COMPLETE, 0);
+            self.share.handles.borrow_mut().make_fence(Fence(fence))
```
kvark: Does this line need to be inside `unsafe{}` ?

src/core/src/device/handle.rs
```diff
@@ -163,6 +163,9 @@ impl<R: Resources> Sampler<R> {
     }
 }
 
+/// Sampler Handle
```
kvark: Comment needed

kvark:
Some small nitpicks, otherwise - great stuff!

src/backend/gl/src/lib.rs
csherratt: the GL call is marked as unsafe, but the `make_fence` is not.

kvark: So we can do something like this instead:
```rust
let fence = Fence(unsafe {
    self.share.context.FenceSync(gl::SYNC_GPU_COMMANDS_COMPLETE, 0)
});
self.share.handles.borrow_mut().make_fence(fence)
```

kvark:
Thanks! Are you ready to merge?

csherratt:
@kvark I think it is good to merge.

kvark:
@homu r+

PR #794, opened 2015-08-22
Added Texture cache to gfx

csherratt:
This creates a texture cache. A texture cache is an LRU that manages the GL_ACTIVE_TEXTURES, it will bind textures until there is no free ACTIVE_TEXTURES, then use the reuse the textures based on their age. older textures are evicted first.


kvark:
Very nice!

@homu r+

csherratt:
@homu retry

PR #795, opened 2015-08-22
Added 'static lifetime to Resources

csherratt:


csherratt:
@homu r+

kvark:
I recall doing this at some point and then falling back because this is too limiting. @csherratt are you sure about the change? Where did you need it?

bvssvni:
Found it https://github.com/gfx-rs/gfx/pull/708.

bvssvni:
Non-static lifetimes in such types should be rare. `Resources` should always be `'static`, I guess, so lifetimes on the associated type does not seem to make sense.

If there is a problem with it then we could fix it in a later version of Gfx.

kvark:
It's a one way road though. Changing it back may break some code.
I'd like to see why exactly it is needed now. Perhaps, we could find a better way?

csherratt:
Actually your change shows another possible solution to my problem. I was hacking on `gfx_scene` and could not figure out a nice way of telling the compiler that the lifetime of the resources was at least as long as the lifetime as the reference. Your change shows a (verbose) way of doing so.

I don't have a problem with this being rolled back.

bvssvni:
I'm OK with this being rolled back too, I don't think it breaks much code if we do it now.

PR #799, opened 2015-08-30
Add program::Builder

csherratt:
This is a breaking change.

This adds a program-builder for creation of shader programs. This exists to avoid new parameters each time we add support for new features that may require new options when calling create_program.


kvark:
Nice step in preparation to TF support!
@homu r+

PR #803, opened 2015-09-18
Avoid using the std::intrinsics module

alexcrichton:
This is actually unstable but [there's a bug][bug] in the compiler allowing it
to be used. Good news is that these functions are available stable elsewhere!

[bug]: https://github.com/rust-lang/rust/issues/28075

### DIFF
--- a/src/core/src/device/draw.rs
+++ b/src/core/src/device/draw.rs
@@ -55,13 +55,13 @@ impl DataBuffer {
     /// Copy a given structure into the buffer, return the offset and the size.
     #[cfg(not(unstable))]
     pub fn add_struct<T: Copy>(&mut self, v: &T) -> DataPointer {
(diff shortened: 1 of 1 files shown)


kvark:
Wow, Alex here! Welcome, and thanks for the PR!

@homu r+

alexcrichton:
:heart:

alexcrichton:
Ah and also, would you be ok releasing a new version with this change? I'm hoping to do a crater run with this release to further evaluate the impact of #28364

kvark:
@alexcrichton sure! @csherratt is on it

alexcrichton:
Awesome, thanks!

csherratt:
@alexcrichton it's published now. Cheers!

alexcrichton:
Thanks @csheratt!

PR #807, opened 2015-10-04
gfx_vertex! potential UB fix

red75prime:
Layout of Rust struct without #[repr(*)] is not defined (yet?)


red75prime:
I do something wrong. Gitwise.

PR #808, opened 2015-10-04
gfx_vertex! potential UB fix

red75prime:
Layout of Rust struct without #[repr(*)] is not defined (yet?)

### DIFF
--- a/src/core/src/macros/mod.rs
+++ b/src/core/src/macros/mod.rs
@@ -30,14 +30,14 @@ macro_rules! gfx_vertex {
                 use $crate::attrib::{Offset, Stride};
                 use $crate::attrib::format::ToFormat;
                 let stride = size_of::<$name>() as Stride;
-                let mut offset = 0 as Offset;
+                let tmp: $name = unsafe{ ::std::mem::uninitialized() };
                 let mut attributes = Vec::new();
                 $(
                     let (count, etype) = <$ty as ToFormat>::describe();
                     let format = $crate::attrib::Format {
                         elem_count: count,
                         elem_type: etype,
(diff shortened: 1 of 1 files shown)


src/core/src/macros/mod.rs
```diff
@@ -30,14 +30,14 @@ macro_rules! gfx_vertex {
                 use $crate::attrib::{Offset, Stride};
                 use $crate::attrib::format::ToFormat;
                 let stride = size_of::<$name>() as Stride;
-                let mut offset = 0 as Offset;
+                let tmp: $name = unsafe{ ::std::mem::uninitialized() };
```
csherratt: My concern with this is that if `$name` supports `drop`. It's possible for this to lead to undefined behavior. But at the same time, I can't imagine anyone using `gfx_vertex!` for anything other then Plain-Old-Data.

It could possibly be made safer by adding `std::mem::forget(tmp)` at the end. That's actually not quite enough, because a `panic!()` caused by `<$ty as ToFormat>::describe()` in the case where a user is doing something _very_ strange could happen.

If we move the creation around tmp just before line 37. and do a forget on line 43. I think it is safe, even from destructors.

But I might just be paranoid.

csherratt:
This can be refined in the future by https://github.com/rust-lang/rfcs/issues/710 but for now I think this approach is a good one.

src/core/src/macros/mod.rs
red75prime: Good point

csherratt:
@homu r+

csherratt:
thanks @red75prime!!!

PR #810, opened 2015-10-08
Fix code which caused E0309 warnings, remove trailing commas from trait bounds lists

burtonageo:
Remove some trailing spaces from the end of trait bounds lists on fns


csherratt:
Thanks!

This is a minor change that actually is an api change. I don't think it should be back ported to 0.7, we can publish it as part of 0.8.

csherratt:
@homu r+

PR #814, opened 2015-11-02
Release 0.8 gfx

csherratt:


bvssvni:
Looks good. Are there any changes in in the v0.7 branch that we should include?

csherratt:
I was just cherry-picking changes into 0.7 from master. So I think this is it.

csherratt:
@homu r+

Issue #815, opened 2015-11-03
Pipeline State Object


kvark:
[PSO](https://msdn.microsoft.com/en-us/library/windows/desktop/dn770370%28v=vs.85%29.aspx) a new concept of DX12, basically incorporating ~~the vertex buffers and~~ format definitions, shader programs, and the output ~~targets~~ target count and formats, depth/stencil/blend states. Looks ~~surprisingly~~ somewhat similar to our `Batch` concept, and considering our overdue overhaul of the concept names (hello #698) it seems straightforward to just adopt the modified concept and borrow the name (PSO).

History-wise, we had the output specified inside batches long time ago, but then these were separated. Too bad we didn't have DX12 at that point to look for inspiration ;) time to revisit this decision and bring the output back. This would allow mapping to DX12 perfectly.

Edit: it's important to list the states that are outside of PSO:
- Vertex/Index buffers
- RT views
- Viewport
- Scissor
- Blend factor
- Stencil ref value
- Topology ordering (not the type!) and adjacency info

kvark:
Another important note w.r.t PSO - we'll need to move all the state caching and verification of the `Renderer::draw` (namely, `bind_output` and `bind_mesh` bodies) into the command buffer implementation. DX12 one will not need too do anything but DX11 and GL ones still need to do whatever we are currently doing.

kvark:
Also note that hiding the output targets inside a batch/PSO would allow us to use heterogeneous render target views with some sort of type safety (over the dimensionality and the type, at least), or at least that's the hope/plan. The target views should totally replace our `Plane` enum, which is GL-atavism.

kvark:
Interestingly, PSO being applied to our current GL backend would totally replace the `Program` concept. We pass shaders and some states in, we get PSO out.

kvark:
Huh, just discovered that Metal also has [MTLRenderPipelineState](https://developer.apple.com/library/ios/documentation/Metal/Reference/MTLRenderPipelineState_Ref/index.html#//apple_ref/occ/intf/MTLRenderPipelineState), and it's exactly matching DX12 one. This means we've chosen the right direction, good luck @sectopod actually implementing it!

PR #820, opened 2015-11-05
Strongly-typed shader objects

sectopod:
Having multiple shaders of the same type linked into a program is a classy GL-ism, and we don't need to support it, since no other API does. `ShaderSet` should be superior to the old `Builder`, albeit it totally cuts the output targets specification logic. Considering that you want PSO (#815), this PR would allow easier transition and you can add the targets later on, to PSO itself.


src/core/src/extra/shade.rs
```diff
@@ -55,7 +55,7 @@ impl<'a> ShaderSource<'a> {
     }
 
     /// Pick one of the stored versions that is the highest supported by the device.
-    pub fn choose(&self, model: ShaderModel) -> Result<&'a [u8], ()> {
```
csherratt: I think choose is the correct word.

csherratt:
I like this direction.

Maybe a better approach would be to use a structure rather then an enum.
```rust
struct ShaderSet<R> {
  vertex: Option<VertexShader<R>>,
  geometry: Option<GeometryShader<R>>,
  pixel: Option<PixelShader<R>>,
  targets: &[&str],
  varying: &[&str] 
}
```
This would give you what you want (not allowing linking of two fragment shaders). But without the half dozen combinations that a modern graphics pipe can manage.

sectopod:
I was actually using this `struct` thing in a previous iteration. Aside from the fact it's less ergonomic for the user (have to specify `None` for a lot of stuff each time, and I do expect the user to actually use `ShaderSet`), it's also not as type-safe (i.e. you could attempt forming invalid combinations).

My `enum` approach the only weakness is *potential* implementation bloatware. But firstly, it's not as important as API clarity and user ergonomics, and secondly, there is not really too many combinations in the end (that we need to support). Maybe up to 5 at most.

kvark:
I don't mind either a `struct` or an `enum`, but overall the PR looks to be driving us in the right direction.

csherratt:
Ok. Let's just use an `enum` for now. If it becomes a problem because of a combinational explosion at some point, we can reevaluate it then.

For now this :+1:

kvark:
@sectopod please revert the `choose` change and we'll be happy to proceed

sectopod:
Done. Btw, if we had `struct` full of `Option` members, we'd had to suffer in ergonomics. Not only we'd have to say `None` for the geometry/tessellation shaders every time, but we'd also need to do something with the pixel shader. Formally, it's not required, since we can run with disabled rasterizer. But really, we use it in 99% of cases, thus it would be noisy to always enclose the pixel shader into `Some()`.

csherratt:
That's the allure of the builder patter when you have lots of optional fields.
```rust
Shader::new().pixel(my_pixel_shader).vertex(my_vertex_shader);
```

kvark:
Doesn't look like we have an obviously superior solution here, unfortunately.
`ShaderSet::Simple(my_vertex_shader, my_pixel_shader)` looks nice to me
Let's proceed, see if we can get any closer to PSO and removing of the shader program and batch concepts.
@homu  r+

kvark:
I have a slight dislike for builders since they introduce yet another level of indirection (or layer).

PR #818, opened 2015-11-04
Use locked versions for dependencies. Add fix for libc breaking changes.

mitchmindtree:

PR #812, opened 2015-10-28
Fixed warning about unecessary parenthesis

bvssvni:

PR #822, opened 2015-11-06
Publish 0.8 w/ the libc fix

csherratt:
First set of cherry-picked changes to backport onto 0.8.

Includes:
- https://github.com/gfx-rs/gfx/pull/812
- https://github.com/gfx-rs/gfx/pull/818

### DIFF
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -15,7 +15,7 @@
 [package]
 
 name = "gfx_device_gl"
-version = "0.6.0"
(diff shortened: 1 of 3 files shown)


csherratt:
@homu r+

bvssvni:
It is a good idea to increase the first non-zero number to avoid breaking existing code.

bvssvni:
Maybe we should yank this version? It causes problems downstream because Cargo tries to link two different versions of libc. We have to wait for the lower level libraries to upgrade, which might take some time.

PR #825, opened 2015-11-08
Not require trailing commas in macros

Binero:
Macros no longer need trailing commas this way.

### DIFF
--- a/src/core/src/macros/mod.rs
+++ b/src/core/src/macros/mod.rs
@@ -17,7 +17,7 @@
 #[macro_export]
 macro_rules! gfx_vertex {
     ($name:ident {
-        $($gl_name:ident@ $field:ident: $ty:ty,)*
+        $($gl_name:ident@ $field:ident: $ty:ty),*
     }) => {
         #[derive(Clone, Copy, Debug)]
         pub struct $name {
@@ -57,7 +57,7 @@ macro_rules! gfx_vertex {
 #[macro_export]
 macro_rules! gfx_parameters {
     ($name:ident {
-        $($gl_name:ident@ $field:ident: $ty:ty,)*
+        $($gl_name:ident@ $field:ident: $ty:ty),*
     }) => {
         #[derive(Clone, Debug)]
         pub struct $name<R: $crate::Resources> {


Binero:
Seems like travis broke.

kvark:
@homu r+

Binero:
The travis fail seems to be unrelated to this commit. It's a conflicting dependency; this commit did not touch any. One of the dependencies might have updated.

kvark:
Might be related to #826. @csherratt, is that correct?

csherratt:
The build break is a new issue (but probably related to `libc 0.2`). `glutin` imports a library  `tempfile` that links to `kernel32-sys 0.2`, `glutin` itself links to `kernel32-sys 0.1` so there is a conflict. I don't think there is anything we can do to fix out build until `glutin` is fixed.

@bvssvni submitted an issue to try and fix `tempfile` https://github.com/Stebalien/tempfile/issues/17 But tomaka might have another idea of how to work around this issue. He's very good at keeping `glutin` working.

csherratt:
I did a local verify of this change and It looks good, but is incomplete as the tests and examples still need to be updated.

I did a quick change to fix it on my local branch (to verify your changes), you can use the change or do it yourself.
https://github.com/csherratt/gfx-rs/tree/pr825

kvark:
Since this would break on the trailing commas, I'm against the change.
@homu r-

Binero:
Maybe add a match arm to allow both?

kvark:
why would you not want to have a trail comma, anyway?
the macro usage is supposed to reflect regular struct declarations, which are recommended to have trailing commas according to rust coding guidelines (IIRC)

Binero:
Going from the guidelines, there is no rule regarding that yet. The proposal says commas should only be printed when the bracket is on a separate line.

https://github.com/rust-lang/rust/blob/219eca11b044de3644b3e9101124513c1a842b09/src/doc/style/style/braces.md#L62

I do realise this is not the most important feature in the world though.

kvark:
@Binero right, I didn't realize that some may prefer one-liner declarations for the structs. I don't think this would look good with `gfx_vertex!` and such... Hence, closing this.

Issue #698, opened 2015-04-24
Find better names for the core concepts


kvark:
`Device`: the same HW device (video card, GPU) can support multiple APIs. Conceptually, `Device` was meant to be an API backend originally. Later on, it lost all the resource management and became just a thing executing command buffers, for which `Device` is not the worst name. Candidates:
- `GraphicsDevice`
  - (+) less changes, i.e. local variable names could still be `device`.
  - (-) confusing with regards to Metal `MTLDevice`, which it is not
- `GraphicsQueue`
  - (+) consistent with Metal/Vulkan, AFAIK
  - (-) it might be a queue on the driver side, but from our point of view it's a sink. Whatever goes in is executed from our perspective.

_Note_: regardless of the `Device` rename, all the `gfx_device_*` crates should be renamed to `gfx_backend_*`.

`Renderer`: not actually rendering anything, but fills up the command buffer. Candidates:
- `GraphicsEncoder`
  - (+) consistent with Metal
- split the resource management aside and deprecate this one in favor of `Stream`

_Note 2_: The prefix `Graphics` here is to indicate the graphics work. When we introduce compute shaders, they have nothing to do with batches or graphics queues. They will need their own queue & encoder.

`Stream`: not actually streaming much, but rather queuing stuff, same as `Renderer`, but with a given target. Candidates:
- `RenderTo`/`EncodeTo`
  - (+) shows that it's derived from `Renderer`/`Encoder`

`Batch`: not related to actual "batching" of anything. As it turned out, `gfx_scene` still uses a bigger structure as a unit of sorting/rendering. Candidates:
- `Unit`
- `Object`
- `DrawCall`
- `PipelineStateObject` (#815)

Please suggest your candidate names!
cc @csherratt @fkaa @bvssvni @cmr

bvssvni:
:+1:

fkaa:
Agree with the `Device -> Backend` for repos. I think `Device` terminology is pretty analogous with the hardware definition. The only difference between gfx's `Device` and Metal's `MTLDevice` right now as I see it is that gfx has decoupled resource creation. The flow of operations in Metal from how i understand it goes like this:
```
MTLDevice <- MTLCommandQueue <- MTLCommandBuffer(s) <- MTL*Encoder(s)
```
[...]

Issue #311, opened 2014-08-29
Match shader samplers type against user-provided texture/sampler pair


sectopod:
Need more safety here! Optimally, we'd like to match at the time user sets up the parameter, not at the draw call itself. This affects MSAA #308. AFAIK, the current approach is to have draw calls unfailable.

bvssvni:
Will this allow changing texture filter at drawing? In the Piston 2D back-ends we would like to choose between linear and nearest filter using the same program.

kvark:
This issue is only about safety checks, not new abilities. Changing the filter mode at drawing is, of course, allowed (see [my other response](https://github.com/PistonDevelopers/gfx_graphics/issues/272#issuecomment-162395962)).

kvark:
Not happening in pre-ll

Issue #284, opened 2014-08-25
[pre-ll] Add support for shader images (UAV objects)


kvark:
GL [specification](http://www.opengl.org/registry/specs/ARB/shader_image_load_store.txt), available on GL-4.2 and DX-11. This is needed for advanced shader model 5.0 techniques, such as forward+ rendering.

dylanede:
What's the status of this now?

kvark:
@dylanede are you looking into using it?
We haven't made any progress on this front.

dylanede:
Well, I'm looking into eventually helping adding support for it, when I have the time. I've got some project ideas that depend upon this.

kvark:
I believe we've had support UAVs for a long time by now.

Issue #161, opened 2014-08-01
Customize texture views for type safety


kvark:
Currently, all the textures are given to the user as `TextureHandle` which bears no compile time (or even run-time) information about the kind of view that the texture is. This has implications:
- shader parameters: we are unable to ensure the types upon connecting parameter struct, the user is free to break the contract at any point later on
- render targets: we could enforce dimensionality match in compile time

Roughly, we could have it done with phantom types:

``` rust
pub struct TextureView<Dim, Type>(TextureHandle);
pub enum Dim2D {}
pub enum Dim3D {}
pub enum TypeFloat {}
pub enum TypeInt {}
...
fn create_texture_2D(...) -> TextureView<Dim2D, TypeFloat> {...}
```

It is unclear though how to separate dimensionality from the base type. For instance, render target doesn't care (yet) about the base type, it only needs a specific dimension. Texture creation as well cares (so far) only about the dimension. Thus, we can implement only the dimension first, and see how it goes.

fkaa:
Would this be using [ARB_texture_view](https://www.opengl.org/registry/specs/ARB/texture_view.txt) when available or is the struct name unrelated?

kvark:
@fkaa I haven't looked into this extension yet, though it does seem to be related.

kvark:
We are using the `ARB_texture_view` now, but overall this issue is no longer needed, given the deprecation of pre-ll.

PR #828, opened 2015-11-09
Pipeline State Objects revolution

### LINKED ISSUE #121: Describe our mesh construction interface in a blog post

### LINKED ISSUE #322: Remove slice from Batch and add .draw_slice
The draw method could use `mesh.as_slice()` by default.

This will make it simpler to render a part of the mesh without having to mutate the batch state.

### LINKED ISSUE #463: Support for specifying index of fragment shader outputs.
The library needs a mechanism for assigning indices to the outputs of the fragment shader outputs. This is necessary when having multiple outputs.

For GLSL < 1.3, we could use gl_FragData[i] for this.
Later on (not sure versions), outputs could be tagged with layout(location = i).

There are some versions in between that can't select indices out the outputs from inside the shader.

I propose we extend the shader! macro to support this construct to flag the shader outputs:
GLSL_150: (["o_Color", "o_Normal", ...], b"out vec4 o_Color; out vec4 o_Normal; ....")

The library will then make sure o_Color is mapped to index 0, o_Normal to index 1, and so on.

The old syntax is still available for earlier GLSL versions and for shaders without fragment outputs.

Comments?

### LINKED ISSUE #571: Make `VertexFormat::generate` to return an iterator
This would avoid an unnecessary allocation, also improving the ergonomics in case the user extends a mesh with multiple vertex buffers.

### LINKED ISSUE #575: Support for ShaderParam multiplexing
It might be desirable to split shader parameters into semi-independent groups to improve de-coupling of systems. For example, in higher-level code (#125) we might have a drawing technique providing its own parameters as a struct, while the object material bears another parameters structure. When drawing, these would need to be combined.

As far as I see, we can not do this:
```rust
impl<A: ShaderParam, B: ShaderParam> ShaderParam for (A, B) {...}
```
... because calling `create_link` on either `A` or `B` would require the coverage of all used shade parameters, which is not the case. Perhaps, we could re-design `create_link` to be more composable.

### LINKED ISSUE #630: Detach clear() semantics from the Frame
Depends on #161. Deprecates #141.

Asking the renderer to clear a framebuffer is the old-gen way to clear surfaces. There are several issues with that:
  - attachments can have different types. Clearing a RGBA8 (unsigned normalized) target with an integer value is a logical mistake we'd like to disallow.
  - there can be multiple color attachments, which requires expanding `ClearData` structure and making it less ergonomic.
  - attempt to clear the depth (for example) when there is no depth attachment is (again) a logical error.
  - clearing is not a quad drawing underneath anyway. It may involve running a compute shader on a meta-data sub-surface, for example. As a consequence directX has clearing on the [target view level](https://msdn.microsoft.com/en-us/library/windows/desktop/bb173539%28v=vs.85%29.aspx).

When #161 is done, we'll be able to enforce the type of the value to clear with at compile time:
```rust
renderer.clear(color_int, 2u);
renderer.clear(color_rgba, (0f32,0.0,0.0,1.0));
renderer.clear(depth_stencil, (0f32, 64u));
```
The drawback is the need to call `clear()` on every surface, but considering the killed `ClearData`, it's going to look way better. The device implementation is free to cache sequential `clear()` calls, but I expect it to be easier to just clear the surfaces one by one as they come. It's not like `clear()` is often a bottleneck, and we can always improve it later while preserving the interface.

Potentially, we can (and should) do the same for blitting.

### LINKED ISSUE #640: Mesh trait
Turning `Mesh` into a trait would provide several benefits:
- can be implemented automatically for `Buffer<T>`, where `T: VertexFormat`, since our buffer info has the size, so we can use it as the number of vertices
- allows mesh proxies without re-allocation of attributes on the heap. For example, an instanced mesh can be a proxy of the per-vertex mesh and per-instance mesh data.

### LINKED ISSUE #694: Streamline the need for R generic parameter to shader_param
Constructs like this make me sad:
```rust
#[shader_param]
struct Param<R: gfx::Resources> {
   x: i32,
   _dummy: PhantomData<R>,
}
```

It would be nice to avoid the need to have `PhantomData` there.

### LINKED ISSUE #703: Create a core library
Thinking of the future I would like to have a core library that contains only the things needed for writing a back-end, such that we could experiment with multiple designs without blocking another. This will also make it easier to release newer versions of Gfx that uses the same core library as older versions, meaning that if some code only depends on the core it does not have to break with major versions.

### LINKED ISSUE #722: Check the layout of uniform buffers
Implementation hint:
- Backends should query the layout of constant buffers, e.g. GL should do this in `query_blocks()`.
- PSO linkage should check for the format to match in `link_constant_buffer()` implementations.

### LINKED ISSUE #761: Memory views
Related to #161, #630

On the lowest level, things like textures or buffers narrow down to chunks of VRAM. What user operates with is some kind of view into these chunks. For example, the same memory could be treated as a texture and a render target given the same tiling mode (compatible with rendering).

Different APIs provide different degrees of freedom w.r.t. memory views. We need to figure out the model we can expose without limiting the use cases, while adding some type safety on the way (*typed* views).

Here is a short recap of what different APIs expose. To be filled with your help!
  - *OpenGL*: textures and buffers are entirely unrelated things. You can't get multiple views into the same memory.
    - `ARB_texture_buffer_object`: allows viewing a buffer through a 1D texture
    - `ARB_texture_view`: allows textures to be viewed via different formats
  - *DirectX10*: there are generic resources (`ID3D10Resource`). One can create multiple shader views (`ID3D10ShaderResourceView`) as well as target views (`ID3D10RenderTargetView`), but the flags enabling these have to be specified upon resource creation.
    - *DX11* also adds UAV views (`ID3D11UnorderedAccessView`) following the same scheme
  - *GNM*: (basically described in the beginning) one allocates memory and treats it as they wish. The compatibility of the views is partially checked by the (optional) validation layer, but the user can still screw up rather badly (by using different sizes, for example).
  - *Vulkan*: TODO

Overall, I believe DX has the most sane model of all, and we should reflect it. Shader resource views should be allowed in unlimited quantities, while the target view should be lifetime bound to the object, thus preventing two writable views from existing simultaneously.

Summoning all the collective mind of @csherratt, @cmr, @fkaa, and even @tomaka from the other side of the moon!

### LINKED ISSUE #815: Pipeline State Object

### LINKED ISSUE #817: Merge RT output with the related states
For a render target setup, we can have depth attached or not. For the state setup, we can have depth test enabled or not. These come as separate things right now (`Output` trait and `DrawState` struct), and the checks are performed at run-time for each draw call we send. Not a big deal performance-wise, but still requires redundant information to be passed in, which is something I'd like to avoid.

Here comes the `OutputMerger` - a struct to replace `Output` trait and contain the relevant states tightly coupled with the surfaces. Will eventually deprecate `Plane` concept as well. Something like that:
```rust
struct OutputMerger {
    depth: Option<(RenderTargetView<Depth>, DepthState)>,
    stencil: Option<(StencilTargetView<Stencil>, StencilState)>,
    colors: Vec<(RenderTargetView<Float4>, TargetBlendState)>, // TODO: heterogeneous MRT?
    state: CommonBlendState,
}
```
This is required by #815. Edit: this is not an obvious dependency. PSO contains the count and formats of the outputs, but *not* the output views themselves. Looks like an opportunity to use some generic parametrization magic - encode the output formats as generic parameters of PSO, since they are typically known at compile time anyway.

### LINKED ISSUE #823: Remove vertex array objects (VAO) from the core
It doesn't look to exist anywhere outside of GL, and isn't that efficient in GL anyway. Internally, our GL backend can still use VAOs if needed.
This functionality is partially covered by PSO (#815), which will define all the vertex layout but not contain the pointers to the actual buffers.

### LINKED ISSUE #842: Where to cache the graphics state?
### How this works now
`Renderer` keeps the `RenderState` and tries to do all the state checks before calling the `CommandBuffer` functions (like `bind_target_surface`). The logic behind it is to minimize the number of commands, and the assumption is that each call to the `CommandBuffer` will result in a separate `Command`.

Now, about this `Command` thing. It's not really used anywhere within `gfx_core`. It is meant to be a convenience serialization format for API backends that don't have their own command buffers exposed (like GL or DX11). So, in a way, it should live in either the GL backend itself (with a potential copy-paste into other backends later on), or in a separate crate (`gfx_command` sounds good) to be used by both GL and DX11.

### The new problem
With the upcoming PSO revolution in #828, the `CommandBuffer` interface will shrink greatly (yay!). The new methods will be about binding PSO objects and their data sets. There is going to be not much opportunity for the `Renderer` to cache. E.g. it will not be able to skip assigning individual buffers any more. Here comes the question in the subject.

One obvious choice is to cache the state on the backend itself. It knows about the granularity of the state, and it can do the finest caching possible. However, it would be happening during the execution as the opposite to the command encoding time (as it is now). The execution time is much more expensive, since it's not possible to execute in parallel.

### Proposed solution
Let's cache everything at the level of `CommandBuffer` implementation! The serialized commands do not have to come 1 to 1 with the calls to the command buffer. The particular implementation can split a fat PSO binding call into a number of smaller states (bind program, assign vertex formats, bind targets, etc). With the knowledge of fine-grained commands comes the opportunity for caching, comparable with what we have now. It's also going to be cheap due to the fact that the command buffer is potentially encoded in parallel. Yet, it will be perfectly compatible with PSO.

I believe that the difference between GL and DX11 is big enough for the `gfx_command` to be unfeasible. Instead, we should cut out all the `Command` logic from `gfx_core`, as well as the state caching from the `Renderer`, and port it to the new command module of the GL backend. As a bonus, we'll have more lightweight `gfx_core` and `gfx` crates with faster compile times, hopefully.

sectopod:
Requires https://github.com/gfx-rs/draw_state/pull/34
Fixes #815, fixes #842, fixes #823, fixes #761, fixes #722, fixes #703, fixes #694, fixes #463
Closes #817, closes #640, closes #630, closes #575, closes #571, closes #322, closes #121
Touches #698, #331, #311, #284, #161

Breaking changes:
- `Batch` concept is gone, PSO is the new paradigm
- `Mesh` is gone, but `Slice` is still with us, even if slightly modified
- shader parameters are superseded by the PSO configuration
- VAO, FBO, `Surface`, `Plane` are all gone, they are now hidden GL implementation details
- Users are required to operate on the strongly-typed target/resource views

TODO (note: this is a vastly incomplete list of what's actually done):
- [x] PSO resource - low level
  - [x] resource
  - [x] binding
  - [x] creation
- [x] High level
  - [x] `RenderTargetView` and `DepthStencilView`
    - [x] low-level bindings
    - [x] raw creation
    - [x] typed creation
    - [x] hook up to PSO
  - [x] `ShaderResourceView` and `UnorderedAccessView`
    - [x] low-level bindings
    - [x] raw creation
    - [x] typed creation
    - [x] hook up to PSO
  - [x] move core into a separate lib (#703)
  - [x] remove the Copy constraint from the resources
  - [x] layout matching
      - [x] vertex attributes
      - [x] shader data
      - [x] pixel targets
  - [x] strongly-typed initialization
  - [x] hook to the `Renderer` for drawing
- [x] helper macros
  - [x] structure definition
  - [x] PSO
  - [x] PSO with static init
  - [x] texture format
  - [x] macros tests
- [x] Universal data format spec
    - [x] serialized form
    - [x] strongly-typed form
    - [x] replace `tex::Format`
    - [x] replace `attrib::Format`
- [x] GL backend
    - [x] rewrite `init()` function for windowing backends
    - [x] rewrite `CommandBuffer` implementation to do the heavy lifting
    - [ ] ~~support for structured buffers~~ (can be postponed)
    - [x] Glutin window
    - [x] GLFW window
- [x] Switch the examples to use PSO
  - [x] blend
  - [x] cube
  - [x] deferred
  - [x] flowmap
  - [x] instancing
  - [x] mipmap
  - [x] performance
  - [x] shadow
  - [x] terrain
  - [x] triangle
  - [x] ubo_tilemap
- [x] Clean up
  - [x] remove old `Resources` subtypes
  - [ ] ~~remove shader programs~~
  - [x] remove old meshes and attribute formats
  - [x] remove old shader parameters
  - [x] remove old texture format
  - [x] remove batches
  - [x] remove VAO and FBO
  - [x] remove `Plane`, `Output`, and `Frame`
- [x] Post-clean
  - [x] rename `NewTexture` to `Texture`
  - [x] remove `primitive` from `mesh::Slice`
- [x] Documentation
  - [x] typed PSO
  - [x] universal format

Or, basically, rewrite the whole damn thing ;)


csherratt:
Sorry @sectopod for creating a conflicts.

sectopod:
@csherratt np!
Heads up - I'm going to change `ShaderParam::Resources` to be a generic parameter rather than an associated type. This would technically allow the same `ShaderParam` to be used with different devices in the same program.

sectopod:
I'm going through several iterations internally, and the progress with this PR is slow but steady. If you need to publish a new version for some other changes - please don't wait for me.
The way this PR is heading now changes quite a bit of perspective in these areas:
  1. Shader resources are serialized into data (`LinkMap`) and passed to the device for program linkage. This is the opposite from what we have now: program linked separately, `ProgramInfo` generated as a serialized form of shader demands, and then they are generically matched against user data.
  2. The concrete resource types, like vertex attributes, have their layout matching logic inside the hidden type (akin `VertexFormat`). This type has the names and formats of the data at compile time (static methods), and the actual linkage to the shader expectations at run time (index data). Vertex structure itself becomes purely data with no logic. This is quite different from the current design, where the generate `Link` structure is just data, and `Vertex` itself has all the logic.
  3. The PSO defines how the data is processed (inputs, outputs, algorithms, parameters, state), but doesn't have the data itself (unlike our old batches). There is going to be less run-time checks during rendering and stronger type guarantees. Prototype PSO definition:
```rust
pub struct PipelineState<
    R: d::Resources,
    I: VertexLayout<R>,
    P: ShaderParam<R>,
    E: PixelLayout<R>
>(d::handle::RawPipelineState<R>, d::PrimitiveType, I, P, E);
```

sectopod:
Why do we even separate vertex attributes from pixel targets from all the other shader data? If we follow #833, we can just define all the shader interface in a single struct like this one:
```rust
gfx_shader_link!( Shader {
    vertex: VertexBuffer<Vertex>, //regular vertex buffer, where `Vertex: VertexFormat`
    vertex_inst: VertexBuffer<Instance>, //instanced buffer, the instance rate is specified somewhere inside the `Instance` struct definition
    const_locals: ConstantBuffer<Local>, //constant buffer
    const_globals: ConstantBuffer<Global>, // another one
    tex_diffuse: TextureView<Dim2, Float4>, // 2D texture SRV
    tex_normal: TextureView<Dim2, Float3>, // same, but of `Float3` type
    sampler_linear: Sampler, // just a sampler, not sure if it would make sense to typify further here
    buf_noise: BufferView<Int4>, // structured buffer SRV
    buf_frequency: UnorderedView<Dim2, Int>, // 2D UAV buffer of `Int`
    pixel_color: TargetView<Dim2, Float4>, // output render target
});
```
Of course, each of the vertex buffers need to have a special struct behind them too (like the current `gfx_vertex`), but still it looks rather sleek to me.

kvark:
@sectopod great idea! so we are going to include attributes and targets into a struct what used to be `ShaderParam`. This would make all the `Output`, `Frame`, and even some `Mesh` stuff irrelevant and simplifies the user side of things: for a draw call, we'd only need a slice, a PSO, and this shader link block. It also simplifies strongly-typed PSO definition: it's only parametrized by the `Resources` and this `ShaderLink` type (maybe `ShaderData` would be better? or `PipelineData`?).

Overall, I think you are not just on the right track, but onto something big. Having all these `*View` types, strong PSO, unified shader link, and with all the old GL cruft removed - gfx would become a very different beast: better suited for next-gen APIs, safer, and friendlier to the users. Keep up!

sectopod:
It took me a fair bit of persistence to implement this scheme, and it could work if finished (got several large commits locally, hesitating to push yet). The reason I stopped is because I feel that this design is incomplete.

Take the blend state for example: providing it separately in the `DrawState` struct, one per color attachment, has no link to the actual target views we provide in this shader link structure. This is not just redundant, it's also a failure to allow using a specific blend mode on a specific target. Similar, even if less severe, is the situation with the depth/stencil - it's possible to provide a depth view without a depth test and the opposite, at run-time, which is sub-optimal.

A logical conclusion for me is to somehow provide the blending, depth state and offset, and stencil state as a part of this [shader link](https://github.com/gfx-rs/gfx/pull/828#issuecomment-156771805) structure. This would mean we no longer need to pass `DrawState` for PSO creation, but instead some sort of extended `Primitive` configuration that would include the topology and multi-sampling (I'd call it `Rasterizer`). However, there is one big problem here - blend/depth/stencil can not be a part of this structure, since it only has the data needed for run-time execution. The configuration for PSO creation is derived from the types themselves (in the current iteration of the design). 

Basically, this shader link has 3 different structures underneath:
 1. PSO initialization structure. It has the rasterizer configuration, blend states, depth/stencil, and the shader names of all the fields. It's currently partly raw-data-driven (`create_pipeline_state(..., &DrawState, ...)`), partly user-defined at compile time (shader symbols), and partly implicitly derived from the types of the declaration fields (e.g. `RenderTargetView<R, T>` derives the target format from `T`).
2. PSO meta-data. It has all the register indices for the used fields, constructed based on the shader reflection. The structure is generated from the field types and managed automatically (though we still need to specify the name because of the current Rust macro system limitation - inability to create new identifiers).
3. PSO run-time data. It has the buffer/view handles of the stuff we are binding on each draw. The structure is the one we are providing to the shader link macro.

So, going back to blend/depth/stencil - we are at the crossroads. Here are some extreme routes we can take, or maybe we can figure out a reasonable compromise in-between (if only anyone would be able to follow my essays...).

1. Extreme types - provide all the initialization data with types. This would mean compile-time fixed depth/blend/stencil states, which is a rather strict limitation. It would also rely heavily on our ability to support more generic versions of the shader links in the macro, which would at least allow some variability of these states. It would also involve a bigger gap between the types we pass at run-time and the types of the declaration, which means more boilerplate code by the user...

2. Extreme data - provide all the init data with data. It could be achieved by generating this PSO init struct by our macro and requiring the user to fill it up and pass in for PSO creation. This structure would contain the symbol names too, so it would provide the highest flexibility we can ever get without compromising the strong types safety.
The downside here is mainly the mental load on the user - now they've got to think about this init structure as well, which format they don't observe visually and can only guess (with compiler help) upon filling. For example, `RenderTargetView<R, T>` in the run-time declaration would imply something like `(&'a str, Option<Blend>, ColorMask)` it the init struct.

Any ideas/opinions on how to proceed? Am I even nearly close to an efficient design here? Without further instructions I'll go with extreme data path to see where the rabbit whole leads...

kvark:
@sectopod nice to see steady progress! It's exciting to see how far we can reach with this shader link concept in terms of safety and usability. BTW, I vote to rename it to `PipelineData` (and `gfx_pipeline_data` correspondingly).

As for your main question, I believe a reasonable compromise between data and types w.r.t `DataLink` would be to have as little types as needed to ensure no possible errors happening at run-time binding. Blend state, for example, is impossible to screw up at run time, so we can have it as data. Symbol names - too - should be data. SRV/DSV/RTV formats though have to match, so we should be having them typed.

kvark:
Ok, I've been thinking about it some more... I still don't know *how* to get this right, but at least I know how to definitely get this wrong ;)

First off, using types only for PSO initialization (`Extreme types`) is a no-go. It harms the flexibility too much (inability to tweak the blend state, depth state, or even the shader-exposed name).

Secondly, generating the PSO init struct in a macro and then expecting the user to instantiate it is another no-go. Everything that user is supposed to create should be either explicitly documented, or declared by the user themselves.

One possible research direction that is left after cutting these paths is - expect the user to provide both init and bind parts explicitly. Reminder: the "init" part is the one passed for PSO creation, the "bind" part is the one provided for each draw call.
```rust
color: gfx::RenderTargetView<R, Float4> | (&str, gfx::BlendState, gfx::ColorMask),
```
This will generate the `color` field in all init/meta/bind structs, where the user will need to provide the `RTV<Float4>` for binding, and `(name, blend, color_mask)` for initialization. If rust typesystem is flexible enough, we'll be able to enforce the `Float4` to be a blendable format too. In other words, we should use the init types to help define what the expectactions for the bind ones are: having a depth state means our DSV should have the depth component, having the blend state means our RTV should be blendable, etc.

kvark:
Another idea. If we don't take the shader variable names into account, the only data that we need to pass for the PSO creation is related to render targets: blend states, color masks, target formats, depth/stencil. This fact makes the targets somewhat different from the rest of our magical struct. Also the fact that there can be only one depth/stencil view, while there can be any repeating number of other types of entries. Perhaps, we should not include the render targets into this struct then? Treating them separately would make the rest of the struct clean enough and reduce the confusion. For the targets, we can use nominal types (tuples) to avoid user-specified names.

sectopod:
@kvark hmm, looks like there is no clearly superior solution in sight. Specifying both the bind data and init data types is too verbose to my taste (I'd rather have the some staff magically derived). Moving render target stuff out would complicate the interface... 
I don't know what to do yet, but I tried to systematize all the data that we need to support for PSO magical struct, minus all the rasterizer stuff. Hope this will help:

prototype 							| init data 				| meta	| bind data
----------------------------------------------------- | ---------------------------------- | ---------- | ----------------------------
`VertexBuffer<T: Structure>`			| `(T::init, InstanceRate)` 	| [v#]	| `BufferHandle<R, T>`
`ConstantBuffer<T: Structure>`			| `(T::init, &str)`			| c#		| `BufferHandle<R, T>`
`Constant<T: Uniform>`				| `(&str, T::format)`		| loc		| `T`
`Texture<T: TextureFormat>`			| `(&str, T::format)`		| t#		| `ShaderResourceView<R, T>`
`Sampler`							| `&str`					| s#	| `Sampler<R>`
`UnorderedBuffer<T: BufferFormat>`	| `(&str, T::format)`		| u#	| `UnorderedAccessView<R, T>`
`RenderTarget<T: RenderFormat>`	| `(&str?, T::format, ColorMask)`		| index	| `RenderTargetView<R, T>`
`BlendTarget<T: BlendFormat>`		| `(&str?, T::format, ColorMask, BlendState)`	| index	| `RenderTargetView<R, T>`
`DepthTarget<T: DepthFormat>`		| `(T::format, DepthState)`	| 	| `DepthStencilView<R, T>`
`StencilTarget<T: StencilFormat>`		| `(T::format, StencilState)`	| 	| `DepthStencilView<R, T>`
`DepthStencilTarget<T: DepthStencilFormat>`	| `(T::format, DepthState, StencilState)`	| 	| `DepthStencilView<R, T>`

sectopod:
I believe I solved the main problem with PSO internal type hierarchy and getting it all defined by helper macros. The biggest issue right now is generating the identifier names in those macros. Since `concat_idents!` is useless, I have to demand all the names provided by the user.

I tried a different approach - to create a new internal module with a user-given name and then just use some predefined simple names inside. However, this turns out to be playing very badly with user-given parameters, which the user expects to be visible, while jumping into a sub-module instantly changes the visibility scope...

As always, any ideas are welcome. It would be nice to get this right from the start, since it affects the API verbosity and complexity.

sectopod:
All the PSO support is complete, including the strongly-typed views/textures and the unified data format. It turned out I could re-use much less stuff from our previous implementation, so it more and more looks like a rewrite of the whole library. Of course, there is more stuff to polish (especially on the GL backend side), but I think we are ready to proceed to phase 2 now: porting the examples.

sectopod:
Here is another update. We are at the end of phase-2: all the examples are now ported, minus some stuff left to fix in the mipmapping example. Moreover, the GL backend got improved massively, with the new format specification supported natively there. So, in other words, no rough spots there.

You can see how the examples became sleeker. There is less magic wrappers and more direct control (e.g. a window is no longer held inside a stream). Some new ergonomic hotspots emerged though, which need to be thought of:
- `gfx::VertexBuffer` initialization is almost always used with `PER_VERTEX`. Optimally, we'd want a separate `gfx::instanceBuffer` with no init data for either. This seems relatively easy to address.
- `gfx_pipeline_init` is used everywhere, and the need for `Init::new()` call when PSO is created. Not sure if we can do much outside of renaming stuff.
- `gfx::RenderTarget` always needs a color mask. We could have a separate `MaskedTarget` for that, but this solution faces a bit of an implementation problem (of how to reduce code duplication in `render/pso.rs`).

The end of phase-2 means we can now start cleaning up all the old stuff (phase-3), at last! This should not only be fun, but also liberating - to improve the compile/link speed, which drags down my productivity at times (especially it used to, before I moved the core into a separate crate).

One thing needs to be reiterated on though, in the meantime - the types of SRV/UAV. Currently, they are typed by `gfx::format::Formatted`, which includes the surface type. With this implementation, a PSO declaration can't seamlessly accept both `SRV<U8Norm>` and `SRV<f32>`, for example, while this should be perfectly possible. One (quick) solution is to bring up the casting functions into public space for SRV/UAV, so that the user can just cast upon assigning to pso data. Another (better) solution would be to re-think the new format specification and detach the view format from the surface format.

examples/cube/main.rs
```diff
@@ -18,39 +18,45 @@ extern crate gfx;
...
         }
     }
 }
 
-// The shader_param attribute makes sure the following struct can be used to
-// pass parameters to a shader.
-gfx_parameters!( Params {
-    u_Transform@ transform: [[f32; 4]; 4],
-    t_Color@ color: gfx::shade::TextureParam<R>,
+gfx_pipeline_init!( PipeData PipeMeta PipeInit {
```
bjz: It's a shame about this long list of idents... I guess there's no way around it though... :(

examples/performance/main.rs
```diff
@@ -119,21 +136,16 @@ fn gfx_main(mut glfw: glfw::Glfw,
...
-
-                let data = Params {
-                    transform: proj.mul_m(&view.mat)
-                                   .mul_m(&model).into_fixed(),
-                    _r: std::marker::PhantomData,
-                };
-                stream.draw(&batch.with(&slice, &data, &state))
-                      .unwrap();
+                data.transform = proj.mul_m(&view.mat)
+                                     .mul_m(&model).into_fixed();
```
bjz: Still on an old version of cgmath? :(

examples/cube/main.rs
sectopod: Indeed. I've been trying to get everyones attention to this issue, but I suppose it requires a bit too much mental investment to even understand how it works. As it stands now, all 3 new types are (or can be) used directly by the user:
- `PipeData` is always needed to be created since it's required for every call
- `PipeMeta` is needed if one needs to store the PSO somewhere, the full type of which is `gfx::PipelineState<R, PipeMeta>`. Out of my head, `ubo_tilemap` example stores it.
- `PipeInit` is needed for PSO creation.

Perhaps, some of them can be re-routed via others by increasing the internal complexity of the type structure a little bit.

examples/performance/main.rs
sectopod: I bet this PR started before your grand cgmath changes ;)

bjz: Yup! :)

examples/cube/main.rs
bjz: Would be kind of nice to have anonymous structs for associated types here:

```rust
impl<'a> PipelineInit for Foo<'a> {
    type Meta = struct {
        ...
    };

    type Data = struct {
        ...
    };
}
```

Perhaps you could use modules here:

```rust
gfx_pipeline_init! {
    mod pipeline {
        ...
    }
}

// expands to:

pub mod pipeline {
    pub struct Init {
        ...
    }

    pub struct Meta {
        ...
    }

    pub struct Data {
        ...
    }
}
```

sectopod: Thanks for trying to tackle this problem! I iterated on the separate module idea earlier:
> I tried a different approach - to create a new internal module with a user-given name and then just use some predefined simple names inside. However, this turns out to be playing very badly with user-given parameters, which the user expects to be visible, while jumping into a sub-module instantly changes the visibility scope...

However, I didn't consider/try doing `use super::*;` at the top of the new module. Perhaps, this would work? As a side effect it may require dependent types (vertex structures) to be declared as public or re-exported as public.

sectopod: @bjz this worked after all. There are some hoops though, like the need to `pub use` instead of just `use` for the stuff used inside the macro, but overall I consider it a success. Please have a look at the latest code in this PR.

sectopod:
I believe all the ergonomics concerns (at least the one mentioned in this thread) are now addressed. Overall, the code is getting polished nicely, and only a few bits left to do:
- better SRV/UAV typing
- documentation
- somebody needs to actually review this code

sectopod:
Hmm. I implemented better typing for SRV/UAV that allow us to avoid the casts (e.g. in deferred example), but now Rust forces me to explicitly specify the format type for the view creation functions in `Factory`, which is a little annoying. Looking into ways to simplify it to avoid explicit type annotations...

sectopod:
We are green, and my checklist is completed. All systems ready to go!

cmr:
@sectopod awesome work, been really excited about this new model! Do you have any recommendations for going about reviewing this code? 138 commits is a bit long, and I imagine at least some of it obsoleted by later commits. Maybe some strategic squashing?

sectopod:
@cmr thank you for the intent of reviewing the code! I understand it may take a while to scan through all the changes, and I don't think it's worth it. The systems I implemented in this PR went through a couple iterations of evolution, at least. It was one hell of an R&D work. Some of this history is reflected in the actual commits, some is not.

Thus, I'd suggest just looking at the final revision with this in mind:
  1. gfx core has been moved into a separate crate (and now I wonder how Travis succeeded, given this crate is not published?..)
  2. lots of old stuff is removed
  3. `Renderer` is renamed to `Encoder`, module name is also changed
  4. new `Texture` is not an evolution of the old one. It was developed separately and then replaced the old name.
  5. quite some bits of logic have migrated from gfx itself into the GL backend: command definition and `CommandBuffer` implementation, Surface concept, and others.

cmr:
Thanks! I'll spend time tomorrow (ie, after 9 hours from now) reviewing this.

bvssvni:
@sectopod Travis CI probably tested the paths in Cargo.toml, which does not require publishing the crates when building the local project.

kvark:
@sectopod I'm glad you managed to finish this epic task. It started as a little feature, and then turned into a rewrite of most of our old code, and re-design of our concepts from the ground up. At some point, I thought the feature creep would just eat you alive ;)

To recap, we drop all the "artificial" concepts we had (batches, frames, planes, meshes, etc), and replace them with a single yet powerful concept of a `PipelineState` (PSO). It's natural, since it closely follows API specification (DX12 + Metal). The `Texture` is just a pixel storage now, and we introduce the accompanying views (shader resource (SRV), unordered access (UAV)) to hook up to PSO. We also get the universal format specification for vertex buffers and texture surfaces.

Everything exists in two worlds: type-level, checked at compile time, and data-level, checked at run-time. Users are supposed to work on the type-level (strongly-typed PSO, views, formats), which gets automatically transcribed into the data-level to be consumed by the non-generic interfaces of `gfx_core` crate, which are implemented by the API backends.

I've been following the development closely and welcome the changes, very excited to see what we can do with the new architecture! I advice everyone interested in the project to have a look to get more familiar with the implementation. Perhaps, you will have some notes for the review too, in order to back up me and @cmr.

Trebgarta:
It is very nice to see that this is close to being merged. I am interested in this project so I want to ask this: When can I expect this to be merged, and should I expect more breaking changes in the future?

cmr:
I'm going to leave reviews in this comment, and append to it as I find more stuff to comment on.

------
https://github.com/sectopod/gfx-rs/blob/68f1265f52186aec73202c98182c4d24e5f85b54/src/core/src/factory.rs#L37

I'm a bit worried about overflow panics here, although maybe I shouldn't be. Using .wrapping_add and .wrapping_mul might be nice, if this function is always used with "safe" slices, to keep code size down. I'm also worried that it is marked `pub` but not `unsafe`, as in general this is very unsafe - it's equivalent to a transmute! - but is safe on "POD" types.
-------

kvark:
@Trebgarta I guess we are waiting for @cmr to finish the review, and for @sectopod to address the issues that @cmr finds.

> should I expect more breaking changes in the future

It would be a miracle if this single PR was final and stable, especially consider the fundamental property of the work. I want to publish 0.9 once this PR is merged, and then we'll start fixing all the rough corners with (hopefully) more people trying out the new interfaces. Then 0.10 should be published with all the fixes, and it should be stable.

sectopod:
@cmr please enumerate the points outlined so that I can refer to them.
About `cast_slice` - I forced `Copy` bound on the arguments, which should make it truly safe now.
I don't understand the point about overflow and `wrapping_*` usage there. Why would `mem::size_of` overflow ever?

kvark:
@niconii suggested to use the following pattern for our PSO struct macro:
```rust
    ($name:ident [ $($t:tt)* ] {
        $( $field:ident : $ty:ty ,)*
    }) => {
        as_item!{struct $name < $($t)* > {
            $( $field: $ty ,)*
        }}
    };
```
Main tricks are:
  1. Using square brackets (`[` and `]`)
  2. Using bare tokens (`tt`)
  3. Adding `as_item!` on top

It would allow us to have generic parallelization of PSO, which is crucial to get some flexibility for the user, given how strict the current type system bounds are.

sectopod:
@kvark  sadly this doesn't seem to be enough. We need to not only provide the generic bounds in a macro but also refer to complete types. I.e. in addition to `struct A<Q:Qu,W:Wu>` we need to have `A<Q,W>` somewhere, and having a token list doesn't help...

@niconii please see [this macro](https://github.com/sectopod/gfx-rs/blob/e2d1e3fbc0ebbfe39ee62c824deba79dace04e4d/src/render/src/macros/pso.rs#L29) if you are willing to help us further.

kvark:
@cmr are you still with us?

cmr:
@sectopod it shouldn't overflow, that's the point! If you don't use wrapping_add/wrapping_mul, the compiler will generate overflow checking code.

cmr:
Sorry if my review is a bit wide-ranging at this point, and not specific to your changes, it's devolved into a full code review.

cmr:
Left some new comments. After they are fixed, I want to do one last look-over and then :shipit:

sectopod:
@cmr thanks! you found quite a few important bits to fix.
(1), (2), (3), (4) - fixed
(6) doesn't work for me. Doing `cargo fmt` produces: `No such file or directory (os error 2)` followed by the usage info (for the `cargo fmt`). We can do it later on.
(7), (8), (9) - fixed
(5), (10) - left for TODO outside of this PR
(11) - doesn't matter for now, since we are only using VAOs whenever GL forces us to do that. EXT versions don't force it. Maybe if we find a better use for them, we'll need to query for EXT stuff too.
(12) not fixed intentionally. It's not a performance critical code, and using `mem_unitialized!` would only confuse us during any debugging.

cmr:
Who gets the honor of pushing the big green button? :)

sectopod:
homu!

cmr:
Oh right, I forget we use homu now!

sectopod:
I wonder if we should volunterely strip ourselves of the power of the green button. Just looking at it makes my skin goose. Only @homu should guard the star gate to upstream.

cmr:
@homu r+ 736b0c3

sectopod:
@cmr hohoho, it's happening!
btw, why did you select the commit explicitly? It's not like I can sneak anything in between your comment and @homu response

cmr:
habit, nothing personal ;)

sectopod:
@cmr here comes the problem: if I push a fix for beta/nightly and politely ask homu to retry, will it retry the same commit you specified?..

cmr:
Not actually sure...

sectopod:
Well, it should be fixed now. The error was real, and I'm glad that beta/nightly caught it.

cmr:
@homu r+ 3f92272

cmr:
:confetti_ball: :tada: Great work @sectopod

gaudecker:
Oh wow, this finally merged? Maybe I should try to raise Rust Theft Auto back from the dead and try this out :party:

kvark:
Hooray! :fireworks: 
@gaudecker this should be fun ;) Don't forget to check up with us on gitter in case you have any questions.

PR #829, opened 2015-11-09
Remove Arc from gfx-rs

### LINKED ISSUE #824: Use std::sync::Arc
Now that it's got `get_mut` method that is stable, we can get rid of our own version of `Arc`.
cc @csherratt

csherratt:
Fixes #824

This is API breaking so it should be part of 0.9. We will probably want to add a way to do `Ord` for our handles as it is needed for `gfx_scene` (This change will break `gfx_scene`).


kvark:
Well, considering the PSO stuff coming in, `gfx_scene` doesn't have no chance to survive without a bit of re-design anyway ;)
Thanks for cleaning this up!
@homu r+

csherratt:
@homu retry

PR #834, opened 2015-11-18
Fix broken example links

mike-marcacci:


kvark:
The cube example is using GLFW, and everything else is on Glutin now. We need to fix not just the links, but the wording too.

mike-marcacci:
Ah, there you go – also, I don't mind at all if you close the PR and fix it yourself if you have something else in mind! Just getting started with gfx-rs and noticed the dead links.

kvark:
Awesome, thank you!
@homu r+

Issue #827, opened 2015-11-09
Update Travis script to the new infrastructure


kvark:
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

kvark:
Fixed with #835

PR #835, opened 2015-11-26
Update .travis.yml to use new build infrastructure

vickenty:
This should be it for #827.


kvark:
Looks sweet, thanks!
@homu r+

PR #836, opened 2015-11-27
Fix maximum mipmap level

vickenty:
Fixes maximum mipmap level passed to ``set_mipmap_range()`` to be one less than number of mipmaps and adds an example to show that things still work correctly after that.

### DIFF
--- a/examples/mipmap/main.rs
+++ b/examples/mipmap/main.rs
@@ -1,0 +1,176 @@
+// Copyright 2014 The Gfx-rs Developers.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


src/backend/gl/src/tex.rs
```diff
@@ -384,7 +384,7 @@ pub fn make_without_storage(gl: &gl::Gl, info: &TextureInfo) ->
         _ => return Err(TextureError::UnsupportedSampling),
     }
 
-    set_mipmap_range(gl, target, (0, info.levels));
+    set_mipmap_range(gl, target, (0, info.levels - 1));
```
kvark: Do we catch `info.levels == 0` case anywhere?

examples/mipmap/main.rs
```diff
@@ -0,0 +1,176 @@
+// Copyright 2014 The Gfx-rs Developers.
```
kvark: 2015 ;)

kvark:
Thanks @vickenty , looks very clean!
Just a couple of minor points to address.

src/backend/gl/src/tex.rs
vickenty: In [Factory.create_texture()](https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/factory.rs#L267)

kvark:
@homu r+

PR #837, opened 2015-12-01
Remove libc dependency

vickenty:
This patch replaces all occurrences of ``libc::c_void`` with ``std::os::raw::c_void``. This type is required for parts that interact with gfx_gl (where it comes from gl_generator), and the rest was changed to the same type for consistency.


kvark:
thanks!
@homu r+

PR #841, opened 2015-12-05
Ignore built-in uniforms.

### LINKED ISSUE #840: Using gl_DepthRange in a fragment shader errors while creating a core batch
Using gl_DepthRange to set gl_FragDepth in a fragment shader causes building of the core to fail with the error: Parameters(MissingUniform("gl_DepthRange.near"))

vickenty:
OpenGL reports built-in uniforms alongside user defined ones, but they
can not be linked with user data and we can just skip over them.

All built-in uniforms have names that start with reserved prefix OpenGL,
which can be used to distinguish them from user-defined ones.

Fixes #840


kvark:
thanks!
@homu r+

PR #843, opened 2015-12-09
Update README URLs based on HTTP redirects

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein

### GitHub Corrected URLs 
Was | Now 
--- | --- 
https://github.com/bjz/glfw-rs/ | https://github.com/PistonDevelopers/glfw-rs 

### HTTPS Corrected URLs 
Was | Now 
--- | --- 
http://en.wikipedia.org/wiki/Fixed-function | https://en.wikipedia.org/wiki/Fixed-function


kvark:
@homu r+

PR #845, opened 2016-01-08
[ARM] Fix compilation on nightly rust 1.7

petevine:


kvark:
@homu r+

sectopod:
@homu retry

sectopod:
Ok, for some reason my #828 passed through Travis just fine without this PR merged. Maybe it's unnecessary at the end of the day? Now, my PR can't be merged automatically due to a conflict :( Perhaps, we could revert this one, then merge mine, if this one is not needed?

petevine:
Go ahead, the change is trivial and can be applied later. 

It's definitely necessary on arm unless you're going to switch to std::os:raw. I think you're forgetting travis tests on x86_64.

PR #846, opened 2016-01-08
Documenting gfx_vertex macro.

Trebgarta:
First try at documenting gfx_vertex macro.

I am very new and I don't know what I'm doing very well, just wanted to contribute this cause docs are important.

So what do you guys think? Could I pull it off?

### DIFF
--- a/src/core/src/macros/mod.rs
+++ b/src/core/src/macros/mod.rs
@@ -14,6 +14,27 @@
 
 //! Macros for deriving `VertexFormat` and `ShaderParam`.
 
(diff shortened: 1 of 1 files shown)


kvark:
@Trebgarta thank you for the PR!
Unfortunately, we can't accept it. gfx is currently undergoing a revolution with #828, which is almost done, as far as I can tell. This PR removes `gfx_vertex!` macro, although it adds `gfx_vertex_struct!`. Perhaps, you'd be able to hold this off for now and then port the comments onto the new version when its done? Closing.

PR #856, opened 2016-01-28
fix get_bit_size

boguszaur:
the doc comment says it should return the number of _bits_, but std::mem::size_of returns the number of _bytes_


kvark:
Indeed, thanks!
@homu r+

PR #861, opened 2016-01-31
DX11 Backend

### LINKED ISSUE #89: Add support for a D3D11 backend
I'm working on this. Everything we do so far is D3D11 compatible. I am not checking for D3D9 compatability, having a D3D9 fallback would be a later project.

### LINKED ISSUE #849: Rename `Phantom` to `InnerRaw`
http://gfx-rs.github.io/gfx/gfx/core/factory/trait.Phantom.html

`Phantom` is a generic name with few associations. I would like it to be easier to remember what it does, even if it is not meant for public use.

### LINKED ISSUE #878: Example launcher
Our monogamy with GL is ending. Now we need all the examples to be written in the backend-agnostic and window-agnostic way. I propose to do the following:
  1. Create a launcher crate in this repository to contain:
    - the application trait that is generic over backends/windows
    - configurable facilities to launch an application with any backend/window
  2. Each example then will need to be updated:
    - will declare a `struct SomeExample`, which implements launcher's `Application` trait, and has all the example logic in it
    - the `main()` body just calls the launcher

### LINKED ISSUE #882: Shader version selection is broken
Our current `ShaderSource` type can't be used for multiple backends. First problem - it knows about possible backends (HLSL, GLSL, etc), while this is supposed to be a backend-agnostic module. Secondly, since the capabilities only expose `ShaderModel` type, we don't know which backend variation to choose among compatible shader models... We need something more directly connected to the backend.

### LINKED ISSUE #897: Reconsider where `update_texture` and `update_buffer` should live
Currently they are a part of the `Factory` trait. However, in DX11 the corresponding `UpdateSubresource` is a method of the immediate context, which is represented by our `Device`. Obviously, requiring the device for resource updates may be less convenient for the end users, but we need to check other APIs (Vulkan, Metal, DX12) to see what's the best place for these methods.

### LINKED ISSUE #902: D3D11 Deferred Context
D3D11 has the concept of a Deferred Context in addition to the Immediate one:
https://msdn.microsoft.com/en-us/library/windows/desktop/ff476880%28v=vs.85%29.aspx#Deferred

It's supposed to work similar to Vulkan/Meta/DX12 command buffers, although not used much for some reason. We could have separate `Device` and `CommandBuffer` implementations in our dx11 backend to work with deferred contexts: each command buffer will just wrap one instead of carrying a list of commands. If we are smart enough, we should be able to share the processing logic between these device implementations.

As an option, we could move the command buffer out from the associated type to the generic parameter of the device, and implement `Device<ImmediateCommandBuffer>` and `Device<DeferredCommandBuffer>` for the same structure, which seems convenient.

When this works, we'll be able to profile both immediate and deferred contexts, tell if there is any performance difference. Pinging @sectopod, in case you want to give it a try within #861.

### LINKED ISSUE #906: Rethink mipmap generation
Our previous approach was to have `generate_mipmap` method on the `Factory`. It doesn't fit DX11 model, since even though there is `GenerateMipmap` method, it is a rendering operation, belongs to `ID3D11DeviceContext`, thus can't be executed by a factory, which only has `ID3D11Device`.

It seems straightforward to just move the mipmap generation into the `CommandBuffer` interface, since this is where our rendering is done. However, this doesn't work well for the major use-case - when we have the mip[0] contents right away and we want to upload it with all the mipmaps generated. Not only this is inconvenient, it also involves blowing up the data buffer, associated with the command buffer on GL and non-deferred DX11. This means - time wasted on the copy and allocation.
Edit: problem is, that we can't create a partially initialized texture in DX11/DX12, that's why we'd have to provide mip[0] contents through the command buffer.

As if it's not difficult enough already, DX12 doesn't have any mechanism to assist in mipmap generation (see #905). You are supposed to do it manually. We can, of course, implement some rendering inside the backend, but this rises the question - should we support mipmap generation at all?.. Once you start being serious about graphics, you end up loading all mipmaps from disk in some optimized GPU-friendly format anyway, since generating them takes time.

sectopod:
Fixes #89, Fixes #849, Fixes #878, Fixes #882, Fixes #897, Fixes #902, Fixes #906  

### Breaking changes
  - `TriangleFan` is removed
  - `R8_G8_B8` and `R3_G3_B2` surface types are removed, and with them the old `Srgb8` shortcut
  - scaled channel formats (integers that are seen as floats) are removed
  - explicit mipmap generation is removed. You can tell gfx to generate mipmaps on texture data initialization instead.
  - `Factory::update_*()` methods are removed in favor of `CommandBuffer`-based updates and providing the initial data for resource creation, which now supports multiple slices and mips.
  - device capabilities are reduced by cutting out GL-specific stuff.
  - there is a new `Bind` flag set that you'll need to pass for some of the object construction methods
  - `Usage` enum has undergone a rewrite
  - there is a new `ShaderSet` type that may need for the PSO (or shader program) construction
  - new structures for SRV/RTV/DSV creation: `ResourceDesc`, `RenderDesc`, and `DepthStencilDesc`
  - CommandBuffer interface has minor changes in the scissor and indexed draw semantics, and a major change in the clearing semantics.
  - `SubmitInfo` is gone. `Encoder` has a single `flush` method instead of `as_buffer` and `reset`.
  - `Phantom` is renamed to `Typed`
  - `gfx_constant_struct` requires the shader-visible names now, the same as `gfx_vertex_struct`
  - shader version selection has been moved to `gfx_app` crate. Both core and render crates are now completely API-unaware.

### Notable non-breaking changes
  - there are new `gfx_device_dx11` and `gfx_window_dxgi` crates. We also set up windows CI, which has GLFW stuff excluded.
  - most, if not all, examples are rewritten to use the new `gfx_app` crate. It is now the only place that knows about all the windowing and API backends in this repository.
  - `Border` mode for samplers
  - new convenience type `your_pso::Bundle` and the constructor `your_pso::bundle()` are generated for you to contain the slice, PSO itself, and PSO run-time data objects.

### Examples update
  - most examples simply call `launch_default`, which selects the default backends for the current platform
  - no longer using `cast_slice` - instead calling `create_texture_const_u8`
  - passing in an array of input slices for texture creation - this saves "skybox" example half a page of code
  - parameters are converted to be using the constant buffers for DX11 compatibility. Some are still being mirrored in globals to be compatible with GL2.
  - shader versions for different models/backends are now properly specified and selected
  - "Target0" name for the final PSO outputs at the moment. DX11 doesn't provide us the name, but it has the type to match on, and GL doesn't care. "Target0" is what we translate built-in DX11 semantic into.

### Why DX11?

It's the most established API for high-performance games and graphics application to date. It has the best all-around driver support, variety of powerful tools, and is not cluttered by the legacy (looking at you, OpenGL).

### Implementation details

The backend supports both native deferred contexts and emulated (in the same way we do with OpenGL). Once you obtained a device instance with `gfx_window_dxgi::init`, you can transform it into `gfx_device_dx11::Deferred` with a simple `into()` call. The application launcher uses native DX11 contexts by default.

### Future work

  - Integrate `winit` when it's ready (#907)
  - Implement state caching for the DX11 command buffers


kvark:
Check out `wio` for better COM handling: https://github.com/retep998/wio-rs/blob/master/src/com.rs

sectopod:
Pretty much blocked now on the absence of `D3DReflect` interface... (cc @retep998) The DX backend may still have flaws in it, but there is only so much I can do without really progressing with shader reflection to get some tests going.

retep998:
Sorry about that, I'll go finish it up right now. :rabbit:

retep998:
I've pushed some stuff to winapi just now. Could you use a cargo local override to test the latest git version of winapi and friends to ensure it works and has everything you need before I cargo publish it?

sectopod:
@retep998 thank you!

sectopod:
@retep998 I've finished the first stage - implementing the basis we use for our examples, including shader reflection with d3dcompile. Now proceeding to the second stage - running the examples, debugging the errors, and fixing them. First problem right away:

> note: ld: cannot find -ld3dcompiler
> ld: cannot find -ldxguid

retep998:
Ah, you're using the `-gnu` targets which come with a bundled MinGW that only has a few of the system import libraries, unlike `-msvc` which would work totally fine.
This should fix it: https://github.com/retep998/winapi-rs/commit/e94ee58e67c042356afea1abc9075667ae2b52e5

sectopod:
Thanks @retep998 , this works! I hope it doesn't conflict with `-msvc`.

retep998:
I've set up the build scripts such that they only use the bundled import libraries for `windows-gnu`, so there's no worries of it causing problems for other targets.

sectopod:
@retep998  Now I'm trying to figure out why `D3DReflect` fails. Here is how I call it:
```rust
let mut reflection = ptr::null_mut();
d3dcompiler::D3DReflect(code.as_ptr() as *const winapi::VOID,
            code.len() as winapi::SIZE_T, &dxguid::IID_ID3D11ShaderReflection, &mut reflection);
```
I'm getting `HRESULT = 0x80004002`, which translates to `E_NOINTERFACE`:
> No such interface supported

Maybe the `dxguid` value for `IID_ID3D11ShaderReflection` is incorrect? My card doesn't support DX11 feature set, but it still runs D3D11, so I expect the code to be supported.

sectopod:
There seem to be [D3D11Reflect](https://msdn.microsoft.com/en-us/library/windows/desktop/ff728670%28v=vs.85%29.aspx) in the docs, that would zero out the chance for improper IID, but it doesn't look like the current bindings provide it.

sectopod:
@retep998 found [something](https://github.com/PixarAnimationStudios/OpenSubdiv/issues/775), makes me a sad panda:
> If a client app links against another version of the DX SDK than OSD was linked with (ex. June 2010 and Windows Kit ), then D3DReflect can return a `E_NOINTERFACE` error, with a null pointer.

retep998:
Oh, there wasn't any `D3D11Reflect` in the Windows 10 SDK but there is in the Windows 8.1 SDK. This is a bit confusing.

retep998:
>Maybe the dxguid value for IID_ID3D11ShaderReflection is incorrect?

I just compared the IID provided by 8, 8.1, and 10 of the Windows SDK headers to the IID in `dxguid-sys` and they all look identical.
8: `DEFINE_GUID(IID_ID3D11ShaderReflection, 0x8d536ca1, 0x0cca, 0x4956, 0xa8, 0x37, 0x78, 0x69, 0x63, 0x75, 0x55, 0x84);`
8.1: `DEFINE_GUID(IID_ID3D11ShaderReflection,0x8d536ca1, 0x0cca, 0x4956, 0xa8, 0x37, 0x78, 0x69, 0x63, 0x75, 0x55, 0x84);`
10: `DEFINE_GUID(IID_ID3D11ShaderReflection, 0x8d536ca1, 0x0cca, 0x4956, 0xa8, 0x37, 0x78, 0x69, 0x63, 0x75, 0x55, 0x84);`
Rust: `DEFINE_GUID!{IID_ID3D11ShaderReflection, 0x8d536ca1, 0x0cca, 0x4956, 0xa8, 0x37, 0x78, 0x69, 0x63, 0x75, 0x55, 0x84}`

Also `D3D11Reflect` is literally just this:

```C
FORCEINLINE HRESULT
D3D11Reflect(_In_reads_bytes_(SrcDataSize) LPCVOID pSrcData,
             _In_ SIZE_T SrcDataSize,
             _Out_ ID3D11ShaderReflection** ppReflector)
{
    return D3DReflect(pSrcData, SrcDataSize,
                      IID_ID3D11ShaderReflection, (void**)ppReflector);
}
```

retep998:
I have DX12 capable hardware so I'd be willing to test it to see if its just a feature level issue. What is a minimal example that I can build and run?

retep998:
Running the triangle example I get this error:
```
C:\msys64\home\Peter\gfx [master ↑ +0 ~2 -0 !]> cargo run --example triangle
     Running `target\debug\examples\triangle.exe`
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: DriverType', ../src/libcore\result.rs:746
stack backtrace:
   0:     0x7ff697595877 - rt::lang_start::h2cbcbe89183bad7eW7y
   1:     0x7ff6975950b6 - rt::lang_start::h2cbcbe89183bad7eW7y
   2:     0x7ff697589eff - sys_common::unwind::begin_unwind_inner::h45b2e3feab309e9aLct
   3:     0x7ff69758a61d - sys_common::unwind::begin_unwind_fmt::h3ab52760803c5fb5Rbt
   4:     0x7ff6975915c2 - rust_begin_unwind
   5:     0x7ff69759aed8 - panicking::panic_fmt::h8430b6988b78563f6YL
   6:     0x7ff69749515d - unwrap_failed<gfx_window_dxgi::InitError>
                        at C:\msys64\home\Peter\src\libcore\macros.rs:29
   7:     0x7ff697491adc - unwrap
                        at C:\msys64\home\Peter\src\libcore\result.rs:687
   8:     0x7ff697491386 - launch
                        at C:\msys64\home\Peter\gfx\src\lib.rs:92
   9:     0x7ff6974912f3 - main
                        at C:\msys64\home\Peter\gfx\examples\triangle\main.rs:71
  10:     0x7ff697594a78 - rt::lang_start::h2cbcbe89183bad7eW7y
  11:     0x7ff697591528 - sys_common::thread_local::StaticKey::lazy_init::h7c119cb8af25fd07S2s
  12:     0x7ff6975948fd - rt::lang_start::h2cbcbe89183bad7eW7y
  13:     0x7ff6974a11c9 - main
  14:     0x7ff69759eb77 - __scrt_common_main_seh
                        at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:264
  15:     0x7fff5ff28101 - BaseThreadInitThunk
Process didn't exit successfully: `target\debug\examples\triangle.exe` (exit code: 101)
```

retep998:
After I installed the optional windows feature "Graphical Tools" I got further!
```
     Running `target\debug\examples\triangle.exe`
Creating swap chain of size 784x561
Success with driver D3D_DRIVER_TYPE(1)
thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: DescriptorInit(VertexImport(0, None))', ../src/libcore\result.rs:746
stack backtrace:
   0:     0x7ff7fa9a53d7 - rt::lang_start::h2cbcbe89183bad7eW7y
   1:     0x7ff7fa9a4c16 - rt::lang_start::h2cbcbe89183bad7eW7y
   2:     0x7ff7fa999a5f - sys_common::unwind::begin_unwind_inner::h45b2e3feab309e9aLct
   3:     0x7ff7fa99a17d - sys_common::unwind::begin_unwind_fmt::h3ab52760803c5fb5Rbt
   4:     0x7ff7fa9a10b2 - rust_begin_unwind
   5:     0x7ff7fa9aaa38 - panicking::panic_fmt::h8430b6988b78563f6YL
   6:     0x7ff7fa8a902b - unwrap_failed<gfx::factory::PipelineStateError>
                        at C:\msys64\home\Peter\src\libcore\macros.rs:29
   7:     0x7ff7fa8a8e00 - unwrap
                        at C:\msys64\home\Peter\src\libcore\result.rs:687
   8:     0x7ff7fa8a8991 - new<gfx_device_dx11::factory::Factory>
                        at C:\msys64\home\Peter\gfx\examples\triangle\main.rs:39
   9:     0x7ff7fa8a1537 - launch
                        at C:\msys64\home\Peter\gfx\src\lib.rs:99
  10:     0x7ff7fa8a1263 - main
                        at C:\msys64\home\Peter\gfx\examples\triangle\main.rs:71
  11:     0x7ff7fa9a45d8 - rt::lang_start::h2cbcbe89183bad7eW7y
  12:     0x7ff7fa9a1018 - sys_common::thread_local::StaticKey::lazy_init::h7c119cb8af25fd07S2s
  13:     0x7ff7fa9a445d - rt::lang_start::h2cbcbe89183bad7eW7y
  14:     0x7ff7fa8b1b59 - main
  15:     0x7ff7fa9ae6e7 - __scrt_common_main_seh
                        at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl:264
  16:     0x7fff5ff28101 - BaseThreadInitThunk
Process didn't exit successfully: `target\debug\examples\triangle.exe` (exit code: 101)
```

sectopod:
Awesome, so we should be good by using `ID3D10ShaderReflection`, for now at least. Too bad it's not exposed by winapi-rs :(

sectopod:
@retep998 Now I'm puzzled.
Added D3D10Shader module myself (https://github.com/retep998/winapi-rs/pull/260), and here is the code now:
```rust
    let hr = unsafe {
        d3dcompiler::D3DReflect(code.as_ptr() as *const winapi::VOID,
            code.len() as winapi::SIZE_T, &dxguid::IID_ID3D10ShaderReflection, &mut reflection)
    };
```
Still getting `E_NOINTERFACE`. Something is missing here.

sectopod:
Maybe it's something wrong with my binaries? After all, this error may be interpret as "your binary doesn't have this interface". I'm compiling the shader with this batch:
```cmd
set FXC="%DXSDK_DIR%\Utilities\bin\x64\fxc.exe"
mkdir data
%FXC% /T vs_4_0 /E Vertex /Fo data/vertex.fx triangle.hlsl
%FXC% /T ps_4_0 /E Pixel /Fo data/pixel.fx triangle.hlsl
```

retep998:
@sectopod Maybe you don't have something installed that is necessary to allow debugging and reflection? I'm on Windows 10 and I had to install the "Graphical Tools" optional system feature which I found out from http://blogs.msdn.com/b/chuckw/archive/2012/11/30/direct3d-sdk-debug-layer-tricks.aspx although not having it caused me to fail at creating the swap chain, not creating the reflection interface. It could possibly also be due to you using the `-gnu` toolchain which I really hope it isn't but I wouldn't put that beyond the ability of the MinGW linker to screw things up like that. I have a bunch of sdk layer DLLs in my `system` folder, do you have them too? http://i.imgur.com/zjmlAHY.png

sectopod:
It's certainly not the "Graphics Tools", since these are installed and only needed for debug runtime.

DLL-wise, I don't have that many:
system32

As for MSVC toolchain, I'm trying this now. Getting something weird upon running the example:
> ../cargo.exe: error while loading shared libraries: ?: cannot open shared object file: No such file or directory

On a side note, my cargo doesn't like `[target.'cfg(windows)'.dependencies]` at all, even though it's clearly documented in the [manifest format](http://doc.crates.io/manifest.html). I need to somehow depend on dx11/dxgi stuff on windows and not depend on glfw in this case. Specifying all the target triples explicitly seems to be rather verbose.

retep998:
That `[target.'cfg(windows)'.dependencies]` thing is very new syntax which was only just added less than a week ago. You'll need a _very_ recent nightly to be able to use that syntax.

As for why the MSVC toolchain isn't working, that is a rather strange error. Did you make sure rustc/cargo are properly in your PATH? If you're having difficulty juggling multiple toolchain installations, consider using [multirust-rs](https://github.com/Diggsey/multirust-rs).

sectopod:
@retep998 I'm building just fine, thus the PATH is correct. The error happens upon running the example. I tried just started the EXE directly and got a better error:
> error while loading shared libraries: D3DCOMPILER_47.dll: cannot open shared object file: No such file or directory

sectopod:
We shouldn't depend on either D3DCOMPILER_47 or D3DCOMPILER_46, since they require Windows 8+. If this is the only way, we'd have to redistribute them somehow.

retep998:
If you use the `d3dcompiler.lib` from the 8.1 or 10 SDK then you end up depending on `d3dcompiler_47.lib` which, if you're using an older Windows, can be found in the Windows SDK (for example `C:\Program Files (x86)\Windows Kits\8.1\bin\x64\d3dcompiler_47.dll`). However, it is one of the files that Microsoft allows for redistribution so you can bundle it with your binary without issue (unless you're using GPL code which makes that situation legally dubious). If you configure your environment with the appropriate vcvars file first (`C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\vcvars64.bat`) then that should add that earlier bin folder to your `PATH` allowing things to work.

kvark:
I wonder why this `D3DCOMPILER_47.dll` only started to happen when @sectopod switched to `msvc` tool chain?

retep998:
Probably because MinGW uses heavily outdated libraries. Their `d3dcompiler.lib` is for `D3DCOMPILER_43.dll` apparently.

kvark:
@sectopod we should use [winit](https://github.com/tomaka/winit) for DXGI window. Example [here](https://github.com/tomaka/vulkano/blob/master/vulkano/examples/image.rs#L28-L29).

retep998:
Actually since it is `d3dcompiler` which provides `D3DReflect` I think what is happening here is that the `IID_ID3D11ShaderReflection` interface provided by `winapi` is for the one in `d3dcompiler_47.dll`, meanwhile when using the `-gnu` targets you end up linking to `d3dcompiler_43.dll` which probably provides a different interface and thus resulting in that error. @sectopod If you do manage to get the `-msvc` stuff compiling and running, I have a suspicion that the `D3DReflect` call will succeed. I'm not sure what the best way to go about solving this is. I suppose I could grab `libd3dcompiler_47.a` from MinGW and use that as the import library, which would at least result in consistency between the `-msvc` and `-gnu` targets although it would require the existence of `d3dcompiler_47.dll` which needs either Windows 8.1/10 or the Windows 8.1/10 SDK. I definitely don't want to redistribute that DLL with the source as it is 4.26 MiB.

sectopod:
Thanks @retep998 
`-msvc` stuff does compile, but running the example I ported to DX still fails, even though I put the damned `D3DCOMPILER_47.DLL` in the same folder as the executable. Output from `gdb`:
> Starting program: C:\Code\3d\gfx\target\debug\examples\triangle.exe
New Thread 2604.0x6b4
During startup program exited with code 0xc000007b.

The dependency on `msvcrt.dll` seems to be screwed up:

triangledeps

retep998:
Are you sure that is the 64-bit `d3dcompiler_47.dll`? You might have accidentally grabbed the 32-bit version.

sectopod:
@retep998 it's hard to believe, but x64 version of this library worked! Thank you!

sectopod:
triangle

kvark:
FYI, @fkaa is waiting for this PR to land

sectopod:
@fkaa I'm getting close to a basic feature set suitable for our examples. Unless you are blocked by this PR, I'll continue converting them (examples).

fkaa:
@sectopod take your time! I'm mainly waiting for the changes in the backend and the `App` things before I start creating a new backend.

sectopod:
Got constant buffer updates working, and the terrain example with them:
terrain

sectopod:
Hooray, we are green!

sectopod:
Sadly, mipmap example fails to run. `R3G3B2` is not supported by DX10+. cc @fkaa - I'll create a bug

kvark:
So technically we can already merge safely, but not publish yet. In case @fkaa wants to start on the Metal backend and is just waiting for this PR.

sectopod:
I got good news and bad news.
Good: I was able to expand Application trait structure to support multi-threaded applications, like our Shadow example. And of course I ported the example itself for the test.
Bad: Under DX11 the shadow example fails to create the blend state of the second PSO (returning `E_INVALIDARG`). This is rather strange for multiple reasons:
  - the first PSO had the same exact descriptor, and it created just fine. DX documentation explicitly mention that if the runtime sees that it already created a state with this descriptor, it will just re-use it.
  - there is absolutely nothing about the descriptor. it's as empty as it could be - Shadow example doesn't use any blending. 
  - following operations, like creating a buffer, fail as well with `E_INVALIDARG`. Again, with nothing suspicious.

Perhaps, there is some sort of a memory stomp. Or, there is a DX runtime error occurring that we need to see somehow.

kvark:
You can use [DebugView](https://technet.microsoft.com/en-us/sysinternals/debugview.aspx) to show DX11 debug runtime output. So we technically don't need any explicit support to enable it. We can use DX11 control panel to switch it on, and then the tool to capture it. Hope this will be helpful!

fkaa:
Reminder that we also need to wait for @retep998 to upload the winapi changes before merging

kvark:
@fkaa right, thanks! We also need to setup a windows auto-builder. Anyone got experience with that?

sectopod:
@fkaa I had to update the shadow example to use 2 constant buffers. It looks like, in general, we'll have to have 2 for each example as long as we want GL compatibility, because on GL side we can't share the same uniform buffer better VS and PS...

So thing is - the second one doesn't get updated or set up properly. I recall you fixed an issue with that, and I'm pretty sure I merged with latest 'master', so I'll need to dig it up and figure if we need to fix this better.

sectopod:
I got `DebugView` tool working and was able to fix a couple of nasty issues thanks to DX11 debug runtime output, so that the Shadow example was running if used with a trivial forward shader. However, upon porting the exact code we use on GL, I discovered that my Quadro FX 4600 only supports Shader Model 4.0, while our usage of comparison sampling with array textures requires SM 4.1 on DX11... So I'm not able to verify if it actually works correct now. Strangely, the same hardware is completely fine with doing the very same thing on GL.

Perhaps, someone else can give it a shot? (@retep998)
All you need is to switch `ApplicationGL2` to `ApplicationD3D11` in the `main()`.

sectopod:
@fkaa Alarm down. I found my own mistake there, your uniform block handling code works good. In fact, Shadow example now has 3 of these buffers, and all are bound correctly.

sectopod:
Got some instancing working:
instancing
Going to proceed to the last big fat example - the deferred rendering one.

sectopod:
Got another problem with the deferred example. We can't expose texture object and sampler under the same name from HLSL. So it becomes incompatible with our GL backend, unless we change the latter to auto-assign a different name for the associated sampler.

sectopod:
Heads up - I'm cutting `R3_G3_B2` and `R8_G8_B8` formats out, and switching the examples to use `Srgba8`. I wanted to work around `R8_G8_B8` but in the end it's nothing but trouble - these are simply not supported by DX11.

sectopod:
Getting blocked by #899 . Will have to address this issue in the scope of the PR.

sectopod:
deferred

retep998:
That looks glorious

sectopod:
Thanks @retep998! I got stumbled a bit by the deferred example, partially due to #899, which will still slow me down for the need to rewrite some parts. But otherwise - we are full steam moving forward.

sectopod:
There are at least 2 major problems with the application launcher that I got:
  1. It doesn't support event handling. This will be solved once tomaka publishes `winit`. I can technically start adopting it right away.
  2. It can't choose the backend at run-time. This could be solved with HKT, but they are not nearly coming to Rust yet, so we need another way. In order for the multiple backends to be created our startup function would have to receive a separate type for each backend:
```rust
gfx_app::launch::<App<_>, App<_>, App<_>>();
```
This doesn't look too hot. Any suggestions are welcome!

sectopod:
Polishing aside, I got only _Flowmap_, _Skybox_, and _UboTilemap_ examples left.

fkaa:
@sectopod great! IIRC `winit` will get published once the linux side has gotten rid of the GL-specific things.

kvark:
I'm not too worried about the app launcher - we can patch it later without breaking core, render, or even device modules. More concerned about #898 at the moment... but I suppose it requires a bit of design/architecture work, not just implementation, so we shouldn't try to shove it in this PR. In other words, :shipit: already!

kvark:
Perhaps, we can ask @cmr to start reviewing everything but the examples? Plus anyone else, who is interested in DX11, please feel free to review!

src/core/src/factory.rs
```diff
@@ -68,32 +84,48 @@ pub enum BufferRole {
...
+    /// GPU: read, CPU: none. Optimal for resourced textures/buffers.
     Const,
-    /// This buffer will be updated "frequently", and will be read from multiple times between
-    /// updates.
+    /// GPU: read, CPU: write.
     Dynamic,
-    /// This buffer always or almost always be updated after each read.
-    Stream,
+    /// GPU: copy, CPU: as specified. Used as a staging buffer,
+    // to be copied back and forth with on-GPU targets.
```
fkaa: Missing a `/`

sectopod:
LOL GLSL:
```glsl
    float lerp = 2.0 * abs(f_Offset0 - .5f);
    vec3 result = mix(t0, t1, lerp);
```
HLSL:
```hlsl
	float mix = 2.0 * abs(u_Offsets.x - 0.5);
	return lerp(t0, t1, mix);
```

sectopod:
Getting partially blocked by #897

cmr:
I can start a review tomorrow. Awesome work as usual, @sectopod!

cmr:
Did a simple pass through, and overall looks good! I see there's a few TODO and `unimplemented!()` scattered around.

sectopod:
@cmr thanks for looking into it, much appreciated!
The plan about TODO and `unimplemented()` stuff is to have as little as we require to properly run all the example. Nothing is perfect, especially from the start, so we'll keep patching it up, implementing missing bits, etc. I hope to get the `gfx_core` stable though.

As for the current status:
Doing #897 now, and it got me down another little rabbit hole. The `CommandBuffer` interface needs to be changed, and this is a core part. So this slows me down a ton, but in  the end we'll have a much cleaner system, with a chance of the stable core. This also means the core will need another look once I'm done with this PR.

sectopod:
Finished porting the examples. Now implementing #897 and #902, plus need to fill up `Map`/`Unmap` since it's now mandatory.

kvark:
@retep998 How is winapi-rs doing? Looks like we are not far from getting blocked by it.

sectopod:
Shadow example running on native DX11 deferred contexts (#902):

shadow

sectopod:
Ok, I think this PR is ready! Blocked by @retep998 now.

I couldn't figure out how to properly generate mipmaps for textures with the input data. Regardless of the `D3D11_RESOURCE_MISC_GENERATE_MIPS` flag, the runtime expects `D3D11_SUBRESOURCE_DATA` provided for every single mip. Obviously, if we had the contents we wouldn't ask it to generate mips in the first place...

Searching the mighty internets got me [this], where basically one is asked to do `UpdateSubresource` and `GenerateMipmaps` directly. This doesn't work for us, since these methods expect a device context, while we only have the device itself on the `Factory` level. In other words, they are rendering operations, and we have rendering and resourcing separated.

Anyhow, if someone knows the answer, please enlighten me. We'll eventually fix the DX11 backend to support mipmaps properly, it's just going to be a patch version bump.

sectopod:
I think I get it now, was misreading the documentation. Specifying this flag doesn't generate mipmaps automatically. It just enables the capability of calling `GenerateMipmap` on an SRV later on. Thus, I'm going to enable the flag by default for any surface that may support it. Also, I'm removing the mipmap generation flag completely, since DX11 appears to be incapable of doing that at the resource creation level.

retep998:
`d3d11-sys = "0.2.0"`
`d3dcompiler-sys` = "0.2.0"`
`dxguid-sys` = "0.2.0"`

sectopod:
@retep998 thank you! Will patch the appveyor now and see if it passes through.

sectopod:
@retep998 don't you have to change the dependencies to "0.2" for these crates and publish the new `winapi-0.2.7`? It doesn't look like your new sub-crates are used: https://ci.appveyor.com/project/kvark/gfx/build/1.0.19

retep998:
`winapi` doesn't depend on the `-sys` crates except as a dev dependency for testing purposes. It is entirely up to gfx to specify the correct version of winapi crates to build against. Your `src/backend/dx11/Cargo.toml` still depends on the older versions of those crates, so you will need to update that file.

sectopod:
@retep998 I see, thank you! Just verified it builds locally without cargo overrides.
For some reason, my build was failing with image-0.7, so I set 0.6 there, and it works (cc @bvssvni).
Will see what AppVeyor does now.

src/backend/dx11/Cargo.toml
```diff
@@ -30,6 +30,6 @@ name = "gfx_device_dx11"
 log = "0.3"
 gfx_core = { path = "../../core", version = "0.2" }
 d3d11-sys = "0.1"
```
retep998: I updated `d3d11-sys` too, you know.

sectopod:
Ok, we are green on windows msvc stable and linux stable/beta. Nightly is broken by #904 rustc regression. GNU build is failing on miniz-sys, which is just a dependency issue, and needs to be addressed on the appveyor script (or platform config) level rather than the code side.

I believe we are ready to :shipit:, at last! @cmr can take the responsibility of :hammer: hitting the big green button.

src/backend/dx11/Cargo.toml
sectopod: Sorry, I missed it. We'll update this at any point later, since it's not necessary and doesn't break the API.

cmr:
Fails on nightly and the -gnu target (wrong setup on appveyor?), merging anyway!

cmr:
:confetti_ball: :balloon: awesome job @sectopod

kvark:
:tada: :tada: :tada:

sectopod:
Thank you everyone for the support!

retep998:
Now gfx can finally live up to its goal of being a a library that abstracts across multiple graphics APIs!

Issue #844, opened 2016-01-05
[ARM breakage] Same old same old 


petevine:
More fallout from https://github.com/rust-lang/rust/issues/29867

```` 
Compiling gfx_device_gl v0.7.0
  /home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/gfx_device_gl-0.7.0/src/shade.rs:314:21: 314:69 error: mismatched types:
 expected `*const u8`,
    found `*const i8`
(expected u8,
    found i8) [E0308]
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/gfx_device_gl-0.7.0/src/shade.rs:314                     target.as_bytes_with_nul().as_ptr() as *const i8);
                                                                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/gfx_device_gl-0.7.0/src/shade.rs:314:21: 314:69 help: run `rustc --explain E0308` to see a detailed explanation
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/gfx_device_gl-0.7.0/src/shade.rs:328:46: 328:94 error: mismatched types:
 expected `*const u8`,
    found `*const i8`
(expected u8,
    found i8) [E0308]
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/gfx_device_gl-0.7.0/src/shade.rs:328                 gl.GetFragDataLocation(name, target.as_bytes_with_nul().as_ptr() as *const i8)
                                                                                                                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/gfx_device_gl-0.7.0/src/shade.rs:328:46: 328:94 help: run `rustc --explain E0308` to see a detailed explanation
/home/odroid/.cargo/registry/src/github.com-48ad6e4054423464/gfx_device_gl-0.7.0/src/info.rs:145:55: 145:60 error: mismatched types:
 expected `*const u8`,
    found `*const i8`
(expected u8,
    found i8) [E0308]
[...]

PR #863, opened 2016-02-08
Additional ARM fix

petevine:
Just one was still remaining .
This should close https://github.com/gfx-rs/gfx/issues/844


kvark:
Thanks!

PR #867, opened 2016-02-14
Add implementation for creating cubemaps with ´TexStorage´. Fixes #866

### LINKED ISSUE #866: Cube texture storage for GL backend
Currently, `Factory.create_texture` and friends fail when passed a `Cube` texture kind. This boils down to `make_with_storage_impl` returning `Err` when a `Cube` kind is used.

Line with error return:
https://github.com/gfx-rs/gfx/blob/64d8fce8b3d32572a9ba16adbf43f126974ba7d2/src/backend/gl/src/tex.rs#L505

Suggested fix would be to mirror the behaviour of `make_widout_storage_impl` as with this line:
https://github.com/gfx-rs/gfx/blob/64d8fce8b3d32572a9ba16adbf43f126974ba7d2/src/backend/gl/src/tex.rs#L372

fkaa:


kvark:
I don't think this is correct. The backend should allocate all 6 sides.

fkaa:
That's what I thought too at first (which the original PR contained), but according to the [reference](https://www.opengl.org/sdk/docs/man4/html/glTexStorage2D.xhtml), only one call to `glTexStorage2D` is required to initialize storage for all sides.

fkaa:
Also tested this with the skybox example I'm working on.

kvark:
@fkaa looks like you are correct!

PR #879, opened 2016-02-20
Window resize support

### LINKED ISSUE #877: Window resize support
The current `gfx_window_glutin` interface doesn't provide an ability to update the color/depth view sizes when the window gets resized. Clients call `gfx_device_gl::create_main_targets` now, which was not supposed to be exposed directly.

One solution would be to have `gfx_window_glutin::update_size(&glutin::Window, &mut handle::SRV, &mut handle::DSV)`. Another - to have a backdoor to the handle's descriptor via a service trait (ala `Phantom`), so that the underlying textures can be patched without re-creating them with the views.

Another little thing - we need to provide the `init_raw` interface for untyped SRV/DSV. It will be handy for the applications that switch formats depending on some settings at run-time. The typed `init` could then be a tiny wrap around the untyped call.

sectopod:
Fixes #877

### DIFF
--- a/src/window/glutin/src/lib.rs
+++ b/src/window/glutin/src/lib.rs
@@ -18,15 +18,14 @@ extern crate gfx_core;
 extern crate gfx_device_gl;
 extern crate glutin;
 
-use gfx_core::{format, handle};
-use gfx_core::tex::Size;
+use gfx_core::{format, handle, tex};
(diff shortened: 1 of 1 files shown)


kvark:
Ok, this would make us ready to publish new versions.
One thing though, maybe we can leave the old format interface, so that `gfx_core` stays at 0.1

PR #883, opened 2016-02-22
Maybe fix macro error

retep998:
It worked for triangle example for me, maybe it fixes everything?


kvark:
Thanks!

kvark:
Unfortunately, this is not a solution, judging by the errors.

PR #884, opened 2016-02-23
Assign block bindings automatically when no explicit layout bindings are given

fkaa:
Will assign the blocks to their indices if there are no bindings found that are *not* zero (default). Tested with `ubo_tilemap` and local `tiled_forward` example.


kvark:
Wonderful PR, and Travis is green again!
Thank you @fkaa !
@home r+

kvark:
@homu r+

PR #888, opened 2016-02-25
Catching up v0.9 with master

kvark:
Everything is backwards compatible:
  - constant buffer binding
  - Srgba8
  - core format view fixes


kvark:
@homu r+

PR #893, opened 2016-03-03
Use R8_G8_B8_A8 format for textures.

### LINKED ISSUE #892: Fix format of mipmapping example
`R3G3B2` is only supported by DX9 and GL due to legacy reasons, but not by DX10 and up.
We should use something else. cc @fkaa

vickenty:
R3_G3_B2 is obsolete and is causing problems for DX11 backend.

Fixes #892.

### DIFF
--- a/examples/mipmap/main.rs
+++ b/examples/mipmap/main.rs
@@ -17,7 +17,7 @@ extern crate gfx;
 extern crate gfx_window_glutin;
 extern crate glutin;
 
-pub use gfx::format::{Srgb8, Depth};
+pub use gfx::format::{Rgba8, Depth};
 
 gfx_vertex_struct!( Vertex {
     pos: [f32; 2] = "a_Pos",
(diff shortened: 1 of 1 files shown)


kvark:
Thank you @vickenty !

May I ask you to redo it slightly though? Problem is, `R8_G8_B8` format is not good either. If you look at the [DXGI_FORMAT](https://msdn.microsoft.com/en-us/library/windows/desktop/bb173059%28v=vs.85%29.aspx) list, there are no 3-component 8 bits (each) types. Presumably, because it's going to contain 32bits (aligned) anyway, so you might as well include the alpha...

Perhaps, you could change the code to use `B5G6R5` instead? That would be wonderful.

If you feel courageous, you can also try removing `R8_G8_B8` and `R3_G3_B2` surface types completely, although that can (and probably should) be done as a separate change, since it may touch different other things (like our `Srgb8` shortcut).

kvark:
Correction: actually, it looks like our DX11 backend [just uses](https://github.com/sectopod/gfx-rs/blob/719ade53d447d3f1df5612af7d207127e80237be/src/backend/dx11/src/data.rs#L54) `RGBA8` instead of requested `RGB8` for compatibility. This is not too sound, and I'd rather have it completely removed.

vickenty:
I changed the format to Rgba8, hopefully it will work everywhere.

It looks like Appveyor failure is not related to this PR, but is there anything else I can improve here?

kvark:
Thanks! Don't worry about AppVeyor - we just set it up, and it's not configured properly yet.

kvark:
Could you squash the commits?
Also, any reason you didn't use B5_G6_R5? It seems to be well supported on all platforms too, including GL ES 2.0.  I feel that it would suit better here to follow the original spirit of the example code that tried to pack data as tightly as possible.

sectopod:
Actually, `B5_G6_R5` has numerous issues. First, it's a reversed format (BGR instead of RGB). Second, while it's defined it's not really supported prior to DX12:
> Direct3D 10 through Direct3D 11:  This value is defined for DXGI. However, Direct3D 10, 10.1, or 11 devices do not support this format.
> Direct3D 11.1:  This value is not supported until Windows 8.

@homu r+

Issue #853, opened 2016-01-28
Weak PSO and components


kvark:
With the PSO revolution in #828 we got strongly-typed PSO and components. Problem is - they are fixed at compile time, well, mostly - unless we do generic PSO definitions, which are blocked on rust macros expressiveness. So when you want to load something from disk, something you don't know exactly, the strong PSO doesn't fit. Hopefully, both PSO and components are interfaced with by traits, so all it takes for us is to introduce weakly typed stuff:
- `WeakVertexBuffer` - to receive the format slice in the init data instead of deriving it from the type
- `WeakPipelineState` - to receive all the components in some serialized form. In fact, the component becomes it's implementation detail, since it's not exposed anywhere.

sectopod:
Most of this task is implemented in #910. The big thing left is loading the whole PSO definition at run-time, although I don't see the use case for it clearly yet.

kvark:
Perhaps, @ebkalderon or other Amethyst devs want to either contribute or express their opinion on having that in gfx-rs.

kvark:
We got a bunch of weak components, closing.

PR #910, opened 2016-03-24
Raw PSO components

sectopod:
This implements a solid part of #853 
It will allow some stuff to be loaded at run-time without any prior compile-time knowledge, like:
 - your vertex layout
 - format of the textures, like you don't know if your texture is RGBA or RGB or something else
 - output target format, like you may want to configure the precision and sRGB properties
 - blending


kvark:
Nice, I recall @bvssvni was asking for it.
@homu r+

Issue #914, opened 2016-03-27
Document the code


eugene2k:
I have finally gotten my rust-sdl app working with gfx! After spending two days working on it. If that doesn't sound bad to you, I'll add that all my app does is draw a triangle. In other words I haven't even done anything other than look into the source code and try to figure out which generic does what. There's a serious lack of comments in the code as well as lack of online documentation (which would be ok if the code was well documented) and it hinders gfx's adoption, in my opinion. 

P.S. Before you refer me to examples/performance please note that that particular example is way out of date.

kvark:
@eugene2k That does sound bad, I agree. Glad you managed to port it at last.

Is `performance` got accidentally disabled on linux build as well? we need to look into that for sure!

Perhaps, you could isolate all the SDL handling stuff inside a separate crate contributed as `gfx_window_sdl`?

kvark:
@eugene2k That does sound bad, I agree. Glad you managed to port it at last.

Is `performance` got accidentally disabled on linux build as well? we need to look into that for sure!

Perhaps, you could isolate all the SDL handling stuff inside a separate crate contributed as `gfx_window_sdl`?

eugene2k:
Performance hasn't been ported, it seems. It appears to be using the pre-PSO interface. I'll give the sdl port a try. Shouldn't be too hard.

eugene2k:
This should do it...
[sdl.patch.gz](https://github.com/gfx-rs/gfx/files/192389/sdl.patch.gz)

kvark:
@eugene2k could you file a PR for it?

eugene2k:
I don't have the repository cloned on github, unfortunately.

Bastacyclop:
redundant with #942

PR #918, opened 2016-03-29
SDL2 window by @eugene2k

sectopod:
Applied the patch from #914 with minor formatting fixes.


eugene2k:
I forgot I left a lifetime param in the init function declaration. It's not needed for anything, so feel free to get rid of it.

kvark:
@eugene2k I think the dependency is correct - the idea behind this root crate is to depend on everything, so that you could test everything with a single "cargo test" command.

Nightly fails on something irrelevant:
>error: the current crate is indistinguishable from one of its dependencies: it has the same crate-name `libc` and was compiled with the same `-C metadata` arguments. This will result in symbol conflicts between the two. [E0519]
> error: Could not document `libc`.

Merging.

PR #919, opened 2016-03-29
gfx_app refactored a little to improve readability

eugene2k:
A very small change to gfx_app that makes parts of the code a bit more readable. Didn't want to just throw it away.


kvark:
Thanks!

PR #921, opened 2016-03-31
Use ColorFormat instead of Srgba8

csherratt:
Srgba8 does not seem to work with my drivers causing none of the examples to work. Switching to Rgba8 seems to fix this issue.


kvark:
Are you hit by https://github.com/tomaka/glutin/issues/748 ?
Anyway, PR is good, thank you!
@homu r+

PR #926, opened 2016-04-03
Move Bundle out of pipeline macro.

### LINKED ISSUE #925: Remove Bundle from pipeline macro
The way it is now, `Bundle` is only useful in a very small number of cases when user has a PSO per visible object, effectively narrowing its use cases to examples and very simple programs.

`Bundle` can be improved, to take ref instead of ownership. But someone may want to draw his objects in several passes, in which case it should be `Vec<Arc<PipelineState>>`. But if I have one global PSO for everything, both `Vec` and `Arc` are not needed. What I want to say here is that there is no one universal way to combine meshes with pipeline state, each application does its own thing.

Which is why I propose Bundle is removed from gfx and either moved to examples support library (gfx_app or other), or just integrated into the examples themselves (it seems possible to replace `Bundle` with a tuple and keep number of lines the same). Or it could be a start of a higher-level drawing library, as was proposed in #876.

cc @sectopod, @kvark

vickenty:
Makes it opt-in without allow(dead_code), also opens way up for multiple
implementations.

Fixes #925.

### DIFF
--- a/examples/blend/main.rs
+++ b/examples/blend/main.rs
@@ -19,6 +19,7 @@ extern crate image;
 
 pub use gfx_app::ColorFormat;
 pub use gfx::format::{Rgba8, DepthStencil};
+use gfx::Bundle;
 
 gfx_vertex_struct!( Vertex {
     pos: [f32; 2] = "a_Pos",
@@ -72,7 +73,7 @@ const BLENDS: [&'static str; 9] = [
 ];
 
 struct App<R: gfx::Resources>{
-    bundle: pipe::Bundle<R>,
+    bundle: Bundle<R, pipe::Data<R>>,
     id: u8,
 }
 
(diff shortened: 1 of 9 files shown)


kvark:
@vickenty Thank you, looks good!
We'll need to not forget to add the proper header to the new source file, like the one all other sources have.
Also note, this is technically a breaking change!
@homu r+

kamil-l:
Tried to compile cube example, got:
`error: unresolved import `gfx::Bundle`. There is no `Bundle` in `gfx`.

kvark:
@kamil-l well, it's clearly included: https://github.com/gfx-rs/gfx/pull/926/files#diff-be2c920c08146924af4d1983d6894d63R59

I just tried locally, and the cube example builds and runs on latest. Must be something local to your changes then?

PR #929, opened 2016-04-04
`get_level_dimensions` keeps unused dims at `0`

### LINKED ISSUE #928: gfx_core::tex::Descriptor::to_image_info yields depth=1 even if depth is 0
I don't understand enough here, but I have tracked down the issue that it's impossible to update a `piston_window::Texture` with the image it was created with.

[This piece of code](https://github.com/gfx-rs/gfx/blob/master/src/core/src/tex.rs#L178) uses `max(1, depth >> level)` to obtain the new depth, but if the starting depth was 0, then it should also yield 0, right?

oli-obk:
fixes #928

### DIFF
--- a/src/core/src/tex.rs
+++ b/src/core/src/tex.rs
@@ -173,9 +173,11 @@ impl Kind {
     }
     /// Get the dimensionality of a particular mipmap level.
     pub fn get_level_dimensions(&self, level: Level) -> Dimensions {
-        use std::cmp::max;
+        use std::cmp::{max, min};
+        // unused dimensions must stay 0, all others must be at least 1
+        let map = |val| max(min(val, 1), val >> level);
         let (w, h, d, _) = self.get_dimensions();
-        (max(1, w >> level), max(1, h >> level), max(1, d >> level), AaMode::Single)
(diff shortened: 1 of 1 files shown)


oli-obk:
technically it should never return `0` but maybe return an `Option` instead that people can `unwrap_or(0)` if they want?

kvark:
Thanks!
@homu r+

kvark:
Hmm, @homu is not responsive. Merging manually.

bvssvni:
Will this get published on crates.io as 0.10.2?

kvark:
Agreed, this change needs to be back-ported to 0.10 (maybe with some other stuff) and published.

kvark:
Looks like this PR broke texture updates since the `ImageInfoCommon` dimensions are now computed to be 0 due to zeroes in the width/height/depth etc that were not there before...
Another way to approach this would be to revert the PR and fix `get_dimensions` instead to use `1` where not applicable instead of zeroes.

sectopod:
@oli-obk note that this fix is now published as a part of gfx_core-0.2.1. You can stop using cargo override if you were relying on this to work.

Issue #922, opened 2016-03-31
Are Users Supposed to Use gfx_app::Application?


kruffin:
I'm trying to figure out if I'm supposed to use gfx_app::Application since that's what all the examples use, but don't see a way to take user input. Looking at the implementation it hard codes the keys it will accept. Is there a way to handle user input with this struct or is this boilerplate I'm supposed to copy and modify?

``` Rust
let mut harness = Harness::new();
main: loop {
    // quit when Esc is pressed.
    for event in window.poll_events() {
        match event {
            glutin::Event::KeyboardInput(_, _, Some(glutin::VirtualKeyCode::Escape)) |
            glutin::Event::Closed => break 'main,
            _ => {},
        }
    }
    // draw a frame
    app.render(&mut device);
    window.swap_buffers().unwrap();
    device.cleanup();
    harness.bump()
}
```

csherratt:
`gfx_app` is a test runner that was added to help with the examples. You can use it as a guide, but I would not recommend building off of it.

kruffin:
Is there something you would recommend? I've tried piston_window but it's difficult trying to find out where the Init{} parameters in Applicaiton line up with with the window returned from it.

kvark:
User input is not yet handled, blocked by #907

bjz:
I wonder if there is a way to make this clearer to people looking at the examples. It might be confusing having it as a separate crate makes it look like it is part of the user facing library. Glium has a support module for example: https://github.com/tomaka/glium/tree/master/examples/support, and examples import it using `mod support;` - making it clear that it is part of client code.

kruffin:
From my perspective, it would be nicer if the examples showed how to set up the boilerplate needed to use the library. I really liked the simplicity of the examples, but feel like it only showed me how to use gfx with gfx_app. What I thought were core concepts of the library (say gfx_app::shade::Source) turned out to actually be stuff I need to re-implement myself. I went from understanding how to use the library, to looking at glutin and seeing what the alternatives are there; raised the bar to entry.

I do like this library, so keep up the good work. It was easier (and more powerful) to get started than say kiss3d, but feel a little lost since there is so much if windows OS do this in the gfx_app crate.

eugene2k:
Well, before any of the examples are switched to showing how to do it right, here's the basic thing you have to do:
1. use gfx_vertex! macro to describe your vertex buffer in the format of a struct
[...]

PR #930, opened 2016-04-04
Remove gfx_app dependency from triangle example

aepsil0n:
This dependency is convenient for re-using code between example. However, it makes it also harder for a user to follow and figure out, what they actually have to implement themselves when using gfx.

Raised in #922.

### DIFF
--- a/examples/triangle/main.rs
+++ b/examples/triangle/main.rs
@@ -49,8 +49,7 @@ pub fn main() {
         .with_vsync();
     let (window, mut device, mut factory, main_color, _main_depth) =
         gfx_window_glutin::init::<ColorFormat, DepthFormat>(builder);
-    let command_buffer = factory.create_command_buffer();
-    let mut encoder: gfx::Encoder<_, _> = command_buffer.into();
+    let mut encoder: gfx::Encoder<_, _> = factory.create_command_buffer().into();
     let pso = factory.create_pipeline_simple(
         include_bytes!("shader/triangle_150.glslv"),
         include_bytes!("shader/triangle_150.glslf"),
--- a/examples/triangle/main.rs
+++ b/examples/triangle/main.rs
@@ -14,11 +14,17 @@
 
 #[macro_use]
 extern crate gfx;
-extern crate gfx_app;
+extern crate gfx_window_glutin;
+extern crate gfx_device_gl;
+extern crate glutin;
 
-pub use gfx_app::ColorFormat;
+use gfx::traits::FactoryExt;
+use gfx::Device;
 
-gfx_vertex_struct!( Vertex {
+pub type ColorFormat = gfx::format::Rgba8;
+pub type DepthFormat = gfx::format::DepthStencil;
+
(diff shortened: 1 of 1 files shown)


examples/triangle/main.rs
```diff
@@ -28,58 +34,51 @@ gfx_pipeline!(pipe {
...
-        let (vbuf, slice) = factory.create_vertex_buffer(&vertex_data);
+pub fn main() {
+    let builder = glutin::WindowBuilder::new()
+        .with_title("Triangle example".to_string())
+        .with_dimensions(1024, 768)
+        .with_vsync();
+    let (window, mut device, mut factory, main_color, _main_depth) =
+        gfx_window_glutin::init::<ColorFormat, DepthFormat>(builder);
+    let command_buffer = factory.create_command_buffer();
+    let mut encoder: gfx::Encoder<_, _> = command_buffer.into();
```
kvark: could be merged with the previous line

examples/triangle/main.rs
```diff
@@ -28,58 +34,51 @@ gfx_pipeline!(pipe {
...
+                glutin::Event::Closed => break 'main,
+                _ => {},
+            }
         }
+        // draw a frame
+        encoder.clear(&data.out, clear_color());
+        encoder.draw(&slice, &pso, &data);
+        window.swap_buffers().unwrap();
+        device.cleanup();
+        encoder.flush(&mut device);
```
kvark: gotta move this between the `draw()` and `swap_buffers()`

examples/triangle/main.rs
```diff
@@ -28,58 +34,51 @@ gfx_pipeline!(pipe {
...
-struct App<R: gfx::Resources> {
-    pso: gfx::PipelineState<R, pipe::Meta>,
-    data: pipe::Data<R>,
-    slice: gfx::Slice<R>,
-}
-
-impl<R: gfx::Resources> gfx_app::Application<R> for App<R> {
-    fn new<F: gfx::Factory<R>>(mut factory: F, init: gfx_app::Init<R>) -> Self {
-        use gfx::traits::FactoryExt;
+fn triangle() -> [Vertex; 3] { [
```
kvark: I'd prefer constants instead of global functions here. First, because constants are more restrictive. Second, because function names should start with a verb, and `triangle()` doesn't satisfy this rule, and `clear_color()` is misleading for it's not really clearing anything.

kvark:
@aepsil0n thank you! I'd much appreciate if you could address the 3 points I mentioned before merging.

examples/triangle/main.rs
aepsil0n: I am a little bit unclear on what you mean by "this"… the loop?

aepsil0n:
Points have been addressed. I figure the failing AppVeyor build has nothing to do with my changes?

kvark:
Marvellous, thank you!
AppVeyor is partially broken still, unrelated to your changes. I created #932 to track it.
@homu r+

PR #931, opened 2016-04-05
Multiplied hidpi_factor on window glutin to get correct resolution on retina displays

paezao:
Hello,

I've had troubles with GFX and the resolution on my retina display. I found out we needed to multiply the results of a function called hidpi_factor from Glutin Window.

Cheers


kvark:
Interesting, thank you for the fix!
@homu r+

PR #937, opened 2016-04-07
Fixed get_texel_count (fixes #928)

### LINKED ISSUE #928: gfx_core::tex::Descriptor::to_image_info yields depth=1 even if depth is 0

sectopod:


sectopod:
@homu r+

PR #939, opened 2016-04-07
Bringing in fixes in get_level_dimensions and get_texel_count from master

sectopod:


sectopod:
Seems harmless.
@homu r+

Issue #936, opened 2016-04-06
Move CullFace out of RasterMethod


kvark:
It's currently only affecting the triangle rendering. It should affect lines and points as well. Currently the state is leaked when rendering lines, which hurts @Binero project.

Binero:
Seems like this issue stretches over multiple crates.

kvark:
@Binero yes, but since they are on crates.io one doesn't have to update all at once - instead one can just start with the bottom (`draw_state`) and go up (`gfx_core`, `gfx`, etc).

Binero:
I will have a look at draw_state.

kvark:
Fixed by #945

PR #945, opened 2016-04-13
Move CullFace out of RasterMethod #936

Binero:
See #936.


kvark:
Looks good! Blocked until draw_state 0.5 is published.

kvark:
@homu r+

PR #949, opened 2016-04-17
Added a gfx_defines! macro as discussed on gitter.

eugene2k:
This macro allows to define the vertex, constant and pipeline formats in one whole block, without needing to use one macro per format. All subblocks are optional and can be defined in several such macros. The macro doesn't interfere with the old approach in any way, so you can use it along with the old gfx_vertex_struct/gfx_constant_struct/gfx_pipeline macros.

### DIFF
--- a/src/render/src/macros/pso.rs
+++ b/src/render/src/macros/pso.rs
@@ -176,18 +176,13 @@ macro_rules! gfx_pipeline_inner {
                 // depth-stencil, scissor
                 for _ in 0 .. 1 {
                     $(
-                        match meta.$field.link_depth_stencil(&self.$field) {
-                            Some(d) => {
-                                assert!(meta.$field.is_active());
-                                desc.depth_stencil = Some(d);
-                                continue;
-                            },
-                            None => (),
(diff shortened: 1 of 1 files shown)


src/render/src/macros/mod.rs
```diff
@@ -29,3 +29,56 @@ macro_rules! gfx_format {
         }
     }
 }
+
+
+/// Helps define vertex, constant and pipeline formats
+/// # Example
+/// ```
+/// gfx_defines! {
```
kvark: Do you mind having a simple examples in the "tests/macros.rs" rather then here? Gotta make sure the macro spews correct code.

eugene2k: No problem, I'll do it tomorrow.

bjz:
I'm wondering if you could just remove the other macros entirely now? Might be nice to have 'one way to do it'.

kvark:
@bjz I'd agree if the new one was not on slightly higher level. Multiple ways to do things are harmful when they are different while being on the same level. With `gfx_defines` though, all the previous rules/syntax apply, so there is not much room for an error.

PR #959, opened 2016-04-24
gfx_pipeline macro and the refactor of the performance example

eugene2k:
Ugh... this was a bit chaotic.
@kvark , please have a look at the performance example. There's a major refactoring there, as I was trying to have as little code duplication as possible.


eugene2k:
Probably should've made that last one a fork...

Cargo.toml
```diff
@@ -52,7 +52,7 @@ path = "examples/deferred/main.rs"
 name = "flowmap"
 path = "examples/flowmap/main.rs"
 
-[[target.x86_64-pc-linux-gnu.example]]
+[[example]]
```
kvark: we need to disable it on windows somehow (see AppVeyor build)... if not possible through `Cargo.toml`, maybe we can make the example code empty for some `#[cfg()]` condition?

src/backend/gl/Cargo.toml
```diff
@@ -28,5 +29,7 @@ name = "gfx_device_gl"
 
 [dependencies]
 log = "0.3"
-gfx_gl = "0.3.1"
 gfx_core = { path = "../../core", version = "0.2" }
+
+[build-dependencies]
+gl_generator = "0.5"
```
kvark: missing new line?

src/backend/gl/src/lib.rs
```diff
@@ -745,3 +745,7 @@ impl gfx_core::DeviceFence<Resources> for Device {
         }
     }
 }
+
+mod gl {
+    include!(concat!(env!("OUT_DIR"), "/gl_bindings.rs"));
```
kvark: Isn't this precisely what `gfx_gl` allowed us to avoid? If I'm not mistaken, one of the points behind that crate was that the generated bindings could be re-used without fetching the GL specs every time. @bjz - please correct me if I'm wrong here.

examples/performance/main.rs
```diff
@@ -15,15 +15,17 @@
...
 extern crate time;
 extern crate gfx_gl as gl;
+extern crate gfx_device_gl;
 
 use time::precise_time_s;
 use cgmath::{SquareMatrix, Matrix, Point3, Vector3, Matrix3, Matrix4};
-use cgmath::{Transform, AffineMatrix3, Vector4};
-pub use gfx::format::{I8Scaled, DepthStencil, Rgba8};
+use cgmath::{Transform, Vector4};
+pub use gfx::format::{DepthStencil, Rgba8 as ColorBuffer};
```
kvark: Please rename to `ColorFormat` to match the examples and `gfx_app`

kvark:
@eugene2k thank you!
I think you did great with the `performance` example. There is a number of points I'd like to be addressed before merging. Also, @csherratt may want to have a look, since he is the original author of this example.

Cargo.toml
eugene2k: Ah! I didn't realize that was the reason for that line!

src/backend/gl/src/lib.rs
eugene2k: Sry, haven't realized that. Though it appears only the performance example uses it. I'll revert it back to using gfx-gl for now.

Cargo.toml
eugene2k: Could we go the opposite way and make it build on windows? I'm using linux, so I can't really test anything, though.

kvark: AFAIK, having GLFW dependency on Windows is an eternal pain...

fkaa: @kvark feature-gating it came up in previous discussion IIRC.

eugene2k: How about switching performance to use glutin instead?

kvark: We still need something to test `glfw`, and `performance` does that currently. Just feature-gate it, or permanently disable on Windows.

eugene2k: I'm thinking of writing doc-tests for glfw and sdl instead. SDL needs something to test it anyway.

kvark:
Thanks!
@homu r+

PR #964, opened 2016-05-05
Added Any to gfx::Resources

csherratt:
This makes `amethyst_renderer` much less verbose and shouldn't really hurt here.

### DIFF
--- a/src/core/Cargo.toml
+++ b/src/core/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_core"
-version = "0.3.0"
+version = "0.3.1"
 description = "Core library of Gfx-rs"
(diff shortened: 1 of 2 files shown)


kvark:
I'm not totally up to this change. For one, some `Any` applications actually use `mopa::Any` instead of `std::any::Any`, like `specs`. How about we discuss the `Amethyst` changes first, and then merge if we decide that this is indeed the way to go?

csherratt:
`mopa::Any` extends off the `std::any::Any` trait. I don't actually really care about `Any` I just want to be able to mark them as `std::marker::Reflect` which is unstable. Right now to do it outside of this library is impossibly verbose.

kvark:
Hmm, that makes sense them. `Reflect` is handy.
@homu r+

PR #965, opened 2016-05-06
Made Buffer only able to hold Copyable types.

Binero:
Title says it all.

### DIFF
--- a/src/core/src/factory.rs
+++ b/src/core/src/factory.rs
@@ -230,7 +230,7 @@ pub trait Factory<R: Resources> {
         self.create_buffer_const_raw(cast_slice(data), mem::size_of::<T>(), role, bind)
             .map(|raw| Typed::new(raw))
     }
-    fn create_buffer_dynamic<T>(&mut self, num: usize, role: BufferRole, bind: Bind)
+    fn create_buffer_dynamic<T: Copy>(&mut self, num: usize, role: BufferRole, bind: Bind)
                                 -> Result<handle::Buffer<R, T>, BufferError> {
         let stride = mem::size_of::<T>();
         let info = BufferInfo {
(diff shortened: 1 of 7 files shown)


kvark:
That adds a little bit of code for the constraints everywhere. So while this specific case is fine, I wonder if this is a good direction in general. Like, should we also constraint `T: format::RenderFormat` in `handle::RenderTargetView`? This would add much more code noise, and then other handles would need to be constrained as well. In the end, will it give us much?

Binero:
It'll add noise only on generic code. Most people actually using the Buffers will know what `T` is, so they don't need to worry about adding an annoying `+ Copy`.

kvark:
Yes, and there is a lot of generic code inside gfx itself, plus whatever user abstracts over us.
What is the real benefit though? Just for the sake of self-documented code?

Binero:
Indeed. Just for the sake of self-documented code.

kvark:
Team, what do you think? @cmr @csherratt @bvssvni @fkaa

bvssvni:
I like to put constraints on methods when possible, not on struct declarations.

Binero:
I would agree with that if there was any way to create such a structure where the restraint wasn't there. But the restraint has to be there for the object to even exist.

kvark:
@Binero But you are never creating  structure in isolation, right? If you are using it, you'll get the compile error anyway. The only hypothetical use-case that is not ideal is if you declare the buffer with some parameter but never use it.

Binero:
Sure. My argument is for self-documenting code. It seems yours is just for short code. I'd prefer the first, but I get the counter argument.

kvark:
It looks like the Rust team is following the same scheme as well. Look at the `HashMap` for example:
http://doc.rust-lang.org/std/collections/struct.HashMap.html

It doesn't restrict the type, even though the implementation only works with stuff that has `Eq` and `Hash`, obviously.
I suppose this is convincing enough. Closing.

PR #976, opened 2016-05-21
Implement Error trait for all errors

burtonageo:
Also edits some comments to have better English, and use more try! handling in render/factory.rs. I mostly got the error messages from doc comments, but would be happy to bikeshed them.


kvark:
Could someone confirm that his is indeed the way errors are meant to be handled in Rust? @cmr @bjz @bvssvni

bvssvni:
Looks right!

kvark:
Thanks @burtonageo  and @bvssvni !

PR #981, opened 2016-05-26
Support i686-windows-msvc rustc target

pravic:
Actually, it worked well time ago and became official in Rust stable since 1.8.0.


kvark:
Looks great, thanks!
@homu r+

Issue #961, opened 2016-04-29
Add the CI tests for subpackages


eugene2k:
Currently Travis and Appveyor only run tests for the top-level gfx_app package. Subpackages, such as gfx_core and gfx_window_* are not tested - only built. Travis and Appveyor scripts need to be changed to test all the packages in gfx repository. Cargo has the ability to run tests for a package given its spec: something like `cargo test -p gfx_core` will test the `gfx_core` package. Adding additional test commands to the scripts should be pretty simple.

eugene2k:
Is there any reason to test gfx on rust with gnu toolchain as well as on rust with msvc toolchain? Seems redundant, since all of gfx is written in rust. I can delete the gnu target from appveyor while I'm working on this issue.

kvark:
@retep998 do you see any use of testing the GNU toolchain?

retep998:
It's nice to test the GNU toolchain as it has somewhat different import libraries that can sometimes have issues (such as when the `libd3dcompiler.a` turned out to be for a severely outdated DLL). While I do wish MinGW would just go away, if you can test GNU for free, I don't see why you wouldn't want to keep testing it just in case.

retep998:
That said, do we really need to install mingw through chocolatey? It spams up the build logs like crazy. If the only way to remove the dependency on that is to remove CI for the gnu toolchain, then I'd be okay with that.

eugene2k:
Wouldn't the same kind of testing be done by winapi-rs? It tests both stable and nightly on gnu and msvc.

eugene2k:
As for chocolatey, it seems appveyor only has mingw32 installed, so the only other way to do it would be to manually download and unpack mingw64.

retep998:
winapi doesn't cover integration tests at all. It has no idea whether things actually function correctly. At best I can test whether things link and whether the size/alignment of a type is correct. The d3dcompiler issue for example I wouldn't be able to catch with testing in winapi.

What does gfx depend on that has to compile C/C++ code and so need mingw? If the entire dependency tree is pure Rust then the mingw that is bundled with Rust is good enough and no separate installation is needed.

eugene2k:
The bindings for the miniz library require gcc to build. Though, I'm not sure what requires the miniz library.

kvark:
Piston-image, IIRC

eugene2k:
[...]

PR #987, opened 2016-06-03
Simplify target dependencies

### LINKED ISSUE #982: Simplify target dependencies
… in [Cargo.toml](https://github.com/gfx-rs/gfx/blob/master/Cargo.toml#L27-L37), it can be simplified as

```
[target.'cfg(unix)'.dependencies]
gfx_window_glfw = { path = "src/window/glfw", version = "0.10" }
gfx_window_sdl = { path = "src/window/sdl", version = "0.2" }

[target.'cfg(windows)'.dependencies]
gfx_device_dx11 = { path = "src/backend/dx11", version = "0.2" }
gfx_window_dxgi = { path = "src/window/dxgi", version = "0.2" }
```

[Manual](http://doc.crates.io/specifying-dependencies.html), available since Cargo 0.9.0 (Rust 1.8.0, which is stable month ago).

Just as suggestion.

eugene2k:
fixes #982

P.S. thx, pravic

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,19 +24,11 @@ gfx_device_gl = { path = "src/backend/gl", version = "0.10" }
 gfx_window_glutin = { path = "src/window/glutin", version = "0.11" }
 gfx = { path = "src/render", version = "0.11" }
 
-[target.x86_64-unknown-linux-gnu.dependencies]
+[target.'cfg(unix)'.dependencies]
 gfx_window_glfw = { path = "src/window/glfw", version = "0.10" }
 gfx_window_sdl = { path = "src/window/sdl", version = "0.2" }
 
-[target.x86_64-pc-windows-gnu.dependencies]
-gfx_device_dx11 = { path = "src/backend/dx11", version = "0.2" }
-gfx_window_dxgi = { path = "src/window/dxgi", version = "0.2" }
-
-[target.x86_64-pc-windows-msvc.dependencies]
-gfx_device_dx11 = { path = "src/backend/dx11", version = "0.2" }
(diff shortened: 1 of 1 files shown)


eugene2k:
On a similar note... what do you guys think of feature-gating sdl and glfw?

kvark:
@eugene2k looks great, thanks!
Feature gating SDL and GLFW is the way to go. Would you want to implement it within this PR?

eugene2k:
Might not need to feature-gate it at all. You said before that the reason why sdl was added as a dependency to gfx_app was so that everything could be tested with a single `cargo test`. That doesn't work anyway (see #961). So perhaps we should just remove sdl and glfw from gfx_app altogether.

kvark:
The fact that this doesn't work is a bug (or missing feature) in cargo, not by design, IIRC. But I can't really see a use case where the app launcher would choose `sdl` or `glfw` over `glutin`, so your point still stands. I suppose we can remove `sdl` and `glfw` from `gfx_app` dependencies then, as you suggest.

eugene2k:
There's a complication with that too. If the dependency is optional or not there, cargo can't find its manifest. So if you want to test say `gfx_window_sdl` you can't just run `cargo test -p gfx_window_sdl` in gfx root and have it tested. Instead you have to specify the manifest of the package (or `cd` to its directory) and then cargo will build not only the package itself, but also all of its dependencies even if you had already built them before, which will slow down travis/appveyor.

So, I guess you should probably merge the PR as it is...

kvark:
Yeah, I've been worrying about this too. Let's leave this for now.
@homu r+

PR #988, opened 2016-06-04
Move GLSL shader code out of main.rs

### LINKED ISSUE #895: Move GLSL shader code out of deferred example source
We don't want to pollute the generic source files with platform-dependent code. Same may apply to other examples. Please wait for #861 before starting.
Edit: it would be even better to dynamically load the code from the disk. It would allow changing shaders without rebuilding.

eugene2k:
Moves GLSL shader code out of main.rs and replace gfx_vertex_struct, gfx_constant_struct and gfx_pipeline macros with gfx_defines

P.S. I seem to have some rendering artifacts when I run the deferred example. Can anyone on windows post a screenshot of what it should look like?

closes #895


kvark:
Looks good, thanks!
Need to figure out those artifacts before merging.

eugene2k:
I don't think they're introduced in the merge, though. I didn't touch the shader code at all.

kvark:
Could you post a shot of what you see?

kvark:
My results here with Linux/GL:
gfx-deferred

eugene2k:
I have the same thing. Are there weird dark triangles on windows as well?

kvark:
Ok, we'll look into the triangles outside of this PR. Thanks!
@homu r+

PR #989, opened 2016-06-05
Add CI tests for subpackages

### LINKED ISSUE #961: Add the CI tests for subpackages

eugene2k:
Changes to appveyor and travis scripts to include testing of submodules.
Changes to appveyor script to only download and install mingw when targetting mingw toolchain.
Code example moved from factory documentation into gfx_window_glutin.
Removed an outdated test, added no_run to some tests to prevent them from failing on CI builds.

fixes #961

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
-sudo: false
+sudo: required
 language: rust
 rust:
-- stable
-- nightly
+  - stable
+  - nightly
 before_script:
-- pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
+  - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
 notifications:
   webhooks:
     urls:
-    - https://webhooks.gitter.im/e/7479b6691b7e5e40716a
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -1,14 +1,14 @@
-sudo: false
+sudo: required
```
kvark: why is this required now?

kvark:
What's up with all the whitespace?

.travis.yml
eugene2k: Can't add the sdl repository otherwise. It's not whitelisted.

eugene2k:
Formatted the files while I was working on them and forgot it would show up in diffs. Is that an issue?

kvark:
@homu r+

PR #991, opened 2016-06-12
GL texture size checks and unsigned int samplers

sectopod:


kvark:
I don't think we are clear to publish the patched version of GL crate but otherwise the changes are good to go!
@homu r+

PR #994, opened 2016-06-17
Special index for gl_FragColor for GLSL 120 to 140 shaders

Vinatorul:
This PR fixes an issue when `gfx` was not working properly with `gl` backend if you are using 120-140 versions of shaders.
Also this fixes `index` variable redefinition in `if name.starts_with("gl_FragData") {` branch.


kvark:
@homu r+

PR #999, opened 2016-07-06
Make cast_slice() work only on Pod types.

### LINKED ISSUE #974: cast_slice is not safe
Hi, I'm looking at gfx-rs and thinking of using it, but I can't drop my usual habit of reviewing unsafe code.

`cast_slice` is not a safe abstraction, it should not be exposed as a safe function. If you make this function private, then it's only a concern for your library internals, but it's not good that it's public.

**Example breakage**. `usize` and `&T` (for any `T`) are both `Copy`, but of course transmuting arbitrary integer into a reference does not make a valid reference (it may be dangling, mutabily aliased, not aligned, or null for example).

Other issues: Breaks type safety. A type may be `Copy` but only allow construction through explicitly exposed functions, that preserve certain invariants for example.

Recommended resolution: Demote the function to private if possible. If you need "pod casting", make an `unsafe trait` that is implemented for explicitly Pod compatible types, like the integer types, and only allow the cast to those types.

vickenty:
This makes structures defined with gfx_vertex_struct! and
gfx_constant_struct! also marked as Pod which is not entirely
correct. These structures should only contain Pod types, but there is no
way to guarantee that.

Fixes #974.

### DIFF
--- a/src/core/src/factory.rs
+++ b/src/core/src/factory.rs
@@ -21,7 +21,7 @@ use std::error::Error;
 use std::fmt;
 use std::mem;
 use {handle, format, mapping, pso, shade, target, tex};
-use {Capabilities, Resources};
+use {Capabilities, Resources, Pod};
 use {VertexShader, GeometryShader, PixelShader, ShaderSet};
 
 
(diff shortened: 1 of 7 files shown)


kvark:
Looks good to me, thanks!
@csherratt care to have a look?

csherratt:
@homu r+

Issue #283, opened 2014-08-25
Use `glBufferStorage` for buffer initialization, if available


kvark:
Available since GL-4.4
```
void glBufferStorage​(GLenum target​, GLsizeiptr size​, const GLvoid * data​, GLbitfield flags​);
```

ebaklund:
Is the instancing example the only test case we have to test the alternative buffer initialization?

kvark:
Fixed by #1026

PR #1001, opened 2016-07-12
Proposal to #283

ebaklund:
A bit worried that init_buffer() is not much used.
Would like to have more tests than only --example instancing.
Also, semantics of usage-flags differ significantly between glBufferStorage and glBufferData.
Kind regards,
Erik.

### DIFF
--- a/src/backend/gl/src/factory.rs
+++ b/src/backend/gl/src/factory.rs
@@ -101,19 +101,35 @@ impl Factory {
     fn init_buffer(&mut self, buffer: Buffer, info: &f::BufferInfo) {
         let gl = &self.share.context;
         let target = role_to_target(info.role);
-        let usage = match info.usage {
-            f::Usage::GpuOnly    => gl::STATIC_DRAW,
-            f::Usage::Const      => gl::STATIC_DRAW,
-            f::Usage::Dynamic    => gl::DYNAMIC_DRAW,
-            f::Usage::CpuOnly(_) => gl::STREAM_DRAW,
-        };
-        unsafe {
-            gl.BindBuffer(target, buffer);
-            gl.BufferData(target,
-                info.size as gl::types::GLsizeiptr,
-                0 as *const gl::types::GLvoid,
-                usage
-            );
+        if self.share.private_caps.buffer_storage_supported {
+            let usage = match info.usage {
+                f::Usage::GpuOnly    => 0,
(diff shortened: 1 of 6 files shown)


ebaklund:
I walked the dog and realized that of course, init_buffer() is indeed used - a lot.
Running through the examples I see that my proposal does not work when animations are involved.
So, it might be that glBufferStorage() cannot be as easy an substitute for glBufferData() as I first thought.

ebaklund:
Closing until solution is found.

kvark:
I'll check it out tomorrow and should come up with some ideas.

ebaklund:
Seems the animated examples are broken upstream. Those should be fixed before we go further with glBufferStorage(),

ebaklund:
Migration to std::time broke the animations.

ebaklund:
To be clear: My glBufferStore() proposal seem to work once the std::time related regressions were fixed.

src/backend/gl/src/factory.rs
```diff
@@ -101,19 +101,35 @@ impl Factory {
...
-            gl.BufferData(target,
-                info.size as gl::types::GLsizeiptr,
-                0 as *const gl::types::GLvoid,
-                usage
-            );
+        if self.share.private_caps.buffer_storage_supported {
+            let usage = match info.usage {
+                f::Usage::GpuOnly    => 0,
+                _                    => gl::MAP_WRITE_BIT | gl::DYNAMIC_STORAGE_BIT,
+            };
```
kvark: don't worry about this too much, the usage bits are to be refactored soon

src/backend/gl/src/info.rs
```diff
@@ -248,6 +249,7 @@ pub fn get(gl: &gl::Gl) -> (Info, Capabilities, PrivateCaps) {
         immutable_storage_supported:       info.is_version_or_extension_supported(4, 2, "GL_ARB_texture_storage"),
         sampler_objects_supported:         info.is_version_or_extension_supported(3, 3, "GL_ARB_sampler_objects"),
         program_interface_supported:       info.is_version_or_extension_supported(4, 3, "GL_ARB_program_interface_query"),
+        buffer_storage_supported:          info.is_version_or_extension_supported(4, 4, "GL_ARB_program_interface_query"),
```
kvark: wrong extension check (`GL_ARB_program_interface_query`)

kvark:
Thanks @ebaklund ! Please fix the extension name to proceed.

kvark:
@homu r+

Issue #898, opened 2016-03-07
Constant buffers are mis-used for structured data


sectopod:
Several examples appear to use constant buffers that only consists of a single array of data:

``` glsl
    uniform u_LightPosBlock {
        vec4 offs[NUM_LIGHTS];
    };
```

The code side effectively treats them as structured buffers: we pass the number of elements on creation, the type is `ConstantBuffer<[f32; 4]>` instead of `ConstantBuffer<LightPosBlock>`, and we even receive a slice of them for updating the buffer. In other words, the buffer API is currently catering to structured buffer users.

We need to fix the constant buffers to stop pretending to be arrays! We need explicit support for structured buffers - these are just regular typed buffers with an SRV for binding to the shader. Problem is - there might be different typing:
1. Simple format, `T: gfx::format::BufferFormat`. Supporting these is straightforward, since we have the same typing for textures.
2. User structure, `T` - not bound by anything.
3. Raw buffer - untyped (or `()` typed for our convenience). Not sure if we have to support these but they exist. The shader to free to load from it in any possible way.

kvark:
This is unfortunate, but not going to be fixed in pre-ll, closing.

PR #1004, opened 2016-07-27
Proper PSO errors, with constant offsets matching

### LINKED ISSUE #722: Check the layout of uniform buffers

### LINKED ISSUE #896: Better PSO initialization errors
Currently what I get is `DescriptorInit(some_id, some_option)`. It would be great to at least return the name of the thing that's getting not found or mismatched. Problem is - we'd optimally want to borrow these names from `ProgramInfo` instead of cloning them in order to avoid heap allocation. However, some of the PSO constructors in `src/render/factory.rs` are linking the program inside them, so they can't return anything out that's bound to the program lifetime...

Perhaps, we could use `Cow<str>` there? Getting understandable errors here is extremely important to assist people debugging their code.

sectopod:
API-breaking change (although no one will notice)

Fixes #722, fixes #896

Status: some DX11 examples fail to match CB elements: `deferred`, `ubo_tilemap`.


kvark:
I assume arrays are the biggest problem, and it interferes with #898.
Perhaps, we can match only the first element offset and the stride of it? No need to go through all the elements.

kvark:
Please provide an example error message now that we can cheer on it ;)

sectopod:
My windows machine HDD has failed. Unfortunately, I'll not be able to finish this fix properly for the DX samples to run. Good news is - all the code is here, in PR, I didn't lose anything locally. If someone wants to pick it up and fix DX samples - would be great!

sectopod:
Heads up - I've got the new windows setup working, will be able to finish this PR myself.

sectopod:
Example error for a mismatched offset inside a constant buffer:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DescriptorInit(ConstantBuffer("Locals", Some(Offset("u_Transform", 0))))', ../src/libcore\result.rs:785

sectopod:
I didn't bump the crate versions, but other than that - the PR is ready to go!

kvark:
At last, it has arrived!
@homu r+

src/backend/dx11/src/mirror.rs
```diff
@@ -86,6 +86,27 @@ fn map_texture_type(tt: winapi::D3D_SRV_DIMENSION) -> s::TextureType {
     }
 }
 
+fn map_container(stype: &winapi::D3D11_SHADER_TYPE_DESC) -> s::ContainerType {
+    use gfx_core::shade::Dimension as Dim;
+    //TODO: use `match` when winapi allows
```
kvark: @retep998 is this something on your radar?

retep998: Oh, the method I use for enums in winapi 0.2 doesn't allow you to match on it? Well, this will be fixed in winapi 0.3 for sure because I will be making enums simple integer constants and you can definitely match on integers.

PR #1006, opened 2016-07-29
(v0.12) Old glClear code bath as a fallback

### LINKED ISSUE #1005: Crash in ZoC on OSX
> `self.context.encoder.flush(&mut self.context.device);`
> error: Process didn't exit successfully: `target/debug/zoc` (signal: 11, SIGSEGV: invalid memory reference)

Need to get an OSX device and check for ourselves:
https://github.com/ozkriff/zoc/tree/WIP_RC_gfx

cc @ozkriff

kvark:
Fixes #1005 in v0.12 branch

### DIFF
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_device_gl"
-version = "0.11.0"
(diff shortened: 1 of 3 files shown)


kvark:
@homu r+

Issue #1008, opened 2016-08-01
Can't bind uniform buffer


tmielcza:
I made a small gfx test that crashes with:
```
thread 'main' panicked at 'Error after executing command BindProgram(0): Invalid Enum'
```
when I try to create data for the draw pipeline:
```rust
    let data = pipe::Data {
        vertex_buffer: v_buf,
        some_uniform: [0., 0.],
        locals: factory.create_constant_buffer(1),
        out_color: target.clone(),
    };
```
After further investigation, I found the guilty call [here](https://github.com/gfx-rs/gfx/blob/d2af23073e303bbfc04ff10999db886dce88fc52/src/backend/gl/src/factory.rs#L126):
```rust
gl.BindBuffer(gl::BUFFER_UNIFORM, buffer);
```
I checked my OpenGL extensions and did not find ```GL_ARB_uniform_buffer_object```.
I'm on OpenGL 2.1.

kvark:
Thanks! We should return an error during the pipeline creation when UBOs are not supported.

tmielcza:
Ok. So, I can't use gfx on my laptop ? :'(

kvark:
Just don't use constant buffers (`gfx::ConstantBuffer<>` component). Instead, pass stuff in globals (`gfx::Global<>` component).

tmielcza:
Okay !
Also there's another bug, I don't know if it's related.
```thread 'main' panicked at 'gl function was not loaded'```
The function seems to be ```gl.BindFramebuffer()```.
```GL_ARB_framebuffer_object``` is missing.

kvark:
Uh, sorry about it. This one is a blocker, I'll try to have a look tonight.

tmielcza:
Thanks a lot!

kvark:
Would you be able to get the latest version and run our examples to verify no more fixes are needed for your laptop? If you run the basics one just fine, I'll publish the fixed version of `gfx_device_gl`.

tmielcza:
Works like a charm for almost every example.
The cube is black, though, I don't know why.
Thanks a lot for your quick intervention !

tmielcza:
Here is the result of apitrace:  [trace.txt](https://github.com/gfx-rs/gfx/files/399177/trace.txt)
The log (I cut it down since It looped):  [log.txt](https://github.com/gfx-rs/gfx/files/399206/log.txt)
Incidentally, near all the examples run with a d3d backend. How can I run them using OpenGL?

[...]

PR #1011, opened 2016-08-03
[GL] Proper handling for FBO and UBO support

kvark:
Fixes most of #1008 
We might need to add some more stuff in here before publishing new GL crate.


kvark:
I thought maybe the black cube of #1008 has anything to do with more fixes, but it appears to be DX11-related, so I'm merging this.
@homu r+

kvark:
Will publish the new version later tonight, hold tight!

PR #1012, opened 2016-08-04
Rebuild all example HLSL shaders

msiglreith:
Rerun the `compile.cmd` for all examples and equalized/fixed the compile scripts.

This should fix the `cube` example as mentioned in #1008. Also tested the `triangle` example to avoid regression (nothing changed).


kvark:
Awesome, thank you!
@homu r+

PR #1013, opened 2016-08-08
Fix incorrect time in performance example

Mattbazooka:
Changed duration_to_f64 to duration_to_ms, which fixes the displayed time for gfx option.


kvark:
Thanks!
@homu r+

PR #1015, opened 2016-08-12
Improve Travis builds

### LINKED ISSUE #998: Speed up Travis builds
https://llogiq.github.io/2016/07/05/travis.html

kvark:
Fixes #998

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,20 @@
-sudo: required
+sudo: false
+os:
+  - linux
+  - osx
 language: rust
+cache: cargo
 rust:
   - stable
   - nightly
+matrix:
+  allow_failures:
+    - os: osx # temporarily
 before_script:
-  - pip install 'travis-cargo<0.2' --user && export PATH=$HOME/.local/bin:$PATH
+  - pip install 'travis-cargo<0.2' --user
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export PATH=$HOME/.local/bin:$PATH ; fi
+  - ls $HOME/Library/Python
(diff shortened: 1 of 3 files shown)


kvark:
Looks to be blocked by https://github.com/travis-ci/apt-source-whitelist/issues/45 due to `libsdl2-dev`

kvark:
Hmm, I'm having problems with building `glfw` and `xcb` stuff without `sudo`. I included the corresponding packages but the build scripts still fail. Any help would be appreciated.

eugene2k:
Yeah, I inserted sudo in there because of libsdl. It worked without sudo and on linux you don't have to put glfw or xcb into apt packages. I figure on macOS you don't need them either. Don't know why you put them in there.

kvark:
So glfw and xcb also require sudo? I put them into packages simply because of the errors in their build scripts, hoping that would help.

eugene2k:
I don't think they do. Sorry, I hadn't looked at the errors travis reports and thought there was something wrong with the travis script. I only just looked and realized rust-xcb isn't compiling. What file is its build script looking for that it can't find?

kvark:
I don't know. One way to get more info is to add RUST_BACKTRACE=1 to Travis script. Or even run it through 'strace'. Or maybe do RUST_LOG=debug... Please feel free to experiment if willing to help.

eugene2k:
I'm stuck at glfw for the moment. glfw now requires a higher version of cmake than the one in the kalakris repository. There are other problems with other repositories that are known to contain cmake. So, at the moment, unless there's a whitelisted repository that I missed, we have to use a repository that isn't whitelisted. And that means using sudo. To resolve the original issue(#998), rather than removing sudo, it might be better to switch to to trusty images.

kvark:
Closing in favour of #1091

PR #1016, opened 2016-08-14
Vulkan: Some groundwork for windows vulkan support

msiglreith:
Correctly locate the vulkan dynamic library on windows, use winit for window handling and implement surface creation.

NOT tested on an unix system so far with xcb.
Setting the correct `"VK_KHR_*_surface` extensions for each platform is left as future work for the moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,9 +39,9 @@ script:
   - travis-cargo test -- -p gfx_window_glutin
   - travis-cargo test -- -p gfx_window_glfw
   - travis-cargo test -- -p gfx_window_sdl
-# - travis-cargo test -- -p gfx_device_vulkan
-# - travis-cargo test -- -p gfx_window_vulkan
-  - travis-cargo test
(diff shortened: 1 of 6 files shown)


kvark:
Looks great, thanks! Now, just need to preserve that linux build (fortunately tested by Travis CI).

kvark:
Also, if vulkan compiles on windows, we need to add `--feature vulkan` to the AppVeyor config, similarly to our Travis/Unix one.

msiglreith:
Indeed! You prefer a separate PR for the CI changes to avoid possible conflicts with #1015 or should I merge them into this one?

kvark:
No, please do it all together. #1015 is not too big, will merge just fine
afterwards ;)

kvark:
Looks like the failure is caused by `cmake` (see [comment](https://github.com/gfx-rs/gfx/pull/1015#issuecomment-239664820)). Merging.

PR #1018, opened 2016-08-16
Vulkan: Working on swapchain creation

msiglreith:
The backend will now load the required extensions for each platform for the surface creation. Additionally improved the swapchain creation to address serveral validation layer errors (checking for support of formats, capabilities, etc.).
Additionally switched to the (more common?) BGRA8 color format for the swapchain images.

I assume we need to shift some responsibilities for the the swapchain creation to the application to select its prefered formats and settings depending on the capabilities offered by the GPU.
(e.g. my system doesn't support RGBA8 for swapchains nor D24_S8 as depth-stencil format)

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -205,7 +205,8 @@ pub fn create(app_name: &str, app_version: u32, layers: &[&str], extensions: &[&
     info!("Chosen physical device {:?} with queue family {}", dev.device, qf_id);
 
     let mvid_id = dev.memory.memoryTypes.iter().take(dev.memory.memoryTypeCount as usize)
-                            .position(|mt| mt.propertyFlags & vk::MEMORY_PROPERTY_DEVICE_LOCAL_BIT != 0)
+                            .position(|mt| (mt.propertyFlags & vk::MEMORY_PROPERTY_DEVICE_LOCAL_BIT != 0)
+                                        && (mt.propertyFlags & vk::MEMORY_PROPERTY_HOST_VISIBLE_BIT != 0))
                             .unwrap() as u32;
     let msys_id = dev.memory.memoryTypes.iter().take(dev.memory.memoryTypeCount as usize)
                             .position(|mt| mt.propertyFlags & vk::MEMORY_PROPERTY_HOST_COHERENT_BIT != 0)
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -80,6 +80,7 @@ pub struct Share {
     inst_pointers: vk::InstancePointers,
     device: vk::Device,
     dev_pointers: vk::DevicePointers,
+    physical_device: vk::PhysicalDevice,
     handles: RefCell<gfx_core::handle::Manager<Resources>>,
 }
 
@@ -91,6 +92,9 @@ impl Share {
     }
     pub fn get_device(&self) -> (vk::Device, &vk::DevicePointers) {
         (self.device, &self.dev_pointers)
+    }
+    pub fn get_physical_device(&self) -> vk::PhysicalDevice {
+        self.physical_device
     }
 }
 
@@ -273,6 +277,7 @@ pub fn create(app_name: &str, app_version: u32, layers: &[&str], extensions: &[&
         inst_pointers: inst_pointers,
         device: device,
         dev_pointers: dev_pointers,
+        physical_device: dev.device,
         handles: RefCell::new(gfx_core::handle::Manager::new()),
     });
     let gfx_device = command::GraphicsQueue::new(share.clone(), queue, qf_id as u32);
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -21,6 +21,7 @@ extern crate vk_sys as vk;
 use std::{fmt, iter, mem, ptr};
 use std::cell::RefCell;
 use std::sync::Arc;
+use std::ffi::CStr;
 use shared_library::dynamic_library::DynamicLibrary;
 
 pub use self::command::{GraphicsQueue, Buffer as CommandBuffer};
@@ -92,6 +93,17 @@ impl Share {
     }
 }
 
+const SURFACE_EXTENSIONS: &'static [&'static str] = &[
+    // Platform-specific WSI extensions
+    "VK_KHR_xlib_surface",
+    "VK_KHR_xcb_surface",
+    "VK_KHR_wayland_surface",
+    "VK_KHR_mir_surface",
+    "VK_KHR_android_surface",
+    "VK_KHR_win32_surface",
+];
+
+
 pub fn create(app_name: &str, app_version: u32, layers: &[&str], extensions: &[&str],
               dev_extensions: &[&str]) -> (command::GraphicsQueue, factory::Factory, SharePointer) {
     use std::ffi::CString;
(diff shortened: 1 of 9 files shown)


kvark:
> e.g. my system doesn't support RGBA8 for swapchains nor D24_S8 as depth-stencil format

Uh, oh, that sucks. I recall @fkaa also complaining that our default `D24_S8` does not exist on Metal.
Looks like we need to have the defaults being different for different backends? Or maybe some feature levels. Either way, it doesn't affect gfx_core or gfx itself, just the example launcher.

src/backend/vulkan/src/lib.rs
```diff
@@ -122,12 +139,38 @@ pub fn create(app_name: &str, app_version: u32, layers: &[&str], extensions: &[&
...
+        let mut out = Vec::with_capacity(num as usize);
+        assert_eq!(vk::SUCCESS, unsafe {
+            entry_points.EnumerateInstanceExtensionProperties(ptr::null(), &mut num, out.as_mut_ptr())
+        });
+        unsafe { out.set_len(num as usize); }
+        out
+    };
 
+    // Check our surface extensions against the available extensions
+    let surface_extensions = SURFACE_EXTENSIONS.iter().filter(|ext| {
```
kvark: maybe `filter_map`?

src/backend/vulkan/src/lib.rs
```diff
@@ -122,12 +139,38 @@ pub fn create(app_name: &str, app_version: u32, layers: &[&str], extensions: &[&
...
+        for inst_ext in instance_extensions.iter() {
```
kvark: maybe 'find()'?

src/backend/vulkan/src/lib.rs
```diff
@@ -151,12 +194,20 @@ pub fn create(app_name: &str, app_version: u32, layers: &[&str], extensions: &[&
...
 
-    let mut physical_devices: [vk::PhysicalDevice; 4] = unsafe { mem::zeroed() };
-    let mut num = physical_devices.len() as u32;
-    assert_eq!(vk::SUCCESS, unsafe {
-        inst_pointers.EnumeratePhysicalDevices(instance, &mut num, physical_devices.as_mut_ptr())
-    });
-    let devices = physical_devices[..num as usize].iter()
+    let physical_devices = {
+        let mut num = 0;
+        assert_eq!(vk::SUCCESS, unsafe {
```
kvark: I'm fine with this code but I don't think it's worth it. Having a maximum bound of the number of devices here helps reducing the code size.

src/backend/vulkan/src/lib.rs
```diff
@@ -167,7 +218,8 @@ pub fn create(app_name: &str, app_version: u32, layers: &[&str], extensions: &[&
     info!("Chosen physical device {:?} with queue family {}", dev.device, qf_id);
 
     let mvid_id = dev.memory.memoryTypes.iter().take(dev.memory.memoryTypeCount as usize)
-                            .position(|mt| mt.propertyFlags & vk::MEMORY_PROPERTY_DEVICE_LOCAL_BIT != 0)
+                            .position(|mt| (mt.propertyFlags & vk::MEMORY_PROPERTY_DEVICE_LOCAL_BIT != 0)
+                                        && (mt.propertyFlags & vk::MEMORY_PROPERTY_HOST_VISIBLE_BIT != 0))
```
kvark: Why does it need to be `HOST_VISIBLE`?

src/window/vulkan/src/lib.rs
```diff
@@ -185,13 +235,13 @@ pub fn init<T: gfx_core::format::RenderFormat>(title: &str, width: u32, height:
...
         imageArrayLayers: 1,
-        imageUsage: vk::IMAGE_USAGE_COLOR_ATTACHMENT_BIT | vk::IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT,
+        imageUsage: vk::IMAGE_USAGE_COLOR_ATTACHMENT_BIT,
         imageSharingMode: vk::SHARING_MODE_EXCLUSIVE,
         queueFamilyIndexCount: 1,
         pQueueFamilyIndices: &0,
         preTransform: vk::SURFACE_TRANSFORM_IDENTITY_BIT_KHR,
         compositeAlpha: vk::COMPOSITE_ALPHA_OPAQUE_BIT_KHR,
-        presentMode: vk::PRESENT_MODE_FIFO_RELAXED_KHR,
+        presentMode: vk::PRESENT_MODE_FIFO_KHR, // required to be supported
```
kvark: Is `RELAXED` not required? Is it not supported on your HW?

kvark:
I have a few ideas to (temporarily) fix glfw breakage (cc @eugene2k):
  1. Use an older glfw-rs package explicitly, one that still required our cmake version
  2. Feature-gate glfw entirely and disable in Travis. Still going to be tested by AppVeyor.

eugene2k:
No need at the moment. I found a whitlisted repo (actually I found I forgot to add an extra package when I tested it before and haven't realized that was the problem) and glfw compiles now. I'm trying to find out what's wrong with xcb now. strace is of absolutely no help, btw :)

kvark:
Awesome, glad to hear that @eugene2k ! If `xcb` is not obvious, please proceed with `glfw` fix PR so that other work (like this PR) shows green.

eugene2k:
#1019 submitted. Don't think the travis build will show green, though

src/backend/vulkan/src/lib.rs
msiglreith: I added it mostly to silence to validation layers, which report a warning or an error if it's not explicitly queried. 
According to the vulkan specs:

> If pPhysicalDeviceCount is smaller than the number of physical devices available, VK_INCOMPLETE will be returned instead of VK_SUCCESS, to indicate that not all the available physical devices were returned.

src/backend/vulkan/src/lib.rs
msiglreith: As we allow to map the `mvid` memory to the host on other points in the code for reading/writing iirc.

src/window/vulkan/src/lib.rs
msiglreith: Only `PRESENT_MODE_FIFO_KHR` is required to be supported following the specs. Checking the queried present modes for better options (ie `PRESENT_MODE_MAILBOX_KHR`) could be done.

msiglreith:
Thanks for reviewing! I'm trying to address the other code lines and fix the error reported by appveyor. I also need to handle the new color format for Metal.

src/backend/vulkan/src/lib.rs
kvark: oh, that's good to know

src/backend/vulkan/src/lib.rs
kvark: hmm. I think we should not request it explicitly, but instead return an error on trying to map it if the memory heap doesn't support it. This doesn't have to be a part of your PR, no worries.

kvark:
Don't worry to much about Metal. It is slightly broken anyway now, so you are not making it much worse :) Let me know when your PR is ready.

msiglreith:
The PR would be ready, unfortunatley the appveyor-gnu buildbot remains broken for some reason.

kvark:
Indeed, the AppVeyor wasn't able to get `mingw` (again) for some reason. Waiting for another Travis pass...

kvark:
Hmm, we need to make GitHub re-merge this with #1019. @msiglreith - could you force-push the last commit or rebase?

kvark:
@homu r+

PR #1019, opened 2016-08-16
Update .travis.yml to use a newer version of cmake

eugene2k:
This fixes the glfw issue. xcb is unsolved for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,13 @@ addons:
   apt:
     sources:
       # install a newer cmake since at this time Travis only has version 2.8.7
-      - kalakris-cmake
-      # add a repository for libsdl2-dev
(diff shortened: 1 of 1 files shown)


msiglreith:
Do we still need the `xcb` crate dependency? #1016 moved all the xcb related window creation to winit. So we might be able to drop it completley.

kvark:
`winit` doesn't support `xcb`, it only uses `xlib` on linux. So we still need `xcb` for systems that don't sport `xlib`, even if behind a feature gate.

kvark:
could you rebase please?

eugene2k:
Rebased. And having a WTH moment here. xcb compiles wiithout complaints now, it seems...

kvark:
@homu r+

PR #1022, opened 2016-08-20
Make trailing comma optional in gfx_defines! declarations

### LINKED ISSUE #1021: gfx_defines! macro throws unintuitive error when missing trailing commas
Following the sample triangle code, I decided to extend the `gfx_defines!` block to include texture coordinates when I discovered:

```rust
gfx_defines!{
	vertex Vertex {
		pos: [f32; 2] = "a_Pos",
		color: [f32; 3] = "a_Color",
		texcoords: [f32; 2] = "a_Texcoords" // <-- error: no rules expect the token `}`
	}
}
```

As I understand it (and I am new to Rust), its idiomatic to use trailing commas in initializer blocks, but not required.

Is this a case where it would be more clear to allow missing a trailing comma?

eugene2k:
fixes #1021

### DIFF
--- a/src/render/src/macros/mod.rs
+++ b/src/render/src/macros/mod.rs
@@ -59,5 +59,15 @@ macro_rules! gfx_defines {
             $keyword $name { $($field : $ty = $e,)+ }
         }
         gfx_defines!($($tail)+);
-    }
+    };
+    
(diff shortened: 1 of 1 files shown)


kvark:
@homu r+
Thanks!

PR #1027, opened 2016-08-23
dx11 tessellation test

ahmedaliadeel:
Ignores Geomtry shader for now, we can create a TessellatedGeomtry
variant for complete pipeline later....

Opengl details are just for sake of filling, i havent tested it yet.


kvark:
Would you be able to patch (not necessarily test) Vulkan and Metal backends too? Travis CI is failing.

src/backend/dx11/src/factory.rs
```diff
@@ -425,11 +441,30 @@ impl core::Factory<R> for Factory {
...
+                    ds: ptr::null_mut(),
                     gs: vs.object as *mut ID3D11GeometryShader,
                     ps: ps.object as *mut ID3D11PixelShader,
                     vs_hash: vs.code_hash,
                 }
             },
+            &core::ShaderSet::Tessellated(ref vs, ref hs, ref ds, ref ps) => {
+                let (vs, hs, ds, ps) = (vs.reference(fh), hs.reference(fh), ds.reference(fh), ps.reference(fh));
+              
+			  populate_info(&mut info, Stage::Vertex,   vs.reflection);
```
kvark: some whitespace messed up

src/core/src/lib.rs
```diff
@@ -145,7 +148,9 @@ pub enum Primitive {
     /// Every three consecutive vertices represent a single triangle. For example, with `[a, b, c,
     /// d]`, `a`, `b`, and `c` form a triangle, and `b`, `c`, and `d` form a triangle.
     TriangleStrip,
-    //Quad,
+    /// Quad,
+    /// 4 point quad patch
+    QuadList
```
kvark: There is a problem here. The implementation you got only supports the quad list for tessellation, while the name doesn't bear this limitation, so some users may be confused trying to render non-tessellated quads.

src/backend/gl/src/command.rs
```diff
@@ -32,6 +32,7 @@ fn primitive_to_gl(primitive: c::Primitive) -> gl::types::GLenum {
         TriangleList => gl::TRIANGLES,
         TriangleStrip => gl::TRIANGLE_STRIP,
         //TriangleFan => gl::TRIANGLE_FAN,
+        QuadList => gl::PATCHES
```
kvark: Note to self: don't forget to call `glPatchParameteri​` somewhere, not to mention the need to fix the GL backend to support tessellation properly and run the new example ;)

src/core/src/lib.rs
kvark: Maybe `QuadListPatch`?

kvark:
Things to address:
  - `QuadList` name
  - Vulkan & Metal backends (compile only)
  - update the screenshot for the new example

sectopod:
Closing in favor of #1087

PR #1031, opened 2016-08-31
Reorganise core

Bastacyclop:
reorganise the core module a bit :)


kvark:
Note: this is a breaking change!
@homu r+

PR #1033, opened 2016-09-02
Make sure the viewport is never of size 0

### LINKED ISSUE #1003: Rendering into 1D texture uses viewport of heigth 1
reported by @Bastacyclop 
2d2826e4a40448ad8f480d5ae1e2a5e2

kvark:
Fixes #1003
cc @Bastacyclop


Bastacyclop:
seems like a better solution than changing the output of `get_dimensions` :)

kvark:
I believe this is not a sound solution, closing.

PR #1035, opened 2016-09-02
Make sure the viewport is never of size 0

### LINKED ISSUE #1003: Rendering into 1D texture uses viewport of heigth 1

Bastacyclop:
Fixes #1003

### DIFF
--- a/src/backend/vulkan/src/command.rs
+++ b/src/backend/vulkan/src/command.rs
@@ -145,15 +145,13 @@ impl command::Buffer<Resources> for Buffer {
     fn bind_samplers(&mut self, _: &[pso::SamplerParam<Resources>]) {}
 
     fn bind_pixel_targets(&mut self, pts: pso::PixelTargetSet<Resources>) {
-        use std::cmp::max;
-
         let (dev, vk) = self.share.get_device();
-        let dim = pts.unwrap_dimensions();
+        let view = pts.get_view();
         let vp = vk::Viewport {
             x: 0.0,
             y: 0.0,
-            width: max(dim.0, 1) as f32,
-            height: max(dim.1, 1) as f32,
+            width: view.0 as f32,
+            height: view.1 as f32,
             minDepth: 0.0,
             maxDepth: 1.0,
         };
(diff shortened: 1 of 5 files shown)


src/core/src/pso.rs
```diff
@@ -258,27 +259,36 @@ impl<R: Resources> PixelTargetSet<R> {
...
             self.stencil = Some(view.clone());
         }
-        self.size = max(self.size, dim);
+        self.set_size(dim);
+    }
+
+    fn set_size(&mut self, dim: texture::Dimensions) {
+        debug_assert!(!self.sized || self.size == dim);
+        self.sized = true;
+        self.size = dim;
```
kvark: let's use `Option<>` instead?

Bastacyclop: should the backend unwrap it or handle the case where there is no pixel target differently ?

kvark: perhaps, use `unwrap_or` at least, may be changed later to some error if we want

src/core/src/pso.rs
```diff
@@ -257,28 +257,42 @@ impl<R: Resources> PixelTargetSet<R> {
...
+        self.set_dimensions(dim);
+    }
+
+    fn set_dimensions(&mut self, dim: texture::Dimensions) {
+        debug_assert!(self.dimensions.map(|d| d == dim).unwrap_or(true));
+        self.dimensions = Some(dim);
+    }
+
+    /// Get the rendering dimensions or empty ones
+    pub fn unwrap_dimensions(&self) -> texture::Dimensions {
```
kvark: Might as well call this `get_viewport` and put a `max` in there ;)

kvark:
Looks good, thanks! Waiting for AppVeyor now.

src/backend/vulkan/src/command.rs
```diff
@@ -146,11 +146,12 @@ impl command::Buffer<Resources> for Buffer {
 
     fn bind_pixel_targets(&mut self, pts: pso::PixelTargetSet<Resources>) {
         let (dev, vk) = self.share.get_device();
+        let view = pts.get_view();
         let vp = vk::Viewport {
             x: 0.0,
             y: 0.0,
-            width: pts.size.0 as f32,
-            height: pts.size.1 as f32,
+            width: max(view.0, 1) as f32,
```
kvark: these `max` should not be needed

kvark:
@homu r+

PR #1037, opened 2016-09-05
Vulkan shader reflection

msiglreith:
First steps into reflecting the SPIR-V modules for defining the descriptor set.

Open issues:
* Some properties aren't correctly reflected (marked as `TODO`). Are they actually needed?
* The current slot based system might need some tweaks to avoid running out of available slots.
* Using a combination of Texture/Sampler requires some special handling, similiar to dx11: They must be defined separately in the shader (sampler has an additional `_` in the name).
* Error handling should be improved (lots of `unwraps()`)


src/backend/vulkan/src/mirror.rs
```diff
@@ -0,0 +1,386 @@
...
+
+        _ => None,
+    }
+}
+
+fn map_instruction_to_type(module: &spirv_utils::RawModule, instr: &instruction::Instruction) -> Option<Type> {
+    use spirv_utils::instruction::{Decoration, Instruction};
+    let id_ty = match *instr {
+        Instruction::TypeBool { result_type } => {
+            Some((result_type, Ty::Basic(BaseType::Bool, ContainerType::Single)))
```
kvark: could re-use `map_scalar_to_basetype`, I suppose

src/backend/vulkan/src/mirror.rs
```diff
@@ -0,0 +1,386 @@
...
+            Some((result_type, Ty::Sampler))
+        },
+        Instruction::TypeImage { result_type, type_id, dim, arrayed, multisampled, .. } => {
+            Some((result_type, Ty::Image(map_scalar_to_basetype(module.def(type_id).unwrap()).unwrap(), map_image_to_texture_type(dim, arrayed, multisampled))))
+        },
+
+        _ => None,
+    };
+
+    if let Some((id, ty)) = id_ty {
```
kvark: using `map` on the option would be prettier

src/backend/vulkan/src/mirror.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    let module = spirv_utils::RawModule::read_module(code).unwrap();
+
+    let mut entry_points = Vec::new();
+    let mut variables = Vec::new();
+    let mut types = Vec::new();
+    for instr in module.instructions() {
+        match *instr {
+            Instruction::EntryPoint { execution_model, ref name, ref interface, .. } => {
+                if let Some(stage) = map_execution_model_to_stage(execution_model) {
```
kvark: will we need to signal some sort of error for that in the future?

src/backend/vulkan/src/mirror.rs
```diff
@@ -0,0 +1,386 @@
...
+        entry_points: entry_points,
+        variables: variables,
+        types: types,
+    }
+}
+
+pub fn populate_info(info: &mut shade::ProgramInfo, stage: shade::Stage, reflection: &SpirvReflection) {
+    if stage == shade::Stage::Vertex {
+        // record vertex attributes
+        if let Some(entry_point) = reflection.entry_points.iter().find(|ep| ep.name == "main" && ep.stage == stage) {
```
kvark: what happens if there is no `main()`? is it even possible?

src/backend/vulkan/src/mirror.rs
```diff
@@ -0,0 +1,386 @@
...
+            for attrib in entry_point.interface.iter() {
+                if let Some(var) = reflection.variables.iter().find(|var| var.id == *attrib && var.storage_class == desc::StorageClass::Input) {
+                    let attrib_name = var.name.clone();
+                    let slot = var.decoration.iter().filter_map(|dec| match *dec {
+                                    instruction::Decoration::Location(slot) => Some(slot),
+                                    _ => None,
+                                }).next().expect("Missing location decoration");
```
kvark: If you are doing `filter_map(..).next().expect(..)`, isn't it the same as just doing `find(..).expect(..)`?

kvark:
The PR looks great! I got a few nitpicks but don't mind merging now if you are unable to address them. Let me know.

For texture+sampler, it would be reasonable to follow DX11 path for consistency (like you did).
For the error propagation, I'm not as much worried about `unwrap()` calls as I am about the places where we simply ignore errors (like, `if let Some(xx) = yy { do_zzz(); } // else { ignore() }`).

src/backend/vulkan/src/mirror.rs
msiglreith: Assuming our module was compiled via glslang it shouldn't be possible. SPIR-V allows to specify any name for entrypoints (not only `main`). If there is no `main()` the code will probably fail somewhere else I assume during the pipeline creation as we are also asking for the main function there.
https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkan/src/factory.rs#L241

src/backend/vulkan/src/mirror.rs
msiglreith: Indeed, I'm trying to change it!

msiglreith:
Thanks for reviewing! I'll try to take care of the points you mentioned in the comments and further add some warnings/errors for currently unhandled cases to avoid confusion in the future.

src/backend/vulkan/src/mirror.rs
kvark: Sounds like it should be using 'expect()' then.

src/backend/vulkan/src/mirror.rs
kvark: I see. It's a minor thing, but I think `find` expresses your intent better.

kvark:
Thanks! Let me know if/when ready.

src/backend/vulkan/src/mirror.rs
msiglreith: I came up with too solutions, but unsure which one is better:
``` Rust
let slot = var.decoration.iter()
                         .map(|dec| if let instruction::Decoration::Location(slot) = *dec { Some(slot) } else { None })
                         .find(|slot| slot.is_some())
                         .expect("Missing location decoration").unwrap();
```
``` Rust
let slot = var.decoration.iter()
                         .find(|dec| if let &instruction::Decoration::Location(..) = *dec { true } else { false })
                         .map(|dec| if let instruction::Decoration::Location(slot) = *dec { Some(slot) } else { None })
                         .expect("Missing location decoration").unwrap();
```
The latter one expresses the `find_map(..)` approach more nicely but is more verbose. Any preferences?

kvark: Ah, I see the problem now. Let's proceed with the least verbose one.

msiglreith:
Should be ready now.

kvark:
Awesome, thank you!
@homu r+

PR #1046, opened 2016-09-11
Fix shadow example on DX backend

fkaa:
Amends the differences between how GL & D3D expects depth values + fixes outdated variable names and missing indices for plane geometry.


kvark:
Thanks!
@homu r+

PR #1047, opened 2016-09-11
Document normalized coordinates

### LINKED ISSUE #1044: Document the differences in normalized vertex positions in the backends
From the chat:
> GL expects Z output in [-1,1], and that's what our math produces (cgmath catering to GL)
DX expects Z in [0,1] (while still treating X and Y in [-1,1] like GL does), so we gotta patch it
note that Z we are patching is in the homogeneous coordinates, so we use W there, and the end result (Z/W) goes into [0,1]

Thus, for DX we need to replace this:
```hlsl
return mul(u_Transform, float4(pos));
```
by this:
```hlsl
float4 ndc = mul(u_Transform, float4(pos));
ndc.z = (ndc.z + ndc.w) * 0.5;
return ndc;
```

Related: #942

Bastacyclop:
Fixes #1044.
Please check those coordinates before merging.


msiglreith:
Vulkan:
* Vulkan 15.1.1: Normalized texel coordinates: The s coordinate goes from 0.0 to 1.0, left to right. The t coordinate goes from 0.0 to 1.0, top to bottom.
* Vulkan A.3: fragment entry points must declare `OriginUpperLeft`.
* SPIR-V (`OriginUpperLeft`): Pixel coordinates appear to originate in the upper left, and increase toward the right and downward.
* Vulkan 23.2: In clip coordinates, the view volume is defined by: `0 <= z <= w`

msiglreith:
Metal:
https://developer.apple.com/library/ios/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/Render-Ctx/Render-Ctx.html#//apple_ref/doc/uid/TP40014221-CH7-SW34

LGTM

Bastacyclop:
Oh thanks I did look at this page but expected some kind of graphical element and missed this text xD.

kvark:
@homu r+

Issue #1048, opened 2016-09-13
Checked out master on OSX 10.11.6 with Rust 1.11.0 unable to compile cube or triangle example


jamiltron:
Hello,

I just cloned gfx-rs on my OSX 10.11.6 machine, tried running the examples, and ran into a number of errors.

Here's the steps I used to arrive at the errors:
```
git clone https://github.com/gfx-rs/gfx.git
cd gfx
cargo run --example cube
````

The errors I'm seeing look related to the metal backend:
```
src/backend/metal/src/command.rs:25:5: 25:27 error: unresolved import `gfx_core::shade::Stage`. Maybe a missing `extern crate gfx_core`? [E0432]
src/backend/metal/src/command.rs:25 use gfx_core::shade::Stage;
                                        ^~~~~~~~~~~~~~~~~~~~~~
src/backend/metal/src/command.rs:25:5: 25:27 help: run `rustc --explain E0432` to see a detailed explanation
src/backend/metal/src/map.rs:443:9: 443:38 error: unresolved import `ore::shade::IsMultiSample::*`. Maybe a missing `extern crate ore`? [E0432]
src/backend/metal/src/map.rs:443     use ore::shade::IsMultiSample::*;
                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/backend/metal/src/map.rs:443:9: 443:38 help: run `rustc --explain E0432` to see a detailed explanation
src/backend/metal/src/factory.rs:40:6: 40:24 error: `core::mapping::Raw` is not a trait [E0404]
src/backend/metal/src/factory.rs:40 impl core::mapping::Raw for RawMapping {
                                         ^~~~~~~~~~~~~~~~~~
src/backend/metal/src/factory.rs:40:6: 40:24 help: run `rustc --explain E0404` to see a detailed explanation
src/backend/metal/src/factory.rs:77:44: 77:63 error: type name `factory::BufferInfo` is undefined or not in scope [E0412]
src/backend/metal/src/factory.rs:77     fn create_buffer_internal(&self, info: factory::BufferInfo, raw_data: Option<*const c_void>)
                                                                               ^~~~~~~~~~~~~~~~~~~
src/backend/metal/src/factory.rs:77:44: 77:63 help: run `rustc --explain E0412` to see a detailed explanation
src/backend/metal/src/factory.rs:77:44: 77:63 help: no candidates by the name of `BufferInfo` found in your project; maybe you misspelled the name or forgot to import an external crate?
[...]

Issue #1014, opened 2016-08-12
Enable CI for Metal backend


kvark:
We've got quite a few backends now. DX11 is tested by AppVeyor, GL and Vulkan are tested by Travis. Metal seems to be rotting a bit, but before fixing it we need to establish the continuous integration to keep it up.
cc @fkaa

kvark:
Travis can do OSX: https://docs.travis-ci.com/user/multi-os/

kvark:
We should make sure to not use `--features vulkan` on OSX too.

fkaa:
Fixed in #1049

PR #1049, opened 2016-09-14
Getting the Metal target to run, adding OSX as a Travis CI target.

jamiltron:
I was able to get the Metal target to both run, and added a Travis CI target to try to keep it steady.

This should fix issue #1048 and issue #1014.

Sorry about my formatting - I realized my editor runs `rustfmt` on save, but it doesn't look like this entire code base is formatted this way, so these commits appear larger than they are. Do you have any kind of editor settings I could use to reformat this if this is a problem?

I know its kind of not a good idea to fix two things in a single pull request, but since it seemed like the Metal target rotted due to not having CI, I thought it might be best to add them in together. That way if I have to change the Metal code its still ran under CI, but also so that my pull request doesn't immediately cause the build to fail, if that stat is important.

I hope this helps out, and I would love pointers, tips, criticisms on things I could improve.

Thank you!

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -21,8 +21,8 @@ extern crate cocoa;
 extern crate gfx_core as core;
 extern crate metal;
 
-//use cocoa::base::{selector, class};
-//use cocoa::foundation::{NSUInteger};
+// use cocoa::base::{selector, class};
+// use cocoa::foundation::{NSUInteger};
 
 use metal::*;
 
@@ -31,7 +31,7 @@ use core::memory::{self, Usage};
 
 use std::cell::RefCell;
 use std::sync::Arc;
-//use std::{mem, ptr};
+// use std::{mem, ptr};
 
 mod factory;
 mod command;
@@ -90,7 +90,7 @@ unsafe impl Sync for InputLayout {}
 
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Shader {
-    func: MTLFunction
+    func: MTLFunction,
 }
 unsafe impl Send for Shader {}
 unsafe impl Sync for Shader {}
@@ -98,7 +98,7 @@ unsafe impl Sync for Shader {}
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Program {
     vs: MTLFunction,
-    ps: MTLFunction
+    ps: MTLFunction,
 }
 unsafe impl Send for Program {}
 unsafe impl Sync for Program {}
@@ -109,7 +109,7 @@ pub struct Pipeline {
     depth_stencil: Option<MTLDepthStencilState>,
     winding: MTLWinding,
     cull: MTLCullMode,
-    fill: MTLTriangleFillMode
+    fill: MTLTriangleFillMode,
 }
 unsafe impl Send for Pipeline {}
 unsafe impl Sync for Pipeline {}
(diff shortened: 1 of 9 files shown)


src/backend/metal/src/lib.rs
```diff
@@ -176,50 +186,83 @@ impl core::Device for Device {
...
+                                                   }, // DSV
+                                                   |_, _v| {
+                                                       // v.sampler.release();
+                                                   }, // sampler
+                                                   |_, _| {
+                                                       // fence
+                                                   },
+                                                   |_, _| {
+                                                       // raw mapping
+                                                   });
```
Bastacyclop: This rustfmt formatting seems awful to me xD

src/backend/metal/src/map.rs
```diff
@@ -440,77 +479,78 @@ pub fn map_base_type(ty: MTLDataType) -> shade::BaseType {
...
 
 pub fn map_texture_usage(usage: Usage) -> (MTLResourceOptions, MTLStorageMode) {
     match usage {
         Usage::GpuOnly => (MTLResourceStorageModePrivate, MTLStorageMode::Private),
-
-        Usage::Const   => (MTLResourceStorageModePrivate, MTLStorageMode::Managed),
+        Usage::Immutable => (MTLResourceStorageModePrivate, MTLStorageMode::Managed),
         Usage::Dynamic => (MTLResourceCPUCacheModeDefaultCache, MTLStorageMode::Managed),
         Usage::CpuOnly(access) => (map_access(access), MTLStorageMode::Managed),
+        Usage::Persistent(access) => (map_access(access), MTLStorageMode::Managed),
```
Bastacyclop: Is this what we should return here ? (I have no idea)

src/backend/metal/src/map.rs
```diff
@@ -440,77 +479,78 @@ pub fn map_base_type(ty: MTLDataType) -> shade::BaseType {
...
+        _ => unreachable!(),
     }
 }
 
 pub fn map_texture_usage(usage: Usage) -> (MTLResourceOptions, MTLStorageMode) {
     match usage {
         Usage::GpuOnly => (MTLResourceStorageModePrivate, MTLStorageMode::Private),
-
-        Usage::Const   => (MTLResourceStorageModePrivate, MTLStorageMode::Managed),
+        Usage::Immutable => (MTLResourceStorageModePrivate, MTLStorageMode::Managed),
```
Bastacyclop: I think this should be [`MTLStorageMode::Private`](https://developer.apple.com/reference/metal/mtlstoragemode/1515797-private). Though you didn't change anything here so `Managed` might be necessary to initialize the immutable memory.

src/backend/metal/src/map.rs
fkaa: IIRC it might be just copy-pasted and I haven't gotten around to changing it yet, but fixing it shouldn't be part of this PR though

fkaa:
I don't think the rustfmt should be too much of a problem (maybe we should do it over the whole codebase every now and then?).

Looks good to me, merging! 👍 
@homu r+

kvark:
Oh wow, this is awesome. Thanks @jamiltron !

src/backend/metal/src/lib.rs
jamiltron: Yeah there's some choices in rustfmt that I am kind of grossed out by, but personally I kind of think just agreeing on one standard kind of makes sense, even if you have to make some compromises on taste :P

src/backend/metal/src/map.rs
jamiltron: Yeah I wanted to keep most of the pre-existing code as is. I think you're right though so I'll check the documentation later and if so I'll give it a spin. Thanks!

src/backend/metal/src/map.rs
jamiltron: Hmm, I guess this depends on the target (iOS vs OSX) unless I'm mistaken ? Right now gfx doesn't support iOS, right? In that case I don't think shared is the right storage choice, given this statement in the documentation:

> In macOS, the shared storage mode is only available for MTLBuffer objects, not MTLTexture objects.

But then also it looks like for iOS and tvOS we'll have to change the storage, as [MTLStorageMode::Managed](https://developer.apple.com/reference/metal/mtlstoragemode/1515488-managed) is not available:

>  In iOS and tvOS, the managed storage mode is not available.

I can investigate further unless you were already working on this @fkaa

Thanks for the code review!

jamiltron:
No problem, happy to help! Thanks for the awesome project!

src/backend/metal/src/map.rs
fkaa: It probably won't run on iOS and I don't have any iOS system to test it on unfortunately (simulator doesn't work either).

I have a bunch of changes locally that I'm going to push today or tomorrow which will make the metal backend actually useable, so I don't think you'd be able to even debug it in its current shape. I haven't touched the storage modes there though, so feel free to pick it up as soon as I hurry up and push the changes! 👍

PR #1050, opened 2016-09-17
Metal fixes and additions

fkaa:
Lots of changes to the metal backend to make it actually useable:
* Memory leaks reduced by a lot (no more kernel panics! 🎉 )
* Now available as a feature rather than being awkwardly squished in between getting commented out and being default for OSX
* Lots more missing stuff implemented (Still missing buffer mapping, I have some ideas here though)
* Moved the metal bindings to crates.io, so we wont break previous published versions

Examples that were added:
* Instancing
* Deferred
* Shadow (*fixed*)

A few things that were changed to make room for Metal:
* Deferred example used two `[i8; 2]` types in one of the vertex attribute macros. Metal sees this as two elements with a stride of 2, and a minimum stride of 4 is required, so it got changed to a `[i8; 4]`. **NOTE**: Changed other shaders but haven't compiled DX
* The `Encoder::draw(...)` function previously bound the pipeline state & vertex buffers before render targets, which Metal requires before everything else. Changed so that order is now `Render targets -> PSO -> Rest`. Tested on GL and nothing was broken, haven't checked DX11, unsure about Vulkan.

Let me know if I missed something

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,8 +9,9 @@ authors = ["The Gfx-rs Developers"]
 documentation = "https://docs.rs/gfx"
 
 [features]
-default = []
+default = ["metal"]
 vulkan = ["gfx_device_vulkan", "gfx_window_vulkan"]
+metal = ["gfx_device_metal", "gfx_window_metal"]
 unstable = []
 
 
@@ -36,6 +37,16 @@ path = "src/window/vulkan"
 version = "0.1"
 optional = true
 
+[dependencies.gfx_device_metal]
+path = "src/backend/metal"
+version = "0.1"
+optional = true
+
+[dependencies.gfx_window_metal]
+path = "src/window/metal"
+version = "0.1"
+optional = true
+
 [target.'cfg(unix)'.dependencies]
 gfx_window_glfw = { path = "src/window/glfw", version = "0.12" }
 gfx_window_sdl = { path = "src/window/sdl", version = "0.4" }
@@ -43,10 +54,6 @@ gfx_window_sdl = { path = "src/window/sdl", version = "0.4" }
 [target.'cfg(windows)'.dependencies]
 gfx_device_dx11 = { path = "src/backend/dx11", version = "0.4" }
 gfx_window_dxgi = { path = "src/window/dxgi", version = "0.4" }
-
-[target.x86_64-apple-darwin.dependencies]
-gfx_device_metal = { path = "src/backend/metal", version = "0.1" }
-gfx_window_metal = { path = "src/window/metal", version = "0.1" }
 
 [[example]]
 name = "blend"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ authors = ["The Gfx-rs Developers"]
 documentation = "https://docs.rs/gfx"
 
 [features]
-default = ["metal"]
+default = []
 vulkan = ["gfx_device_vulkan", "gfx_window_vulkan"]
 metal = ["gfx_device_metal", "gfx_window_metal"]
 unstable = []
--- a/examples/deferred/main.rs
+++ b/examples/deferred/main.rs
@@ -92,7 +92,6 @@ gfx_defines!{
     constant CubeLocals {
         transform: [[f32; 4]; 4] = "u_Transform",
         radius: f32 = "u_Radius",
-        _padding: [f32; 3] = "",
     }
 
     pipeline light {
@@ -518,7 +517,6 @@ impl<R: gfx::Resources> gfx_app::Application<R> for App<R> {
         let mut cube_locals = CubeLocals {
             transform: (proj * view.mat).into(),
             radius: LIGHT_RADIUS,
-            _padding: [0.0; 3]
         };
         encoder.update_constant_buffer(&self.light.data.locals_vs, &cube_locals);
         cube_locals.radius = EMITTER_RADIUS;
--- a/examples/deferred/main.rs
+++ b/examples/deferred/main.rs
@@ -39,7 +39,6 @@ extern crate noise;
 use rand::Rng;
 use cgmath::{SquareMatrix, Matrix4, Point3, Vector3, EuclideanVector, deg};
 use cgmath::{Transform, AffineMatrix3};
-pub use gfx::format::Depth;
 pub use gfx_app::ColorFormat;
 use gfx::{Bundle, texture};
 use genmesh::{Vertices, Triangulate};
@@ -47,6 +46,11 @@ use genmesh::generators::{SharedVertex, IndexedPolygon};
 use std::time::{Instant};
 
 use noise::{Seed, perlin2};
+
+#[cfg(feature="metal")]
+pub use gfx::format::Depth32F as Depth;
+#[cfg(not(feature="metal"))]
+pub use gfx::format::Depth;
 
 // Remember to also change the constants in the shaders
 const NUM_LIGHTS: usize = 250;
@@ -68,8 +72,7 @@ gfx_defines!{
     }
 
     vertex BlitVertex {
-        pos: [i8; 2] = "a_Pos",
-        tex_coord: [i8; 2] = "a_TexCoord",
+        pos_tex: [i8; 4] = "a_PosTexCoord",
     }
 
     vertex CubeVertex {
@@ -89,6 +92,7 @@ gfx_defines!{
     constant CubeLocals {
         transform: [[f32; 4]; 4] = "u_Transform",
         radius: f32 = "u_Radius",
+        _padding: [f32; 3] = "",
     }
 
     pipeline light {
@@ -171,7 +175,11 @@ struct ViewPair<R: gfx::Resources, T: gfx::format::Formatted> {
 // need a custom depth format in order to view SRV depth as float4
 struct DepthFormat;
 impl gfx::format::Formatted for DepthFormat {
+    #[cfg(feature="metal")]
+    type Surface = gfx::format::D32;
+    #[cfg(not(feature="metal"))]
     type Surface = gfx::format::D24;
+
     type Channel = gfx::format::Unorm;
     type View = [f32; 4];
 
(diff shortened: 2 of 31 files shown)


examples/deferred/main.rs
```diff
@@ -48,6 +47,11 @@ use std::time::{Instant};
 
 use noise::{Seed, perlin2};
 
+#[cfg(feature="metal")]
+pub use gfx::format::Depth32F as Depth;
```
kvark: we could probably use it for all the backends, at least for the examples that require anything higher than GL 2.1

src/backend/metal/Cargo.toml
```diff
@@ -28,8 +28,9 @@ name = "gfx_device_metal"
 [dependencies]
 log = "0.3"
 gfx_core = { path = "../../core", version = "0.5" }
-cocoa = "0.3"
+cocoa = "0.2"
```
kvark: oh, why is that?

src/backend/metal/src/command.rs
```diff
@@ -14,416 +14,112 @@
 
 #![allow(missing_docs)]
 
-// use cocoa::foundation::NSRange;
+//use cocoa::foundation::NSRange;
 
-use core::{command, pso, shade, state, target, texture as tex};
+use core::{pso, shade, state, target, texture, command};
```
kvark: command should probably go first since this list is in the alphabetical order, but that's minor

src/backend/metal/src/command.rs
```diff
@@ -14,416 +14,112 @@
...
-        for i in 0..1 {
+        for i in 0..MAX_VERTEX_ATTRIBUTES {
             if let Some((buffer, offset)) = vbs.0[i] {
-                buffers[i] = buffer.0;
-                offsets[i] = offset as u64;
-                indices[i] = i as u64;
+                // TODO(fkaa): is it okay if we cascade downwards with vertex
+                //             buffers? don't want to collide with user defined
+                //             values on accident!
+                self.encoder.set_vertex_buffer(30 - i as u64, offset as u64, unsafe { *(buffer.0).0 });
```
kvark: 30 is a magic constant?

src/backend/metal/src/command.rs
```diff
@@ -14,416 +14,112 @@
...
-            let mut count = 0;
-            for cbuf in cbs.iter() {
-                if cbuf.1.contains(mask) {
-                    buffers[cbuf.2 as usize] = (cbuf.0).0;
-                    count += 1;
+            for cb in cbs.iter() {
+                if cb.1.contains(mask) {
+                    if let Stage::Vertex = stage {
+                        self.encoder.set_vertex_buffer(cb.2 as u64, 0, unsafe { *((cb.0).0).0 });
+                    } else {
```
kvark: just to clarify: no geometry shaders, right? might be better to use a `match` here with some `unimplemented!` paths

src/backend/metal/src/command.rs
```diff
@@ -433,18 +129,16 @@ impl command::Buffer<Resources> for CommandBuffer {
...
     fn bind_resource_views(&mut self, rvs: &[pso::ResourceViewParam<Resources>]) {
         for &stage in [Stage::Vertex, Stage::Pixel].iter() {
-            let mut views = [native::Srv(ptr::null_mut()); MAX_RESOURCE_VIEWS];
             let mask = stage.into();
-            let mut count = 0;
             for view in rvs.iter() {
                 if view.1.contains(mask) {
-                    views[view.2 as usize] = view.0;
-                    count += 1;
+                    if let Stage::Vertex = stage {
```
kvark: same - use `match`

Cargo.toml
```diff
@@ -44,10 +55,6 @@ gfx_window_sdl = { path = "src/window/sdl", version = "0.4" }
 gfx_device_dx11 = { path = "src/backend/dx11", version = "0.4" }
```
kvark: I wonder if we should feature gate DX11 the same way for consistency
the fact that DX11 is used by default doesn't help the users too much anyway

src/backend/metal/src/command.rs
```diff
@@ -453,176 +147,239 @@ impl command::Buffer<Resources> for CommandBuffer {
...
-            let mut samplers = [native::Sampler(MTLSamplerState::nil()); MAX_SAMPLERS];
             let mask = stage.into();
-            let mut count = 0;
-            for sm in ss.iter() {
-                if sm.1.contains(mask) {
-                    samplers[sm.2 as usize] = sm.0;
-                    count += 1;
+            for sampler in ss.iter() {
+                if sampler.1.contains(mask) {
+                    // TODO(fkaa): what about min/max lods?
```
kvark: We definitely [pass those](https://github.com/gfx-rs/gfx/blob/a015a3a6e0dac56d877e54a5427d28714ca30e72/src/core/src/texture.rs#L424) in `SamplerInfo`, so we'll need to use them here. TODO

src/backend/metal/src/command.rs
```diff
@@ -453,176 +147,239 @@ impl command::Buffer<Resources> for CommandBuffer {
...
     fn set_scissor(&mut self, rect: target::Rect) {
-        self.buf.push(Command::SetScissor(MTLScissorRect {
+        // TODO(fkaa): why are getting 1x1 scissor?
+        /*self.encoder.set_scissor_rect(MTLScissorRect {
             x: rect.x as u64,
             y: rect.y as u64,
             width: (rect.x + rect.w) as u64,
             height: (rect.y + rect.h) as u64,
-        }));
+        });*/
```
kvark: should we put `unimplemented` here?

src/backend/metal/src/command.rs
```diff
@@ -453,176 +147,239 @@ impl command::Buffer<Resources> for CommandBuffer {
...
+                *(buf.0).0 = self.device.new_buffer_with_data(
+                    data.as_ptr() as _,
+                    data.len() as _,
+                    map_buffer_usage(buf.1));
+
+                // invalidate old buffer in cache
+                self.encoder.invalidate_buffer(b);
+            }
+        } else {
+            // TODO(fkaa): how do we (eventually) handle updating buffers when
```
kvark: good question. We need a solid synchronization model for resources (buffers and textures)

src/backend/metal/src/command.rs
```diff
@@ -154,21 +91,22 @@ impl command::Buffer<Resources> for CommandBuffer {
...
-        //self.buf.push(Command::BindPipeline(pso));
     }
 
     fn bind_vertex_buffers(&mut self, vbs: pso::VertexBufferSet<Resources>) {
         for i in 0..MAX_VERTEX_ATTRIBUTES {
             if let Some((buffer, offset)) = vbs.0[i] {
+                // TODO(fkaa): is it okay if we cascade downwards with vertex
+                //             buffers? don't want to collide with user defined
+                //             values on accident!
                 self.encoder.set_vertex_buffer(30 - i as u64, offset as u64, unsafe { *(buffer.0).0 });
```
jamiltron: Why 30?

src/backend/metal/src/command.rs
fkaa: Right, so I was constantly getting a 1x1 scissor rect and I'm not sure why (default value?) So I just temporarily commented it out since nothing was using scissors from what I could tell

src/backend/metal/src/command.rs
fkaa: I think it complains at PSO creation if you try anything other than VS + PS, but matching is probably the better option nonetheless

src/backend/metal/src/command.rs
fkaa: Forgot to use `MTL_MAX_BUFFER_BINDINGS` (31). Otherwise is it OK that vertex buffers (not mentioned in shaders, only PSO macro) are automatically assigned from top to bottom? would be unpleasant if you bind your own buffers at say 0 & 1 and it collides with another buffer. Should still warn/error at runtime though.

src/backend/metal/Cargo.toml
fkaa: There was some version mismatch between metal bindings and the one used in the gfx backend, didn't feel like pushing a new version so I took the lazy way out

Cargo.toml
fkaa: Do we have an issue open on cargo to allow for default features depending on target? Useful for users

src/backend/metal/src/command.rs
Bastacyclop: The scissor [data](https://github.com/gfx-rs/gfx/blob/a015a3a6e0dac56d877e54a5427d28714ca30e72/src/render/src/pso/mod.rs#L72) is initialized and cleared as a 1x1 rect. So if there is no `Scissor` component in the PSO to override this, you get this 1x1 value ? I don't understand why the other backends are fine with this though. Maybe the scissor rect is set to 1x1 but the scissor test is not activated ?

Cargo.toml
kvark: Agreed, I don't think we have...

src/backend/metal/src/command.rs
kvark: Yes, in other backends there is an explicit switch to turn it ON/OFF. If Metal doesn't have the switch, you need to cache that flag and then use the full-sized scissor if it means to be OFF.

src/backend/metal/src/command.rs
kvark: You mean, assign to 0 and 1 from the shader side? I thought this code should be the only thing in control of what vertex buffers are assigned to the slots.

fkaa: Well vertex buffers share the same slot-space as constant buffers etc., so they would overlap if they were assigned from 0 up and in shader another buffer was bound to slot 0 as well. How does DX/HLSL handle this? (assuming it has something similar to `in` attributes in GLSL)

src/backend/metal/src/command.rs
fkaa: Gotcha, full-sized = size of the viewport?

kvark: Yep. Or put the maximum value, i.e. (0, 32768, 0, 32768) or something

src/backend/metal/src/command.rs
kvark: In DX, constant buffers are assigned to different set of registers. Are you sure about Metal stuff? Can you drop a link with some info to read up?

fkaa: [Vertex attributes and [[ stage_in ]]](https://developer.apple.com/library/content/documentation/Metal/Reference/MetalShadingLanguageGuide/func-var-qual/func-var-qual.html#//apple_ref/doc/uid/TP40014364-CH4-SW11). Basically the `[[ attribute(n) ]]` are assigned to a buffer slot [here](https://github.com/fkaa/gfx/blob/151ed922cbca6d3763ed3ce346a5bd29ed0a1f4b/src/backend/metal/src/factory.rs#L331-L369), and [this](https://developer.apple.com/reference/metal/mtlvertexattributedescriptor/1515502-bufferindex?language=objc) mentions that the vertex attributes point to a buffer in the argument table, which looks like this:

kvark: So it sounds like we should track the buffers assigned to constant buffers and avoid those indices when binding the vertex attributes. (doesn't have to be a part of this PR)

kvark:
@homu r+

PR #1055, opened 2016-09-24
[GL] bind output "Target{n}" to n if  isn't available

fkaa:
Where `n` is in the range of `0..core::MAX_COLOR_TARGETS` (4). Done before linking, safe [even if outputs are not present](https://www.opengl.org/wiki/Fragment_Shader#Output_buffers):
> Note that it is perfectly legal to assign names to fragment colors that are not mentioned in the fragment shader. The linking process will only use the names that are actually mentioned in the fragment shader. Because of that, it is also perfectly legal to assign multiple names to the same number; this is only an error if you attempt to link a program that uses both names.

### DIFF
--- a/src/backend/gl/src/shade.rs
+++ b/src/backend/gl/src/shade.rs
@@ -478,6 +478,15 @@ pub fn create_program(gl: &gl::Gl, caps: &c::Capabilities, private: &PrivateCaps
         unsafe { gl.AttachShader(name, sh) };
     }
 
+    if !private.program_interface_supported {
+        for i in 0..c::MAX_COLOR_TARGETS {
+            let color_name = format!("Target{}\0", i);
+            unsafe {
+                gl.BindFragDataLocation(name, i as u32, (&color_name[..]).as_ptr() as *mut gl::types::GLchar);
+            }
+         }
+    }
+
     unsafe { gl.LinkProgram(name) };
     info!("\tLinked program {}", name);
 


msiglreith:
Nice PR 👍 

How does this influence the way of specifying the PSO and writing shaders?
I assume for muliple render targets the names have to formatted like this `Target{N}`.
If so, we probably should document this to avoid confusion and enforce it in the PSO description.

fkaa:
It shouldn't break any current programs (that weren't already broken). I'd say this is a temporary solution until we find a better one. If we publish a new version before that It'd be worth a mention (this only affects those without `ARB_program_interface_query`). Hopefully having Target*n* won't be necessary.

Ideally we would use [GetFragDataLocation](http://docs.gl/gl3/glGetFragDataLocation) but there's the problem that program introspection needs to be done independent of any PSO information currently, and we need the name of the output for it to work.

Perhaps we could defer the lookup to when the PSO validation for a color output fails, which allows us to check with `GetFragDataLocation` and insert it into the program info if it exists. Even in the case when a program is shared between multiple PSOs and using different outputs it shouldn't affect one another, as only the PSO definitions are required to exist in the program info, not the other way around. Would need some hacks inside PSO macros though.

cc @kvark

kvark:
This is actually perfectly backwards-compatible. Thanks @fkaa !
@homu r+

PR #1056, opened 2016-09-24
Update sdl2 dependency in gfk_window_sdl

jplatte:
`cargo build` and `cargo clippy` still complete without warnings.


kvark:
Thanks!
@homu r+

PR #1060, opened 2016-10-05
Implement init_existing and init_existing_raw for gfx_window_glutin

AnthIste:
## Rationale

I would like to be able to combine Piston and gfx in a more modular fashion. Currently, there are 2 options to create and initialise a Glutin window with gfx:

1. Use PistonWindow<GlutinWindow>. This creates the window and initialises gfx. The version of gfx is the same one that ships with Piston (potentially outdated).
2. Use gfx_window_glutin. This creates the window and initialises gfx. There is no way to wrap this window in a way Piston understands ([GlutinWindow](https://docs.rs/pistoncore-glutin_window/0.30.0/glutin_window/struct.GlutinWindow.html))

Each library is taking responsibility of creating the window from a builder. I would like to split the responsibility of creating the window, from initialising gfx.

## Implementation

This is a simple overloaded API. Instead of using a passed builder to construct the glutin::Window directly, the window is passed as a parameter. This can be borrowed as no ownership is needed, and therefore it is not returned.

## Safety

What will happen if you try and initialise an existing window multiple times? The existing API creates the window so this is not possible.

## Versioning

I have bumped the minor version in Cargo.toml to 0.13.1. The change should be 100% backwards compatible.

### DIFF
--- a/src/window/glutin/Cargo.toml
+++ b/src/window/glutin/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_window_glutin"
-version = "0.13.0"
+version = "0.13.1"
 description = "Glutin window for gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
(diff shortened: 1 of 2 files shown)


kvark:
Thanks!
@homu r+

PR #1064, opened 2016-10-09
A few formatting fixes

jplatte:
The more important thing here is the first commit, which fixes the documentation formatting. Before:

bildschirmfoto am 2016-10-09 um 18 17 30

After:

bildschirmfoto am 2016-10-09 um 18 17 43


kvark:
Oh my, it looks sweet now, thanks!
@homu r+

PR #1068, opened 2016-10-22
Enable PROGRAM_POINT_SIZE in OpenGL backend

lidavidm:
This lets the vertex shader write to gl_PointSize, so that point primitives can be rendered as different sizes.

The flag is enabled during initialization of the OpenGL backend.


lidavidm:
Looks like the Vulkan builds fail?

sectopod:
There is some sort of a macro-related breakage on the nightly, although no idea why it only affected Linux build and not OSX.

kvark:
@lidavidm thanks for your PR! The breakage is issued in #1070, and we don't need to block on it here.

PR #1072, opened 2016-10-24
Add particle example using a geometry shader

Diggsey:


kvark:
That's impressive, thank you!
It would be really neat at some point (not necessarily now) to move the particle processing code to GPU. This would require either transform feedback, or compute shaders, or simply storing the particle data in textures (possible today). Could show millions of them ;)

PR #1073, opened 2016-10-27
Use winit to create dx11 windows

Diggsey:


src/window/dxgi/src/lib.rs
```diff
@@ -84,11 +73,17 @@ where Cf: format::RenderFormat
 /// Initialize with a given size. Raw format version.
 pub fn init_raw(title: &str, requested_width: winapi::INT, requested_height: winapi::INT, color_format: format::Format)
                 -> Result<(Window, Device, Factory, core::handle::RawRenderTargetView<Resources>), InitError> {
-    let hwnd = match window::create(title, requested_width, requested_height) {
-        Ok(h) => h,
-        Err(()) => return Err(InitError::Window),
+    let inner = match winit::WindowBuilder::new()
+        .with_title(title.to_owned())
+        .with_dimensions(requested_width as u32, requested_height as u32)
+        .with_min_dimensions(requested_width as u32, requested_height as u32)
```
kvark: I don't think we need to set the min/max dimensions here.

src/window/dxgi/src/lib.rs
```diff
@@ -19,20 +19,19 @@ extern crate log;
 extern crate kernel32;
```
kvark: please remove both `kernel32` and `user32` dependencies from here as no longer used

src/lib.rs
```diff
@@ -275,11 +275,20 @@ impl<A: ApplicationBase<gfx_device_dx11::Resources, D3D11CommandBuffer>> Applica
...
+                    winit::Event::Closed => break 'main,
+                    _ => {}
+                }
+            }
+            // draw a frame
             app.render(&mut device);
             window.swap_buffers(1);
             device.cleanup();
-            harness.bump();
+            harness.bump()
```
kvark: any reason for this line to be changed?

src/window/dxgi/src/lib.rs
Diggsey: We don't currently resize the backbuffer in response to the window being resized, so setting the min and max just prevents resizing altogether (which most closely matches the previous behaviour without winit)

kvark: hmm, I see the reasoning. We'll revise this later then, when the resize support is better.

kvark:
Thanks!

PR #1075, opened 2016-11-01
Fix pipeline definition macro

### LINKED ISSUE #1070: GFX is broken on nightly
https://travis-ci.org/ozkriff/zoc/jobs/169961437

``` plain
...
cargo build --release
   Compiling core v0.0.1 (file:///home/travis/build/ozkriff/zoc/core)
   Compiling glutin v0.6.1
   Compiling tempfile v2.1.4
   Compiling wayland-window v0.2.3
   Compiling gfx_window_glutin v0.12.0
   Compiling zoc v0.0.1 (file:///home/travis/build/ozkriff/zoc)
error[E0432]: unresolved import `$crate`
  --> src/pipeline.rs:8:1
   |
8  | gfx_defines! {
   | ^ no `$crate` in the root
   |
   = note: this error originates in a macro outside of the current crate
error[E0433]: failed to resolve. Use of undeclared type or module `gfx`
  --> src/pipeline.rs:15:22
   |
15 |         basic_color: gfx::Global<[f32; 4]> = "u_Basic_color",
   |                      ^^^^^^^^^^^^^^^^^^^^^ Use of undeclared type or module `gfx`
...
```

msiglreith:
`use $crate;` is not allowed due to recent restrictions and `use super::*;` won't import the gfx crate. (allowed in future but feature gated https://github.com/rust-lang/rfcs/pull/1560)

Closes #1070

### DIFF
--- a/src/render/src/macros/pso.rs
+++ b/src/render/src/macros/pso.rs
@@ -226,9 +226,9 @@ macro_rules! gfx_pipeline_base {
         $( $field:ident: $ty:ty, )*
     }) => {
         pub mod $module {
-            use $crate;
             #[allow(unused_imports)]
             use super::*;
+            use super::gfx;
             gfx_pipeline_inner!{ $(
                 $field: $ty,
             )*}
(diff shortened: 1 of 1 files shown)


kvark:
Thanks!
@homu r+

mitchmindtree:
I'm not sure that this PR results in the same behaviour?

My understanding is that the new `use super::gfx;` line assumes that either `extern crate gfx;` or `pub use gfx;` exists in the `super` module, while the original `use $crate;` does not. This PR would break any downstream crates that don't happen to invoke `gfx_pipeline!` in the same module that they import the gfx crate, right?

PR #1078, opened 2016-11-10
Add missing lifetime bound in DerefMut impl for RWer

jplatte:
This isn't caught by the current stable compiler, but to me it looks like it was always wrong (the impl has fewer bounds on T than the struct) and the current nightly compiler errors out without this change.


kvark:
Neat, thanks!
@homu r+

PR #1079, opened 2016-11-18
Gamma example added

thedarkduscher:
I've added an example based on the triangle example which should help to check the gamma settings.


kvark:
Thanks @thedarkduscher !
I'm a bit worried about the growing number of examples though. They have a lot of shared code, and become increasingly difficult to maintain given that our contributor number doesn't grow as quick.
@homu r+

PR #1080, opened 2016-11-20
Add attributes to gfx_defines macros

djmcgill:
For this commit, it only works on the `vertex` and `constant` structs, and NOT any of their fields.

### DIFF
--- a/src/render/src/macros/mod.rs
+++ b/src/render/src/macros/mod.rs
@@ -34,16 +34,16 @@ macro_rules! gfx_format {
 /// Defines vertex, constant and pipeline formats in one block
 #[macro_export]
 macro_rules! gfx_defines {
-    (vertex $name:ident {
+    ($(#[$attr:meta])* vertex $name:ident {
             $( $field:ident : $ty:ty = $e:expr, )+
     }) => {
-        gfx_vertex_struct!($name {$($field:$ty = $e,)+});
+        gfx_vertex_struct_meta!($(#[$attr])* vertex_struct_meta $name {$($field:$ty = $e,)+});
     };
 
-    (constant $name:ident {
+    ($(#[$attr:meta])* constant $name:ident {
             $( $field:ident : $ty:ty = $e:expr, )+
     }) => {
-        gfx_constant_struct!($name {$($field:$ty = $e,)+});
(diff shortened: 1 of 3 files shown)


djmcgill:
This is a workaround for the issue https://github.com/rust-lang/rust/issues/24189 which causes the character '#' to be considered a valid identifier character. Thus, the macro pattern
```
$(#[$attr:meta])* $name:ident
``` (which is greedy with no lookahead) fails because the parse is ambiguous. So whenever we want a pattern like this, there needs to be something separating the two tokens.

Lazy status uses commas: https://github.com/rust-lang-nursery/lazy-static.rs/blob/v0.2.1/src/lib.rs
Bitflags uses access modifiers: https://github.com/rust-lang-nursery/bitflags/blob/master/src/lib.rs#L183

This commit uses the "keywords" `vertex` and/or `constant`.

File: src/render/src/macros/structure.rs [hunk 1]
```diff
@@ -18,8 +18,20 @@
...
         $( $field:ident: $ty:ty = $name:expr, )*
+    }) => (gfx_impl_struct_meta! {
+        impl_struct_meta $runtime_format : $compile_format = $root {
+            $( $field : $ty = $name, )*
+        }
+    })
+}
+
+#[macro_export]
+macro_rules! gfx_impl_struct_meta {
```
kvark: any reason you decided to not just put the extra stuff into `gfx_impl_struct` instead of routing through the new `gfx_impl_struct_meta`?

src/render/src/macros/mod.rs
```diff
@@ -52,6 +52,29 @@ macro_rules! gfx_defines {
         gfx_pipeline!($name {$($field:$ty = $e,)+});
     };
 
+    // The recursive case for vertex structs
+    ($(#[$attr:meta])* vertex $name:ident {
```
kvark: is this ever going to hit, given that the same pattern is parsed above?

src/render/src/macros/structure.rs
djmcgill: I could probably have two patterns inside `gfx_impl_struct`, one with `$(#[$attr:meta])* impl_struct_meta` and one without, but then I believe I'd have to duplicate the whole body, which is worse to me than having an extra name.

kvark:
Thanks @djmcgill ! I hope you'll be around if/when somebody faces the macros bugs ;)
@homu r+

PR #1081, opened 2016-11-28
Remove gfx import in pipeline macro

msiglreith:
As RFC 1560 (https://github.com/rust-lang/rfcs/pull/1560) is now stabilized on nightly 1.14.0 `use super::*` will now import all occurences of `use gfx;` from the parent module. This allows to compile code like following:
``` Rust
#[macro_use]
extern crate gfx as gfx_temp;

mod foo {
    // Import `gfx` into this module.
    // The new pipeline module created in the macro has access
    // to this import via `use super::*;`
    use gfx_temp; 

    gfx_defines!{
        vertex Vertex {
            pos: [f32; 4] = "a_Pos",
            tex_coord: [f32; 2] = "a_TexCoord",
        }

        pipeline pipe {
            vbuf: gfx_temp::VertexBuffer<Vertex> = (),
        }
    }
}
```
So as long as the `gfx` module is imported in the module where you define the pipeline everything should work. Following this condition the regression https://github.com/gfx-rs/gfx/issues/1070 would be resolved.

### DIFF
--- a/src/render/src/macros/pso.rs
+++ b/src/render/src/macros/pso.rs
@@ -228,7 +228,6 @@ macro_rules! gfx_pipeline_base {
         pub mod $module {
             #[allow(unused_imports)]
             use super::*;
-            use super::gfx;
             gfx_pipeline_inner!{ $(
                 $field: $ty,
             )*}
@@ -244,7 +243,6 @@ macro_rules! gfx_pipeline {
         pub mod $module {
             #[allow(unused_imports)]
             use super::*;
-            use super::gfx;
             gfx_pipeline_inner!{ $(
                 $field: $ty,
             )*}


msiglreith:
Is there an option for conditional compilation depening on rustc version? Rust compiler <1.14.0 still require the `use $crates;` approach, while newer versions would make use of the `item_like_imports` feature.

kvark:
Hmm. Everything is great, except that it doesn't build on Linux stable (according to Travis)...

msiglreith:
Rust stable is still 1.13.0 which doesn't support `item_like_imports` and still requires the legacy `use $crates;` way. 1.14.0 will go out of beta around end of december.

sectopod:
Is this still needed? I was under impression that gfx now compiles fine on nightly/stable

msiglreith:
I think the example above won't compile with the current master branch, but I can double-check this!
1.14 is scheduled to become stable at 22nd december, then this PR should also build without problems on stable.

kvark:
@msiglreith ok, so you are saying we should be safe to merge this for 1.14? We'd also need to update the appveyor to use the new version.

msiglreith:
@kvark Yes, but I'll try to check other projects (e.g zoc) if this might cause some issues.

msiglreith:
Hm, looks like the `item_like_imports` feature didn't make it into the stable release. Closing this for the moment and create a new PR, when it finally becomes active in the stable channel.

PR #1086, opened 2016-12-08
[GL] Make sure VAO is binded before calling VertexAttribPointer

adahn:
## Rationale:
gfx_device_gl uses single VAO, which is binded once during device initialization. This causes problems when gfx is used with other libraries and middlewares, which can use their own VAO.

## Changes:
This PR adds a command which binds VAO owned by gfx, before binding VBO and calling VertexAttribPointer.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -109,8 +109,9 @@ pub const COLOR_DEFAULT: s::Color = s::Color {
     blend: None,
 };
 
-pub const RESET: [Command; 13] = [
+pub const RESET: [Command; 14] = [
     Command::BindProgram(0),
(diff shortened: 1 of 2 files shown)


kvark:
If you want to make sure the VAO is bound, the best way would be to bind it on start of the command buffer: https://github.com/gfx-rs/gfx/blob/f1c0f20d9fc76d6f934e2efb2b034efe585f3884/src/backend/gl/src/command.rs#L111

adahn:
I've put it inside bind_vertex_buffers function, because it is connected with vertex buffer commands, but making it one of RESET commands is also ok for me. I'll make new commit.

sectopod:
Thanks!

PR #1087, opened 2016-12-14
Basic Tessellation

sectopod:
Derived from #1027
Only tested on DX11 so far. Feel free to hack on GL side!
Note: the actual screenshot doesn't match the visuals. The example is configured for wireframe in order to illustrate the generated polys.

cc @ahmedaliadeel


kvark:
@homu r+
somebody will need to fix GL afterwards

Issue #1098, opened 2016-12-21
Persistent mapping on D3D11


kvark:
D3D11 doesn't appear to support it. cc @Bastacyclop

Bastacyclop:
@kvark Do you mean our backend, that would be expected since nobody did it yet (#1020), or D3D11 itself ? If D3D11 only supports temporary mapping we could do the same thing I did when the OpenGL capabilities do not include persistent mapping. Instead of synchronising the accesses to the persistent mapping we simply map/unmap the temporary mapping (the user API doesn't change). The question might be: is this solution good enough ?

Bastacyclop:
To be more precise, whenever the user wants to access the mapping we ensure that it is mapped by checking a boolean and calling `Map` if necessary. Whenever the GPU needs access to the mapping we ensure that it is unmapped by checking the same boolean and calling `Unmap` if necessary.

Bastacyclop:
This makes me consider entirely moving the mapping `Option<Fence>` management from core to the backends.

snuk182:
Does this mean we should mimic the Persistent Buffers in DX11 with, say, [this](https://gamedev.stackexchange.com/questions/15651/mapping-a-vertex-buffer-in-directx11)?

Bastacyclop:
That's what I would do, yes (without looking into the details of your link). We may also want to avoid explicitly saying to the user that the mapping is "persistent" since it might fall back to temporary mapping and that could be confusing.

snuk182:
Then would not it be more clear to name it not Persistent, but Mappable ?

kvark:
Sounds good! looks like you are on the right track :)

snuk182:
Crap. The [buffer mapping api](https://retep998.github.io/doc/winapi/d3d11/struct.ID3D11DeviceContext.html#method.Map) attached to the DeviceContext, which is a part of Device. How can I get it in a Factory?

Bastacyclop:
@snuk182 I hope we have not been working on the same parts :p

snuk182:
Please go ahead. I never worked with D3D so just playing around.

Bastacyclop:
I simply made the PR above, I never worked with D3D either and I won't have access to a Windows environment for a while.

snuk182:
Ok then I gonna torture my copy of gfx a bit.

snuk182:
[...]

PR #1099, opened 2016-12-22
Prepare mapping API for D3D11

Bastacyclop:
Following #1098.
@snuk182 @kvark

### DIFF
--- a/src/backend/gl/src/factory.rs
+++ b/src/backend/gl/src/factory.rs
@@ -129,7 +129,7 @@ impl Factory {
             let usage = match info.usage {
                 GpuOnly | Immutable => 0,
                 Dynamic => gl::DYNAMIC_STORAGE_BIT,
-                Persistent(access) => access_to_map_bits(access) | gl::MAP_PERSISTENT_BIT,
+                Mappable(access) => access_to_map_bits(access) | gl::MAP_PERSISTENT_BIT,
                 CpuOnly(_) => gl::DYNAMIC_STORAGE_BIT,
             };
             unsafe {
@@ -146,7 +146,7 @@ impl Factory {
                 GpuOnly => gl::STATIC_DRAW,
                 Immutable => gl::STATIC_DRAW,
                 Dynamic => gl::STREAM_DRAW,
-                Persistent(access) => match access {
+                Mappable(access) => match access {
                     memory::RW => gl::DYNAMIC_COPY,
                     memory::READ => gl::DYNAMIC_READ,
                     memory::WRITE => gl::DYNAMIC_DRAW,
--- a/src/backend/gl/src/factory.rs
+++ b/src/backend/gl/src/factory.rs
@@ -216,13 +216,15 @@ pub enum MappingKind {
     Temporary,
 }
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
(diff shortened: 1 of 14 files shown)


src/backend/gl/src/factory.rs
```diff
@@ -216,13 +216,15 @@ pub enum MappingKind {
...
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Debug, Eq, Hash, PartialEq)]
 #[allow(missing_copy_implementations)]
 pub struct MappingGate {
     pub kind: MappingKind,
     pub pointer: *mut ::std::os::raw::c_void,
     pub target: gl::types::GLenum,
-    pub is_mapped: bool,
+    // TODO: put those inside the kind enum or something ?
+    pub status: mapping::Status<R>, // Persistent Only
```
kvark: yes, please, let's use `enum`

src/core/src/lib.rs
```diff
@@ -182,7 +182,7 @@ pub trait Resources:          Clone + Hash + Debug + Eq + PartialEq + Any {
     type DepthStencilView:    Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync;
     type Sampler:             Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync + Copy;
     type Fence:               Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync;
-    type Mapping:             Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync + mapping::Gate<Self>;
+    type Mapping:             Hash + Debug + Eq + PartialEq + Any + Send + Sync + mapping::Gate<Self>;
```
kvark: I hope this does not prevent automatic deriving of `Clone` for all the handles parametrized by `R`

Bastacyclop: Well `R` is still `Clone` right, I don't think an associated type will change that ?

kvark: ok, good

sectopod:
Instancing example fails on D3D11 because of the `unimplemented!` arm in `map_usage`.
Even if I force it to be a dynamic buffer, D3D11 doesn't like it. There are no resources in D3D11 that are both readable and writable by the CPU, and accessible by the GPU. A dynamic resource is only writable on CPU, and staging resources can't be used for anything on GPU other than transfer operations.

One solution I could propose is:
  - change the instancing example (and others) to use Write-only mapping
  - fix `map_usage` to use the dynamic buffer type, but also make sure Read is not requested there on D3D11 only, returning an error

snuk182:
@sectopod fixing `map_usage` itself would not help, as there is a ban on using mappings for anything except Persistent ( @Bastacyclop will rename it to Mappable for clarity ), so either the example should be rewritten from `map_buffer` to `update_buffer` or mapping for DX11 should be eventually implemented (working on it, but not as fast as I'd want to). And yes, `RW` map access paradigm is indeed not portable, so should be rethinked...

Bastacyclop:
I am in favor of rewriting the example to use write-only mapping and returning something like `mapping::Error::AccessNotSupported(access)` when appropriate.

kvark:
ok, then, looks like a progress :)
@homu r+

PR #1103, opened 2016-12-24
App launcher upgrade

### LINKED ISSUE #907: Integrate winit
https://github.com/tomaka/winit is not published yet (cc @tomaka). When it releases, we should be able to tightly integrate with it (the list is moved here from #861):
- pass the universal window builder to the `gfx_window_*` initialization
- use the even loop for the application launcher to respond on key presses
- reduce the cruft of the DXGI window implementation

kvark:
Fixes #907

This PR simplifies the internals of gfx_app a bit, removing `Wrap*` typedefs as well as `Application*` traits.
It also properly routes the events to the application, making the whole API useful now.
The examples are not only updated, but also the controls are re-implemented for the `deferred`, `ubo_tilemap`, and `blend`.
A follow-up task #1104 has been created to make it even better.

Example of default backend with auto window:
```rust
    use gfx_app::Application;
    App::launch_simple("MyTitle");
```
Example of default backend with manual window:
```rust
    use gfx_app::Application;
    let wb = winit::WindowBuilder::new().with_title("MyTitle");
    App::launch_default(wb);
```
Example of forced GL3 backend:
```rust
    let wb = winit::WindowBuilder::new().with_title("MyTitle");
    gfx_app::launch_gl3::<App<_, _>>(wb);
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ addons:
 script:
   - export RUST_BACKTRACE=1
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cargo build --features vulkan; else cargo build; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cargo build --features metal; else cargo build; fi
   - cargo test -p gfx_core
   - cargo test -p gfx
   - cargo test -p gfx_device_gl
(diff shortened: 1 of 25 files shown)


kvark:
Travis is green, and given that all the backends are now properly tested as of this PR, we can proceed safely.
@homu r+

PR #1116, opened 2017-01-01
[v0.13] Capping the level by MAX_LEVEL

### LINKED ISSUE #1115: Bitshifting overflow while loading texture
I get this error while trying to load a texture:

    thread 'main' panicked at 'attempt to shift right with overflow', /gfx_core-0.4.0/src/tex.rs:205

It looks like it's coming from this line:

https://github.com/gfx-rs/gfx/blob/master/src/core/src/texture.rs#L272

Specifically, the `level` argument gets set to 16 at some point, and `w`/`h`/`d` are `u16`, which triggers a bitshifting overflow.

What's odd is that this only happens for debug mode with `cargo run`. Using `cargo run --release` makes everything work just fine. Doing some debugging showed that in release mode, `level` is always 0.

This is on Mac OSX with the OpenGL backend.

kvark:
Fixes #1115

### DIFF
--- a/src/core/Cargo.toml
+++ b/src/core/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_core"
-version = "0.5.0"
+version = "0.5.1"
 description = "Core library of Gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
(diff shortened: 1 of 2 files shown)


kvark:
@homu r+

PR #1120, opened 2017-01-04
WIP: Add on_resize to ApplicationBase.

djmcgill:

PR #1121, opened 2017-01-05
[APP] Resize support

### LINKED ISSUE #1104: [APP] Handle window resize
gfx_app should be providing the updated color/depth views upon window resize to the application.
Bonus points for updating the `triangle` example to handle the resize.
Extra bonus for converting the `gamma` into using `gfx_app`.

kvark:
Builds upon #1120
Fixes #1104

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,8 +47,7 @@ pub type DepthFormat = gfx::format::Depth32F;
 #[cfg(not(feature = "metal"))]
 pub type DepthFormat = gfx::format::DepthStencil;
 
-pub struct Init<R: gfx::Resources> {
-    pub backend: shade::Backend,
+pub struct WindowTargets<R: gfx::Resources> {
     pub color: gfx::handle::RenderTargetView<R, ColorFormat>,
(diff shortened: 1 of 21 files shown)


src/lib.rs
```diff
@@ -118,27 +121,37 @@ A: Sized + ApplicationBase<gfx_device_gl::Resources, gfx_device_gl::CommandBuffe
...
 
     let mut harness = Harness::new();
     loop {
         for event in window.poll_events() {
-            if !app.on(event) {
-                return
+            match event {
+                winit::Event::Closed => return,
+                winit::Event::KeyboardInput(winit::ElementState::Pressed, _, key) if key == A::get_exit_key() => return,
+                winit::Event::Resized(width, height) => if width != orig_width || height != orig_height {
```
djmcgill: If you resize the window (this event fires as expected), and then resize again back to the original dimensions then this event will not fire even though it should.

kvark: Nicely spotted, thanks! Fixed now

kvark:
Note: Dx11 resize doesn't work properly yet, but it's no fatal. See  #1122
@homu r+

Issue #423, opened 2014-11-07
Implement buffer copying


cmr:
See https://www.opengl.org/registry/specs/ARB/copy_buffer.txt

Bastacyclop:
@kvark I think me and @Androcas will simply add the following methods:
```rust
command::Buffer::copy_buffer(src: R::Buffer, dst: R::Buffer,
                             src_offset_bytes: usize, dst_offset_bytes: usize,
                             size_bytes: usize);

Encoder::copy_buffer(src: handle::Buffer, dst: handle::Buffer,
                     src_offset: usize, dst_offset: usize,
                     size: usize) -> Result<(), CopyError>;
```
Using [glCopyBufferSubData](https://www.opengl.org/sdk/docs/man/html/glCopyBufferSubData.xhtml) for OpenGL, [vkCmdCopyBuffer](https://www.khronos.org/registry/vulkan/specs/1.0/xhtml/vkspec.html#vkCmdCopyBuffer) for Vulkan (we won't handle synchronization) and [DeviceContext::CopySubresourceRegion](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476394(v=vs.85).aspx) for D3D11.

kvark:
@Bastacyclop sounds good!

PR #1123, opened 2017-01-06
Change usage semantics

### LINKED ISSUE #970: Change Usage semantics
The current `Usage` enum is heavily inspired by DX11. There is nothing wrong with it in particular, it maps nicely to both DX11 and GL. But looking forward, we need more flexibility, since other APIs are not locked to these usage types. I propose a set of flags instead:

``` rust
bitflags!(
    /// CPU transfer operations.
    /// In general, CPU access (read/write) forces a texture to have linear tiling, which reduces the GPU access pattern efficiency (hence, the rendering performance). Addition costs apply to synchronize CPU-GPU interaction by the driver.
    /// It is recommended though (if not enforced in the future) to not have renderable textures that are read/write. Instead, you create 2 separate textures - one for GPU only rendering with `TRANS_COPY` flag. Another - for CPU-only read/write texture ("staging texture" in DX11) that you will copy.
    pub flags Transfer: u8 {
        /// Resource is readable by CPU.
        pub TRANS_READ = 0x1,
        /// Resource is writable by CPU.
        pub TRANS_WRITE = 0x2,
        /// Resource can be copied from/into CPU visible memory. This flag barely affects performance.
        pub TRANS_COPY = 0x4,
    }
);
```

The backend will be able to derive the appropriate internal usage type from the `(Transfer, Bind)` flags. If a resource doesn't have any flags, it's considered immutable (in DX11 terms). `TRANS_WRITE` alone means the resource is dynamic. Both read/write mean the resource is staging. The backends are free to error out on an incompatible set of flags.

Bastacyclop:
Planned:
- [x] Modify core and OpenGL backend
- [x] Simplify the mapping code
- [x] Modify the other backends
- [x] Update gfx's examples and run them

Fixes #970
#423 will become necessary for staging.

### DIFF
--- a/examples/instancing/main.rs
+++ b/examples/instancing/main.rs
@@ -112,17 +112,23 @@ impl<R: gfx::Resources> gfx_app::Application<R> for App<R> {
          let size = 1.6 / instances_per_length as f32;
         println!("size: {}", size);
 
-        // the buffer might not be device local, for faster GPU access
-        // one may want to transfer its contents to a buffer that is
+        let staging = factory
+            .create_buffer(MAX_INSTANCE_COUNT,
+                           gfx::buffer::Role::Staging,
+                           gfx::memory::Usage::Upload,
+                           gfx::TRANSFER_SRC).unwrap();
+        {
+            let mut writer = factory.write_mapping(&staging).unwrap();
+            fill_instances(&mut writer, instances_per_length, size);
+        }
+
         let instances = factory
             .create_buffer(MAX_INSTANCE_COUNT,
                            gfx::buffer::Role::Vertex,
-                           gfx::memory::Usage::Upload,
-                           gfx::Bind::empty()).unwrap();
-        {
-            let mut writer = factory.write_mapping(&instances).unwrap();
-            fill_instances(&mut writer, instances_per_length, size);
-        }
+                           gfx::memory::Usage::Data,
+                           gfx::TRANSFER_DST).unwrap();
+
+        unimplemented!(); // transfer data
 
         let (quad_vertices, mut slice) = factory
             .create_vertex_buffer_with_slice(&QUAD_VERTICES, &QUAD_INDICES[..]);
--- a/examples/instancing/main.rs
+++ b/examples/instancing/main.rs
@@ -112,13 +112,16 @@ impl<R: gfx::Resources> gfx_app::Application<R> for App<R> {
          let size = 1.6 / instances_per_length as f32;
         println!("size: {}", size);
 
-        let (instance_buffer, mut instance_mapping) = factory
-            .create_mapped_buffer_writable(MAX_INSTANCE_COUNT,
-                                     gfx::buffer::Role::Vertex,
-                                     gfx::Bind::empty());
+        // the buffer might not be device local, for faster GPU access
+        // one may want to transfer its contents to a buffer that is
+        let instances = factory
+            .create_buffer(MAX_INSTANCE_COUNT,
+                           gfx::buffer::Role::Vertex,
+                           gfx::memory::Usage::Upload,
+                           gfx::Bind::empty()).unwrap();
         {
-            let mut instances = factory.write_mapping(&mut instance_mapping);
-            fill_instances(&mut instances, instances_per_length, size);
+            let mut writer = factory.write_mapping(&instances).unwrap();
+            fill_instances(&mut writer, instances_per_length, size);
         }
 
         let (quad_vertices, mut slice) = factory
@@ -134,7 +137,7 @@ impl<R: gfx::Resources> gfx_app::Application<R> for App<R> {
                 ).unwrap(),
             data: pipe::Data {
                 vertex: quad_vertices,
-                instance: instance_buffer,
+                instance: instances,
                 scale: size,
                 locals: factory
                     .create_buffer_immutable(&[locals], gfx::buffer::Role::Constant, gfx::Bind::empty())
--- a/src/backend/dx11/src/execute.rs
+++ b/src/backend/dx11/src/execute.rs
@@ -22,21 +22,8 @@ use {Buffer, Texture};
 pub fn update_buffer(context: *mut winapi::ID3D11DeviceContext, buffer: &Buffer,
                      data: &[u8], offset_bytes: usize) {
     let dst_resource = (buffer.0).0 as *mut winapi::ID3D11Resource;
-    /* TODO: do we want to allow updates for Data ?
-    let dst_box = winapi::D3D11_BOX {
-        left:   offset_bytes as winapi::UINT,
-        top:    0,
-        front:  0,
-        right:  (offset_bytes + data.len()) as winapi::UINT,
-        bottom: 1,
-        back:   1,
-    };
-    let ptr = data.as_ptr() as *const _;
-    unsafe {
-        (*context).UpdateSubresource(dst_resource, 0, &dst_box, ptr, 0, 0)
-    };
-    */
 
+    // DYNAMIC only
     let map_type = winapi::D3D11_MAP_WRITE_DISCARD;
     let hr = unsafe {
         let mut sub = mem::zeroed();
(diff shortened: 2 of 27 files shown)


src/render/src/factory.rs
```diff
@@ -125,51 +125,40 @@ pub trait FactoryExt<R: Resources>: Factory<R> {
...
+    fn create_constant_buffer<T>(&mut self, num: usize)
+                                 -> (handle::Buffer<R, T>,
+                                     mapping::WritableOnly<R, T>)
         where T: Copy
     {
-        let buffer = self.create_buffer_mappable(num, role, bind, memory::READ)
-            .unwrap();
-        let mapping = self.map_buffer_readable(&buffer).unwrap();
-        (buffer, mapping)
+        self.create_upload_buffer(num, buffer::Role::Constant, Bind::empty())
```
Bastacyclop: Or create a `Data` buffer with `TRANSFER_DST` ?

src/core/src/buffer.rs
```diff
@@ -103,6 +103,8 @@ pub enum Role {
     Index,
     /// Constant buffer
     Constant,
+    /// Staging buffer
+    Staging,
```
kvark: I don't think it makes sense to be in `Role`. Staging is a property of underlying memory, not the role. Any role buffer can be stages.

src/backend/gl/src/factory.rs
```diff
@@ -143,19 +147,9 @@ impl Factory {
...
-                    memory::READ => gl::DYNAMIC_READ,
-                    memory::WRITE => gl::DYNAMIC_DRAW,
-                    _ => unreachable!(),
-                },
-                CpuOnly(access) => match access {
-                    memory::READ => gl::STREAM_READ,
-                    _ => gl::DYNAMIC_DRAW,
-                }
+                Data => gl::STATIC_DRAW,
+                Upload => gl::DYNAMIC_DRAW,
```
kvark: not `STREAM_DRAW`?

src/core/src/mapping.rs
```diff
@@ -165,6 +166,8 @@ impl<'a, R: Resources, T: 'a + Copy> DerefMut for RWer<'a, R, T> {
     fn deref_mut(&mut self) -> &mut [T] { self.slice }
 }
 
+// TODO: should we keep the RW mapping stuff ?
```
kvark: I say kill it ;)

src/render/src/factory.rs
kvark: Yes, I think the constant buffers should be `Data`. Their CPU copies are attached to the command buffers, after all

src/core/src/buffer.rs
Bastacyclop: What happens when a buffer is not used for anything but staging ? It does not make much sense for it to be any of the current roles.

Bastacyclop: Maybe a `None` variant ?

kvark: I think you are right. Staging buffers don't fulfil any role directly (other than being staging, obviously), so your initial approach is correct. For GL, I suspect the staging buffer will really be a PBO target.

src/backend/gl/src/factory.rs
```diff
@@ -148,7 +148,7 @@ impl Factory {
         else {
             let usage = match info.usage {
                 Data => gl::STATIC_DRAW,
```
kvark: Warning: that would be inefficient for our constant buffers, which change often

Bastacyclop: I don't see any way to know if the data is going to change often with the new usage, should we reintroduce the `Dynamic` variant then ?

Bastacyclop: Another way could be to use `Upload` + mapping hidden behind the update commands.

Bastacyclop: I think a `Dynamic` variant is more flexible for the backend side.

kvark: I thought about it some more, let it be `STATIC_DRAW`, see the [discussion](https://gitter.im/gfx-rs/gfx?at=58724f1cc02c1a3959fedc57) in gitter.

src/backend/dx11/src/factory.rs
```diff
@@ -962,32 +930,29 @@ pub fn ensure_mapped(pointer: &mut *mut ::std::os::raw::c_void,
             DepthPitch: 0,
         };
             
-            
+        let dst = raw_handle.to_resource() as *mut winapi::d3d11::ID3D11Resource;
+        let map_type = winapi::d3d11::D3D11_MAP_WRITE; // no DISCARD because we are STAGING
```
Bastacyclop: I will need to put `MAP_READ` here when mapping for read.

src/backend/dx11/src/execute.rs
```diff
@@ -14,48 +14,40 @@
...
-                let dst = (sub.pData as *mut u8).offset(offset_bytes as isize);
-                ptr::copy_nonoverlapping(data.as_ptr(), dst, data.len());
-                (*context).Unmap(dst_resource, 0);
-                hr
-            };
-            if !winapi::SUCCEEDED(hr) {
-                error!("Buffer {:?} failed to map, error {:x}", buffer, hr);
-            }
-        },
+    /* TODO: do we want to allow updates for Data ?
```
Bastacyclop: I will remove this

src/backend/dx11/src/execute.rs
```diff
@@ -77,32 +69,40 @@ pub fn update_texture(context: *mut winapi::ID3D11DeviceContext, texture: &Textu
...
-                back:   (image.zoffset + image.depth) as UINT,
-            };
-            let ptr = data.as_ptr() as *const _;
-            unsafe {
-                //let subres = winapi::D3D11CalcSubresource(image.mipmap, array_slice, num_mipmap_levels);
-                (*context).UpdateSubresource(dst_resource, subres, &dst_box, ptr, row_pitch, depth_pitch)
-            };
-        },
-        Usage::Dynamic | Usage::CpuOnly(_) | Usage::Mappable(_) => unimplemented!(),
+    /* TODO: do we want to allow updates for Data ?
```
Bastacyclop: I will remove this too

src/backend/dx11/src/execute.rs
```diff
@@ -77,32 +69,40 @@ pub fn update_texture(context: *mut winapi::ID3D11DeviceContext, texture: &Textu
...
+        back:   (image.zoffset + image.depth) as UINT,
+    };
+    let ptr = data.as_ptr() as *const _;
+    unsafe {
+        //let subres = winapi::D3D11CalcSubresource(image.mipmap, array_slice, num_mipmap_levels);
+        (*context).UpdateSubresource(dst_resource, subres, &dst_box, ptr, row_pitch, depth_pitch)
+    };
+    */
+
+    // FIXME: probably wrong
```
Bastacyclop: I need to verify this part, I think it did not take many details into account.

Bastacyclop:
`create_buffer_immutable` may not produce an immutable buffer if the bind flags are "mutable".

Bastacyclop:
@kvark What do you think about it ? I think it's ready for merge (publishing should probably wait a bit), I changed dx11's [texture update](https://github.com/gfx-rs/gfx/pull/1123/files#diff-e53e2203bb8a16e2d7c3c80d560bb067R58) without actually testing it. `Upload` and `Download` are basically unusable and untestable without #423.

kvark:
@Bastacyclop thanks! I'll do another review pass today, and then we need to make sure it still works on Windows.
Then we'll probably jump on #423 before publishing.

Bastacyclop:
I ran the examples on linux and windows and did not see any issue (except that "instancing" cannot run), but there is no texture update in them (I think).

examples/instancing/main.rs
```diff
@@ -112,15 +112,24 @@ impl<R: gfx::Resources> gfx_app::Application<R> for App<R> {
...
 
-        let (instance_buffer, mut instance_mapping) = factory
-            .create_mapped_buffer_writable(MAX_INSTANCE_COUNT,
-                                     gfx::buffer::Role::Vertex,
-                                     gfx::Bind::empty());
+        let staging = factory
+            .create_buffer(MAX_INSTANCE_COUNT,
+                           gfx::buffer::Role::Staging,
+                           gfx::memory::Usage::Upload,
+                           gfx::TRANSFER_SRC).unwrap();
```
kvark: we might need a helper for the upload buffer constructor

kvark:
@homu r+

Issue #1102, opened 2016-12-24
Lower level GPU abstraction


kvark:
We need to provide a level of GPU abstraction that doesn't have any overhead over using some API bindings directly. That means *no* resource lifetime management and preferably no heap allocation.

If we try to implement such a layer from scratch, we'd end up re-using most of the existing code from gfx_core and the backends. Thus, I believe we should steer the gfx_core towards this no-overhead goal, reaching it before we stabilize the API.

One effect of this transition would be - clearer separation between gfx_render and gfx_core. Another - opened use cases for applications demanding no overhead, like [webrender](https://github.com/servo/webrender/). These applications may have limited use cases, allowing them to manage resource lifetimes more efficiently than us.

kvark:
We already have these 3 levels of resource type abstraction:
  1. Raw untracked structs (gfx_core, `R::Texture`).
  2. Tracked `Arc`/`Rc` handles with metadata (gfx_render, `handle::RawTexture<R>`).
  3. Tracked generically-typed handles (gfx_render, `handle::Texture<R, Type>`).

So the transition would be to make gfx_core operate on `R::*` structs entirely, not be aware of the tracked handles (which will live in gfx_render only), and minimize any heap allocations.

msiglreith:
> We need to provide a level of GPU abstraction that doesn't have any overhead over using some API bindings directly.

This is really difficult with regards to Vulkan/D3D12 as we would have to provide a very thin layer above those two APIs exposing concepts like resource barriers and explicit memory control. There was a talk by Chajdas concerning this some time ago (http://gpuopen.com/wp-content/uploads/2016/03/d3d12_vulkan_lessons_learned.pdf), which states that such a low level wrapper might be the best way for implementation.
(For gfx I started implementing something similar but I was kinda stuck at descriptor stuff and other projects had priority https://github.com/msiglreith/gfx/tree/gfx_next/src/core_next/src)

I assume the new architecture would be to implement `core` as some kind of bare metal API, which is kind of unsafe for the user. `Render` is then an higher-level entry point for users, which rather want an easy and safe gfx layer to use for quickly building applications, but don't need the full control over the GPU.

kvark:
@msiglreith thanks for the info! You seem to not only understand the goal clearly, but also already having the first bits of the implementation. That's pretty impressive and exciting to see 👍

msiglreith:
Small status update for `corell`:
[...]

PR #1125, opened 2017-01-06
 Initial groundwork for a low level api

msiglreith:
## Motivation
The next gen APIs (Vulkan and D3D12) offer an low level interface of the GPU in comparison to older APIs like OpenGL or D3D11. To avoid overhead due to high level abstraction we want to investigate the implementation of a thin wrapper above the mentioned low level APIs to address https://github.com/gfx-rs/gfx/issues/1102.

## Goals
This PR creates a new low level core API `corell` and two backend implementations for the low level core. corell is heavily based on the vulkan API and shares a lot of concepts like `Instance`, `Surface`, etc., but tries to also take d3d12 into account. In future steps further abstractions could be considered to reduce the amount of setup code.

Planned coverage for this PR:
- [x] Instance/Context
- [x] Surface
- [x] Swapchain
- [x] PhysicalDevice

Note: Also includes some dummy types for other objects.
`format` and `memory` are mainly copied from `core`.

Hoping for early feedback if this is the intended way to go with regards to #1102.

cc @MaikKlein for quick checking ash based vulkan implementation if you have time 😄

### DIFF
--- a/src/backend/vulkanll/src/lib.rs
+++ b/src/backend/vulkanll/src/lib.rs
@@ -12,17 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-extern crate gfx_corell as core;
 extern crate ash;
+extern crate gfx_corell as core;
+extern crate kernel32;
+#[macro_use]
+extern crate lazy_static;
+extern crate winit;
 
 use ash::version::{EntryV1_0, InstanceV1_0, V1_0};
 use ash::vk;
+use ash::Entry;
 
 use std::ffi::{CStr, CString};
 use std::iter;
 use std::mem;
 use std::ptr;
 use std::sync::Arc;
+
+lazy_static! {
+    static ref VK_ENTRY: Result<Entry<V1_0>, ash::entry::LoadingError> = Entry::new();
+}
 
 pub struct PhysicalDevice {
     info: core::PhysicalDeviceInfo,
@@ -60,12 +69,85 @@ impl core::CommandQueue for CommandQueue {
     }
 }
 
-pub struct Surface {
+struct SurfaceInner {
+    handle: vk::SurfaceKHR,
+    instance: Arc<InstanceInner>,
+    loader: vk::SurfaceFn,
+}
 
+impl Drop for SurfaceInner {
+    fn drop(&mut self) {
+        unsafe { self.loader.destroy_surface_khr(self.instance.0.handle(), self.handle, ptr::null()); }
+    }
+}
+
+pub struct Surface {
+    // Vk (EXT) specs [29.2.7 Platform-Independent Information]
+    // For vkDestroySurfaceKHR: Host access to surface must be externally synchronized
+    inner: Arc<SurfaceInner>,
+}
+
+impl Surface {
+    fn from_raw(instance: &Instance, surface: vk::SurfaceKHR) -> Surface {
+        let entry = VK_ENTRY.as_ref().expect("Unable to load vulkan entry points");
+
+        let loader = vk::SurfaceFn::load(|name| {
+                unsafe {
+                    mem::transmute(entry.get_instance_proc_addr(
+                        instance.inner.0.handle(),
+                        name.as_ptr()))
+                }
+            }).expect("Unable to load surface functions");
+
+        let inner = Arc::new(SurfaceInner {
+            handle: surface,
+            instance: instance.inner.clone(),
+            loader: loader,
+        });
+
+        Surface {
+            inner: inner,
+        }
+    }
 }
 
 impl core::Surface for Surface {
-    
+    type B = Backend;
+    type Window = winit::Window;
+
+    #[cfg(target_os = "windows")]
+    fn from_window(window: &winit::Window, instance: &Instance) -> Surface {
+        use winit::os::windows::WindowExt;
+        let entry = VK_ENTRY.as_ref().expect("Unable to load vulkan entry points");
+        let win32_loader = ash::extensions::Win32Surface::new(entry, &instance.inner.0)
+                        .expect("Unable to load win32 surface functions");
+
+        let surface = unsafe {
+            let info = vk::Win32SurfaceCreateInfoKHR {
+                s_type: vk::StructureType::Win32SurfaceCreateInfoKhr,
+                p_next: ptr::null(),
+                flags: vk::Win32SurfaceCreateFlagsKHR::empty(),
+                hinstance: unsafe { kernel32::GetModuleHandleW(ptr::null()) } as *mut _,
+                hwnd: window.get_hwnd() as *mut _,
+            };
+
+            win32_loader.create_win32_surface_khr(&info, None)
+                .expect("Error on surface creation")
+        };
+
+        Self::from_raw(instance, surface)
+    }
+
+    #[cfg(not(target_os = "windows"))]
+    fn from_window(window: &winit::Window, instance: &Instance) -> Surface {
+        unimplemented!()
+    }
+
+    fn build_swapchain<T: core::format::RenderFormat>(
+                    &self, width: u32, height: u32,
+                    present_queue: &CommandQueue) -> SwapChain {
+        unimplemented!()
+    }
 }
 
 pub struct SwapChain {
@@ -111,7 +193,7 @@ impl core::Instance for Instance {
 
     fn create() -> Instance {
         // TODO: return errors instead of panic
-        let entry = ash::Entry::new().expect("Unable to load vulkan entry points");
+        let entry = VK_ENTRY.as_ref().expect("Unable to load vulkan entry points");
 
         let app_info = vk::ApplicationInfo {
             s_type: vk::StructureType::ApplicationInfo,
--- a/src/backend/vulkanll/src/lib.rs
+++ b/src/backend/vulkanll/src/lib.rs
@@ -44,6 +44,7 @@ pub struct QueueFamily {
     instance: Arc<InstanceInner>,
     device: vk::PhysicalDevice,
     family_index: u32,
+    queue_type: vk::QueueFlags,
     queue_count: u32,
 }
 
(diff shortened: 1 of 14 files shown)


src/backend/vulkanll/src/lib.rs
```diff
@@ -45,14 +66,117 @@ pub struct SwapChain {
...
 
-pub struct Instance {
+struct InstanceInner(pub ash::Instance<V1_0>);
+impl Drop for InstanceInner {
+    fn drop(&mut self) {
+        unsafe { self.0.destroy_instance(None); }
+    }
+}
+
+const SURFACE_EXTENSIONS: &'static [&'static str] = &[
```
MaikKlein: You can also use Ash's extensions names but it is not necessary. `Surface::name().as_ptr()` would give you a const* i8.

src/backend/vulkanll/src/lib.rs
```diff
@@ -45,14 +66,117 @@ pub struct SwapChain {
...
+
+    fn enumerate_physical_devices(&self) -> Vec<PhysicalDevice> {
+        self.inner.0.enumerate_physical_devices()
+            .expect("Unable to enumerate physical devices")
+            .iter()
+            .map(|&device| {
+                // TODO: add an ash function for this
+                let properties = unsafe {
+                    let mut out = mem::zeroed();
+                    self.inner.0.fp_v1_0().get_physical_device_properties(device, &mut out);
```
MaikKlein: Seems that I didn't wrap that function yet, I'll do it tomorrow. If you encounter that again tell me and I'll add them for you.

MaikKlein:
The ash usage looks good.

src/corell/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// The device is based on a software rasterizer
+    pub software_rendering: bool,
+}
+
+pub trait Device {
+    
+}
+
+pub trait CommandQueue {
+    type B: Backend;
```
kvark: So the `Backend` points to all the things, like `CommandQueue`, `Device`, or even `Resources`. This is great, a real step forward from the Zoo of traits we got. But why do these traits have to point back at `Backend`? It seems like it would be more convenient to have `type CommandBuffer: CommandBuffer` here instead. Would there be any issues?

src/corell/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Submits a `CommandBuffer` to the GPU queue for execution.
+    fn submit(&mut self, cmd_buffer: &<<Self as CommandQueue>::B as Backend>::CommandBuffer);
+}
+
+pub trait SwapChain {
```
kvark: For GL, it looks like this would require glutin to be a part of the GL backend.
I would not mind, if not the fact that there is a few people who uses SDL/GLFW with gfx-rs at the moment.

src/corell/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    type ShaderResourceView:  Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync + Copy;
+    type UnorderedAccessView: Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync + Copy;
+    type RenderTargetView:    Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync + Copy;
+    type DepthStencilView:    Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync;
+    type Sampler:             Clone + Hash + Debug + Eq + PartialEq + Any + Send + Sync + Copy;
+}
+
+/// Different types of a specific API.
+pub trait Backend {
+    type CommandBuffer;
```
kvark: if there is no `CommandBuffer` trait, perhaps we need some `RenderEncoder`, `TransferEncoder` and `ComputeEncoder` ones?

src/corell/src/lib.rs
msiglreith: This wouldn't cause any issues.
I noticed only that it might only become a bit more verbose (e.g https://github.com/msiglreith/gfx/blob/gfx_next/src/core_next/src/lib.rs#L101)

src/corell/src/lib.rs
msiglreith: Yes, I'm currently not sure atm how to split up the different command buffer types, also how to properly support secondary command buffers and bundles. But this will be part of future work!

msiglreith:
## Current state
New way to setup the backend:
```Rust
fn main() {
    let window = winit::WindowBuilder::new()
        .with_dimensions(1024, 768)
        .build()
        .unwrap();

    // instantiate backend
    let instance = back::Instance::create();
    let physical_devices = instance.enumerate_physical_devices();
    let queue_descs = physical_devices[0].get_queue_families().iter().map(|family| { (family, family.num_queues()) }).collect();

    // build a new device and associated command queues
    let (device, queues) = physical_devices[0].open(queue_descs);

    let surface = back::Surface::from_window(&window, &instance);
    let mut swap_chain = surface.build_swapchain::<ColorFormat>(1024, 768, &queues[0]);

    'main: loop {
        for event in window.poll_events() {

        }

        let frame = swap_chain.acquire_frame();

        // rendering

        // present frame
        swap_chain.present();
    }
}
```
`Instance` is the main entry point for an application. It's similar to an OpenGL context to some extent. We'll mainly use it for enumerating the available GPU devices on the system.

`PhyiscalDevice` and `Surface` are host abstraction of the corresponding gpu device and window. They are needed to build `Device` and `Swapchain` (driver/application-level counterparts).

Every GPU and virtual adapter (e.g d3d12 WARP) will be represented as `PhysicalDevice` and the user can query information about these devices. Probably most interesting properties are the supported command queues, which will be created together with the device.

## Design decisions
### Thread behavior (Vulkan)
Some objects need to be externally synchronized (e.g vkDevice on destruction or vkQueue on submit).

I tried to avoid locking for these objects to minimize the overhead and enforced the synchronization using `Arc`/`Rc`. Reference counting also has some additional overhead but it should only used for 'large' objects like device and command queue, which internals won't be cloned very often (on startup usually).

An alternative approach would be trying to remove all kind of reference counting and entirely relay on references `&/& mut`, but this might become very tricky to ensure the lifetimes and handle the trait interfaces.

### Command queues
Only partially linked to this PR but I think worth to discuss early on.
Vulkan and D3D12 have different kind of command queues (also Metal's encoders, but I'm not familiar with the API):
- D3D12 allows 3 different queue types/families (http://images.slideplayer.com/11/3326273/slides/slide_20.jpg) with an hierarchical: Graphics queues support also compute and transfer/copy operations.
- Vulkan slightly differs: Graphics and Compute queues also support transfer/copy operations. If a physical device exposes a graphics queue, it also has to support a graphics+compute queue.

Possible queue combinations: Graphics/Copy, Graphics/Compute/Copy, Compute/Copy, Copy
Command buffers of the different types can then be submitted to the queues.
Difficulty here is the representation of the command buffers and command queues. We can either try to separate the different types to ensure at compile time that e.g a compute command buffer doesn't call `draw` or detect it at runtime.

## Minor issues
- I'm not totally happy with the name `PhysicalDevice` as it can also includes virtual devices like the mentioned WARP implementation, but couldn't find a good name (maybe adapter?)
- Often returning references to `Vec<_>` (e.g https://github.com/gfx-rs/gfx/pull/1125/files#diff-4e5f58801cbe25ef39fdaf3aff780d49R91), any suggestions for improvement here?

If I forgot to mention something I'll add it later!
Any feedback appreciated!

Bastacyclop:
> Often returning references to Vec<_>

Why not return iterators like `QueueFamiliyIter` or `QueueFamilies` (can be type aliases) ?

kvark:
@msiglreith thanks for the awesome writup!

```rust
let queue_descs = physical_devices[0].get_queue_families().iter().map(|family| { (family, family.num_queues()) }).collect();
```
Why do we need to process this? Could `physical_devices[0].open` just accept queue families directly?

```rust
let surface = back::Surface::from_window(&window, &instance);
```
This might look better as `let surface = instance.create_surface(&window);`

```rust
let mut swap_chain = surface.build_swapchain::<ColorFormat>(1024, 768, &queues[0]);
```
Doesn't `surface` already know about the size (or has the size)?

> Every GPU and virtual adapter (e.g d3d12 WARP) will be represented as PhysicalDevice and the user can query information about these devices

Nice to see d3d concepts fitting into the queue families!

> Reference counting also has some additional overhead but it should only used for 'large' objects like device and command queue, which internals won't be cloned very often (on startup usually).

It sounds reasonable, yes.

> Vulkan and D3D12 have different kind of command queues (also Metal's encoders, but I'm not familiar with the API):

Metal's encoders are close to our `Encoder`, it's not related to the command queue.

> Difficulty here is the representation of the command buffers and command queues. 

I have some ideas, will express them in the follow-ups.

> I'm not totally happy with the name PhysicalDevice as it can also includes virtual devices like the mentioned WARP implementation, but couldn't find a good name (maybe adapter?)

Yes, `Adapter` sounds good!

> Often returning references to `Vec<_>`

I'm not aware of any valid use case to return `&Vec<>`. If you have these queue families locally, you can return slices `&[]`.

msiglreith:
@Bastacyclop @kvark First of all thank you for the feedback!

> Why not return iterators like QueueFamiliyIter or QueueFamilies (can be type aliases of Vec iterator) ?

> I'm not aware of any valid use case to return &Vec<>. If you have these queue families locally, you can return slices &[].

Probably will try to use iterators and slices as fallback depending on the complexity and usability.

> Why do we need to process this? Could physical_devices[0].open just accept queue families directly?

In order to specify how many queues of a certain family we want to create. D3D12 theoretically supports infinite software queues and doesn't allows us to query how many hardware queues there are iirc.

> This might look better as let surface = instance.create_surface(&window);

> Yes, Adapter sounds good!

Agree! Going to change it.

> Doesn't surface already know about the size (or has the size)?

Actually yes. Vulkan allows to create swapchains images with different dimensions from the surface (vk specs extensions, 29.5 Surface Queries), but I can't imagine a use-case at the moment. Also switching this to the default windows size.

kvark:
@msiglreith the latest changes look good 👍

msiglreith:
I'm quite fine with the current design iteration.
Both backends should work fine (without crashing), Vulkan validation layers throw some errors due to lack of synchronization primitives and need to create the swapchain images.

All points mentioned in the feedback has been addressed iirc. Added TODOs throughout the source code for future work.

Upcoming topics to work on:

- Shaders, RenderPasses and Pipelines
- Resources and Memory
- Descriptor handling
- Synchronization
- Command Buffers

kvark:
My biggest concern here is the code duplication between our current gfx_core and the low-level counter-part. I'd prefer a much gradual transition, but it looks like we are too invested already, so merging this.

Note to all developers - any changes would have to go into both the current and new cores. Hence, we should make the swap and deprecate the old one as soon as possible!

@homu r+

PR #1137, opened 2017-01-17
Error improvements.

burtonageo:
Exposes more sub-errors through the `Error::cause` method, and imps the `Error` trait on more Error structs.


kvark:
Looks good, thank you!
@homu r+

PR #1138, opened 2017-01-17
Fix for the PSO error bounds

kvark:
We used to `impl<S: Error> Error for PipelineStateError` while actually using `S = String` and `S = &str`, which don't satisfy the `Error` bound, making the whole thing a bit useless. This PR fixes that.


kvark:
Well, since there are no objections, I'm merging this.
@homu r+

PR #1139, opened 2017-01-17
Add a gfx::Slice::split_at method.

nical:
A simple method to split a slice in two in a similar way as [std::slice::split_at](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at). This is useful when grouping the geometry for several draw calls in the same buffer.


kvark:
Thanks @nical ! You made it just in time for v0.14 (see #1140) :)
Waiting for CI and then merging. Let me know if you got any more changes in progress.

kvark:
@homu r+

PR #1143, opened 2017-01-17
[Metal] Better creation error tracking

kvark:
Sadly, Metal is not supported on my old Mac Mini 2010.
If someone got a newer one, please try running our examples with "--feature metal" flag. It would be nice to publish that Metal backend at last.

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -265,6 +265,11 @@ impl core::Device for Device {
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum InitError {
+    FeatureSet,
+}
+
 pub fn create(format: core::format::Format,
               width: u32,
               height: u32)
@@ -273,7 +278,7 @@ pub fn create(format: core::format::Format,
                          handle::RawRenderTargetView<Resources>,
                          *mut CAMetalDrawable,
                          *mut MTLTexture),
-                        ()> {
+                        InitError> {
     use core::handle::Producer;
 
     let share = Share {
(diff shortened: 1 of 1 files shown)


doppioslash:
@kvark:
On a 10.12.2 Macbook Pro from 2015:

- triangle, gamma, and shadow   succeed
- cube, deferred, flowmap and terrain fail with:
```
thread 'main' panicked at 'not yet implemented', src/backend/metal/src/command.rs:303
```
- blend, particle, mipmap, ubo_tilemap  fail with:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SelectError(Msl(11))', ../src/libcore/result.rs:799
```
- skybox crashes with:
```
-[MTLTextureDescriptorInternal validateWithDevice:]:774: failed assertion `MTLTextureDescriptor width must equal height for cube textures.'
```
- terrain_tessellated fail with:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Hull(StageNotSupported(Hull))', ../src/libcore/result.rs:799
```
- instancing fails with:
```
32 instances per length
1024 instances
size: 0.05
thread 'main' panicked at 'not yet implemented', src/backend/metal/src/factory.rs:760
```

kvark:
@doppioslash incredible, thank you! We'll revisit this comment many times for sure when implementing missing features. Please feel free to tackle them as well if you want!
@homu r+

kvark:
> triangle, gamma, and shadow succeed

Triangle and Gamma use glutin directly. Shadow example defaults to GL.
So it appears that Metal is no-functional at the moment...

Bastacyclop:
@kvark This is probably partially because of this from recent PRs: https://github.com/gfx-rs/gfx/blob/master/src/backend/metal/src/command.rs#L303:
1. There is an [usage redundance](https://github.com/gfx-rs/gfx/blob/master/src/backend/metal/src/lib.rs#L135) used in buffer updates just like the one I removed in DX11 (Vulkan may have it too).
1. [map_buffer_usage](https://github.com/gfx-rs/gfx/blob/master/src/backend/metal/src/map.rs#L531) now requires `Bind` too, so the latter is not enough to call it anymore.

However buffer updates are currently `Dynamic` only so things could be hardcoded for it.

kvark:
@Bastacyclop thanks for the info!
It would be really great to have someone picking up the Metal backend, given @fkaa 's absence. I wouldn't mind maintaining it myself, but my Mac is too old. Help wanted!

PR #1144, opened 2017-01-17
Make the CreateProgramError a proper struct

burtonageo:
Makes the CreateProgramError type a proper struct which impls the `Error` trait. This means that you can now get the cause of a `ProgramError::Link` error.


kvark:
Thanks!
@homu r+

PR #1146, opened 2017-01-18
Tracking the index buffer mapping

### LINKED ISSUE #1145: IndexBuffer does not support mappability
OpenGL, gfx-0.14.
https://github.com/snuk182/nuklear-backend-gfx
https://github.com/snuk182/nuklear-test/tree/gfx0.13
If we use Upload buffer to stream index data into, it is not unmapped before actual draw calls, which results in a crash:
`thread 'main' panicked at 'Error InvalidOperation executing command: DrawIndexed(4, 5123, RawOffset(0x0), 174, 0, None)', C:\Users\plyhuns\.cargo\registry\src\github.com-1ecc6299db9ec823\gfx_device_gl-0.13.0\src\lib.rs:753`

`GL_INVALID_OPERATION is generated if a non-zero buffer object name is bound to an enabled array or the element array and the buffer object's data store is currently mapped.`

This is because only PSO-data-based buffers go through the data binding mechanism which serves all the `ensure_(un)mapped` calls. Should adapt this to the `IndexBuffer` from a `Slice` as well.

Also we cannot override this by putting IndexBuffer into a pipe:
> the trait bound `gfx::IndexBuffer<u16>: gfx::pso::DataLink<'a>` is not satisfied [E0277]

kvark:
Fixes #1145
r? @Bastacyclop

### DIFF
--- a/src/render/src/encoder.rs
+++ b/src/render/src/encoder.rs
@@ -186,6 +186,7 @@ impl<R: Resources, C: command::Buffer<R>> Encoder<R, C> {
         if !dst.get_info().bind.contains(memory::TRANSFER_DST) {
             return Err(CopyError::NoDstBindFlag);
         }
+
         let size_bytes = mem::size_of::<T>() * size;
         let src_offset_bytes = mem::size_of::<T>() * src_offset;
         let src_copy_end = src_offset_bytes + size_bytes;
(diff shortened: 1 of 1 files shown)


snuk182:
Fix works.

kvark:
@homu r+

PR #1147, opened 2017-01-18
Use HashSet for AccessInfo

### LINKED ISSUE #1133: Use HashSet instead of Vec for AccessInfo
Follow up on #1129

Bastacyclop:
Also moves the structure from `core::pso` to `core::command` to highlight its relation with the command submission process.
Closes #1133


kvark:
@Bastacyclop looks great, thanks!
I suppose that's not a breaking change, right? Thinking of including it in gfx_core-0.6.1 if we are asked to update it.
@homu r+

Bastacyclop:
Yes, it shouldn't break anything.

PR #1150, opened 2017-01-20
Work around nonconforming OpenGL drivers returning the length of a uniform name without the trailing null byte.

CensoredUsername:
This fixes shader reflection on some OpenGL drivers when long
uniform names or nested structs are used in a shader. On conforming
drivers it has no effect.


kvark:
Perfect, thanks @CensoredUsername !
@homu r+

PR #1154, opened 2017-01-23
Add primitives with adjacency information

### LINKED ISSUE #1152: Adjacency information for geometry shaders
This enables effects such as silhouette detection, shadow volume extrusion, and other effects which rely on additional vertices in the geometry shader stage.

tocubed:
Fixes #1152 by adding additional variants to `Primitive`.


kvark:
Looks great, thanks @tocubed !
@homu r+

PR #1159, opened 2017-01-26
Forcing the command buffer to be sendable

kvark:


kvark:
Oh wow, getting some Vulkan and D3D11 failures. Will need to fix those...

kvark:
Closing in favour of #1177

Issue #1122, opened 2017-01-05
[DX11] Resize problem


kvark:
This is a follow-up to #1121
From https://msdn.microsoft.com/en-us/library/windows/desktop/bb174577(v=vs.85).aspx

> You can't resize a swap chain unless you release all outstanding references to its back buffers. You must release all of its direct and indirect references on the back buffers in order for ResizeBuffers to succeed. 

First problem is - we can't guarantee that the user has freed all the handles to our back buffer. Perhaps, we can invalidate it internally and then issue an error if one is trying to be used? Besides that we allow resize to fail with a console error but not fatal.

Another problem is - we don't have the new back buffer to replace the old back buffer before we call `ResizeBuffers`. One possible workaround is:
  1. create a dummy surface of size 1x1 first
  2. call `Application::on_resize()` with that dummy surface, expect it to replace all the old gfx-level references
  3. call `Device::clear_state()` and `Device::cleanup()`, expect the all the DX11-level references to the old back buffer to actually be deleted
  4. do the `ResizeBuffers`
  5. get the real back buffer
  6. call `Application::on_resize()` again with the real back buffer

PR #1160, opened 2017-01-26
DX11 resize fix

plyhun:
https://github.com/gfx-rs/gfx/issues/1122


snuk182:
Did not include a `gfx_app` fix here cause found it inconsistent with the current approach (`gfx_app` instance does not own the RTV so cannot unsure all its references are detached). Subject to discuss.

kvark:
Sweet, thanks!
@homu r+

PR #1162, opened 2017-01-29
Migrate low level core API changes to core.

msiglreith:
Add Surface, Instance, Swapchain, CommandQueue, QueueFamily and Adapter traits to core from corell.

The backends would need updates later-on. The dx12ll and vulkanll backends can be used for reference.
After the backend changes, the window crates shouldn't be needed anymore and replaced by the `Surface` and `SwapChain` parts.

Open question would be the support for other windows like SDL and glfw, because the current surface trait uses an associated window type.

### DIFF
--- a/src/core/src/lib.rs
+++ b/src/core/src/lib.rs
@@ -27,6 +27,7 @@ extern crate draw_state;
 use std::fmt::Debug;
 use std::hash::Hash;
 use std::any::Any;
+use std::slice::Iter;
 
 pub use draw_state::{state, target};
 pub use self::factory::Factory;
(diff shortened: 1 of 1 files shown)


Bastacyclop:
Could the window crates handle the `Instance` creation, each having their own types ?

msiglreith:
Oh, indeed this probably works fine as well.

src/core/src/lib.rs
```diff
@@ -238,3 +239,107 @@ pub trait Device: Sized {
     /// Cleanup unused resources. This should be called between frames. 
     fn cleanup(&mut self);
 }
+
+
+/// An `Instance` holds per-application state for a specific backend
+pub trait Instance {
+    /// Associated `Adapter` type.
+    type Adapter: Adapter;
```
kvark: is there a single adapter associated with the instance?
I thought the instance is the main entry point, and it can have multiple different adapters exposed.
These adapters may have different queue types/families, so must have different types.

src/core/src/lib.rs
```diff
@@ -238,3 +239,107 @@ pub trait Device: Sized {
...
+
+    /// Create a new device and command queues.
+    fn open<'a, I>(&self, queue_descs: I) -> (Self::Device, Vec<Self::CommandQueue>)
+        where I: Iterator<Item=(&'a Self::QueueFamily, u32)>;
+
+    /// Get the `AdapterInfo` for this adapater.
+    fn get_info(&self) -> &AdapterInfo;
+
+    /// Return the supported queue families for this adapter.
+    fn get_queue_families(&self) -> Iter<Self::QueueFamily>;
```
kvark: any reason it can't just return `&[Self::QueueFamily]`?
Also, why is `QueueFamily` a trait?

src/core/src/lib.rs
```diff
@@ -238,3 +239,107 @@ pub trait Device: Sized {
...
+
+/// A `Surface` abstracts the surface of a native window, which will be presented
+pub trait Surface {
+    /// Associated `CommandQueue` type.
+    type CommandQueue: CommandQueue;
+    /// Associated `SwapChain` type.
+    type SwapChain: SwapChain;
+
+    /// Create a new swapchain from the current surface with an associated present queue.
+    fn build_swapchain<T: format::RenderFormat>(&self, present_queue: &Self::CommandQueue)
```
kvark: I wonder... if a surface is only used to build the swap chain, why do we need the concept of a surface at all? Can't the instance return a swap chain directly, bypassing the surface (or hiding it)?

src/core/src/lib.rs
msiglreith: I considered it easier to use to return `Iter` instead of the slice. (It's basically the iterator equivalent to &[]).
Will change it if you prefer returning a slice.

`QueueFamily` is defined as trait as it was easier to implement and wasn't sure if it can be implemented as single struct for all backends. I can try to define a struct for it. Difficulties arised because of 
checking for presentation support for a surface.

src/core/src/lib.rs
msiglreith: I would also prefer to remove it. Similar issue as above: require a queue, which supports presentation on a surface to create a swapchain. But I need the surface first to check presentation support.
Again, I'll re-think it to see if I come up with a better solution!

src/core/src/lib.rs
msiglreith: There is a single adapter type associated, which is more like a backend abstraction for all adapters.
Adapters then define hardware devices and virtual devices and there capabilities (supported queue types).
I don't really see the issue here unfortunately.

kvark: Why should we have `Instance` and `Adapter` be separate types? They both kinda converge to representing a backend.

src/core/src/lib.rs
kvark: If it's always an array/slice on the implementation side, I'd strongly be for `&[]` returned value since it has less magic than returning an iterator.
As for the `QueueFamily` trait - I suppose it's fine, especially since most apps will not care about it.

kvark:
Thanks @msiglreith ! I think this has a few rough edges but this is totally fine for a WIP patch. Ready to merge when you are.

msiglreith:
Replaced `Iter` with `&[]` and split the `Instance` trait into separate functions for `Adapter` and `Surface`.
This might make a little bit more difficult to implement for the backends but hopefully improves usability.

kvark:
Sounds good, thanks!
@homu r+

kvark:
@homu retry

kvark:
Travis CI is having a hard time with OSX bots at the moment... Could anyone confirm that this PR builds on OSX, so that we can just force-merge it?

kvark:
@homu retry

JohnColanduoni:
Builds fine on OS X 10.12

PR #1163, opened 2017-01-30
Fix vertex attrib size for 16-bit attributes with more than one component

Nitori-:
I use U16Norm for texture coordinates and noticed that the GL calls weren't quite right.


kvark:
Nicely spotted, thanks!
@homu r+

PR #1165, opened 2017-01-30
Fix Metal backend

JohnColanduoni:
I've updated the parts of the Metal backend that were out of date, allowing the examples to run. I also fixed the stencil buffer support and improved the mechanism for applying clear/clear_depth/clear_stencil commands.


examples/triangle/main.rs
```diff
@@ -12,16 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+extern crate cgmath;
 #[macro_use]
 extern crate gfx;
-extern crate gfx_window_glutin;
-extern crate glutin;
+extern crate gfx_app;
```
kvark: I'm sorry, but we'd prefer to keep `triangle` example to be straight without basing on `gfx_app`.

src/backend/metal/src/command.rs
```diff
@@ -34,12 +34,17 @@ use native::{Rtv, Srv, Dsv};
 use metal::*;
 
 use std::ptr;
+use std::collections::HashMap;
+use std::collections::hash_map::Entry;
```
kvark: nit: could include both from `hash_map::`

src/backend/metal/src/command.rs
```diff
@@ -214,27 +220,64 @@ impl command::Buffer<Resources> for CommandBuffer {
...
+            if let Some(stencil_attachment) = stencil_attachment {
+                stencil_attachment.set_store_action(MTLStoreAction::Store);
+            }
 
-            let clear = unsafe { *depth.2 };
-
-            if let Some(val) = clear {
-                attachment.set_load_action(MTLLoadAction::Clear);
-                attachment.set_clear_depth(val as f64);
+            if let Entry::Occupied(mut clear_entry) = self.dsv_clear.entry(depth) {
```
kvark: I don't think you need to use the Entry API here, just a regular `get_mut` should do

JohnColanduoni: The reason I used it is that the entry may need to be deleted entirely, not just modified. If get_mut is used there would be two lookups in that case. For the depth target this is conditional so I can't simplify it to a remove(), but the render target analogue always removes the entry if present. I'll change that to remove() instead of mucking about with Entries.

kvark:
Oh, sounds like Entry makes sense here then, no need to touch that code!

src/backend/metal/src/factory.rs
```diff
@@ -247,6 +247,10 @@ impl core::Factory<Resources> for Factory {
                     .object_at(0)
                     .set_pixel_format(MTLPixelFormat::BGRA8Unorm_sRGB);
 
+                // We need fake depth attachments in case explicit writes to the depth buffer are required
+                pso_descriptor.set_depth_attachment_pixel_format(MTLPixelFormat::Depth24Unorm_Stencil8);
```
fkaa: I get an error for this:
```/Library/Caches/com.apple.xbs/Sources/Metal/Metal-85.82.1/Framework/MTLRenderPipeline.mm:1571: failed assertion `depthAttachmentPixelFormat is not a valid MTLPixelFormat.'```
Switching out for `MTLPixelFormat::Depth32_Stencil8` works though

fkaa:
Other than the triggered assertion on my end looks good to merge 👍

src/backend/metal/src/factory.rs
JohnColanduoni: Good catch, apparently D24S8 is only supported on *some* OS X GPUs (and not at all on iOS). According to the docs all feature sets support Depth32_Stencil8 so that's a good choice.

kvark:
Awesome, thanks @JohnColanduoni and @fkaa !
@homu r+

PR #1167, opened 2017-01-31
Add the rendering::graphics-api category

nical:
That way gfx will show up here: https://crates.io/categories/rendering::graphics-api.


nical:
Note that at the moment you need to publish cargo nightly for the category to show up on crates.io.

kvark:
Thanks!
@homu r+

PR #1174, opened 2017-02-03
Fix OSX compilation of trianglell example 

msiglreith:
OSX doesn't support our d3d12 or vulkan backend, so in this case the example will just do nothing.


kvark:
Thanks!
@homu r+

PR #1175, opened 2017-02-04
Added Metal shader library support

### LINKED ISSUE #1168: Metal shader libraries
Metal can load shaders at runtime in one of two formats: shader source or precompiled libraries (.metallib). Either way an MTLLibrary object is created, which is able to expose multiple functions by name. 

Because core only allows construction of VertexShader/PixelShader structs though a single method in the Factory trait, the Metal backend can only load shaders from source, and requires the shader function to have a specific name (vert for vertex, frag for fragment). This forces you to have at most one vertex and one fragment shader in each file, and even then the shader ends up being compiled twice at runtime.

I'm happy to do the work for this, but unfortunately this change cannot be entirely confined to the Metal backend because the VertexShader/PixelShader structures' internals are private. Would it be acceptable to make them public, or at least provide a doc(hidden) function to construct them inside the backend?

JohnColanduoni:
Fixes #1168

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -28,9 +28,9 @@ name = "gfx_device_metal"
 [dependencies]
 log = "0.3"
 gfx_core = { path = "../../core", version = "0.6" }
-cocoa = "0.2.5"
+cocoa = "0.5"
 libc = "0.2"
 objc = "0.1.8"
 objc-foundation = "0.1"
 bit-set = "0.4"
-metal-rs = "0.1"
+metal-rs = { git = "https://github.com/fkaa/metal-rs", rev = "5263898c2ce1fb8263653272014366eba45ef88d" }
--- a/src/backend/metal/src/factory.rs
+++ b/src/backend/metal/src/factory.rs
@@ -15,6 +15,7 @@
 use std::os::raw::c_void;
 use std::sync::Arc;
 use std::{mem, slice, str};
+use std::path::Path;
 
 // use cocoa::base::{selector, class};
 // use cocoa::foundation::{NSUInteger};
@@ -29,7 +30,7 @@ use command::CommandBuffer;
 
 use MTL_MAX_BUFFER_BINDINGS;
 
-use {Resources, Share, Texture, Buffer, Shader, Program, Pipeline};
+use {Resources, Share, Texture, Buffer, Shader, Program, ShaderLibrary, Pipeline};
 use native;
 use mirror;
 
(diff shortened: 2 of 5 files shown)


src/backend/metal/Cargo.toml
```diff
@@ -33,4 +33,4 @@ libc = "0.2"
 objc = "0.1.8"
 objc-foundation = "0.1"
 bit-set = "0.4"
-metal-rs = "0.1"
+metal-rs = { git = "https://github.com/fkaa/metal-rs", rev = "5263898c2ce1fb8263653272014366eba45ef88d" }
```
kvark: Why do you need a git revision here?
We may just ask @fkaa to publish a new version instead.
Although, I don't mind patching this later on.

src/backend/metal/src/factory.rs
```diff
@@ -150,6 +151,64 @@ impl Factory {
 
         self.device.new_depth_stencil_state(desc)
     }
+
+    pub fn create_library<P: AsRef<Path>>
+        (&mut self,
```
kvark: that's some unusual code formatting

src/backend/metal/src/factory.rs
```diff
@@ -150,6 +151,64 @@ impl Factory {
...
+         file: P)
+         -> Result<ShaderLibrary, core::shade::CreateShaderError> {
+        use core::shade::CreateShaderError;
+
+        let lib = match self.device.new_library_with_file(file) {
```
kvark: nit: might as well make this expression the last one:
```rust
match self.device.new_library_with_file(file) {
  Ok(lib) => ShaderLibrary { lib: lib },
  Err(err) => Err(CreateShaderError::CompilationFailed(err.into())),
}
```

kvark:
Thanks @JohnColanduoni !
Let me know if you don't want to address the minor issues I got, we can deal with them afterwards.

src/backend/metal/Cargo.toml
JohnColanduoni: That revision includes the merging of a pull request I recently submitted (to metal-rs) to add the binding for the newLibraryFromFile method.

I think it might be better to hold off because I plan to keep working on the metal backend which may require some more additions to the bindings, and the metal backend is not on crates.io itself at the moment, but that works too.

src/backend/metal/src/factory.rs
JohnColanduoni: Sorry, I was replicating what was done elsewhere in the file (e.g. Factory::create_buffer_immutable_raw on line 226) when there was a long return type that wouldn't fit if the usual line breaking (i.e. with &mut self on the same line) wouldn't work. How should I format it?

src/backend/metal/src/factory.rs
JohnColanduoni: :P Good catch!

kvark:
Thanks!
Formatting is fine, we'll clean everything up with rustfmt anyway.
@homu r+

JohnColanduoni:
The functions wouldn't be much good if they were private XD

kvark:
@homu r-

kvark:
@homu r+

kvark:
Could you squash it as well please?

kvark:
@homu r-

kvark:
Thanks!
@homu r+

PR #1177, opened 2017-02-04
Make command buffers sendable

kvark:


kvark:
Looks like Windows builds are good now. Homu will stop if Travis does not build.
@homu r+

Issue #1136, opened 2017-01-16
Crash with an empty vertex buffer


kvark:
> if you call `factory.create_vertex_buffer_with_slice(&[], &[] as &[u16])`, `encoder.flush(&mut device)` will panic afterwards
> the panic is 'Error after executing command BindProgram(0): InvalidValue'

Anttonii:
Did some digging around and the problem seems to caused by the fact that the program is attempting to store data into a 0-sized vertex buffer which results in invalid value error.

More information here: https://www.khronos.org/opengl/wiki/GLAPI/glBufferStorage

> GL_INVALID_VALUE is generated if size​ is less than or equal to zero.

kvark:
I'm not convinced that we should support the empty buffers/textures at all.
I think the `Factory` can just return an error for this case.

PR #1178, opened 2017-02-05
Use glBufferData instead of glBufferStorage for zero sized buffers

fintelia:
Zero sized buffers are allowed in OpenGL. However, attempting to create one using glBufferStorage will result in an [invalid operation error](https://www.opengl.org/sdk/docs/man/html/glBufferStorage.xhtml#errors). The glBufferData function does not have this limitation (only negative sizes are considered invalid) and should be used instead for this case.

### DIFF
--- a/src/backend/gl/src/factory.rs
+++ b/src/backend/gl/src/factory.rs
@@ -125,7 +125,7 @@ impl Factory {
             0 as *const gl::types::GLvoid
         };
 
-        if self.share.private_caps.buffer_storage_supported {
+        if self.share.private_caps.buffer_storage_supported && info.size > 0 {
             let usage = match info.usage {
                 Data => 0,
                 // TODO: we could use mapping instead of glBufferSubData


Bastacyclop:
@fintelia Why do you need zero-sized buffers ? Returning a creation error seems more appropriate to me.

Bastacyclop:
related to #1136

kvark:
Problem with this PR is that there is other code that assumes that a buffer was allocated via `glBufferStorage` based on `share.private_caps.buffer_storage_supported`.

fintelia:
@Bastacyclop Zero-sized buffers are useful when all vertex attributes will be computed by the vertex shader using `gl_VertexID`. For instance when rendering a heightmap, the Y value will often be sourced from a texture, while the X and Z coordinates can be easily computed from the vertex id. In this case, `mem::size_of::<Vertex>() == 0` so the total size of the buffer will be zero despite logically having a non-zero number of elements. It would be possible for the vertex struct to contain a dummy value (or actual vertex positions), but that seems kind of wasteful, especially considering that OpenGL doesn't impose such a limitation.

fintelia:
@kvark Would it be better to change `Factory::create_buffer` and `Factory::create_buffer_immutable` so that if the element type has size zero, they instead create buffers of size one byte (filled with garbage data)? This way glBufferStorage could still be used to create the buffer while user code wouldn't even have to know that the buffer was larger than requested.

kvark:
@fintelia 
> Zero-sized buffers are useful when all vertex attributes will be computed by the vertex shader using gl_VertexID

Did you consider not having a vertex buffer in the PSO at all then?
Sounds like what you need is not a zero-sized vertex buffer, but rather just a dummy buffer.

fintelia:
@kvark I didn't realize it was possible to create a PSO without any vertex buffer

kvark:
@fintelia FYI: the PSO components, in general, are allowed to not show up, or show up multiple times (btw, this applies to vertex buffers too), with a few exceptions: it doesn't make sense to specify more than one scissor component or depth/stencil component.

PR #1182, opened 2017-02-10
Better PSO element errors

kvark:
Example:
> thread 'main' panicked at 'Could not create PSO for `DrawShaded`!: DescriptorInit(ConstantBuffer("u_PointLights", Some(Offset { name: "plight[1].center", shader_offset: 48, code_offset: 44 })))'

cc @ebkalderon


kvark:
@homu r+

PR #1183, opened 2017-02-10
[WIP] Macros 1.1

kvark:


kvark:
r? anyone

src/macros/src/lib.rs
```diff
@@ -0,0 +1,72 @@
+#![recursion_limit = "192"]
+
+extern crate proc_macro;
+extern crate syn;
+#[macro_use] extern crate quote;
+
+use proc_macro::TokenStream;
+
+
+#[proc_macro_derive(GfxVertexFormat)]
```
Bastacyclop: is the `Gfx` prefix necessary ? (I hope there is a way to rename a macro on the user side if he has clashes)

src/macros/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+    let gen = structure(ast, quote!(gfx::shade::Formatted), quote!(gfx::shade::ConstFormat));
+    gen.parse().unwrap()
+}
+
+fn structure(ast: syn::DeriveInput, ty_compile: quote::Tokens, ty_run: quote::Tokens)
+             -> quote::Tokens {
+    let name = &ast.ident;
+    let fields = match ast.body {
+        syn::Body::Struct(syn::VariantData::Struct(ref fields)) => fields,
+        _ => panic!("`GfxStruct` can only be derived for structs"),
```
Bastacyclop: We are not really deriving `GfxStruct`, I think rustc already wraps the derive panics inside a pretty message with the derive name and location.

kvark:
Thanks @Bastacyclop ! Fixed now, merging in an hour.

src/macros/src/lib.rs
kvark: Nicely spotted! Fixed in #1207

ebkalderon:
This is fantastic! ❤️ I think this also means that my previous issues with `#![deny(missing_docs)]` is no longer a problem, correct?

Issue #1127, opened 2017-01-10
Travis is green with errors


Bastacyclop:
Travis is [green with errors](https://travis-ci.org/gfx-rs/gfx/jobs/190336743#L493). I think this is because of the `;` command chaining in [.travis.yml](https://github.com/gfx-rs/gfx/blob/14706a41c0f3c2b2800a67bedf1118bbcae91ee2/.travis.yml#L51) ?

kvark:
Not seeing any errors any more, and #1185 has landed, closing.

PR #1185, opened 2017-02-10
Using cargo test --all

kvark:


Bastacyclop:
I think this would fix #1127 ?

kvark:
@homu r+

kvark:
@Bastacyclop if we still have errors, it wouldn't pass the CI if it fixed #1127

PR #1187, opened 2017-02-13
[GL] Work around 0-sized storage buffers

### LINKED ISSUE #1136: Crash with an empty vertex buffer

kvark:
Fixes #1136 

cc @stalkerg


kvark:
@homu r+

PR #1191, opened 2017-02-17
bump image dependency version

lxnt:
Existing dependency is on 0.6, which spews a bunch of warnings. This is annoying, so let's bump the dependecy to the latest version. 

Tested: examples.


kvark:
Thanks!
@homu r+

PR #1193, opened 2017-02-18
Implement FactoryExt::create_index_buffer.

nical:
A simple addition to the Factory.
While it was easy to create an index buffer separately before (in terms of required boilerplate) using the ToIndexBuffer trait, it was quite hard to find out that is is possible to do at all.


kvark:
Thanks!
@homu r+

PR #1194, opened 2017-02-18
[WIP] [DX11] fix crash when window is minimized

### LINKED ISSUE #1156: [DX11] All examples crash when window is minimized
Windows / DX11 / Gfx-0.14.0 and master branch, too.

gfx_window_dxgi::window.swap_buffers() panics at 'assertion failed: `(left == right)` (left: `0`, right: `142213121`)'. This happens when you minimize application window. Can be reproduced on every example code.

Androcas:
Closes #1156 

I fixed the minimization crash by ignoring `DXGI_STATUS_OCCLUDED`, so the window will be resized to (0, 0). 
I called `update_views` in gfx_app and I'm not sure if `device.clear_state()` is necessary.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -217,8 +217,8 @@ A: Sized + ApplicationBase<gfx_device_dx11::Resources, D3D11CommandBuffer>
             }
         }
         if let Some((width, height)) = new_size {
-            device.clear_state();
-            match window.resize_swap_chain(&mut factory, width, height) {
+            use gfx_window_dxgi::update_views;
+            match update_views(&mut window, &mut factory, &mut device, width, height) {
                 Ok(new_color) => {
                     let new_depth = factory.create_depth_stencil_view_only(width, height).unwrap();
                     app.on_resize(&mut factory, WindowTargets {
@@ -227,7 +227,7 @@ A: Sized + ApplicationBase<gfx_device_dx11::Resources, D3D11CommandBuffer>
                         aspect_ratio: width as f32 / height as f32,
                     });
                 },
-                Err(hr) => error!("Resize failed with code {:X}", hr),
+                Err(e) => error!("Resize failed: {}", e),
             }
             continue;
         }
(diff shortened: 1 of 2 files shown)


src/window/dxgi/src/lib.rs
```diff
@@ -178,16 +179,16 @@ pub fn init_raw(wb: winit::WindowBuilder, color_format: format::Format)
 }
 
 /// Update the internal dimensions of the main framebuffer targets. Generic version over the format.
-pub fn update_views<Cf>(window: &mut Window, factory: &mut Factory, device: &mut Device, width: u16, height: u16)
+pub fn update_views<Cf, D>(window: &mut Window, factory: &mut Factory, device: &mut D, width: u16, height: u16)
             -> Result<h::RenderTargetView<Resources, Cf>, f::TargetViewError>
-where Cf: format::RenderFormat
+where Cf: format::RenderFormat, D: core::Device
```
kvark: why not use `Device` directly as it used to?

src/window/dxgi/src/lib.rs
```diff
@@ -178,16 +179,16 @@ pub fn init_raw(wb: winit::WindowBuilder, color_format: format::Format)
...
 {
     use core::Device;
     
     factory.cleanup();
-    device.clear_state();
+    //device.clear_state();
```
kvark: `clear_state` is still needed in order to really free the old back buffer

src/lib.rs
```diff
@@ -217,8 +217,9 @@ A: Sized + ApplicationBase<gfx_device_dx11::Resources, D3D11CommandBuffer>
             }
         }
         if let Some((width, height)) = new_size {
-            device.clear_state();
-            match window.resize_swap_chain(&mut factory, width, height) {
+            println!("{:?}", (width, height));
```
kvark: temporary `println`?

Androcas:
I used a generic to manage the two devices of DX11 (Deferred and Device)

Androcas:
By the way, the examples are not releasing the targets so they all raise the error.

kvark:
Thanks!
Is this still WIP?

Androcas:
I think it's ready to be merged

kvark:
@homu r+

ngreduce:
Hi. When this fix will be on crates.io?

PR #1196, opened 2017-02-19
[ll] Draft for pipeline and shader handling

msiglreith:
The interesting parts are in the `corell` directory (scroll down to the bottom!). Especially in the `factory` and `pso` module.

## API design
- `ShaderLibrary`: A collection of shader entry points following the SPIR-V module concept from Vulkan and the metallib from Metal (`LibrarySource::Module`). On other backends these will be manually constructed from single shaders with a corresponding entry name (`LibrarySource::Composed`).
- `PipelineStateObject`: Mostly matches the current graphics pipeline state object. Additionally, it depends now on the subpass of renderpass and requires to set the shader entrypoints.

## Upcoming changes
- `RenderPass`: A more high-level concept from Vulkan, consisting of multiple subpasses. They will be emulated on other backends, but will have zero/low overhead.
- `PipelineSignature`: Specifies the layout of the corresponding pipeline and must match the layout in the shaders. This concept corresponds to `PipelineLayout`s in Vulkan and `RootSignatures` in D3D12, where we will follow the `PipelineLayout` from Vulkan more closely as they are more restricted. Also, having to look into details here if we require to also define `DescriptorSets`.

## Difficulties
Vulkan and D3D12 are bindless APIs using descriptors, which is not supported in older APIs. This will probably result into some conflicts. Unsure how to solve this in the future.

Example usage of the new API for the trianglell example:
```Rust
    // for the D3D12 backend
    let shader_lib = device.create_shader_library(
            shade::LibrarySource::Composed(&[
                ("vs_main", include_bytes!("data/vs_main.o")),
                ("ps_main", include_bytes!("data/ps_main.o"))])
        ).expect("Error on creating shader lib");

    let shader_entries = pso::GraphicsShaderSet {
        vertex_shader: "vs_main",
        hull_shader: None,
        domain_shader: None,
        geometry_shader: None,
        pixel_shader: "ps_main",
    };

    let pipeline_signature = device.create_pipeline_signature();

    let pipeline_desc = pso::GraphicsPipelineDesc::new(
        Primitive::TriangleList,
        state::Rasterizer::new_fill(),
        shader_entries);

    // `SubPass` requires a `RenderPass`, haven't updated the example at this point
    let pipelines = device.create_graphics_pipelines(&[
        (&shader_lib, &pipeline_signature, 0, &pipeline_desc) 
    ]);
```

Feedback is welcome!


src/backend/dx12ll/src/factory.rs
```diff
@@ -0,0 +1,196 @@
...
+                        winapi::D3D12_SHADER_BYTECODE {
+                            pShaderBytecode: ptr::null(),
+                            BytecodeLength: 0,
+                        }
+                    }
+                }
+            };
+
+            let vs = build_shader(shader_lib, Some(desc.shader_entries.vertex_shader));
+            let ps = build_shader(shader_lib, Some(desc.shader_entries.pixel_shader));
```
kvark: this one may be optional too

src/backend/dx12ll/src/factory.rs
```diff
@@ -0,0 +1,196 @@
...
+use std::collections::BTreeMap;
+
+use core::{self, shade};
+use core::SubPass;
+use core::pso::{self, EntryPoint, GraphicsShaderSet};
+use {data, state, native};
+use {Device, Resources as R};
+
+impl core::Factory<R> for Device {
+    fn create_shader_library<'a, 'b>(&mut self, sources: shade::LibrarySource<'a, 'b>) -> Result<native::ShaderLib, shade::CreateShaderError> {
```
msiglreith: Thanks for taking a look!
> It appears to me that the choice between Module/Composed is defined by the backend itself,

Exactly! I'm also not totally happy with this API but it was the best way I found for the current iteration.
Are you suggesting to remove the function from the trait and implement separate functions (different signature) individually for each backend?

kvark: Yes, although you probably see the picture better at this point. I don't think we should be baking into run-time what is known at compile time. Shader linking and loading is something very specific to the backends (at least until we get the unified language) anyway, so it wouldn't be too much harm to move `create_shader_library`  out of the trait. At the end of the day, we can consider this to be "staging" API, as in - we'll observe how it's used to make decisions on what to "stabilize" by moving it into the core trait.

kvark:
@homu delegate=msiglreith

examples/trianglell/main.rs
```diff
@@ -45,10 +52,62 @@ fn main() {
...
+    let render_pass = device.create_renderpass();
+
+    //
+    let mut pipeline_desc = pso::GraphicsPipelineDesc::new(
+        Primitive::TriangleList,
+        state::Rasterizer::new_fill(),
+        shader_entries);
+
+    pipeline_desc.color_targets[0] = Some((ColorFormat::get_format(), state::MASK_ALL.into()));
+    pipeline_desc.vertex_buffers.push(pso::VertexBufferDesc {
```
kvark: O_o, run-time pipeline setup? great!

src/backend/dx12ll/src/factory.rs
```diff
@@ -14,35 +14,84 @@
...
         // TODO: handle entry points with the same name
         for &(entry_point, byte_code) in shaders {
-            shader_map.insert(entry_point, byte_code.into());
+            let mut blob = ComPtr::<winapi::ID3DBlob>::new(ptr::null_mut());
+            let hr = unsafe {
+                d3dcompiler::D3DCreateBlob(
+                    byte_code.len() as u64,
+                    blob.as_mut() as *mut *mut _)
+            };
+            println!("{:?}", hr); // TODO: error handling
```
kvark: temporary?

src/backend/dx12ll/src/mirror.rs
```diff
@@ -0,0 +1,75 @@
...
+    ) -> Vec<InputElemDesc>
+{
+    let shader_desc = unsafe {
+        let mut desc = mem::zeroed();
+        vertex_reflection.GetDesc(&mut desc);
+        desc
+    };
+
+    // DEBUG:
+    println!("{:?}", shader_desc);
```
kvark: just use `debug!`, `warn!`, and `info!` instead

src/corell/src/pso.rs
```diff
@@ -142,6 +176,8 @@ impl GraphicsPipelineDesc {
             depth_stencil: None,
             shader_entries: shader_entries,
             color_targets: [None; MAX_COLOR_TARGETS],
+            vertex_buffers: Vec::new(),
+            attributes: Vec::new(),
         }
     }
 }
```
kvark: nit: new line missing

kvark:
I find it amusing that the PR conflicts with `trianglell` changes, since I assume you are the only one working on `trianglell` :)

msiglreith:
Rebased and fixed merge conflicts. Also addressed the points mentioned (thanks for reviewing!).
Going to shift proper `RenderPass` and `PipelineSignature` handling into the next PR as the current one already is pretty large.
Would be ready for merging if everything is fine (:

kvark:
Thank you! Let's make CI pass before we merge.

kvark:
Wonderful!
@homu r+

Issue #1111, opened 2016-12-27
Bump version number in README


pcstl:
When explaining how to add gfx as a dependency, the current README in the repo specifies version 0.12 as the version to be added to the user's Cargo.toml. I installed it myself while following the README's instructions. However, if version 0.12 is specified by itself, Cargo will install the latest version of gfx_core, causing a series of errors mainly related to trait bounds (which I spent a few hours trying to figure out until realizing the current version was 0.13)

Bumping the version number will avoid having new users deal with unfamiliar error messages when they first install gfx. I fixed this in #1110.

PR #1200, opened 2017-02-26
Update README.md

DaseinPhaos:
Bump version number. #1111 
Add reference to @icefoxen's `LearningGfx` post.
Change the usage of term "module" into "crate", which feels more appropriate.


kvark:
Very nice, thank you!
@homu r+

PR #1205, opened 2017-03-05
Update objc crate version

burtonageo:
Update objc crate version for Metal backends.


kvark:
Thanks for the PR!
It appears that OSX nightly fails because something else is still using objc-1.8:
https://travis-ci.org/gfx-rs/gfx/jobs/207862091#L1134

PR #1209, opened 2017-03-10
Bump metal-rs version

fkaa:
Should hopefully fix build errors in #1205


kvark:
@homu r+

kvark:
Looks like we are still missing something

kvark:
@homu r+
yay, thanks @fkaa !

PR #1210, opened 2017-03-11
[ll] Low level command buffer support

msiglreith:
First iteration of type safe command buffer handling including `CommandPool` and `CommandQueue`.
The `Device` trait got removed for the moment and replaced by a combination of `Factory` and `CommandQueue`s, where the first one is responsible for resource creation and the queues for command buffer allocation and submission.

There are 3 different fundamental types of work: graphics, compute and transfer.
Vulkan allows queues which support: graphics+compute+transfer, graphics+transfer, compute+transfer and transfer.
D3D12 allows graphics+compute+transfer, compute+transfer and transfer.
The rest of the APIs only have one general graphics+compute+transfer queue type.

We followed the vulkan model and defined a `GeneralCommandQueue`, `GraphicsCommandQueue`, `ComputeCommandQueue` and `TransferCommandQueue`.

The command queues can only submit command buffers with supported type of work, e.g. `GraphicsCommandQueue` only support `GraphicsCommandBuffer` and `TransferCommandBuffer`.

In a similar fashion, command buffers are divided. Additionally, there is a `SubpassCommandBuffer`, which can be executed in subpasses of a RenderPass. These command buffers will map to a subset of secondary command buffers for vulkan to allow parallel command recording of subpasses. (We currently don't have specific support for d3d12 bundles!)

CommandBuffers are allocated from CommandPools with the corresponding type. These pools are associated with a command queue and non-Send\Sync as CommandBuffers. Future work will focus on handle submission due to the non-Send restriction.

Work left for this PR:
- [x] Implement Vulkan command pool and queue support
- [x] ~~Implement command buffer for at least one backend~~
(Edit: Requires to some extent resource handling, which would be too much for this PR)
- [x] Improve documentation

As always, feedback and further suggestions (or also questions) appreciated!


examples/trianglell/main.rs
```diff
@@ -108,12 +108,15 @@ fn main() {    env_logger::init().unwrap();
...
 
     //
-    let pipelines = device.create_graphics_pipelines(&[
+    let pipelines = factory.create_graphics_pipelines(&[
         (&shader_lib, &pipeline_signature, SubPass { index: 0, main_pass: &render_pass }, &pipeline_desc)
     ]);
 
     println!("{:?}", pipelines);
 
+    let mutgraphics_pool = queue::GraphicsCommandPool::<back::CommandPool>::from_queue(&mut general_queues[0], 16);
```
Bastacyclop: typo here?

Bastacyclop:
Looks promising :)
If command buffers are not `Send`/`Sync`, how is multithreaded rendering going to work ? Are there going to be submitted directly on the owning thread ?

msiglreith:
> If command buffers are not Send/Sync, how is multithreaded rendering going to work ? Are there going to be submitted directly on the owning thread ?

My current idea is to create some work submission object, which are created once we finished recording of a command buffer. These submission objects are send/sync as you can't operate on them and they only carry the internal command buffer. Instead of directly submitting a command buffer as in the current implementation, we will submit these special objects.

msiglreith:
Implemented the submit approach (need to enfore Send/Sync though). It was actually required due to allow recording multiple command buffers sequential and submit at the end to avoid conflicts due to multiple `&mut` borrows.
Additionally, refactor command pools a bit and command buffers are now recorded with an `Encoder` struct to enforce that we don't try to add new commands after `finish()`.

msiglreith:
Would be ready to merge. A lot of parts are still unimplemented or only stubs.
Current design will maybe change in future, depending of how implementations will work out (e.g remove command buffer traits and only expose an unsafe command buffer trait with safe wrapper command buffers). Some parts like re-submitable command buffers and reset synchronization will be part of another PR in the future.

Next steps will aim at resource creation, descriptor handling and synchronization primitves to get some first visible results! I assume this will be a first stage to futher work on improving the API and move some more 'fixed' parts into the `core` API.

examples/trianglell/main.rs
```diff
@@ -83,8 +84,8 @@ fn main() {    env_logger::init().unwrap();
         pixel_shader: Some("ps_main"),
     };
 
-    let pipeline_signature = device.create_pipeline_signature();
-    let render_pass = device.create_renderpass();
+    let pipeline_signature = factory.create_pipeline_signature();
```
kvark: let's call it a pipeline layout instead?

src/corell/src/pool.rs
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+
+use std::ops::{DerefMut};
+use {CommandPool, CommandQueue};
+pub use queue::{GeneralQueue, GraphicsQueue, ComputeQueue, TransferQueue};
+
+/// General command pool can allocate general command buffers.
+pub trait GeneralCommandPool: CommandPool {
+    fn from_queue<Q>(queue: &mut Q, capacity: usize) -> Self
+        where Q: Into<GeneralQueue<<Self as CommandPool>::Queue>> +
```
kvark: in what case would `Q` not be `GeneralQueue<Self::Queue>`?

src/corell/src/queue.rs
```diff
@@ -0,0 +1,164 @@
...
+
+/// General command queue, which can execute graphics, compute and transfer command buffers.
+pub struct GeneralQueue<Q: CommandQueue>(Q);
+impl<Q: CommandQueue> GeneralQueue<Q> {
+    #[doc(hidden)]
+    pub unsafe fn new(queue: Q) -> Self {
+        GeneralQueue(queue)
+    }
+
+    pub fn submit_general(&mut self, cmd_buffers: &[Submit<<Q as CommandQueue>::GeneralCommandBuffer>]) {
```
kvark: I don't think you need to write `<Q as CommandQueue>` here and below, can just use `Q`

src/corell/src/pool.rs
msiglreith: Under no circumstances, but makes the implementation a bit easier for the backends.
Can change it if you prefer.

examples/trianglell/main.rs
msiglreith: Yes, as it will follow more the vulkan pipeline layout model than root signatures.

kvark:
Awesome, thanks, :ship: it!
@homu r+

PR #1214, opened 2017-03-23
Delete `gfx_app::Backend`

burtonageo:
It has been completely succeeded by `gfx_app::shade::Backend`.


kvark:
@burtonageo it's not succeeded by the `shade::Backend`. `gfx_app::Backend` was meant to be a run-time option to gfx_app itself (but we haven't found a way for this to work), while `shade::Backend` is just a piece of information for an application.

kvark:
Closing due to inactivity

PR #1215, opened 2017-03-25
[ll] A first triangle!

msiglreith:
(PR got larger than I expected..)

## RenderPass
The render pass approach follows vulkan here including attachments, framebuffers, image layout transitions etc. Other backends will need to emulate the necessary transition steps.
The API could be a bit safer to enforce same attachments for PSO, RenderPass and Framebuffers.

## Memory Management
Vulkan and d3d12 expose memory as heaps with different properties  (caching protocol and memory location). From a heap, memory can be allocated, which can be bound to specific resources (buffer/image).
The heap model from vulkan is more feature rich in comparison to d3d12, which exposes `DEFAULT`, `UPLOAD`, `READBACK` (+ `CUSTOM`) heaps, so we follow vulkan here regarding heap types.
From heap types we can allocate `Heap`s, in Vulkan only called `DeviceMemory`, where we can bind memory to resources (immutable step).

## RenderPassEncoders
Subpasses can be encoded in two different ways: inline in the primary command buffer or via secondary buffers. Vulkan allows only one encoding mode per subpass. Inline encoding doesn't allow multi threaded recording for the subpass, but probably has less overhead for secondary submission and easier to emulate on other backends.

The current triangle vertex attributes are hardcoded as vertex buffer support is still WIP (memory map/unmap). Also, the d3d12 is disabled for the moment and will be updated in the next PRs else this one would be become way too big.

## Future work
Main components missing in corell are support for memory map/unmap, synchronization primitives and descriptor sets.

Edit: just noticed that I need to rebase!


src/corell/src/pso.rs
```diff
@@ -187,14 +188,46 @@ impl GraphicsPipelineDesc {
...
+        const VERTEX_SHADER = 0x8,
+        /// Hull shader execution.
+        const HULL_SHADER = 0x10,
+        /// Domain shader execution.
+        const DOMAIN_SHADER = 0x20,
+        /// Geometry shader execution.
+        const GEOMETRY_SHADER = 0x40,
+        /// Pixel shader execution.
+        const PIXEL_SHADER = 0x80,
+        const EARLY_FRAGMENT_TESTS = 0x100,
```
msiglreith: Should be before `PIXEL_SHADER`, my vulkan specification was unclear at this point, but it doesn't make sense and I double-checked against the 1.0.42, which explicitly states the logical order (early before pixel shader).

src/backend/vulkanll/src/factory.rs
```diff
@@ -178,128 +278,345 @@ impl core::Factory<R> for Factory {
...
+                            let pipeline = native::GraphicsPipeline {
+                                pipeline: *pipelines.next().unwrap(),
+                            };
+                            Ok(pipeline)
+                        }
+                        Err(ref err) => Err(err.clone()),
+                    }
+                }).collect::<Vec<_>>()
+            },
+            Err(err) => {
```
msiglreith: On error I should check each returned handle for `VK_NULL_HANDLE` and return error on null handle. Currently returning error if *one* pipeline creation fails, which isn't really useful. I'm going to fix this!

kvark:
@msiglreith this is pretty scary for the review. Let's discuss the issues you are most unsure about separately from submitting PRs, if possible. Also, amazing work here!
@homu r+

PR #1218, opened 2017-04-02
[ll] Image creation and memory mapping

msiglreith:
Add basic memory mapping support, a simplified approach based on the current core mapping. Doesn't include tracking for ensuring range flushing or overlapping mapping, but is more lightweight.

Also fixed vertex attribute description and the trianglell example recieves data now from a vertex buffer.


src/corell/src/mapping.rs
```diff
@@ -0,0 +1,93 @@
...
+        match *self {
+            AccessOverlap => "The requested mapping access overlaps with another"
+        }
+    }
+}
+
+/// Mapping reader
+pub struct Reader<'a, R: Resources, T: 'a + Copy> {
+    slice: &'a [T],
+    #[allow(dead_code)] mapping: R::Mapping,
```
Bastacyclop: is this needed ?

src/corell/src/mapping.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    fn deref(&self) -> &[T] { self.slice }
+}
+
+/// Mapping writer.
+/// Currently is not possible to make write-only slice so while it is technically possible
+/// to read from Writer, it will lead to an undefined behavior. Please do not read from it.
+pub struct Writer<'a, R: Resources, T: 'a + Copy> {
+    slice: &'a mut [T],
+    #[allow(dead_code)] mapping: R::Mapping,
```
Bastacyclop: (same)

src/corell/src/mapping.rs
msiglreith: Do you mean the `#[allow(dead_code)]`? If so, I copied it from the `core` reader mapping implementation.
`mapping` itself is mainly used for handling automatic unmapping of the underlying memory once the reader/writer drops.

Bastacyclop: I mean `mapping` itself, you don't seem to use it. In the current `core` it is only used for tracking and there is no unmapping on drop.

msiglreith: This one should be called once `Reader`\`Writer` goes out of scope.
https://github.com/gfx-rs/gfx/pull/1218/files#diff-2aa247755db5577c6193abe291447143R38

Bastacyclop: I see, so if I understand correctly the usage of `Reader` and `Writer` changes a lot. In `core`, the buffers are (conceptually) persistently mapped at creation time (not really when the backend does not support persistent mapping) and then you use `Reader`/`Writer` when you want to take exclusive access to use the mapping. In `corell`, `Reader`/`Writer` would represent the actual mapping, maybe the names could highlight that difference ?
For example:
`factory.map_readable() -> mapping::Readable`
`factory.map_writable() -> mapping::Writable`

src/backend/vulkanll/src/command.rs
```diff
@@ -60,16 +60,50 @@ impl CommandBuffer {
...
+                    y: region.image_offset.y,
+                    z: region.image_offset.z,
+                },
+                image_extent: vk::Extent3D {
+                    width:  region.image_extent.width,
+                    height: region.image_extent.height,
+                    depth:  region.image_extent.depth,
+                },
+            }
+        }).collect::<Vec<_>>();
```
kvark: we'd like to avoid heap allocations in the future, but this is totally fine at this stage of corell

msiglreith: There are quite a few of them unfortunately.
What would be an alternative? SmallVec? Could switch to it for later occurrences.

msiglreith:
Fixed up the memory heap property flags a bit as I wasn't happy with them. They were kinda wrong and now matches vulkan a bit better but also stays compatible with d3d12 cpu page properties.

Pulling the current `core` mapping into `corell` would be redundant work as the plan is to merge it into `core`. Keeping to simplified model for now and will take care of it once we pull the corresponding pieces.

kvark:
@Bastacyclop please merge when ready
@homu delegate=Bastacyclop

Bastacyclop:
@homu r+

Issue #1198, opened 2017-02-24
GL backend performance: Tracking state is fast, resetting state is slow


icefoxen:
At least on NVidia GPU's, it seems that calling `glUseProgram(0)` and then setting to another program is slow.  This makes `Encoder::draw()` slow, because `gfx_backend_gl::CommandBuffer::bind_pipeline_state()` is slow, because `Command::BindProgram(pso.program)` is slow.  Caching the current `pso` program and not bothering to issue a bind command when it doesn't change makes the `performance` example about 2x faster...  WHEN you remove the `self.reset_state()` call from `gfx_backend_gl::no_fence_submit()`.  It also opens the door 

I assume that not resetting the state on each submission call will break some things, so this isn't a general-purpose solution without some re-architecting to make the state reset unnecessary.  But, if someone is interested in OpenGL performance, this is a pretty good place to start.

Characteristic timings from the `performance` example, on an NVidia 660 graphics card:

No changes:
```
total time:		51.97ms
	create list:	6.94ms
	submit:		44.18ms
	gpu wait:	0.85ms

```

No redundant program binds:
```
total time:		24.20ms
	create list:	7.22ms
	submit:		16.28ms
	gpu wait:	0.70ms

```

Only setting necessary values in `gfx_backend_gl::CommandBuffer::bind_pipeline_state()`:
```
total time:		21.47ms
	create list:	6.84ms
	submit:		13.86ms
	gpu wait:	0.77ms
```

Additionally, because of the multi-layered design of gfx, effort should be made to avoid as much work as possible at as high an abstraction level as possible.  It's much faster to never submit a command than it is to submit a command, check it, and then decide it's redundant.

kvark:
Thanks @icefoxen !

First of all, the general approach for gfx-rs state caching is to do it on a backend level, i.e. in the `CommandBuffer` implementation. Our backends already do some work:
```rust
        if self.cache.rasterizer != pso.rasterizer {
            self.cache.rasterizer = pso.rasterizer;
            self.parser.parse(Command::SetRasterizer(pso.rasterizer));
}
```

Of course, it's not nearly complete, and we can be much more aggressive in optimizing this.

Caching things on the `Encoder` level may save us a few lines in the backends, however it is arguable: backends have more information and thus can do a better job at caching what they need. At the end of the day, the `Encoder`-level caching may become redundant and just add overhead, if we consider the backends doing the optimization ideally. Thus, I'd propose focusing on the backends directly.

[...]

PR #1220, opened 2017-04-05
Attempts at making the GL backend cache more state

icefoxen:
Would resolve (a large part of) issue #1198 .

It's quite a brute-force solution, and I am deeply uncertain that it tracks everything correctly, but I tried to err on the side of safety, allowing redundant state-setting calls rather than getting rid of them.  It is also incomplete: there is some stuff, such as stencil buffers, that it doesn't try to track (because I don't really know how).  However, it appears to work; I don't *see* any differences in any of the examples, and apitrace output for the performance example is about equivalent to the gl example (which is good!).

Still, please sanity-check this, and help make sure it works with fancy stuff like multiple PSO and vertex type definitions.  I'll be doing the same if/when I have time.


icefoxen:
Dammit, hang on...

msiglreith:
Looks like your branch is based on the `master` branch but the PR is opened for merging into `0.14`

Bastacyclop:
@icefoxen It seems that your commits are all before the "random" ones, maybe you could try to do `git reset   67ee0eb` and see if `git log` is ok then ?

icefoxen:
@msiglreith It was SUPPOSED to be based on 0.14... let me try again.  x_X

icefoxen:
Closed in favor of #1221

PR #1221, opened 2017-04-05
Adding more caching to GL backend.

icefoxen:
Like #1220 except based on the right starting point.  Sorry.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -205,82 +205,78 @@ impl Cache {
         }
     }
 
-    fn push(&self, buf: &mut Vec<Command>, command: Command) {
-        buf.push(command)
-    }
-
-    fn bind_program(&mut self, buf: &mut Vec<Command>, program: Program) {
+    fn bind_program(&mut self, program: Program) -> Option<Command> {
         if program == self.program {
-            return;
+            return None;
         }
         self.program = program;
-        buf.push(Command::BindProgram(program));
+        Some(Command::BindProgram(program))
     }
 
-    fn bind_constant_buffer(&mut self, buf: &mut Vec<Command>, constant_buffer: c::pso::ConstantBufferParam<Resources>) {
+    fn bind_constant_buffer(&mut self, constant_buffer: c::pso::ConstantBufferParam<Resources>) -> Option<Command> {
         if self.constant_buffer == Some(constant_buffer) {
-            return;
+            return None;
         }
         self.constant_buffer = Some(constant_buffer);
-        buf.push(Command::BindConstantBuffer(constant_buffer));
+        Some(Command::BindConstantBuffer(constant_buffer))
     }
 
-    fn bind_resource_view(&mut self, buf: &mut Vec<Command>, resource_view: c::pso::ResourceViewParam<Resources>) {
+    fn bind_resource_view(&mut self, resource_view: c::pso::ResourceViewParam<Resources>) -> Option<Command> {
         if self.resource_view == Some(resource_view) {
-            return;
+            return None;
         }
         self.resource_view = Some(resource_view);
-        buf.push(Command::BindResourceView(resource_view));
+        Some(Command::BindResourceView(resource_view))
     }
 
-    fn bind_index(&mut self, buf: &mut Vec<Command>, buffer: Buffer, itype: c::IndexType) {
+    fn bind_index(&mut self, buffer: Buffer, itype: c::IndexType) -> Option<Command> {
         if self.index == buffer && itype == self.index_type {
-            return;
+            return None;
         }
         self.index_type = itype;
         self.index = buffer;
-        buf.push(Command::BindIndex(buffer));
+        Some(Command::BindIndex(buffer))
     }
 
-    fn bind_framebuffer(&mut self, buf: &mut Vec<Command>, access: Access, fb: FrameBuffer) {
+    fn bind_framebuffer(&mut self, access: Access, fb: FrameBuffer) -> Option<Command> {
         if self.framebuffer == Some((access, fb)) {
-            return;
+            return None;
         }
         self.framebuffer = Some((access, fb));
-        buf.push(Command::BindFrameBuffer(access, fb));
+        Some(Command::BindFrameBuffer(access, fb))
     }
 
-    fn set_rasterizer(&mut self, buf: &mut Vec<Command>, rasterizer: s::Rasterizer) {
+    fn set_rasterizer(&mut self, rasterizer: s::Rasterizer) -> Option<Command> {
         if self.rasterizer == Some(rasterizer) {
-            return;
+            return None;
         }
         self.rasterizer = Some(rasterizer);
-        buf.push(Command::SetRasterizer(rasterizer));
+        Some(Command::SetRasterizer(rasterizer))
     }
 
-    fn set_viewport(&mut self, buf: &mut Vec<Command>, rect: Rect) {
+    fn set_viewport(&mut self, rect: Rect) -> Option<Command> {
         if self.viewport == Some(rect) {
-            return;
+            return None;
         }
         self.viewport = Some(rect);
-        buf.push(Command::SetViewport(rect));
+        Some(Command::SetViewport(rect))
     }
 
-    fn set_scissor(&mut self, buf: &mut Vec<Command>, rect: Option<Rect>) {
+    fn set_scissor(&mut self, rect: Option<Rect>) -> Option<Command> {
         if self.scissor_test == rect {
-            return;
+            return None;
         }
         self.scissor_test = rect;
-        buf.push(Command::SetScissor(rect));
+        Some(Command::SetScissor(rect))
     }
-    fn set_depth_state(&mut self, buf: &mut Vec<Command>, depth: Option<s::Depth>) {
+    fn set_depth_state(&mut self, depth: Option<s::Depth>) -> Option<Command> {
         if self.depth_state == depth {
-            return;
+            return None;
         }
         self.depth_state = depth;
-        buf.push(Command::SetDepthState(depth));
+        Some(Command::SetDepthState(depth))
     }
-    fn set_stencil_state(&mut self, buf: &mut Vec<Command>, option_stencil: Option<s::Stencil>, stencils: (Stencil, Stencil), cullface: s:: CullFace) {
+    fn set_stencil_state(&mut self, option_stencil: Option<s::Stencil>, stencils: (Stencil, Stencil), cullface: s:: CullFace) -> Option<Command> {
         // This is a little more complex 'cause if option_stencil
         // is None the stencil state is disabled, it it's Some
         // then it's enabled and parameters are set.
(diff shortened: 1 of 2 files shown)


src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+    fn bind_program(&mut self, buf: &mut Vec<Command>, program: Program) {
+        if program == self.program {
+            return;
+        }
+        self.program = program;
+        buf.push(Command::BindProgram(program));
+    }
+
+    fn bind_constant_buffer(&mut self, buf: &mut Vec<Command>, constant_buffer: c::pso::ConstantBufferParam<Resources>) {
+        if let Some(cb) = self.constant_buffer {
```
kvark: nit: could be simplified to `if self.constant_buffer == Some(constant_buffer)`
if it complains about the type being moved and such, you can compare by reference `if self.constant_buffer.as_ref() == Some(&constant_buffer)`

src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+    }),
     Command::SetScissor(None),
     Command::SetDepthState(None),
     Command::SetStencilState(None, (0, 0), s::CullFace::Nothing),
     Command::SetBlendState(0, COLOR_DEFAULT),
     Command::SetBlendState(1, COLOR_DEFAULT),
     Command::SetBlendState(2, COLOR_DEFAULT),
     Command::SetBlendState(3, COLOR_DEFAULT),
-    Command::SetBlendColor([0f32; 4]),
+    Command::SetBlendColor([0f32; 4])
```
kvark: is that a `rustfmt` decision to remove the comma?

src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+            if cb == constant_buffer {
+                return;
+            }
         }
+        self.constant_buffer = Some(constant_buffer);
+        buf.push(Command::BindConstantBuffer(constant_buffer));
     }
+
+    fn bind_resource_view(&mut self, buf: &mut Vec<Command>, resource_view: c::pso::ResourceViewParam<Resources>) {
+        if let Some(rv) = self.resource_view {
```
kvark: similarly, this can be rewritten

src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+        if self.index == buffer && itype == self.index_type {
+            return;
+        }
+        self.index_type = itype;
+        self.index = buffer;
+        buf.push(Command::BindIndex(buffer));
+    }
+
+    fn bind_framebuffer(&mut self, buf: &mut Vec<Command>, access: Access, fb: FrameBuffer) {
+
```
kvark: extra line?

src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+        buf.push(Command::SetDepthState(depth));
+    }
+    fn set_stencil_state(&mut self, buf: &mut Vec<Command>, option_stencil: Option<s::Stencil>, stencils: (Stencil, Stencil), cullface: s:: CullFace) {
+        // This is a little more complex 'cause if option_stencil
+        // is None the stencil state is disabled, it it's Some
+        // then it's enabled and parameters are set.
+        // That's actually bad because it makes it impossible
+        // to completely remove all redundant calls if the
+        // stencil is enabled;
+        // we'll be re-enabling it over and over.
```
kvark: agreed

src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+        // For now though, we just don't handle it.
+        buf.push(Command::SetStencilState(option_stencil, stencils, cullface));
+    }
+    fn set_blend_state(&mut self, buf: &mut Vec<Command>, color_slot: c::ColorSlot, color: s::Color) {
+        if let Some(bs) = self.blend_state {
```
kvark: nit: simplify

src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+            blend_state: None,
+            blend_color: None,
+            viewport: None,
+            rasterizer: None,
+            framebuffer: None,
+            index: 0,
+        }
+    }
+
+    fn push(&self, buf: &mut Vec<Command>, command: Command) {
```
kvark: what is the purpose of this method?

src/backend/gl/src/command.rs
```diff
@@ -127,47 +134,188 @@ pub const RESET: [Command; 14] = [
...
+        buf.push(command)
+    }
+
+    fn bind_program(&mut self, buf: &mut Vec<Command>, program: Program) {
```
kvark: I feel like this is a bit verbose on the `Cache` side. How about the `bind_*` methods returning `Option<Command>`, and the caller just doing something like `self.buf.extend(self.cache.bind_program(program))`? I wonder if we could chain those options to call `extend` for multiple results at once.

src/backend/gl/src/command.rs
```diff
@@ -209,160 +358,199 @@ impl command::Buffer<Resources> for CommandBuffer {
...
-                    self.active_attribs |= 1<<i;
-                },
-                (_, None) if self.active_attribs & (1<<i) != 0 => {
-                    self.buf.push(Command::UnbindAttribute(i as c::AttributeSlot));
-                    self.active_attribs ^= 1<<i;
-                },
-                (_, None) => (),
+        if self.cache.current_vbs == Some(vbs) {
+            return
+        } else {
```
kvark: no need for `else` branch here

src/backend/gl/src/command.rs
```diff
@@ -209,160 +358,199 @@ impl command::Buffer<Resources> for CommandBuffer {
...
+                   dst: Buffer,
+                   src_offset_bytes: usize,
+                   dst_offset_bytes: usize,
                    size_bytes: usize) {
-        self.buf.push(Command::CopyBuffer(src, dst,
-                                          src_offset_bytes as gl::types::GLintptr,
-                                          dst_offset_bytes as gl::types::GLintptr,
-                                          size_bytes as gl::types::GLsizeiptr));
+        self.cache.push(&mut self.buf,
+                                     Command::CopyBuffer(src,
```
kvark: indentation is broken here

src/backend/gl/src/command.rs
```diff
@@ -209,160 +358,199 @@ impl command::Buffer<Resources> for CommandBuffer {
...
                       img: c::texture::RawImageInfo) {
         let ptr = self.data.add(data);
         match ntex {
-            NewTexture::Texture(t) =>
-                self.buf.push(Command::UpdateTexture(t, kind, face, ptr, img)),
-            NewTexture::Surface(s) =>
-                error!("GL: unable to update the contents of a Surface({})", s),
+            NewTexture::Texture(t) => {
+                self.cache.push(&mut self.buf,
+                                             Command::UpdateTexture(t, kind, face, ptr, img))
```
kvark: indentation here

src/backend/gl/src/command.rs
```diff
@@ -371,22 +559,33 @@ impl command::Buffer<Resources> for CommandBuffer {
...
 
-    fn call_draw(&mut self, start: c::VertexCount,
-                 count: c::VertexCount, instances: Option<command::InstanceParams>) {
-        self.buf.push(Command::Draw(self.cache.primitive, start, count, instances));
+    fn call_draw(&mut self,
+                 start: c::VertexCount,
+                 count: c::VertexCount,
+                 instances: Option<command::InstanceParams>) {
+        self.cache.push(&mut self.buf,
+                                     Command::Draw(self.cache.primitive, start, count, instances));
```
kvark: indentation

src/backend/gl/src/command.rs
```diff
@@ -371,22 +559,33 @@ impl command::Buffer<Resources> for CommandBuffer {
...
                          instances: Option<command::InstanceParams>) {
         let (offset, gl_index) = match self.cache.index_type {
             c::IndexType::U16 => (start * 2u32, gl::UNSIGNED_SHORT),
             c::IndexType::U32 => (start * 4u32, gl::UNSIGNED_INT),
         };
-        self.buf.push(Command::DrawIndexed(self.cache.primitive,
-            gl_index, RawOffset(offset as *const gl::types::GLvoid), count, base, instances));
+        self.cache
+            .push(&mut self.buf,
+                         Command::DrawIndexed(self.cache.primitive,
```
kvark: indentation

icefoxen:
What is the policy on rustfmt?  You don't appear to be using it as far as I can tell, so a lot of the indentation stuff is just emacs's auto-formatting.  If there's some specific formatting guide or such I should follow that'd make life easier.

src/backend/gl/src/command.rs
icefoxen: Basically a layer of indirection.  It used to have a bit more functionality but that ended up getting pulled out into other functions.  Now it's basically there to say "the cache modifies the command buffer whenever necessary, nothing else needs to touch it directly."

src/backend/gl/src/command.rs
icefoxen: A bit verbose, yes.  Like I've said before, this is a pretty brute-force attempt...

kvark:
@icefoxen I haven't looked into `rustfmt` and there is no official policy. My rough understanding is that it's not ready for us yet, so no need to vigorously format anything :)

icefoxen:
All right.  I was more looking for ways to vigorously *not* reformat everything, since I usually code with rustfmt running automatically every time I save a file and it makes a terrible mess of projects that don't use it.

icefoxen:
Okay, I think I've fixed all the formatting bugs and simplified where indicated.  The only thing I haven't done is gotten rid of `Cache::push()` and altered the cache functions to return `Option` instead of maybe pushing the commands themselves.  I feel like the latter can work better because it gives the cache more control over what goes on, and makes the caller have to do less work.  All the caller has to do is call the `Cache`'s function to add a command.

I think that ideally every command should be added to the buffer via a method on `Cache`; if the cache can do nothing, it just pushes the command directly, but it would open the door to having more state be tracked by the `Cache` without having to change anything about how the caller operates.

kvark:
@icefoxen thanks for updating it!!!

> I feel like the latter can work better because it gives the cache more control over what goes on, and makes the caller have to do less work. All the caller has to do is call the Cache's function to add a command.
> I think that ideally every command should be added to the buffer via a method on Cache; if the cache can do nothing, it just pushes the command directly, but it would open the door to having more state be tracked by the Cache without having to change anything about how the caller operates.

Don't worry about the caller. There is only one, so it doesn't make much sense to try to shove more logic into the `Cache`, making it sort of an opaque layer that fills up the command buffer (AFAIK, that's what your PR is doing). Thing is, `CommandBuffer` IS this layer already, so what `Cache` needs to be is just a structure holding some state, possibly with a few helpers if that make sense.

Does this sound reasonable?

icefoxen:
Fair enough!  It's hard sometimes, you look at stuff and say "hey, a layer of abstraction could go here!" and why not put one in?

This will probably take me a few days to get around to doing, but I'll refactor it so the functions just return an option.

icefoxen:
Refactor complete; it IS nicer, thank you for the suggestion.

kvark:
Thanks @icefoxen ! Looks great now! 👍 
@homu r+

src/core/src/pso.rs
```diff
@@ -208,7 +208,7 @@ impl Descriptor {
 }
 
 /// A complete set of vertex buffers to be used for vertex import in PSO.
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
kvark: are these `core` changes required for your PR?
We need to bump the version numbers and (if core is changed) change the dependency of core to the new version.

src/backend/gl/src/command.rs
```diff
@@ -141,33 +153,158 @@ pub const RESET: [Command; 14] = [
...
+            blend_color: None,
+            viewport: None,
+            rasterizer: None,
+            framebuffer: None,
+            index: 0,
+        }
+    }
+
+    fn bind_program(&mut self, program: Program) -> Option<Command> {
+        if program == self.program {
```
kvark: nit: I think a simple `if/else` for the case of small branches should be preferred over `return`

src/core/src/pso.rs
icefoxen: I'm afraid so because it needs to be able to compare the existing `VertexBufferSet` against the one it has cached.  It can probably be worked around by just storing the *contents* of the `VertexBufferSet`, but that seems a little redundant.

kvark: Ok, I see. Please bump both version numbers then and make sure the new core is requested

kvark:
Ok, we can bump the versions ourselves and do the refactor, if needed.

kvark:
@icefoxen your code has sailed in https://crates.io/crates/gfx_device_gl/0.13.1

PR #1223, opened 2017-04-07
[ll] vulkan: Update to latest ash and fix graphics pipeline creation

msiglreith:
Once a error occurs during pipeline creation we now recover all piplines, which were sucessfully built, instead of returning an error for all of them.


kvark:
Great, thank you!
@homu r+

Issue #1216, opened 2017-03-29
[ll] Resource destruction


msiglreith:
*Part of the lower level core abstraction rewrite #1102*

## Motivation
We want to minimize API overhead for our core library. At the moment we track resources via reference counted handles in `core`. In contrast, `corell` API directly works with the raw `R::*` structs. Some backends (e.g Vulkan) have explicit `vkDestroy*` functions, which require an handle to the device.
d3d11/12 solves this by internal reference counting.

## 1st Approach (Explicit destruction)
In order to destruct resources the API will expose `destroy_*` functions, which need to be explicitly called.

**Pros**: Doesn't require to store the device handle in each resource.
**Cons**: Requires the user to explicitly take track of the it's resources and avoid duplicated destruction.

## 2nd Approach (Lifetime wrapper)
We use wrapper structs `ResourceObject<'a, R>` for the resources `R::*` to have an associated lifetime. Resources now can't outlive the creating factory.

**Pros**: We only need to store the 'bare' device handle in the resources for destruction.
**Cons**: Requires all Factory methods to use `&self` instead of `&mut self`, else we won't be able to reborrow the factory for creating new resources.

## 3rd Approach (RAII)
The most intuitive way for user would be RAII, resources will be destructed once the handle drops.

**Pros**: Safe and easy to use
**Cons**: Requires to keep track of the underlying device handle via reference counting to avoid the resource outliving the device.

cc @kvark

kvark:
I'm strongly in favour of 1st approach - explicit destruction. One of the core benefits of the new core API that I'm looking to get is zero overhead for the most demanding applications.

Not sure how the 2nd will solve the problem, perhaps you need to explain it more.

msiglreith:
Sounds reasonable.
Regarding the 2nd approach I tried to create a small code snippet to show the underlying idea:
https://is.gd/c6zdFR
Motivation for this approach was to avoid the reference counting costs and try to couple resource lifetime to the factory/device lifetime.

minecrawler:
I do not like the first one, because it is Rust we use and Rust takes care of memory stuff for us. It just feels like a step back to C/C++ (though gfx is abstracting over something which was initially defined with those languages in mind). In my opinion, the third approach would be the correct one to guarantee lifetimes and memory safety on the user side.

Bastacyclop:
[...]

PR #1227, opened 2017-04-13
[ll] Add explicit resource destruction

msiglreith:
Add explicit resource destruction as proposed in #1216.
Implemented the new factory methods for the vulkan backend.

I went with a resource consuming interface (takes ownership of the resource) as I liked the concept of the factory giving ownership to the resource on creationg and taking it away on destruction.
```Rust
fn destroy_*(&mut self, R::Resource);
```

### DIFF
--- a/examples/trianglell/main.rs
+++ b/examples/trianglell/main.rs
@@ -308,6 +308,28 @@ fn main() {
         // present frame
         swap_chain.present();
     }
+
+    // cleanup!
+    factory.destroy_shader_lib(shader_lib);
+    factory.destroy_pipeline_layout(pipeline_layout);
+    factory.destroy_renderpass(render_pass);
+    factory.destroy_heap(heap);
+    factory.destroy_heap(image_heap);
+    factory.destroy_heap(image_upload_heap);
+    factory.destroy_buffer(vertex_buffer);
+    factory.destroy_buffer(image_upload_buffer);
+    factory.destroy_image(image_logo);
+    for pipeline in pipelines {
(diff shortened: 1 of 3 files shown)


examples/trianglell/main.rs
```diff
@@ -308,6 +308,28 @@ fn main() {
...
+    factory.destroy_shader_lib(shader_lib);
+    factory.destroy_pipeline_layout(pipeline_layout);
+    factory.destroy_renderpass(render_pass);
+    factory.destroy_heap(heap);
+    factory.destroy_heap(image_heap);
+    factory.destroy_heap(image_upload_heap);
+    factory.destroy_buffer(vertex_buffer);
+    factory.destroy_buffer(image_upload_buffer);
+    factory.destroy_image(image_logo);
+    for pipeline in pipelines.into_iter() {
```
kvark: nit: I don't think `into_iter()` is needed?

msiglreith: Should be required as we need ownership of the pipeline and `iter()/iter_mut()` iterate only over references.
Couldn't make the compiler happy without using `into_iter()` either.

kvark: `into_iter` is equivalent of not calling anything at all, e.g.
```rust
for pipeline in pipelines {...}
```

msiglreith: That's pretty nice!
Fixed and squashed.

kvark:
@homu r+

kvark:
@homu retry

PR #1228, opened 2017-04-17
Change changelog order into new to old and add links for mentioned issues

ama0:
Since all changelogs go new to old and it's common to link the issues, I thought why not :)
And @kvark mentioned on reddit he would welcome a PR in this regard.


kvark:
Beautiful, thank you!
@homu r+

Issue #1219, opened 2017-04-05
[ll] [RFC] Resource binding


msiglreith:
_Part of the lower level core abstraction rewrite #1102_

Resource Binding
====================

Pipelines require access to resources in the shader stages. Each pipeline has an interface/layout for the different resources.
Resources can be constant buffers, textures, samplers, push constants, ...

# Overview (bindless APIs)
## Vulkan
In Vulkan resources are grouped together as `DescriptorSet`s, which define the layout of the pipeline. A descriptor is a lightweight handle to an buffer or image view. Views allow shaders to operate on the underlying objects and give access to a subrange of their memory with associated meta information.
To give a shader access to a resource, a view for the resource needs to be created first and a descriptor set matching the pipeline layout.
Descriptors are then written into the set, containing the view and the descriptor type (access type for the shader).

References:
https://developer.nvidia.com/vulkan-shader-resource-binding
https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html

## Directx 12
Pipelines have a root signature, consisting of references to descriptor tables (range in a descriptor heap), push constants and inline resources. In d3d12 we can differ between CPU and GPU descriptors. CPU descriptors are returned on view creation. GPU descriptors have to be allocated on a descriptor heap and CPU descriptors can be written into GPU descriptors. There can only be bound 1 Srv/Cbv/Uav descriptor heap and 1 sampler heap at the same time.

References:
https://msdn.microsoft.com/en-us/library/windows/desktop/dn899206(v=vs.85).aspx
https://software.intel.com/en-us/articles/introduction-to-resource-binding-in-microsoft-directx-12

# Draft proposal
The descriptor table approach from d3d12 is powerful and quite straightforward, but hard and inefficient to emulate via descriptor sets. Therefore we will have to follow vulkan's concept of descriptor sets here.

Both APIs share the view concept, which we can build upon. d3d12 stores views inside CPU descriptor heaps (to be symmetric to GPU descriptors..), but we can hide them internally in the `Factory`.
Our current factory API in core fits the model quite well and we can take it over.

Root Signatures are bound to 64*4bytes at maximum. (Descriptor tables: 4bytes, Root Constants: 4bytes, Root Descriptor: 4bytes, static sampler: 0)

Still, an open topic is descriptor creation and organization in heaps/pool.
[...]

PR #1230, opened 2017-04-18
[ll] Resource binding

msiglreith:
First pieces of an implementation of https://github.com/gfx-rs/gfx/issues/1219

- [x] Resource destruction api
- [x] View creation for SRV/UAV/CBV (edit: no CBVs so far and SRV/UAV only for images)
- [x] Descriptor update/copy (edit: no copy, update only for a few descriptor types)

edit:
- [x] Sampler creation/destruction
- [x] Pipeline barrier support for vulkan (image barriers only)

Groundwork done for future work on the resource binding. For UAV, in particular, compute shader support would be appealing. Therefore, I would like to cut the PR at this point (already became so large again..sry!).

Example upgraded from triangles to textured sprites!
trianglell


kvark:
@homu r+

PR #1231, opened 2017-04-18
gfx_device_gl: Add a new field allowing to use an externally managed framebuffer object as the display target.

manuel-rhdt:
Adds a new public field `display_fb` to the `CommandBuffer` structure in the `gfx_device_gl` backend. If not changed by the user it will have the value 0. This allows to use an existing framebuffer object as the drawing destination instead of the Framebuffer with name 0. Useful on iOS/tvOS where no default frame buffer exists and drawing can only occur on system provided renderbuffers.

I expect this be used with the `create_main_targets_raw` function to create the default color and depth targets. These will automatically use the framebuffer object given in the `display_fb` field.

If you are happy with this I would like to do more pull requests for better iOS support (e.g. only use supported OpenGL ES 3.0 features).

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -186,6 +186,7 @@ pub struct CommandBuffer {
     pub buf: Vec<Command>,
     pub data: DataBuffer,
     fbo: FrameBuffer,
+    pub display_fb: FrameBuffer,
     cache: Cache,
     active_attribs: usize,
 }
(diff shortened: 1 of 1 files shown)


kvark:
Thank you! This is some neatly detailed and documented piece of code, a pleasure to read ;)

I think the whole `CommandBuffer` struct of GL backend is quite unsafe w.r.t. exposed public fields, so that's something to note (and potentially address later on).

> If you are happy with this I would like to do more pull requests for better iOS support (e.g. only use supported OpenGL ES 3.0 features).

Please do! 👏 👍 

@homu r+

PR #1234, opened 2017-04-21
Add optional serialization support, version bumps, formatting fixes

### LINKED ISSUE #1229: Serde integration
Example:
```rust
#[cfg_attrib(feature="serialize", derive(Serialize, Deserialize))]
```
cc @ebkalderon @msiglreith

ebkalderon:
> Note: This is the season finale for our two-part series. Prepare for `gfx` 0.15! :tada:

### Added
* Add `serialize` feature to `gfx_app` and `gfx_core`, courtesy of `draw_state` 0.7.0 (fixes #1229)
* Added API docs to `Copy{Buffer,BufferTexture,TextureBuffer}Result` types

### Changed
* Minor reformatting and alphabetical ordering of `#[derive]` statements
* Switch one manual `Eq` implementation to `#[derive]` instead
* Encoder now uses `#[derive(Debug)]`
* Bumped minor version numbers of affected crates (`gfx` is now up to 0.15.0)
* Updated several third-party dependencies
* Build against Rust 1.16.0 on AppVeyor

### Fixed
* Expose `CopyError` and `Copy{Buffer,BufferTexture,TextureBuffer}Result` types from `gfx`

CC @kvark

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,7 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cargo build --features vulkan; else cargo build; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cargo build --features metal; else cargo build; fi
   - cargo test --all
+  - cargo test -p gfx -p gfx_core --features serialize
   - cargo test -p gfx_window_sdl
   - cargo test -p gfx_device_gl
   - cargo test -p gfx_window_glutin
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx_app"
-version = "0.4.0"
+version = "0.5.0"
 description = "GFX example application framework"
 homepage = "https://github.com/gfx-rs/gfx"
 keywords = ["graphics", "gamedev"]
@@ -13,27 +13,27 @@ default = []
 sdl = ["gfx_window_sdl"]
 vulkan = ["gfx_device_vulkan", "gfx_device_vulkanll", "gfx_window_vulkan"]
 metal = ["gfx_device_metal", "gfx_window_metal"]
+serialize = ["gfx/serialize", "gfx_core/serialize"]
 unstable = []
-
 
 [lib]
 name = "gfx_app"
 
 [dependencies]
 log = "0.3"
-env_logger = "0.3"
+env_logger = "0.4"
 glutin = "0.7.1"
 winit = "0.5.1"
-gfx_core = { path = "src/core", version = "0.6" }
+gfx_core = { path = "src/core", version = "0.7" }
 gfx_corell = { path = "src/corell", version = "0.1" }
-gfx = { path = "src/render", version = "0.14" }
+gfx = { path = "src/render", version = "0.15" }
 gfx_macros = { path = "src/macros", version = "0.2" }
-gfx_device_gl = { path = "src/backend/gl", version = "0.13" }
-gfx_window_glutin = { path = "src/window/glutin", version = "0.14" }
+gfx_device_gl = { path = "src/backend/gl", version = "0.14" }
+gfx_window_glutin = { path = "src/window/glutin", version = "0.15" }
 
 [dependencies.gfx_device_vulkan]
 path = "src/backend/vulkan"
-version = "0.1"
+version = "0.2"
 optional = true
 
 [dependencies.gfx_device_vulkanll]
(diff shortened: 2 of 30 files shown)


.travis.yml
```diff
@@ -51,7 +51,7 @@ script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export LD_LIBRARY_PATH=$LIBRARY_PATH ; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then cargo build --features vulkan; else cargo build; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cargo build --features metal; else cargo build; fi
-  - cargo test --all
+  - cargo test --all --features serialize
```
torkleyy: Maybe test with and without the feature?

appveyor.yml
```diff
@@ -15,6 +15,6 @@ install:
   - rustc -vV
   - cargo -vV
 build_script:
-  - cargo build --features vulkan
+  - cargo build --features "vulkan serialize"
 test_script:
-  - cargo test --all --features vulkan
+  - cargo test --all --features "vulkan serialize"
```
torkleyy: Same as in `.travis.yml`

.travis.yml
ebkalderon: Given that the feature only makes the types derive from `Serialize `/`Deserialize` and nothing more, it just has to compile correctly to pass. But if you really want, I can duplicate the line to re-run all tests without it.

torkleyy: Just a thought, you can leave it as is. We could also add the extra test later, should we implement anything more advanced.

ebkalderon: Good point! I think I'll add it in.

ebkalderon:
@kvark Alright, I've updated the PR and it's ready for review!

src/core/src/handle.rs
```diff
@@ -32,15 +32,13 @@ impl<R: Resources> ops::Deref for RawBuffer<R> {
...
 /// Type-safe buffer handle
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq)]
 pub struct Buffer<R: Resources, T>(RawBuffer<R>, PhantomData<T>);
 
 impl<R: Resources, T> cmp::PartialEq for Buffer<R, T> {
     fn eq(&self, other: &Self) -> bool { self.0.eq(&other.0) }
 }
 
-impl<R: Resources, T> cmp::Eq for Buffer<R, T> {}
```
kvark: we shouldn't be removing this line. The difference with `derive` is that we don't require `T: Eq` here, which is essential

Cargo.toml
```diff
@@ -13,27 +13,27 @@ default = []
...
 name = "gfx_app"
 
 [dependencies]
 log = "0.3"
-env_logger = "0.3"
-glutin = "0.7.1"
-winit = "0.5.1"
-gfx_core = { path = "src/core", version = "0.6" }
+env_logger = "0.4"
+glutin = "0.7"
```
kvark: you shouldn't strip the patch version from here - it is likely for a reason

src/render/Cargo.toml
```diff
@@ -30,9 +30,10 @@ name = "gfx"
 path = "src/lib.rs"
 
 [features]
+serialize = ["draw_state/serialize"]
```
kvark: shouldn't it also say `gfx_core/serialize` here?

src/window/dxgi/Cargo.toml
```diff
@@ -29,9 +29,9 @@ name = "gfx_window_dxgi"
...
 log = "0.3"
 kernel32-sys = "0.2"
-user32-sys = "0.1"
+user32-sys = "0.2"
 dxguid-sys = "0.2"
 winapi = "0.2"
-winit = "0.5.2"
-gfx_core = { path = "../../core", version = "0.6" }
-gfx_device_dx11 = { path = "../../backend/dx11", version = "0.5" }
+winit = "0.5"
```
kvark: again, please don't strip this patch requirement

Cargo.toml
```diff
@@ -13,27 +13,27 @@ default = []
 sdl = ["gfx_window_sdl"]
 vulkan = ["gfx_device_vulkan", "gfx_device_vulkanll", "gfx_window_vulkan"]
 metal = ["gfx_device_metal", "gfx_window_metal"]
+serialize = ["gfx_core/serialize"]
```
kvark: should probably also say `gfx/serialize`

src/core/src/handle.rs
ebkalderon: Oh, I see. I'll revert this change, then.

Cargo.toml
ebkalderon: According to semantic versioning, the patch number is for backwards-compatible bug fixes only. By removing the patch version from the `Cargo.toml`, we can get these bug fixes whenever they come out, so we don't have to manually increment the version number over and over again.

Cargo.toml
ebkalderon: Correct, I'll fix that.

src/render/Cargo.toml
ebkalderon: Correct, I'll fix that.

Cargo.toml
kvark: I think you are misreading the spec slightly. For 0.x.y, AFAIK, `x` is considered a "major" and `y` as "minor". Adding features just requires you to bump the patch version then, as long as it's backwards compatible. If something **depends** on new features, it needs to have the patch version specified for that minimal version (what we have). `cargo update` would still use the newer patch versions, if available, but it will guarantee we'll not get lower ones.

kvark:
I think we are ready. Thank you!
@homu r+

ebkalderon:
@kvark So the PR is complete! Commenting here, as requested.

PR #1236, opened 2017-04-22
Reexport SubmissionError and SubmissionResult

ebkalderon:
### Fixed

- Reexport `SubmissionError` and `SubmissionResult` types.

CC @kvark @msiglreith


kvark:
@homu r+

PR #1238, opened 2017-04-25
Derive Hash and Eq for Slice and IndexBuffer

ebkalderon:
### Changed

* Derive `Hash` and `Eq` traits for `Slice` and `IndexBuffer`.


kvark:
@homu r+

PR #1239, opened 2017-04-25
[ll] Add semaphore and fence sync primitives

msiglreith:
First steps regarding synchronization support based on the vulkan API.
It should be possible to emulate Semaphore and Fence in dx12 using ID3D12Fence.

Additionally, extended the queue submit to allow fence and semaphore control, and batch multiple submits.

Following PR will address destruction, reset and further control options for image frame acquisition. Aiming to remove all remaining validation errors left in the trianglell example.


kvark:
👍 
@homu r+

Issue #1213, opened 2017-03-14
Link to arcsynthesis on OpenGL state machine is dead


burtonageo:
The [link](http://www.arcsynthesis.org/gltut/Basics/Intro%20What%20is%20OpenGL.html#d0e887) on the Readme about binding and unbinding objects leads to a parked domain website.

ama0:
I think you can close this issue, @kvark. #1240 fixed it.

kvark:
Thanks @ama0 !

PR #1240, opened 2017-04-25
Update README.md to v0.15, fix broken link (#1213)

ama0:
- Update build dependency information from "gfx = 0.14" to "gfx = 0.15"
- Fixes "Link to arcsynthesis on OpenGL state machine is dead" (#1213)
  - Since the old link is dead and I can't find it in the Wayback Machine Archives my guess is just based on the old URL (gltut[orial]/Basics/Intro What is OpenGL)


kvark:
Thank you!
@homu r+

PR #1242, opened 2017-04-26
Allow cgmath types in constant buffers

ebkalderon:
### Added
* Allow `cgmath::{Deg, Rad, Matrix2, Matrix3, Matrix4, Point2, Point3}` types to be used in constant buffers, disabled by default under the `math` feature gate.
* Pull in `cgmath` 0.13 as an optional dependency.

### Changed
* Convert examples from `cgmath` version 0.7 to 0.13.


kvark:
This PR looks great! Very clean and straightforward positive.
Just one thing that I'd like to be changed is the feature name. "math" is ambiguous. We should have either "cgmath", if possible, or something with "cgmath" in it, e.g. "cgmath-types".

ebkalderon:
@kvark Thanks for the feedback! I have revised the feature name to `"cgmath-types"`, as requested.

kvark:
Sweet, thank you!
@homu r+

PR #1243, opened 2017-04-27
[ll] More on sync and minimize errors

msiglreith:
Improve synchronization for submission and frame acquisition.
Fix issue on vulkan backend that queue priority vec only lifes on the stack and can get overwritten.

Current synchronization strategy in the example is pretty bad as it blocks forces a pipeline flush each frame.

Compute shaders next!

### DIFF
--- a/examples/trianglell/main.rs
+++ b/examples/trianglell/main.rs
@@ -23,9 +23,9 @@ extern crate winit;
 extern crate image;
 
 use gfx_corell::{buffer, command, format, pass, pso, shade, state, target, 
-    Device, CommandPool, GraphicsCommandPool,
+    Device, CommandPool, GraphicsCommandPool, CommandQueue,
     GraphicsCommandBuffer, ProcessingCommandBuffer, TransferCommandBuffer, PrimaryCommandBuffer,
-    Primitive, Instance, Adapter, Surface, SwapChain, QueueFamily, QueueSubmit, Factory, SubPass};
+    Primitive, Instance, Adapter, Surface, SwapChain, QueueFamily, QueueSubmit, Factory, SubPass, FrameSync};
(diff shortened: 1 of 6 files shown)


examples/trianglell/main.rs
```diff
@@ -375,9 +376,10 @@ fn main() {
             }
         }
 
-        graphics_pool.reset();
+        general_queues[0].wait_idle(); // SLOW!
```
kvark: why are we waiting here?

msiglreith:
I changed them to `&mut` to ensure that multiple threads can't access them at the same time.

kvark:
@msiglreith ok, I suppose it's reasonable. And those `waits` are temporary.
@homu r+

examples/trianglell/main.rs
msiglreith: Hm, not totally sure if this can be removed. The pipeline barrier enforces that the data has been uploaded to the gpu?
Most/All examples I've seen so far wait after the texture upload, wondering if this is just over-conservative.
(I'm still not feeling 100% safe when it comes to synchronization/ordering guarantees...)

kvark:
@msiglreith 

>> general_queues[0].wait_idle(); // SLOW!
> Hm, not totally sure if this can be removed. The pipeline barrier enforces that the data has been uploaded to the gpu?

I might be missing something here, but it seems to me that waiting for a queue to become idle each frame is a no-go right away. Instead, the CPU should be arbitraging synchronization of GPU with semaphores.

So, the queue that did execute data transfer should signal a semaphore on completion. Currently it does not: `signal_semaphores: &[],`. Then this semaphore is to be used as a wait parameter for your first frame queue execution.

PR #1244, opened 2017-04-29
[ll] Add Constant Buffer Views

msiglreith:
Constant buffer views (Cbv) are a concept introduced by dx12, which uses views for everything now.
Cbvs needs to be explicitly created in contrast to e.g vertex buffer views.
In vulkan these are represented as the underlying buffer and a range.


kvark:
nice!
@homu r+

Issue #1052, opened 2016-09-19
[pre-ll] Compute shaders support


kvark:
Compute is supported on GL4, DX11, Metal, and Vulkan. Can be very useful for some general computations, especially on the hardware with async compute support.

Optimally, we'd want to have the compute stuff existing separately from the graphics:
- `ComputeCommandBuffer` trait versus `GraphicsCommandBuffer`. Correspondingly, separate `Computer` struct (or, `ComputeEncoder`, which doesn't sound nearly as pretty).
- `ComputeDevice` trait versus `GraphicsDevice`. The backends are free to either provide structs that implement both of these traits, or any of them.

dowoncha:
I am interested in implementing Compute for Vulkan support. Any tips on starting? Do the compute traits already exist?

kvark:
@dowoncha we are in the process rewriting the whole core layer, and I think @msiglreith is putting the compute support from the start. If you want to try adding the compute to the old layer, you are still welcome to, as long as you are accepting that the results might be partially discarded. Also, the current Vulkan backend is not yet functional, but that's (again) something that the ongoing rewrite seems to address.

msiglreith:
The `corell` rewrite already comes with compute support as kvark mentioned. The `ComputeCommandBuffer` trait already exists and is implemented for vulkan iirc.
If you are interested in working at the vulkanll backend (yay!): starting point would be the implementation of `create_compute_pipelines` (https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkanll/src/factory.rs#L499-L501) and corresponding compute shader loading.

Due to the lack of descriptor set/pipeline layout support it probably won't be really usable atm besides running very simple shaders.

You can reach me on gitter if you face some issues with the low level API or rough edges of the current design iteration (very experimental!).

msiglreith:
I'm going to implement it for corell in the next days if noone else is working on it atm.

kvark:
Pretty much out of scope now, closing.

PR #1245, opened 2017-04-29
[ll] Compute pipeline

msiglreith:
Add compute pipeline creation
cc #1052 

(corell now should have support for most of the core features from vulkan/dx12.)


kvark:
@homu r+

PR #1246, opened 2017-04-30
Add ProgramError::Geometry enum variant

ebkalderon:
### Added
* Add `ProgramError::Geometry` enum variant.
* Add `create_shader_set_geometry` method to `FactoryExt`.

### Changed
* Clarified `create_shader_set_tessellation` doc comments.

### Fixed
* Corrected usage of "pixel shader" to the correct shader types in `ProgramError` doc comments.

CC @kvark @msiglreith

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -141,7 +141,7 @@ gfx_gl = "0.3"
 rand = "0.3"
 genmesh = "0.4"
 noise = "0.1"
-image = "0.12.3"
+image = "0.13"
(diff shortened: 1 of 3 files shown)


kvark:
Thanks, looks great!

Note that this is technically a breaking change.

@homu r+

torkleyy:
I guess the patch should be included in todays nightly.

kvark:
@homu r+

PR #1252, opened 2017-05-03
[ll] Core update [01/..]

msiglreith:
_[01/..] Part of a longer project for merging corell into core_

Merge device setup and surface creation part from corell.
Adapater, Surface, Swapchain, etc. will be gradually updated to match corell.
Device trait will be replaced by a struct containing queues, heaps and a resource factory.

### DIFF
--- a/src/core/src/lib.rs
+++ b/src/core/src/lib.rs
@@ -37,7 +37,9 @@ use std::hash::Hash;
 use std::any::Any;
 
 pub use draw_state::{state, target};
+pub use self::command::CommandBuffer;
 pub use self::factory::Factory;
+pub use self::queue::{GeneralQueue, GraphicsQueue, ComputeQueue, TransferQueue};
 
 pub mod buffer;
 pub mod command;
@@ -48,6 +50,7 @@ pub mod handle;
 pub mod mapping;
 pub mod memory;
 pub mod pso;
+pub mod queue;
 pub mod shade;
 pub mod texture;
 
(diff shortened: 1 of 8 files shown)


src/core/src/lib.rs
```diff
@@ -285,20 +331,43 @@ pub trait Device: Sized {
     fn cleanup(&mut self);
 }
 
+///
```
msiglreith: Documentation added in a later commit (not included in this PR)

src/core/src/lib.rs
```diff
@@ -336,25 +405,52 @@ pub trait QueueFamily: 'static {
...
     type Window;
 
     /// Create a new surface from a native window.
     fn from_window(window: &Self::Window) -> Self;
 
     /// Create a new swapchain from the current surface with an associated present queue.
-    fn build_swapchain<T: format::RenderFormat>(&self, present_queue: &Self::CommandQueue)
-        -> Self::SwapChain;
+    fn build_swapchain<S, T>(&self, present_queue: &Self::CommandQueue) -> S
+        where S: SwapChain, T: format::RenderFormat;
```
msiglreith: Surface got some further changes when I implemented the GL and vulkan backend.
The current Swapchain generic got removed again.
Surfaces will be moved towards the current window crates and one of the two entry points for creation (besides Adapter).
So the user is supposed to create a surface from a window (usually winit::Window) and further create an associated swapchain. The swapchain has to be recreated on resize or other events.

src/core/src/lib.rs
```diff
@@ -285,20 +331,43 @@ pub trait Device: Sized {
     fn cleanup(&mut self);
 }
 
+///
+pub struct Device_<R: Resources, F: Factory<R>, Q: CommandQueue> {
```
kvark: that's a neat approach

PR #1253, opened 2017-05-04
[ll] Core update (vulkan) [02/..]

msiglreith:
Merge new parts into the existing vulkan backend from vulkanll.
Minor adaptions due to removal of the `Instance` trait and changes to `Surface`.
Including addition of queue types for the core.

Removed most of the current vulkan backend as it wasn't working atm.


kvark:
👍 we aren't going to miss the non-functional backend code

PR #1255, opened 2017-05-08
[ll] Core update (GL) [03/..]

msiglreith:
- Implement device creation part for GL
- Move SwapChain to the window crate because OpenGL requires the glutin window for swapping the buffers on present

Command buffer submission on GL is not optimal atm due to cloning the internal data and cmd buffers.
I'll rebase the `ll` branch after the PR on the latest master to avoid conflicts for the future merge.

Next task would be the d3d11 and d3d12 backend, updating the `gfx_app` and some smaller changes to `render`. Hopefully we can reach an mergable state then.
Unfortunately I don't have access to an OS X machine so I can't take care of the Metal backend.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -52,6 +52,7 @@ pub struct DataPointer {
     size: u32,
 }
 
+#[derive(Clone)]
 pub struct DataBuffer(Vec<u8>);
 impl DataBuffer {
     /// Create a new empty data buffer.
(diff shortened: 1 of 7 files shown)


src/backend/gl/src/command.rs
```diff
@@ -628,3 +634,37 @@ impl command::Buffer<Resources> for CommandBuffer {
                       instances));
     }
 }
+
+impl c::CommandBuffer for CommandBuffer {
+    type SubmitInfo = SubmitInfo;
+    unsafe fn end(&mut self) -> SubmitInfo {
```
kvark: can we have `fn end(&mut self) -> &SubmitInfo`, perhaps? or anything else that borrows it instead of creating

msiglreith: hm this could work! Changing `SubmitInfo` to vk::CommandBuffer on vulkan would minimize code changes.

PR #1256, opened 2017-05-08
Run-time specified instance rate

kvark:
The new `ToInstanceRate` trait allows the same PSO declaration and shader to be used with both instanced and non-instanced attribute. There appears to be a real use case for it - sometimes we know the color (for example) on a per-object basis, sometimes it's per-vertex.

### DIFF
--- a/src/render/src/pso/buffer.rs
+++ b/src/render/src/pso/buffer.rs
@@ -36,20 +36,46 @@ type AttributeSlotSet = usize;
 /// Service struct to simplify the implementations of `VertexBuffer` and `InstanceBuffer`.
 #[derive(Derivative)]
 #[derivative(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct VertexBufferCommon<T, I>(
+pub struct VertexBufferCommon<T, I=InstanceRate>(
     RawVertexBuffer,
     #[derivative(Hash = "ignore", PartialEq = "ignore")]
     PhantomData<(T, I)>
 );
 
+/// Helper trait for `VertexBufferCommon` to support variable instance rate.
+pub trait ToInstanceRate {
+    /// The associated init type for PSO component.
+    type Init;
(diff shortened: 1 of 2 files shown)


src/render/src/pso/buffer.rs
```diff
@@ -36,20 +36,46 @@ type AttributeSlotSet = usize;
...
 #[derive(Derivative)]
 #[derivative(Clone, Debug, Eq, Hash, PartialEq)]
-pub struct VertexBufferCommon<T, I>(
+pub struct VertexBufferCommon<T, I=InstanceRate>(
     RawVertexBuffer,
     #[derivative(Hash = "ignore", PartialEq = "ignore")]
     PhantomData<(T, I)>
 );
 
+/// Helpe trait for `VertexBufferCommon` to support variable instance rate.
```
msiglreith: typo: `Helpe`/`Helper`

kvark:
Thanks @msiglreith ! Should be fixed now (in both PRs)

kvark:
Also note that I feel a bit uneasy about versioning here. Technically, `VertexBufferCommon` is a public PSO component, and I'm changing its API. So there might be users that get a breakage, but I don't expect anyone to actually use it directly just yet.

msiglreith:
I see, a bit unfortunate. I would prefer to ship it with the next 0.16 release, except if there is some urgent need for a new release with this changes.
Thanks for addressing the mentioned nits! I think we can merge it into master once tests passes.

kvark:
@msiglreith yeah, hmm. Problem is - I need it for `vange-rs` (hence the PR in the first place), and we don't have 0.16 on the roadmap even.
I guess I'll use the override for the mean time, should be fine since I don't expect many people trying to build fresh vange-rs on their machines anyway, at this point at least.

kvark:
@homu r+

PR #1257, opened 2017-05-08
[v0.15] Run-time specified instance rate

kvark:
Same as #1256 , but for our v0.15 release


kvark:
Closing as not backwards-compatible, as discussed in #1256

Issue #1199, opened 2017-02-25
Window creation examples for SDL and GLFW


kvark:
> Basic examples can be found at https://github.com/icefoxen/gfx/tree/master/examples/triangle-glfw and https://github.com/icefoxen/gfx/tree/master/examples/triangle-sdl

cc @icefoxen

icefoxen:
See, now I'm going to have to eat my words about "keeping those up to date is trivial", since they're not up to date...  XD

kellpossible:
and an example with glutin too?

alexheretic:
#1271 provides an sdl window example, run with `(cd src/window/sdl && cargo run --example window)`

kvark:
Fixed by #1258 and #1271
Thanks everyone!

PR #1258, opened 2017-05-08
Triangle example using GLFW

Haggus:
GLFW triangle example from #1199


kvark:
The problem here is minimizing the code duplication with existing code. This PR copies a lot of stuff (code, screenshot, shaders, etc). I wonder if we can have everything as optional code paths inside the same triangle example.

Haggus:
Oh, I thought that was the point. To make the triangle example use glfw, and make it as close to glutin implementation as possible :)

By "optional code paths", do you mean having triangle examples share the same assets?

kvark:
@Haggus I'm sorry, there is a lot of discussion context between me and @icefoxen that got lost and not expressed in the original issue. I believe there is a solution that would require minimum code/maintenance and would be helpful for anyone choosing GLFW/SDL. This solution is to:
  1. Have the examples to be *inside* the corresponding window backend folders (`src/window/glfw/examples/`)
  2. Don't draw anything, just clear the screen. We can have a comment referring to the triangle example for the actual drawing code, it's not going to be any different.

How does this sound to you?

Haggus:
@kvark that sounds good. I don't mind changing it, so if this is how you guys want it, then it's OK with me. I'll submit changes as soon as I can!

Haggus:
@kvark let me know what you think about these changes.

Also, now that I think about it, this example should probably be called "GLFW Example".

src/window/glfw/Cargo.toml
```diff
@@ -28,6 +28,11 @@ documentation = "https://docs.rs/gfx_window_glfw"
 name = "gfx_window_glfw"
 
 [dependencies]
-glfw = "0.13"
+glfw = "0.14"
+gfx = "0.15"
```
kvark: this should go to `[dev-dependencies]`

src/window/glfw/Cargo.toml
```diff
@@ -28,6 +28,11 @@ documentation = "https://docs.rs/gfx_window_glfw"
...
 gfx_core = { path = "../../core", version = "0.7" }
 gfx_device_gl = { path = "../../backend/gl", version = "0.14" }
+
+[[example]]
+name = "window"
+path = "examples/window/main.rs"
```
kvark: let's simplify it by placing the example as `examples/window.rs`, so that you don't even need this `path=` line here

src/window/glfw/examples/window/main.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    let (mut window, events) = glfw.create_window(1024, 768, "Triangle example", glfw::WindowMode::Windowed)
+        .expect("Failed to create GLFW window.");
+
+    window.set_key_polling(true);
+    window.set_close_polling(true);
+    window.make_current();
+    glfw.set_error_callback(glfw::FAIL_ON_ERRORS);
+    let (_, _, _, _) = gfx_window_glfw::init(&mut window);
+
```
kvark: let's add a `//Note: actual drawing code is no different from the triangle example, or any other`.

src/window/glfw/examples/window/main.rs
```diff
@@ -0,0 +1,53 @@
...
+pub fn main() {
+    let mut glfw = glfw::init(glfw::FAIL_ON_ERRORS)
+        .ok()
+        .expect("Failed to initialize GLFW");
+
+    glfw.window_hint(glfw::WindowHint::ContextVersion(3, 2));
+    glfw.window_hint(glfw::WindowHint::OpenGlForwardCompat(true));
+    glfw.window_hint(glfw::WindowHint::OpenGlProfile(glfw::OpenGlProfileHint::Core));
+
+    let (mut window, events) = glfw.create_window(1024, 768, "Triangle example", glfw::WindowMode::Windowed)
```
kvark: let's change the title here

Haggus:
Should be good now. Let me know if you'd like any other changes! :)

Cargo.toml
```diff
@@ -145,4 +145,4 @@ image = "0.13"
 winit = "0.6"
 
 [target.x86_64-unknown-linux-gnu.dev-dependencies]
-glfw = "0.12"
+glfw = "0.14"
```
kvark: Actually, I don't know why it's even here? gfx_app doesn't use glfw or SDL anyway.
Let's remove it?

src/window/glfw/Cargo.toml
```diff
@@ -28,6 +28,9 @@ documentation = "https://docs.rs/gfx_window_glfw"
 name = "gfx_window_glfw"
 
 [dependencies]
-glfw = "0.13"
+glfw = "0.14"
```
kvark: this needs a version bump for this crate

src/window/glfw/examples/window.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright 2015 The Gfx-rs Developers.
```
kvark: we are 2 years in the future, yay!

src/window/glfw/Cargo.toml
```diff
@@ -28,6 +28,9 @@ documentation = "https://docs.rs/gfx_window_glfw"
...
 gfx_core = { path = "../../core", version = "0.7" }
 gfx_device_gl = { path = "../../backend/gl", version = "0.14" }
+
+[dev-dependencies]
+gfx = "0.15"
```
kvark: actually, you need to provide both the local path and the version, similar to the dependencies above

src/window/glfw/Cargo.toml
Haggus: Isn't 0.14 the latest version?

kvark: I mean, we should bump the version of `gfx_window_glfw` to `0.15`, given that dependency change

src/window/glfw/Cargo.toml
```diff
@@ -28,6 +28,9 @@ documentation = "https://docs.rs/gfx_window_glfw"
 name = "gfx_window_glfw"
 
 [dependencies]
-glfw = "0.13"
+glfw = "0.14"
 gfx_core = { path = "../../core", version = "0.7" }
 gfx_device_gl = { path = "../../backend/gl", version = "0.14" }
+
+[dev-dependencies]
+gfx = { path = "../../../src/render", version = "0.15" }
```
kvark: I'd assume you can go with "../../render". It still doesn't explain why travis fails to compile. Does it compile for you?

Haggus: if I do "../../render", it doesn't compile for me. With "src" it does without issues.

Haggus: I did "cargo clean", it recompiled and works.

kvark:
There is one small fix needed in the root Cargo.toml now to point to gfx_window_glfw 0.15

Haggus:
Should be good now :)

kvark:
Thanks! 🤞

kvark:
This looks like a bug :( I can repro it on my machine. `cargo test -p gfx_window_glfw` doesn't use `dev-dependencies` section for some reason. I propose to move it into the regular dependencies, but add a `//TODO` note mentioning the problem. Also, the path should really just be "../../render".

Haggus:
Here is the bug: https://github.com/rust-lang/cargo/issues/860

kvark:
Thanks @Haggus for finding the issue! I commented on there.

Issue #1204, opened 2017-03-05
Cube example renders differently with OpenGL vs Metal on MacOS


burtonageo:
When rendering with Metal:

<img width="1136" alt="gfx_cube_example_metal" src="https://cloud.githubusercontent.com/assets/1732920/23586045/35972738-0185-11e7-993e-eeeb78d8c56c.png">

When rendering with OpenGL:

<img width="1136" alt="gfx_cube_example_opengl" src="https://cloud.githubusercontent.com/assets/1732920/23586053/462c7ca6-0185-11e7-98ca-51c562369a83.png">

<img width="582" alt="my_specs" src="https://cloud.githubusercontent.com/assets/1732920/23586062/7ce89e96-0185-11e7-8103-1a91e77e9d67.png">

kvark:
Thanks for filing the issue! I wonder if we are confused here about `ARGB2101010` format. Seems rather unusual for a back buffer. Do you have HDR-enabled screen?

eldyer:
Same problem here on my mac with NVIDIA GeForce GT 650M. Pixel depth is 32-Bit Color (ARGB8888).

JohnColanduoni:
It looks like there are two issues (possibly with a single cause):

1. The single texel "texture" is being interpreted as BGRA event though gfx::format::Rgba8 is specified. The cube shows up as blue if you swap the blue and red in the cube's main.rs.
2. There's something funny going on with the gamma. Tweaking the format mapping in the metal backend by sending (Rgba8) to BGRAUnorm_sRGB at map.rs:241 (instead of BGRAUnorm) makes the background the right color, but the cube is still too dark.

My guess is this has something to do with the backend silently converting Rgba8 to Bgra8 for render targets, which we do to isolate the user from the fact that CAMetalLayers (i.e. the window) only supports BGRAUnorm, BGRAUnorm_sRGB and RGBA16Float. However Vulkan already requires the same juggling so we should probably stop doing the remapping.

GabrielMajeri:
I think this was fixed by #1259 and can be closed.

PR #1259, opened 2017-05-09
Fix Metal backend's color handling

JohnColanduoni:
Two issues were causing distored colors when using the Metal backend:

1. RGBA8 was mapped to BGRA8 when rendering, to accomodate CAMetalLayer's requirement that it only use BGRA ordering for 8-bit backbuffer formats. This however created issues when handling textures, so instead we expose this difference to the user. It should only require them to choose a backbuffer format based on the backend used, which is already needed for Vulkan.

2. CAMetalLayer appears to interpret the buffer as sRGB whether you want it to or not. To accomodate this we use the backbuffer color format (B8_G8_R8_A8, Srgb) in gfx_app when using Metal, so writes to the backbuffer are transformed properly. This may actually be an error on our side, because we do some hacky things with regards to backbuffer management, but until those are fixed this will at least produce correct results.

Addresses issue #1204

### DIFF
--- a/src/backend/metal/src/map.rs
+++ b/src/backend/metal/src/map.rs
@@ -237,16 +237,13 @@ pub fn map_format(format: Format, is_target: bool) -> Option<MTLPixelFormat> {
             Int   => RGBA8Sint,
             Uint  => RGBA8Uint,
             Inorm => RGBA8Snorm,
-            Unorm => if is_target {
-                         BGRA8Unorm
-                     } else {
-                         RGBA8Unorm
-                     },
-            Srgb  => if is_target {
-                         BGRA8Unorm_sRGB
-                     } else {
-                         RGBA8Unorm_sRGB
-                     },
+            Unorm => RGBA8Unorm,
+            Srgb  => RGBA8Unorm_sRGB,
+            _ => return None,
+        },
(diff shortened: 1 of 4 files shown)


src/backend/metal/src/map.rs
```diff
@@ -237,16 +237,13 @@ pub fn map_format(format: Format, is_target: bool) -> Option<MTLPixelFormat> {
             Int   => RGBA8Sint,
             Uint  => RGBA8Uint,
             Inorm => RGBA8Snorm,
-            Unorm => if is_target {
-                         BGRA8Unorm
```
kvark: there surely were reasons for this, even if it's definitely cleaner in the PR.
@fkaa do you recall why BGRA was used here?

fkaa: IIRC it's as mentioned in the PR description, the backbuffer requires BGRA and gfx_app used RGBA everywhere and I did the conversion (stupidly) in that conversion function (`map_format`). This looks much cleaner (and most importantly more functional!)

fkaa:
LGTM! 👍

kvark:
Thanks for taking a look, @fkaa !
@homu r+

PR #1260, opened 2017-05-09
Use full namespaces in macro-generated fns which return ::std types

burtonageo:
This avoids clash when you define your own `Result` type. e.g.

```rust
gfx_defines! {
     vertex Vertex { /* ... */ }
     pipeline pipe { /* ... */ }
}

pub type Result<T> = ::std::result::Result<T, MyError>;
```

Will cause a compile failure:

```
error[E0244]: wrong number of type arguments: expected 1, found 2
  --> src/file.rs:12:1
   |
12 | gfx_defines! {
   | ^ expected 1 type argument
   |
   = note: this error originates in a macro outside of the current crate

```


kvark:
Seems reasonable, although I don't recall seeing this elsewhere.
Thank you!
@homu r+

kvark:
@homu retry

PR #1262, opened 2017-05-10
Implement gfx::pso::buffer::RawGlobal

ebkalderon:
### Added

* Implement `gfx::pso::buffer::RawGlobal`.
* Add global link compatibility error into `gfx::pso::InitError` instead of `()`.

### Fixed

* Perform format compatibility checking for typed `Global`s.
* Initial implementation of `TODO` described in `match_attribute()` function.

CC @kvark @msiglreith

### DIFF
--- a/src/render/src/pso/buffer.rs
+++ b/src/render/src/pso/buffer.rs
@@ -104,21 +104,58 @@ pub struct ConstantBuffer<T: Structure<shade::ConstFormat>>(
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct RawConstantBuffer(Option<(Usage, ConstantBufferSlot)>);
 
-/// Global (uniform) constant component. Describes a free-standing value passed into
-/// the shader, which is not enclosed into any constant buffer. Deprecated in DX10 and higher.
+/// Global (uniform) constant component. Describes a free-standing value passed
+/// into the shader, which is not enclosed into any constant buffer.
 ///
 /// - init: `&str` = name of the constant
 /// - data: `T` = value
 #[derive(Derivative)]
 #[derivative(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Global<T: ToUniform>(
-    Option<shade::Location>,
+    RawGlobal,
     #[derivative(Hash = "ignore", PartialEq = "ignore")]
     PhantomData<T>
 );
 
-fn match_attribute(_: &shade::AttributeVar, _: Format) -> bool {
-    true //TODO
+/// Raw global (uniform) constant component. Describes a free-standing value
+/// passed into the shader, which is not enclosed in any constant buffer.
+///
+/// - init: `&str` = name of the constant
+/// - data: `UniformValue` = value
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct RawGlobal(Option<shade::Location>);
+
+fn match_attribute(attr: &shade::AttributeVar, fmt: Format) -> bool {
+    use core::shade::{BaseType, ContainerType};
+    use core::format::ChannelType;
+
+    // TODO: Fill in `_` with compatible `core::format::SurfaceType` variants.
+
+    match (attr.base_type, fmt.1) {
+        (BaseType::I32, ChannelType::Int) |
+        (BaseType::F32, ChannelType::Float) |
+        (BaseType::F32, ChannelType::Inorm) |
+        (BaseType::F32, ChannelType::Srgb) |
+        (BaseType::F32, ChannelType::Unorm) |
+        (BaseType::U32, ChannelType::Uint) => match (attr.container, fmt.0) {
+                (ContainerType::Single, _) |
+                (ContainerType::Vector(2), _) |
+                (ContainerType::Vector(3), _) |
+                (ContainerType::Vector(4), _) => true,
+                _ => false,
+        },
+        (BaseType::F64, ChannelType::Float) |
+        (BaseType::F64, ChannelType::Inorm) |
+        (BaseType::F64, ChannelType::Srgb) |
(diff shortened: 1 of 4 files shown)


src/render/src/pso/buffer.rs
```diff
@@ -104,19 +104,29 @@ pub struct ConstantBuffer<T: Structure<shade::ConstFormat>>(
...
 #[derive(Derivative)]
 #[derivative(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct Global<T: ToUniform>(
-    Option<shade::Location>,
+    RawGlobal,
     #[derivative(Hash = "ignore", PartialEq = "ignore")]
     PhantomData<T>
 );
 
+/// Raw global (uniform) constant component. Describes a free-standing value
```
kvark: let's explain a little on the difference between this and a regular `Global`.
Also, let's remove the "Deprecated in ..." comment both here and above, since you are at it ;) thanks!

src/render/src/pso/buffer.rs
```diff
@@ -295,33 +305,58 @@ impl<R: Resources> DataBind<R> for RawConstantBuffer {
...
     fn new() -> Self {
-        Global(None, PhantomData)
+        Global(RawGlobal::new(), PhantomData)
+    }
+    fn is_active(&self) -> bool {
+        self.0.is_active()
+    }
+    fn link_global_constant(&mut self, var: &shade::ConstVar, init: &Self::Init) ->
+                            Option<Result<(), shade::UniformValue>> {
+        self.0.link_global_constant(var, init)
```
kvark: let's add `//TODO: check `T` compatibility with `var`

src/render/src/pso/buffer.rs
```diff
@@ -295,33 +305,58 @@ impl<R: Resources> DataBind<R> for RawConstantBuffer {
...
+        RawGlobal(None)
     }
     fn is_active(&self) -> bool {
         self.0.is_some()
     }
     fn link_global_constant(&mut self, var: &shade::ConstVar, init: &Self::Init) ->
                             Option<Result<(), shade::UniformValue>> {
-        if &var.name == *init {
+        if var.name.as_str() == *init {
             //if match_constant(var, ())
```
kvark: let's remove this comment

src/render/src/pso/buffer.rs
ebkalderon: No need for a comment; I can add this functionality myself!

kvark: I believe it should be here in `link_global_constant` return type, so that you can pass it through. The old error type `shade::UniformValue` didn't make much sense and wasn't used anyway.
Btw, you should first check for the underlying compatibility of the raw stuff (`self.0.link...`) and then (literally, using `and_then`) check for compatibility.

src/render/src/pso/buffer.rs
```diff
@@ -104,19 +104,29 @@ pub struct ConstantBuffer<T: Structure<shade::ConstFormat>>(
...
+/// Raw global (uniform) constant component. Describes a free-standing value
+/// passed into the shader, which is not enclosed in any constant buffer.
+/// Deprecated in DX10 and higher.
+///
+/// - init: `&str` = name of the constant
+/// - data: `UniformValue` = value
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct RawGlobal(Option<shade::Location>);
+
 fn match_attribute(_: &shade::AttributeVar, _: Format) -> bool {
```
ebkalderon: Shouldn't this be `ConstFormat` and not `Format`?

kvark:
Hah, this function is for vertex attributes, thus not applicable here

ebkalderon:
@kvark Yes, I realized that. I'm just implementing it as well for good measure, so we can eliminate the `TODO`.

src/render/src/pso/buffer.rs
```diff
@@ -104,21 +104,50 @@ pub struct ConstantBuffer<T: Structure<shade::ConstFormat>>(
...
-    true //TODO
+/// Raw global (uniform) constant component. Describes a free-standing value
+/// passed into the shader, which is not enclosed in any constant buffer.
+///
+/// - init: `&str` = name of the constant
+/// - data: `UniformValue` = value
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct RawGlobal(Option<shade::Location>);
+
+fn match_attribute(attr: &shade::AttributeVar, fmt: Format) -> bool {
```
ebkalderon: After implementing this function, all the examples still compile and run just fine. My implementation is incomplete, however.

1. What should I fill in for the `_` in each entry?
2. Should I handle `ContainerType::Matrix` variants as well?

kvark: 1. let's leave this for the follow-ups
  2. nope, those can't be attributes
  3. yes

src/render/src/pso/buffer.rs
```diff
@@ -104,21 +104,50 @@ pub struct ConstantBuffer<T: Structure<shade::ConstFormat>>(
...
+    use core::shade::{BaseType, ContainerType};
+    use core::format::{SurfaceType, ChannelType};
+
+    match ((attr.base_type, attr.container), (fmt.0, fmt.1)) {
+        ((BaseType::I32, ContainerType::Single), (_, ChannelType::Int)) => true,
+        ((BaseType::F32, ContainerType::Single), (_, ChannelType::Float)) => true,
```
kvark: `F32` is also compatible with `Snorm`, `Unorm`, and `Srgb` channel types

ebkalderon: Okay. Is `F64` also compatible with the same channel types as `F32`?

ebkalderon:
Done! Hopefully this PR is good to go.

kvark:
@homu r+

PR #1263, opened 2017-05-10
Add resizing support for gfx_window_sdl

felipeagc:
Added some functions similar to those in gfx_window_glutin so you can resize the viewport according to the window size.


kvark:
Nice, thank you!

kvark:
Published in https://crates.io/crates/gfx_window_sdl/0.7.0

PR #1266, opened 2017-05-13
Only enable PROGRAM_POINT_SIZE for desktop

torkleyy:
In OpenGL ES, it is enabled by default
and the flag is not supported.


kvark:
@homu r+

PR #1271, opened 2017-05-15
Add sdl window example, update to 0.30

alexheretic:
*gfx_window_sdl* requires sdl2 = "0.29" stopping usage of later sdl2 versions or requiring gfx to be updated manually for each sdl2 minor release. 

Since this lib is used as a helper for sdl and isn't particularly likely to break with newer minor versions, it is more helpful to declare a minimum sdl2 version `sdl2 = ">= 0.29, < 1"`. This allows a top-level consumer to declare, for example, sdl2 = "0" and have it work properly (ie currently fetching v0.30.0).

In the case newer versions cause issues, it is still clear which version(s) can be tried by viewing the minimum declared.

I've also added an sdl version of the triangle example, this adds a little compile assurance to gfx_window_sdl when running `cargo test`. I added the gfx dev-depencency sdl version to be `sdl2 = "0"` meaning the example will always test the latest 0 minor version, but this could be changed to `sdl2 = "0.30"` if greater build consistency is required. (However, I think its better as 0, as it matches the >= 0.29 better)

What are this projects views on forward compatibility of this sort?

### DIFF
--- a/src/window/sdl/Cargo.toml
+++ b/src/window/sdl/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_window_sdl"
-version = "0.6.0"
+version = "0.7.0"
 description = "SDL2 window for gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
@@ -28,6 +28,14 @@ name = "gfx_window_sdl"
 
 [dependencies]
 log = "0.3"
-sdl2 = "0.29"
+sdl2 = "0.30"
 gfx_core = { path = "../../core", version = "0.7" }
 gfx_device_gl = { path = "../../backend/gl", version = "0.14" }
+# Currently there is an issue with cargo and dev-dependencies:
+# https://github.com/rust-lang/cargo/issues/860
+# TODO: move gfx to [dev-dependencies] once it gets resolved.
+gfx = { path = "../../render", version = "*" }
+
+[[example]]
+name = "window"
+path = "examples/window.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,12 +59,12 @@ optional = true
 
 [dependencies.gfx_window_sdl]
 path = "src/window/sdl"
-version = "0.6"
+version = "0.7"
 optional = true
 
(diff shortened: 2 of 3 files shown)


kvark:
I'd like us to avoid supporting another triangle example in the code base. Please check out #1258 for the way GLFW is tested.

As for `sdl2 = ">= 0.29, < 1"` condition, I think it goes against semver. We can't automatically use newer minor versions of non-stable dependencies.

> gfx_window_sdl requires sdl2 = "0.29" stopping usage of later sdl2 versions or requiring gfx to be updated manually for each sdl2 minor release.

You don't have to use the newest release at every point in time, unless there are features you require. I suppose you can pick a version that `gfx_window_sdl` links to and work with that unless there is a blocker, in which case you can make a simple PR (even through Github web interface) to our crate and use a local override in the mean time. I don't think this situation is going to occur often. Let me know if your experience tell otherwise, or there are particular reasons to stick with the latest version of sdl all the time.

src/window/sdl/Cargo.toml
```diff
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_window_sdl"
-version = "0.6.0"
+version = "0.6.1"
```
kvark: I believe this should be "0.7.0", since the `sdl2` dependency gets exposed in the API

alexheretic:
Thanks for taking a look. In that case I may as well convert this into updating sdl2 0.29 -> 0.30, and adding the window example.

To be honest though, I'll probably end up using my own fork of *gfx_window_sdl* when the next sdl update I fancy appears. But I can understand you wanting stronger consistency from the gfx point of view.

kvark:
Thanks @alexheretic ! Code looks good, waiting for CI to test it.

kvark:
Please squash the commits and fix CI errors to proceed.

alexheretic:
Yeah just noticed the dev-dependencies thing, worked around it in the same way as glfw. For the commits can't you just click squash & merge ala

torkleyy:
@alexheretic We're using homu, which does not support this.

alexheretic:
@torkleyy Ok I've done it manually.

kvark:
Thank you!
@homu r+

PR #1272, opened 2017-05-16
[ll] Command Pools and application [05/..]

msiglreith:
A larger PR therefore a bit more explanation!

**Goal**: Merge device setup and 'frontend' (adapters, queues, swapchains, etc.) parts into current core. This doesn't include changes to the factory and command buffers.

- **Deprecate the `Device` trait**, the functionality will be split into the queues and the factory. (not everything ported so far)
- Adding **command pools**, which are necessary for creating command buffers/encoders. Current plan for gfx_app is that the user still will receive an `Encoder` (renamed to `GraphicsEncoder`) when implementing `render`. So keeping the changes to the examples to a minimum for the moment.
- We can rethink the application design in the future if there are some compute examples or multi-queue stuff etc. Also due to the unification of the window creation (Surface + Swapchain) we could think of some helper method for easier setup and maybe **merge application somehow into the render**.
- In contrast to OpenGL the other APIs usually expose **multiple rendertargets for rendering to the backbuffers** instead of just one framebuffer. The current d3d11 backend simply uses one backbuffer iirc, but that's poor workaround. We probably need some changes to the examples which select the correct RTV depending on the current frame. It could be either done by the user or the application only sends the current frame color RTV.

### DIFF
--- a/src/backend/vulkan/src/pool.rs
+++ b/src/backend/vulkan/src/pool.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::BorrowMut;
 use std::ops::DerefMut;
 use std::ptr;
 use std::sync::Arc;
@@ -77,9 +78,11 @@ macro_rules! impl_pool {
         }
 
         impl pool::$pool for $pool {
-            fn from_queue<Q>(queue: &mut Q, capacity: usize) -> $pool
-                where Q: Into<$queue<CommandQueue>> + DerefMut<Target=CommandQueue>
+            fn from_queue<Q>(mut queue: Q, capacity: usize) -> $pool
+                where Q: Into<$queue<CommandQueue>> + BorrowMut<CommandQueue>
             {
+                let queue = queue.borrow_mut();
+
                 // Create command pool
                 let info = vk::CommandPoolCreateInfo {
                     s_type: vk::StructureType::CommandPoolCreateInfo,
@@ -113,7 +116,7 @@ macro_rules! impl_pool {
                             device: queue.device.clone(),
                         }
                     )
-                }).collect::<Vec<_>>();
+                }).collect();
 
                 $pool {
                     pool: command_pool,
--- a/src/backend/vulkan/src/pool.rs
+++ b/src/backend/vulkan/src/pool.rs
@@ -1,0 +1,133 @@
+// Copyright 2017 The Gfx-rs Developers.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::ops::DerefMut;
+use std::ptr;
+use std::sync::Arc;
+use ash::vk;
+use ash::version::DeviceV1_0;
+
+use core::{self, pool};
+use core::command::{Encoder};
+use core::{CommandPool, GeneralQueue, GraphicsQueue, ComputeQueue, TransferQueue};
+use command::CommandBuffer;
+use native::{self, GeneralCommandBuffer, GraphicsCommandBuffer, ComputeCommandBuffer, TransferCommandBuffer, SubpassCommandBuffer};
+use {CommandQueue, RawDevice};
+
+macro_rules! impl_pool {
+    ($pool:ident, $queue:ident, $buffer:ident) => (
+        pub struct $pool {
+            pool: vk::CommandPool,
+            command_buffers: Vec<$buffer>,
+            next_buffer: usize,
+            device: Arc<RawDevice>,
+        }
+
+        impl core::CommandPool for $pool {
+            type Queue = CommandQueue;
+            type PoolBuffer = $buffer;
+
+            fn acquire_command_buffer<'a>(&'a mut self) -> Encoder<'a, $buffer> {
+                let available_buffers = self.command_buffers.len() as isize - self.next_buffer as isize;
+                if available_buffers <= 0 {
+                    self.reserve((-available_buffers) as usize + 1);
+                }
+
+                let buffer = &mut self.command_buffers[self.next_buffer];
+                self.next_buffer += 1;
+
+                let info = vk::CommandBufferBeginInfo {
+                    s_type: vk::StructureType::CommandBufferBeginInfo,
+                    p_next: ptr::null(),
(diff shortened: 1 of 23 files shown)


src/core/src/queue.rs
```diff
@@ -58,19 +58,51 @@ impl<Q: CommandQueue> DerefMut for GeneralQueue<Q> {
...
-impl<Q: CommandQueue> Into<TransferQueue<Q>> for GeneralQueue<Q> {
-    fn into(self) -> TransferQueue<Q> {
-        TransferQueue(self.0)
+impl<Q: CommandQueue> From<GeneralQueue<Q>> for TransferQueue<Q> {
+    fn from(queue: GeneralQueue<Q>) -> TransferQueue<Q> {
+        TransferQueue(queue.0)
+    }
+}
+
+impl<'a, Q: CommandQueue> From<&'a GeneralQueue<Q>> for &'a GraphicsQueue<Q> {
```
msiglreith: These conversion implementations are mostly for convenience. I noticed during writing the App part that we probably can't relay on APIs to expose at least one GeneralQueue, but at least a GeneralQueue or a GraphicsQueue (if the backend doesnt support compute). Due to typesafety it becomes quite verbose to select the correct queue. With these impls queue types can be `downcasted` which makes it easier.

src/core/src/queue.rs
```diff
@@ -104,9 +136,9 @@ impl<Q: CommandQueue> DerefMut for GraphicsQueue<Q> {
     }
 }
 
-impl<Q: CommandQueue> Into<TransferQueue<Q>> for GraphicsQueue<Q> {
-    fn into(self) -> TransferQueue<Q> {
-        TransferQueue(self.0)
+impl<Q: CommandQueue> From<GraphicsQueue<Q>> for TransferQueue<Q> {
```
msiglreith: Switched to `From` as it seems to be better practice.

src/lib.rs
```diff
@@ -284,44 +257,75 @@ A: Sized + ApplicationBase<gfx_device_metal::Resources, gfx_device_metal::Comman
...
+        colors: main_colors,
         depth: main_depth.2,
         aspect_ratio: width as f32 / height as f32, //TODO
     });
 
     let mut harness = Harness::new();
     let mut running = true;
+    let mut frame_semaphore = factory.create_semaphore();
+
+    // TODO: How can we get rid of this checks all the time?
```
msiglreith: Outdated comment, will remove in the next PR..

src/lib.rs
```diff
@@ -333,9 +337,14 @@ A: Sized + ApplicationBase<gfx_device_vulkan::Resources, gfx_device_vulkan::Comm
                 _ => app.on(event),
             }
         });
-        let mut frame = win.start_frame();
-        app.render(frame.get_queue());
-        frame.get_queue().cleanup();
+        let frame = swap_chain.acquire_frame(FrameSync::Semaphore(&mut frame_semaphore));
```
msiglreith: The new rendering loop for application.

src/backend/vulkan/src/pool.rs
```diff
@@ -0,0 +1,133 @@
...
+                                  .expect("Error on command buffer allocation") // TODO: better error handling
+                };
+                let command_buffers = command_buffers.into_iter().map(|buffer| {
+                    $buffer(
+                        CommandBuffer {
+                            raw: buffer,
+                            device: queue.device.clone(),
+                        }
+                    )
+                }).collect::<Vec<_>>();
```
kvark: nit: I don't think you need to explicitly specify the collect type here, since it's inferred.

src/backend/vulkan/src/pool.rs
```diff
@@ -0,0 +1,133 @@
...
+use core::{self, pool};
+use core::command::{Encoder};
+use core::{CommandPool, GeneralQueue, GraphicsQueue, ComputeQueue, TransferQueue};
+use command::CommandBuffer;
+use native::{self, GeneralCommandBuffer, GraphicsCommandBuffer, ComputeCommandBuffer, TransferCommandBuffer, SubpassCommandBuffer};
+use {CommandQueue, RawDevice};
+
+macro_rules! impl_pool {
+    ($pool:ident, $queue:ident, $buffer:ident) => (
+        pub struct $pool {
```
kvark: can we convert this code to do more generic work then macros?
e.g.
```rust
pub struct GenericPool<B> {
...
   command_buffers: Vec<B>,
}
impl<B> core::CommandPool for GenericPool<B> {
...
}
```

src/core/src/pool.rs
```diff
@@ -0,0 +1,59 @@
...
+/// Transfer command pool can allocate transfer command buffers.
+pub trait TransferCommandPool: CommandPool {
+    ///
+    fn from_queue<Q>(queue: &mut Q, capacity: usize) -> Self
+        where Q: Into<TransferQueue<Self::Queue>> +
+                 DerefMut<Target=Self::Queue>;
+}
+
+/// Subpass command pool can allocate subpass command buffers.
+pub trait SubpassCommandPool: CommandPool {
```
kvark: similarly, is there a particular reason for these to be separate traits?
i.e. could we have a generic `pub trait SpecificCommandPool<Something>: CommandPool {...}`, parametrized by either something existing, or maybe a phantom type?

src/core/src/queue.rs
```diff
@@ -58,19 +58,51 @@ impl<Q: CommandQueue> DerefMut for GeneralQueue<Q> {
...
-        TransferQueue(self.0)
+impl<Q: CommandQueue> From<GeneralQueue<Q>> for TransferQueue<Q> {
+    fn from(queue: GeneralQueue<Q>) -> TransferQueue<Q> {
+        TransferQueue(queue.0)
+    }
+}
+
+impl<'a, Q: CommandQueue> From<&'a GeneralQueue<Q>> for &'a GraphicsQueue<Q> {
+    fn from(queue: &'a GeneralQueue<Q>) -> &'a GraphicsQueue<Q> {
+        unsafe { &*(queue as *const _ as *const GraphicsQueue<Q>) }
```
kvark: why is this safe?

src/lib.rs
```diff
@@ -284,44 +257,75 @@ A: Sized + ApplicationBase<gfx_device_metal::Resources, gfx_device_metal::Comman
...
+    let (surface, adapters) = window.get_surface_and_adapters();
+
+    // Init device
+    let queue_descs = adapters[0].get_queue_families().iter()
+                                 .filter(|family| surface.supports_queue(&family) )
+                                 .map(|family| { (family, family.num_queues()) })
+                                 .collect::<Vec<_>>();
+    let gfx_core::Device_ { mut factory, mut general_queues, mut graphics_queues, .. } = adapters[0].open(&queue_descs);
+
+    let queue = if !general_queues.is_empty() {
```
kvark: nit: could be `let queue = match general_queues.first_mut() {...}`

src/core/src/queue.rs
msiglreith: Right.. this only made sense for the vulkan backend..
I'm refactoring this towards something like `SpecificQueueBase<Q: CommandQueue, R: Repr<Q>>(R, PhantomData<Q>>` which should allow to deal with it in a safer way!

PR #1273, opened 2017-05-17
corell Metal backend

JohnColanduoni:
This is a first pass at a corell Metal backend. It has just enough working to run the trianglell example (which I didn't know displayed a rectangle now XD). A couple of important notes:

* The corell SwapChain API doesn't work well with the usual method of drawing to windows with Metal (namely CAMetalLayer). CAMetalLayer maintains a set of textures internally for the backbuffer, but it handles reuse via reference counting and doesn't give us access to the persistent ordered list that we need. To work around this, I've implemented the swap chain manually, taking advantage of the ability to create a Metal texture from a GPU-resident IOSurface (which appears to be how CAMetalLayer does it). There's definitely a way to directly use an IOSurface as a backing store for a CALayer to prevent the need for any copying, but until I can figure out how to do that I used a single-buffered OpenGL context and glBlitFramebuffer to handle presentation. This only imposes the overhead of a single GPU-GPU copy per frame.

* Metal has heaps and fences, but they are unavailable on macOS. The code that is specific to heap/fence emulation is marked with a "native_heap" /"native_fence" feature so it is clear what needs to change to implement unemulated variants in the future.


msiglreith:
Awesome! 💯  I'll have a look later on!

kvark:
WOOT!
@homu delegate=msiglreith

examples/trianglell/main.rs
```diff
@@ -92,6 +97,11 @@ fn main() {
             ("ps_main", include_bytes!("data/ps_main.spv"))]
         ).expect("Error on creating shader lib");
 
+    #[cfg(feature = "metal")]
+    let shader_lib = factory.create_shader_library_from_source(
```
msiglreith: Good to see the actual library feature in use!

src/backend/metalll/src/command.rs
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    fn bind_vertex_buffers(&mut self, buffer: pso::VertexBufferSet<Resources>) {
+        self.vertex_buffers.clear();
+        self.vertex_buffers.extend(buffer.0.iter().map(|&(buffer, offset)| (buffer.0, offset)));
+    }
+
+    fn set_viewports(&mut self, rects: &[target::Rect]) {
+        if rects.len() != 1 {
+            panic!("Metal supports only one viewport");
```
msiglreith: We could propagate the errors in the future and emit then on `finish()`

src/backend/metalll/src/command.rs
```diff
@@ -0,0 +1,613 @@
...
+            if let Some(viewport) = command_buffer.viewport {
+                render_encoder.set_viewport(viewport);
+            }
+            if let Some(scissors) = command_buffer.scissors {
+                render_encoder.set_scissor_rect(scissors);
+            }
+            if let Some(pipeline_state) = command_buffer.pipeline_state {
+                render_encoder.set_render_pipeline_state(pipeline_state);
+            } else {
+                panic!("missing bound pipeline state object");
```
msiglreith: Is this a requirement for Metal Encoders to have always an bound PSO?
I've seen various other constraints in some backends (e.g vertex bind slots need to defined etc.).
We should document them somewhere but that's out of scope.

msiglreith:
Thanks again, great work!
Yeah, it should display the gfx-rs logo, `trianglell` might not be the best name :grinning: 
@homu r+

src/backend/metalll/src/command.rs
JohnColanduoni: The docs say its required for draw calls, but doesn't specify whether you can call end_encoding without setting one. I'll have to test this to see if you can use an empty encoder to clear renderbuffers.

PR #1274, opened 2017-05-17
Improved Metal rendering and presentation for corell

JohnColanduoni:
Metal's SwapChain implementation now presents the backbuffers directly, without any copying,
by setting the IOSurface backing the Metal texture as the contents of a CALayer. This hands
our renderbuffer straight to the compositor, just as the native swapchain in CAMetalLayer does.

Additionally, the trianglell example now displays almost correctly when using Metal; it now samples
the texture properly instead of displaying the texture coordinates as colors. However blend mode
interpretation is still not implemented, so the background of the rectangle is white, not transparent.

### DIFF
--- a/examples/trianglell/shader/triangle.metal
+++ b/examples/trianglell/shader/triangle.metal
@@ -7,17 +7,21 @@ struct VsInput {
 
 struct VsOutput {
   float4 pos [[position]];
-  float3 color;
+  float2 uv;
 };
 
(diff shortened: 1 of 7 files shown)


kvark:
Wow, that was fast!
@homu delegate=msiglreith

JohnColanduoni:
Yeah, turns out the reason the presentation part wasn't working was I specified the backbuffer format as RGBC instead of RGBA XD

msiglreith:
@homu r+

Issue #1267, opened 2017-05-13
Android: "gl function was not loaded"


ozkriff:
`gl.BindFragDataLocation` call (added in https://github.com/gfx-rs/gfx/commit/20ce6da1188b4ec3ec049619d268e04d3249724e) causes 

```plain
thread 'main' panicked at 'gl function was not loaded', /home/ozkriff/zoc/main/target/arm-linux-androideabi/debug/build/gfx_gl-15c7fd5127b4e8d8/out/gl_bindings.rs:1479
```

on android.

torkleyy:
The reason this function is not available under OpenGL ES is that there is no way to have multiple targets :worried: 

Here the steps needed to fix the issue:

* find a way to access version and capabilities in `backend::gl::shade::create_program`
* in `backend::gl::shade::create_program`, only bind the frag data location in case it's not embedded
* eventually warn in case we're compiling for Android and there is a `Target1`

PR #1276, opened 2017-05-18
Only enable gl.BindFragDataLocation for desktop

ozkriff:
Fixes https://github.com/gfx-rs/gfx/issues/1267


ozkriff:
>  instead of sharing Info, can we maybe have a boolean flag in PrivateCaps for this?

Closed in favor of https://github.com/gfx-rs/gfx/pull/1277

PR #1277, opened 2017-05-18
Only enable gl.BindFragDataLocation for desktop (v2)

ozkriff:
Fixes #1267

PR #1278, opened 2017-05-18
[v0.16] GL fixes for Android

kvark:
Includes #1277 and #1266
cc @ozkriff


kvark:
Pretty sure it's safe, given that we haven't diverged from `master` too much, and it's already been CI-tested there.
@homu r+

PR #1279, opened 2017-05-22
Enable Bgra8 on OpenGL backend

zakorgy:
Replacing the TODO parts related to `B8_G8_R8_A8`, since using Bgra8 format type with OpenGL backend causes runtime panics, because of these unhandled cases.

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1279)

### DIFF
--- a/src/backend/gl/src/tex.rs
+++ b/src/backend/gl/src/tex.rs
@@ -55,9 +55,9 @@ fn kind_face_to_gl(kind: t::Kind, face: Option<t::CubeFace>) -> GLenum {
 fn format_to_glpixel(format: NewFormat) -> GLenum {
     use core::format::SurfaceType as S;
     use core::format::ChannelType as C;
-    let (r, rg, rgb, rgba) = match format.1 {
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/tex.rs
```diff
@@ -208,7 +207,10 @@ fn format_to_glfull(format: NewFormat) -> Result<GLenum, ()> {
             C::Float => gl::RGBA32F,
             _ => return Err(()),
         },
-        S::B8_G8_R8_A8 => return Err(()), // TODO
+        S::B8_G8_R8_A8 => match cty {
+            C::Unorm => gl::RGBA8,
```
kvark: So, if I understand correctly, the texture data is still stored as `RGBA8` but what you want to specify under `BGRA` format is that the data you provide is swizzled?

kvark:
@homu r+

src/backend/gl/src/tex.rs
zakorgy: Yes, that was the idea.

PR #1284, opened 2017-05-30
Fix typo and remove trailing whitespace in slice.rs

### LINKED ISSUE #1283: Tiny typo in docs for IndexBuffer
In [the docs for `IndexBuffer`](https://docs.rs/gfx/0.16.0/gfx/enum.IndexBuffer.html), there's a part that reads:

> Do note that the `Slice`' start and end restrictions still apply for this variant.

There should be an "s" after the apostrophe. I would make a PR to fix this, but my editor removes trailing whitespace on save, so it's a bit annoying for me to make the minimal diff to fix this.

ucarion:
Fixes #1283 .


kvark:
Thank you!
Appveyor goes nuts:
> fatal runtime error: out of memory

@homu r+

PR #1286, opened 2017-05-31
Improved GLES capability list

ColinKinloch:
I think these values are correct.
I compared the header at https://www.khronos.org/registry/OpenGL/api/GLES3/gl32.h to the functions listed in the extension specifications at https://www.khronos.org/registry/OpenGL/extensions/ARB/

### DIFF
--- a/src/backend/gl/src/factory.rs
+++ b/src/backend/gl/src/factory.rs
@@ -496,7 +496,7 @@ impl f::Factory<R> for Factory {
 
     fn create_sampler(&mut self, info: t::SamplerInfo) -> handle::Sampler<R> {
         let name = if self.share.private_caps.sampler_objects_supported {
-            tex::make_sampler(&self.share.context, &info)
+            tex::make_sampler(&self.share.context, &info, &self.share.private_caps)
         } else {
             0
         };
--- a/src/backend/gl/src/info.rs
+++ b/src/backend/gl/src/info.rs
@@ -171,6 +171,7 @@ pub struct PrivateCaps {
     pub buffer_storage_supported: bool,
     pub clear_buffer_supported: bool,
     pub frag_data_location_supported: bool,
+    pub sampler_lod_bias_supported: bool,
 }
 
 /// OpenGL implementation information
@@ -184,6 +185,13 @@ pub struct Info {
     pub shading_language: Version,
     /// The extensions supported by the implementation
     pub extensions: HashSet<&'static str>,
+}
+
+#[derive(Copy, Clone)]
+enum Requirement {
+    Core(u32,u32),
+    Es(u32, u32),
+    Ext(&'static str),
 }
 
 impl Info {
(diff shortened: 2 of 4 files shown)


src/backend/gl/src/info.rs
```diff
@@ -171,6 +171,7 @@ pub struct PrivateCaps {
     pub buffer_storage_supported: bool,
     pub clear_buffer_supported: bool,
     pub frag_data_location_supported: bool,
+    pub is_embedded: bool,
```
kvark: let's make the flag more specific, i.e. `sampler_lod_bias_supported`

src/backend/gl/src/factory.rs
```diff
@@ -496,7 +496,7 @@ impl f::Factory<R> for Factory {
 
     fn create_sampler(&mut self, info: t::SamplerInfo) -> handle::Sampler<R> {
         let name = if self.share.private_caps.sampler_objects_supported {
-            tex::make_sampler(&self.share.context, &info)
+            tex::make_sampler(&self.share.context, &info, self.share.private_caps.is_embedded)
```
kvark: let's pass the whole `private_caps` in there by referece

src/backend/gl/src/info.rs
```diff
@@ -237,28 +246,57 @@ pub fn get(gl: &gl::Gl) -> (Info, Capabilities, PrivateCaps) {
         max_texture_size: get_usize(gl, gl::MAX_TEXTURE_SIZE),
         max_patch_size: if tessellation_supported { get_usize(gl, gl::MAX_PATCH_VERTICES) } else {0},
 
-        instance_base_supported:           info.is_version_or_extension_supported(4, 2, "GL_ARB_base_instance"),
-        instance_call_supported:           info.is_version_or_extension_supported(3, 1, "GL_ARB_draw_instanced"),
-        instance_rate_supported:           info.is_version_or_extension_supported(3, 3, "GL_ARB_instanced_arrays"),
-        vertex_base_supported:             info.is_version_or_extension_supported(3, 2, "GL_ARB_draw_elements_base_vertex"),
-        srgb_color_supported:              info.is_version_or_extension_supported(3, 2, "GL_ARB_framebuffer_sRGB"),
-        constant_buffer_supported:         info.is_version_or_extension_supported(3, 1, "GL_ARB_uniform_buffer_object"),
+        instance_base_supported:           info.is_supported(Some((4, 2)),
```
kvark: I like how structured it became, although there are holes still (e.g. with "GL_NV_copy_buffer" check).
I'd prefer this code to be less verbose though. For example, we could pass extensions as `&[&str]` list instead of `Option<&str>`. And for the GL versions we could even go for `f32` values (`3.1`, `3.2`, `NaN`).

src/backend/gl/src/info.rs
ColinKinloch: I don't disagree, I was using the same naming scheme as elsewhere in the code:
[src/backend/gl/src/info.rs#L24](https://github.com/gfx-rs/gfx/blob/6a300e46a239bfbac8394d500759f7210b692568/src/backend/gl/src/info.rs#L24)
[src/backend/gl/src/tex.rs#L559](https://github.com/gfx-rs/gfx/blob/38a6aa384794f77ff11361b6ca507064b20a88e7/src/backend/gl/src/tex.rs#L559)

There are enough silly little differences between GL and GLES that an embedded flag to cover all bases may be preferable.

src/backend/gl/src/info.rs
ColinKinloch: I was wondering why the logic there was so complex, I didn't want to cause trouble for someone more versed in the intricacies of the copy buffer extension.
I suppose in the case that there were sufficient difference between the NV and ARB versions it wouldn't be a boolean value.

I don't think that representing GL version number as a floating point is a good idea the [docs](https://www.khronos.org/opengl/wiki/Get_Context_Info#OpenGL_version_number) define it as two integers GL_MAJOR_VERSION and GL_MINOR_VERSION.
I get your point though, I've created a simple macro that arguably makes the code more expressive.

src/backend/gl/src/info.rs
kvark: The first code link uses it correctly, since `Info` is not a capability struct.
The second one is indeed a bad example on our part, I'd prefer it to be fixed in the way I described here.

src/backend/gl/src/info.rs
```diff
@@ -233,68 +225,88 @@ impl Info {
...
+        false
+    };
 }
 
 /// Load the information pertaining to the driver and the corresponding device
 /// capabilities.
 pub fn get(gl: &gl::Gl) -> (Info, Capabilities, PrivateCaps) {
     let info = Info::get(gl);
-    let tessellation_supported =           info.is_version_or_extension_supported(4, 0, "GL_ARB_tessellation_shader");
+    let tessellation_supported =           cap_support!(info, core (4,0),
```
kvark: this is much better! I think we can improve here by not using a macro. Instead, you could have
```
enum RequiredVersion {
    Core(u32,u32),
    Es(u32, u32),
}
```
which is included as `use RequiredVersion::*`, so that the ergonomics of using it inside this function stays as good as it is with the macro

kvark:
Beautiful, thank you!
I'm going to squash & merge after CI is done, unless you want to squash on your side.

ColinKinloch:
I just added the renaming of the flag.
You go ahead with the squash if your happy with it.

kvark:
@ColinKinloch thank you! I'm very happy to see the code you came up with, awesome job!
The benefit of you squashing rather than me is that we are able to use @homu then to merge it.

kvark:
@homu r+

Issue #1291, opened 2017-06-05
[ll] Backend Generics


JohnColanduoni:
Currently it is pretty difficult to create generics over backends when using corell:

- Neither Adapter nor Backend require that their CommandQueue associated types use the same Resources as them. Adding that constraint is also not simple, because CommandQueue<R=Self::Resources> doesn't immediately propagate to the command buffer associated types so they must be separately constrained.
- RenderPassInlineEncoder and RenderPassSecondaryEncoder's lifetime parameters make their inclusion in any Backend trait very difficult, because you then need to make that Backend trait have 4 lifetime parameters which then infects any trait or structure you try to make generic over the Backend trait.

I think the best way to handle this is to extend the techniques we use for other traits in corell to these:

- Make CommandQueue take Resources as a generic parameter like CommandBuffer does.
- Make RenderPassInlineEncoder/RenderPassSecondaryEncoder structs with references to the command buffer, RenderPass, and FrameBuffer with appropriate lifetimes that contains a backend-specific type inside, to which the structs proxy all of the calls (like Encoder does).

If this sounds good I'd be happy to implement it :smile:

msiglreith:
Highly agree! Noticed this issue also while fiddling with `render`. I have something similar cooking on my experimentale [branch](https://github.com/gfx-rs/gfx/compare/ll...msiglreith:pr_asteria_05) and wanted to do a PR for the `ll` integration branch but couldn't test it on my dektop so far.

My current approach was based on re-introducing the [`Backend`](https://github.com/msiglreith/gfx/blob/pr_asteria_06/src/core/src/lib.rs#L228-L239) trait, which defines the types .
Replacing most of the associated types with a generic `B: Backend` works quite well so far.
How does this sound?

JohnColanduoni:
Yeah that sounds perfect! I'm playing around with turning RenderPassI*Encoder into structs to work around the lifetimes, the rest of the API seems pretty generic-friendly so far.

PR #1292, opened 2017-06-06
[ll] Re-introduce Backend trait [06/..]

msiglreith:
Addresses the mentioned Backend trait from #1291 (first bullet point) for `ll` integration.

Introducing a new `Backend` trait in similar fashion to the current `Resource` trait simplifies the API and helps higher-level layers for operating on the core. For more details regarding the current problems with associated trait approach please checkout the mentioned issue!
Implemented the Backend changes for GL, Vulkan and D3D12 backend so far,

For reviewing: [Backend trait](https://github.com/msiglreith/gfx/blob/pr_asteria_06v2/src/core/src/lib.rs#L229-L242)

Other notable changes:
- Remove `Send` from Buffer, anticipating the upcoming changes towards allocation of command pools instead of sending around Encoder. Also, added dummy implementation for the Buffer trait to allow merging the first parts of the integration branch once the setup stuff is done without needing to additionally rewrite all the command buffer stuff.
- Macrofiy the `Queue` definitions [src/core/queue.rs](https://github.com/msiglreith/gfx/blob/pr_asteria_06v2/src/core/src/queue.rs), helps to reduce code duplication. Will be extended if further conversions and trait impls are needed. Some occurences of `Borrow<Queue>` etc will be revised in the future..

cc @JohnColanduoni

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -13,9 +13,11 @@
 // limitations under the License.
 
 use comptr::ComPtr;
-use core::command;
+use core::{command, pso, shade, state, target, texture as tex};
+use core::{IndexType, VertexCount};
 use native::{GeneralCommandBuffer, GraphicsCommandBuffer, ComputeCommandBuffer, TransferCommandBuffer, SubpassCommandBuffer};
 use winapi;
+use {Backend, Resources};
 
 pub struct CommandBuffer {
     pub raw: ComPtr<winapi::ID3D12GraphicsCommandList>,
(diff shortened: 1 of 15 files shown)


kvark:
Thanks @msiglreith !
I'm not quite sure about the giant macros thing (e.g. implementing GL command buffers). Could you explain the need for them?

msiglreith:
In particular for the command buffers: `ll` currently splits the command buffer functionalities depending on functionality (compute/graphics/transfer), but this hasn't been integrated yet. To be able to merge the current branch into master we need to at least implement `Buffer` for now (mainly graphics functionalities).

PR #1296, opened 2017-06-08
[ll] Working on app unification [09/..]

msiglreith:
Currently trying to merge the current `launch_` functions into one backend-agnostic `run` function for app. Only surface and adapter setup should be then backend specific.

Other changes induced by the mentioned goal above (note: focusing on GL and Vulkan atm):
* Implement command pools for GL (maybe pool implementations can be moved towards `core` similar to CommandQueues, which would reduce code size.. )
* Add `SwapchainConfig`, which allows to specify color and depth-stencil formats for the backbuffers. Should be extended in the future for present modes (if possible), swapchain size and possible other options.
* `present` now takes the present queue, which allows to remove the `Mutex` for CommandQueues on Vulkan

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -37,10 +37,10 @@ mod command;
 mod factory;
 mod info;
 mod native;
+mod pool;
 mod shade;
 mod state;
 mod tex;
-
 
 pub type Buffer         = gl::types::GLuint;
 pub type ArrayBuffer    = gl::types::GLuint;
@@ -59,6 +59,7 @@ unsafe impl Sync for Fence {}
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub enum Backend {}
 impl c::Backend for Backend {
+    type Adapter = Adapter;
     type Resources = Resources;
     type CommandQueue = CommandQueue;
     type GeneralCommandBuffer = command::GeneralCommandBuffer;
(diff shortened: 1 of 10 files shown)


kvark:
Could you elaborate on why `Compatible` is needed and how this empty trait helps?

msiglreith:
`Compatible` encodes the supported queue flairs (compute, graphics and transfer).
e.g `GraphicsQueue` supports graphics and transfer, so we can build `GraphicsCommandPools` and `TransferCommandPools`, and be able to submit graphics and transfer command buffers to the queue.

Atm Pools are created from a queue, but we need to ensure that the queue supports the functionality of the pool command buffers (e.g shouldn't be able to create a GraphicsPool from a ComputeQueue).

kvark:
So it's fine for `Compatible` to not have any actual logic then, other than serving as a strong type protector? I'd expect something like `AsRef<TargetCommandBuffer>` to serve both, maybe.

msiglreith:
> I'd expect something like AsRef<TargetCommandBuffer> to serve both, maybe.

This isn't possible due having to construct the the `TargetCommandBuffer` but I will try to improve that point.

kvark:
@msiglreith feel free to merge the PR and then follow-up at will. I'm happy with the code, given you addressed my concerns.

msiglreith:
Ok, cool. I'll look to combine it with this one depending on how many changes this will need, else split it. (Not totally sure right now how do address it)

msiglreith:
I'm afraid I don't see a solution atm, which would improve the current situation.
To clarify the my current design decisions:
* `AsRef` and `From` derives are atm soley for the library users to allow 'downcasting' queues (e.g GeneralQueue -> GraphicsQueue). This is useful if you want a GraphicsQueue but the backend will only create a GeneralQueue, so you can downcast it (or the associated Ref/Mut) .
* The backends only need to ensure a compatible feature set (`Compatible<Q>`) and to have access to the internal `CommandQueue` impl.

PR #1300, opened 2017-06-09
Separate Buffer/Encoder for RenderPass Encoders

JohnColanduoni:
To avoid needing to include traits with lifetime parameters in
the GraphicsCommandBuffer trait, we split the RenderPass\*Encoder trait
into RenderPass\*Buffer (a backend-implemented trait with no lifetime
parameters) and a struct RenderPass*Encoder containing the necessary
lifetime parameters and associated references.

These structs proxy all their calls to the Buffer trait implementations
by passing a reference to the whole struct that includes the buffer as
a member, as well as the lifetime-bound references. As a result the
backend implementations have access to all the data they had previously
without having to resort to lifetime parameters on the trait or
unsafety.

### DIFF
--- a/examples/trianglell/main.rs
+++ b/examples/trianglell/main.rs
@@ -405,8 +405,7 @@ fn main() {
             cmd_buffer.bind_graphics_descriptor_sets(&pipeline_layout, 0, &[&set0[0], &set1[0]]);
 
             {
-                let mut encoder = back::RenderPassInlineEncoder::begin(
-                    &mut cmd_buffer,
+                let mut encoder = cmd_buffer.begin_render_pass_inline(
                     &render_pass,
                     &framebuffers[frame.id()],
                     target::Rect { x: 0, y: 0, w: pixel_width, h: pixel_height },
--- a/src/backend/dx12ll/src/command.rs
+++ b/src/backend/dx12ll/src/command.rs
@@ -20,6 +20,7 @@ use winapi::*;
 
 use core::{self, command, memory, pso, state, target, IndexType, VertexCount, VertexOffset};
 use core::buffer::IndexBufferView;
+use core::command::{RenderPassInlineEncoder, RenderPassSecondaryEncoder, Encoder};
 use data;
 use native::{self, CommandBuffer, GeneralCommandBuffer, GraphicsCommandBuffer,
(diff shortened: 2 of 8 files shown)


JohnColanduoni:
I'll spin up Windows and fix the DX12 backend later tonight, but the Metal and Vulkan backends worked when I tested them.

examples/trianglell/main.rs
```diff
@@ -405,8 +405,7 @@ fn main() {
             cmd_buffer.bind_graphics_descriptor_sets(&pipeline_layout, 0, &[&set0[0], &set1[0]]);
 
             {
-                let mut encoder = back::RenderPassInlineEncoder::begin(
-                    &mut cmd_buffer,
+                let mut encoder = cmd_buffer.begin_render_pass_inline(
```
msiglreith: Great! Much better.

msiglreith:
For dx12, our current goal should be that it compiles, `trianglell` won't run anyway at the current state iirc.
Where did appveyor go again?

JohnColanduoni:
On further reflection I think we might actually want to make the RenderPassBuffer traits hidden in their entirety. They never show up in any capacity unless you're writing a backend; all users see is a GraphicsCommandBuffer as a generic parameter for the RenderPassEncoder struct. The duplication of the method names could be quite confusing when reading the documentation.

kvark:
So we'll have multiple visibility rings, logically:
  - inner ring - private to gfx_core
  - backend implementors ring
  - users ring, including gfx_render

JohnColanduoni:
The build is only failing on nightly, from an out of memory error when compiling dxguid-sys (which appears to a [known bug](https://github.com/rust-lang/rust/issues/42544) in rustc).

src/backend/dx12ll/src/command.rs
```diff
@@ -20,6 +20,7 @@ use winapi::*;
 
 use core::{self, command, memory, pso, state, target, IndexType, VertexCount, VertexOffset};
 use core::buffer::IndexBufferView;
+use core::command::{RenderPassInlineEncoder, RenderPassSecondaryEncoder, Encoder};
```
kvark: nit: `command` is already included, so there should be either just `command` or all the types we use from it for consistency

kvark:
Thank you!
@homu r+

PR #1301, opened 2017-06-09
Replace cgmath bindings by `mint`

### LINKED ISSUE #1290: Replace cgmath bindings by `mint`
https://github.com/kvark/mint

cc @ebkalderon 

Also need to cleanup the dependencies. I.e. instead of having this in the root Cargo.toml:
```toml
cgmath-types = ["gfx/cgmath-types", "gfx_core/cgmath-types"]
```
We could just have this:
```toml
mint = "gfx/mint"
```
Which would then require "gfx_core/mint" internally.

vitvakatu:
Closes #1290 

Unfortunately (or maybe not), `Rad` and `Deg` types were removed.

### DIFF
--- a/src/core/Cargo.toml
+++ b/src/core/Cargo.toml
@@ -29,7 +29,7 @@ path = "src/lib.rs"
 
 [dependencies]
 bitflags = "0.8"
-cgmath = { version = "0.14", optional = true }
+mint = { version = "0.3", optional = true }
 derivative = "1.0"
 draw_state = "0.7"
 log = "0.3"
@@ -37,6 +37,5 @@ serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }
 
 [features]
-cgmath-types = ["cgmath"]
 serialize = ["serde", "serde_derive", "draw_state/serialize"]
 unstable = []
--- a/src/core/Cargo.toml
+++ b/src/core/Cargo.toml
@@ -29,7 +29,7 @@ path = "src/lib.rs"
 
 [dependencies]
 bitflags = "0.8"
-mint = { version = "0.3", optional = true }
+mint = { version = "0.4.1", optional = true }
 derivative = "1.0"
 draw_state = "0.7"
 log = "0.3"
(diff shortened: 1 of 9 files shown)


src/core/Cargo.toml
```diff
@@ -29,14 +29,13 @@ path = "src/lib.rs"
 
 [dependencies]
 bitflags = "0.8"
-cgmath = { version = "0.14", optional = true }
+mint = { version = "0.3", optional = true }
```
kvark: please use "0.4.1" here

src/core/src/shade.rs
```diff
@@ -21,8 +21,13 @@ use std::error::Error;
 use {Resources};
 use {AttributeSlot, ColorSlot, ConstantBufferSlot, ResourceViewSlot, SamplerSlot, UnorderedViewSlot};
 
-#[cfg(feature = "cgmath-types")]
-use cgmath::{Deg, Matrix2, Matrix3, Matrix4, Point2, Point3, Rad, Vector2, Vector3, Vector4};
+#[cfg(feature = "mint")]
+use mint::{
+    Point2, Point3,
```
kvark: I don't think it makes sense to use all the types explicitly as opposed to a simple `use mint;`, given that all the usage is in macros, so there are no ergonomics concerns

src/render/Cargo.toml
```diff
@@ -30,12 +30,11 @@ name = "gfx"
 path = "src/lib.rs"
 
 [features]
-cgmath-types = ["gfx_core/cgmath-types", "cgmath"]
 serialize = ["gfx_core/serialize", "draw_state/serialize"]
 unstable = []
 
 [dependencies]
-cgmath = { version = "0.14", optional = true }
+mint = { version = "0.3", optional = true }
```
kvark: same, please use "0.4.1"

src/render/src/shade.rs
```diff
@@ -14,8 +14,12 @@
 
 //! Shader parameter handling.
 
-#[cfg(feature = "cgmath-types")]
-use cgmath::{Deg, Matrix2, Matrix3, Matrix4, Rad, Point2, Point3, Vector2, Vector3, Vector4};
+#[cfg(feature = "mint")]
+use mint::{
+    Point2, Point3,
```
kvark: similarly, please just do `use mint`

vitvakatu:
Appveyor nightly fails because of https://github.com/rust-lang/rust/issues/42544
I had the same problem on my local machine

PR #1302, opened 2017-06-10
Add ggez engine to "Who's using it"

vitvakatu:
Closes #1295

PR #1303, opened 2017-06-10
Fix version in readme

### LINKED ISSUE #1111: Bump version number in README

vitvakatu:
Includes #1302 
Closes #1111 

We should carefully look at instructions in README to avoid new user's problems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Hardware features:
 
 Biggest open-source projects are:
   - [Amethyst](https://github.com/amethyst/amethyst) engine
+  - [ggez](https://github.com/ggez/ggez) engine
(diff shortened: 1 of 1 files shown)


kvark:
@homu r+

PR #1308, opened 2017-06-11
[ll] app: Adopt GL to the application refactor [10/..]

msiglreith:
Current focus lays on refactoring the app to minimize the backend specific code to window setup.

### DIFF
--- a/src/backend/gl/src/factory.rs
+++ b/src/backend/gl/src/factory.rs
@@ -23,7 +23,7 @@ use core::handle::{self, Producer};
 use core::target::{Layer, Level};
 
 use command::{COLOR_DEFAULT};
-use {Resources as R, Share, OutputMerger};
+use {Info, Resources as R, Share, OutputMerger};
 use {Buffer, BufferElement, FatSampler, NewTexture,
      PipelineState, ResourceView, TargetView, Fence};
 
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -344,6 +289,22 @@ fn run<A, B, S>(backend: shade::Backend,
     }
 }
 
+trait EventsLoop {
```
msiglreith: `winit` and `glutin` both implement the same methods but don't share a trait. Therefore, I created an own for our specific needs..

src/backend/gl/src/factory.rs
```diff
@@ -88,6 +88,11 @@ impl Factory {
         }
     }
 
+    /// Get the OpenGL-specific driver information
+    pub fn get_info<'a>(&'a self) -> &'a Info {
```
kvark: this is highly subjective, but we could just pass `info.shading_language`  to the factory on construction and avoid sharing the `Info` struct

msiglreith: That could also be done, indeed. But I still would stick with the current solution because `Share` (containing `Info`) is wrapped in a `Rc` and we need `Info` anyways for the CommandQueue later, either stored directly inside the struct or also in `Share` if we want to expose multiple queues for the GL backend.

kvark:
Ok, makes sense.
Please go ahead and merge then.

PR #1314, opened 2017-06-14
[ll] Port cube example and fix gl queue submission [13/..]

msiglreith:
* `Cube` demo now runs on GL (excluding resizing..)
* Fix queue submission on GL as mentioned on the prior PR, based on the old `Device` interface

(Removed unneeded attribute from the last PR)

### DIFF
--- a/examples/cube/main.rs
+++ b/examples/cube/main.rs
@@ -16,11 +16,13 @@ extern crate cgmath;
 #[macro_use]
 extern crate gfx;
 extern crate gfx_app;
+extern crate gfx_core;
 
-pub use gfx_app::{ColorFormat, DepthFormat};
+pub use gfx_app::{BackbufferView, ColorFormat, DepthFormat};
 
 use cgmath::{Deg, Matrix4, Point3, Vector3};
-use gfx::{Bundle, texture};
+use gfx::{Bundle, Factory, texture};
+use gfx::GraphicsPoolExt;
 
 // Declare the vertex format suitable for drawing,
 // as well as the constants used by the shaders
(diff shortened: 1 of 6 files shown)


examples/cube/main.rs
```diff
@@ -16,11 +16,13 @@ extern crate cgmath;
 #[macro_use]
 extern crate gfx;
 extern crate gfx_app;
+extern crate gfx_core;
```
kvark: we need to make it not a required dependency, at the end of the day

src/core/src/lib.rs
```diff
@@ -398,10 +398,22 @@ pub struct QueueSubmit<'a, B: Backend + 'a> {
 /// CommandBuffers will be later submitted to command queues instead of the device.
 pub trait CommandQueue<B: Backend> {
     /// Submit command buffers to queue for execution.
-    unsafe fn submit(&mut self, submit_infos: &[QueueSubmit<B>], fence: Option<&mut <B::Resources as Resources>::Fence>);
+    /// `fence` will be signalled after submission and _must_ be unsignalled.
+    // TODO: `access` legacy (handle API)
+    #[doc(hidden)]
+    unsafe fn submit(&mut self, submit_infos: &[QueueSubmit<B>], fence: Option<&mut <B::Resources as Resources>::Fence>,
```
kvark: why is it unsafe btw?

msiglreith: I marked it as `unsafe` as it's not supposed to be called. You could for example submit `GraphicsCommandBuffers` to a `ComputeQueue`, which will probably result in some driver error.

examples/cube/main.rs
msiglreith: noted!

PR #1315, opened 2017-06-15
[ll] Refactor command buffer types [14/..]

msiglreith:
All APIs have only a single command buffer type, except vulkan's secondary command buffers differ a bit. Therefore it seems reasonable to only require gfx backends to only implement one `RawCommandBuffer`  and have backend-agnostic safe wrappers in the core, which will implement the support functionalities depending on the type (`GraphicsCommandBuffer`, `ComputeCommandBuffer`, etc.)

The current change adds a bit more code, but this should have positive effect in the future by reducing the need of macros or additional code for the backend implementations
(SubpassCommandBuffer will stay a separate type atm, as they are a bit special, but could maybe merged as well in the future.)

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -323,331 +323,343 @@ impl Cache {
 
 }
 
-// CommandBuffer trait implementation
-macro_rules! impl_cmd_buffer {
-    ($buffer:ident) => (
-        /// A command buffer abstraction for OpenGL.
-        ///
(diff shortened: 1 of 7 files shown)


src/backend/gl/src/command.rs
```diff
@@ -323,331 +323,343 @@ impl Cache {
 
 }
 
-// CommandBuffer trait implementation
-macro_rules! impl_cmd_buffer {
```
kvark: hooray macros are gone!

PR #1316, opened 2017-06-16
Add Resource trait for generic resource destruction

JohnColanduoni:
I'm not sure if this is something that others want/need but I was playing around with it and thought I'd put it up for comments.

Adds a trait Resource implemented by all Resources associated types that have a destroy_* method in corell::Factory. This makes it much easier to write abstractions/macros for cleanup that work over multiple resource types. DepthStencilView also appeared to be missing a corresponding method in Factory so I added one.

This functionality is basically impossible to implement on top of corell due to coherency issues; any implementations of a trait for both Resources::X and Resources::Y will be marked as conflicting even if they never do in actuality.

### DIFF
--- a/src/backend/metalll/src/native.rs
+++ b/src/backend/metalll/src/native.rs
@@ -235,6 +235,8 @@ mod fence_related {
     pub struct Fence {}
 }
 
+gfx_impl_resources!();
+
 pub unsafe fn objc_err_description(object: *mut objc::runtime::Object) -> String {
     let description: *mut objc::runtime::Object = msg_send![object, localizedDescription];
     let utf16_len: NSUInteger = msg_send![description, length];
--- a/src/corell/src/lib.rs
+++ b/src/corell/src/lib.rs
@@ -261,30 +261,35 @@ pub trait SwapChain {
 
 /// Different resource types of a specific API. 
 pub trait Resources:          Clone + Hash + Debug + Any {
-    type ShaderLib:           Debug + Any + Send + Sync;
-    type RenderPass:          Debug + Any + Send + Sync;
-    type PipelineLayout:      Debug + Any + Send + Sync;
-    type GraphicsPipeline:    Debug + Any + Send + Sync;
-    type ComputePipeline:     Debug + Any + Send + Sync;
+    type ShaderLib:           Debug + Any + Send + Sync + Resource<Self>;
+    type RenderPass:          Debug + Any + Send + Sync + Resource<Self>;
+    type PipelineLayout:      Debug + Any + Send + Sync + Resource<Self>;
+    type GraphicsPipeline:    Debug + Any + Send + Sync + Resource<Self>;
+    type ComputePipeline:     Debug + Any + Send + Sync + Resource<Self>;
     type UnboundBuffer:       Debug + Any + Send + Sync;
-    type Buffer:              Debug + Any + Send + Sync;
+    type Buffer:              Debug + Any + Send + Sync + Resource<Self>;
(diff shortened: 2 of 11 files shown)


src/backend/metalll/src/native.rs
```diff
@@ -235,6 +235,39 @@ mod fence_related {
     pub struct Fence {}
 }
 
+macro_rules! impl_resource {
```
kvark: can we move that macro into core?

JohnColanduoni:
No, I intended it for non-backend clients of the corell library that need to handle resources manually. I'm not sure if the corell-based render crate will aggregate resources enough for this to be valuable, but I came up with this while writing a procedural macro to handle resource destruction for a project I'm moving to using corell directly.

Basically I have structures containing multiple corell resources (static resources related to the same part of the render pipeline, multiple resources associated with a particular geometry, etc.) that need to implement explicit destruction via a factory instance. I can't use Drop for the same reason corell doesn't; it would require reference counting a handle to the Factory. To make this less error-prone, I wrote a custom derive for a trait Destroyable that I simply derive for each structure that contains resources. It implements a destroy method that recursively calls the destroy method for each field in turn, basically like Drop but with a parameter (the factory).

Unfortunately, it's not possible to implement that trait for the Resources associated types, even for a specific backend, due to the coherency issues I mentioned. I would need to use wrapper structs, which would then break various places where corell requires you to pass slices, as well as be incredibly tedious. I suspect that other projects using corell without implementing reference counting on top would run into the same issue.

kvark:
Oh, thanks for the explanation! It makes more sense to me now.
Could you elaborate on that further?
> This functionality is basically impossible to implement on top of corell due to coherency issues; any implementations of a trait for both Resources::X and Resources::Y will be marked as conflicting even if they never do in actuality.

JohnColanduoni:
If I have some trait Destroyable defined in my crate and try to implement it via either:

```
impl<R: Resources> Destroyable<R> for R::Image {
...
}
```

or

```
use gfx_device_vulkanll::Resources as R;
impl Destroyable<R> for R::Image {
}
```

It works just fine... until I try to add (e.g.) a similar implementation for RenderTargetView, 
and the compiler complains that the two implementations conflict. In theory Image and RenderTargetView could be the same type, and there's no way to assert that they aren't.

kvark:
Ok, I see now. I believe this makes sense then.
@JohnColanduoni could you make `Resource` having `Debug + Any` bound, since they always come together anyway?

JohnColanduoni:
Sure, I'll move the macro (and the resource/destructor method list) into corell as well.

src/corell/src/lib.rs
```diff
@@ -261,30 +261,35 @@ pub trait SwapChain {
...
-    type ShaderResourceView:  Debug + Any + Send + Sync;
-    type UnorderedAccessView: Debug + Any + Send + Sync;
-    type RenderTargetView:    Debug + Any + Send + Sync;
-    type DepthStencilView:    Debug + Any + Send + Sync;
-    type FrameBuffer:         Debug + Any + Send + Sync;
-    type Sampler:             Debug + Any + Send + Sync;
-    type Semaphore:           Debug + Any + Send + Sync;
-    type Fence:               Debug + Any + Send + Sync;
-    type Heap:                Debug + Any;
+    type Image:               Debug + Any + Send + Sync + Resource<Self>;
```
kvark: we need to remove `Debug+Any` bounds from here now

kvark:
Thanks!

PR #1320, opened 2017-06-17
Fix depth pixel format in GL

matthew-mcallister:
The mistake was using `GL_DEPTH`, which is the depth buffer handle, instead of one of the [allowed pixel formats](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexSubImage2D.xhtml).


kvark:
Nice find, thank you!
@homu r+

kvark:
Hmm, homu is not responding

Issue #1321, opened 2017-06-17
[ll] Integration: Device [01/03]


msiglreith:
_Metaissue for tracking the progress of the remaining issues for the first `ll` integration batch #1102 ._
_This integration step includes changes to window setup, device creation and command buffer submission_

- [x] Implement dx12 backend
- [x] Implement vulkan backend
- [x] Implement metal backend
- [x] Update glfw window
- [x] Update SDL2 window
- [ ] Improve error handling
- [ ] Improve documentation (descriptions, restrictions, examples, and possible module splits)
- [x] render: Allow setting synchronization primitives for `Encoder` flushing (e.g sync swapchain image acquistion)
- [ ] Swapchain recreation/resizing
- [x] Deprecate queues `wait_idle` (GPU side fence signal + CPU side wait)
- [x] Add handle for semaphores
- [x] Update examples
  - [x] `blend`
  - [x] `deferred`
  - [x] `flowmap`
  - [x] `gamma`
  - [x] `instancing`
  - [x] `mipmap`
  - [x] `particle`
  - [x] `performance`
  - [x] `shadow`
  - [x] `terrain`
  - [x] `terrain_tessellated`
  - [x] `triangle`
  - [x] `ubo_tilemap`

msiglreith:
As there seems to confusion regarding the `ll` branch, `corell` and the integration process I created some slides which hopefully communicate the current process better.
[gfx-rs LL integration.pdf](https://github.com/gfx-rs/gfx/files/1145735/gfx-rs.LL.integration.pdf)

msiglreith:
Swapchain recreation the only point missing from this list besides better error handling and some more love for the documentation. Recreation was harder to implement due to our handle based API.
Current development happens inside the `ll` branch which includes the integration of the other 2/3 of the process.

PR #1322, opened 2017-06-18
[ll] Update examples [17/..]

msiglreith:
* Update blend, flowmap and skybox examples
* Remove `gfx_core` dependency from cube example

cc #1321

### DIFF
--- a/examples/cube/main.rs
+++ b/examples/cube/main.rs
@@ -16,9 +16,8 @@ extern crate cgmath;
 #[macro_use]
 extern crate gfx;
 extern crate gfx_app;
-extern crate gfx_core;
 
-pub use gfx_app::{BackbufferView, ColorFormat, DepthFormat};
+use gfx_app::{BackbufferView, ColorFormat, DepthFormat};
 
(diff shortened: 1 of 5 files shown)


examples/cube/main.rs
```diff
@@ -167,9 +166,9 @@ impl<B: gfx::Backend> gfx_app::Application<B> for App<B> {
         }
     }
 
-    fn render<Gp>(&mut self, (frame, semaphore): (gfx_core::Frame, &<B::Resources as gfx::Resources>::Semaphore),
-                  pool: &mut Gp, queue: &mut gfx_core::queue::GraphicsQueueMut<B>)
-        where Gp: gfx_core::GraphicsCommandPool<B>
+    fn render<Gp>(&mut self, (frame, semaphore): (gfx::Frame, &<B::Resources as gfx::Resources>::Semaphore),
```
kvark: hopefully, the semaphore will eventually become something like `handle::Semaphore<B::Resources>`

msiglreith: Ah right, I will set it on the issue list!

Issue #1328, opened 2017-06-21
[pre-ll] Select proper buffer format when viewing as SRV


kvark:
Current code:
```rust
let format = gl::R8; //TODO: get from the buffer handle
```

cc @HeroesGrave

kvark:
Viewing buffer as SRV is not implemented yet in gfx-hal, closing the issue. We can open a new one for implementing the thing in the first place.

PR #1329, opened 2017-06-22
Implement buffer texture format selection for GL backend

HeroesGrave:
"Fixes" #1328 (although buffer textures aren't implemented at all for other backends)

Hopefully I managed to update all the function signatures for the different backends. I could only test the GL backend on my computer.

### DIFF
--- a/src/backend/gl/src/factory.rs
+++ b/src/backend/gl/src/factory.rs
@@ -18,7 +18,7 @@ use std::{slice, ptr};
 use {gl, tex};
 use core::{self as d, factory as f, texture as t, buffer, mapping};
 use core::memory::{self, Bind, SHADER_RESOURCE, UNORDERED_ACCESS, Typed};
-use core::format::ChannelType;
+use core::format::{ChannelType, Format};
 use core::handle::{self, Producer};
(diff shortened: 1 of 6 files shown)


src/backend/gl/src/factory.rs
```diff
@@ -436,12 +436,13 @@ impl f::Factory<R> for Factory {
...
 
-    fn view_buffer_as_shader_resource_raw(&mut self, hbuf: &handle::RawBuffer<R>)
+    fn view_buffer_as_shader_resource_raw(&mut self, hbuf: &handle::RawBuffer<R>, format: Format)
                                       -> Result<handle::RawShaderResourceView<R>, f::ResourceViewError> {
         let gl = &self.share.context;
         let mut name = 0 as gl::types::GLuint;
         let buf_name = *self.frame_handles.ref_buffer(hbuf);
-        let format = gl::R8; //TODO: get from the buffer handle
+        let format = tex::format_to_glfull(format)
+            .map_err(|_| f::ResourceViewError::Unsupported)?;
```
kvark: too much sugar for my taste, given that a single `match` would be almost as concise

PR #1330, opened 2017-06-22
[ll] D3D12 backend MVP

### LINKED ISSUE #1353: [ll] Mapping size/offset semantics
Currently, `write_mapping` gets the `offset` and `size` parameters in `T` units. This appears slightly inconsistent with other `Factory` methods (e.g. `create_buffer`) that work with bytes. I believe that even though we technically can treat the buffer as typed, given the `T` knowledge, we should be consistent with our `offset` and `size`, and treat them in bytes.

@msiglreith thoughts?

kvark:
Fixes #1353
Also brings some improvements to core and metal backends.

### DIFF
--- a/src/backend/dx12ll/src/command.rs
+++ b/src/backend/dx12ll/src/command.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use comptr::ComPtr;
-use std::{cmp, mem, ptr, ops};
+use std::{cmp, mem, ptr};
 use winapi::{self, UINT, UINT64, FLOAT};
 
 use core::{self, command, memory, pso, state, target, IndexType, VertexCount, VertexOffset};
--- a/src/backend/dx12ll/src/command.rs
+++ b/src/backend/dx12ll/src/command.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use comptr::ComPtr;
-use std::{mem, ptr, ops};
+use std::{cmp, mem, ptr, ops};
 use winapi::{self, UINT, UINT64, FLOAT};
 
 use core::{self, command, memory, pso, state, target, IndexType, VertexCount, VertexOffset};
@@ -110,7 +110,6 @@ impl CommandBuffer {
     fn copy_buffer_to_image(&mut self, buffer: &native::Buffer, image: &native::Image,
                             _layout: memory::ImageLayout, regions: &[command::BufferImageCopy]) {
         //TODO: handle layout?
-        assert_eq!(image.row_pitch % winapi::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT as usize, 0);
         let mut src = winapi::D3D12_TEXTURE_COPY_LOCATION {
             pResource: buffer.resource.as_mut_ptr(),
             Type: winapi::D3D12_TEXTURE_COPY_TYPE_PLACED_FOOTPRINT,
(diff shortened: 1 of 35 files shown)


src/backend/dx12ll/src/factory.rs
```diff
@@ -97,17 +97,71 @@ impl Factory {
         }
         Ok(native::ShaderLib { shaders: shader_map })
     }
+
+    pub fn create_descriptor_heap_impl(device: &mut ComPtr<winapi::ID3D12Device>,
```
msiglreith: Could you annotate that this as an non-shader visible/cpu heap?

src/backend/dx12ll/src/lib.rs
```diff
@@ -130,8 +136,26 @@ impl core::Adapter for Adapter {
...
+        // https://msdn.microsoft.com/en-us/library/windows/desktop/dn788678(v=vs.85).aspx
+        let heap_types = vec![
+            core::HeapType {
+                id: 0,
+                properties: memory::DEVICE_LOCAL,
+                heap_index: 1,
+            },
+            core::HeapType {
+                id: 1,
+                properties: memory::CPU_VISIBLE | memory::CPU_CACHED,
```
msiglreith: Does D3D12 have non-coherent memory? Vulkan requires to explicitly flush non-coherent memory via `vkFlush/InvalidateMappedMemoryRanges`. I don't recall if there is a similar restriction on d3d12 (map/unmap?).

kvark:
Progress is real, landing imminent:
gfx-dx12

src/backend/dx12ll/src/command.rs
```diff
@@ -1,575 +1,726 @@
...
+    }
+
+    fn copy_image_to_buffer(&mut self) {
+        unimplemented!()
+    }
+
+    fn clear_color(&mut self, rtv: &native::RenderTargetView, value: command::ClearColor) {
+        let clear_color = match value {
+            command::ClearColor::Float(c) => c,
+            command::ClearColor::Int(c) => [c[0] as FLOAT, c[1] as FLOAT, c[2] as FLOAT, c[3] as FLOAT], // TODO: error?
```
msiglreith: (Casts are probably wrong here, requiring to keep the same bits but that's something we rather need to test in the future..)

src/backend/dx12ll/src/command.rs
```diff
@@ -1,575 +1,726 @@
...
+macro_rules! impl_render_pass_inline_buffer {
+    ($graphics_buffer:ident) => {
+        impl command::RenderPassInlineBuffer<$graphics_buffer, R> for RenderPassInlineBuffer {
+            fn begin(command_buffer: &mut Encoder<$graphics_buffer>,
+                     render_pass: &RenderPass,
+                     framebuffer: &FrameBuffer,
+                     render_area: target::Rect,
+                     clear_values: &[command::ClearValue]
+            ) -> Self {
+				let color_views = framebuffer.color.iter().map(|view| view.handle).collect::<Vec<_>>();
```
msiglreith: nit: formatting

src/backend/dx12ll/src/command.rs
```diff
@@ -1,575 +1,726 @@
...
+       			assert!(framebuffer.depth_stencil.len() <= 1);
```
msiglreith: Whats the motivation behind this check?

src/backend/dx12ll/src/native.rs
```diff
@@ -115,28 +125,92 @@ pub struct DepthStencilView {
...
+#[derive(Clone, Debug)]
+pub struct DualHandle {
+    pub cpu: winapi::D3D12_CPU_DESCRIPTOR_HANDLE,
+    pub gpu: winapi::D3D12_GPU_DESCRIPTOR_HANDLE,
+}
+
+#[derive(Clone, Debug)]
 pub struct DescriptorHeap {
+    pub inner: ComPtr<winapi::ID3D12DescriptorHeap>,
+    //TODO: hide members?
```
msiglreith: Sounds like a good place for `pub(crate)` 😄

kvark:
Looks green enough :) Merging, at last!

Issue #1304, opened 2017-06-10
[ll] [RFC] GL Swapchains


msiglreith:
In contrast to the other APIs, GL doesn't differ between device and swapchain as two independent entities, but merged them as one Context object.

Our current design for device and swapchain creation looks like this:
```Rust
let window = WindowBuilder::build();
let (surface, adapters) = window.create_adapters_and_surface();
// find adapter which supports surfaces and a queue family
// which supports present on this surface ..
let swapchain = surface.build_swapchain(config);
let device = adapters[0].open();
```
This approach is based on vulkans model, but causes some issue with GL (and glutin):
* Both swapchain and device need access to the GL context
* We have to create the context at swapchain creation to correctly set the pixel formats (swapchain -> adapter)
* Swapchain creation requires the specific present queue instance for dx12 (adapter -> swapchain)

Therefore I propose that we should create the swapchain and the adapter at the same time to avoid these conflicts:
```Rust
// Extend adapter trait
pub trait Adapter<B: Backend>: Sized {
    /// Create a new device and command queues.
    fn open(&self, queue_descs: &[(&B::QueueFamily, u32)]) -> Device<B>;

   /// NEW!
   // The present queue will be the first queue instance of present queue family
   fn open_with_swapchain<S>(&self, queue_descs: &[(&B::QueueFamily, u32)],
           swapchain_config: SwapchainConfig, present_queue_family: &B::QueueFamily)
         -> (Device<B>, S::Swapchain) where S: Surface<B>;

    /// Get the `AdapterInfo` for this adapater.
    fn get_info(&self) -> &AdapterInfo;

    /// Return the supported queue families for this adapter.
    fn get_queue_families(&self) -> &[B::QueueFamily];
}
```

The mentioned issue with `glutin`:
Context and window creation are coupled, but we would need to first create the window and later build a context for the window on `open_with_swapchain`.
SDL2 seems to support this, so I'm quite confident that the glutin API could be adapted to support this (which I would try to do).

kvark:
> Rethinking the issue: for the moment we could also simply enforce the user to set the pixel format on start for GL and only allow matching formats for the swapchain. This could be relaxed on a later stage with the described glutin changes, but it's not high priority imo.

I suppose that's the consensus then ;)

msiglreith:
Latest idea was to hide the actual default framebuffer of OpenGL and only expose custom created images.
Pro:
- Less issues with sRGB
- Avoid magic multisampling resolves of GL driver
[...]

PR #1335, opened 2017-06-23
[ll] Update glfw and sdl window [21/..]

msiglreith:
Adapt glfw and sdl window creation to the new model (Window-> Surface -> SwapChain).
The swapchains of these backends are restricted as mentioned in #1304 (set surface format at the beginning and swapchain config has to match those).

Backbuffer image creation could be abstracted in a followup PR as it's similar on all three (glutin, glfw, sdl) window backends.

Edit: I need to fix sdl build on my linux system later - done


msiglreith:
Should be ready now

PR #1337, opened 2017-06-27
Add layout specifiers to some examples that didn't have them.

icefoxen:
Needs verification from someone who knows what they're doing, though.


kvark:
@homu r+

kvark:
Thanks!

PR #1338, opened 2017-06-28
[ll] Update some examples and add synced flushing for encoder [23/..]

msiglreith:
* render: Add `synced_flush`, which works like `flush` but additionally allows to specify synchronization primitives. Important for ensuring that the acquired backbuffer image is available for rendering.
* Update particle, terrain and instancing example
* Fix `handle` test
* Add dx11 and dx12 test to appveyor

cc #1321


examples/instancing/main.rs
```diff
@@ -146,29 +150,37 @@ impl<R: gfx::Resources> gfx_app::Application<R> for App<R> {
...
+            views: window_targets.views,
         }
     }
 
-    fn render<C: gfx::CommandBuffer<R>>(&mut self, encoder: &mut gfx::Encoder<R, C>) {
+    fn render<Gp>(&mut self, (frame, semaphore): (gfx::Frame, &gfx::handle::Semaphore<B::Resources>),
+                  pool: &mut Gp, queue: &mut gfx::queue::GraphicsQueueMut<B>)
+        where Gp: gfx::GraphicsCommandPool<B>
+    {
+        let mut encoder = pool.acquire_graphics_encoder();
```
kvark: can `gfx::Application` hide the acquiring and flushing of `Encoder` for us?

msiglreith: Possible, a valid request as a single Encoder the common case for our examples. I'll look into it after updated the other examples.

PR #1345, opened 2017-07-04
Added generate_mipmap

kvark:
This is very basic and just exposes the existing API of `CommandBuffer`.


kvark:
@ebkalderon good point! Added now.

kvark:
Ok, published in 0.16.1 now

kvark:
Now that I think of it, T should have a BlendFormat bound.

PR #1355, opened 2017-07-06
[ll] Improve documentation and formatting [26/..]

msiglreith:
* Improve formatting of core
* Improve capabilities documentation
* Remove unused dx11 code

### DIFF
--- a/src/core/src/lib.rs
+++ b/src/core/src/lib.rs
@@ -33,23 +33,21 @@ extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 
-use std::fmt::{self, Debug};
-use std::error::Error;
-use std::hash::Hash;
+use format::Formatted;
 use std::any::Any;
 use std::borrow::Borrow;
-use format::Formatted;
+use std::error::Error;
+use std::fmt::{self, Debug};
+use std::hash::Hash;
 
-pub use draw_state::{state, target};
 pub use self::command::CommandBuffer;
 pub use self::factory::Factory;
-pub use self::pool::{
-    GeneralCommandPool, GraphicsCommandPool, ComputeCommandPool,
-    TransferCommandPool, SubpassCommandPool};
-pub use self::queue::{
-    GeneralQueue, GraphicsQueue, ComputeQueue, TransferQueue};
-pub use self::window::{
-    Backbuffer, Frame, FrameSync, Surface, SwapChain, SwapchainConfig, WindowExt};
+pub use self::pool::{ComputeCommandPool, GeneralCommandPool, GraphicsCommandPool,
+                     SubpassCommandPool, TransferCommandPool};
+pub use self::queue::{ComputeQueue, GeneralQueue, GraphicsQueue, TransferQueue};
+pub use self::window::{Backbuffer, Frame, FrameSync, Surface, SwapChain, SwapchainConfig,
+                       WindowExt};
+pub use draw_state::{state, target};
 
 pub mod buffer;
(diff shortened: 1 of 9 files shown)


nical:
If you want to improve the formatting you might as well move to block indent instead of visual indents since that's the official rust style now (assuming gfx-rs is to follow the official rust style guide).

src/core/src/lib.rs
```diff
@@ -145,25 +145,38 @@ impl<R: Resources> ShaderSet<R> {
...
 /// Features that the device supports.
-#[allow(missing_docs)] // pretty self-explanatory fields!
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
 pub struct Capabilities {
-    pub max_vertex_count: usize,
-    pub max_index_count: usize,
+    /// Recommended maximum number of vertices per draw call.
+    /// If there is no recommendation, the value is `None`.
+    pub max_vertex_count: Option<VertexCount>,
```
kvark: mm, I don't see it as "recommended" but rather as a hard limit, in which case having an `Option` doesn't make sense

src/core/src/lib.rs
```diff
@@ -145,25 +145,38 @@ impl<R: Resources> ShaderSet<R> {
...
 
+    /// Support offsets for instanced drawing.
     pub instance_base_supported: bool,
+    /// Support instanced drawing.
     pub instance_call_supported: bool,
+    /// Support manually specified vertex attribute rates (divisors).
     pub instance_rate_supported: bool,
+    /// Support base vertex offset for indexed drawing.
     pub vertex_base_supported: bool,
+    /// Support sRGB rendertargets.
```
kvark: ... and textures

src/core/src/lib.rs
msiglreith: It seems to be only used in the gl backend and the refpages state that it's only a recommendation (above the limit the performance may decrease).
We could also switch to `VertexCount::MAX` instead of `Option` for the other backends.

msiglreith:
I switched to block indents as proposed by @nical, I'll squash the commits at the end.

src/core/src/lib.rs
```diff
@@ -145,25 +145,38 @@ impl<R: Resources> ShaderSet<R> {
...
     pub srgb_color_supported: bool,
+    /// Support constant buffers.
     pub constant_buffer_supported: bool,
+    /// Support unordered-access views.
     pub unordered_access_view_supported: bool,
+    /// Support specifying the blend function and equation for each color target.
     pub separate_blending_slots_supported: bool,
+    /// Support accelerated buffer copy. (unused)
     pub copy_buffer_supported: bool,
```
msiglreith: Should we remove this because it's not used at all?

src/core/src/lib.rs
kvark: Ah, I see that D3D11 just have those counts as 2^32. How about we remove them entirely then?

src/core/src/lib.rs
kvark: We do have `CommandBuffer::copy_buffer` so the proper course of action would be check for the support instead of removing the flag

msiglreith: Ah, right! I'll open a separate issue for it.

msiglreith:
All proposed changes should have been addressed, thanks for the reviews!

kvark:
Thanks!

PR #1357, opened 2017-07-06
[ll] Update triangle example [27/..]

msiglreith:
- Port the triangle example to the new core. Due to the ll changes this requires a lot more boilerplate for the setup. Interesting direction would be providing some shortcuts for the user for specific scenarios (e.g single queue setup)

- Add API support for waiting on semaphores on present to allow patterns like acquire frame -> wait -> draw -> signal -> present. Not implemented for the backends so far (vk and dx12 are non-functional atm anyway..)

cc #1321

### DIFF
--- a/examples/triangle/main.rs
+++ b/examples/triangle/main.rs
@@ -102,7 +102,8 @@ pub fn main() {
     ).unwrap();
     let (vertex_buffer, slice) = factory.create_vertex_buffer_with_slice(&TRIANGLE, ());
     let mut graphics_pool = <device_gl::Backend as gfx::Backend>::GraphicsCommandPool::from_queue(graphics_queue.as_ref(), 1);
-    let semaphore = factory.create_semaphore();
+    let frame_semaphore = factory.create_semaphore();
+    let draw_semaphore = factory.create_semaphore();
 
     let mut data = pipe::Data {
         vbuf: vertex_buffer,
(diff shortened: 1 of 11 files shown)


examples/triangle/main.rs
```diff
@@ -43,26 +49,68 @@ const TRIANGLE: [Vertex; 3] = [
...
+        swap_chain
+            .get_backbuffers()
+            .iter()
+            .map(|&(ref color, ref ds)| {
+                let color_desc = texture::RenderDesc {
+                    channel: ColorFormat::get_format().1,
+                    level: 0,
+                    layer: None,
+                };
+                let rtv = factory.view_texture_as_render_target_raw(color, color_desc)
```
kvark: hmm, why is `color` a raw texture (leading to the need to use `device_gl::Resources>`)? It's taken from `swap_chain`, which builds from `config`, which has the color type, technically.

msiglreith: The color format in `SwapchainConfig` is currently set on runtime, so that we could query for support the surface in the future in case the prefered format might not be supported.

kvark: It is almost required to get rid of the direct device_gl dependency. Do you see a way to get this within the PR, or would you prefer to address this later?

msiglreith:
Added an extension trait for SwapChains allowing to easier create the rtvs, removing the explicit device_gl notation.
Still directly relays on the gl crate due to command pool creation, which I dont know how to address atm.

msiglreith:
Shall I add https://github.com/msiglreith/gfx/commit/908b1d63d4f97700c47ba0252fd90a33d014e6d0 to this PR in order to remove the `device_gl` direct dep?

kvark:
Oh wow. Let's do it.

msiglreith:
Done! :bell:

kvark:
Please go ahead and merge.

PR #1359, opened 2017-07-08
[ll] Refactor command pools [28/..]

msiglreith:
- Refactor command pools following the command buffer refactoring by splitting pools into `RawCommandPools` and `SubpassCommandPools`. These changes ares motivated after noticing in the backend implementation that every backend uses macros for implementation of the CommandPool types for primary command buffers.
`General/Graphics/Compute/TransferCommandPool` are now safe-wrappers around `RawCommandPool`, `SubpassCommandPool` stays as separate pool trait but currently doesn't provide any functionality as the design isn't fleshed out and it doesn't block #1321 
- Remove `Compatible` queue trait as this is now encoded by the `create_command_pool` functions (https://github.com/gfx-rs/gfx/pull/1357)
- Update gamma example, similar to triangle
- Cleanup queue resources after each frame


src/backend/dx11/src/pool.rs
```diff
@@ -14,77 +14,62 @@
...
 
-macro_rules! impl_pool {
-    ($pool:ident, $queue:ident, $buffer:ident) => (
-        pub struct $pool {
-            command_buffers: Vec<$buffer<Backend>>,
-            next_buffer: usize,
-        }
+pub struct RawCommandPool {
+    command_buffers: Vec<RawCommandBuffer<CommandList>>,
+    next_buffer: usize,
```
kvark: good to see less macros usage! 👍

src/core/src/command.rs
```diff
@@ -50,38 +50,38 @@ impl<B: Backend> Submit<B> {
 ///
 /// Pools will always return an Encoder on `acquire_command_buffer` to provide a safe interface.
 #[derive(Debug)]
-pub struct Encoder<'a, B: Backend, C: CommandBuffer<B> + 'a>(&'a mut C, PhantomData<B>);
+pub struct Encoder<B: Backend, C: CommandBuffer<B>>(C, PhantomData<B>);
```
kvark: oh wow, pretty! I think we should not specify `C` here though (even shorter!):
```rust
pub struct Encoder<B, C>(C, PhantomData<B>);
```
I know it may sound controversial. On one hand, this version is shorter and allows any higher-level wrappers to be shorter too. On the other hand, specifying the dependencies here allows for earlier error catching and clearer error messages. That point is rather weak though, and I recall the recommendation (not to use the bounds on struct definitions) to be in the guidelines.

src/core/src/command.rs
```diff
@@ -50,38 +50,38 @@ impl<B: Backend> Submit<B> {
...
     fn deref(&self) -> &C {
-        self.0
+        &self.0
     }
 }
 
-impl<'a, B, C> DerefMut for Encoder<'a, B, C>
-    where B: Backend, C: CommandBuffer<B> + 'a
+impl<B, C> DerefMut for Encoder<B, C>
+    where B: Backend, C: CommandBuffer<B>
```
kvark: correspondingly, this `where` clause will not be needed

src/core/src/pool.rs
```diff
@@ -14,90 +14,80 @@
...
-
-/// Subpass command pool can allocate subpass command buffers.
-pub trait SubpassCommandPool<B: Backend>: CommandPool<B> {
-    ///
-    fn from_queue<Q>(queue: Q, capacity: usize) -> Self
-    where
-        Q: Compatible<GraphicsQueue<B>> + AsRef<B::CommandQueue>;
-
+///
+pub struct TransferCommandPool<B: Backend>(pub(crate) B::RawCommandPool);
```
kvark: oh, aren't you quick to use the new shiny features? 😁  I don't mind at all, just need to get used to it as well

examples/gamma/main.rs
```diff
@@ -50,18 +52,43 @@ pub fn main() {
...
+    let window = gfx_window_glutin::build(builder, &events_loop, ColorFormat::get_format(), DepthFormat::get_format());
+
+    let (mut surface, adapters) = gfx_window_glutin::Window(&window).get_surface_and_adapters();
+    let queue_descs = adapters[0].get_queue_families().iter()
+                                 .filter(|family| surface.supports_queue(&family) )
+                                 .map(|family| { (family, 1) })
+                                 .collect::<Vec<_>>();
+
+    // Open device (factory and queues)
+    let gfx::Device { mut factory, mut general_queues, mut graphics_queues, .. } = adapters[0].open(&queue_descs);
```
msiglreith: Looks good!

examples/blend/main.rs
```diff
@@ -148,9 +148,8 @@ impl<B: gfx::Backend> gfx_app::Application<B> for App<B> {
         }
     }
 
-    fn render<Gp>(&mut self, (frame, semaphore): (gfx::Frame, &gfx::handle::Semaphore<B::Resources>),
-                  pool: &mut Gp, queue: &mut gfx::queue::GraphicsQueueMut<B>)
-        where Gp: gfx::GraphicsCommandPool<B>
+    fn render(&mut self, (frame, semaphore): (gfx::Frame, &gfx::handle::Semaphore<B::Resources>),
+              pool: &mut gfx::GraphicsCommandPool<B>, queue: &mut gfx::queue::GraphicsQueueMut<B>)
```
msiglreith: I think this would be possible, `shadow` requires additional command pools, which should be created at initialization and probably some additional synchronization.

examples/gamma/main.rs
msiglreith: I just understand correctly what you meant with the second option (in particular with removing 10 lines). Imo the advantage of this is too small to add it as additional function. In order to remove the queue cast below we would need to change the queue family, which we could consider in the future.

Just requesting a `GraphicsCommandQueue` could be conflicting with some backends like GL. The current approach I'm aiming at for GL is to either export a `GraphicsQueue` or `GeneralQueue` depending on compute support.

PR #1360, opened 2017-07-10
[ll] Remove trait bounds and simplify device opening [29/..]

msiglreith:
Follow to #1359 
- Remove `Encoder` trait bounds
- Add `open_with` to `Adapter` which takes an closure returning the number of queues for each family

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -149,11 +149,8 @@ fn run<A, B, S, EL>((width, height): (u32, u32),
     use gfx::texture;
 
     // Init device
-    let queue_descs = adapters[0].get_queue_families().iter()
-                                 .filter(|family| surface.supports_queue(&family) )
-                                 .map(|family| { (family, family.num_queues()) })
-                                 .collect::<Vec<_>>();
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -149,11 +149,8 @@ fn run<A, B, S, EL>((width, height): (u32, u32),
     use gfx::texture;
 
     // Init device
-    let queue_descs = adapters[0].get_queue_families().iter()
-                                 .filter(|family| surface.supports_queue(&family) )
-                                 .map(|family| { (family, family.num_queues()) })
-                                 .collect::<Vec<_>>();
-    let gfx_core::Device { mut factory, mut general_queues, mut graphics_queues, .. } = adapters[0].open(&queue_descs);
+    let gfx_core::Device { mut factory, mut general_queues, mut graphics_queues, .. } =
+        adapters[0].open_with(|family| if surface.supports_queue(&family) { 1 } else { 0 });
```
kvark: we can even do `surface.supports_queue(&family) as u32`

PR #1362, opened 2017-07-11
Update dependencies & examples to glutin 0.9 / winit 0.7

### LINKED ISSUE #1348: Update cocoa dependencies
This should make Travis happy again.

alexheretic:
This PR brings gfx up to date with the latest glutin & winit releases. Having a `ContextBuilder` in addition to the `WindowBuilder` and a rework of the event structures are the main changes for gfx users. The events are particularly groansome for refactoring. I'm particularly interested in getting gfx up to date so I can finally have XWayland window support in Linux. See [glutin ContextBuilder addition PR](https://github.com/tomaka/glutin/pull/901) & [winit event rework PR](https://github.com/tomaka/winit/pull/164)

Fixes #1348

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,22 +110,23 @@ impl Factory<gfx_device_gl::Resources> for gfx_device_gl::Factory {
     }
 }
 
-pub fn launch_gl3<A>(wb: winit::WindowBuilder) where
+pub fn launch_gl3<A>(window: winit::WindowBuilder) where
 A: Sized + ApplicationBase<gfx_device_gl::Resources, gfx_device_gl::CommandBuffer>
 {
     use gfx::traits::Device;
+    use glutin::GlContext;
 
     env_logger::init().unwrap();
     let gl_version = glutin::GlRequest::GlThenGles {
         opengl_version: (3, 2), // TODO: try more versions
         opengles_version: (2, 0),
     };
-    let builder = glutin::WindowBuilder::from_winit_builder(wb)
-                                        .with_gl(gl_version)
-                                        .with_vsync();
-    let events_loop = glutin::EventsLoop::new();
+    let context = glutin::ContextBuilder::new()
+        .with_gl(gl_version)
+        .with_vsync(true);
+    let mut events_loop = glutin::EventsLoop::new();
     let (window, mut device, mut factory, main_color, main_depth) =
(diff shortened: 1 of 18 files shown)


brendanzab:
Updated the description so that #1348 will auto-close.

alexheretic:
I've amended the commit to fix Vulkan compile issue, and apply the same fixes to mac/windows. I couldn't actually test these though.

src/lib.rs
```diff
@@ -207,18 +217,27 @@ A: Sized + ApplicationBase<gfx_device_dx11::Resources, D3D11CommandBuffer>
...
-                winit::WindowEvent::KeyboardInput(winit::ElementState::Pressed, _, key, _) if key == A::get_exit_key() => return,
-                winit::WindowEvent::Resized(width, height) => {
-                    let size = (width as gfx::texture::Size, height as gfx::texture::Size);
-                    if size != window.size {
-                        // working around the borrow checker: window is already borrowed here
-                        new_size = Some(size);
-                    }
-                },
-                _ => app.on(event),
+        events_loop.poll_events(|event| {
```
kvark: any reason the `poll_events` blocks are rewritten? is it `rustfmt` doing the work?

alexheretic: Code had to be reworked to handle the [`winit::Event`](https://docs.rs/winit/0.7.2/winit/enum.Event.html) rework. Events seem to be split into Device & Window now, plus in general all winit events are more nested and have switched to named variable enums.

I've simply reworked the existing code into the new structure.

mitchmindtree:
@alexheretic you might have already addressed this somewhere else, but I couldn't find it in this PR so just in case I wanted to add that `GlContext::resize` must be called now when either `EventsLoop::poll_events` or `EventsLoop::run_forever` receives `Resized` events, e.g. [see here for a demonstration in the updated glutin example](https://github.com/tomaka/glutin/pull/901/files#diff-04c6e90faac2675aa89e2176d2eec7d8R69). This is because the `glutin::Context` does not know about events being emitted by the eventloop, and on some platforms (macos comes to mind) GL contexts require being updated when the window is resized.

Originally this was not necessary as glutin wrapped the entire `EventsLoop` and so this could be hidden from the glutin user.

alexheretic:
Thanks @mitchmindtree, nice spot. Not *ideal* from a consumers perspective. I'll raise a PR to fix the examples.

PR #1366, opened 2017-07-12
[ll] Update performance and change windows [30/..]

msiglreith:
- Update the performance example
- Optimize GL submission and fix command buffer reset
- Windows now take ownership of the raw windows
- Fix rebasing issues

cc #1321

### DIFF
--- a/examples/gamma/main.rs
+++ b/examples/gamma/main.rs
@@ -53,16 +53,12 @@ pub fn main() {
         .with_dimensions(1024, 768)
         .with_vsync();
     let win = gfx_window_glutin::build(builder, &events_loop, ColorFormat::get_format(), DepthFormat::get_format());
-    let mut window = gfx_window_glutin::Window::new(window);
+    let mut window = gfx_window_glutin::Window::new(win);
 
     let (mut surface, adapters) = window.get_surface_and_adapters();
-    let queue_descs = adapters[0].get_queue_families().iter()
-                                 .filter(|family| surface.supports_queue(&family) )
-                                 .map(|family| { (family, 1) })
-                                 .collect::<Vec<_>>();
-
     // Open device (factory and queues)
-    let gfx::Device { mut factory, mut general_queues, mut graphics_queues, .. } = adapters[0].open(&queue_descs);
+    let gfx::Device { mut factory, mut general_queues, mut graphics_queues, .. } =
+        adapters[0].open_with(|family| surface.supports_queue(&family) as u32);
     let mut graphics_queue = if let Some(queue) = general_queues.first_mut() {
         queue.as_mut().into()
     } else if let Some(queue) = graphics_queues.first_mut() {
--- a/examples/gamma/main.rs
+++ b/examples/gamma/main.rs
@@ -52,9 +52,10 @@ pub fn main() {
         .with_title("Gamma example".to_string())
         .with_dimensions(1024, 768)
         .with_vsync();
-    let window = gfx_window_glutin::build(builder, &events_loop, ColorFormat::get_format(), DepthFormat::get_format());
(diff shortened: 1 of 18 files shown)


examples/gamma/main.rs
```diff
@@ -52,9 +52,10 @@ pub fn main() {
         .with_title("Gamma example".to_string())
         .with_dimensions(1024, 768)
         .with_vsync();
-    let window = gfx_window_glutin::build(builder, &events_loop, ColorFormat::get_format(), DepthFormat::get_format());
+    let win = gfx_window_glutin::build(builder, &events_loop, ColorFormat::get_format(), DepthFormat::get_format());
+    let mut window = gfx_window_glutin::Window::new(window);
```
kvark: that must be a mistake? there is no `window`

examples/gamma/main.rs
```diff
@@ -114,6 +117,7 @@ pub fn main() {
         encoder.draw(&slice, &pso, &data);
         encoder.synced_flush(&mut graphics_queue, &[&frame_semaphore], &[&draw_semaphore], None);
         swap_chain.present(&mut graphics_queue, &[&draw_semaphore]);
+        graphics_queue.wait_idle();
```
kvark: that one is gone, btw

examples/performance/main.rs
```diff
@@ -89,100 +91,143 @@ trait Renderer: Drop {
...
+        let events_loop = glutin::EventsLoop::new();
+        let builder = glutin::WindowBuilder::new()
+            .with_title("Triangle example".to_string())
+            .with_dimensions(1024, 768)
+            .with_vsync();
+        let win = gfx_window_glutin::build(builder, &events_loop, ColorFormat::get_format(), DepthStencil::get_format());
+        let mut window = gfx_window_glutin::Window::new(win);
+        // Acquire surface and adapters
+        let (mut surface, adapters) = window.get_surface_and_adapters();
+        let queue_descs = adapters[0].get_queue_families().iter()
```
kvark: can we use the new (shorter) API for these?

examples/performance/main.rs
```diff
@@ -89,100 +91,143 @@ trait Renderer: Drop {
...
+                                    .filter(|family| surface.supports_queue(&family) )
+                                    .map(|family| { (family, 1) })
+                                    .collect::<Vec<_>>();
+
+        // Open device (factory and queues)
+        let gfx::Device { mut factory, mut general_queues, mut graphics_queues, .. } = adapters[0].open(&queue_descs);
+        let mut queue = if let Some(queue) = general_queues.pop() {
```
kvark: what I was thinking is having a wrapper enum (that is either general or graphics queue) that is dereffable to the graphics one. In this case, that giant `if/else` section could be omitted

src/backend/gl/src/command.rs
```diff
@@ -72,9 +72,17 @@ impl DataBuffer {
     }
 }
 
+#[allow(missing_copy_implementations)]
 pub struct SubmitInfo {
-    pub buf: Vec<Command>,
-    pub data: DataBuffer,
+    // Raw pointer optimization:
+    // Command buffers are stored inside the command pools.
```
kvark: a little bit scary, but the comment sounds reasonable

examples/gamma/main.rs
msiglreith: I haven't implemented it for the `ll` branch so far (core, not corell), but it's on my todo!

examples/performance/main.rs
msiglreith: These if/else clauses are really annoying, but I would probably rather try to 'downcast' the queue families upfront so we only get GraphicsQueue.

PR #1368, opened 2017-07-13
[ll] Update shadow example and adjust core for multi-threading

msiglreith:
* core Add necessary trait bounds for pools and backend
* render: Allow to split finish and submit for GraphicsEncoder, allowing to send submission across threads
* shadow: Instead of having an encoder per thread, we now have a pool per thread, record a command buffer on each thread, finish and send back the submit info to the main thread, where we submit the commands atm.

cc #1321

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,5 +19,5 @@ build_script:
 test_script:
   - cargo test --all --features vulkan
   - cargo test -p gfx -p gfx_core --features "mint serialize"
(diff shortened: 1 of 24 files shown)


msiglreith:
Most of the builds looks pretty green besides metal ofc and headless glutin is broken atm.

kvark:
@msiglreith FYI, metal should be green as soon as you rebase on master

PR #1369, opened 2017-07-13
[ll] Support queue family casting and remove queue ref wrappers

msiglreith:
Allow to specify the queue type on initialization (e.g create a graphics queue from a general queue family).
Due to this queue 'casting' the queue reference wrappers become more or less obsolete as the user can now specify the wished queue type upfront and doesn't need to cast after they have been created.

### DIFF
--- a/examples/gamma/main.rs
+++ b/examples/gamma/main.rs
@@ -61,11 +61,7 @@ pub fn main() {
         adapters[0].open_with(|family, ty| {
             ((ty.supports_graphics() && surface.supports_queue(&family)) as u32, gfx::QueueType::Graphics)
         });
-    let mut graphics_queue = if let Some(queue) = graphics_queues.pop() {
-        queue
-    } else {
-        panic!("Unable to find a graphics queue.");
-    };
+    let mut graphics_queue = graphics_queues.pop().expect("Unable to find a graphics queue.");
 
     // Create swapchain
     let config = gfx::SwapchainConfig::new()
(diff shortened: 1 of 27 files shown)


examples/gamma/main.rs
```diff
@@ -57,14 +57,14 @@ pub fn main() {
...
-        adapters[0].open_with(|family| surface.supports_queue(&family) as u32);
-    let mut graphics_queue = if let Some(queue) = general_queues.first_mut() {
-        queue.as_mut().into()
-    } else if let Some(queue) = graphics_queues.first_mut() {
-        queue.as_mut()
+    let gfx::Device { mut factory, mut graphics_queues, .. } =
+        adapters[0].open_with(|family, ty| {
+            ((ty.supports_graphics() && surface.supports_queue(&family)) as u32, gfx::QueueType::Graphics)
+        });
+    let mut graphics_queue = if let Some(queue) = graphics_queues.pop() {
```
kvark: could just be `graphics_queues.pop().expect("...")`

src/backend/vulkan/src/lib.rs
```diff
@@ -63,6 +63,21 @@ pub struct Instance {
     pub surface_extensions: Vec<&'static str>,
 }
 
+fn map_queue_type(flags: vk::QueueFlags) -> QueueType {
+    if flags.intersects(vk::QUEUE_GRAPHICS_BIT | vk::QUEUE_COMPUTE_BIT) { // TRANSER_BIT optional
```
kvark: should be `contains` instead of `intersects`

msiglreith: The bitflag implementation of `ash` doesn't provide `contains` (https://docs.rs/ash/0.18.0/ash/vk/types/struct.QueueFlags.html)

kvark: It has `subset` instead

msiglreith: Ah right, I used the wrong operator. Good find.

PR #1371, opened 2017-07-14
[ll] Headless initialization [33/..]

msiglreith:
Adding support for headless init as alternative to acquiring a surface and adapters from `WindowsExt`.


kvark:
Nice!

PR #1373, opened 2017-07-14
Support field attributes on vertex and constant structures

mzmonsour:
Adds support for attributes on struct fields, such as doc comments. This doesn't work on stable because of a macro parser bug. Wait until rust-lang/rust#42913 is in a stable release (1.20 probably) before merging this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,8 @@
 environment:
   matrix:
-    - TARGET: 1.18.0-x86_64-pc-windows
+    - TARGET: 1.20.0-x86_64-pc-windows
       COMPILER: gnu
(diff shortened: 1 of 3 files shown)


kvark:
@mzmonsour is this ready to ship?
Note that you need to switch appveyor config to use 1.20.
Also, this PR should probably be aimed at `pre-ll` and/or `v0.16` as opposed to `master`

mzmonsour:
The PR should be ready to merge. I tried adjusting the appveyor config, although I may have messed something up in the process of targeting pre-ll, since appveyor is complaining about the PR. Let me know if there's something else I should fix.

PR #1375, opened 2017-07-15
[ll] Replace wait_idle with fence based sync [34/..]

msiglreith:
Mostly based on the current API design from `corell`.
cc #1321 

WIP:
* gl: fence handling needs a bit more polishing
* d3d11: Unsure if we can do GPU-CPU synchronization via fences (queries?) or it's an implementation detail (map/unmap)


msiglreith:
Fences could be implemented for d3d11 via a `QueryEvent` and spinlocking until the query event returns true. d3d11.4 adds support for d3d12 fences.

PR #1380, opened 2017-07-18
Use MTLHeap when possible for resource allocation

JohnColanduoni:
Adds MTLHeap support to the Metal backend, and uses it whenever possible for memory allocations.

This isn't quite ready to merge, https://github.com/gfx-rs/metal-rs/pull/7 needs to make it into crates.io first.

### DIFF
--- a/src/backend/metalll/src/conversions.rs
+++ b/src/backend/metalll/src/conversions.rs
@@ -1,5 +1,5 @@
 use core;
-use core::pass;
+use core::{pass, image, memory};
 use core::format::Format;
 use metal::*;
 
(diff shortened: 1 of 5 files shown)


src/backend/metalll/src/conversions.rs
```diff
@@ -120,3 +120,62 @@ pub fn map_vertex_format(format: Format) -> Option<MTLVertexFormat> {
...
+    if properties.contains(memory::WRITE_COMBINED) {
+        options |= MTLResourceOptionCPUCacheModeWriteCombined;
+    }
+    options
+}
+
+pub fn map_heap_properties_to_storage_and_cache(properties: memory::HeapProperties) -> (MTLStorageMode, MTLCPUCacheMode) {
+    let storage;
+    let cpu;
+    if properties.contains(memory::CPU_VISIBLE) {
```
kvark: any reason you don't do `let cpu = if {..} else {..};` and the same for `storage`?

src/backend/metalll/src/factory.rs
```diff
@@ -45,9 +47,24 @@ impl Drop for Factory {
...
+        MTLFeatureSet::iOS_GPUFamily3_v2,
+        MTLFeatureSet::iOS_GPUFamily1_v4,
+        MTLFeatureSet::iOS_GPUFamily2_v4,
+        MTLFeatureSet::iOS_GPUFamily3_v3,
+
+        MTLFeatureSet::tvOS_GPUFamily1_v2,
+        MTLFeatureSet::tvOS_GPUFamily1_v3,
+
+        MTLFeatureSet::macOS_GPUFamily1_v3,
+    ].iter().cloned().any(|x| device.supports_feature_set(x));
```
kvark: nice!

src/backend/metalll/src/factory.rs
```diff
@@ -642,43 +659,83 @@ impl CoreFactory<Resources> for Factory {
...
+            size: max_size,
+            alignment: max_alignment,
         }
     }
 
-    #[cfg(not(feature = "native_heap"))]
     fn bind_buffer_memory(&mut self, heap: &n::Heap, offset: u64, buffer: n::UnboundBuffer) -> Result<n::Buffer, buffer::CreationError> {
-        Ok(n::Buffer(buffer.0))
+        let bound_buffer;
+        match *heap {
```
kvark: similarly, please convert to functional style aka `let bound_buffer = match *heap {...};`

kvark:
@JohnColanduoni thanks for updating it!
Do you think we still need to merge this in the light of #1384? Or should we postpone the decision until we figure out the plan on how exactly to make `MTLHeap` fitting our API?

JohnColanduoni:
@kvark This at least fixes the handling of memory types (previously everything was allocated in coherent, cpu visible memory) with needs to be done regardless of what we decide on #1384. Also if we decide to use heaps and fall back on driver-managed allocation when that fails we're just a few lines away.

Issue #1379, opened 2017-07-17
Allocating heap memory for low level DX12 backend fails


ConorLPBoyle:
When attempting to compile and run the trianglell example on Windows 10 64 bit MSVC with the default DX12 backend, the program panics at runtime with the following message:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `-2147024809`', src\backend\dx12ll\src\factory.rs:215:8
```
The [create_heap function](https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12ll/src/factory.rs#L215) asserts that
```
assert_eq!(winapi::S_OK, unsafe {
            self.inner.CreateHeap(&desc, &dxguid::IID_ID3D12Heap, &mut heap)
        });
```
Commenting this out, the example simply fails to run without providing a backtrace.
```
error: process didn't exit successfully: `target\debug\examples\trianglell.exe` (exit code: 2173)
```
The example works fine compiled with with `--features vulkan`.

kvark:
Interesting! We don't require that much memory for the heaps, so I wonder why it fails.

It'd be great to see the [DebugView](https://technet.microsoft.com/en-us/sysinternals/debugview.aspx) output from this crash. Debug runtime should have a cleaner error message than just the code.

ConorLPBoyle:
Alright, this should be more helpful.
`
[12408] D3D12 ERROR: ID3D12Device::CreateHeap: D3D12_HEAP_FLAGS has invalid flag combinations set. The following flags may not all be set simultaneously. Exactly one must be left unset, or all may be left unset when the adapter supports D3D12_RESOURCE_HEAP_TIER_2 or creating a heap in conjunction with D3D12_HEAP_FLAG_SHARED_CROSS_ADAPTER: D3D12_FEATURE_DATA_D3D12_OPTIONS::ResourceHeapTier = D3D12_RESOURCE_HEAP_TIER_1, D3D12_HEAP_FLAG_SHARED_CROSS_ADAPTER = 0, D3D12_HEAP_FLAG_DENY_NON_RT_DS_TEXTURES = 0, D3D12_HEAP_FLAG_DENY_RT_DS_TEXTURES = 0, and D3D12_HEAP_FLAG_DENY_BUFFERS = 0. [ STATE_CREATION ERROR #631: CREATEHEAP_INVALIDMISCFLAGS]
`

kvark:
Interesting, thanks for the info!
[...]

Issue #1354, opened 2017-07-05
[meta][ll] Vulkan portability issues


kvark:
Here is a rough list of areas where Vulkan translation to other next-gen APIs is not straightforward.
Related to #1226, #1241, and #1102
cc @msiglreith @JohnColanduoni 

## Direct3D 12
  - #1352: `CopyTextureRegion` from a source buffer requires minimum alignment of 0x100 for the rows and 0x200 for the depth slices, while Vulkan supports tight packing.

## Metal

JohnColanduoni:
The SwapChain implementation I wrote uses an API that is considered private by Apple on iOS prior to the upcoming iOS 11. If we want to support Metal 1 on iOS that's another sticking point.

msiglreith:
Events will probably be hard to implement for both APIs, requiring to split command buffers.

kvark:
I think we should migrate all the points from  #1241 here, describe them in great detail, and close the other issue as a duplicate.

Kangz:
Not sure if this is the right place to share some of our findings so let me know if I should move this.

 - ```vkQueueWaitIdle``` that can be translated to D3D12 by adding a Signal to a queue, linking it to an Event object and waiting on the event. In Metal you can have a callback when a command buffer finishes executing, and have that trigger a condition variable. glFinish for OpenGL.
 - The placed footprint in D3D12 force an alignment of 512 for the offset from the start of the buffer. We have found that it is possible to work around this by sometimes doing two copies instead of one (for 2D textures).

kvark:
@Kangz Thanks for chiming in! I split your comment into corresponding issues (#1383 and #1352). I'd prefer to leave this space for more general concerns.

msiglreith:
# Issues (Portability API)
Aspects where vulkan portability is not straightforward. In particular, hard to map towards D3D12, which exposes a smaller and more restricted API surface:

## Memory mapping

**Summary:** Vulkan allows to map arbitrary slices of `Heap`/`DeviceMemory`. in other APIs mapping is coupled with (sub)resources.

### Vulkan
* Allocate `Heap`/`DeviceMemory` via `AllocateMemory`, which can be bound to `Buffer` or `Image`.
* Map/Unmap `DeviceMemory` via `(Un)MapBuffer` to read/write data to buffers or images.
* `DeviceMemory` can be only mapped once until it's unmapped again.

### D3D12
* Allocate memory in heaps and create placed resources inside the heaps, similar to vulkan
[...]

PR #1381, opened 2017-07-18
Added new create_heap functions to corell factory

ConorLPBoyle:
See #1379

### DIFF
--- a/src/backend/dx12ll/src/factory.rs
+++ b/src/backend/dx12ll/src/factory.rs
@@ -200,6 +200,35 @@ impl Factory {
             }
         }
     }
+fn create_flagged_heap(&mut self, heap_type: &core::HeapType, size: u64, flags: winapi::D3D12_HEAP_FLAGS) -> native::Heap {
+        let mut heap = ptr::null_mut();
+        let desc = winapi::D3D12_HEAP_DESC {
+            SizeInBytes: size,
+            Properties: data::map_heap_properties(heap_type.properties),
+            Alignment: 0,
+            Flags: flags,
+        };
+
+        assert_eq!(winapi::S_OK, unsafe {
+            self.inner.CreateHeap(&desc, &dxguid::IID_ID3D12Heap, &mut heap)
+        });
+
+        native::Heap {
+            inner: ComPtr::new(heap as *mut _),
+            ty: heap_type.clone(),
+            size: size,
+            //TODO: merge with `map_heap_properties`
+            default_state: if !heap_type.properties.contains(memory::CPU_VISIBLE) {
+                winapi::D3D12_RESOURCE_STATE_COMMON
+            } else if heap_type.properties.contains(memory::COHERENT) {
+                winapi::D3D12_RESOURCE_STATE_GENERIC_READ
+            } else {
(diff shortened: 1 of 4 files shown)


kvark:
Thank you for doing the investigation and filing this PR!

It's a bit unfortunate that we have to provide the queue type, since it is another derail from Vulkan API (#1354). I'll try to investigate a bit more, but so far there doesn't seem to be an easy workaround.

For the actual API change, one of the following would be preferred:
  1. Either `create_xxx_heap`  returns `Heap<SomePhantomType>` for strong typing, or
  2. it's still `create_heap` but there is a parameter of `enum` type to say what resources are supposed to be created on it

Another side-task is to double-check with Metal to see if the type may be required there as well.
cc @msiglreith @JohnColanduoni

msiglreith:
Eh, looks like we need to stray away further from the light..
I would prefer going with the 2nd option and expose the limitation in `Capabilities`.
The currently implemented methods from the 1st option could still be provided for easier usage, especially for the `render` layer.

src/backend/dx12ll/src/factory.rs
```diff
@@ -231,6 +260,21 @@ impl core::Factory<R> for Factory {
         }
     }
 
+    fn create_buffer_heap(&mut self, heap_type: &core::HeapType, size: u64) -> native::Heap {
+        self.create_flagged_heap(heap_type, size, 
+                                 winapi::D3D12_HEAP_FLAGS(0) | winapi::D3D12_HEAP_FLAG_DENY_NON_RT_DS_TEXTURES | winapi::D3D12_HEAP_FLAG_DENY_RT_DS_TEXTURES)
```
msiglreith: There are aliases which should be used here instead of manually assembling the flags:
* `D3D12_HEAP_FLAG_ALLOW_ONLY_BUFFERS`
* `D3D12_HEAP_FLAG_ALLOW_ONLY_NON_RT_DS_TEXTURES`
* `D3D12_HEAP_FLAG_ALLOW_ONLY_RT_DS_TEXTURES`

kvark:
@msiglreith 
> I would prefer going with the 2nd option and expose the limitation in Capabilities.
The currently implemented methods from the 1st option could still be provided for easier usage, especially for the render layer.

We happen to already use some of the strong typing in the coreLL, e.g. for command queues. I expect users to know ahead of time which heaps are going to be used for allocating of what types of resources, so having strong typing there wouldn't hurt, would it?

msiglreith:
> We happen to already use some of the strong typing in the coreLL, e.g. for command queues. I expect users to know ahead of time which heaps are going to be used for allocating of what types of resources, so having strong typing there wouldn't hurt, would it?

Fair enough, my motivation was to give users additional control for memory management if possibe but it probably won't make any difference in reality.

kvark:
We need to come up with a set of guidelines for typing, e.g. strong types are provided if:
  - user is expected to know them ahead of time
  - no overhead is implied

JohnColanduoni:
Metal heaps are much more hands off than either DX12 or Vulkan (they don't even allow you to specify allocation offsets), and only accept size and memory properties at creation. AFAIK the only restriction for heap resource allocation is that coherent memory cannot be allocated from a heap.

PR #1385, opened 2017-07-19
[ll] Heap types and device capabilities

### LINKED ISSUE #1379: Allocating heap memory for low level DX12 backend fails

kvark:
Fixes #1379
Closes #1381


src/corell/src/lib.rs
```diff
@@ -123,6 +135,8 @@ pub struct Device<R: Resources, F: Factory<R>, Q: CommandQueue> {
     pub transfer_queues: Vec<TransferQueue<Q>>,
     pub heap_types: Vec<HeapType>,
     pub memory_heaps: Vec<u64>,
+    /// Device capabilities
+    pub caps: Capabilities,
```
msiglreith: Could we already have access to these capabilities from `Adapter`?
So users could have more control about which device to open beside the current `AdapterInfo`.

kvark: That was my original design, but unfortunately D3D12 only queries those via `D3D12Device`, which we only get in `open()`, so I had to move stuff under `Device`. Anyhow, I'll be happy to revise this later.

PR #1387, opened 2017-07-20
Some doc improvements

nical:
When I have a bit of time I'll go through things that I find lacking or confusing in the doc and try to improve a bit with what I know. I'll add more doc to this PR as I go but don't hesitate to give feedback and corrections along the way.

First thing to note is that rustdoc doesn't do justice to some of the attempts at documenting the code. For example it's hard to see from the docs what the different bit patterns for a particular bitflags are. Landing on the page for `Bind` for instance isn't very informative: https://docs.rs/gfx/0.16.1/gfx/struct.Bind.html
It could probably be improved by the bit_flags macro but in the mean time, listing the different bit patterns in the doc of the bitflag itself helps a great deal I think (from the point of view of someone who tries to make something out of docs.rs.

### DIFF
--- a/src/core/src/memory.rs
+++ b/src/core/src/memory.rs
@@ -12,11 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Memory stuff
+//! Types to describe the properties of memory allocated for gfx resources.
 
 use std::mem;
 
-/// How this memory will be used.
+/// How this memory will be used on the GPU and/or the CPU.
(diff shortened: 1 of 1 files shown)


kvark:
Ugh
> error: internal compiler error: /checkout/src/librustc/infer/mod.rs:573: Encountered errors `[FulfillmentError(Obligation(predicate=Binder(TraitPredicate(<CB as audio::AudioCallback>)),depth=0),Unimplemented)]` resolving bounds after type-checking

PR #1388, opened 2017-07-22
More documentation

nical:
This is a work in progress. This PR adds the gfx::doc::terminology module which only contains rustdoc documentation. The idea is to have a centralized place with information about what various gfx concepts map to in other APIs. I am not quite certain what the best way to present this information is. For example we could also have sections "OpenGL to gfx", "D3D11 to gfx", etc which would only link to the documentation of each term in gfx concisely (like: " - GL thing: [gfx thing](link to thing in the doc)".

What I am trying to address here is that for someone who does not know the internals of gfx, using this crate currently is a small exercise in reverse-engineering, especially when you know what you want to do (but not how it is called in gfx).

The other thing that I want to move towards, is for each module (or at least each crate) to have a large doc block with sections `Overview` and `Examples` that give some info about the general design and what the most important concepts are with link to teh relevant pieces. Rustoc is not good at making the important things stand out, so it has to be done somewhat manually in the crate/module doc.

Ultimately, gfx should strive to get to a documentation at the level of [glium's](https://docs.rs/glium) which I took a lot of inspiration from when documenting [lyon](https://docs.rs/lyon). I received some positive feedback from the work on lyon and I think that gfx could really use some doc that is structured similarly.

I don't think should land as is with all of the TODO's. Don't hesitate to review this while the work is in progress to help me fill the holes and rewrite/rearrange as needed.

### DIFF
--- a/src/core/src/factory.rs
+++ b/src/core/src/factory.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Resource factory
+//! # Resource factory
 //!
 //! This module exposes the `Factory` trait, used for creating and managing graphics resources, and
 //! includes several items to facilitate this.
@@ -187,17 +187,52 @@ pub enum WaitFor {
     All,
 }
 
+/// # Overview
+///
 /// A `Factory` is responsible for creating and managing resources for the context it was created
 /// with.
 ///
-/// # Construction and Handling
+/// ## Construction and Handling
+///
 /// A `Factory` is typically created along with other objects using a helper function of the
 /// appropriate gfx_window module (e.g. gfx_window_glutin::init()).
 ///
 /// This factory structure can then be used to create and manage different resources, like buffers,
 /// shader programs and textures. See the individual methods for more information.
 ///
-/// Also see the `FactoryExt` trait inside the `gfx` module for additional methods.
+/// This trait is extended by the [`gfx::FactoryExt` trait](https://docs.rs/gfx/*/gfx/traits/trait.FactoryExt.html).
+/// All types implementing `Factory` also implement `FactoryExt`.
+///
+/// ## Immutable resources
+///
+/// Immutable buffers and textures can only be read by the GPU. They cannot be written by the GPU and
+/// cannot be accessed at all by the CPU.
+///
+/// See:
+///  - [`Factory::create_texture_immutable`](trait.Factory.html#tymethod.create_texture_immutable),
+///  - [`Factory::create_buffer_immutable`](trait.Factory.html#tymethod.create_buffer_immutable).
+///
+/// ## Raw resources
+///
+/// The term "raw" is used in the context of types of functions that have a strongly typed and an
+/// untyped equivalent, to refer to the untyped equivalent.
+///
+/// For example ['Encloder::create_buffer_raw'](trait.Factory.html#tymethod.create_buffer_raw) and
+/// ['Encloder::create_buffer'](trait.Factory.html#tymethod.create_buffer)
+///
+/// ## Shader resource views and unordered access views
+///
+/// This terminology is borrowed from D3D.
+///
+/// Shader resource views typically wrap textures and buffers to provide read-only access in shaders.
+/// An unordered access view provides similar functionality, but enables reading and writing to
+/// the buffer or texture in any order.
+///
+/// See:
+///
+/// - [The gfx::UNORDERED_ACCESS bit in the gfx::Bind flags](../gfx/struct.Bind.html).
+/// - [Factory::view_buffer_as_unordered_access](trait.Factory.html#method.view_buffer_as_unordered_access).
+///
 #[allow(missing_docs)]
 pub trait Factory<R: Resources> {
     /// Returns the capabilities of this `Factory`. This usually depends on the graphics API being
--- a/src/core/src/factory.rs
+++ b/src/core/src/factory.rs
@@ -217,8 +217,8 @@ pub enum WaitFor {
 /// The term "raw" is used in the context of types of functions that have a strongly typed and an
 /// untyped equivalent, to refer to the untyped equivalent.
 ///
-/// For example ['Encloder::create_buffer_raw'](trait.Factory.html#tymethod.create_buffer_raw) and
-/// ['Encloder::create_buffer'](trait.Factory.html#tymethod.create_buffer)
+/// For example ['Factory::create_buffer_raw'](trait.Factory.html#tymethod.create_buffer_raw) and
+/// ['Factory::create_buffer'](trait.Factory.html#tymethod.create_buffer)
 ///
 /// ## Shader resource views and unordered access views
 ///
--- a/src/render/src/lib.rs
+++ b/src/render/src/lib.rs
@@ -14,8 +14,79 @@
 
 #![deny(missing_docs)]
 
-//! An efficient, low-level, bindless graphics API for Rust. See [the
-//! blog](http://gfx-rs.github.io/) for explanations and annotated examples.
+//! # gfx
+//!
+//! An efficient, low-level, bindless graphics API for Rust.
+//!
+//! # Overview
+//!
+//! ## Command buffers and encoders
+//!
+//! A command buffer is a serialized list of drawing and compute commands.
+//! Unlike with vulkan, command buffers are not what you use to create commands, but only
+//! the result of creating these commands. Gfx, borrowing metal's terminology, uses
+//! encoders to build command buffers. This means that, in general, users of the gfx crate
+//! don't manipulate command buffers directly much and interact mostly with encoders.
+//!
+//! Manipulating an `Encoder` in gfx corresponds to interacting with:
+//!
+//! - a `VkCommandBuffer` in vulkan,
+//! - a `MTLCommandEncoder` in metal,
+//! - an `ID3D12GraphicsCommandList` in D3D12.
+//!
+//! OpenGL and earlier versions of D3D don't have an explicit notion of command buffers
+//! or encoders (with the exception of draw indirect commands in late versions of OpenGL,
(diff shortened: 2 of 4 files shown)


nical:
It might be better to make the terminology page just a list of links to various areas of the documentation and move the extra explanation closer to their types in the doc.

src/core/src/factory.rs
```diff
@@ -12,10 +12,41 @@
...
+//!
+//! The term "raw" is used in the context of types of functions that have a strongly typed and an
+//! untyped equivalent, to refer to the untyped equivalent.
+//!
+//! For example ['Encloder::create_buffer_raw'](../../trait.Factory.html#tymethod.create_buffer_raw) and
+//! ['Encloder::create_buffer'](../../trait.Factory.html#tymethod.create_buffer)
+//!
+//! ## Unordered access
+//!
+//! Unordered access views refer to buffers that can be written to by a shader.
```
msiglreith: We should annotate here that they support read and write to avoid confusion.

src/render/src/doc/terminology.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2016 The Gfx-rs Developers.
```
msiglreith: nit: year

src/render/src/doc/terminology.rs
```diff
@@ -0,0 +1,49 @@
...
+//!
+//! In OpenGL, handles are simple integers (contrary to gfx's high level API, OpenGL does not
+//! provide any automatic memory management).
+//!
+//! Note: The comments about memory management above apply to the high level API, but not the
+//! low level one (ll).
+//!
+//! ## Slice
+//!
+//! A `Slice` in gfx describes for a givien draw call the range of vertices to process, their order
```
msiglreith: typo: `givien`/`given`

src/render/src/lib.rs
```diff
@@ -14,8 +14,63 @@
...
-//! blog](http://gfx-rs.github.io/) for explanations and annotated examples.
+//! # gfx
+//!
+//! An efficient, low-level, bindless graphics API for Rust.
+//!
+//! # Overview
+//!
+//! ## Command buffers and encoders
+//!
+//! A command buffer is an already built serialized list drawing and compute commands.
```
msiglreith: This sentence is a bit confusing to me, is there some word missing?

src/render/src/lib.rs
```diff
@@ -14,8 +14,63 @@
...
+//! Unlike with vulkan, command buffers are what you use to create commands, but only
```
jonas-schievink: Is there a "not" missing in this sentence?

src/render/src/lib.rs
```diff
@@ -14,8 +14,63 @@
...
+//! - The [`Command buffer` trait documentation](trait.CommandBuffer.html).
+//!
+//! ## Factory
+//!
+//! The factory is what lets you allocate GPU resources such as buffers and textures.
+//!
+//! Each gfx backend provides its own factory type which implements both:
+//!
+//! - The [`factory` trait documentation](traits/trait.Factory.html#overview).
+//! - The [`factoryExt` trait documentation](traits/trait.FactoryExt.html).
```
jonas-schievink: "documentation" seems out of place here. Also, traits should be capitalized correctly.

src/core/src/factory.rs
kvark: UAVs don't refer to just buffers, also can view into textures

src/render/src/doc/terminology.rs
```diff
@@ -0,0 +1,49 @@
...
+
+//! # Terminology
+//!
+//! ## Buffer
+//!
+//! TODO!
+//!
+//! ## Handle
+//!
+//! Handles are the mechanism primarily gfx uses to refer to most GPU resources.
```
kvark: "gfx primarily" - a swap of these sounds better to me

src/render/src/doc/terminology.rs
```diff
@@ -0,0 +1,49 @@
...
+//! Most handle types use an atomically reference counted pointer internally to automatically
+//! deal with memory management. This means that when using the gfx's high level API, there is
+//! no need to explicitly manage the lifetime and destroy handles.
+//!
+//! In OpenGL, handles are simple integers (contrary to gfx's high level API, OpenGL does not
+//! provide any automatic memory management).
+//!
+//! Note: The comments about memory management above apply to the high level API, but not the
```
kvark: I see "high level" here being a confusing/over-loaded term

src/render/src/doc/terminology.rs
```diff
@@ -0,0 +1,49 @@
...
+//! low level one (ll).
+//!
+//! ## Slice
+//!
+//! A `Slice` in gfx describes for a givien draw call the range of vertices to process, their order
+//! and the number of instances.
+//! The slice is how index buffers are specfied in a draw call.
+//!
+//! Not to confuse with the term slice as employed more broadly in rust (`&[T]`) which is a view into
```
kvark: `Slice` is just not a very good name for our concept...

src/render/src/lib.rs
```diff
@@ -14,8 +14,63 @@
...
+//!
+//! OpenGL does not have a notion of factory (resources are created directly off of the global
+//! context). D3D11 has a DXGI factroy but it is only used to interface with other processes
```
kvark: "factroy"

src/render/src/lib.rs
```diff
@@ -14,8 +14,63 @@
...
+//! and the window manager, resources like textures are usually created using the device.
```
kvark: In fact `gfx::Factory` is roughly equivalent to `ID3D11Device`, `MTLDevice` and `VkDevice`.
@msiglreith - we should consider renaming it to `Device`. Even if confusing, at least it goes in line with the underlying backends

msiglreith: Sounds reasonable. Factory interacts now with fences therefore `Device` would be more suitable. We need to find a new name for our `Device` struct then!

src/render/src/doc/terminology.rs
nical: I agree. It confused me until I looked into its code, and the notion of slice is so central in rust that I think it would be good to avoid overloading it.

src/render/src/doc/terminology.rs
nical: Agreed. How would you name the high level api as opposed to the ll one?

nical:
I see that Factory has ` *_raw` methods for untyped buffer creation and `*_u8` for textures. Do they mean different things?

kvark:
@nical 
> Agreed. How would you name the high level api as opposed to the ll one?

Maybe "current core" as opposed to "low level core"? I don't see a good name :(

> I see that Factory has *_raw methods for untyped buffer creation and *_u8 for textures. Do they mean different things?

`_u8` has been added on top, it's not soundly matching the existing APIs. It still operates on typed handles, just expects byte slices of data.

@msiglreith 
> We need to find a new name for our Device struct then!

If you are referring to CoreLL `Device`, then it looks rather auxiliary, so the name doesn't matter much. Could be `ActiveAdapter` or something.

src/core/src/factory.rs
```diff
@@ -187,17 +187,52 @@ pub enum WaitFor {
...
+/// See:
+///  - [`Factory::create_texture_immutable`](trait.Factory.html#tymethod.create_texture_immutable),
+///  - [`Factory::create_buffer_immutable`](trait.Factory.html#tymethod.create_buffer_immutable).
+///
+/// ## Raw resources
+///
+/// The term "raw" is used in the context of types of functions that have a strongly typed and an
+/// untyped equivalent, to refer to the untyped equivalent.
+///
+/// For example ['Encloder::create_buffer_raw'](trait.Factory.html#tymethod.create_buffer_raw) and
```
msiglreith: `Encloder`?

src/render/src/lib.rs
```diff
@@ -14,8 +14,79 @@
...
+//! - a `MTLCommandEncoder` in metal,
+//! - an `ID3D12GraphicsCommandList` in D3D12.
+//!
+//! OpenGL and earlier versions of D3D don't have an explicit notion of command buffers
+//! or encoders (with the exception of draw indirect commands in late versions of OpenGL,
+//! which can be seen as a GPU-side command buffer). They are managed implicitly by the driver.
+//!
+//! See:
+//!
+//! - The [`Encoder` struct documentation](struct.Encoder.html).
```
msiglreith: We only expose `GraphicsEncoder` atm (after the ll integration got merged, the old `Encoder` got renamed)

src/render/src/lib.rs
```diff
@@ -14,8 +14,79 @@
...
+//! - `ID3D11Device` in D3D11,
+//! - `MTLDevice` in metal.
+//!
+//! OpenGL does not have a notion of factory (resources are created directly off of the global
+//! context). D3D11 has a DXGI factory but it is only used to interface with other processes
+//! and the window manager, resources like textures are usually created using the device.
+//!
+//! ## Device
+//!
+//! See [the `gfx::Device` trait](trait.Device.html).
```
msiglreith: `Device` is now a struct in `core`

src/render/src/lib.rs
```diff
@@ -14,8 +14,79 @@
...
+//!
+//! ## Pipeline state (PSO)
+//!
+//! See [the documentation of the gfx::pso module](pso/index.html).
+//!
+//! ## Memory management
+//!
+//! Handles internally use atomically reference counted pointers to deal with memory management.
+//! GPU resources are not destroyed right away when all references to them are gone. Instead they
+//! are destroyed the next time [Device::cleanup](trait.Device.html#tymethod.cleanup) is called.
```
msiglreith: `cleanup` is now a method of command queues

nical:
> A few things changed with the first ll integration (sorry for the extra work!). Would you mind addressing the points I mentioned?

No problem, on it.

PR #1398, opened 2017-07-26
Check for CopyBufferSubData, fixes #1356 and adds an alternative path

### LINKED ISSUE #1356: [GL] Check for CopyBufferSubData support
We currently use the `glCopyBufferSubData` in the command buffer, but don't check for support (`copy_buffer_supported`) for this.

keringar:
I tested it in a VM that only supported OpenGL 3.0 and it seems to work fine. However, I don't know if the assert is necessary, should it be removed?

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -656,14 +656,34 @@ impl CommandQueue {
             },
             Command::CopyBuffer(src, dst, src_offset, dst_offset, size) => {
                 let gl = &self.share.context;
-                unsafe {
-                    gl.BindBuffer(gl::COPY_READ_BUFFER, src);
-                    gl.BindBuffer(gl::COPY_WRITE_BUFFER, dst);
-                    gl.CopyBufferSubData(gl::COPY_READ_BUFFER,
-                                         gl::COPY_WRITE_BUFFER,
-                                         src_offset,
-                                         dst_offset,
-                                         size);
+
+                if self.share.capabilities.copy_buffer_supported {
+                    unsafe {
+                        gl.BindBuffer(gl::COPY_READ_BUFFER, src);
+                        gl.BindBuffer(gl::COPY_WRITE_BUFFER, dst);
+                        gl.CopyBufferSubData(gl::COPY_READ_BUFFER,
+                                            gl::COPY_WRITE_BUFFER,
+                                            src_offset,
+                                            dst_offset,
+                                            size);
+                    }
+                } else {
+                    debug_assert!(self.share.private_caps.buffer_storage_supported == false);
+
+                    unsafe {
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/lib.rs
```diff
@@ -656,14 +656,31 @@ impl CommandQueue {
...
+                                            src_offset,
+                                            dst_offset,
+                                            size);
+                    }
+                } else {
+                    assert!(size > 0);
+
+                    unsafe {
+                        let mut src_ptr = 0 as *mut ::std::os::raw::c_void;
+                        factory::temporary_ensure_mapped(&mut src_ptr, gl::COPY_READ_BUFFER, src, memory::READ, gl);
```
Bastacyclop: You are bypassing GFX's mapping tracking mecanism so the [automatic unmapping](https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/lib.rs#L877) will not happen and the buffer might already be mapped. IIRC this means that the next time you will try to use this buffer for something like mapping it again or drawing with it, you will get an `INVALID_OPERATION`.

keringar: Are there any alternatives to glCopyBufferSubData than? Unless I'm misunderstanding, with a version of OpenGL that old, persistent mappings would not be supported and I wouldn't be able to upload data because the buffer is already mapped.

keringar:
I'm going to work on this some more I had an idea on an alternative.

Bastacyclop:
I don't know much about GL < 3.2, but without taking performances into account simply unmapping both buffers after the copy should be enough when persistent mapping is not supported (is it always true when `glCopyBufferSubData` is not supported ?). When using temporary mappings, you know that both buffers are unmapped when executing this command because `glCopyBufferSubData` needs this invariant too and the current tracking mecanism ensures it. So if you only map them locally (i.e. you unmap them after use) you are preserving the invariant for the following operations.

keringar:
Oh I understand what you mean now. Can you check this once you get the chance?

src/backend/gl/src/lib.rs
```diff
@@ -656,14 +656,32 @@ impl CommandQueue {
...
+                    unsafe {
+                        gl.BindBuffer(gl::COPY_READ_BUFFER, src);
+                        gl.BindBuffer(gl::COPY_WRITE_BUFFER, dst);
+                        gl.CopyBufferSubData(gl::COPY_READ_BUFFER,
+                                            gl::COPY_WRITE_BUFFER,
+                                            src_offset,
+                                            dst_offset,
+                                            size);
+                    }
+                } else {
```
Bastacyclop: Maybe add `debug_assert!(self.share.private_caps.buffer_storage_supported == false)` to make sure there is no persistent mapping going on.

keringar: Added

PR #1401, opened 2017-07-26
Exclude the nightly compiler from bors builds

notriddle:
The [internal compiler errors][ICE] are annoying.

[ICE]: https://travis-ci.org/gfx-rs/gfx/jobs/257898756#L500


kvark:
Looks awesome, thank you!

PR #1403, opened 2017-07-27
dx11/dx12: Replace comptr with wio

### LINKED ISSUE #1402: Use wio-rs for ComPtr
We currently use [`comptr-rs`](https://github.com/msiglreith/comptr-rs), which is unpublished (and I don't really consider to publish it on crates.io). Therefore I would propose switching over to [`wio-rs`](https://github.com/retep998/wio-rs) as their implementation is probably also more mature.

Blocked on https://github.com/retep998/wio-rs/pull/9

msiglreith:
Temporary refering to my `wio` fork as we are blocked on winapi 0.3 (or a new 0.2 release) anyway and this might require an `wio` version bump again.

Closes #1402

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -22,7 +22,7 @@ extern crate d3dcompiler;
 extern crate dxgi;
 extern crate dxguid;
 extern crate winapi;
-extern crate comptr;
+extern crate wio;
 
 pub use self::data::map_format;
 pub use self::factory::Factory;
@@ -37,7 +37,7 @@ mod pool;
 mod state;
 
 use core::{command as com, handle};
-use comptr::ComPtr;
+use wio::com::ComPtr;
 use std::cell::RefCell;
 use std::ptr;
 use std::sync::Arc;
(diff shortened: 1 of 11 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -59,19 +59,18 @@ impl Instance {
     #[doc(hidden)]
     pub fn create() -> Self {
         // Create DXGI factory
-        let mut dxgi_factory = ComPtr::<winapi::IDXGIFactory2>::new(ptr::null_mut());
-
+        let mut dxgi_factory: *mut winapi::IDXGIFactory2 = ptr::null_mut();
         let hr = unsafe {
             dxgi::CreateDXGIFactory1(
                 &dxguid::IID_IDXGIFactory2,
-                dxgi_factory.as_mut() as *mut *mut _ as *mut *mut c_void)
```
kvark: could you add `as_mut` method in the fork/upsteam? It seems quite useful on multiple occasions in our code

msiglreith: Sry, I can't follow as there is already a `as_mut` in wio's `ComPtr` and I don't see how/where a new method would help.
The only thing I missed a bit was a function returning `&mut *mut T` which resulted in first creating a pointer and then store it later into the `ComPtr` wrapper.

torkleyy:
@notriddle Do you know why there is no status report?

notriddle:
According to the log, bors failed to push the commit to your master branch (`{:error, :push}`). The log is right here, though crash reports are only visible to reviewers and admins: https://app.bors.tech/repositories/848/log

You need to turn off the "Review required" thing. It's in Settings -> Branches -> Protected Branches -> master

kvark:
Thanks @notriddle !
bors retry

PR #1406, opened 2017-07-30
core: Rework queue submission

msiglreith:
Current queue submission allowed to submit all kind of buffers to a queue which could cause issues on some backends. We provide now untyped and a strong-typed interface for command buffer submission employing a builder pattern internally.

Queue capabilities and hierarchies are fully encoded into the type systems. Submissions will be upcasted to the match the required capabilities needed for executing the passed command buffers. Only queue with matching capabilities can execute this submissions.

A missing point which will be addressed in a followup would to allow submission of different Submission objects, currently required to have the same capability, possibly with an additional builder for a collection of submissions.

Example for submission building:
```rust
let submission_a =
    Submission::new()
        .wait_on(&[&landscape_semaphore, &sky_semaphore]) // wait_semaphores
        .wait_on(&[&ocean_height_sync])                   // wait_semaphores
        .submit(&[buffer_tree1, buffer_tree2])            // push graphics command buffers
        .submit(&[buffer_ocean])                          // push more graphics command buffers!
        .submit(&[ocean_propagation])                      // push some compute stuff
        .signal(&[&terrain_semaphore]);                   // signal semaphores
```
The resulting submission would require `General` capability as it contains compute and graphics commands.

### DIFF
--- a/src/backend/dx11/src/command.rs
+++ b/src/backend/dx11/src/command.rs
@@ -26,6 +26,7 @@ use core::{MAX_VERTEX_ATTRIBUTES, MAX_CONSTANT_BUFFERS,
            MAX_SAMPLERS, MAX_COLOR_TARGETS};
 use {native, Backend, CommandList, Resources, InputLayout, Buffer, Texture, Pipeline, Program};
 
+#[derive(Clone)]
 pub struct SubmitInfo<P> {
     pub parser: P,
 }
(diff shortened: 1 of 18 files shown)


msiglreith:
Metal backend doesn't pass CI atm. I'm unsure how to properly fix it (without a machine to test) as the currently implementation for checking for the last command buffer during submission looks a bit scary.

kvark:
Awesome, thank you!
I'm not going to have access to a Mac for a week, could @JohnColanduoni take a look?

msiglreith:
CI should pass now (implemented an custom iterator extension..), but unsure if it will run (metal backend could be broken anyway atm due to the integration merge),

PR #1413, opened 2017-08-03
examples: Fix warnings

burtonageo:


examples/performance/main.rs
```diff
@@ -94,7 +94,6 @@ trait Renderer: Drop {
 struct GFX {
     dimension: i16,
     window: gfx_window_glutin::Window,
-    surface: gfx_window_glutin::Surface,
```
msiglreith: I'm not totally sure if dropping it early is safe, haven't looked into details about the required lifetime of `Surface`s but it's on my TODO. But for GL it shouldn't be a problem iirc.

burtonageo: I can take it out, and just replace it with `_surface` instead, if you prefer

kvark:
bors r=msiglreith

PR #1414, opened 2017-08-03
Rename factory to device and device to gpu

### LINKED ISSUE #698: Find better names for the core concepts

msiglreith:
Closes #698

### DIFF
--- a/examples/blend/main.rs
+++ b/examples/blend/main.rs
@@ -52,15 +52,15 @@ impl Vertex {
     }
 }
 
-fn load_texture<R, F>(factory: &mut F, data: &[u8])
+fn load_texture<R, D>(device: &mut D, data: &[u8])
                 -> Result<gfx::handle::ShaderResourceView<R, [f32; 4]>, String> where
-                R: gfx::Resources, F: gfx::Factory<R> {
+                R: gfx::Resources, D: gfx::Device<R> {
     use std::io::Cursor;
     use gfx::texture as t;
     let img = image::load(Cursor::new(data), image::PNG).unwrap().to_rgba();
     let (width, height) = img.dimensions();
     let kind = t::Kind::D2(width as t::Size, height as t::Size, t::AaMode::Single);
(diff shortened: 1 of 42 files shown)


brendanzab:
Awesome stuff! Was never really happy with the `Factory` name! 👏

bvssvni:
Nice! Like the name `Gpu`.

PR #1250, opened 2017-05-02
[WIP] Use cargo workspace. Move gfx_app in its own crate

ama0:
This is a first prototype on using cargo workspace for the whole project. (#1248)
gfx_app is no longer the 'main crate'. It is moved into src/gfx_app

I left the old dependencies in the main `Cargo.toml` to make it easier to compare and see what may be a cause of error.

What is working:
- cargo build
- cargo run --example [all examples work]
- cargo test

What is not working:
- cargo build --all
- cargo test --all
Both throw a bunch of errors because the platform is wrong. See [Provide a way to run cargo subcommands on all crates in the workspace #2878](https://github.com/rust-lang/cargo/issues/2878#issuecomment-292986795) This seems to be the blocker right now.

Issue #1248, opened 2017-05-01
Use cargo workspaces


kvark:
https://doc.rust-lang.org/beta/book/second-edition/ch14-03-cargo-workspaces.html

Benefits:
  - more idiomatic
  - re-uses the same "target" folder - faster publishing, cleaner
  - magnificent `cargo test --all` to test everything at once

TODO:
  - add `[workspace]` to the root cargo toml
  - adjust `.gitignore` to only care about the `/target` (as opposed to any `target`)
  - simplify Travis script to use `test --all`

torkleyy:
Plus a `workspace` key is needed in every crate's `Cargo.toml`.

I would also suggest converting `gfx_app` into a workspace member, which is clearer IMO.

kvark:
@torkleyy right. With an ability to build everything in a workspace, we no longer strictly need a project like `gfx_app` to depend on everything. It would be great to move it out of root, to avoid confusion.

ama0:
I'm currently looking into it. Don't hold your breath though. :)

> Plus a `workspace` key is needed in every crate's `Cargo.toml`.

If I understand it correctly, this isn't the case. See:  [1525-cargo-workspace.md#implicit-relations](https://github.com/rust-lang/rfcs/blob/master/text/1525-cargo-workspace.md#implicit-relations)
>The package.workspace can be omitted if it would only contain ../ (or some repetition of it). That is, if the root of a workspace is hierarchically the first Cargo.toml with [workspace] above a crate in the filesystem, then that crate can omit the package.workspace key.

Another thing that may become an issue for people using [cargo-graph](https://github.com/kbknapp/cargo-graph) is an incorrect `[root]` in `Cargo.lock`. See [Incorrect [root] in Cargo.lock when using workspaces #3203](https://github.com/rust-lang/cargo/issues/3203)
But this shouldn't affect gfx-rs in any way.

torkleyy:
@ama0 Oops, sorry for the useless note then. Thought it would be necessary, thanks for pointing this out.

ama0:
> TODO:
>
> *  adjust .gitignore to only care about the /target (as opposed to any target)

This conflicts with the current goal to have simple window examples in all window providers. `window_sdl` currently has one and running this example from `/src/window/sdl` creates a target folder _in_ `/src/window/sdl`. Therefore I think it would be better to leave the current "ignore all target folders" entry in `.gitignore`.

xtian:
[...]

PR #1417, opened 2017-08-05
Move examples to explicit gfx_support crate/module. Also move to virtual cargo workspace.

bjadamson:
I noticed [this issue about gfx_app](https://github.com/gfx-rs/gfx/issues/922) has pretty much stopped, when I was just looking, and realized the exact same thing had been confusing me trying to learn gfx. In the thread, it was brought up maybe it was a good idea to move gfx_app to it's own module, to help users with learning the API (whom are reading the example code).

Since progress has slowed down in [the PR for cargo workspaces](https://github.com/gfx-rs/gfx/pull/1250) I thought I would submit this as *some* progress. I thought it was good timing since the last commit to the gfx repository just accepted a [pull-request doing a similar rename](https://github.com/gfx-rs/gfx/commit/d1e99ce6ac0b1c1bb06feb73518f66662001f859), it seems smart to do another rename pull-request like this (since others maintaining branches will already need to rebase). I could be wrong about this, but the alternative would for me to sit back and wait until another point in the future to do this.

I think renaming gfx_app to gfx_support, and more importantly moving the code out of gfx's root (and under examples/) will help users understand what is "gfx library code" vs "gfx support library code". The distinction is subtle but important, I believe. It certainty wasn't clear to me until quite a while of digging by myself.

I have a few things I'd like to discuss moving forward, this pull request certainly isn't ready as is.
1. I'd like to gather consensus that this is the right direction to move gfx. Are there any objections to the direction of this pull-request?
2. Is gfx_project a good name for the "root project"?
3. I'm not sure that duplicating the Cargo.toml file is the right way to express what I'm intending. Right now I'm duplicating the dependencies section in both Cargo.toml files.  Apart from going through each dependency one-by-one, and seeing whether examples/ or gfx/ requires the dependency, is there a way to not duplicate the dependencies your aware of? Is duplicating them a problem?
4. In the other thread glium was brought up as an example, in glium's example code, the helper code lives in the [examples/support/](https://github.com/glium/glium/tree/master/examples/support) directory, causing all examples to have to add  [explicitly mod support it](https://github.com/glium/glium/blob/master/examples/deferred.rs#L10). I tried emulating the same setup, but I wasn't able to get it to compile. I would need to spend more time comparing with the glium project to emulate that setup.
5. I need to try using the support code from another package, to make sure it's visible the way I think it should be. This is a TODO.

Anyways, I stole this list from the other thread. Here is what is working:

cargo build
cargo run --example (they all seem to work)

What is not working:

cargo test -- this isn't working for me on master/, not completely sure abouit this one.
cargo build --all (error: native frameworks are only available on macOS targets)
cargo test --all (error: Could not compile `core-foundation-sys`)

Pretty tired now when writing this up, so posting this looking to get some direction/guidance to move forward with this change. 

edit:
Resolves:
* #1248
* #1424


msiglreith:
Hi and thanks for that opening this PR!

> I was just looking, and realized the exact same thing had been confusing me trying to learn gfx

I had a similar experience when trying to understand the repository structure. Therefore I'm also in favor of reorganizing the crates.

> I could be wrong about this, but the alternative would for me to sit back and wait until another point in the future to do this.

Seems perfectly fine to do it now! Even though I don't have a strong opinion for `gfx_app` or `gfx_support`. If we move towards `support`, we could also think about abandoning the current `gfx_core` crate as the backend-specific setup code became quite [small](https://github.com/gfx-rs/gfx/blob/master/src/lib.rs#L236-L303) after the window changes.

> Is gfx_project a good name for the "root project"?

What would be the root project? IMO we should move towards a workspace based structure and clearly separate the crates. e.g
```
- examples
- backends
--- gl
--- dx11
--- vulkan
- gfx (current core)
- gfx_app/gfx_support (or merge this gfx_example)
- gfx_render
- window
--- dxgi
--- metal
--- glutin
```
Opinions?

torkleyy:
I like that directory structure. Examples could probably just be a crate with many binaries.

bjadamson:
That structure makes sense to me, but I'm seeing some blockers in the other thread that seemed to have [halted progress](https://github.com/gfx-rs/gfx/pull/1250). Does it make sense to use cargo workspaces if they can't pass the criteria? (the things not working in #1250.

I guess, does it matter if cargo build --all and cargo test --all won't work for now?

torkleyy:
I fixed that issue already IIRC.

bjadamson:
I also got platform errors when running cargo test --all and cargo build --all. Is that what you mean you fixed?

torkleyy:
I added an --exclude flag to cargo for that.

bjadamson:
Ok, I moved the directory structure around to match what @msiglreith proposed above.

One thing I need to request help with, I'm not sure how to best remove duplication between the root level Cargo.toml and gfx_support/Cargo.toml files. Both crate's need access to the cfg features, so it seems I need to declare the gl/dx/vulkan/etc... dependencies in both Cargo.toml files.. Does anyone have any ideas to only have to declare the list of "features" and dependencies once?

Here's exactly what I mean by duplication:
https://github.com/bjadamson/gfx/blob/mod-support/Cargo.toml#L12-L84
https://github.com/bjadamson/gfx/blob/mod-support/gfx_support/Cargo.toml#L30-L101

You'll notice in the two links, a subset of dependencies and config flags are duplication. The flags seem really bad for DRY practices, maybe the duplicated dependencies list is unavoidable though?

> What would be the root project? IMO we should move towards a workspace based structure and clearly separate the crates. e.g

That makes sense, but I think moving to a workspace structure is another step after this (if anything to reduce the scope of the amount I'm trying to learn all at once). That is unless you believe a workspace structure would help me with my duplication problem above?

My issue with using workspaces I'm having trouble finding any real documentation on cargo workspaces. Specifically, I could not find a good example repository to emulate for gfx. Because of this, I wasn't able to work through the errors I was getting trying to switch to using cargo workspaces.

msiglreith:
Regarding workspaces you could take a look [here](https://github.com/msiglreith/panopaea) for example. (shameless self-promotion..)
The root `Cargo.toml` only defines the repository hierarchy.

Summoning @kvark for opinions on the proposed structure and the overall direction

bjadamson:
Thank you for the link! I'll look through your repo (very nice logo btw) and see what I can accomplish.

bjadamson:
I'm unsure the best way to move forward now. My last commit fully switches the project to using a cargo workspace, based off your example repo. I've gotten to the point where cargo build --all fails, but I can run the examples (trying to build packages previously controlled by cfg section of top-level Cargo.toml file).

Since previously the crate inclusion was controlled by the [top level Cargo.toml file,](https://github.com/gfx-rs/gfx/blob/master/Cargo.toml#L12-L24) I'm assuming we should use the --exclude flag for cargo @torkleyy implemented. I'm unsure how to control which packages get built, I doubt the user experience we want is to manually type the --exclude paths ourselves. What did you have in mind here? 

I'm assuming we want to be able to say, "cargo build --all" and "cargo test -all", and have the --exclude's figured out by the compiler. Is this assumption correct?

kvark:
@bjadamson 

>I'd like to gather consensus that this is the right direction to move gfx. Are there any objections to the direction of this pull-request?

No objections here.

> In glium's example code, the helper code lives in the examples/support/ directory, causing all examples to have to add explicitly mod support it. 

We need to have it as a crate, not just a module, so other clients are able to use it.

> Is gfx_project a good name for the "root project"?
I'm not sure that duplicating the Cargo.toml file is the right way to express what I'm intending. Right now I'm duplicating the dependencies section in both Cargo.toml files. Apart from going through each dependency one-by-one, and seeing whether examples/ or gfx/ requires the dependency, is there a way to not duplicate the dependencies your aware of? Is duplicating them a problem?

I don't think we need the root project at all. Duplicating contents of `Cargo.toml` doesn't make sense. We already made steps to have separate examples in glfw/sdl windows. We just need to push it all the way down:
  - `trianglell` will be inside `gfx_core/examples`
  - `triangle` will be inside `gfx_render/examples`
  - glfw/sdl examples will be left untouched
  - all other examples will be in `gfx_support/examples`

This strictly corresponds to what level of abstraction an example is based on.
Doing this would mean the root `Cargo.toml` has nothing but the list of workspace members, similar to https://github.com/servo/webrender/blob/master/Cargo.toml

Running examples would be a little less convenient, but that's a minor thing to worry about. Building everything would be done with `cargo test --all --exclude XXX`, where XXX depends on the platform.

@msiglreith 
>we could also think about abandoning the current gfx_core crate

This crate is only going to be used directly by a handful of clients, no need to strive making the name shorter (would only cause confusion).

msiglreith:
@kvark Sorry, I meant `gfx_app` not `gfx_core` (edited it now)

kvark:
Dropping `gfx_app` completely would be great 👍 
If there is some glue code still to be shared between examples, I'd be fine with glium-style `support` module as originally mentioned. My objection to it was based on the assumption that `gfx_app` still has some logic in it.

bjadamson:
> everything would be done with cargo test --all --exclude XXX, where XXX depends on the platform

Generating XXX is what I've been trying to figure out the past few days. Previously before switching to virtual workspace, gfx uses cfg flags in the root Cargo.toml file, now that's not possible. Do you have an idea on how that would work?

> Running examples would be a little less convenient, but that's a minor thing to worry about. Building everything would be done with cargo test --all --exclude XXX, where XXX depends on the platform.

Maybe this could be solved with a CLI tool? One you can install with Cargo, like some other projects have.

kvark:
@bjadamson I agree it's rather annoying to not being able to just do `cargo test --all`. On the other hand, there is way more users of gfx-rs (who don't need to care about this) than developers. So we could start by just documenting the `XXX` in the README and then see if we can do more with a cargo plugin.

bjadamson:
Ok, I have moved the examples as you stated you would have them. The build is going to fail, right now I think that all is left is modifying the [appvoyer config file](https://github.com/gfx-rs/gfx/blob/master/appveyor.yml)

I need to think a bit on what it should be changed too, since test --all won't work at all, but we still want platform testing. I think.

bjadamson:
I got the build to succeed, using this [new appvoyer configuration](https://github.com/bjadamson/gfx/blob/mod-support/appveyor.yml#L17-L23). Could someone take a look and suggest any improvements?

I tried to make sure we're executing the same tests as before. It's just now I'm using gfx_support as the "build target", as it seems to consume all the other crates. I'm open to ideas that are better than setting your current working directory to the gfx_support directory and executing "cargo build"

bjadamson:
For some reason, the latest appvoyer execution didn't pick up my latest changes to the appvoyer config file. It was still executing using cargo test --all, which I changed. I'm pushing a README.md change to try and hopefully get appvoyer to pick up the newest config.

bjadamson:
Oops, I'm going to move the examples/ to gfx_support/examples

bjadamson:
I'm going to give the queue a chance to catch up now, hopefully fix any errors that come up.

From my end, once the tests pass, all that is left is to review:

https://github.com/bjadamson/gfx/blob/mod-support/gfx_render/examples/Cargo.toml

https://github.com/bjadamson/gfx/blob/mod-support/gfx_core/examples/Cargo.toml

https://github.com/bjadamson/gfx/blob/mod-support/gfx_support/examples/Cargo.toml

Right now, I'm unsure which configuration belongs in which. Right now the platform dependencies are duplicated across all 3 Cargo.toml files.

bjadamson:
Weird, the [build](https://travis-ci.org/gfx-rs/gfx/builds/262561385?utm_source=github_status&utm_medium=notification) failed on two of the hosts, for a CMake error. Passed on the other two hosts though, I wonder if it's a fluke. Going to push dummy commit to force rebuild.

bjadamson:
Ok, I believe everything to be correct. It seems a [known issue](https://www.traviscistatus.com/incidents/p3wv8zvkz3zb) is preventing two of the hosts from executing the tests correctly. Going to see if they work tomorrow.

In the meantime, this is ready for review. Unless I think of anything else, I don't have any more changes to make. Hopefully once this get merged I'll have some motivation to write up some doc's for gfx based on what I've learned using it.

msiglreith:
Nice progress so far 👍 
The build scripts can be simplified to something like
`cargo test --all --exclude gfx_window_sdl --exclude gfx_device_metal --exclude gfx_device_metalll --exclude gfx_window_metal` (example for windows), which should cover everything imo.

The current issue I see with our structure is that it's hard to find the examples. When I see a new project on github I usually check for a `examples` folder or `test` folder and take a quick look at the README to find a way to get started. Our usual examples are inside `gfx_support` which is probably unintuitive if you are not familiar with the project even though I understand why they should be put there.

bjadamson:
I agree.

One solution, not saying it's good, is to create an example/ directory in the root and put a small README.md file that explains to the user where the examples are.

kvark:
True. We can put a simple Readme into the root examples folder explaining where to look (examples structure)

bjadamson:
README.md file added. Working on trying to find a version of cargo test --all with --exclude's that works.

bjadamson:
--exclude only seems  to work on nightly. Do we want to use --exclude if it only works on nightly?

kvark:
> cargo test --all --exclude gfx_window_sdl --exclude gfx_device_metal --exclude gfx_device_metalll --exclude gfx_window_metal

wow, if this is really that long, we should have a helper, e.g. a `Makefile`.

bjadamson:
I think the --exclude list is above my understanding atm. I'm not sure which packages should be excluded from which builds.

And it only seems to work with nightly. Do we want to use --exclude list for nightly only? I'm not sure its worth it, what do you all think.

msiglreith:
> --exclude only seems to work on nightly. Do we want to use --exclude if it only works on nightly?

- appveyor fails because we currently specify rust 1.18 instead of 1.19, need to bump the version here
- travis OSX fails because vulkan is not excluded
- travis linux fails due to gflw issue

bjadamson:
@msiglreith thanks. Pushing a new version, I'll keep an eye on it.

kvark:
@msiglreith when CI breaks like that, it feels like the world is falling apart :)

bjadamson:
Not sure why it can't find Vulkan on linux yet. Gonna work on this later.

msiglreith:
Build issue left atm:
* Exclude `gfx_device_dx12` on appveyor-gnu, which could become an issue as `gfx_window_dxgi` depends on dx11 and dx12
* Exclude gfx_device/window_vulkan` on OS X as Mac doesn't support vulkan (duh)

kvark:
Assuming this works out, please squash the commit history up to a few bigger commits.

bjadamson:
Are you ok with one large commit for the feature?

kvark:
Sure thing

kvark:
looks like it's trying to build DX stuff on linux

bjadamson:
Yeah, I don't see why yet. I just found a typo in the travis build script, so I need to let that build. Anyways the windows packages are explicitly excluded, I haven't figured out why there still being compiled.

from build:

> +cargo build --all --features vulkan --exclude gfx_window_glutin --exclude gfx_window_sdl --exclude gfx_device_dx11 --exclude gfx_device_dx12ll --exclude gfx_device_dx12 --exclude gfx_device_metal --exclude gfx_device_metalll --exclude gfx_window_metal

https://github.com/bjadamson/gfx/blob/mod-support/scripts/travis-script.sh#L15-L17

https://github.com/bjadamson/gfx/blob/mod-support/scripts/travis-script.sh#L39

msiglreith:
`gfx_window_dxgi` (the window handling lib for the D3D family) depends on both `gfx_device_dx11` and `gfx_device_dx12`, which isn't listed in the excludes.

bjadamson:
I see. So excluded packages are ignored in this case. Thanks!

bjadamson:
Ok. Rebased ontop of master. Going to wait and see if all ci passes this time.

bjadamson:
I did a merge commit, and it ended up using the previous commit. It did this before, and I had to push another commit to fix it. I'm happy with the commits exactly as they are, and don't want to make a nonsense change just to get the CI to use the latest commit. Any ideas on how to rerun the ci?

bjadamson:
Apparently closing and re-opening the pull-request causes travis to rerun. Trying that

bjadamson:
Ok. This is ready for review. The CI tests are passing.

appveyor.yml
```diff
@@ -15,9 +15,15 @@ install:
...
+    --exclude gfx_device_metal
+    --exclude gfx_device_metalll
+    --exclude gfx_window_metal
 test_script:
-  - cargo test --all --features vulkan
-  - cargo test -p gfx -p gfx_core --features "mint serialize"
-  - if %COMPILER%==msvc cargo test --all --features dx11
-  - if %COMPILER%==msvc cargo test --all --features dx12
+  # re: gfx is gfx_render
+  - cargo test -p gfx -p gfx_core
```
msiglreith: Appveyor should transit to `cargo test --all --excludes ...` later as not everything is tested as far as I can see. But that requires some more changes to the current `gfx_window_dxgi`, which we can/should address later.

examples/README.md
```diff
@@ -0,0 +1,24 @@
+<!--
+    Copyright 2014 The Gfx-rs Developers.
```
msiglreith: nit: year 😁

examples/README.md
```diff
@@ -0,0 +1,24 @@
...
+-->
+
+
+## Where are the examples?
+The examples are split across three crates: gfx_core, gfx_render, gfx_support.
+
+* [gfx_core/examples](../gfx_core/examples) contains low level examples.
+* [gfx_render/examples](../gfx_render/examples) contains examples that show how to use the gfx_render crate directly.
+* [gfx_support/examples](../gfx_support/examples) contains examples that utilize the gfx_support module, showing you how you can build an
+application with minimal setup.
```
msiglreith: nit: unnecessary newline

gfx_core/core/Cargo.toml
```diff
@@ -37,6 +42,5 @@ serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }
 smallvec = "0.4.1"
 
-[features]
-serialize = ["serde", "serde_derive", "draw_state/serialize"]
-unstable = []
+[dev-dependencies]
+gfx = { path = "../../gfx_render", version = "0.16" }
```
msiglreith: We should definitely try to get rid of this dependency in a followup!

gfx_render/Cargo.toml
```diff
@@ -37,5 +38,7 @@ unstable = []
 mint = { version = "0.4.1", optional = true }
 derivative = "1.0"
 draw_state = "0.7"
-gfx_core = { path = "../core", version = "0.7.1" }
+gfx_core = { path = "../gfx_core/core", version = "0.7.1" }
 log = "0.3"
+
+
```
msiglreith: nit: newlines

msiglreith:
> Last thing I'd like to see is that core's Cargo.toml is in gfx_core direction, whereas gfx_corell and gfx_macros live in their own folders in the root.

:+1: 

> Another thing - perhaps gfx_ prefix is unnecessary, given that we are in gfx repo, and we don't have this prefix for the backends?

I'm not having a strong opinion on this one. I'm leaning towards keeping the current prefixes as it's a bit easier to differ between library code/crates and additional stuff. The amethyst project seems employ the same approach.

kvark:
I just see it as inconsistency:
we have `window/glutin` mapping to `gfx_window_glutin`
so I'd expect `core` to map to `gfx_core`
and so on

appveyor.yml
bjadamson: Thanks, yeah I played with test --all, but kept getting a linker error with d312. I wasn't sure, so I opted to keep it as before this time.

examples/README.md
bjadamson: I never know about these... do you want me to change them all? The others are 2014, atleast the one I copy/pasted from

bjadamson:
Ok, I've made the requested changes. I squashed it back down to two commits, one for each issue. The only think I didn't touch (I believe) is the 2014 dates, as they seem completely inconsistent across the project. I can update them all if you want, unless I'm missing the point of the dates. :)

Oh I also didn't make any gfx_ prefix changes.

Here's a convenience link for viewing new directory structure:
https://github.com/bjadamson/gfx/tree/mod-support

examples/README.md
bjadamson: addressed

gfx_core/core/Cargo.toml
bjadamson: agreed!

bjadamson:
Ok, so I have another set of changes I would like to propose for review/merge. However, it is based on this branch. I am unable to figure out how to setup a pull-request for this repository that is based on **my gfx fork** mod-support branch. If I base the pull-request against this repository, it will contain all of the changes from the mod-support branch AND my new changes, which is obviously not what I want.

> Features bundled include:
> 
> * Disallow dead code within gfx_support module.
> * move load_texture to gfx_support public interface and re-use implementation.
> * remove all build warnings from whole project (except one, a platform specific config, not sure how to solve it).
> * Update all copyright dates to 2017. This was brought up in the other pull request, I hope it's ok to lump in with here.
> * Implement triangle example using gfx_support module. Now there are two triangle examples.
> * Allow users of gfx_support to override the default event handling. This allows strictly more control inside the example implementations.
> * Allow examples to run on my older pc. My pc doesn't support gl 150 core, so running the triangle example required some code changes. I went through all the examples, and tried to make them run using the lowest required version of GL available. Now most of the examples can run on my computer, without code change.

Since I couldn't figure that out, I opened a pull-request against my own fork. The pull-request contains a bunch of QOL improvements that I found myself when working with gfx and gfx_support.
https://github.com/bjadamson/gfx/pull/1

I'd love to discuss that pull request, but until this gets merged, I think this is the only way to move forward.

msiglreith:
Thanks for the changes!
I noticed that [`gfx_core`](https://github.com/bjadamson/gfx/tree/mod-support/gfx_core) contains the `trianglell` example which should be inside `gfx_corell`. Would be also nice to get rid of the additional `Cargo.toml` for the example.

@kvark So we could either add missing prefixes or remove them alltogether. As mentioned earlier I'm leaning to towards keeping them as it's easier then to see what's a crate and what to search for on crates.io, even though `window/gfx_window_dxgi` or `backend/gfx_device_vulkan` might be a bit redundant.
(We could also just put everything into the root directory)

msiglreith:
Regarding the follow-up PR you mentioned. I haven't looked into it in detail but my thoughts about the feature list:

> - Disallow dead code within gfx_support module.
> - move load_texture to gfx_support public interface and re-use implementation.
> - Update all copyright dates to 2017. This was brought up in the other pull request, I hope it's ok to lump in with here.
> - Implement triangle example using gfx_support module. Now there are two triangle examples.

Sounds good! 👍

> - remove all build warnings from whole project (except one, a platform specific config, not sure how to solve it).

Unsure about this because a lot of them are caused by WIP code and sometimes just indicate that a function hasn't been implemented, where `#[allow(unused)]` (or whatever it's called) would be nice. At some cases it could also be missing implementation. Requires careful reviewing here!

> - Allow users of gfx_support to override the default event handling. This allows strictly more control inside the example implementations.

Do we have a use case here? Even though I'm not opposed to the change

> - I went through all the examples, and tried to make them run using the lowest required version of GL available (1.10 in most cases).

Sweet, looking forward to it! 1.10 would be GL 2.0 iirc?

kvark:
@bjadamson sorry for blocking your PR on a naming problem! naming is hard...
I do believe that we need to sort this question nicely at this time, there will be no other mass rename PRs in the nearest future (I hope!).

>> remove all build warnings from whole project (except one, a platform specific config, not sure how to solve it).
> Unsure about this because a lot of them are caused by WIP code ...

I sort of agree. It's a valid warning for something to be fixed/implemented. Sort of a compile-time variant of `unimplemented!`, so I think it should stay.

@msiglreith 
ok, so to recap, we have the following options:
  1. omit `gfx_` prefix consistently for all crates, thus having `/core`, `/macros`, `/render` as well as `/window/glutin`, etc
  2. have `gfx_` prefix consistently, e.g. `/gfx_core`, `/gfx_macros`, `/window/gfx_window_glutin`, etc
  3. have `gfx_` prefix inconsistently for root crates only, like it is now
  4. have all crates stay in the root with full names, e.g. `/gfx_core`, `/gfx_window_glutin`

I require consistency, so (3) is not an option for me.
(2) seems quite a bit verbose, especially since it's not DRY (repeating the `window` part, for example)
(4) looks too crowded in the root for my taste... I'd like to have a clear difference between folders that are crates and folders that are standard to cargo (e.g. `/target`, `/examples`, etc).
(1) actually suffers from the same problem as (4)

Thus, my proposal is to modify (4) to solve the problem:
  4a. put all the fully qualified crates under `/libs` folder, e.g. `/libs/gfx_core`, `/libs/gfx_window_glutin`, etc

kvark:
Actually, the proposed 4a is worse than what we have now:
5. `/src/core`, `/src/render`, `/src/window/glutin`, etc

That is my vote then - keep this scheme.

bjadamson:
> * remove all build warnings from whole project (except one, a platform specific config, not sure how to solve it).
>
> Unsure about this because a lot of them are caused by WIP code and sometimes just indicate that a function hasn't been implemented, where #[allow(unused)] (or whatever it's called) would be nice. At some cases it could also be missing implementation. Requires careful reviewing here!

Yah. The only thing I did was go through the project and rename them mostly to have underscores infront of their name. Careful review might find a renamed some of them to plain "_" instead of leaving the original name and just prefixing it with a _. It shouldn't be many instances.

> Do we have a use case here? Even though I'm not opposed to the change

You don't, but I do. I would like to start using the gfx_support machinery outside of this crate (in a limited way first). I guess I hadn't brought this up until now. I'd like to not make the maintencance of this project suffer at all because of it, but I can't really find a valid reason not to reuse gfx_support in my own game project. It hides a lot of the backend initialization.

> Sweet, looking forward to it! 1.10 would be GL 2.0 iirc?
According to [this](https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)#OpenGL_and_GLSL_versions), yeah.

bjadamson:
> I noticed that gfx_core contains the trianglell example which should be inside gfx_corell. Would be also nice to get rid of the additional Cargo.toml for the example.

Whoops! That's an easy fix though. Thanks!

bjadamson:
Ok the push is hot. Fingers crossed on the CI.

bjadamson:
Whoops, I pushed all the commits on the wrong branch. I'll upload another one

src/backend/vulkanll/src/lib.rs
```diff
@@ -53,9 +52,11 @@ lazy_static! {
 }
 
 pub struct QueueFamily {
+    #[allow(dead_code)]
```
kvark: please remove those too

src/core/src/dummy.rs
```diff
@@ -83,41 +83,38 @@ impl Device<DummyResources> for DummyDevice {
...
         unimplemented!()
     }
-    fn create_buffer_immutable_raw(
-        &mut self,
-        _: &[u8],
-        _: usize,
-        _: buffer::Role,
-        _: Bind,
-    ) -> Result<handle::RawBuffer<DummyResources>, buffer::CreationError> {
+    fn create_buffer_immutable_raw
```
kvark: please, no. This is changing from the new style to the old one.

src/core/src/lib.rs
```diff
@@ -265,51 +267,50 @@ pub trait Backend: 'static + Sized {
...
-    type Fence:               Debug + Any + Send + Sync;
-    type Heap:                Debug + Any;
-    type Mapping;
-    type DescriptorHeap:      Debug + Any;
-    type DescriptorSetPool:   Debug + Any;
-    type DescriptorSet:       Debug + Any;
-    type DescriptorSetLayout: Debug + Any;
-}
-*/
+// Different resource types of a specific API.
```
kvark: I don't understand why these comments needed to be changed

src/core/src/memory.rs
```diff
@@ -41,55 +41,55 @@ pub enum Usage {
...
-    /// - [`DEPTH_STENCIL`](constant.DEPTH_STENCIL.html)
-    /// - [`SHADER_RESOURCE`](constant.SHADER_RESOURCE.html)
-    /// - [`UNORDERED_ACCESS`](constant.UNORDERED_ACCESS.html)
-    /// - [`TRANSFER_SRC`](constant.TRANSFER_SRC.html)
-    /// - [`TRANSFER_DST`](constant.TRANSFER_DST.html)
-    ///
-    ///
-    /// This information is used to create resources
-    /// (see [gfx::Factory](trait.Factory.html#overview)).
+/// Flags providing information about the usage of a resource.
```
kvark: similarly, here and in *all other* bitflag usages:
I think having indentation looks much cleaner, I don't see a point of removing it from the comments

src/corell/Cargo.toml
```diff
@@ -30,3 +31,6 @@ path = "src/lib.rs"
 bitflags = "0.8"
 log = "0.3"
 draw_state = "0.6"
+
+[dev-dependencies]
+gfx = { path = "../render", version = "0.16" }
```
kvark: needs a //TODO: remove

src/corell/examples/Cargo.toml
```diff
@@ -0,0 +1,67 @@
+[package]
+name = "gfx-corell-examples"
```
kvark: I might be missing something here. Why do we have a `Cargo.toml` in the examples folder?

src/corell/examples/trianglell/main.rs
```diff
@@ -50,15 +51,31 @@ struct Vertex {
...
-const TRIANGLE: [Vertex; 6] = [
-    Vertex { a_Pos: [ -0.5, 0.33 ], a_Uv: [0.0, 1.0] },
-    Vertex { a_Pos: [  0.5, 0.33 ], a_Uv: [1.0, 1.0] },
-    Vertex { a_Pos: [  0.5,-0.33 ], a_Uv: [1.0, 0.0] },
-
-    Vertex { a_Pos: [ -0.5, 0.33 ], a_Uv: [0.0, 1.0] },
-    Vertex { a_Pos: [  0.5,-0.33 ], a_Uv: [1.0, 0.0] },
-    Vertex { a_Pos: [ -0.5,-0.33 ], a_Uv: [0.0, 0.0] },
-];
+const TRIANGLE: [Vertex; 6] = [Vertex {
```
kvark: please no

src/corell/examples/trianglell/main.rs
```diff
@@ -122,25 +142,19 @@ fn main() {
...
-    let set1_layout = factory.create_descriptor_set_layout(&[
-            DescriptorSetLayoutBinding {
-                binding: 0,
-                ty: DescriptorType::Sampler,
-                count: 1,
-                stage_flags: shade::STAGE_PIXEL,
-            }
-        ],
-    );
+    let set0_layout = factory.create_descriptor_set_layout(&[DescriptorSetLayoutBinding {
```
kvark: again, please let's keep the old style

src/render/examples/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "gfx-render-examples"
```
kvark: similarly, why do we have a `Cargo.toml` here?

src/support/examples/blend/main.rs
```diff
@@ -45,85 +43,67 @@ gfx_defines!{
...
-    "Dodge",
-    "Burn",
-    "Overlay",
-    "Multiply",
-    "Add",
-    "Divide",
-    "Grain Extract",
-    "Grain Merge",
-];
+const BLENDS: [&'static str; 9] = ["Screen",
```
kvark: mmm, no

src/support/examples/cube/main.rs
```diff
@@ -58,80 +59,74 @@ impl Vertex {
...
-
-        let index_data: &[u16] = &[
-             0,  1,  2,  2,  3,  0, // top
-             4,  5,  6,  6,  7,  4, // bottom
-             8,  9, 10, 10, 11,  8, // right
-            12, 13, 14, 14, 15, 12, // left
-            16, 17, 18, 18, 19, 16, // front
-            20, 21, 22, 22, 23, 20, // back
-        ];
+        let vertex_data = [// top (0, 0, 1)
```
kvark: the new code doesn't look better

src/support/examples/deferred/shader/blit.glslf
```diff
@@ -1,4 +1,4 @@
-#version 150 core
+#version 140
```
kvark: I don't think you can use GLSL_140 here since it has `out vec4 Target0`. You'll need to change the use into `gl_FragColor`

bjadamson:
Oh man, thank you for reviewing! I didn't actually mean to push those commits together.. I'm still working on making the final push. I screwed up, sorry! I didn't mean to make you review this meaninglessly yet :(

src/backend/gl/src/info.rs
```diff
@@ -229,7 +235,11 @@ impl Info {
         self.extensions.contains(&s)
     }
 
-    pub fn is_version_or_extension_supported(&self, major: u32, minor: u32, ext: &'static str) -> bool {
+    pub fn is_version_or_extension_supported(&self,
```
bjadamson: All these formatting changes I didn't mean to push on this branch

appveyor.yml
kvark: Well, if you ask me, I'd rather remove *all* those headers entirely. They are just cluttering the source.
Last time we discussed this with the gfx-rs core team (~3 years ago? lol), the main argument for keeping the comments was that Rust uses the same scheme. Hard to counter that  :(

README.md
```diff
@@ -104,12 +104,35 @@ Those examples are automatically downloaded if you clone the gfx directory:
 	$ git clone https://github.com/gfx-rs/gfx
 
 where `<my_dir>` is a directory name of your choice. Once gfx is downloaded you can build any of the gfx examples.
-The examples are listed in the `<my_dir>/gfx/Cargo.toml` file.
-For example try:
+The examples are split across three crates: gfx_core, gfx_render, gfx_support.
```
bjadamson: Done.

README.md
```diff
@@ -122,7 +145,7 @@ For gfx to work, it needs access to the graphics system of the OS. This is typic
 gfx can use a couple of those to acquire graphical contexts.
 For example; [glfw](https://github.com/PistonDevelopers/glfw-rs) or [glutin](https://github.com/tomaka/glutin/).
 
-To see how the graphic context is acquired, see the [cube example](https://github.com/gfx-rs/gfx/tree/master/examples/cube) or the [triangle example](https://github.com/gfx-rs/gfx/tree/master/examples/triangle).
+To see how the graphic context is acquired, see the [cube example](https://github.com/gfx-rs/gfx/tree/master/gfx_support/examples/cube) or the [triangle example](https://github.com/gfx-rs/gfx/tree/master/gfx_render/examples/triangle).
```
bjadamson: fixed, thanks!

examples/README.md
```diff
@@ -0,0 +1,23 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+
+## Where are the examples?
+The examples are split across three crates: core, render, support.
+
+* [gfx_corell/examples](../src/corell/examples) contains low level examples.
```
bjadamson: done

info/contrib.md
```diff
@@ -28,15 +28,8 @@ Finally, feel free to hop on [#rust-gamedev](http://chat.mibbit.com/?server=irc.
...
 
-* _examples_ : GFX's examples
-* _tests_ : GFX's tests
-* _info_ : Information and documentation
-* _src_ : gfx_app, an application framework for GFX
-* _src/core_ : gfx_core, core structures and the interface that backends must provide
-* _src/backend_ : The backends implementations
-* _src/render_ : The main gfx package
-* _src/window_ : Different backends to create windows and initialize their graphics
+* in progress, since this repository is under such change at the moment maintaining this list seems unimportant right now.
```
bjadamson: You are correct.

src/support/examples/triangle/shader/triangle_110.glslv
```diff
@@ -1,8 +1,8 @@
-#version 150 core
+#version 110
```
kvark: we still need GLSL 150 versions of triangle shaders, since OSX doesn't support compatibility GL profile and can't run GLSL 110 shaders...

README.md
```diff
@@ -104,25 +104,52 @@ Those examples are automatically downloaded if you clone the gfx directory:
 	$ git clone https://github.com/gfx-rs/gfx
 
 where `<my_dir>` is a directory name of your choice. Once gfx is downloaded you can build any of the gfx examples.
-The examples are listed in the `<my_dir>/gfx/Cargo.toml` file.
-For example try:
+The examples are split across three crates: core, render, support. You can find
 
-	$ cd gfx
+1. core/examples contains low level examples.
```
kvark: should be `src/core/examples`, and so on

src/corell/Cargo.toml
bjadamson: added

src/corell/examples/Cargo.toml
bjadamson: I wasn't originally able to make it work using one Cargo.toml file. I could probably figure it out now that I've been working on this so much, but right now it keeps the dependencies separated. I'm not saying that's a good reason to keep it this way... I'm just not super interested in iterating again today.

If you want to block on this, I'm happy to pick it up again tomorrow and see if I can figure it out. Up to you

src/render/examples/Cargo.toml
bjadamson: same

README.md
bjadamson: all fixed

appveyor.yml
bjadamson: Can we just a LICENSE file in the root directory?

src/support/examples/triangle/shader/triangle_110.glslv
bjadamson: I didn't mean to make any of these changes, sorry for the confusion and extra code to review!

bjadamson:
Okay. Letting the CI run for a while. I have one more thing to do (figure out the extra Cargo.toml files) inside the example directories, see if I can get rid of them. Probably figure that out sometime in the next  24 hours, we'll see though!

bjadamson:
Ok, yay CI passed!

Moving on... I went through the exercise of trying to remove the extra Cargo.toml files within the examples/ directories. I remember now why I moved the examples/ to their own Cargo.toml file. Without it, I'm running into cyclic dependencies.

See the corell/triangle example. The corell/triangle example [depends on gfx/corell](https://github.com/bjadamson/gfx/blob/mod-support/src/corell/examples/Cargo.toml#L23). Please note these dependencies are not optional dependencies.

The corell/triangle example also [depends on gfx_device_gl](https://github.com/bjadamson/gfx/blob/mod-support/src/corell/examples/Cargo.toml#L31-L34).

When I moved src/corell/examples/Cargo.toml's contents up one directory to the src/corell/Cargo.toml file, I find a cyclic dependency. src/corell depends on src/backend/vulkanll (this cannot be a dev dependency, cargo does not allow dev dependencies to be optional). src/backend/vulkanll has a [dependency on src/corell](https://github.com/bjadamson/gfx/blob/mod-support/src/backend/vulkanll/Cargo.toml#L33).

Because of this cyclic dependency, I remember (now) opting for an extra Cargo.toml file within the examples/ directory. I'm open to a better solution, but I feel like the extra Cargo.toml file is the best option I am aware of.

kvark:
@bjadamson ah, ok, I see now. Thanks for the explanation!
I find `cd src/corell/examples && cargo run --example trianglell` to be rather verbose and not too convenient either. If we are to have a project for examples, we might as well just have it owning *all* of the examples and placed at the root, e.g.
`cd examples && cargo run --example trianglell`

bjadamson:
I agree... but I'm unsure how to satisfy both of your requests. Earlier in this thread, you requested the examples moved into their appropriate repository.

> We just need to push it all the way down:

> trianglell will be inside gfx_core/examples
> triangle will be inside gfx_render/examples
> glfw/sdl examples will be left untouched
> all other examples will be in gfx_support/examples

Point 2.
> I find cd src/corell/examples && cargo run --example trianglell to be rather verbose and not too convenient either

I agree it is verbose, but has not been that inconvenient in practice. If I'm committing to using GFX, it isn't a big deal to set my working directory appropriately. What do you want me to do?

kvark:
@bjadamson Unless an example is heavily tied to the windowing subsystem (like the ones we have for sdl and glfw), let's put all of them into `/examples` for simplicity. Having multiple examples folders would only be good if these were real examples without a separate project.

Also, given a separate project in `/examples`, I think we can have them as binaries in order to do:
```
cd /examples && cargo run --bin trianglell`
```

bjadamson:
Ok, mind changed, I'll do that.

bjadamson:
I just realized a side effect is I'll need different names for the two triangle examples.

I just want to note one down side is that is again unclear which examples use the support modules and which don't, without looking in the source of the examples itself.

bjadamson:
Actually @kvark, does my last point do anything to sway your opinion at all? I'm currently working on the changes you requested, but I'm actually finding it to be a reduction in clarity for me as a user of the lib.

kvark:
@bjadamson hold on, I'm not sure about the best solution here :(

We have different sorts of examples:
  - window-tied ones (SDL/GLFW): trivially put into the corresponding window crates
  - support ones: could be put into the support crate
  - triangle, gamma: ??? - use gfx_render as well as gfx_window_glutin
  - trianglell: ??? - uses gfx_corell, gfx_device_vulkanll, gfx_device_metall, gfx_device_d3d12ll

For the `triangle` & `gamma` - how about we put `gfx_window_glutin` as a dev dependency of gfx_render and move those examples into /src/render/examples`? That wouldn't have any dependency cycles, and would not require a separate solution

Only unclear part is `trianglell`, and for a good reason - we aren't going to have the compile time backend selection for other LL examples if we have any. One choice would be to have a basic `supportll` crate, having `triangle` example. I assume it will be obsolete when LL transition is done. What do you think, @msiglreith?

kvark:
@bjadamson 
> I'm not thrilled about making more changes like this...

By no means I want to pile more work on you for this PR. All I want is to wrap it up nicely.

Would it be most straightforward if we have an `/examples/Cargo.toml` with a single binary pointing to `/examples/triangell/` and the corresponding dependencies?

bjadamson:
I have a tendency to give up early... character flaw I'm recognizing..so I'll try not to give up :)

Can you help me understand if this is what you intend?

examples/Cargo.toml -- new, contains dependencies for trianglell example
examples/triangleell/ -- moved from src/corell/examples??

src/support/examples -- stays?
src/render/examples -- stays?

Users still have to change directory to the right examples directory. Do I understand correctly?

kvark:
@bjadamson 
> so I'll try not to give up

That's the spirit!

> examples/Cargo.toml -- new, contains dependencies for trianglell example
> examples/triangleell/ -- moved from src/corell/examples??

correct. `corell/examples` is not needed then

src/support/examples -- stays?
src/render/examples -- stays?

Correct, but without dedicated `Cargo.toml` configs in the examples folders, so you'd run those as:
```
cd src/render && cargo run --example triangle
```

> Users still have to change directory to the right examples directory. Do I understand correctly?

Yes

msiglreith:
The cargo limitations are quite unfortunate 😟 
I think it's safe to say that the window crates should stay where they are as they don't relay on multiple other crates.

My suggestion would be to move the others into `examples` with a single root Cargo.toml and create subdirectories for `core`/`corell` for the triangle examples.
Our examples mostly show how to achieve a certain rendering effect or how to use an API feature, which is probably the main aspect of them (showcase). Therefore I don't really care that much about inconsistency here (sorry!). The goal of `gfx_app/support` was to avoid the overhead of writing the initialization for our samples/demos/examples.
Splitting the `core`/`corell` parts because they are more low-level and due to the upcoming widening of the gap between core and render we might have need for more `core`-only examples in the future.

Regarding the length of this discussion, we could also make a cut here and split it into 2 PRs depending on how long it takes to find a consensus to avoid rushed decision!

bjadamson:
Allright, well while you guys all decide, I'll just wait for consensus.

My personal vote would be what's checked in now.

bjadamson:
Reading through gitter, it seems a lot of people have a desire for a more structured project directory. It just seems naming is hard, and it will take some time to agree on a final solution. Reading through the gitter, we're still in the early stages of deciding on a "best" structure for the repository.

I'm very eager to get this branch checked in in some capacity.. I would like to work on other aspects of gfx, but I don't very much look forward to doing work on a separate branch, to only have to rebase that work after this eventually gets merged.

Would a compromise be considered from the core team?

Right now, the commit as-is is an improvement over the status quo. If anything for switching the project to use a virtual workspace, we gained cargo test --all on OSX and linux platforms. Would an acceptable path forward be to open a new issue targeting the naming of the different projects/structures with as much context (gitter chat, links to this PR) be sufficient to allow the conversation to continue, but also gain the (perceived) benefits of merging this?

The main downside I see is that sometime down the line, it is possible (who knows maybe noone will care) we'll need to do another naming refactor. It's not any different than today though in that regard.

bjadamson:
If there's anything that's been decided, I'm happy to do that work ASAP.

msiglreith:
Summarizing the small gitter discussion from yesterday:
kvark's extension proposal
```rust
/examples
  Cargo.toml // specifying paths for **binaries**
  /corell
    /triangle
  /render
    /triangle
    /gamma
  /support
    /deferred
    /blend
    /...
```
I would agree on this structure being the interim goal for this PR. The window examples are left in the current state.
cc @kvark

bjadamson:
That structure seems reasonable to me. If there's consensus I can make one last switch to that.

bjadamson:
Specifying paths to the binaries though, I don't think I'm used to programming with rust enough yet to understand exactly what is meant by that. Could you elaborate a little bit?

msiglreith:
@bjadamson Specifying the path as done [here](https://github.com/gfx-rs/gfx/blob/master/Cargo.toml#L88-L90).

bjadamson:
@msiglreith Oh ok, thanks.

kvark:
@bjadamson sorry for not giving you an open path for finishing this PR. I'd hope to see this merged as much as you do.
Problem is - we are in a situation where no good solution for the directory structure satisfies everyone, and gaining making any changes to this design involves shuffling lots of stuff around (e.g. moving examples, renaming `support`, etc)...

The proposal @msiglreith linked to is a compromise in itself, given that I wanted to only have LL examples in the `/examples` folder and everything else in the corresponding projects. Nobody is happy about this, so it's hard to universally accept :)

> That structure seems reasonable to me. If there's consensus I can make one last switch to that.

It doesn't appear that we found a better one, so let's accept this as a consensus.

> Specifying paths to the binaries though, I don't think I'm used to programming with rust enough yet to understand exactly what is meant by that. Could you elaborate a little bit?

I mean having the examples as `[[bin]]` entries as opposed to `[[example]]`, given that we are going into `/examples` folder to launch them anyway, and `bin` is shorter. See http://doc.crates.io/manifest.html

bjadamson:
Ok, I'm going to do that work. If another consensus comes up in the interim I'll be on gitter if you'd be kind enough to stop me (;

bjadamson:
Oh one thing. I have two triangle examples. One using render, one use support. Since it is requested that there is one Cargo.toml file with all the examples laid out, they each need a unique name.

How should I name them? I see two strategies, but there are more I'm sure.

triangle_render
triangle?

This would mean all the support examples keep their same name.

or
triangle
triangle_support

This would involve renaming all the support examples to have  a "_support " suffix. This is of course to maintain consistency.

bjadamson:
Ok, I sent the build off. Assuming I didn't make any mistakes (or missing any documentation that needs another last minute update to point to the right directory) then it's ready minus the issue I mentioned above.

Here's the link though so it's concrete in case my description was unclear.
https://github.com/bjadamson/gfx/blob/mod-support/examples/Cargo.toml#L160-L163
https://github.com/bjadamson/gfx/blob/mod-support/examples/Cargo.toml#L151-L153

Just need an arbitrary naming convention for the examples that is consistent. I don't think triangle_render is particularly good, I'm just unsure if renaming all of the examples to have a "_support" suffix is desired other.

README.md
```diff
@@ -133,13 +122,18 @@ To use `glutin`, for example, your `Cargo.toml` must be extended with the follow
...
+[accompanying documentation](./examples/README.md) for understanding and running those examples.
+
+
 ## Structure and current versions
 `gfx` consist of several crates. You can find all of them in this repository.
 
 | Core functionality: | Graphic backends: | Window backends: |
 | :---: | :---: | :---: |
 | [![gfx on crates.io](http://img.shields.io/crates/v/gfx.svg?label=gfx)](http://crates.io/crates/gfx) | [![gfx_device_gl on crates.io](http://img.shields.io/crates/v/gfx_device_gl.svg?label=gfx_device_gl)](http://crates.io/crates/gfx_device_gl) | [![gfx_window_sdl on crates.io](http://img.shields.io/crates/v/gfx_window_sdl.svg?label=gfx_window_sdl)](http://crates.io/crates/gfx_window_sdl) |
-| [![gfx_app on crates.io](http://img.shields.io/crates/v/gfx_app.svg?label=gfx_app)](http://crates.io/crates/gfx_app) | [![gfx_device_dx11 on crates.io](http://img.shields.io/crates/v/gfx_device_dx11.svg?label=gfx_device_dx11)](http://crates.io/crates/gfx_device_dx11) | [![gfx_window_dxgi on crates.io](http://img.shields.io/crates/v/gfx_window_dxgi.svg?label=gfx_window_dxgi)](http://crates.io/crates/gfx_window_dxgi) |
```
msiglreith: Why is this row removed?

examples/README.md
```diff
@@ -0,0 +1,33 @@
...
+The gfx-rs git repository contains a number of examples.
+Those examples are automatically downloaded if you clone the gfx directory:
+
+	$ cd <my_dir>
+	$ git clone https://github.com/gfx-rs/gfx
+
+where `<my_dir>` is a directory name of your choice. Once gfx is downloaded you can build any of the gfx examples.
+The examples are split across three directories, each pertaining to the GFX library they are using.
+
+1. examples/core contains low level examples.
```
msiglreith: should be `examples/corell` as we currently don't have pure `core` examples

examples/corell/Cargo.toml
```diff
@@ -0,0 +1,67 @@
+[package]
```
msiglreith: This file seems outdated.

examples/render/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
```
msiglreith: This file seems outdated.

examples/render/triangle/shader/triangle_120.glslf
```diff
@@ -1,4 +1,4 @@
-#version 120
+#version 110
```
msiglreith: Is this changed intended? If so, please rename the file.

examples/render/triangle/shader/triangle_120.glslv
```diff
@@ -1,4 +1,4 @@
-#version 120
+#version 110
```
msiglreith: Same as above.

examples/support/triangle/README.md
```diff
@@ -0,0 +1,25 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+
+# Triangle Example
+
+A simple example showing how to render a colored cube using vertex and index
```
msiglreith: Description is wrong, should be for a 🔺

examples/support/triangle/shader/triangle_120.glslf
```diff
@@ -1,4 +1,4 @@
-#version 120
+#version 110
```
msiglreith: Same as above.

examples/support/triangle/shader/triangle_120.glslv
```diff
@@ -1,4 +1,4 @@
-#version 120
+#version 110
```
msiglreith: Same as above.

info/contrib.md
```diff
@@ -28,15 +28,17 @@ Finally, feel free to hop on [#rust-gamedev](http://chat.mibbit.com/?server=irc.
...
-* _examples_ : GFX's examples
-* _tests_ : GFX's tests
-* _info_ : Information and documentation
-* _src_ : gfx_app, an application framework for GFX
-* _src/core_ : gfx_core, core structures and the interface that backends must provide
-* _src/backend_ : The backends implementations
-* _src/render_ : The main gfx package
+* _src/backend : The different backends GFX supports.
+* _src/core : gfx_core, core structures and the interface that backends must provide
+* _src/corell : WIP, will be absorbed into src/core and be.
```
msiglreith: 'and be.'
Not a native speaker, but I assume there is missing something.

src/corell/Cargo.toml
```diff
@@ -30,3 +31,7 @@ path = "src/lib.rs"
 bitflags = "0.8"
 log = "0.3"
 draw_state = "0.6"
+
+# TODO: remove
+[dev-dependencies]
+gfx = { path = "../render", version = "0.16" }
```
msiglreith: I think we can remove this again?

src/support/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+documentation = "https://docs.rs/gfx_support"
+workspace = "../.."
+
+[lib]
+name = "gfx_support"
+
+[features]
+default = ["gl"]
+gl = ["gfx_device_gl", "gfx_window_glutin"]
+
```
msiglreith: Other backend features should be added here also.

README.md
bjadamson: It was indicated earlier in the discussion that it was desirable to kill of gfx_app all together.

examples/README.md
bjadamson: fixed

examples/corell/Cargo.toml
bjadamson: Oops! File is entirely removed now.

examples/render/Cargo.toml
bjadamson: Oops! File is entirely removed now.

examples/render/triangle/shader/triangle_120.glslf
bjadamson: no, not for this PR. Reverted.

examples/render/triangle/shader/triangle_120.glslv
bjadamson: fixed

examples/support/triangle/README.md
bjadamson: fixed

info/contrib.md
bjadamson: I'll just get rid of the "and be". I was going to indicate something along the lines that corell and core are going to be merged.. but it doesn't seem useful now.

src/corell/Cargo.toml
bjadamson: Ok, I removed it. I can't run the vulkan examples obviously, but the behavior seems correct.

src/support/Cargo.toml
bjadamson: Yeah, fixed. Downsides of testing on one platform.

bjadamson:
Allright, changes requested pushed. I'm wondering if the dependency on dxgi in support/ is necessary, but I don't have a windows box for another week to test it on. It complains that dx11 and dx12 require dxgi, so I just left it like that. Thanks for such careful reviews @msiglreith, I think I'm becoming blind to the mistakes when reviewing because I've looked at it to many times.

README.md
msiglreith: Ok, but what about the other crates?

info/contrib.md
```diff
@@ -28,15 +28,17 @@ Finally, feel free to hop on [#rust-gamedev](http://chat.mibbit.com/?server=irc.
...
 
-* _examples_ : GFX's examples
-* _tests_ : GFX's tests
-* _info_ : Information and documentation
-* _src_ : gfx_app, an application framework for GFX
-* _src/core_ : gfx_core, core structures and the interface that backends must provide
-* _src/backend_ : The backends implementations
-* _src/render_ : The main gfx package
+* _src/backend : The different backends GFX supports.
+* _src/core : gfx_core, core structures and the interface that backends must provide
```
kvark: We need to be consistent here by either naming the crates explicitly for all entries, or omitting them for all (simpler).
Also, trailing point is missing at the end.

examples/Cargo.toml
```diff
@@ -0,0 +1,163 @@
+[package]
+name = "gfx-examples"
```
kvark: should be `gfx_examples` for consistency with other names

examples/Cargo.toml
```diff
@@ -0,0 +1,163 @@
...
+version = "0.7.0"
+description = "GFX example application framework"
```
kvark: need this updated, it's not a framework

examples/Cargo.toml
```diff
@@ -0,0 +1,163 @@
+[package]
+name = "gfx-examples"
+version = "0.7.0"
```
kvark: should be 0.1.0
also, quite a few of the fields are not needed here since this project is never going to be published

examples/Cargo.toml
```diff
@@ -0,0 +1,163 @@
...
+description = "GFX example application framework"
+homepage = "https://github.com/gfx-rs/gfx"
+keywords = ["graphics", "gamedev"]
+license = "Apache-2.0"
+authors = ["The Gfx-rs Developers"]
+documentation = "https://docs.rs/gfx_project"
```
kvark: broken link, better remove this at all

examples/Cargo.toml
```diff
@@ -0,0 +1,163 @@
...
+
+cgmath = "0.15"
+gfx_gl = "0.3"
+rand = "0.3"
+genmesh = "0.5"
+noise = "0.4"
+image = "0.15"
+
+gfx = { path = "../src/render", version = "0.16" }
+gfx_core = { path = "../src/core", version = "0.7.1" }
```
kvark: do we actually need *all* the dependencies here?
I'd assume that we only need corell, render, support, window_glutin, and a few things for trianglell

examples/support/triangle/shader/triangle_120.glslf
kvark: still not fixed

examples/support/triangle/shader/triangle_120.glslv
kvark: still not fixed

info/contrib.md
bjadamson: Fixed trailing point, removed reference to gfx_core. Did I understand correctly?

examples/Cargo.toml
bjadamson: Fixed and I removed all non-essential fields (keep name, version, workspace)

examples/Cargo.toml
bjadamson: fixed

examples/Cargo.toml
bjadamson: removed (unless you want this kept I can add it back)

examples/Cargo.toml
bjadamson: removed

examples/Cargo.toml
bjadamson: I was able to remove:
gfx_macros

I needed to keep all the rest to build the examples, since they all use one Cargo.toml file now. Unless I'm missing something, I don't see how we would remove any more dependencies listed in this file and still have the examples runnable.

bjadamson:
All fixed. Sorry about the 110 version not being fixed, I don't know what I did to undo that fix before pushing last time. Maybe it was my rebase, I'll be more careful.

info/contrib.md
bjadamson: assuming thumbs up means I did

examples/Cargo.toml
kvark: Thanks! We can always clean up more later on.

README.md
bjadamson: Didn't realize multiple items on one line. Fixed!

bjadamson:
Ok, I believe it's all in order finally. Except for the last squash. GNU windows build is the only CI that doesn't pass. I have no idea why it's failing to link only on that platform. Does anyone have an idea why?

I could add an if check to see if it's compiling GNU in the appvoyer script, and try excluding the dx12ll backend package, but that feels like a hack.

msiglreith:
Excluding it for GNU would be fine as there seems to be no support for d3d12 yet

bjadamson:
Ok, GNU excluded. Because of that, I was able to put in "cargo test --all" into the appvoyer builds. So that *should* be cargo test --all for each platform.

Syntax is a bit wieldly though, yml is weird and I don't see an obvious way to reduce repetition and maintain if/else branching.
https://github.com/bjadamson/gfx/blob/mod-support/appveyor.yml#L18-L56

bjadamson:
Oh, also gained this line, I assume it's desirable to add. I'm not exactly sure how test --all and --features flags work. I'm not sure if the three lines testing dx11 dx12 and dx12ll are redundant with cargo test --all.
https://github.com/bjadamson/gfx/blob/mod-support/appveyor.yml#L56

bjadamson:
Looks like all the tests passed. The host running the "linux nightly" compiler seems to have died mid run, I'm pushing a noop update to force the CI to rerun.

Is there a better way to get the travis hosts to rerun when this happens?

Anyways, I think it's ready.

examples/render/triangle/main.rs
```diff
@@ -71,8 +71,8 @@ pub fn main() {
     let views = swap_chain.create_color_views(&mut device);
 
     let pso = device.create_pipeline_simple(
-        include_bytes!("shader/triangle_150.glslv"),
-        include_bytes!("shader/triangle_150.glslf"),
+        include_bytes!("shader/triangle_120.glslv"),
```
kvark: doing that would make it fail on OSX, lt's leave GLSL 150 there

examples/skybox/compile.cmd
```diff
@@ -1,5 +0,0 @@
-@echo off
```
kvark: this file should probably be moved instead of removed

examples/support/triangle/main.rs
```diff
@@ -0,0 +1,111 @@
+// Copyright 2017 The Gfx-rs Developers.
```
kvark: any idea why this file is marked as new as opposed to moved?

src/support/Cargo.toml
```diff
@@ -0,0 +1,98 @@
...
+env_logger = "0.4"
+glutin = "0.9"
+winit = "0.7"
+gfx_core = { path = "../core", version = "0.7.1" }
+gfx = { path = "../render", version = "0.16" }
+
+gfx_device_gl = { path = "../backend/gl", version = "0.14", optional = true }
+gfx_window_glutin = { path = "../window/glutin", version = "0.17", optional = true }
+
+[dev-dependencies]
```
kvark: this whole section should no longer be needed

bjadamson:
Holy crap it happened. I'll try and take a look at your mentions in a few days (packing up my pc tonight to move).

What timing, thanks again for all the encouragement :)

examples/support/triangle/main.rs
bjadamson: I'm assuming because it didn't exist before hand. I rewrote the basic triangle example using the support module, so users could compare implementing a basic triangle with and without the support module.

Issue #1416, opened 2017-08-04
[ll] Command Buffer Rework


msiglreith:
_This is part of the intergration process and the low level API development._
# Summary
[summary]: #summary

Rework the current `Buffer` to match the API exposed by the next-gen APIs (Vulkan, D3D12 and Metal). Including a strong-typed interface to ensure compatibilty with queues and pools (don't allow to submit compute commands to graphics queues!) 

# Motivation
[motivation]: #motivation

This rework is motivated by the on-going shift towards a low-level API. Allowing us provide a command buffer interface with minimal overhead. Additionally, provide currently lacking compute support.

> Note: Reworking the cmd buffers will affect large parts of the current core and render API, but we don't list all points which might require changes here, focusing only of the major parts/approach.

# Detailed design

We provide a `RawCommandBuffer` trait which needs to be implemented for each backend. It's functionality is similar to `vkCommandBuffer` and `ID3D12GraphicsCommandLists`. We noticed that the APIs only expose one command buffer type and don't differ between functionality. This allows us to provide safe shim-wrappers for our `RawCommandBuffer` which ensure a certain functionality.

## Strong-typed Command Buffer shims

```rust
pub struct ComputeCommandBuffer<'a, B: Backend>(pub(crate) &'a mut B::RawCommandBuffer)
where B::RawCommandBuffer: 'a;

impl<'a, B: Backend> Capability for ComputeCommandBuffer<'a, B> {
    type Capability = Compute;
}

impl<'a, B: Backend> ComputeCommandBuffer<'a, B> {
    /// Finish recording commands to the command buffers.
    ///
    /// The command buffer will be consumed and can't be modified further.
    /// The command pool must be reset to able to re-record commands.
    pub fn finish(mut self) -> Submit<B, Compute> {
        Submit::new(self.0.finish())
    }

    // Implement compute and transfer operations here, forwarding to raw command buffer calls.
}
```
These shims are similar to Metals Encoders but in contrast to them the user doesn't/can't switch between different encoders. This means that e.g `ComputeCommandBuffer` implements methods for compute and transfer operations.

## RawCommandBuffer

The `RawCommandBuffer` denotes _primary_ command buffers which can be submitted to queues.
It abstracts the command buffers of the different APIs, where we focus on the next-gen APIs. For GL the command buffers are implemented by storing command operations manually inside a Vec as currently implemented.

Incomplete interface:
```rust
pub trait RawCommandBuffer<B> {
[...]

PR #1419, opened 2017-08-07
[ll] Initial work for command buffer rework [37/..]

msiglreith:
Following the approach presented by #1416 

* Rework `Buffer` trait to match the new `RawCommandBuffer` interface for primary command buffers
* Implement strong-typed shims for command buffer types.
* Add  `RenderPassInlineEncoder` as additional shim for renderpass handling
* Start implementation of vulkan/dx12/(gl) command buffer reworks based on the `corell` backends
* Merge `Resources` into `Backend` as it's easier to work with..
* Add a few new resource types (renderpass, framebuffer, descriptorset, ...) and corresponding device creation methods but mostly to get the command buffer interface correct atm.

Current implementation issues:
* dx12: Renderpasses need to be emulated, requiring to cache the current renderpass information in the command buffer. In the future we could optimize memory storage by storing these per command pool as we would only need one cache/pool instead of one cache/command buffer.

* dx12: Indirect calls are more advanced compared to vulkan, therefore we need to create a few `CommandSignatures` upfront on initialization and share them with each command buffer, but shouldn't be a blocker.

cc #1102 

Edit: separate branch similar to the old `ll` branch. It doesn't build yet, but to keep you updated from time to time..


kvark:
The merging of `Resources` and `Backend` seems to do well.
All good stuff!

kvark:
> Edit: separate branch similar to the old ll branch. It doesn't build yet, but to keep you updated from time to time..

Hmm, this PR is targeted at `master` though. Is this a mistake?

msiglreith:
> Hmm, this PR is targeted at master though. Is this a mistake?

Oh, I need to create an empty branch first and push it..

msiglreith:
Updated the `ll` branch to be even with `master` and retargeted the branch to `ll`.

PR #1423, opened 2017-08-10
Fix travis linux build

msiglreith:
Add missing glfw dependencies for the newer trusty images.

Same as #1422 for 0.16 branch


msiglreith:
I assume it's safe to merge this now, given the other was merged/worked. So we can unblock the other pending PRs

bors r+

PR #1425, opened 2017-08-11
[ll] Implement clear, copy and resolve commands for vulkan [38/..]

msiglreith:
* Adjust the core command buffer API for clear/copy/resolve commands to matching the vulkan API
* Implement `clear_depth_stencil`, `copy_*` and `resolve_image` for vulkan
* Smaller cleanup (formating, merge barriers into an enum, ...)

cc #1416

### DIFF
--- a/src/backend/vulkan/src/command.rs
+++ b/src/backend/vulkan/src/command.rs
@@ -87,9 +87,9 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
         let contents = map_subpass_contents(first_subpass);
         unsafe {
             self.device.0.cmd_begin_render_pass(
-                self.raw, // commandBuffer
-                &info,    // pRenderPassBegin
-                contents, // contents
+                self.raw,
+                &info,
+                contents,
             );
         }
     }
@@ -211,12 +211,12 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
             .collect();
         unsafe {
             self.device.0.cmd_resolve_image(
-                self.raw,                           // commandBuffer
-                src.raw,                            // srcImage
-                data::map_image_layout(src_layout), // srcImageLayout
-                dst.raw,                            // dstImage
-                data::map_image_layout(dst_layout), // dstImageLayout
-                &regions,                           // pRegions
+                self.raw,
+                src.raw,
+                data::map_image_layout(src_layout),
+                dst.raw,
+                data::map_image_layout(dst_layout),
+                &regions,
             );
         }
     }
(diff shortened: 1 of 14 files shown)


src/core/src/memory.rs
```diff
@@ -178,12 +178,11 @@ bitflags!(
...
-///
-#[derive(Clone, Copy, Debug)]
-pub struct BufferBarrier;
-
-///
-#[derive(Clone, Copy, Debug)]
-pub struct ImageBarrier;
+pub enum Barrier {
+    ///
+    MemoryBarrier,
```
kvark: I don't think we need suffices here, just `Memory`, `Image`, and `Buffer` should do

src/backend/vulkan/src/command.rs
```diff
@@ -337,20 +441,27 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
-            }).collect();
+            })
+            .collect();
 
         unsafe {
             self.device.0.cmd_copy_buffer_to_image(
-                self.raw, // commandBuffer
-                src.raw, // srcBuffer
-                dst.raw, // dstImage
+                self.raw,                       // commandBuffer
```
kvark: I don't think we need the comments here, it's mostly self-describing.
For non-obvious things, we could add local variables, e.g. `let command_buffer = self.raw;`

src/backend/vulkan/src/command.rs
```diff
@@ -283,47 +363,71 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
+            self.device
+                .0
+                .cmd_dispatch_indirect(self.raw, buffer.raw, offset)
         }
     }
 
-    /*
-    fn update_buffer(&mut self, buffer: &n::Buffer, data: &[u8], offset: usize) {
+    fn copy_buffer(&mut self, src: &n::Buffer, dst: &n::Buffer, regions: &[BufferCopy]) {
+        let regions: SmallVec<[vk::BufferCopy; 16]> = regions
```
kvark: collecting into a smallvec - neat!

src/backend/vulkan/src/command.rs
```diff
@@ -248,25 +316,37 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
+        sets: &[&n::DescriptorSet],
+    ) {
+        let sets: SmallVec<[vk::DescriptorSet; 16]> = sets.iter().map(|set| set.raw).collect();
+
+        unsafe {
+            self.device.0.cmd_bind_descriptor_sets(
+                self.raw,                        // commandBuffer
+                vk::PipelineBindPoint::Graphics, // pipelineBindPoint
+                layout.raw,                      // layout
+                first_set as u32,                // firstSet
```
kvark: similarly here, let's not have comments, and instead have local vars where needed:
`let dynamic_offsets = []; //TODO`

src/backend/vulkan/src/command.rs
```diff
@@ -143,73 +141,146 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
+                        &region.dst_subresource,
+                    ),
+                    dst_offset: data::map_offset(region.dst_offset),
+                    extent: data::map_extent(region.extent),
+                }
+            })
+            .collect();
+        unsafe {
+            self.device.0.cmd_resolve_image(
+                self.raw,                           // commandBuffer
```
kvark: similarly, let's remove comments

msiglreith:
Fair enough! Updated.

kvark:
Let me know if you need help porting this to any of the backends!

Issue #1376, opened 2017-07-17
[ll] Descriptor pools maping


kvark:
`vkDescriptorPool` can be mapped to an indirect argument buffer of Metal. However, D3D12 model is much different:
  - `ID3D12DescriptorHeap` can only store CBV/SRV/UAV, or samplers, or RTV, or DSV. Mixing is not allowed.
  - only one heap of each type can be active (bound to `ID3D12CommandList`) at any given time.

Current approach:
  - have both heaps and pools, allocated on those heaps
  - pool on D3D12 is just a slice of the heap
  - a heap on Vulkan is nothing

This works, but it's an unfortunate derail from Vulkan API. It would be great to not expose heaps directly and map from Vulkan.

msiglreith:
On the topic of d3d12's descriptor heaps (for reference):
https://github.com/gpuweb/gpuweb/issues/19#issuecomment-316569697

kvark:
The current situation is worrisome. Basically, there is hardware that can only do what D3D12 model allows, but not what Vulkan's does. This means possible re-allocation of descriptors between heaps **at draw time** by the vulkan driver... :(

I'm still waiting for Vulkan group to respond, but the worst thing is that it's not even clear how to nicely encode this capability difference in flags. If we just have a flag saying that "one can bind descriptor sets from different pools" (and only expose Vulkan style `DescriptorPool`), then it doesn't allow the case where samplers are provided from a different heap... And trying to make the rules more complicated means greater headache on the ISV side when using the library.

If the situation is not resolved by then, I'll try to engage both parties at SIGGRAPH.

kvark:
For the sake of Vk Portability, we should probably just support Vk descriptor pools 1:1. D3D12 would then keep a heap for samplers/descriptors, a transfer command queue, and re-allocate the heap when new pools are requested.

kvark:
Details of the proposal:
  - when `allocate_descriptors` is called, and we are out of space, we:
    1.  allocate the new heap, increasing the assigned "generation" count. This generation gets carried by all descriptor sets referring to this heap
    2. issue a copy operation to get all the old descriptors in the new heap, remember the semaphore/fence.
    3. don't remove the old heap. Only do this once no command buffers contain references to it.
- when a command buffer starts recording:
    1. take the fresh-most heap and binds it with `BindDescriptorHeaps`
    2. remember the generation ID as the current one
- when `bind_descriptor_sets` is called
[...]

PR #1430, opened 2017-08-16
[ll] Groundwork for descriptor sets [39/..]

msiglreith:
Initial work for support of descriptor sets, which are required as a part of the command buffer rework.
Descriptor sets follow the model from vulkan, including associated concepts like descriptor pools and layouts. From D3D12 we borrow the concept of descriptor heaps #1376 

## Overview
* dx12: Implement a few more commands, mainly copy and clears. The implementations are based upon our dx12ll backend but slightly modified (i.e support multiple layer clearing, fix subresource calculation and subregion support)
* Refactor `pso` module: Now split into multiple modules for different pipeline aspects. Graphics pipeline description follows now vulkan's approach more closely.
* Add creation methods for the new resources for `Device` and `handle::Manager`

## Future work
* `DescriptorPool` will be become a trait and removed from the pure resources as it's similar to `CommandPool`
* Focusing on getting the GL backend running as we have a first draft API now. Aiming on getting a running MVP with the GL backend to build upon.

cc #1102


src/core/src/handle.rs
```diff
@@ -382,7 +461,12 @@ impl<Bd: Backend> Producer<Bd> for Manager<Bd> {
         J: Fn(&mut T, &mut Bd::Sampler),
         K: Fn(&mut T, &mut Mutex<Bd::Fence>),
         L: Fn(&mut T, &mut Mutex<Bd::Semaphore>),
-    >(&mut self, param: &mut T, fa: A, fe: E, ff: F, fg: G, fh: H, fi: I, fj: J, fk: K, fl: L) {
+        M: Fn(&mut T, &mut Bd::RenderPass),
```
msiglreith: duplicated: fixed on local repository

src/core/src/handle.rs
```diff
@@ -267,6 +279,13 @@ pub struct Manager<B: Backend> {
     samplers:      Vec<Arc<B::Sampler>>,
     fences:        Vec<Arc<Mutex<B::Fence>>>,
     semaphores:    Vec<Arc<Mutex<B::Semaphore>>>,
+    graphics_pipelines:     Vec<Arc<B::GraphicsPipeline>>,
```
kvark: ugh, unfortunate to have this code added only to be removed later on

src/backend/dx12/src/command.rs
```diff
@@ -95,35 +98,84 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
+        regions: &[ImageResolve],
+    ) {
+        for region in regions {
+            for l in 0..region.num_layers as UINT {
+                unsafe {
+                    self.raw.ResolveSubresource(
+                        src.resource,
+                        src.calc_subresource(region.src_subresource.0 as UINT, l + region.src_subresource.1 as UINT),
+                        dst.resource,
+                        dst.calc_subresource(region.src_subresource.0 as UINT, l + region.src_subresource.1 as UINT),
```
msiglreith: nicely spotted!

PR #1431, opened 2017-08-17
[ll] Start reworking GL command buffers [40/..]

msiglreith:
## Overview
* Remove most of the older command buffer and pso setup code, and align the backend implementation more with the other backends (moving resource defintions into native)
* Implement draw and dispatch commands including more safety checks for compatibility
* Add new capability flags regarding draw command support to better match functionality of different GL version
* Start adding queue side caching, which allows us to reduce the GL calls for command buffer resetting to a minimum in the future. The old command buffer caching will still be preserved as it reduces memory allocation and is suitable for possible command buffer reuse in the future.

## Reviewing notes
* The later 2 commits should be reviewed independently of the first one (adfdbcb) as it's more easily to see the changes.
* `tex.rs` got renamed to `texture.rs` but I forgot to read in the 'cleanup' commit


kvark:
Could you squash the first 2 commits together so that git recognizes tex -> texture move?
Otherwise looks decent!

msiglreith:
Thanks, squased those two.
I'm currently thinking about moving cmd buffer caching more over to queue caching as it's much more powerful. The old caching seems more useful for as general optimization technique if the users emit redundant commands in a _single_ command buffer.

PR #1432, opened 2017-08-17
[ll] Further work on the gl backend [41/..]

msiglreith:
* gl: Implement index buffer binding command
* gl: Create a `GeneralQueue` if we have compute support, else fall back to `GraphicsQueue`
* core: Split `Capabilities` into `Features` (specify which aspects of the our API are supported) and `Limits` (numerical limits like max_texture_size, max_viewports, ..)

### DIFF
--- a/src/backend/gl/src/info.rs
+++ b/src/backend/gl/src/info.rs
@@ -15,7 +15,7 @@
 use std::collections::HashSet;
 use std::{ffi, fmt, mem, str};
 use gl;
-use core::{Capabilities, Features, IndexCount, Limits, VertexCount};
+use core::{Features, IndexCount, Limits, VertexCount};
 
 /// A version number for a specific component of an OpenGL implementation
 #[derive(Copy, Clone, Eq, Ord, PartialEq, PartialOrd)]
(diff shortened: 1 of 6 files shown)


src/backend/gl/src/info.rs
```diff
@@ -256,41 +256,46 @@ pub fn get(gl: &gl::Gl) -> (Info, Capabilities, PrivateCaps) {
...
+        unordered_access_view:              info.is_supported(&[Core(4,0)]), // TODO: extension
+        separate_blending_slots:            info.is_supported(&[Core(4,0),
+                                                                Es  (3,0),
+                                                                Ext ("GL_ARB_draw_buffers_blend")]),
+        copy_buffer:                        info.is_supported(&[Core(3,1),
+                                                                Es  (3,0),
+                                                                Ext ("GL_ARB_copy_buffer"),
+                                                                Ext ("GL_NV_copy_buffer")]),
+    };
+    let caps = Capabilities {
```
kvark: not sure if `Capabilities` makes any sense now. Might consider just storing features and limits directly

kvark:
Do you think we can remove the `Capabilities` completely?

msiglreith:
I'm fine with removing it, give me a few minutes 😺

kvark:
Awesome, thank you!

PR #1433, opened 2017-08-18
[ll] gl: Implement setting viewports and scissors [42/..]

msiglreith:
Implement `set_viewports` and `set_scissors`:
* Our old implementation only allowed one viewport and scissor. To minimize heap allocations we store everything into the `DataBuffer` in the hope to reuse the memory in upcoming frames (we could actually share the internal buffers across all command buffers of a pool!).
* Add validation checks on the command buffer side
* Enabling and Disabling of scissors test will done on when binding the pipeline. Scissor disabling seems to only make sense when rasterization is disabled as vulkan requires at least one viewport/scissor.

### Open Issues
* Disabling (zeroing) all non-active viewports and scissors requires a bit more logic. This is required to be consitent with D3D12.

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -22,14 +22,16 @@
 extern crate log;
 extern crate gfx_gl as gl;
 extern crate gfx_core as core;
+extern crate smallvec;
 
 use std::cell::RefCell;
 use std::rc::Rc;
-use core::{self as c, handle, format, texture as t, memory, command as com, buffer};
+use core::{self as c, handle, format, target, texture as t, memory, command as com, buffer, Viewport};
 use core::QueueType;
 use core::target::{Layer, Level};
 use command::{Command, DataBuffer};
 use device::MappingKind;
+use smallvec::SmallVec;
 
 pub use self::device::Device;
 pub use self::info::{Info, PlatformName, Version};
@@ -342,6 +344,10 @@ struct State {
     // Currently bound index/element buffer.
     // None denotes that we don't know what is currently bound.
     index_buffer: Option<gl::types::GLuint>,
+    // Currently set viewports.
+    num_viewports: usize,
+    // Currently set scissor rects.
+    num_scissors: usize,
 }
 
 impl State {
@@ -351,6 +357,8 @@ impl State {
         State {
             vao: false,
             index_buffer: None,
+            num_viewports: 0,
+            num_scissors: 0,
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/backend/gl/src/lib.rs
```diff
@@ -486,6 +497,36 @@ impl CommandQueue {
...
             }
         }
+
+        // Reset viewports
+        if self.state.num_viewports == 1 {
+            unsafe { gl.Viewport(0, 0, 0, 0) };
+            unsafe { gl.DepthRange(0.0, 1.0) };
+        } else if self.state.num_viewports > 1 {
+            // 16 viewports is a common limit set in drivers.
+            let viewports: SmallVec<[[f32; 4]; 16]> =
```
kvark: could just have a fixed size `[[0.0, 0.0, 0.0, 0.0]; 16]` instead of smallvec

src/backend/gl/src/lib.rs
```diff
@@ -486,6 +497,36 @@ impl CommandQueue {
...
+            unsafe { gl.ViewportArrayv(0, viewports.len() as i32, viewports.as_ptr() as *const _)};
+            unsafe { gl.DepthRangeArrayv(0, depth_ranges.len() as i32, depth_ranges.as_ptr() as *const _)};
+        }
+
+        // Reset scissors
+        if self.state.num_scissors == 1 {
+            unsafe { gl.Scissor(0, 0, 0, 0) };
+        } else if self.state.num_scissors > 1 {
+            // 16 viewports is a common limit set in drivers.
+            let scissors: SmallVec<[[i32; 4]; 16]> =
```
kvark: same here

src/core/src/command/raw.rs
```diff
@@ -54,10 +54,47 @@ pub trait RawCommandBuffer<B: Backend> {
...
+    /// this call, will be disabled.
+    ///
+    /// Ensure that the number of set viewports at draw time
+    /// is equal (or higher) to the number specified in the
+    /// bound pipeline.
+    ///
+    /// # Errors
+    ///
+    /// This function does not return an error.
+    /// Invalid usage of this function will result
```
kvark: somewhat surprising wrap length of the line?

kvark:
thank you!
r=me regardless if you decide to address the nits

src/backend/gl/src/lib.rs
msiglreith: 16 is not the upper bound, even though from gpu info database, it seems to be the one reported by most drivers atm. So in the theory/future we could have more than 16.

PR #1434, opened 2017-08-21
[ll] implement `Default` for `WrapMmode`

burtonageo:
From the doc comments:

```rust
pub enum WrapMode {
    /// Tile the texture. That is, sample the coordinate modulo `1.0`. This is
    /// the default.
    Tile,
```

### DIFF
--- a/src/corell/src/image.rs
+++ b/src/corell/src/image.rs
@@ -217,6 +217,12 @@ pub enum WrapMode {
     Border,
 }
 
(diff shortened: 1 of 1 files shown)


kvark:
Thank you for the PR!
I think we should rather remove this deprecated sentence from the comment than implement default for the wrap mode.

kvark:
Closing due to inactivity, please reopen if needed

PR #1436, opened 2017-08-22
[ll] Extend pso functionality and work on gl backend [43/..]

msiglreith:
* core: Add further configuration possibilities for graphics pipelines (still not finished so far: multisampling, depth-stencil state and dynamic states)
* core: Split the setting of reference values (blend constants and stencil ref) into two commands
* glutin: Fix window backend after the merge of `Resources` and `Backend`
* gl: Implement color clearing based on the old implementation

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -489,6 +489,9 @@ impl CommandQueue {
             self.state.vao = true
         }
 
+        // Reset indirect draw buffer
+        unsafe { gl.BindBuffer(gl::DRAW_INDIRECT_BUFFER, 0) };
+
         // Unbind index buffers
         match self.state.index_buffer {
             Some(0) => (), // Nothing to do
@@ -852,9 +855,6 @@ impl CommandQueue {
                 }else if false {
                     error!("Separate blending slots are not supported");
                 }
-            },
-            Command::SetBlendColor(color) => {
-                state::set_blend_color(&self.share.context, color);
             },
             Command::SetPatches(num) => {
                 let gl = &self.share.context;
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -39,7 +39,7 @@ pub use self::info::{Info, PlatformName, Version};
 mod command;
 mod device;
 mod info;
-mod native;
+pub mod native;
 mod pool;
 mod state;
 mod texture;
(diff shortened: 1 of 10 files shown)


src/backend/gl/src/lib.rs
```diff
@@ -489,6 +473,9 @@ impl CommandQueue {
             self.state.vao = true
         }
 
+        // Reset indirect draw buffer
+        unsafe { gl.BindBuffer(gl::DRAW_INDIRECT_BUFFER, 0) };
```
kvark: that would surely confuse some older driver, we need to gate this by a device feature

src/core/src/pso/graphics.rs
```diff
@@ -15,10 +15,19 @@
...
-use Primitive;
 use super::EntryPoint;
-use super::input_assembler::{AttributeDesc, VertexBufferDesc};
-use super::output_merger::{ColorTargetDesc, DepthStencilDesc};
+use super::input_assembler::{AttributeDesc, InputAssemblerDesc, VertexBufferDesc};
+use super::output_merger::{ColorInfo, DepthStencilDesc};
+
+// Vulkan:
+//  - SpecializationInfo not provided per shader
+//  - TODO: infer rasterization discard from shaders?
```
kvark: we absolutely could, yeah, but Vk portability would be a bit easier if we had this flag explicitly exposed

src/core/src/pso/graphics.rs
```diff
@@ -37,35 +46,104 @@ pub struct GraphicsShaderSet {
...
-            shader_entries: shader_entries,
-            color_targets: Vec::new(),
-            vertex_buffers: Vec::new(),
-            attributes: Vec::new(),
-        }
-    }
+///
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature="serialize", derive(Serialize, Deserialize))]
+pub struct DepthBias {
```
kvark: I guess that makes us completely detach from `draw_state`?

msiglreith: Yes, more and more /:

src/core/src/pso/graphics.rs
msiglreith: okay

src/backend/gl/src/lib.rs
msiglreith: ah right, getting all them correct is hard! Thanks for reviewing!

PR #1439, opened 2017-08-22
[ll] small Vulkan update

kvark:


src/backend/vulkan/src/command.rs
```diff
@@ -206,7 +206,7 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
                         region.src_subresource,
                         region.num_layers,
                     ),
-                    src_offset: offset,
+                    src_offset: offset.clone(),
```
msiglreith: Ugh, `vkOffset3d` not implementing `Copy`.
We should address this in `ash` in the future.

src/backend/vulkan/src/lib.rs
```diff
@@ -348,6 +348,11 @@ impl core::Adapter<Backend> for Adapter {
 
 #[doc(hidden)]
 pub struct RawDevice(pub ash::Device<V1_0>);
+impl fmt::Debug for RawDevice {
```
msiglreith: Does the compiler require it?

src/backend/vulkan/src/command.rs
kvark: I don't mind `Clone` either. `Copy` would work better if we need arrays of that stuff

src/backend/vulkan/src/lib.rs
kvark: It does since I included `Arc<RawDevice>` in `DescriptorPool`

PR #1437, opened 2017-08-22
[WIP] Settle on `Swapchain` naming convention.

Binero:

PR #1441, opened 2017-08-23
Fix appveyor for dx12 builds

msiglreith:
Exclude all the necessary crates for building with appveyor.
`gfx_window_dxgi` is also excluded as it depends on dx12, which we need to feature-gate in the a followup PR but this would block #1437


kvark:
thank you!
bors r+

PR #1442, opened 2017-08-23
[ll] new core example

kvark:
The example doesn't run yet, since the command buffers have not been ported yet from LL.

Sorry @Bastacyclop , I removed the advanced mapping tracker. We'll need to resurrect it at the `gfx_render` level, but I can't afford jumping on it right now - need the core to be finished first.


Bastacyclop:
No problem. Are you going to keep the persistent mapping interface (with temporary mapping fallback behind the scenes) for `core` ?

kvark:
@Bastacyclop not sure yet. We want to be explicit and lightweight, and tracking those persistent mappings appears to involve a bit of logic.
It's tempting to drop support for platforms that don't have persistent mapping...

Bastacyclop:
Well wouldn't that drop support for D3D11 ?

Bastacyclop:
Otherwise there could be a way to query whether the platform uses persistent or temporary mapping and only `render` would abstract over that.

PR #1443, opened 2017-08-24
[ll] Running Vulkan example

kvark:
~~Don't merge - this is for demonstration purposes only ;)~~
This is ready for merging: core is looking good, and Vulkan runs a basic example.


kvark:
@msiglreith I think it's ready for merging. The quad draws at last (Vulkan only).

I had to go through quite a lot of plumbing on the way here, which was complicated by the growing divergence between coreLL and core. Fortunately, I faced no really bad/frustrating issues: it went rather smooth thanks to the solid foundation laid out by @msiglreith and the sheer power of Rust.

Having 3 branches (pre-ll, master, ll) doesn't help much now. We need to finish the transition in `ll` branch, merge into master, and remove all the old xxxLL crates. Speaking of finishing, the roughly following things are TODO:
  1. Backends: Metal, D3D12, GL, potentially D3D11 in some limited way
  2. fill the gaps (and improve) the gfx_core API, e.g. have a nicer typed render pass encoding (potentially evolved from coreLL).
  3. `gfx_render` is hugely important and will require an effort

It draws, at last:
gfx-core-quad

src/core/src/command/graphics.rs
```diff
@@ -12,142 +12,165 @@
...
+use {Backend, InstanceCount, VertexCount, Viewport};
 use buffer::IndexBufferView;
-use queue::capability::{Capability, Graphics};
-use super::{BufferCopy, BufferImageCopy, ClearColor, CommandBufferShim, ImageCopy, RawCommandBuffer, Submit};
+use image::ImageLayout;
+use queue::capability::{Graphics, Supports};
+use super::{CommandBuffer, RawCommandBuffer, RenderPassInlineEncoder};
+
+
+/// A universal clear color supporting integet formats
```
msiglreith: small typo: `integet`

src/backend/vulkan/src/lib.rs
```diff
@@ -394,18 +419,54 @@ impl CommandQueue {
     }
 }
 
-impl core::CommandQueue<Backend> for CommandQueue {
-    unsafe fn submit_raw<'a, I>(
-        &mut self,
-        submit_infos: I,
+impl core::RawCommandQueue<Backend> for CommandQueue {
+    unsafe fn submit_raw(&mut self,
+        submission: core::RawSubmission<Backend>,
```
msiglreith: Why only single submission per submit?

src/backend/empty/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+description = "Empty backend for gfx-rs"
+publish = false
+workspace = "../../.."
+
+[lib]
+name = "empty"
+
+[dependencies]
+gfx_core = { path = "../../core", version = "0.7" }
```
msiglreith: I'm not totally sure if moving dummy out works as it's required for the docs of `core`.
Have you tested it?

src/backend/vulkan/src/lib.rs
```diff
@@ -318,20 +346,17 @@ impl core::Adapter<Backend> for Adapter {
...
         let queues = queue_infos.iter().flat_map(|info| {
             (0..info.queue_count).map(|id| {
-                let queue = unsafe {
+                let queue_raw = unsafe {
                     device.raw.0.get_device_queue(info.queue_family_index, id)
                 };
-                unimplemented!()
-                /*
-                // TODO:
+                let queue = CommandQueue {
```
msiglreith: I assume we still need to differ between queue types here?

msiglreith:
The capability based implementations look really neat!
I tried it in earlier iterations multiple times (before `Support`) but always hit at some compiler complaints, so I'm really excited to see it no, awesome!  ❤️

src/backend/vulkan/src/lib.rs
kvark: I'm on it: https://github.com/gfx-rs/gfx/pull/1449/commits/4fab52897c6d65107c745bf20532c2896142ed5e

src/backend/vulkan/src/lib.rs
kvark: Looking at it again, I think we should get it back to multiple submissions per call, just to match Vulkan if nothing else...

src/backend/empty/Cargo.toml
msiglreith: Would be nice to keep it the current way to keep `core` small and be a bit more consitent.
I'm currently trying out if this is possible.

PR #1445, opened 2017-08-24
Remove all Apache license headers

### LINKED ISSUE #1428: Remove all license headers from source files
They only clutter the source, make it more difficult to navigate in plain text.
According to Apache license [rules](https://www.apache.org/legal/src-headers.html), we'd only need the headers if we were to transfer the project under ASF, which is not on the roadmap ;)

xtian:
Fixes #1428

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,3 @@
-<!--
-    Copyright 2014 The Gfx-rs Developers.
-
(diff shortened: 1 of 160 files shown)


kvark:
Neat, thank you!
bors r+

PR #1447, opened 2017-08-26
Minor documentation improvements.

Binero:
Added some documentation to help new users figure out how the various structures are typically obtained, as well as an example (& doc-test) for `gfx_window_glutin`.

It's a very minor improvement, but every little bit helps! :D

### DIFF
--- a/src/core/src/adapter.rs
+++ b/src/core/src/adapter.rs
@@ -5,6 +5,9 @@
 use {Backend, Gpu, QueueType};
 
 /// Represents a physical or virtual device, which is capable of running the backend.
+///
+/// The `Adapter` is typically obtained from objects implementing `gfx::WindowExt` or
+/// `gfx::Headless`. How these types are created is backend-specific.
 pub trait Adapter<B: Backend>: Sized {
     /// Create a new logical gpu with the specified queues.
     ///
--- a/src/window/glutin/src/lib.rs
+++ b/src/window/glutin/src/lib.rs
@@ -1,3 +1,27 @@
+//! Window creation using glutin for gfx.
+//!
+//! # Examples
+//! The following code creates a `gfx::WindowExt` using glutin.
+//!
+//! ```no_run
+//! extern crate glutin;
+//! extern crate gfx_device_gl;
+//! extern crate gfx_window_glutin;
+//! 
+//! fn main() {
+//!     use gfx_window_glutin::Window;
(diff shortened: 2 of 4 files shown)


src/window/glutin/src/lib.rs
```diff
@@ -1,3 +1,27 @@
+//! Window creation using glutin for gfx.
+//!
+//! # Examples
+//! The following code creates a `gfx::WindowExt` using glutin.
```
kvark: hmm, it doens't make sense to duplicate the same code here and in the glutin window example. Let's make this one fully compilable and remove the latter?

src/core/src/adapter.rs
```diff
@@ -5,6 +5,8 @@
 use {Backend, Gpu, QueueType};
 
 /// Represents a physical or virtual device, which is capable of running the backend.
+///
+/// The `Adapter` is obtained from objects implementing `gfx::WindowExt`.
```
kvark: ... or by backend-specific functions

src/window/glutin/src/lib.rs
Binero: Sorry, what other glutin window example is that?

kvark: Ah, pardon me, I was thinking about GLFW/SDL

kvark:
Thank you! Need to fix CI, and we'll be good.

src/core/src/adapter.rs
```diff
@@ -5,6 +5,9 @@
 use {Backend, Gpu, QueueType};
 
 /// Represents a physical or virtual device, which is capable of running the backend.
+///
+/// The `Adapter` is typically obtained from objects implementing `gfx::WindowExt`. Different
```
msiglreith: We also have [`Headless`](https://github.com/gfx-rs/gfx/blob/master/src/core/src/lib.rs#L346-L353) now. The difference between `WindowExt` and `Headless` is the additional `Surface` as we basically have 2 entrypoints for the library `Adapter` for the physical GPU side and `Surface` for the physical display side. Both open/create logical equivalents on the library side (`Gpu` and `Swapchain`) for working with the API.

Binero:
@msiglreith Fixed the adapter docs.
@kvark Squashed all commits into one.

kvark:
Thank you!
bors r+

kvark:
@Binero that new test needs to be marked as no-run

torkleyy:
One could also feature flag the main function (requiring feature `headless`) and make it a no-op otherwise.

PR #1449, opened 2017-08-28
[ll] renamed backend raw associated types

kvark:


src/backend/vulkan/src/lib.rs
```diff
@@ -343,29 +369,13 @@ impl core::Adapter<Backend> for Adapter {
...
-        }).collect();
-
+        let device_arc = device.raw.clone();
         core::Gpu {
             device,
-            general_queues: queues,
-            graphics_queues: Vec::new(),
-            compute_queues: Vec::new(),
-            transfer_queues: Vec::new(),
+            general_queues: collect_queues(queue_descs, &device_arc, QueueType::General),
```
kvark: @msiglreith 
it was also tempting to have queue capability trait with a method returning `QueueType` (aka our `Formatted::get_format()`), but I'd rather not make this complication if this initialization code is the only use case :)

PR #1451, opened 2017-08-28
[pre][GL] removal of Factory::frame_handles

kvark:
Port of #1397 to 0.16
required us to bump gfx_core as well, unfortunately


kvark:
Ffs, what's holding this one?
bors r+

PR #1457, opened 2017-08-29
[ll] core: Refactor command pools

msiglreith:
Refactor the current command pool implementation and remove `SubmitInfo`.
Motivated by providing a lower level design which is more vulkan flavored.

We basically follow the (internal) paradigm of providing `Raw` objects (close to the metal, zero overhead, C-ish) and having safe wrappers on top of them which enforce some constraints at compile time and come with minimal overhead.

### Lower layer
* The `RawCommandBuffer` isn't supposed to store the command buffers internally and moves that task onto the library user.
* `SubmitInfo` was a safe object to pass over thread boundaries to prevent users from recording multiple command buffers at the same time. We now move towards sending (+ cloning) `RawCommandBuffers` which requires the users to take care of not recording multiple buffers from one pool at the same time.

### Higher layer
* `CommandPool` fullfills the part of a higher wrapper around `RawCommandPool` to ensure capabilities for the underlying buffers and prevent multiple buffer encoding. Stores command buffers internally in a `Vec` queue.
* `Submit` internally uses now `RawCommandBuffers` instead of `SubmitInfo`.

## TODO
- [x] ~~Ensure that `CommandBuffer` is `!Send` (how?)~~ edit: Can be `Send`..
- [x] Update the Vulkan backend

### DIFF
--- a/src/core/src/command/mod.rs
+++ b/src/core/src/command/mod.rs
@@ -58,10 +58,13 @@ impl<'a, B: Backend, C> CommandBuffer<'a, B, C> {
     ///
     /// The command buffer will be consumed and can't be modified further.
     /// The command pool must be reset to able to re-record commands.
-    pub fn finish(mut self) -> Submit<B, C> {
+    pub fn finish(self) -> Submit<B, C> {
+        unsafe { Submit::new(self) }
+    }
+}
+
+impl<'a, B: Backend, C> Drop for CommandBuffer<'a, B, C> {
+    fn drop(&mut self) {
         self.raw.finish();
-        unsafe {
-            Submit::new(self)
-        }
     }
 }
--- a/src/core/src/command/mod.rs
+++ b/src/core/src/command/mod.rs
@@ -15,7 +15,6 @@
 //!
 
 use Backend;
-use pool::{RawCommandPool};
 use std::marker::PhantomData;
 
 mod compute;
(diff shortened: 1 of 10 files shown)


src/core/src/command/mod.rs
```diff
@@ -29,42 +28,28 @@ pub use self::raw::RawCommandBuffer;
...
 }
 
 /// Command buffer with compute, graphics and transfer functionality.
-pub struct CommandBuffer<'a, B: 'a + Backend, C> {
-    pub(crate) raw: B::CommandBuffer,
-    pool: &'a mut B::CommandPool,
-    _capability: PhantomData<C>,
+pub struct CommandBuffer<'a, B: Backend, C> {
+    pub(crate) raw: &'a mut B::CommandBuffer,
+    _capability: PhantomData<*const C>,
```
kvark: why `*const C`?

src/core/src/queue/submission.rs
```diff
@@ -25,7 +25,7 @@ use smallvec::SmallVec;
 /// Raw submission information for a command queue.
 pub struct RawSubmission<'a, B: Backend + 'a> {
     /// Command buffers to submit.
-    pub cmd_buffers: &'a [B::SubmitInfo],
+    pub cmd_buffers: &'a [B::CommandBuffer],
```
kvark: this certainly looks better now

src/core/src/command/mod.rs
```diff
@@ -74,10 +59,9 @@ impl<'a, B: Backend, C> CommandBuffer<'a, B, C> {
     /// The command buffer will be consumed and can't be modified further.
     /// The command pool must be reset to able to re-record commands.
     pub fn finish(mut self) -> Submit<B, C> {
-        let submit = self.raw.finish();
+        self.raw.finish();
```
kvark: we might also want to call `self.raw.finish()` in the destructor

src/core/src/command/mod.rs
msiglreith: leftover from my attempt to enforce `!Send`, fixed.

src/core/src/command/mod.rs
msiglreith: Would you mind elaborating?

kvark:
Thank you!

PR #1458, opened 2017-08-30
Mipmap generation for GL

snuk182:


src/backend/gl/src/tex.rs
```diff
@@ -220,10 +220,41 @@ pub fn format_to_glfull(format: NewFormat) -> Result<GLenum, ()> {
     })
 }
 
+fn determine_mipmap_level(desc: &t::Info, mip: Option<t::Mipmap>) -> t::Level {
+    use std::cmp::{min, max};
+    
+    match mip {
```
kvark: please use the existing `get_num_levels` instead: https://github.com/gfx-rs/gfx/blob/7f30638ff6707bf473fb08e3ce3b2b87e92ebacd/src/core/src/texture.rs#L289

src/backend/gl/src/tex.rs
```diff
@@ -220,10 +220,41 @@ pub fn format_to_glfull(format: NewFormat) -> Result<GLenum, ()> {
...
+        },
+        _ => desc.levels
+    }
+}
+
 fn set_mipmap_range(gl: &gl::Gl, target: GLenum, (base, max): (u8, u8)) { unsafe {
     gl.TexParameteri(target, gl::TEXTURE_BASE_LEVEL, base as GLint);
     gl.TexParameteri(target, gl::TEXTURE_MAX_LEVEL, max as GLint);
 }}
+fn set_mipmap_creation_parameters(gl: &gl::Gl, target: GLenum) { unsafe {
```
kvark: these are not creation parameters, it's the filtering. Also, not sure why you need this?

src/core/src/texture.rs
```diff
@@ -314,6 +314,18 @@ impl Kind {
     }
 }
 
+/// The marker for the texture initializer to generate extra space for the mipmap generation.
+#[allow(missing_docs)]
```
kvark: no need for missing docs here

src/core/src/texture.rs
```diff
@@ -314,6 +314,18 @@ impl Kind {
...
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
+#[repr(u8)]
+pub enum Mipmap {
```
kvark: I don't think you need to pass `Option<Mipmap>` around as opposed to just `Mipmap`, seems to unnecessarily complicate things

kvark:
Also, appveyor fails: https://ci.appveyor.com/project/kvark/gfx/build/1.0.1340/job/uqldnd6k9jjgg2x9#L131

src/backend/metal/src/factory.rs
```diff
@@ -483,8 +483,8 @@ impl core::Factory<Resources> for Factory {
     fn create_texture_raw
         (&mut self,
          desc: core::texture::Info,
+         _: core::texture::Mipmap,
```
kvark: I believe this parameter is meaningless when the `data_opt.is_none()`.
Therefore, we should move it behind the data `Option<>`

src/backend/gl/src/tex.rs
```diff
@@ -220,11 +220,11 @@ pub fn format_to_glfull(format: NewFormat) -> Result<GLenum, ()> {
...
 }
 
-fn determine_mipmap_level(desc: &t::Info, mip: Option<t::Mipmap>) -> t::Level {
+fn get_num_levels(desc: &t::Info, mip: t::Mipmap) -> t::Level {
     use std::cmp::{min, max};
     
     match mip {
-        Some(t::Mipmap::Allocated) => {
+        t::Mipmap::Allocated => {
             fn mip_level1(w: u16) -> u8 {
```
kvark: can we just use `desc.kind.get_num_levels()`?

src/core/src/factory.rs
```diff
@@ -331,7 +331,7 @@ pub trait Factory<R: Resources> {
         -> Result<handle::RawDepthStencilView<R>, TargetViewError>;
 
     fn create_texture<S>(&mut self, kind: texture::Kind, levels: target::Level,
-                      bind: Bind, usage: Usage, channel_hint: Option<format::ChannelType>, mipmap: Option<texture::Mipmap>)
+                      bind: Bind, usage: Usage, mipmap: texture::Mipmap, channel_hint: Option<format::ChannelType>)
```
kvark: consequently, we don't need this function to change

src/core/src/factory.rs
```diff
@@ -427,7 +427,7 @@ pub trait Factory<R: Resources> {
         self.view_texture_as_depth_stencil(tex, 0, None, texture::DepthStencilFlags::empty())
     }
 	
-	fn create_texture_immutable_u8_mipmapped<T: format::TextureFormat>(&mut self, kind: texture::Kind, data: &[&[u8]], mipmap: Option<texture::Mipmap>)
+	fn create_texture_immutable_u8_mipmapped<T: format::TextureFormat>(&mut self, kind: texture::Kind, mipmap: texture::Mipmap, data: &[&[u8]])
```
kvark: given that we are changing the API anyway in a breaking fashion, I don't think it makes sense to try to preserve the old `create_texture_immutable_u8`. Let's just add a new parameter in there.

snuk182: I'd also suggest either naming the `Mipmap` options more self-describing (`ProvidedWithBitmapData`, `AllocatedForGeneration`) or document it better to eliminate usage `Allocated` without further generation call.

kvark:
Something went off the rails: +4,899 −4,893

snuk182:
Purge this. Git ruins everything.

PR #1459, opened 2017-08-30
Mipmap generation for GL

snuk182:


src/backend/gl/src/tex.rs
```diff
@@ -404,45 +411,31 @@ pub fn make_without_storage(gl: &gl::Gl, desc: &t::Info, cty: ChannelType) ->
         Ok(t) => t,
         Err(_) => return Err(t::CreationError::Format(desc.format, Some(cty))),
     };
-
+    
+    let levels = get_num_levels(desc, mip);
     let fixed_loc = desc.bind.contains(SHADER_RESOURCE);
-    make_widout_storage_impl(gl, desc.kind, gl_format, gl_pixel_format, gl_data_type,
-                             desc.levels, fixed_loc)
+    make_without_storage_impl(gl, desc.kind, gl_format, gl_pixel_format, gl_data_type,
```
kvark: I missed this last time, but this is not how we are supposed to determine the level count.
It *always* comes from `t::Info`, unconditionally.
What your new parameter is supposed to change is how the data slices are being treated.

src/core/src/factory.rs
```diff
@@ -330,7 +331,7 @@ pub trait Factory<R: Resources> {
         -> Result<handle::RawDepthStencilView<R>, TargetViewError>;
 
     fn create_texture<S>(&mut self, kind: texture::Kind, levels: target::Level,
-                      bind: Bind, usage: Usage, channel_hint: Option<format::ChannelType>)
+                      bind: Bind, usage: Usage, channel_hint: Option<format::ChannelType>)
```
kvark: extra spaces at the end?

src/backend/gl/src/tex.rs
snuk182: Did not get it. With `Mipmap::Generated` we *change* the levels count, because it is actually changed.

kvark: You need to understand what a texture is. It's a number of 2D texel buffers, basically, allocated in VRAM. That allocation is precisely described by the `Info` struct. You do **not** need to change it.
What you are doing is saying: hey, I'm providing that texel data, but it's only for the base mipmap level, and I want the other levels to be left uninitialized. That's why I asked you to move the parameter into the data `Option`.

snuk182: Ok, so what should be instead then?

snuk182:
Next iteration comes shortly.

PR #1460, opened 2017-08-30
[ll] Working on gl backend (build, fences and samplers)

msiglreith:
* Fix GL backend build after command pool changes #1457 (pool reset still WIP)
* Implement fence and semaphore handling (ported from the old implementation)
* Implement sampler creation, adding few more features to properly reflect capabilities of the backend

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -15,14 +15,15 @@
 use std::rc::Rc;
 use std::{slice, ptr};
 
-use {gl};
+use gl;
+use gl::types::{GLenum, GLuint, GLint, GLfloat, GLsizei, GLvoid};
 use core::{self as c, device as d, image as i, pass, pso, buffer, mapping};
 use core::memory::{self, Bind, SHADER_RESOURCE, UNORDERED_ACCESS, Typed};
 use core::format::{ChannelType, Format};
 use core::target::{Layer, Level};
 
 use {Info, Backend as B, Share};
-use {conv, device, native as n, pool};
+use {conv, device, native as n, pool, state};
 
 fn access_to_map_bits(access: memory::Access) -> gl::types::GLenum {
     let mut r = 0;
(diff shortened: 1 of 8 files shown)


src/backend/gl/src/device.rs
```diff
@@ -90,8 +91,8 @@ impl d::Device<B> for Device {
         &self.share.limits
     }
 
-    fn create_heap(&mut self, _: &c::HeapType, _: d::ResourceHeapType, _: u64) -> Result<n::Heap, d::ResourceHeapError> {
-        unimplemented!()
+    fn create_heap(&mut self, heap_type: &c::HeapType, resource_type: d::ResourceHeapType, size: u64) -> Result<n::Heap, d::ResourceHeapError> {
```
kvark: how do you bear with all those "unused variable warnings"? :)

src/backend/gl/src/device.rs
```diff
@@ -121,9 +124,69 @@ impl d::Device<B> for Device {
         unimplemented!()
     }
 
-    fn create_sampler(&mut self, info: i::SamplerInfo) -> n::Sampler {
-        unimplemented!()
+    fn create_sampler(&mut self, info: i::SamplerInfo) -> n::FatSampler {
+        let object = if self.share.features.sampler_objects {
```
kvark: can we have a fast-exit path here? e.g.
```rust
if !self.share.features.sampler_objects {
  return XXX;
}
```

src/backend/gl/src/lib.rs
```diff
@@ -66,7 +65,7 @@ impl c::Backend for Backend {
     type Buffer = native::Buffer;
     type UnboundImage = device::UnboundImage;
     type Image = native::Image;
-    type Sampler = native::Sampler;
+    type Sampler = native::FatSampler;
```
kvark: we could have a sampler as an `enum` to avoid dragging the info with us
I don't think it's a big deal though

src/backend/gl/src/native.rs
```diff
@@ -75,6 +81,8 @@ pub enum Image {
 }
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
+// Additionally storing the `SamplerInfo` for older OpenGL versions, which
```
kvark: could be doc comments?

src/backend/gl/src/device.rs
msiglreith: you mean those 'friendly reminders' :>?

src/backend/gl/src/lib.rs
msiglreith: Implemented, we probably gain nothing from it (size) but it makes the difference a bit clearer.

msiglreith:
vulkan fixes for the new feature flags in the next PR (command pool/buffer reset).

PR #1462, opened 2017-09-01
Core serialization, Vulkan debug layers

kvark:


src/backend/vulkan/src/lib.rs
```diff
@@ -128,19 +175,39 @@ impl Instance {
                 p_next: ptr::null(),
                 flags: vk::InstanceCreateFlags::empty(),
                 p_application_info: &app_info,
-                enabled_layer_count: 0,
-                pp_enabled_layer_names: ptr::null(),
-                enabled_extension_count: str_pointers.len() as u32,
-                pp_enabled_extension_names: str_pointers.as_ptr(),
+                enabled_layer_count: layers.len() as _,
+                pp_enabled_layer_names: str_pointers.as_ptr(),
```
msiglreith: What will happen here if the layers are not present? Can we add checks here to filter out some layers?

src/window/vulkan/src/lib.rs
```diff
@@ -291,7 +291,8 @@ impl core::WindowExt<backend::Backend> for Window {
     type Adapter = backend::Adapter;
 
     fn get_surface_and_adapters(&mut self) -> (Surface, Vec<backend::Adapter>) {
-        let instance = Instance::create("gfx-rs demo", 0);
+        let enable_debug = true; //TEMP?
```
msiglreith: hmm, maybe enable it for debug builds only? For release it's still possible to activate layers via env variables.

PR #1463, opened 2017-09-01
[ll] gl: Add more safety for the command pool memory storage

msiglreith:
* Rework command pool memory storage to work now with a combination of `UnsafeCell` + shared atomic to denote if safe access to the underlying pool and memory is possible. It works similar to a `Mutex` but doesn't block and should be faster. We can revisit the implementation approach on a later stage when we have implemented the rest of the backend imo.

* Fix behavior of `RawCommandBuffer::begin` which should implicitly reset when individual reset flag is set on pool creation
* Emit an error when `reset` is called without the mentioned flag

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -22,16 +22,16 @@ use core::command::{BufferCopy, BufferImageCopy, ClearValue, ImageCopy, ImageRes
 use core::target::{ColorValue, Stencil};
 use {native as n, Backend};
 use pool::{self, BufferMemory};
-use std::cell::RefCell;
+use std::cell::UnsafeCell;
 use std::mem;
 use std::sync::Arc;
+use std::sync::atomic::{self, AtomicBool};
 
 // Command buffer implementation details:
 //
 // The underlying commands and data are stored inside the associated command pool.
 // See the comments for further safety requirements.
-// Each command buffer holds a (growable) slice of the buffers (`CommandBuffer`
-// and `DataBuffer`) in the pool.
+// Each command buffer holds a (growable) slice of the buffers in the pool.
 //
 // Command buffers are recorded one-after-another for each command pool.
 // Actual storage depends on the resetting behavior of the pool.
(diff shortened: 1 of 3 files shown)


src/backend/gl/src/command.rs
```diff
@@ -153,11 +153,15 @@ impl From<c::Limits> for Limits {
...
 pub struct RawCommandBuffer {
-    pub(crate) memory: Arc<RefCell<pool::OwnedBuffer>>,
+    pub(crate) memory: Arc<pool::SharedBuffer>,
     pub(crate) buf: BufferSlice,
     // Buffer id for the owning command pool.
     // Only relevant if individual resets are allowed.
     pub(crate) id: u64,
+    // Associated pool and shared buffer is accessible.
+    // Mainly for safety reasons, not necessary but otherwise could result in weird crashes.
+    accessible: Arc<AtomicBool>,
```
Bastacyclop: Is there a reason not to put the atomic inside `SharedBuffer` with `Mutex`/`RefCell`-like operations ?

src/backend/gl/src/command.rs
```diff
@@ -209,6 +216,16 @@ impl RawCommandBuffer {
         }
     }
 
+    /// Try to take access of the pool and shared memory.
+    pub(crate) fn take_access(&self) -> bool {
+        self.accessible.swap(false, atomic::Ordering::SeqCst)
```
Bastacyclop: Maybe this ordering is better suited https://doc.rust-lang.org/nomicon/atomics.html#acquire-release ?

src/backend/gl/src/command.rs
msiglreith: For `Individual` we have a HashMap for `SharedBuffer` and I want to protect the whole map and not the individual allocations, that's why I decided to move it out.

src/backend/gl/src/command.rs
msiglreith: Switched to `Acquire` on `take_access` and `Release` on `release_access`.

PR #1464, opened 2017-09-01
Mipmap generation for GL

snuk182:


src/core/src/factory.rs
```diff
@@ -425,24 +425,28 @@ pub trait Factory<R: Resources> {
...
                                    -> Result<(handle::Texture<R, T::Surface>,
                                               handle::ShaderResourceView<R, T::View>),
                                              CombinedError>
     {
         let surface = <T::Surface as format::SurfaceTyped>::get_surface_type();
         let num_slices = kind.get_num_slices().unwrap_or(1) as usize;
         let num_faces = if kind.is_cube() {6} else {1};
+        let mut texture_level = (data.len() / (num_slices * num_faces)) as texture::Level;
+        if let texture::Mipmap::Allocated = mipmap {
+	        texture_level *= kind.get_num_levels();
```
kvark: this is a rather non-trivial code for something that should be as trivial as
```rust
let levels = match mipmap {
   Allocated => {
               if data.len() != num_slices * num_faces {
                         return Err(...)
               }
               kind.get_num_levels()
   }
   Provided => {
               let num_slices = kind.get_num_slices().unwrap_or(1) as usize;
               let num_faces = if kind.is_cube() {6} else {1};
               (data.len() / (num_slices * num_faces)) as texture::Level
   }
};
```

src/core/src/factory.rs
```diff
@@ -462,7 +466,7 @@ pub trait Factory<R: Resources> {
         for (rd, d) in raw_data.iter_mut().zip(data.iter()) {
             *rd = cast_slice(*d);
         }
-        self.create_texture_immutable_u8::<T>(kind, &raw_data[.. data.len()])
+        self.create_texture_immutable_u8::<T>(kind, texture::Mipmap::Provided, &raw_data[.. data.len()])
```
kvark: the API of `create_texture_immutable` should be the same yes `create_texture_immutable_u8` except for the data type

src/backend/dx11/src/factory.rs
```diff
@@ -599,7 +599,7 @@ impl core::Factory<R> for Factory {
     }
 
     fn create_texture_raw(&mut self, desc: texture::Info, hint: Option<core::format::ChannelType>,
-                          data_opt: Option<&[&[u8]]>) -> Result<h::RawTexture<R>, texture::CreationError> {
+                          data_opt: Option<(&[&[u8]], texture::Mipmap)>) -> Result<h::RawTexture<R>, texture::CreationError> {
```
kvark: clearly, D3D11 needs more changes to get this working
at least you can return an error if it's `Mipmap::Generated`

snuk182: This will narrow the code portability, so the transfer GL->d3d11 will require code patch, while the mode is actually supported, just unparameterized.

kvark: I don't see how this mode is "actually supported" on D3D11. We construct `D3D11_SUBRESOURCE_DATA` per data slice. Having less of those just out of nowhere would cause the confusion in the driver.

src/core/src/factory.rs
snuk182:
```rust
let levels = match mipmap {
		    texture::Mipmap::Allocated => if data.len() != num_slices * num_faces {
                return Err(CombinedError::Texture(texture::CreationError::Mipmap));
            } else {
                kind.get_num_levels()
            },
		    texture::Mipmap::Provided => (data.len() / (num_slices * num_faces)) as texture::Level
		};
        let desc = texture::Info {
            kind: kind,
            levels: levels,
            format: surface,
            bind: SHADER_RESOURCE,
            usage: Usage::Data,
        };
```

src/core/src/factory.rs
snuk182: What did you mean here?

kvark: the parameters of `create_texture_immutable` should now be:
```rust
         &mut self,
         kind: texture::Kind,
         mipmap: texture::Mipmap,
         data: &[&[<T::Surface as format::SurfaceTyped>::DataType]])
```

snuk182:
Next PR committed.

PR #1465, opened 2017-09-01
[ll] associated constants for format types

kvark:
This is how it was meant to be played in the first place :)
Unfortunately, Rust-1.20 ICEs... Waiting for Beta to catch up, see https://github.com/rust-lang/rust/issues/29924


examples/core/quad/main.rs
```diff
@@ -17,8 +17,6 @@ extern crate gfx_core as core;
 #[cfg(feature = "dx12")]
 extern crate gfx_device_dx12 as back;
 #[cfg(feature = "vulkan")]
-extern crate gfx_device_vulkan as back;
```
msiglreith: Intentional?

kvark: yep. Apparently, it's not used.

kvark:
Replaced by https://github.com/gfx-rs/gfx/pull/1554/commits/6e773b6028a12532367c0e92019a16e53727d878
Strangely, no longer ICEs

Issue #1466, opened 2017-09-02
Rebuild high level on top of LL


Bastacyclop:
## 1 - Rebuild `gfx_render`

First, we need to rebuild `gfx_render` on top of the new low level core.
It will incorporate the old features from `gfx_core` and `gfx_render`:
- Resource handles
- Mapping tracking
- `Factory` and `Device`, with the new terminology `Device` and `Queue`
- `Encoder` will not be `Send` anymore and rely on pools
- `Slice`, see #1392
- Pipeline mecanisms, see #1429

## 2 - Split `gfx_app`

Then, we want our crates to have clearer identities as discussed in [gitter](https://gitter.im/gfx-rs/gfx?at=599370f7ee5c9a4c5fc4032c). That's why the old `gfx_app` will be split into three components:
- Utility code for the examples will live in the `examples` directory
- Utility code for the user will be moved into `gfx_render` (backend-agnostic)
- Backend selection will be the responsibility of a new `gfx_frontend` crate

## 3 - Improve

After these two steps, we will probably have a lot of polishing to do!

msiglreith:
Thanks for taking a stab at this topic!
The mentioned points sound good so far.

Relevant discussion (but also partially outdated): #1281

Bastacyclop:
Things to do after the first `render` implementation:
- [ ] Apply the conclusions of [this `Format` discussion](https://github.com/gfx-rs/gfx/pull/1476#discussion_r137530687).
- [ ] Clean it up and handle errors
- [ ] Write the documentation

kvark:
This is somewhat addressed by [Rendy](https://github.com/rustgd/rendy) now
cc @omni-viral

PR #1467, opened 2017-09-02
[ll][WIP] Gpu, Device and handles for render

Bastacyclop:
Working on #1466, starting with a fresh `cleanup()` implementation.
I commented out a lot of code to be able to compile it.

### DIFF
--- a/src/render/src/device.rs
+++ b/src/render/src/device.rs
@@ -1,20 +1,14 @@
-//! Device extension.
-//!
-//! This module serves as an extension to the `device` module in the `gfx_core` crate. This module
-//! exposes extension functions and shortcuts to aid with creating and managing graphics resources.
-//! See the `DeviceExt` trait for more information.
+// use std::error::Error;
+// use std::fmt;
+// use core::{self, buffer, format, state};
+use core::{Backend, Primitive, HeapType};
+// use core::pso::{CreationError, Descriptor};
+// use core::memory::{self, Bind, Pod};
+// use pso;
+// use shade::ProgramError;
+use handle::GarbageSender;
 
-use std::error::Error;
-use std::fmt;
-use core::{buffer, format, handle, texture, state};
-use core::{Backend, Primitive, ShaderSet};
-use core::device::Device;
-use core::pso::{CreationError, Descriptor};
-use core::memory::{self, Bind, Pod};
-use slice::{Slice, IndexBuffer, IntoIndexBuffer};
-use pso;
-use shade::ProgramError;
-
+/*
 /// Error creating a PipelineState
 #[derive(Clone, PartialEq, Debug)]
 pub enum PipelineStateError<S> {
@@ -82,10 +76,36 @@ impl<S> From<CreationError> for PipelineStateError<S> {
         PipelineStateError::DeviceCreate(e)
     }
 }
+*/
 
-/// This trait is responsible for creating and managing graphics resources, much like the `Device`
-/// trait in the `gfx_core` crate. Every `Device` automatically implements `DeviceExt`.
-pub trait DeviceExt<B: Backend>: Device<B> {
+#[derive(Clone)]
+pub struct Device<B: Backend> {
+    raw: B::Device,
+    // TODO: could be shared instead of cloned
+    heap_types: Vec<HeapType>,
+    memory_heaps: Vec<u64>,
+    garbage: GarbageSender<B>,
+}
+
+impl<B: Backend> Device<B> {
+    pub(crate) fn new(
+        raw: B::Device,
+        heap_types: Vec<HeapType>,
+        memory_heaps: Vec<u64>,
+        garbage: GarbageSender<B>) -> Self
+    {
+        Device { raw, heap_types, memory_heaps, garbage }
+    }
+
+    pub fn ref_raw(&self) -> &B::Device {
+        &self.raw
+    }
+
+    pub fn mut_raw(&mut self) -> &mut B::Device {
+        &mut self.raw
+    }
+
+/*
     /// Creates an immutable vertex buffer from the supplied vertices.
     /// A `Slice` will have to manually be constructed.
     fn create_vertex_buffer<T>(&mut self, vertices: &[T])
@@ -276,6 +296,5 @@ pub trait DeviceExt<B: Backend>: Device<B> {
             texture::WrapMode::Clamp,
         ))
     }
+    */
 }
-
-impl<B: Backend, F: Device<B>> DeviceExt<B> for F {}
--- a/src/render/src/handle.rs
+++ b/src/render/src/handle.rs
@@ -1,0 +1,172 @@
+use std::marker::PhantomData;
+use std::{ops, cmp, hash};
+use std::sync::mpsc;
+
+use Backend;
+
+pub(crate) type GarbageSender<B> = mpsc::Sender<Garbage<B>>;
+pub(crate) type GarbageReceiver<B> = mpsc::Receiver<Garbage<B>>;
+pub(crate) fn garbage_channel<B: Backend>() -> (GarbageSender<B>, GarbageReceiver<B>) {
+    mpsc::channel()
+}
+
+macro_rules! define_resources {
+    ($($name:ident: $info:path,)*) => {
+        pub enum Garbage<B: Backend> {
+            $( $name(B::$name), )*
+        }
+
+        pub mod inner {
+            use Backend;
+            use super::{Garbage, GarbageSender};
+            use std::{cmp, hash};
+
+            $(
+            
+            #[derive(Clone, Debug)]
+            pub struct $name<B: Backend> {
+                resource: Option<B::$name>,
+                info: $info,
+                garbage: GarbageSender<B>
+            }
+
+            impl<B: Backend> $name<B> {
+                pub(crate) fn new(
+                    resource: B::$name,
+                    info: $info,
+                    garbage: GarbageSender<B>) -> Self
+                {
+                    $name {
+                        resource: Some(resource),
+                        info,
+                        garbage,
+                    }
+                }
+
+                pub fn resource(&self) -> &B::$name {
+                    self.resource.as_ref().unwrap()
+                }
+
+                pub fn info(&self) -> &$info {
+                    &self.info
+                }
+            }
+
+            impl<B: Backend> cmp::PartialEq for $name<B>
+                where B::$name: cmp::PartialEq
+            {
+                fn eq(&self, other: &$name<B>) -> bool {
+                    self.resource().eq(&other.resource())
+                }
+            }
+
+            impl<B: Backend> cmp::Eq for $name<B>
+                where B::$name: cmp::Eq
+            {}
+
+            impl<B: Backend> hash::Hash for $name<B>
+                where B::$name: hash::Hash
+            {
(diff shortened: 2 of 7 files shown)


src/render/src/handle.rs
```diff
@@ -0,0 +1,85 @@
...
+pub(crate) type GarbageSender<B> = mpsc::Sender<Garbage<B>>;
+pub(crate) type GarbageReceiver<B> = mpsc::Receiver<Garbage<B>>;
+pub(crate) fn garbage_channel<B: Backend>() -> (GarbageSender<B>, GarbageReceiver<B>) {
+    mpsc::channel()
+}
+
+macro_rules! resource_handle {
+    ($handle:ident, $guard:ident, $resource:ident) => {
+        #[derive(Clone, Debug)]
+        struct $guard<B: Backend>(Option<B::$resource>, GarbageSender<B>);
```
msiglreith: Would you mind explaining why we need `Option<_>` here?

src/render/src/lib.rs
```diff
@@ -143,3 +148,80 @@ pub mod pso;
...
 /// Convenience macros
 pub mod macros;
+*/
+
+use std::ops;
+use std::sync::mpsc;
+
+pub struct Gpu<B: Backend> {
+    device: Device<B>,
+    queue: CommandQueue<B, core::General>,
```
msiglreith: should be `General` or `Graphics`. OpenGL backend will only expose graphics queues if there is no compute support,
Else I think every backend exposes general but need to double-check.

src/render/src/lib.rs
```diff
@@ -143,3 +148,80 @@ pub mod pso;
...
+    garbage: handle::GarbageReceiver<B>,
+}
+
+impl<B: Backend> Gpu<B> {
+    pub fn new<A: core::Adapter<B>>(adapter: &A) -> Self {
+        let queue_descs = adapter.get_queue_families()
+            .iter()
+            .map(|&(ref family, qtype)| (family, qtype, family.num_queues()) )
+            .collect::<Vec<_>>();
+
```
msiglreith: Somehow we need to filter for queues which can be used for presentation with the display surface.

src/render/src/lib.rs
```diff
@@ -143,3 +148,80 @@ pub mod pso;
...
+                ConstantBufferView(cbv) => dev.destroy_constant_buffer_view(cbv),
+                ShaderResourceView(srv) => dev.destroy_shader_resource_view(srv),
+                UnorderedAccessView(uav) => dev.destroy_unordered_access_view(uav),
+                Sampler(s) => dev.destroy_sampler(s),
+            }
+        }
+    }
+}
+
+impl<B: Backend> ops::Deref for Gpu<B> {
```
msiglreith: We might rather expose explicit `get_device`/`ref_device`/..(?) functions than `Deref`.

src/render/src/handle.rs
Bastacyclop: I need ownership on `Drop` to send the resource on the garbage channel for future cleanup.

src/render/src/lib.rs
Bastacyclop: Would a third `Compute`-only alternative be useful for some use cases ?

msiglreith: I would say barely

src/render/src/lib.rs
Bastacyclop: This will become a bit more verbose for users that are not actually using the `Device` split, but I agree that the use of `Deref` might be a bit abusive here.

src/render/src/device.rs
```diff
@@ -82,10 +76,19 @@ impl<S> From<CreationError> for PipelineStateError<S> {
         PipelineStateError::DeviceCreate(e)
     }
 }
+*/
+
+#[derive(Clone)]
+pub struct Device<B: Backend> {
+    pub(crate) device: B::Device,
```
kvark: convention would probably be to name it `raw`

src/render/src/device.rs
```diff
@@ -82,10 +76,19 @@ impl<S> From<CreationError> for PipelineStateError<S> {
...
+    // TODO: could be shared instead of cloned
+    pub(crate) heap_types: Vec<HeapType>,
```
kvark: why do all of them need to be `pub(crate)`?

src/render/src/device.rs
```diff
@@ -276,6 +279,5 @@ pub trait DeviceExt<B: Backend>: Device<B> {
             texture::WrapMode::Clamp,
         ))
     }
-}
-
-impl<B: Backend, F: Device<B>> DeviceExt<B> for F {}
+    */
+}
```
kvark: nit: newline

src/render/src/handle.rs
```diff
@@ -0,0 +1,85 @@
...
+            fn drop(&mut self) {
+                self.1.send(Garbage::$resource(self.0.take().unwrap()))
+                    .unwrap_or_else(|e| {
+                        error!("Could not drop {}: {}", stringify!($resource), e)
+                    });
+            }
+        }
+
+        #[derive(Clone, Debug)] // TODO: PartialEq, Eq, Hash
+        pub struct $handle<B: Backend>(Arc<$guard<B>>);
```
kvark: we might need more stuff carried out by the handle, e.g. stuff like `BufferInfo`, `ProgramInfo`, etc

src/render/src/handle.rs
```diff
@@ -0,0 +1,85 @@
...
+
+resource_handle! { ShaderLib, ShaderLibGuard, ShaderLib }
+resource_handle! { Buffer, BufferGuard, Buffer }
+resource_handle! { Image, ImageGuard, Image }
+resource_handle! { RenderTargetView, RenderTargetViewGuard, RenderTargetView }
+resource_handle! { DepthStencilView, DepthStencilViewGuard, DepthStencilView }
+resource_handle! { ConstantBufferView, ConstantBufferViewGuard, ConstantBufferView }
+resource_handle! { ShaderResourceView, ShaderResourceViewGuard, ShaderResourceView }
+resource_handle! { UnorderedAccessView, UnorderedAccessViewGuard, UnorderedAccessView }
+resource_handle! { Sampler, SamplerGuard, Sampler }
```
kvark: please configure your editor to add new lines automatically

src/render/src/handle.rs
```diff
@@ -0,0 +1,85 @@
...
+
+        impl<B: Backend> $handle<B> {
+            pub fn resource(&self) -> &B::$resource {
+                &self.0.resource()
+            }
+        }
+    }
+}
+
+resource_handle! { ShaderLib, ShaderLibGuard, ShaderLib }
```
kvark: The macro parameters should be different, i.e. `{Image, ImageInfo}`.
The 3rd parameter matches the 1st, so no need to provide it.
The 2nd parameter could be avoided if we have `mod guard { //all the guards with the same name as their resources`, so that `guard::Image` is the one being a part of `Image` handle. This may require specifying all the handle types in a single macro invocation, so that you can open and close the `guard` mod once.

src/render/src/device.rs
Bastacyclop: It was mainly to avoid writing a redundant `Device::new` but I can remove them.

src/render/src/lib.rs
```diff
@@ -143,3 +148,80 @@ pub mod pso;
...
+    }
+
+    pub fn cleanup(&mut self) {
+        use core::Device;
+
+        let dev = &mut self.device.device;
+        for garbage in self.garbage.try_iter() {
+            use handle::Garbage::*;
+            match garbage {
+                ShaderLib(sl) => dev.destroy_shader_lib(sl),
```
msiglreith: We lack documentation for the `core` api so you probably aren't aware of it but resources need to _externally synchronized_ (in vulkan terms) on destruction. This means that they (or derivates like image views) are not in flight anymore/process by the GPU on destruction. This requires fencing to create a sync point between CPU-GPU. To hide costs and reduce blocking by fences the usual approach is to have basically ringbuffer of 'frames' (e.g size of 3). The approach then would be to re-use existing fences from these frames and cleanup at frame boundaries with a bit of delay: e.g queue up for destruction at frame 0, waiting until beginning of the next frame 0 to destroy. This gives the GPU 2 frames time to catch up and signal frame 0 fence.
The concept extends to command buffers/pools as well. In our `quad` example we wait til the GPU idles after every frame which is slow but currently more a testing ground than best practice.

materials:
- http://developer.download.nvidia.com/gameworks/events/GDC2016/Vulkan_Essentials_GDC16_tlorach.pdf (slide 25, graphical visualization)
- https://www.khronos.org/assets/uploads/developers/library/2016-siggraph/3D-BOF-SIGGRAPH_Jul16.pdf (slide 54, example approach from UE4)

Bastacyclop: Thanks for the explanation. I think holding resources handles until synchronisation has been done would be a good way to ensure that cleanup is not done early.

src/render/src/handle.rs
```diff
@@ -1,85 +1,153 @@
...
 }
 
-macro_rules! resource_handle {
-    ($handle:ident, $guard:ident, $resource:ident) => {
-        #[derive(Clone, Debug)]
-        struct $guard<B: Backend>(Option<B::$resource>, GarbageSender<B>);
+macro_rules! define_resources {
+    ($($name:ident: $info:path,)*) => {
+        pub enum Garbage<B: Backend> {
+            $( $name(B::$name), )*
```
kvark: nice!

src/render/src/handle.rs
```diff
@@ -1,85 +1,153 @@
...
+        }
+
+        pub mod inner {
```
kvark: does it have to be `pub`?

src/render/src/handle.rs
```diff
@@ -1,85 +1,153 @@
...
+                            error!("Could not drop {}: {}", stringify!($name), e));
+                }
             }
+
+            )*
+        }
+
+        pub mod raw {
+            use std::sync::Arc;
+            $( pub type $name<B> = Arc<super::inner::$name<B>>; )*
```
kvark: heh, cool

src/render/src/lib.rs
kvark: might be easier with #1468

src/render/src/lib.rs
```diff
@@ -150,36 +152,44 @@ pub mod shade;
...
-        let core::Gpu { device, mut general_queues, heap_types, memory_heaps, .. } =
-            adapter.open(&queue_descs);
+        let core::Gpu {
+            device,
+            mut general_queues,
+            mut graphics_queues,
+            heap_types,
+            memory_heaps,
+            ..
+        } = adapter.open(&queue_descs);
```
kvark: note: there is `open_with`, which is a bit shorter

src/render/src/handle.rs
Bastacyclop: Since I'm using a type alias for the handle and not a wrapper, I think so. It also allows to call `resource()` and `info()` without any boilerplate on our side. It's the equivalent of the old [`buffer::Raw`](https://github.com/gfx-rs/gfx/blob/v0.16/src/core/src/buffer.rs#L24) for example.

Bastacyclop:
@kvark I think it's ready for merge unless you'd rather have me continue on this PR. The queue filtering can be resolved later IMO (after #1468 is merged ?).

PR #1469, opened 2017-09-05
[ll][gl] glutin integration

kvark:
A follow-up to #1468
I think `gl` needs an `Instance` type, and every other backend needs, too.
Perhaps, the `Instance` should be implementing the `Backend` trait even, to make the passing around and initialization more consistent.


src/backend/gl/Cargo.toml
```diff
@@ -12,10 +12,15 @@ documentation = "https://docs.rs/gfx_device_gl"
 workspace = "../../.."
 
 [lib]
-name = "gfx_device_gl"
+name = "gfx_backend_gl"
+
+[features]
+default = ["glutin"]
+headless = ["glutin"]
```
msiglreith: I think we can remove the `headless` feature as it got promoted to a first-class feature with the introduction of `Headless` trait.

src/backend/gl/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "gfx_device_gl"
-version = "0.14.0"
+name = "gfx_backend_gl"
```
Bastacyclop: Nice to see it renamed! :D Although a crates.io search for "gfx" might look weird.. Is there a way to remove a crate from public search like a full crate yank ?

msiglreith: I would prefer a large `[DEPRECATED]` sign in the crate description text for the `gfx_device_` and `gfx_window_` crates.

kvark:
@msiglreith PTAL, I think this is ready to merge now

src/backend/gl/src/window/glutin.rs
```diff
@@ -0,0 +1,172 @@
...
+        };
+
+        Swapchain {
+            window: self.window.clone(),
+            backbuffer: [backbuffer; 1],
+        }
+    }
+}
+
+impl core::Instance<B> for Surface {
```
msiglreith: ha! That's an interesting approach :D

PR #1474, opened 2017-09-06
[ll] big refactor, ranges

kvark:


src/core/src/command/raw.rs
```diff
@@ -19,10 +20,35 @@ pub trait RawCommandBuffer<B: Backend>: Clone + Send {
...
     ///
-    fn pipeline_barrier(&mut self, &[Barrier<B>]);
+    fn pipeline_barrier(
+        &mut self,
+        stages: Range<pso::PipelineStage>,
+        barriers: &[Barrier<B>],
+    );
+
+    ///
+    fn fill_buffer(
```
msiglreith: How would you map this to D3D12?

src/backend/gl/src/lib.rs
```diff
@@ -528,115 +528,116 @@ impl CommandQueue {
...
+                                base_vertex as _,
                             );
-                        } else if base_vertex == 0 {
-                            gl.DrawElements(
+                        }
+                    } else {
+                        error!("Base vertex with indexed drawing not supported");
+                    }
+                } else if features.draw_indexed_instanced {
+                    if base_vertex == 0 && instances.start == 0 {
```
msiglreith: Can we reduce the checks for `instances.start == 0`? Currently 3 times in row

kvark: it would require more nesting/depth to the code structure, I'd rather have this condition repeated instead

src/core/src/command/raw.rs
kvark: (planned)

PR #1475, opened 2017-09-07
[ll] vk: Fix win32 initialization

msiglreith:
* Fix build for surface creation
* Filter request layers to only initialize with available layers, else instance creation can fail (e.g debug layers may not be registered when installing LunarG SDK on windows afaik)

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -65,10 +65,10 @@ impl Drop for RawInstance {
 pub struct Instance {
     pub raw: Arc<RawInstance>,
 
-    /// Supported surface extensions of this instance.
-    pub surface_extensions: Vec<&'static str>,
+    /// Supported extensions of this instance.
+    pub extensions: Vec<&'static str>,
 
-    //TODO: move into `RawInstance`, destroy in `drop`
+    // TODO: move into `RawInstance`, destroy in `drop`
     _debug_report: Option<(ext::DebugReport, vk::DebugReportCallbackEXT)>,
 }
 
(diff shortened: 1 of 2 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -131,20 +131,43 @@ impl Instance {
...
+            .iter()
+            .chain(EXTENSIONS.iter())
+            .filter_map(|ext| {
+                instance_extensions
+                    .iter()
+                    .find(|inst_ext| unsafe {
+                        CStr::from_ptr(inst_ext.extension_name.as_ptr()) ==
+                            CStr::from_ptr(ext.as_ptr() as *const i8)
+                    })
+                    .and_then(|_| Some(*ext))
```
kvark: `and_then` doesn't seem useful in this context, did you mean `map`?

src/backend/vulkan/src/lib.rs
```diff
@@ -131,20 +131,43 @@ impl Instance {
...
+        let layers = LAYERS
+            .iter()
+            .filter_map(|layer| {
+                instance_layers
+                    .iter()
+                    .find(|inst_layer| unsafe {
+                        CStr::from_ptr(inst_layer.layer_name.as_ptr()) ==
+                            CStr::from_ptr(layer.as_ptr() as *const i8)
+                    })
+                    .and_then(|_| Some(*layer))
```
kvark: same here

src/backend/vulkan/src/lib.rs
```diff
@@ -131,20 +131,43 @@ impl Instance {
...
+            })
+            .collect::<Vec<&str>>();
+
+        // Check requested layers against the available layers
+        let layers = LAYERS
+            .iter()
+            .filter_map(|layer| {
+                instance_layers
+                    .iter()
+                    .find(|inst_layer| unsafe {
```
kvark: perhaps, we should `warn!` for layers/extensions that are not found

msiglreith:
@kvark Thanks for the review, could you take a look again?

PR #1476, opened 2017-09-07
[ll] Frames and encoders for render

Bastacyclop:
Working on #1466.
This is an untested draft, I hope there are not too many mistakes.
The interesting stuff is in [core/pool](https://github.com/gfx-rs/gfx/compare/ll...Bastacyclop:ll?expand=1#diff-a8c8cf11be50bbdb6bab730b503bfe00), [render/device](https://github.com/gfx-rs/gfx/compare/ll...Bastacyclop:ll?expand=1#diff-a505eeca0919c44fc9d868f9a1cbdfe3) and more importantly [render/encoder](https://github.com/gfx-rs/gfx/compare/ll...Bastacyclop:ll?expand=1#diff-e76accb5ddad96024c8ebca0ad05ee68) and [render/lib](https://github.com/gfx-rs/gfx/compare/ll...Bastacyclop:ll?expand=1#diff-be2c920c08146924af4d1983d6894d63).
My next step will be to test it in a copy of the quad example.

You can acquire `encoder::Scope`s from which you can get actual `Encoder`s. An `encoder::Scope` is basically a more convenient command buffer pool. So there are two reasons to use multiple scopes: to use multiple encoders at the same time (multithreading mostly) and to release all the acquired encoders independently (different encoding scopes). Maybe scope is not a good name and I should name it simply pool instead.


src/render/src/device.rs
```diff
@@ -105,6 +135,198 @@ impl<B: Backend> Device<B> {
...
+            ).into())
+    }
+
+    pub fn view_buffer_as_constant<T>(&mut self,
+        buffer: &handle::Buffer<B, T>,
+        range: Range<u64>,
+    ) -> Result<handle::ConstantBufferView<B, T>, TargetViewError>
+    {
+        self.view_buffer_as_constant_raw(buffer, range)
+            .map(|raw| Typed::new(raw))
```
kvark: nit: could be shorter

src/render/src/device.rs
```diff
@@ -105,6 +135,198 @@ impl<B: Backend> Device<B> {
...
+                srv,
+                image.into(),
+                self.garbage.clone()
+            ).into())
+    }
+
+    // TODO: rename to simply ViewError ?
+    pub fn view_image_as_shader_resource<F>(&mut self, image: &handle::Image<B, F>)
+        -> Result<handle::ShaderResourceView<B, F>, TargetViewError>
+        where F: format::TextureFormat
```
kvark: note to self: need to rename it to `format::ImageFormat`

src/render/src/encoder.rs
```diff
@@ -1,19 +1,91 @@
...
-use core::memory::{self, cast_slice, Typed, Pod, Usage};
-use slice;
-use pso;
+use core::{self, command, format};
+use core::command::CommandBuffer;
+use memory::{self, cast_slice, Typed, Pod, Usage};
+use {handle, image};
+
+// This is the unique owner of the inner UnsafeCell.
+pub struct Scope<B: Backend, C>(Arc<UnsafeCell<ScopeInner<B, C>>>);
```
kvark: could use `Arc<Mutex<_>>` instead. it's safer and is fast with `futex` backing when no contention is present

src/render/src/encoder.rs
```diff
@@ -1,19 +1,91 @@
...
+        inner.mut_pool().reserve(additional);
+    }
+
+    pub fn acquire_encoder<'a>(&'a mut self) -> Encoder<'a, B, C> {
+        let inner = unsafe { &mut *self.0.get() };
+        Encoder {
+            buffer: inner.mut_pool().acquire_command_buffer(),
+            // raw_data: pso::RawDataSet::new(),
+            // access_info: command::AccessInfo::new(),
+            handles: handle::Bag::new(),
```
kvark: lol, great name!

src/render/src/encoder.rs
```diff
@@ -39,11 +111,11 @@ pub enum CopyError<S, D> {
 /// Result type returned when copying a buffer into another buffer.
 pub type CopyBufferResult = Result<(), CopyError<usize, usize>>;
 
-/// Result type returned when copying buffer data into a texture.
-pub type CopyBufferTextureResult = Result<(), CopyError<usize, [texture::Size; 3]>>;
+/// Result type returned when copying buffer data into an image.
+pub type CopyBufferImageResult = Result<(), CopyError<usize, [image::Size; 3]>>;
```
kvark: btw, could use `image::Dimensions`?

src/render/src/handle.rs
```diff
@@ -101,72 +115,67 @@ define_resources! {
...
+pub type ShaderResourceView<B, T> = Typed<raw::ShaderResourceView<B>, T>;
+pub type UnorderedAccessView<B, T> = Typed<raw::UnorderedAccessView<B>, T>;
+
+pub use self::raw::Sampler;
+
+#[derive(Debug, Clone)]
+pub enum ViewSource<B: Backend> {
+    Image(raw::Image<B>),
+    Buffer(raw::Buffer<B>),
+    Backbuffer(B::Image, ::image::Info),
```
kvark: looks great IMO

src/render/src/lib.rs
```diff
@@ -146,70 +140,259 @@ pub mod shade;
...
 }
+*/
 
-pub struct Gpu<B: Backend> {
+// TODO: could be named Instance too ?
+pub struct Context<B: Backend> {
+    surface: B::Surface,
     device: Device<B>,
-    queue: Queue<B>,
+    queue: Queue<B, core::Graphics>,
```
kvark: no compute yet?

src/core/src/pool.rs
```diff
@@ -88,9 +91,7 @@ impl<B: Backend, C> CommandPool<B, C> {
     /// If more command buffers are requested than allocated, new buffers will be reserved.
     /// The command buffer will be returned in 'recording' state.
     pub fn acquire_command_buffer<'a>(&'a mut self) -> CommandBuffer<'a, B, C> {
-        if self.buffers.len() <= self.next_buffer {
-            self.reserve(1);
-        }
+        self.reserve(1);
 
         let buffer = &mut self.buffers[self.next_buffer];
         self.next_buffer += 1;
```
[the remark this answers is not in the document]
msiglreith: set to 0 on `reset`

src/render/src/device.rs
```diff
@@ -105,6 +135,198 @@ impl<B: Backend> Device<B> {
...
+        mip_levels: image::Level,
+        format: format::Format,
+        bind: memory::Bind,
+        usage: memory::Usage
+    ) -> Result<handle::raw::Image<B>, image::CreationError>
+    {
+        unimplemented!()
+    }
+
+    // FIXME?: used to be format::SurfaceTyped
```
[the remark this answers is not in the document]
msiglreith: Image view doesn't need to have the same format as the image they are created from (can differ in surface and channel type). The format only needs to be compatible. In particular, on GL this is only possible with texture view support (4.3+).

src/render/src/lib.rs
```diff
@@ -146,70 +140,259 @@ pub mod shade;
...
-            .collect::<Vec<_>>();
-        
+pub struct Backbuffer<B: Backend, Cf: RenderFormat> {
+    color: handle::RenderTargetView<B, Cf>,
+    // TODO: depth
+}
+
+use self::SyncState::*;
+#[derive(PartialEq)]
+enum SyncState {
```
msiglreith: What's the deal with these sync states?

src/render/src/lib.rs
```diff
@@ -146,70 +140,259 @@ pub mod shade;
...
+    }
+
+    pub fn acquire_graphics_scope(&mut self) -> encoder::Scope<B, core::Graphics> {
+        self.queue.acquire_encoder_scope()
     }
 
-    pub fn cleanup(&mut self) {
-        use core::Device;
+    // TODO: allow submissions before present
+    pub fn present(&mut self, submits: Vec<encoder::Submit<B, core::Graphics>>) {
```
msiglreith: Is there something preventing me from submitting command buffers from a pool which I requested in a previous frame?

src/render/src/device.rs
```diff
@@ -105,6 +135,198 @@ impl<B: Backend> Device<B> {
...
+            self.raw.create_sampler(info), info, self.garbage.clone()
+        ).into()
+    }
+
+    pub fn view_buffer_as_constant_raw(&mut self,
+        buffer: &handle::raw::Buffer<B>,
+        range: Range<u64>,
+    ) -> Result<handle::raw::ConstantBufferView<B>, TargetViewError>
+    {
+        use handle::inner::*;
```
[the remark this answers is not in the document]
Bastacyclop: no, I'm just lazy :p

src/render/src/encoder.rs
Bastacyclop: If you want, but there is not even dynamic ownership.

kvark: I don't strongly mind either way, it just seems that having less `unsafe` in the codebase is slightly more appealing, if there is no cost to it

src/render/src/encoder.rs
Bastacyclop: It can be both dimensions or an offset, so I don't think that `AaMode` makes sense here.

src/render/src/lib.rs
Bastacyclop: Not yet, it was simpler to write. I thought about having two methods to `acquire_general/graphics_scope` but both can fail (even graphics since there is no downcasting). This would not be very convenient to use. I also thought about having an enum inside the `Encoder` but it seemed annoying to match all the time in the implementation.

src/render/src/lib.rs
Bastacyclop: The names might need some work but I use it to track synchronisation state at runtime.
For semaphores:
`Unreached` means that it has not been setup.
`Reached` means that it has been setup.
For fences:
`Unreached` means that the fence has not been waited upon.
`Reached` means that the fence has been waited upon.

It's also probably not necessary for all three primitives used here but it's easy to remove where not needed.

src/render/src/lib.rs
Bastacyclop: No, that would extend the lifetime of the associated scope (with `ScopeDependency`) and keep it from being reset until unused.

src/render/src/device.rs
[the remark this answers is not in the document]
Bastacyclop: I could also put the glob import in the beginning of the file, everything is used somewhere in the module.

src/render/src/encoder.rs
kvark: ok

src/render/src/lib.rs
kvark: Can we lift the `C` generic to the `Gpu` level (in `gfx_render`) and have 2 kind of constructors: one with compute (so, `Generic`) and other without (so, `Graphics`). More often than not, the client knows if they need to run compute.

src/render/src/lib.rs
msiglreith: The pool gets send back to the queue and added to the current frame once everything has been submitted if I understand it correctly?

src/render/src/device.rs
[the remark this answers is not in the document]
msiglreith: (from the gitter discussion)
Vulkan (if flag set) and Metal only require the pixel size of the format to be the same (excluding some restriction for depth-stencil formats), so we can specify SurfaceType and ChannelType for views.
D3D only allows changing the ChannelType, but also the SurfaceType for UAVs.

Imo we should stick to specifying the full format for creation, channel for SRV and full format for UAVs.

src/render/src/lib.rs
Bastacyclop: `Encoder` and `Submit` hold a scope dependency. On submission it is also added to `FrameBundle`. Once a frame fence is waited upon, the associated `FrameBundle` dependencies are cleared. The pool gets send back to the queue when there is no encoder, submit or frame still referencing the scope. Only then it can be reset and recycled into another scope.

src/render/src/encoder.rs
```diff
@@ -1,19 +1,101 @@
...
-use slice;
-use pso;
+use core::{self, command, format};
+use core::command::CommandBuffer;
+use memory::{self, cast_slice, Typed, Pod, Usage};
+use {handle, image};
+
+// This is the unique owner of the inner struct.
+pub struct Scope<B: Backend, C>(Arc<Mutex<ScopeInner<B, C>>>);
+pub struct ScopeGuard<'a, B: Backend, C: 'a> {
```
Bastacyclop: Here is the `Mutex` version, I couldn't find a better way to write it. The API suffers a bit IMO.

src/render/src/lib.rs
```diff
@@ -146,70 +140,259 @@ pub mod shade;
...
+        let mut device = Device::new(device, heap_types, memory_heaps, garbage_sender);
+
+        let backbuffers = backbuffers.into_iter()
+            .map(|raw| {
+                Backbuffer {
+                    color: Typed::new(device
+                        .view_backbuffer_as_render_target_raw(
+                            raw.color,
+                            unimplemented!(),
+                            unimplemented!(),
```
Bastacyclop: How should I retrieve the backbuffer dimensions for `Kind` ? And what should I put in `Bind` ?

kvark: Perhaps, we should add `Surface::get_kind()` method to the trait. For `Bind`, I think having `RENDER_TARGET` is required, and maybe we can have `TRANSFER_SRC`.

src/render/src/lib.rs
msiglreith: Okay, I think I grasped everything. Thanks for explanation! Looks solid!

src/render/src/encoder.rs
Bastacyclop: @kvark So do you think the internal safety is worth it here ?

src/render/src/lib.rs
```diff
@@ -146,70 +140,259 @@ pub mod shade;
...
-        } else {
-            Queue::General(general_queues.remove(0))
-        };
+        let queue = Queue::new(graphics_queues.remove(0));
+
+        let swap_config = core::SwapchainConfig::new()
+            .with_color::<Cf>();
+        // TODO: call supports_queue
+        let mut swapchain = surface.build_swapchain(swap_config, &queue.inner);
+        let backbuffers = Vec::<core::Backbuffer<B>>::new(); // TODO: take swapchain backbuffers
```
Bastacyclop: @msiglreith Should I modify `core` for this or can you do it ?

src/render/src/encoder.rs
Bastacyclop: I would say no, unless we find a way to keep the same API.

src/render/src/lib.rs
Bastacyclop: Works great :)

Bastacyclop:
I should probably avoid the `quad` data copy and only have the `main.rs` and cargo setup.

Cargo.toml
```diff
@@ -17,5 +17,6 @@ members = [
     #"src/window/metal",
     #"src/window/sdl",
     #"examples",
-    "examples/core/quad"
+    "examples/core/quad",
+    "examples/render/quad_render",
```
kvark: why not converting the existing triangle example (written against `gfx`) instead of adding a new one?

Bastacyclop: I would have to rewrite the whole PSO stuff using `core` until it's ready in `render`, so I'd rather not. But I only plan to keep this new example temporarily. Once `render` will be more complete we can remove it and rewrite the old ones.

kvark:
let's not hold this off the ground for too long ;)
bors r+

kvark:
We should exclude the examples from `cargo test --all` on OSX, given that Metal backend is not ready yet. Also, nightly builds doing nothing for some reason?

msiglreith:
I assume the issue with OS X, is the default feature `vulkan` of the quad example and our vulkan backend isn't built with OS X support in mind (obviously..).

PR #1479, opened 2017-09-07
[ll] gl: Mutexes for command pool storage

msiglreith:
Switch from `UnsafeCell`+`AtomicBool` over to `Mutex` as discussed on gitter.
Motivated by low performance penality for the (usual) uncontained case for the memory access.

Now shares the complete `BufferMemory` instead of only `OwnedBuffer` because there seems no library support to `map` a mutex or share it for multiple data structures.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -11,10 +11,8 @@ use {native as n, Backend};
 use pool::{self, BufferMemory};
 
 use std::mem;
-use std::cell::UnsafeCell;
 use std::ops::Range;
-use std::sync::Arc;
-use std::sync::atomic::{self, AtomicBool};
+use std::sync::{Arc, Mutex};
 
 // Command buffer implementation details:
 //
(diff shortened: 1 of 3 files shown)


src/backend/gl/src/command.rs
```diff
@@ -167,34 +162,30 @@ pub struct RawCommandBuffer {
...
-            BufferMemory::Individual { ref mut storage, ref mut next_buffer_id } => {
-                // Add a new pair of buffers
-                let buffer = Arc::new(UnsafeCell::new(pool::OwnedBuffer::new()));
-                storage.insert(*next_buffer_id, buffer.clone());
-                let id = *next_buffer_id;
-                *next_buffer_id += 1;
-                (buffer, id, true)
+        let (id, individual_reset) = {
+            let mut memory = memory.lock().unwrap();
+            match *memory {
```
kvark: nit: why not just `match memory.lock().unwrap()` or even `match memory.lock()`?

src/backend/gl/src/command.rs
```diff
@@ -222,18 +203,21 @@ impl RawCommandBuffer {
     }
 
     fn push_cmd(&mut self, cmd: Command) {
-        if self.take_access() {
-            let mut cmd_buffer = unsafe { &mut (*self.memory.get()).commands };
+        let slice = {
+            let mut memory = self.memory.lock().unwrap();
```
Bastacyclop: So pushing command while the pool is in access is now legal ? You could also use `try_lock` and keep the restriction.

kvark: sounds like a good idea

PR #1481, opened 2017-09-07
Mapping improvements: raw API + ranges

kvark:


kvark:
> I would rename it to match the vulkan implementation

I see certain value in keeping it consistent here, e.g. `read_mapping` vs `read_mapping_raw`. What naming do you have in mind?

> Won't we need access to features like invalidation and flush too ?

Probably will need it.

> As a side note, we could still use write-only pointers/slices.

Could you elaborate?

Bastacyclop:
I would rename both raw and typed version actually, to `map(_buffer)?_(readable|writable)[_raw]` or something.
Write-only would be nice to avoid reading from write-only mappings, if it was possible.

PR #1486, opened 2017-09-11
Indexed triangle strip cut for TriangleLL

kvark:
This change was done for a GPUWeb test (positive!), but I don't think why we'd want this wasted ;)

### DIFF
--- a/src/backend/dx12ll/src/factory.rs
+++ b/src/backend/dx12ll/src/factory.rs
@@ -421,7 +421,8 @@ impl core::Factory<R> for Factory {
                     pInputElementDescs: input_element_descs.as_ptr(),
                     NumElements: input_element_descs.len() as u32,
                 },
-                IBStripCutValue: winapi::D3D12_INDEX_BUFFER_STRIP_CUT_VALUE_DISABLED,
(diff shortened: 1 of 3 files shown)


src/backend/dx12ll/src/factory.rs
```diff
@@ -421,7 +421,8 @@ impl core::Factory<R> for Factory {
                     pInputElementDescs: input_element_descs.as_ptr(),
                     NumElements: input_element_descs.len() as u32,
                 },
-                IBStripCutValue: winapi::D3D12_INDEX_BUFFER_STRIP_CUT_VALUE_DISABLED,
+                //IBStripCutValue: winapi::D3D12_INDEX_BUFFER_STRIP_CUT_VALUE_DISABLED,
+                IBStripCutValue: winapi::D3D12_INDEX_BUFFER_STRIP_CUT_VALUE_0xFFFFFFFF, //note: this works with U16 too!
```
msiglreith: Looks a bit hackish, but if it works ..

kvark:
Oh great:
> error: An unknown error occurred

msiglreith:
Travis also fails with unknown errors, let's try again ..

bors r+

msiglreith:
Travis passed in the 2nd run and appveyor only fails on nightly, probably some temporary regression.
Merging it manually.

PR #1489, opened 2017-09-11
[ll] shader module API rework

kvark:
I made it pretty Vulkan-ic :)
In the future, we may add specialization to `EntryPoint` struct.

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -5,7 +5,7 @@ use core::{Features, Limits, HeapType};
 use core::memory::Requirements;
 use native as n;
 use std::{mem, ptr};
-use std::collections::BTreeMap;
+use std::ffi::CString;
 use std::ops::Range;
 use std::sync::Arc;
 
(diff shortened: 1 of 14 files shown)


src/backend/vulkan/src/device.rs
```diff
@@ -573,61 +501,49 @@ impl d::Device<B> for Device {
...
-                        Ok(_) => {
-                            let pipeline = *pipelines.next().unwrap();
-
-                            // Check if pipeline compiled correctly
-                            if pipeline == vk::Pipeline::null() {
-                                Err(pso::CreationError::Other) // TODO
-                            } else {
-                                Ok(n::GraphicsPipeline(pipeline))
-                            }
+            Ok(pipelines) |
```
msiglreith: Nice one! I remember that this logic was a pain to write :)

src/backend/vulkan/src/device.rs
```diff
@@ -711,12 +612,34 @@ impl d::Device<B> for Device {
...
             self.raw.0.create_framebuffer(&info, None)
-                        .expect("error on framebuffer creation")
-        };
+        }.expect("error on framebuffer creation");
 
         n::FrameBuffer { raw: framebuffer }
     }
 
+    fn create_shader_module(&mut self, spirv_data: &[u8]) -> Result<n::ShaderModule, d::ShaderError> {
+        assert_eq!(spirv_data.len() & 3, 0);
```
msiglreith: Could we add a comment for this. The size of SPIR-V modules requiring to be a multiple of the size of DWORD is not that obvious imo.

src/core/src/lib.rs
```diff
@@ -214,6 +214,34 @@ pub struct HeapType {
     pub heap_index: usize,
 }
 
+/// Shader entry point.
+#[derive(Debug)]
+pub struct EntryPoint<'a, B: Backend> {
```
msiglreith: Could be moved into `pso` imo.

kvark:
@msiglreith thanks for reviewing! Everything should be addressed now

PR #1494, opened 2017-09-13
[ll] Pixel to Fragment rename

kvark:
I believe `fragment` is a better term (it's not always pixels being processed) and more consistent with Vulkan. Sorry for bikeshedding!
Also removed explicit rasterizer discard, we'll instead just check for the fragment shader presence.

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -268,7 +268,7 @@ impl d::Device<B> for Device {
                 stages.push(make_stage(vk::SHADER_STAGE_VERTEX_BIT, shaders.vertex));
             }
             // Pixel stage
-            if let Some(entry) = shaders.pixel {
+            if let Some(entry) = shaders.fragment {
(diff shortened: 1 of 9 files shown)


src/backend/vulkan/src/device.rs
```diff
@@ -340,8 +340,8 @@ impl d::Device<B> for Device {
                 s_type: vk::StructureType::PipelineRasterizationStateCreateInfo,
                 p_next: ptr::null(),
                 flags: vk::PipelineRasterizationStateCreateFlags::empty(),
-                depth_clamp_enable: vk::VK_TRUE, // TODO
-                rasterizer_discard_enable: vk::VK_FALSE, // TODO
+                depth_clamp_enable: if desc.rasterizer.depth_clamping { vk::VK_TRUE } else { vk::VK_FALSE },
+                rasterizer_discard_enable: if shaders.fragment.is_none() { vk::VK_TRUE } else { vk::VK_FALSE },
```
Bastacyclop: Could be nice to have a `bool` to vulkan bool method somewhere.

PR #1496, opened 2017-09-14
[ll] Port dx12ll to dx12

msiglreith:
### Overview
* Port over (most/all) parts of the old dx12ll backend to the dx12 backend
* Remove dx12ll backend
* Implement first parts of hidden descriptor heaps for dx12 using a free-list allocator (deallocation missing)
* Allow descriptor sets to be mixed (srv/cbv/uav and samplers), but will split internally and requires up to 2x more root signature slots.

### TODO
* Pipeline barriers support is pretty basic atm, but it needs more love in general in `core`
* Renderpasses also only ported over and missing large chunks of functionality
* Descriptor pool deallocation
* Some functions haven't been implemented so far (including destruction calls), only trying to get `quad` running atm (and it works when manually specifiying the barriers),


examples/core/quad/main.rs
```diff
@@ -173,15 +211,15 @@ fn main() {
     });
 
     pipeline_desc.attributes.push(pso::AttributeDesc {
-        location: 1,
+        location: 0,
```
kvark: doesn't this make the existing Vk shaders incompatible?

src/backend/dx12/src/command.rs
```diff
@@ -250,9 +351,43 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
+            //       Resize while recording?
+            let mut heaps = [
+                sets[0].heap_srv_cbv_uav.as_mut() as *mut _,
+                sets[0].heap_samplers.as_mut() as *mut _
+            ];
+            self.raw.SetDescriptorHeaps(2, heaps.as_mut_ptr())
+        }
+
+        let mut table_id = 0;
+        for (i, table) in layout.tables.iter().enumerate() {
```
kvark: let's split this into 2 loops, one `for table in &layout.tables[..first_set]`, and another `for (set, _) in sets.iter_mut().zip(layout.tables[first_set..].iter())`

src/backend/dx12/src/device.rs
```diff
@@ -1,82 +1,1144 @@
...
+                String::from_utf8_lossy(slice).into_owned()
+            };
+            return Err(d::ShaderError::CompilationFailed(message))
+        }
+
+        shader_map.insert(entry_point.into(), blob);
+        Ok(n::ShaderModule { shaders: shader_map })
+    }
+
+    pub fn create_descriptor_heap_impl(
```
kvark: nit: no need _impl suffix any more, given no descriptor heaps in the core trait

src/backend/dx12/src/device.rs
```diff
@@ -1,82 +1,1144 @@
...
+            subpasses,
+        }
+    }
+
+    fn create_pipeline_layout(&mut self, sets: &[&n::DescriptorSetLayout]) -> n::PipelineLayout {
+        // Pipeline layouts are implemented as RootSignature for D3D12.
+        //
+        // Each descriptor set layout will be one table entry of the root signature.
+        // We have the additional restriction that SRV/CBV/UAV and samplers need to be
+        // separated, so each set layout will actually occupy 2 entries!
```
kvark: could we have a smarter logic that only splits a set layout if it has both samplers and non-samplers?

src/backend/dx12/src/device.rs
```diff
@@ -1,82 +1,1144 @@
...
+
+            // Get color attachment formats from subpass
+            let (rtvs, num_rtvs) = {
+                let mut rtvs = [winapi::DXGI_FORMAT_UNKNOWN; 8];
+                let mut num_rtvs = 0;
+                for (mut rtv, target) in rtvs.iter_mut()
+                                             .zip(pass.color_attachments.iter())
+                {
+                    let format = subpass.main_pass.attachments[target.0].format;
+                    println!("{:?}", format);
```
kvark: debug print?

src/backend/dx12/src/device.rs
```diff
@@ -1,82 +1,1144 @@
...
+
+        Ok(UnboundBuffer {
+            requirements,
+            stride,
+            usage,
+        })
+    }
+
+    fn get_buffer_requirements(&mut self, _buffer: &UnboundBuffer) -> Requirements {
         unimplemented!()
```
kvark: so lazy! :)

examples/core/quad/main.rs
msiglreith: No, it currently was wrongly implemented, switching uv and pos. Vulkan hasn't used the location but only the order they have been added to the attributes list. I think I included a fix for vulkan in this PR.

src/backend/dx12/src/device.rs
msiglreith: It basically works this way, no empty tables. Comment doesn't reflect it properly (:

src/backend/dx12/src/device.rs
msiglreith: :P

torkleyy:
I'd recommend increasing timeout limit https://github.com/amethyst/amethyst/blob/develop/bors.toml#L20

kvark:
Huge OSX backlog, as always :(

PR #1497, opened 2017-09-14
[ll] More mapping

Bastacyclop:
Modifies the mapping API for `core` as we discussed.
~~Tracking is not yet implemented for `render`.~~
~~Currently using an unpublished version of `ash` for the Vulkan backend.~~
A nice side effect is that the render example now works with `StackAllocator` :)

### DIFF
--- a/examples/core/quad/main.rs
+++ b/examples/core/quad/main.rs
@@ -288,13 +288,11 @@ fn main() {
     };
 
     // TODO: check transitions: read/write mapping and vertex buffer read
-    {
-        let mut vertices = device
-            .acquire_mapping_writer::<Vertex>(&vertex_buffer, 0..buffer_len)
-            .unwrap();
-        vertices.copy_from_slice(&QUAD);
-        device.release_mapping_writer(vertices);
-    }
+    let mut vertices = device
+        .acquire_mapping_writer::<Vertex>(&vertex_buffer, 0..buffer_len)
+        .unwrap();
+    vertices.copy_from_slice(&QUAD);
+    device.release_mapping_writer(vertices);
 
     // Image
     let img_data = include_bytes!("data/logo.png");
@@ -315,17 +313,15 @@ fn main() {
     };
 
     // copy image data into staging buffer
-    {
-        let mut data = device
-            .acquire_mapping_writer::<u8>(&image_upload_buffer, 0..upload_size)
-            .unwrap();
-        for y in 0 .. height as usize {
-            let row = &(*img)[y*(width as usize)*image_stride .. (y+1)*(width as usize)*image_stride];
-            let dest_base = y * row_pitch as usize;
-            data[dest_base .. dest_base + row.len()].copy_from_slice(row);
-        }
-        device.release_mapping_writer(data);
+    let mut data = device
+        .acquire_mapping_writer::<u8>(&image_upload_buffer, 0..upload_size)
+        .unwrap();
+    for y in 0 .. height as usize {
+        let row = &(*img)[y*(width as usize)*image_stride .. (y+1)*(width as usize)*image_stride];
+        let dest_base = y * row_pitch as usize;
+        data[dest_base .. dest_base + row.len()].copy_from_slice(row);
     }
+    device.release_mapping_writer(data);
 
     let image = device.create_image(kind, 1, ColorFormat::get_format(), i::TRANSFER_DST | i::SAMPLED).unwrap(); // TODO: usage
     println!("{:?}", image);
--- a/examples/core/quad/main.rs
+++ b/examples/core/quad/main.rs
@@ -288,9 +288,13 @@ fn main() {
     };
 
     // TODO: check transitions: read/write mapping and vertex buffer read
-    device.write_mapping::<Vertex>(&vertex_buffer, 0..buffer_len)
-        .unwrap()
-        .copy_from_slice(&QUAD);
+    {
+        let mut vertices = device
+            .acquire_mapping_writer::<Vertex>(&vertex_buffer, 0..buffer_len)
+            .unwrap();
+        vertices.copy_from_slice(&QUAD);
+        device.release_mapping_writer(vertices);
+    }
 
     // Image
     let img_data = include_bytes!("data/logo.png");
@@ -311,12 +315,16 @@ fn main() {
     };
 
     // copy image data into staging buffer
-    if let Ok(mut mapping) = device.write_mapping::<u8>(&image_upload_buffer, 0..upload_size) {
+    {
+        let mut data = device
+            .acquire_mapping_writer::<u8>(&image_upload_buffer, 0..upload_size)
+            .unwrap();
         for y in 0 .. height as usize {
             let row = &(*img)[y*(width as usize)*image_stride .. (y+1)*(width as usize)*image_stride];
             let dest_base = y * row_pitch as usize;
-            mapping[dest_base .. dest_base + row.len()].copy_from_slice(row);
+            data[dest_base .. dest_base + row.len()].copy_from_slice(row);
         }
+        device.release_mapping_writer(data);
     }
 
     let image = device.create_image(kind, 1, ColorFormat::get_format(), i::TRANSFER_DST | i::SAMPLED).unwrap(); // TODO: usage
--- a/examples/render/quad_render/main.rs
+++ b/examples/render/quad_render/main.rs
@@ -1,14 +1,7 @@
 extern crate env_logger;
 extern crate gfx;
 extern crate gfx_core as core;
-#[cfg(feature = "dx12")]
-extern crate gfx_device_dx12 as back;
-#[cfg(feature = "vulkan")]
 extern crate gfx_backend_vulkan as back;
-#[cfg(feature = "metal")]
-extern crate gfx_device_metal as back;
-#[cfg(feature = "gl")]
-extern crate gfx_backend_gl as back;
 
 extern crate winit;
 extern crate image;
@@ -20,7 +13,7 @@ use core::format::{Formatted, Srgba8 as ColorFormat, Vec2};
 use core::pass::Subpass;
 use core::queue::Submission;
 use core::target::Rect;
-use gfx::allocators::BoxedAllocator as Allocator;
+use gfx::allocators::StackAllocator as Allocator;
 
 use std::io::Cursor;
 
@@ -41,7 +34,6 @@ const QUAD: [Vertex; 6] = [
     Vertex { a_Pos: [ -0.5,-0.33 ], a_Uv: [0.0, 0.0] },
 ];
 
-#[cfg(any(feature = "vulkan", feature = "dx12", feature = "metal"))]
 fn main() {
     env_logger::init().unwrap();
     let mut events_loop = winit::EventsLoop::new();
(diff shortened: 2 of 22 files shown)


src/render/src/device.rs
```diff
@@ -225,6 +195,73 @@ impl<B: Backend> Device<B> {
...
+        buffer: &'a T,
+        range: Range<u64>,
+    ) -> Result<mapping::Reader<'a, B, T::Data>, mapping::Error>
+        where T: MaybeTypedBuffer<B>
+    {
+        let (resource, info) = buffer.as_raw().resource_info();
+        Ok(mapping::Reader {
+            inner: self.raw.acquire_mapping_reader(
+                resource,
+                range_in_bytes::<T>(range)
```
Bastacyclop: Ups, need `T::Data` here.

Bastacyclop:
Tracking is currently simpler than what we had previously. There is no stall on mapping access, instead the frames involving the buffer need to have been waited for during `acquire_frame`. More flexiblity will probably be needed at some point.

examples/core/quad/main.rs
```diff
@@ -288,9 +288,13 @@ fn main() {
     };
 
     // TODO: check transitions: read/write mapping and vertex buffer read
-    device.write_mapping::<Vertex>(&vertex_buffer, 0..buffer_len)
-        .unwrap()
-        .copy_from_slice(&QUAD);
+    {
+        let mut vertices = device
+            .acquire_mapping_writer::<Vertex>(&vertex_buffer, 0..buffer_len)
```
kvark: I think we can provide better ergonomics here (on top of the current API), something like
```rust
device.map_scope(&vertex_buffer, 0..buffer_len)
    .unwrap()
    .copy_from_slice(&QUAD);
```
So `map_scope` would return a temporary object holding a mutable reference to the `device` and releasing the mapping automatically

I suspect we can get the type from the buffer format, since it's typed?

examples/core/quad/main.rs
```diff
@@ -311,12 +315,16 @@ fn main() {
     };
 
     // copy image data into staging buffer
-    if let Ok(mut mapping) = device.write_mapping::<u8>(&image_upload_buffer, 0..upload_size) {
+    {
+        let mut data = device
```
kvark: same ergonomic improvement could be made here, using
```rust
if let Ok(mut map) = device.map_scope(&image_upload_buffer, 0..upload_size) {
...
}
```

examples/render/quad_render/main.rs
```diff
@@ -71,31 +63,8 @@ fn main() {
     let device: &mut back::Device = render_device.mut_raw();
 
     // Setup renderpass and pipeline
-    // dx12 runtime shader compilation
```
kvark: great :)

examples/render/quad_render/main.rs
```diff
@@ -164,15 +133,15 @@ fn main() {
     });
 
     pipeline_desc.attributes.push(pso::AttributeDesc {
-        location: 1,
+        location: 0,
```
kvark: that is reverting @msiglreith change?

src/backend/dx12/src/device.rs
```diff
@@ -966,37 +963,43 @@ impl d::Device<B> for Device {
...
-            return Err(mapping::Error::OutOfBounds);
-        }
-
-        let range = winapi::D3D12_RANGE {
-            Begin: range.start,
-            End: range.end,
+    fn acquire_mapping_raw(&mut self, buf: &n::Buffer, read: Option<Range<u64>>)
+        -> Result<*mut u8, mapping::Error>
+    {
+        let read_range = if let Some(read) = read {
```
kvark: it's a matter of preference, but I somewhat like `match` more than `if let xxx=yyy {...} else {..}` because the latter is more sugar for nothing in return

src/backend/gl/src/device.rs
```diff
@@ -381,19 +368,13 @@ impl d::Device<B> for Device {
...
-    fn write_mapping_raw(&mut self, _: &n::Buffer, _: Range<u64>)
-        -> Result<(*mut u8, Mapping), mapping::Error>
+    fn acquire_mapping_raw(&mut self, buf: &n::Buffer, read: Option<Range<u64>>)
+        -> Result<*mut u8, mapping::Error>
     {
         unimplemented!()
     }
 
-    fn unmap_mapping_raw(&mut self, _: Mapping) {
+    fn release_mapping_raw(&mut self, buf: &n::Buffer, wrote: Option<Range<u64>>) {
```
kvark: I'd be curious if GL fits the current design. Having some rough implementation here would be great as a proof of concept.

src/backend/vulkan/Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "gfx_backend_vulkan"
 log = "0.3"
 lazy_static = "0.2"
 shared_library = "0.1"
-ash = "0.18.4"
+ash = "0.18.5"
```
kvark: nice 👍

src/core/src/device.rs
```diff
@@ -312,53 +312,80 @@ pub trait Device<B: Backend>: Clone {
...
         let count = (range.end - range.start) as usize / mem::size_of::<T>();
-        self.read_mapping_raw(buf, range)
-            .map(|(ptr, mapping)| mapping::Reader {
-                slice: unsafe { slice::from_raw_parts(ptr as *const _, count) },
-                _mapping: mapping,
+        self.acquire_mapping_raw(buffer, Some(range.clone()))
+            .map(|ptr| {
+                let start_ptr = unsafe { ptr.offset(range.start as isize) } as *const _;
+                mapping::Reader {
+                    slice: unsafe { slice::from_raw_parts(start_ptr, count) },
```
kvark: nit: let's combine these 2 unsafe blocks into one

src/core/src/device.rs
```diff
@@ -312,53 +312,80 @@ pub trait Device<B: Backend>: Clone {
...
         let count = (range.end - range.start) as usize / mem::size_of::<T>();
-        self.write_mapping_raw(buf, range)
-            .map(|(ptr, mapping)| mapping::Writer {
-                slice: unsafe { slice::from_raw_parts_mut(ptr as *mut _, count) },
-                _mapping: mapping,
+        self.acquire_mapping_raw(buffer, None)
+            .map(|ptr| {
+                let start_ptr = unsafe { ptr.offset(range.start as isize) } as *mut _;
+                mapping::Writer {
+                    slice: unsafe { slice::from_raw_parts_mut(start_ptr, count) },
```
kvark: same here

src/core/src/mapping.rs
```diff
@@ -40,7 +41,14 @@ impl StdError for Error {
...
 pub struct Reader<'a, B: Backend, T: 'a + Copy> {
     pub(crate) slice: &'a [T],
-    pub(crate) _mapping: B::Mapping,
+    pub(crate) buffer: &'a B::Buffer,
+    pub(crate) released: bool,
+}
+
+impl<'a, B: Backend, T: 'a + Copy> Drop for Reader<'a, B, T> {
+    fn drop(&mut self) {
+        if !self.released { panic!("a mapping reader was not released"); }
```
kvark: wouldn't it be more straightforward to just use `assert!(self.released, "a mapping reader was not released")`?

src/core/src/mapping.rs
```diff
@@ -54,7 +62,15 @@ impl<'a, B: Backend, T: 'a + Copy> ops::Deref for Reader<'a, B, T> {
...
     pub(crate) slice: &'a mut [T],
-    pub(crate) _mapping: B::Mapping,
+    pub(crate) buffer: &'a B::Buffer,
+    pub(crate) range: Range<u64>,
+    pub(crate) released: bool,
+}
+
+impl<'a, B: Backend, T: 'a + Copy> Drop for Writer<'a, B, T> {
+    fn drop(&mut self) {
+        if !self.released { panic!("a mapping writer was not released"); }
```
kvark: same here

src/core/src/mapping.rs
```diff
@@ -54,7 +62,15 @@ impl<'a, B: Backend, T: 'a + Copy> ops::Deref for Reader<'a, B, T> {
 /// to read from Writer, it will lead to an undefined behavior. Please do not read from it.
 pub struct Writer<'a, B: Backend, T: 'a + Copy> {
```
kvark: I don't think having `Copy` bound on the struct here does anything for us

src/render/src/buffer.rs
```diff
@@ -17,5 +19,64 @@ pub struct Info {
...
+            cpu: AtomicBool::new(false),
+            gpu: AtomicUsize::new(0),
+        };
+        Info { usage, memory, size, stride, access }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct Access {
+    cpu: AtomicBool,
```
kvark: I wonder if we could use `Arc<()>` for that or something, to reduce the atomic logic

examples/core/quad/main.rs
Bastacyclop: The scope could be nice indeed, I was more thinking about a closure but I think that error handling might be less convenient with a closure.
This is the `core` example so there is no typed handle here.

examples/render/quad_render/main.rs
Bastacyclop: Should be actually applying it on the `quad_render` example copy.

src/core/src/mapping.rs
Bastacyclop: I always forget that `assert!` can take messages :)

src/render/src/buffer.rs
Bastacyclop: `AtomicBool` can probably be replaced by `Mutex<()>` but then you have to drag the `MutexGuard`s everywhere and the lifetimes will probably not be manageable. Not sure how you would use `Arc<()>`.

examples/render/quad_render/main.rs
```diff
@@ -231,14 +228,13 @@ fn main() {
 
     {
         let mut data = context.mut_device()
-            .acquire_mapping_writer(&image_upload_buffer, 0..upload_size)
+            .write_mapping(&image_upload_buffer, 0..upload_size)
```
kvark: I believe having the whole block in `if let Ok(mapping) = device.write_mapping(...)` looks much nicer

src/render/src/device.rs
```diff
@@ -262,13 +260,39 @@ impl<B: Backend> Device<B> {
...
-    )
-        where T: Copy
-    {
+    ) {
         self.raw.release_mapping_writer(writer.inner);
         writer.info.access.release_exclusive();
     }
 
+    /// Sugar to acquire and release a mapping reader.
+    pub fn read_mapping<'a, MTB>(&'a mut self,
```
kvark: nit: I think `self` parameter should also start next line, not here

kvark:
a few more nits, if you will ;)

examples/render/quad_render/main.rs
Bastacyclop: It won't crash if there is an error though, but maybe that's what you want ? Otherwise I don't think the block is necessary.

kvark: yeah, it wouldn't, so it's not much better, just looks somewhat nicer

kvark:
bors r+

Why would we need to batch mapping if even Bulkan doesn't?

kvark:
Doesn't look like my online merge fix worked. @Bastacyclop please rebase and overwrite my commit.

msiglreith:
@kvark Vulkan does allow batching of flushes and invalidation calls (would safe us a few internal barriers on some platforms). We basically follow dx12's model and couple mapping (actual mapping done internally I assume) and flush/invalidation. So we basically 'remap' everytime we want to flush data for reading. Tight ranges are interesting for mobile hardware (as suggested by the d3d12 debug layer when mapping the whole buffer).

kvark:
Interesting info, thanks @msiglreith!
So you propose to release multiple mappings at once? Sounds good to me.

PR #1501, opened 2017-09-16
[ll] dx12: Compute pipelines and cleanup

msiglreith:
* Fix build issue in `quad` for dx12 backend
* Improve documentation for root signature layout. Better reflect internal layout.
* Implement compute pipeline creation
* Increase default descriptor heap size for CBV/SRV/UAV heaps to match Tier 1 limit
* Implement a few more destroy functions
* Only enable debug layer for debug builds following Vulkan


kvark:
Please feel free to force merge

PR #1502, opened 2017-09-17
[ll][WIP] Encoding and state transitions

Bastacyclop:
Working on #1466.
I'm not familiar with the concepts yet so tell me if I'm missing something.


examples/render/quad_render/main.rs
```diff
@@ -280,56 +285,30 @@ fn main() {
...
+            buffer_offset: 0,
+            buffer_row_pitch: row_pitch,
+            buffer_slice_pitch: row_pitch * (height as u32),
+            image_aspect: i::ASPECT_COLOR,
+            image_subresource: (0, 0..1),
+            image_offset: command::Offset { x: 0, y: 0, z: 0 },
+            image_extent: d::Extent { width, height, depth: 1 },
+        }]);
+
+    let init_submit = init_encoder.finish();
```
kvark: That's what I'm worried about: passing the encoder in the device creation doesn't make it safe: user is still responsible to actually submit it. So I think we should change to tokens right away:
  - resource creation returns a token
  - encoder can do `init_resources(&[Token])`, which consumes the tokens.

This is not safer but I believe is more universal.

src/render/src/buffer.rs
```diff
@@ -80,3 +82,24 @@ impl Access {
...
+    fn as_raw(&self) -> &handle::raw::Buffer<B> { &self }
+}
+
+impl<B: Backend, T: Pod> MaybeTyped<B> for handle::Buffer<B, T> {
+    type Data = T;
+    fn as_raw(&self) -> &handle::raw::Buffer<B> { &self }
+}
+
+impl<B: Backend> handle::raw::Buffer<B> {
+    pub fn stable_state(&self) -> core::buffer::State {
```
kvark: `pub(crate)` ?

src/render/src/device.rs
```diff
@@ -301,33 +297,41 @@ impl<B: Backend> Device<B> {
...
         allocator: &mut A,
+        encoder: &mut Encoder<B, C>,
         usage: image::Usage,
         kind: image::Kind,
         mip_levels: image::Level,
         format: format::Format
     ) -> Result<handle::raw::Image<B>, image::CreationError>
-        where A: Allocator<B>
+        where A: Allocator<B>,
+              C: core::queue::Supports<core::Transfer>
```
kvark: uh, yeah, that certainly looks out of place here. Device shouldn't know anything about queues and their capabilities

src/render/src/encoder.rs
```diff
@@ -84,6 +90,41 @@ pub struct Submit<B: Backend, C> {
     pub(crate) pool: PoolDependency<B, C>
 }
 
+// TODO: coalescing?
+struct ImageStates {
+    states: Vec<core::image::State>,
```
kvark: We could probably just add the level/slice into the hashmap key instead of introducing this struct.
Do you think the struct gives us an advantage?

src/render/src/encoder.rs
```diff
@@ -150,356 +172,373 @@ impl<B: Backend> AccessInfo<B> {
...
-            });
+    pub fn require_state(
+        &mut self,
+        stage: core::pso::PipelineStage,
+        buffer_states: &[(&handle::raw::Buffer<B>, core::buffer::State)],
+        image_states: &[(&handle::raw::Image<B>, image::Subresource, core::image::State)],
+    ) {
+        let mut barriers = Vec::new();
+        for &(buffer, state) in buffer_states {
+            self.require_buffer_state(buffer, state)
```
kvark: nit: could be `barriers.extend(self.require_buffer_state(...));`

src/render/src/encoder.rs
```diff
@@ -150,356 +172,373 @@ impl<B: Backend> AccessInfo<B> {
...
-        if !info.is_inside(dim) {
-            let (w, h, d, _) = dim;
-            return Err(CopyError::OutOfDstBounds {
-                size: [w, h, d],
-                copy_end: [info.xoffset + info.width,
-                           info.yoffset + info.height,
-                           info.zoffset + info.depth]
-            });
+        for &(image, (level, layer), state) in image_states {
+            self.require_image_state(image, level, layer, state)
```
kvark: same here

src/render/src/encoder.rs
```diff
@@ -150,356 +172,373 @@ impl<B: Backend> AccessInfo<B> {
...
+                .map(|b| barriers.push(b));
+        }
+        let current_stage = mem::replace(&mut self.pipeline_stage, stage);
+        if (current_stage != stage) || !barriers.is_empty() {
+            println!("barrier: {:?}, {:?}", current_stage..stage, &barriers[..]);
```
kvark: please change to `debug!`

src/render/src/image.rs
```diff
@@ -21,26 +23,26 @@ pub struct Info {
...
-            format: (),
-            mipmap: mip,
-        }
-    }
 
-    /// Get the raw image info for a given mip.
-    pub fn to_raw_image_info(&self, mip: Level) -> RawImageInfo {
-        self.to_image_info(mip).convert(self.format)
+impl<B: Backend> handle::raw::Image<B> {
+    pub fn stable_state(&self) -> core::image::State {
```
kvark: I think we should store this as a constant instead of recomputing

src/render/src/encoder.rs
Bastacyclop: I was wondering if it would be worth it to coalesce the state slices instead of always having a state for each level/layer. Otherwise, I don't see any advantage.

examples/render/quad_render/main.rs
Bastacyclop: That's also better for batching into a single barrier command :)

src/render/src/encoder.rs
Bastacyclop: So you will want to keep it ?

src/render/src/image.rs
Bastacyclop: A constant like a field in `Info`, or something more global ?

src/render/src/encoder.rs
kvark: Ok, let's have it this way for now, will reconsider later.

src/render/src/encoder.rs
kvark: removing works too, I don't mind either way

src/render/src/image.rs
kvark: yep, a field in `Info` would work

examples/render/quad_render/main.rs
```diff
@@ -36,6 +37,13 @@ const QUAD: [Vertex; 6] = [
     Vertex { a_Pos: [ -0.5,-0.33 ], a_Uv: [0.0, 0.0] },
 ];
 
+gfx_descriptor_struct! {
```
kvark: yammy!

examples/render/quad_render/main.rs
```diff
@@ -68,27 +76,10 @@ fn main() {
...
-
-    let set1_layout = device.create_descriptor_set_layout(&[
-            pso::DescriptorSetLayoutBinding {
-                binding: 0,
-                ty: pso::DescriptorType::Sampler,
-                count: 1,
-                stage_flags: pso::STAGE_FRAGMENT,
-            }
-        ],
+    gfx_allocate_descriptor_structs!(context.mut_device();
```
kvark: hmm, not sure why this has to be a macro?

src/render/src/device.rs
```diff
@@ -492,6 +492,51 @@ impl<B: Backend> Device<B> {
             .map(Typed::new)
     }
 
+    #[doc(hidden)]
+    pub fn allocate_descriptor_sets(
+        &mut self,
+        sets_bindings: &[&[core::pso::DescriptorSetLayoutBinding]],
```
kvark: why is this a 2d array?

Bastacyclop: You can allocate multiple sets in a single batch, allowing to share the underlying descriptor pool.

examples/render/quad_render/main.rs
Bastacyclop: Since you allocate a (potential) batch of descriptor structs, the result can consist of different types and cannot be a vector (unless dynamically typed). We could probably have some tricky tuple imbrication using a builder but it seems worse to use.

Bastacyclop: We could also come up with another descriptor allocation API, obviously. It seemed like an easy way to avoid explicit pools while still allowing some batching.

examples/render/quad_render/main.rs
```diff
@@ -66,124 +81,46 @@ fn main() {
...
-        let frame_rtv = backbuffer.color.resource();
+    let frame_rtvs = backbuffers.iter().map(|backbuffer| {
+        context.mut_device()
+            .view_image_as_render_target(&backbuffer.color, (0..1, 0..1))
+            .unwrap()
+    }).collect::<Vec<_>>();
+    let framebuffers = frame_rtvs.iter().map(|rtv| {
         let extent = d::Extent { width: pixel_width as _, height: pixel_height as _, depth: 1 };
-        device.create_framebuffer(&render_pass, &[frame_rtv], &[], extent)
+        device.create_framebuffer(pipeline.render_pass(), &[rtv.resource()], &[], extent)
```
msiglreith: I don't know if you have planned it but how difficult would you estimate to add custom renderpass support?

examples/render/quad_render/main.rs
```diff
@@ -254,20 +195,10 @@ fn main() {
...
-        },
-        pso::DescriptorSetWrite {
-            set: &set1[0],
-            binding: 0,
-            array_offset: 0,
-            write: pso::DescriptorWrite::Sampler(vec![sampler.resource()]),
-        },
-    ]);
+    context.mut_device().update_descriptor_sets()
+        .write(sample_desc.sampled_image(), 0, vec![&image_srv])
```
msiglreith: Why do we need `vec![..]` here?

examples/render/quad_render/main.rs
```diff
@@ -349,50 +257,38 @@ fn main() {
...
-
-            }
-            encoder.finish()
-        };
-
-        context.present(vec![submit]);
+        }
+        
+        submits.push(encoder.finish());
+        context.present(mem::replace(&mut submits, Vec::new()));
```
msiglreith: hm, do we have alternatives here? `drain(..)` or sth like this?

examples/render/quad_render/main.rs
```diff
@@ -349,50 +257,38 @@ fn main() {
...
     }
 
     println!("cleanup!");
-    device.destroy_descriptor_pool(srv_pool);
```
msiglreith: praise the garbage collector \o/

src/render/src/macros.rs
```diff
@@ -0,0 +1,214 @@
...
+                                });
+                                color_attachments.push((attach_id, attach.required_layout));
+                            }
+                        )*
+                        let subpass = cpass::SubpassDesc {
+                            color_attachments: &color_attachments[..],
+                        };
+
+                        // TODO:
+                        let dependency = cpass::SubpassDependency {
```
msiglreith: Vulkan's spec (and probably we also) implicitly add subpass dependencies from EXTERNAl -> firstSubpass and lastSubpass -> EXTERNAL, which should cover your use-case also. So this dependencies shouldn't be needed.

src/render/src/pso/mod.rs
```diff
@@ -1,334 +1,284 @@
...
-        match *self {
-            ElementError::NotFound(ref s) => write!(f, "{}: {:?}", self.description(), s),
-            ElementError::Offset{ ref name, ref shader_offset, ref code_offset } =>
-                write!(f, "{}: ({:?}, {:?}, {:?})", self.description(), name, shader_offset, code_offset),
-            ElementError::Format{ ref name, ref shader_format, ref code_format } =>
-                write!(f, "{}: ({:?}, {:?}, {:?})", self.description(), name, shader_format, code_format),
-        }
-    }
+pub trait Bind {
+    fn desc_type() -> core::pso::DescriptorType;
```
kvark: associated constant?

examples/render/quad_render/main.rs
Bastacyclop: I didn't think about it yet, I'm focusing on getting simple pipelines working for this PR.

src/render/src/macros.rs
Bastacyclop: nice!

examples/render/quad_render/main.rs
```diff
@@ -68,118 +81,36 @@ fn main() {
...
-        rate: 0,
-    });
-
-    pipeline_desc.attributes.push(pso::AttributeDesc {
-        location: 0,
-        binding: 0,
-        element: pso::Element {
-            format: <Vec2<f32> as Formatted>::get_format(),
-            offset: 0,
+    let sample_desc = context.mut_device().create_descriptors(1).pop().unwrap();
```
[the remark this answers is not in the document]
Bastacyclop: We could do `let mut device = context.ref_device().clone();` at the beginning.

examples/render/quad_render/main.rs
```diff
@@ -68,118 +81,36 @@ fn main() {
...
+    let pipe_init = pipe::Init {
+        sample: &sample_desc,
+        color: pso::ColorInfo {
```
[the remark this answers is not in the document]
Bastacyclop: No reason, I simply kept the original code here.

examples/render/quad_render/main.rs
```diff
@@ -344,15 +265,15 @@ fn main() {
 
             cmd_buffer.set_viewports(&[viewport]);
             cmd_buffer.set_scissors(&[scissor]);
-            cmd_buffer.bind_graphics_pipeline(&pipelines[0].as_ref().unwrap());
+            cmd_buffer.bind_graphics_pipeline(pipeline.pipeline());
```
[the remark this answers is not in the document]
Bastacyclop: Yeah that should not stay once `encoder.draw` is done :p

src/render/src/device.rs
```diff
@@ -492,6 +422,96 @@ impl<B: Backend> Device<B> {
             .map(Typed::new)
     }
 
+    // TODO: smarter allocation
+    pub fn create_descriptors<D>(&mut self, count: usize) -> Vec<D>
+        where D: pso::Descriptors<B>
+    {
+        use core::DescriptorPool as CDP;
+
+        let bindings = &D::layout_bindings()[..];
```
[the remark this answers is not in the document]
Bastacyclop: Problem is I can't make a `&'static` slice without macro tricks to get the binding index (maybe easier with a custom derive ?).

examples/render/quad_render/main.rs
```diff
@@ -34,16 +33,16 @@ const QUAD: [Vertex; 6] = [
...
+    desc {
         sampled_image: gfx::pso::SampledImage,
         sampler: gfx::pso::Sampler,
     }
 }
 
 gfx_graphics_pipeline! {
     pipe {
-        sample: gfx::pso::DescriptorSet<SampleDesc<B>>, // TODO remove <B>
+        desc: gfx::pso::DescriptorSet<desc::Set<B>>, // TODO: improve
```
kvark: we still do need to get rid of `B` here since no other PSO component is backend dependent

src/render/src/device.rs
```diff
@@ -515,6 +514,28 @@ impl<B: Backend> Device<B> {
...
+        extent: Extent,
+    ) -> handle::raw::FrameBuffer<B>
+        where P: pso::GraphicsPipelineMeta<B>
+    {
+        let rtv_res: Vec<_> = rtvs.iter().map(|&rtv| rtv.resource()).collect();
+        let dsv_res: Vec<_> = dsvs.iter().map(|&dsv| dsv.resource()).collect();
+        let buffer = self.raw.create_framebuffer(
+            pipeline.render_pass(), &rtv_res[..], &dsv_res[..], extent);
+        let info = handle::FrameBufferInfo {
+            rtvs: rtvs.iter().map(|&rtv| rtv.clone()).collect(),
```
kvark: could just have used `rtvs.iter().cloned().collect()` ?

src/render/src/encoder.rs
```diff
@@ -593,50 +595,19 @@ impl<'a, B: Backend, C> Encoder<'a, B, C>
...
-        for &(location, value) in &self.raw_pso_data.global_constants {
-            self.command_buffer.bind_global_constant(location, value);
-        }
-        self.command_buffer.bind_unordered_views(&self.raw_pso_data.unordered_views);
-        //Note: it's important to bind RTV, DSV, and UAV before SRV
-        self.command_buffer.bind_resource_views(&self.raw_pso_data.resource_views);
-        self.command_buffer.bind_samplers(&self.raw_pso_data.samplers);
-        self.draw_slice(slice, slice.instances);
+        // TODO: instances
+        data.begin_renderpass(self, pipeline).draw(vertices, 0..1);
```
kvark: starting and ending a render pas for each draw call is not perfect.
I think one way we cold make this nicer is to have encoders behave in the same way they do in Metal (it is a nice API, no matter what political negativity is around it):
  - have command buffers visible to the user
  - user creates a graphics encoder, which starts a render pass given the targets, and borrows the command buffer mutably
  - user submits a bunch of gaphics/draw commands
  - user drops the encoder, which ends the render pass and releases the command buffer borrow, which can now be submitted

cc @msiglreith

src/render/src/handle.rs
```diff
@@ -230,7 +231,7 @@ define_resources! {
     PipelineLayout: (),
     GraphicsPipeline: (),
     // ComputePipeline
-    // FrameBuffer
+    FrameBuffer: ::handle::FrameBufferInfo<B>,
```
kvark: nit: would be nicer to not use global imports

src/render/src/macros.rs
```diff
@@ -45,59 +45,89 @@ macro_rules! gfx_buffer_struct {
...
+                type Data = Data<B>;
+
+                fn from_raw(
+                    layout: handle::raw::DescriptorSetLayout<B>,
+                    raw: pso::RawDescriptorSet<B>
+                ) -> (Self, Self::Data) {
+                    let mut binding = 0;
+                    let mut next_binding = || {let b = binding; binding += 1; b };
+                    (Set {
+                        $( $field: next_binding(), )*
```
kvark: nit: heh, not sure what we are getting by moving `next_binding()` up as opposed to just inlining it here

src/render/src/pso/mod.rs
```diff
@@ -21,86 +21,117 @@ impl<B: Backend> RawDescriptorSet<B> {
...
+        &mut Vec<(&'a handle::raw::Buffer<B>, core::buffer::State)>,
+        &mut Vec<(&'a handle::raw::Image<B>, image::Subresource, core::image::State)>
+    );
 }
 
-pub trait Bind {
-    fn desc_type() -> core::pso::DescriptorType;
-    fn desc_count() -> usize;
+pub trait BindDesc {
+    const TYPE: core::pso::DescriptorType;
```
kvark: niiice!

src/render/src/encoder.rs
msiglreith: I would be totally happy if we can get a Metal-ish API, but I also see why you went with the renderpass for each pipeline approach. Sounds like a hard problem due to required compatibility between framebuffer, pipeline and renderpass.
We might have to dynamically compile renderpasses (with single renderpass) and pipelines here to get this working or require the user to do it upfront or a mix of the both approaches.

kvark: The way I see it:
  - user declares a render pass using a macro, like the other things. The information required for the macro is exactly what Vulkan considers to **not** be a part of render pass compatibility.
  - the "compatible" pieces would then be supplied at run-time during render pass creation by the user
  - the pipeline definition would then point to a render pass definition, thus statically enforcing which pipelines are going to be accepted in the graphics encoder corresponding to the given render pass definition (not the instance)

Does it make sense?

msiglreith: Sounds reasonable even though I'm unsure if it's that easy to split the renderpass definition in compatible and non-compatible parts..

examples/render/quad_render/main.rs
Bastacyclop: The macro could generate a `desc::Component`, how does that sound ?

src/render/src/device.rs
Bastacyclop: That would require `cloned().cloned()` because this is a `&&RTV`.

src/render/src/encoder.rs
Bastacyclop: It would be nice but I think it deserves a separate PR.

src/render/src/macros.rs
Bastacyclop: we avoid getting a warning for the last unused `+=` operation.

src/render/src/encoder.rs
```diff
@@ -150,456 +178,436 @@ impl<B: Backend> AccessInfo<B> {
...
-            return Err(CopyError::Overlap {
-                src_offset: src_offset_bytes,
-                dst_offset: dst_offset_bytes,
-                size: size_bytes,
-            });
+    fn init_image<'b>(&mut self, image: &'b handle::raw::Image<B>)
+        -> core::memory::Barrier<'b, B>
+    {
+        let creation_state = (core::image::Access::empty(), ImageLayout::Undefined);
+        let levels = image.info().kind.get_num_levels();
```
Bastacyclop: Looks like this does not give the actual image levels, should I use `image.info().mip_levels` instead ?.

src/render/src/encoder.rs
kvark: that's fair

src/render/src/encoder.rs
kvark: I'd expect them to match, plus minus the differences between 0 and 1

Bastacyclop: The way I see it, `kind.get_num_levels()` is how many mip levels you should have if you had mip mapping while `info().mip_levels` is how many mip levels you really have (so it is valid even if you do not have mip mapping).

kvark: I think you just need to remove `Info::mip_levels` completely. Image `Kind` is the full description of what an image has in terms of dimensions/levels.

Bastacyclop: Do we agree that:

1. You can have images with arbitrary mip levels (1 most likely, like for backbuffers).
2. `kind.get_num_levels()` will not return this number, but rather `(1..).find(|level| dominant>>level <= 1).unwrap()`, which is like the maximum number of mip levels you can have.

And that we need the actual level count and not the maximum level count ?

kvark: @Bastacyclop Sorry, I got confused. Used the `get_num_layers` lately and assumed `get_num_levels` does a similar thing. You are correct, it doesn't return the actually stored mipmap count, this method can only be used for estimation during the image creation, and is rather pointless afterwards.

So in your case using `image.info().mip_levels` is the way to proceed.

examples/render/quad_render/main.rs
```diff
@@ -58,134 +71,52 @@ fn main() {
...
-            offset: 8
+    let vs_module = device.mut_raw().create_shader_module(include_bytes!("data/vs_main.spv")).unwrap();
+    let fs_module = device.mut_raw().create_shader_module(include_bytes!("data/ps_main.spv")).unwrap();
+
+    let (desc, mut desc_data) = device.create_descriptors(1).pop().unwrap();
+    let pipe_init = pipe::Init {
+        desc: &desc,
+        color: pso::ColorInfo {
+            mask: state::MASK_ALL,
+            color: Some(state::BlendChannel {
```
kvark: nit: let's use the presets. We can do as a follow-up

examples/render/quad_render/main.rs
```diff
@@ -58,134 +71,52 @@ fn main() {
...
-    });
-
-    //
-    let pipelines = {
-        let shader_entries = pso::GraphicsShaderSet {
+        vertices: (),
+    };
+    let pipeline = device.create_graphics_pipeline(
+        pso::GraphicsShaderSet {
             vertex: pso::EntryPoint { entry: "main", module: &vs_module },
```
kvark: you know what would be **really** cool? to write:
```rust
vertex: vs_module["main"],
```

examples/render/quad_render/main.rs
```diff
@@ -197,39 +128,40 @@ fn main() {
...
+    let (vertex_buffer, token) = device.create_buffer::<Vertex, _>(
         &mut upload,
         gfx::buffer::VERTEX,
         vertex_count
     ).unwrap();
+    init_tokens.push(token);
 
-    context.mut_device()
-        .write_mapping(&vertex_buffer, 0..vertex_count)
+    device.write_mapping(&vertex_buffer, 0..vertex_count)
```
kvark: I wonder if you can actually do anything with that buffer, including the mapping, before it's initialized. E.g. doesn't Vulkan require the HOST access to be set?

examples/render/quad_render/main.rs
```diff
@@ -197,39 +128,40 @@ fn main() {
...
         image_stride as u64
     ).unwrap();
+    init_tokens.push(token);
 
     println!("copy image data into staging buffer");
 
-    if let Ok(mut image_data) = context.mut_device()
-        .write_mapping(&image_upload_buffer, 0..upload_size)
+    if let Ok(mut image_data) = device.write_mapping(&image_upload_buffer, 0..upload_size)
     {
```
kvark: nit: move to the end of the previous line

examples/render/quad_render/main.rs
```diff
@@ -238,38 +170,28 @@ fn main() {
...
-        },
-        pso::DescriptorSetWrite {
-            set: &set1[0],
-            binding: 0,
-            array_offset: 0,
-            write: pso::DescriptorWrite::Sampler(vec![sampler.resource()]),
-        },
-    ]);
+    device.update_descriptor_sets()
+        .write(desc_data.sampled_image(&desc), 0, &[&image_srv as _])
```
kvark: I wonder if we can get rid of those `X as _` things?

examples/render/quad_render/main.rs
```diff
@@ -282,56 +204,24 @@ fn main() {
...
-            .submit(&[submit]);
-        context.mut_queue().submit(submission, Some(&mut fence));
-
-        device.wait_for_fences(&[&fence], d::WaitFor::All, !0);
-    }
-
-    device.destroy_fence(fence);
+    let mut encoder_pool = context.acquire_encoder_pool();
+    let mut init_encoder = encoder_pool.acquire_encoder();
+    init_encoder.init_resources(init_tokens);
```
kvark: perhaps, this could just put everything inline:
```rust
init_encoder.init_resources(vec![
    vertex_token,
    image_token,
    ...
]);
```

src/backend/dx12/src/device.rs
```diff
@@ -726,7 +726,7 @@ impl d::Device<B> for Device {
         color_attachments: &[&n::RenderTargetView],
         depth_stencil_attachments: &[&n::DepthStencilView],
         _extent: d::Extent,
-    ) -> Result<n::FrameBuffer, d::FramebufferError> {
+    ) -> Result<n::FrameBuffer, d::FrameBufferError> {
```
kvark: Let's keep it the old way. I believe saying `Framebuffer` and `Renderpass` consistently across core/backends would be nicer

src/backend/gl/src/native.rs
```diff
@@ -96,6 +96,8 @@ pub struct DescriptorSetLayout;
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
 pub struct DescriptorSet;
 
+#[allow(missing_copy_implementations)]
```
kvark: I find this one to be the most annoying warning of all. I'd prefer to just put it as a top-level thing. In fact, it **is already**: https://github.com/gfx-rs/gfx/blob/ad34a1eb2590b6fed48c0d8889ef104b490f1eb3/src/backend/gl/src/lib.rs#L4

src/backend/metal/src/device.rs
```diff
@@ -512,7 +512,7 @@ impl core::Device<Backend> for Device {
             let base_ptr = buf.0.contents() as *mut u8;
 
             if base_ptr.is_null() {
-                return Err(mapping::Error::InvalidAccess);
+                return Err(unimplemented!());
```
kvark: why `unimplemented`?

src/core/src/device.rs
```diff
@@ -289,7 +289,11 @@ pub trait Device<B: Backend>: Clone {
     ///
     fn view_image_as_render_target(&mut self, &B::Image, format::Format, image::SubresourceLayers) -> Result<B::RenderTargetView, TargetViewError>;
 
-    // TODO: view_image_as_depth_stencil
+    ///
+    #[allow(unused_variables)]
+    fn view_image_as_depth_stencil(&mut self, image: &B::Image, format: format::Format, range: image::SubresourceRange) -> Result<B::DepthStencilView, TargetViewError> {
```
kvark: should use `SubresourceLayers`

src/render/src/device.rs
```diff
@@ -311,13 +244,38 @@ impl<B: Backend> Device<B> {
...
+        use core::image::ImageLayout;
+
         let image = self.raw.create_image(kind, mip_levels, format, usage)?;
         let (image, memory) = allocator.allocate_image(self, usage, image);
-        let info = image::Info { usage, kind, mip_levels, format, memory };
-        Ok(Image::new(image, info, self.garbage.clone()).into())
+        let origin = image::Origin::User(memory);
+        let stable_access = core::image::Access::empty();
+        let stable_layout = match usage {
+            _ if usage.contains(COLOR_ATTACHMENT) =>
```
kvark: interesting matching

src/backend/dx12/src/device.rs
Bastacyclop: They are named `FrameBuffer` and `RenderPass` in the backend trait though.

src/backend/metal/src/device.rs
Bastacyclop: We had removed `core::mapping::Error::InvalidAccess` because it was originally for render, but it seems it came back with another meaning. Will put it back.

examples/render/quad_render/main.rs
Bastacyclop: Unfortunately type inference does not seem to go through without it.

src/backend/dx12/src/device.rs
kvark: Let's rename the backend trait types then. I believe we did discuss this naming issue recently and agreed upon the Vulkan-like naming, right @msiglreith ?

kvark: Note: we'll rename it outside of this PR. It's already big enough

kvark:
@Bastacyclop Here is one thing I think we should change w.r.t. typing.

Currently, `Typed` trait implements `Deref`/`DerefMut` for the wrapped type. I believe this is semantically incorrect. You only need `deref` when you call methods on the wrapped object. In our case there are no methods - the contained stuff is just opaque resource handles.

The way you use `Deref` is to be able to pass it down to functions that don't care about `Typed`, hence you sometimes get type inference failing - it's not supposed to use `Deref` there.

Instead, let's have `Typed` implementing `AsRef`, and have all the accepting methods be generic over `R: AsRef<T>`, calling `as_ref()` in the body explicitly. This will be more ergonomic for the user (less `X as _` casting, better error messages, etc). Also, perhaps this will eliminate `buffer::MaybeTyped` trait.

Bastacyclop:
`AsRef` sounds good, but it won't eliminate `MaybeTyped` because the latter is mainly used to say that raw operations are expecting `u8` data.

kvark:
@Bastacyclop ok, so it could be defined as
```rust
pub trait MaybeTyped<B: Backend>: AsRef<handle::raw::Buffer<B>> {
    type Data: Pod;
}
```

kvark:
Ok, the branch got accidentally pushed to `ll` by mistake, and that branch was not protected (it is now!).

PR #1504, opened 2017-09-18
[ll] Metal core port

kvark:
Based on the code by @JohnColanduoni . The problem with original code was that it tried to re-use the command buffers already used for submission. Say thanks to @jrmuizel for investigating this ;)

The example either gets stuck or not releasing the resources properly, but it renders 🎆 . More work to follow, hopefully.


src/backend/metal/src/command.rs
```diff
@@ -1,477 +1,634 @@
...
+    fn bind_graphics_descriptor_sets(
+        &mut self,
+        layout: &native::PipelineLayout,
+        first_set: usize,
+        sets: &[&native::DescriptorSet],
+    ) {
+        while self.inner().descriptor_sets.len() < first_set + sets.len() {
+            self.inner().descriptor_sets.push(None);
+        }
+        for (out, &set) in self.inner().descriptor_sets[first_set ..].iter_mut().zip(sets) {
```
msiglreith: nit: Should be written with iterators only.

src/backend/metal/src/command.rs
```diff
@@ -1,477 +1,634 @@
...
+            height: dst.0.height(),
+            depth: dst.0.depth(),
+        }};
+        // FIXME: layout
+
+        for region in regions {
+            let image_offset = &region.image_offset;
+
+            // TODO multiple layers
+            assert!((region.image_subresource.1).len() == 1, "multiple layer copies not implemented");
```
msiglreith: We can't loop over all layers?

src/backend/metal/src/command.rs
```diff
@@ -1,477 +1,634 @@
...
+        dst: &native::Buffer,
+        regions: &[BufferImageCopy],
+    ) {
+        unimplemented!()
+    }
+    
+    fn draw(
+        &mut self,
+        vertices: Range<VertexCount>,
+        instances: Range<InstanceCount>,
```
msiglreith: instancing is not supported for metal?

src/backend/vulkan/src/window.rs
```diff
@@ -44,7 +44,7 @@ impl Instance {
             .iter()
             .map(|&extension| {
                 match extension {
-                    #[cfg(not(target_os = "windows"))]
+                    #[cfg(target_os = "unix")]
```
msiglreith: nice, might even fix our travis errors :+1:

src/backend/metal/src/command.rs
```diff
@@ -1,477 +1,634 @@
...
+
+unsafe impl Send for CommandPool {
+}
+unsafe impl Sync for CommandPool {
+}
 
 #[derive(Clone)]
-pub struct SubmitInfo {
-    pub(crate) command_buffer: MTLCommandBuffer
+pub struct CommandBuffer(Arc<UnsafeCell<CommandBufferInner>>);
```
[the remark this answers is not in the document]
JohnColanduoni: So the CommandBuffer can be cloned. The Metal CommandBuffer needs some ancillary data so it's not sufficient to simply add a reference to MTLCommandBuffer.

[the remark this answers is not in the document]
msiglreith: `Clone` is only required/valid for finished command buffers to allow multi submission (probably not support atm in metal) or for convenience. The command buffer doesn't need to be in a valid state afterwards and calling commands on a cloned CommandBuffer is undefined behavior. I assume a normal clone would be enough here.

Note: I should have added clone inside `RawCommandBuffer` instead of the standard `clone`.

examples/core/quad/Cargo.toml
```diff
@@ -35,15 +35,8 @@ path = "../../../src/backend/vulkan"
...
-#[dependencies.gfx_device_metal]
-#path = "../../../src/backend/metal"
-#version = "0.3"
-#optional = true
-
-#[dependencies.gfx_window_metal]
-#path = "../../../src/window/metal"
-#version = "0.4"
-#optional = true
+[target.'cfg(macos)'.dependencies]
```
JohnColanduoni: I had to change this to `cfg(target_os = "macos")` to get it to compile.

src/backend/metal/src/command.rs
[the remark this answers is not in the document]
JohnColanduoni: I tried removing the Arc and then realized we still need it for `CommandPool.reset` to work (since MTLCommandBuffers can't actually be reset) :/

src/backend/metal/src/command.rs
JohnColanduoni: I think for array/cube textures extent.depth might actually map to layer count (otherwise sourceBytesPerImage doesn't make a ton of sense). I think it might be better to wait until we have a test case before we try to implement it that way though.

kvark:
Let's see if CI is happy
bors r+

kvark:
oh, travis, can't wait for you too long
windows/linux builds are passing, good to go

PR #1506, opened 2017-09-19
[ll] single descriptor set/pool for the core/quad example

kvark:


examples/core/quad/main.rs
```diff
@@ -266,17 +270,20 @@ fn main() {
...
-        1, // sets
-        &[pso::DescriptorRangeDesc { ty: pso::DescriptorType::SampledImage, count: 1 }],
-    );
-    let set0 = srv_pool.allocate_sets(&[&set0_layout]);
-
-    let mut sampler_pool = device.create_descriptor_pool(
-        1, // sets
-        &[pso::DescriptorRangeDesc { ty: pso::DescriptorType::Sampler, count: 1 }],
+    let mut desc_pool = device.create_descriptor_pool(
+        2, // sets
```
msiglreith: Why 2 sets?

kvark: should be one, but doesn't really matter :)

kvark:
bors r=msiglreith

PR #1511, opened 2017-09-20
[ll][dx12] Basic subpass dependencies

kvark:
(can't test it on AMD atm, Intel worked either way)

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -4,7 +4,7 @@ use core::{IndexCount, IndexType, InstanceCount, VertexCount, VertexOffset, View
 use core::buffer::IndexBufferView;
 use core::command::{BufferCopy, BufferImageCopy, ClearColor, ClearValue, ImageCopy, ImageResolve,
                     SubpassContents};
-use core::pass::{Attachment, AttachmentLoadOp, AttachmentOps};
+use core::pass::{Attachment, AttachmentLoadOp, AttachmentOps, SubpassRef};
 use winapi::{self, UINT64, UINT};
 use {conv, native as n, Backend};
 use smallvec::SmallVec;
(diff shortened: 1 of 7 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -24,29 +24,89 @@ fn get_rect(rect: &target::Rect) -> winapi::D3D12_RECT {
...
+                            StateBefore: state_src,
+                            StateAfter: state_dst,
+                        },
+                    }
+                );
+            }
+        }
+
+        unsafe {
+            self.raw.clone().ResourceBarrier(
```
msiglreith: Why do we need to clone? I assume we always have mutable access during command buffer execution.

src/backend/dx12/src/command.rs
```diff
@@ -24,29 +24,89 @@ fn get_rect(rect: &target::Rect) -> winapi::D3D12_RECT {
...
+                let state_src = conv::map_image_resource_state(dependency.accesses.start, att.layouts.start);
+                let state_dst = conv::map_image_resource_state(dependency.accesses.end, att.layouts.end);
+
+                if state_src == state_dst {
+                    continue;
+                }
+
+                let resource = match state.frame_buffer.color.get(i) {
+                    Some(view) => view.resource,
+                    None => state.frame_buffer.depth_stencil[0].resource,
```
msiglreith: We can't have multiple ds views in a framebuffer?

src/backend/dx12/src/device.rs
```diff
@@ -275,6 +277,7 @@ impl d::Device<B> for Device {
         n::RenderPass {
             attachments: attachments.to_vec(),
             subpasses,
+            dependencies: dependencies.to_vec(),
```
msiglreith: Something to keep in mind (for a follow-up) are the implicit dependencies at the start and end of a renderpass to EXTERNAL if there are any.

src/backend/dx12/src/command.rs
kvark: we should have, this will be TODO, esp considering that even a single depth buffer is not supported by our rough backends yet

src/backend/dx12/src/command.rs
kvark: we don't want to borrow `&mut self`

src/backend/dx12/src/device.rs
kvark: It's actually handled by the PR. Not correctly yet, but handled.

PR #1513, opened 2017-09-21
[ll] Surface capability querying

msiglreith:
Support querying surface properties to obtain reasonable values for swapchain creation.
Next step are extending swapchain create infos to make use of the queried results.

Metal currently unimplemented

### DIFF
--- a/src/backend/dx12/src/window.rs
+++ b/src/backend/dx12/src/window.rs
@@ -6,7 +6,7 @@ use std::ptr;
 use winit;
 use winapi;
 use wio::com::ComPtr;
-use {conv, native as n, Backend, Instance, QueueFamily};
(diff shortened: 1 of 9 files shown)


src/backend/dx12/src/window.rs
```diff
@@ -38,6 +38,20 @@ impl core::Surface<Backend> for Surface {
...
+    fn surface_capabilities(&self, _: &Adapter) -> core::SurfaceCapabilities {
+        let extent = core::window::Extent2d {
+            width: self.width,
+            height: self.height,
+        };
+
+        core::SurfaceCapabilities {
+            image_count: 2..16, // we currently use a flip effect which supports 2..16 buffers
+            current_extent: Some(extent),
+            extents: extent..extent,
```
kvark: mmm, nice ranges!

PR #1515, opened 2017-09-21
[ll][dx12] Proper resource flags, image to buffer copies

kvark:
This PR gets DX12 backend to a state capable of running 05-image sample of webgpu-servo

### DIFF
--- a/src/backend/dx12/src/native.rs
+++ b/src/backend/dx12/src/native.rs
@@ -1,4 +1,4 @@
-
+use core::device::ResourceHeapType;
 use core::pass::{Attachment, AttachmentRef, SubpassDependency};
 use core::pso::DescriptorSetLayoutBinding;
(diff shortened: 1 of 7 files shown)


src/backend/dx12/src/native.rs
```diff
@@ -256,7 +257,8 @@ pub struct DescriptorPool {
     pub pools: Vec<pso::DescriptorRangeDesc>,
     pub max_size: u64,
 }
-unsafe impl Send for DescriptorPool { }
+unsafe impl Send for DescriptorPool {}
+unsafe impl Sync for DescriptorPool {}
```
msiglreith: Why derive `Sync` here? Do you need it?

kvark: I don't think this is correct :( but yes I happen to need this at the moment.
Guess it's a good case for branching out then

Issue #1498, opened 2017-09-15
[ll] Memory Barriers portability


kvark:
Basic idea is to map Vulkan's pipeline barriers to all other APIs, where appropriate.
Investigation details here - https://github.com/gpuweb/gpuweb/issues/27

### Vulkan
Has resource access masks for buffers and images, plus image layouts for the latter.
Pipeline barriers are separate for buffers and images, there are targeted versions (single resource) as well as bulk ones (all resources of this type).

### D3D12
Has resource states, which are split into mutable/immutable. A resource can be in either a single mutable state, or in a combination of immutable ones. Some of the transitions are implicit to/from "generic" state.
Pipeline barriers allow transitioning a single resource between states, aliasing transitions, and special UAV handling.

Problems:
- [ ] actual precise mapping between Vk access+layout to D3D12 resource state
- [ ] bulk transitions (all resources of a type)
- [ ] render pass dependencies with `VK_SUBPASS_EXTERNAL`

kvark:
More info on the last issue from @msiglreith (please feel free to edit/append):
>Renderpasses allows to define dependencies between subpassess which form a DAG. For non-attachments these dependencies equals to global memory pipeline barriers and for attachments these are basically image barriers (shouldn't be an issue). Synchronization scopes are defined by `srcSubpass` and `dstSubpass`. I'm unsure it we can simply skip `VK_SUBPASS_EXTERNAL` as a parameter .

msiglreith:
* My current analysis for transition states (comparison between vulkan and d3d12): https://gist.github.com/msiglreith/1f23d900560dea1e6a8b07c7e91535a6

* Global memory barriers _could_ be emulated via https://github.com/gpuweb/gpuweb/issues/27#issuecomment-330719329

* D3D12's UAV barriers can be inferred by by SHADER_READ/SHADER_WRITE occurences.

* D3D12's alias barriers are needed for accessing aliased resources. Vulkan requires transitions for images with aliased memory from `VK_IMAGE_LAYOUT_UNDEFINED` or `VK_IMAGE_LAYOUT_PREINITIALIZED` before usage. Need to look further into buffer aliasing. We might be able to infer the d3d12 barriers also.

Kangz:
First off, nice investigation! This comment will be a good reference.

> Global memory barriers could be emulated via gpuweb/gpuweb#27 (comment)

[...]

PR #1517, opened 2017-09-23
[ll] dx12: Implement pipeline barriers

msiglreith:
Implement pipeline barriers for dx12 based on vulkan's access model.

Requires a few workarounds and restrictions to make this work:
* We differ between read and read/write states as it's a limitation by dx12. Allowing to only have one write bit set but multiple read bits if no write bit is set.
* `TRANSFER_WRITE` for images is used for all copy commands (including resolve). We add transitions barriers for resolve operations before and afterwards (can be optimized maybe) to have a consistent state accross command list boundaries.
* `SHADER_READ` is limited to SRV reads, which we need to enforce on the shader side for HLSL to get it working. `SHADER_WRITE` is used for write access to UAV and can be combined with `SHADER_READ` to obtain read access to UAV (but not SRV).
* Buffers can only be transistioned as in a whole, not on a range basis. (TODO: image only whole subresources or whole image).
* Global memory barriers currently unsupported (not totally clear if we can support them)
* UAV and aliasing barriers are always inserted atm, needs further investigation on how we can do this smarter.

Imo these are valid restrictions to vulkan's specification.
cc #1498

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -208,24 +208,20 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
         _stages: Range<pso::PipelineStage>,
         barriers: &[memory::Barrier<Backend>],
     ) {
-        // TODO: very much WIP
-        warn!("pipeline barriers unimplemented");
+        let mut raw_barriers = Vec::new();
 
-        let mut transition_barriers = Vec::new();
-
+        // transition barriers
         for barrier in barriers {
             match *barrier {
-                memory::Barrier::Image { ref states, target, ref range } => {
-                    let _ = range; //TODO: use subresource range
-                    let state_src = conv::map_image_resource_state(states.start.0, states.start.1);
-                    let state_dst = conv::map_image_resource_state(states.end.0, states.end.1);
+                memory::Barrier::Buffer { ref states, target } => {
+                    let state_src = conv::map_buffer_resource_state(states.start);
+                    let state_dst = conv::map_buffer_resource_state(states.end);
 
                     if state_src == state_dst {
-                        warn!("Image pipeline barrier requested with no effect: {:?}", barrier);
                         continue;
                     }
 
-                    transition_barriers.push(
+                    raw_barriers.push(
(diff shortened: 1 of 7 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -316,6 +383,22 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
         }
+
+        {
+            // Insert barrier for back transition from `RESOLVE_DEST` to `COPY_DEST`.
+            let transition_barrier = winapi::D3D12_RESOURCE_BARRIER {
+                Type: winapi::D3D12_RESOURCE_BARRIER_TYPE_TRANSITION,
+                Flags: winapi::D3D12_RESOURCE_BARRIER_FLAG_NONE,
+                u: winapi::D3D12_RESOURCE_TRANSITION_BARRIER {
+                    pResource: dst.resource,
+                    Subresource: winapi::D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES,
```
kvark: here is a TODO item, I suppose

src/backend/dx12/src/conv.rs
```diff
@@ -421,27 +421,82 @@ pub fn map_function(fun: Comparison) -> D3D12_COMPARISON_FUNC {
     }
 }
 
-pub fn map_image_resource_state(access: image::Access, layout: image::ImageLayout) -> D3D12_RESOURCE_STATES {
+pub fn map_buffer_resource_state(access: buffer::Access) -> D3D12_RESOURCE_STATES {
+    // Mutable states
+    if access.contains(buffer::SHADER_WRITE) {
```
kvark: do we check anywhere that only a single mutable flag is specified?

src/backend/dx12/src/command.rs
msiglreith: Yes! Good find.

src/backend/dx12/src/conv.rs
msiglreith: no, I would shift this into the validation layer.

kvark:
Please proceed with or without bors

PR #1518, opened 2017-09-25
[ll] Metal Backend: Support multiple windowing libs

davll:
Metal backend in `ll` branch needed winit object to create surface. However, it was inconvenient for users with sdl2, glew3 windowing library instead of winit library. Thus I made a trait NativeWindow to bridge window objects and the metal Instance in gfx-rs.

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -87,6 +87,7 @@ impl Instance {
         }
     }
 
+    #[cfg(feature = "winit")]
     pub fn create_surface(&self, window: &winit::Window) -> Surface {
         use winit::os::macos::WindowExt;
         self.create_surface_from_nsview(window.get_nsview())
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -41,18 +41,6 @@ use core_graphics::geometry::CGRect;
 pub struct Instance {
 }
 
-pub trait NativeWindow {
-    fn get_nsview(&self) -> *mut c_void;
-}
-
-#[cfg(feature = "winit")]
-impl NativeWindow for winit::Window {
-    fn get_nsview(&self) -> *mut c_void {
-        use winit::os::macos::WindowExt;
-        WindowExt::get_nsview(self)
-    }
-}
-
 impl core::Instance<Backend> for Instance {
     fn enumerate_adapters(&self) -> Vec<Adapter> {
         // TODO: enumerate all devices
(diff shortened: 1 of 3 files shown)


kvark:
I feel that we better have `winit` as default (either built-in or as a default feature), with extra windowing stuff like sdl/glfw coming as extra features. `winit` is available everywhere, and it's rusty, means that it's a natural fit to work with gfx-rs. We should encourage it, not gate it away.

davll:
@msiglreith Thanks for approval :D
However, I discovered `NativeWindow trait` approach might be problematic for Vulkan backend and other backends, since it has to detect WinAPI, Xlib, XCB, and Wayland. Should `NativeWindow trait` be specific for gfx_backend_* or defined in gfx_backend_common (the shared lib for gfx backends)?

@kvark Is it OK to make `winit` feature default?

davll:
I've rebased the commits :)

src/backend/metal/src/lib.rs
```diff
@@ -61,11 +77,9 @@ impl Instance {
         Instance {}
     }
 
-    pub fn create_surface(&self, window: &winit::Window) -> Surface {
+    pub fn create_surface(&self, window: &NativeWindow) -> Surface {
```
kvark: I don't think having a trait here is really necessary. We don't need to be generic here. Instead, we could have:
```rust
pub fn create_surface_from_nsview(&self, ns_view: cocoa::base::id) -> Surface {...}
#[cfg(feature="winit")]
pub fn create_surface(&self, window: &winit::Window) -> Surface {
    use winit::os::macos::WindowExt;
    self.create_surface_from_nsview(WindowExt::get_nsview(window))
}
```

davll:
@kvark Nice point! Thanks for the review. `NativeWindow` trait might be redundant because `gfx_backend_metal` is actually macOS/iOS specific. And providing native interface gives better extensibility and room for abstraction in future. 👍

src/backend/metal/src/lib.rs
```diff
@@ -84,6 +86,11 @@ impl Instance {
             }))
         }
     }
+
+    pub fn create_surface(&self, window: &winit::Window) -> Surface {
```
msiglreith: Should be feature gated imo.

davll: Oops, my mistake :(

msiglreith: np, thanks for the quick fix!

msiglreith:
Btw do you want to include further changes in this PR (asking because of the `[WIP]`)?
Else the PR looks good to me in the current state and we can proceed merging imo.

davll:
Yes, the PR is done and waiting to be merged :D

PR #1520, opened 2017-09-25
[ll][dx12] Subpass Dependencies

kvark:
I'm not yet confident that this is going to be enough, but the implementation is somewhat sound, and it will get us going for the nearest future. Basic logic is the following:

We ignore the access masks and pipeline stages of dependencies, but use the source/dest subpass indexes in order to build an order of subpasses, which fulfills the dependencies (i.e. traversing the graph in this order means that the dependencies of each next subpass are resolved by the time we get to it).

Within the produced order, for each attachment we figure if there is a particular state it needs to be, or preserved, or we don't care. Once we see a required state, we determine where the transition needs to start (at some previous pass, or right here in place), record new barriers, and update the current values for the attachment.

This barrier recording is done in `create_renderpass`, which means we can't store them directly as `D3D12_RESOURCE_STATES`, because we don't know the exact resources yet. Thus, we keep them in `native::BarrierDesc` objects, which get resolved in a simple way during the render pass encoding.

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -1,10 +1,9 @@
 use wio::com::ComPtr;
-use core::{command, image, memory, pso, target};
+use core::{command, image, memory, pass, pso, target};
 use core::{IndexCount, IndexType, InstanceCount, VertexCount, VertexOffset, Viewport};
 use core::buffer::IndexBufferView;
 use core::command::{BufferCopy, BufferImageCopy, ClearColor, ClearValue, ImageCopy, ImageResolve,
                     SubpassContents};
-use core::pass::{Attachment, AttachmentLoadOp, AttachmentOps, SubpassRef};
 use winapi::{self, UINT64, UINT};
 use {conv, native as n, Backend};
 use smallvec::SmallVec;
@@ -28,6 +27,15 @@ pub struct RenderPassCache {
     clear_values: Vec<ClearValue>,
 }
 
+impl RenderPassCache {
+    fn get_resource(&self, att_id: pass::AttachmentId) -> *mut winapi::ID3D12Resource {
+        match self.frame_buffer.color.get(att_id) {
+            Some(view) => view.resource,
+            None => self.frame_buffer.depth_stencil[att_id - self.frame_buffer.color.len()].resource,
+        }
+    }
+}
+
 #[derive(Clone)]
 pub struct CommandBuffer {
     raw: ComPtr<winapi::ID3D12GraphicsCommandList>,
@@ -35,7 +43,7 @@ pub struct CommandBuffer {
 
     // Cache renderpasses for graphics operations
     pass_cache: Option<RenderPassCache>,
-    next_subpass: usize,
+    cur_subpass: usize,
 }
 
 unsafe impl Send for CommandBuffer { }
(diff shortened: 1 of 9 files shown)


kvark:
Note to self: need to have a small tweak to at least not fail on subpass self-dependencies.

kvark:
Vulkan was broken due to unrelated reason, should be good now on CI.

src/backend/dx12/src/command.rs
```diff
@@ -133,6 +120,16 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
     ) {
         assert!(framebuffer.depth_stencil.len() <= 1);
```
msiglreith: I think we can safely remove this.

src/backend/dx12/src/device.rs
```diff
@@ -266,20 +266,133 @@ impl d::Device<B> for Device {
...
-        n::RenderPass {
+        let mut rp = n::RenderPass {
             attachments: attachments.to_vec(),
-            subpasses,
-            dependencies: dependencies.to_vec(),
+            subpasses: Vec::new(),
+            post_barriers: Vec::new(),
+        };
+
+        while let Some(sid) = deps_left.iter().position(|count| *count == 0) {
```
msiglreith: Intuitively, I would say this should be `*count != 0` because we check above for which subpasses we need to insert dependencies, shouldn't it?

src/backend/dx12/src/device.rs
```diff
@@ -266,20 +266,133 @@ impl d::Device<B> for Device {
...
+            for (att_id, ai) in att_infos.iter_mut().enumerate() {
+                let state_dst = match ai.sub_states[sid] {
+                    SubState::Preserve => {
+                        ai.barrier_start_index = rp.subpasses.len() + 1;
+                        continue;
+                    },
+                    SubState::New(state) if state != ai.last_state => state,
+                    _ => continue,
+                };
+                let barrier = n::BarrierDesc::new(att_id, ai.last_state .. state_dst);;
```
msiglreith: Nit: Unneeded ';'

src/backend/dx12/src/device.rs
```diff
@@ -266,20 +266,133 @@ impl d::Device<B> for Device {
...
+        };
+
+        while let Some(sid) = deps_left.iter().position(|count| *count == 0) {
+            //TODO:
+            // for each attachment used by this subpass:
+            //   - figure out the previous layout
+            //   - figure out the previous access flags
+            //   - insert either a split or a single barrier
+            // clean up used dependencies
+            deps_left[sid] = 0xFFFF; //dummy
```
msiglreith: We can't use `Option<_>` here due to the `vec![..]` macro?

src/backend/dx12/src/device.rs
kvark: The array stores the number of dependencies that this subpass needs to be fulfilled before it can start. So `count==0` would mean we are good to go, after which we scan over all the dependencies that start off this subpass and deduce them from the descination subpasses.

kvark:
One problem with ignoring the access states is that we'd like to avoid going going to/from COMMON state at the renderpass boundaries. This can be addressed as a follow-up though.

msiglreith:
Do you want to address the appveyor build failure?

PR #1521, opened 2017-09-26
[ll] heap creation -> memory allocation

kvark:
This PR renames heaps to memory, which is allocated/freed instead of creation/destruction.
Seeing how we are getting further from DX12 heaps and leaning towards thinner Vk portability layer, naming memory allocations for what they are should make sense.

Also, the DX12 backend update will include splitting the memory types for Resource Heap Tier1 hardware (like NV Vulkan driver does), which is what allows us to get rid of an unsound `ResourceHeapType`.

Todo:
- [x] core + dummy
- [x] render + example
- [x] gl
- [x] vulkan (more work needed for error detection)
- [x] dx12
- [x] metal
- [x] core example

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -1,6 +1,6 @@
 use conv;
 use core::{buffer, device as d, format, image, mapping, memory, pass, pso, state};
-use core::{Features, HeapType, Limits};
+use core::{Features, Limits, MemoryType};
 use core::memory::Requirements;
 use d3d12;
 use d3dcompiler;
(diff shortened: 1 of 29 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -772,6 +768,7 @@ impl d::Device<B> for Device {
         let requirements = memory::Requirements {
             size,
             alignment: winapi::D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT as u64,
+            type_mask: if self.private_caps.heterogeneous_resource_heaps { 0x7 } else { 0x7<<2 },
```
msiglreith: Should be `0x7<<4` if I understood the bitlayout correctly.

PR #1522, opened 2017-09-26
[0.16] Add allow unused super::gfx import for pso macro

alexheretic:
Can produce 
```
warning: unused import: `super::gfx`
```

### DIFF
--- a/src/render/src/macros/pso.rs
+++ b/src/render/src/macros/pso.rs
@@ -231,6 +231,7 @@ macro_rules! gfx_pipeline_base {
         pub mod $module {
             #[allow(unused_imports)]
             use super::*;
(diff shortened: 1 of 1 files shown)


kvark:
I believe we need `gfx` imported since this is where PSO components are defined. Also would be worth to see if we can do `use $crate`, but that's not too important.
bors r+

kvark:
note: this needs to go into `pre-ll` (at least) as well

alexheretic:
`use $crate` produces a warning saying it will be a hard error in future. This `use` statements is for gfx *consumers*, not gfx itself (Though I personally don't use it, just want the warnings to go). 

Would you like me to raise this for `pre-ll` as well?

PR #1525, opened 2017-09-29
[ll] Clear attachments and small fixes

msiglreith:
* Add `clear_attachments` command for clears inside renderpasses
* dx12: Implement `Reset` command
* dx12: Fix subresource ranges for image transitions


kvark:
Looks like DX12 build is broken too?..

msiglreith:
Latest `spirv_cross` versions `0.1.x` were breaking backwards compatibility. Should be fixed again with the changes from #1526

PR #1526, opened 2017-09-29
[ll] dx12: Update spirv-cross version and set correct shader model

msiglreith:


src/backend/dx12/src/device.rs
```diff
@@ -6,8 +6,8 @@ use d3d12;
 use d3dcompiler;
 use dxguid;
 use kernel32;
-use spirv_cross::{hlsl, spirv, CompileTarget};
-use spirv_cross::spirv::ExecutionModel;
+use spirv_cross::hlsl::{self, ShaderModel};
```
kvark: nit: can we just do `use spirv_cross::{hlsl, spirv}` instead?

msiglreith: sure, done!

src/backend/dx12/src/device.rs
```diff
@@ -751,25 +762,31 @@ impl d::Device<B> for Device {
 
         let module = spirv::Module::new(spirv_data);
 
-        let parse_options = spirv::ParserOptions::new();
-        let ast = spirv::Parser::new(CompileTarget::Hlsl)
+        let parse_options = spirv::ParserOptions::default();
+        let parsed = spirv::Parser::new()
             .parse(&module, &parse_options)
             .map_err(|_| d::ShaderError::CompilationFailed("Unknown parsing error".into()))?;
```
grovesNL: FYI `spirv_cross` [tries to retrieve an error message now in `ErrorCode::CompilationError(String)`](https://github.com/grovesNL/spirv_cross/blob/master/spirv_cross/src/lib.rs#L26) - could be useful for some initial debugging if we update the quad shader.

kvark:
Metal is broken:
> error[E0412]: cannot find type `ShaderStageFlags` in this scope
   --> /Users/travis/build/gfx-rs/gfx/src/backend/metal/src/native.rs:165:22
    |
165 |     pub stage_flags: ShaderStageFlags,
    |                      ^^^^^^^^^^^^^^^^ not found in this scope
    |

We'll fix it afterwards.

Issue #1226, opened 2017-04-10
[ll] Vulkan portable implementation FFI


kvark:
If #1102 brings us to a Vulkan-like API of the `gfx_core`, there is one interesting option to explore - create a C FFI layer that with Vulkan-like header (or better - an actual one from upstream) that would work with our core library.

This would allow any application targeting Vulkan to work on a variety of platforms, and will be solidly backed by our Rust implementation. Essentially, the [Angle](https://github.com/google/angle) of the new era, taking on MoltenVK at the same time. The future has many use of such a layer ;)

Another absolutely crazy extension of this path is a possible use of `vulkano` as a high-level layer instead of `gfx_render`. It can just be an alternative, but convergence is great!

Bastacyclop:
Wouldn't this need shader support too ?

kvark:
Yes, via existing SPIRV-Cross library

kvark:
It is happening in https://github.com/kvark/portability

PR #1527, opened 2017-09-29
[ll] Copy implementations for most of the resource types

kvark:
~~requires https://github.com/MaikKlein/ash/pull/31~~

Needed for #1226, so that we can expose most of the types as copyable things without boxing them.
I'm going to implement the `copy` feature on other backends too, but still not sure if `gfx_core` needs to force `Copy` on some of the `Backend` associated types, or just having them in the backends is enough.

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -26,13 +26,56 @@ mod window;
 
 use core::{memory, Features, Limits, QueueType};
 use spirv_cross::hlsl;
-use wio::com::ComPtr;
+
+#[cfg(not(feature = "copy"))]
+pub use wio::com::ComPtr;
+#[cfg(feature = "copy")]
+use copy_ptr::CopyPtr as ComPtr;
 
 use std::{mem, ptr};
 use std::os::raw::c_void;
 use std::os::windows::ffi::OsStringExt;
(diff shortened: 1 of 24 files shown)


src/backend/dx12/src/lib.rs
```diff
@@ -26,7 +26,11 @@ mod window;
 
 use core::{memory, Features, Limits, QueueType};
 use spirv_cross::hlsl;
+
+#[cfg(not(feature = "copy"))]
 use wio::com::ComPtr;
+#[cfg(feature = "copy")]
+use core::copy::Pointer as ComPtr;
```
msiglreith: An issue I see with this approach is the missing `Release` calls once a `ComPtr` drops.

src/core/Cargo.toml
```diff
@@ -13,6 +13,7 @@ workspace = "../.."
 [features]
 #TODO: switch to pure `serde` feature once `draw_state` is no longer used
 serialize = ["serde", "serde_derive", "draw_state/serialize"]
+copy = []
```
msiglreith: Have you planned further portability specific changes? If so, we might want to rename this feature as it's probably only needed for FFI related tasks.

kvark:
I'll see how useful this is for the portability first, then come back later.

PR #1528, opened 2017-10-01
Metal: primitive types, image to buffer copies

kvark:
This change has a few //TODO items, with an assumption that we need to have a follow-up pass that clarifies Send/Sync usage on resource anyway.

There are still issues with Metal backend: WebGPU prototype crashes while recording an MTL blit encoder, (improper reset/cb re-use most likely), but at least it shows the first frame correctly with a texture now.
The backend also needs to properly implement binding inheritance (e.g. don't require PSO to be set outside of the encoder), but that's for future PRs.

### DIFF
--- a/examples/core/quad/main.rs
+++ b/examples/core/quad/main.rs
@@ -73,7 +73,7 @@ fn main() {
     let pixel_height = window_size.1 as u16;
 
     // instantiate backend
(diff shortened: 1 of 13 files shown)


kvark:
Replaced by  #1529

PR #1533, opened 2017-10-02
[ll] updated Readme, Changelog, fixed dx12 nightly build

kvark:
The old README/CHANGELOG contents make little sense for `ll`. We'll keep them in `pre-ll`, where we'll be forwarding people who needs compatibility stuff.


kvark:
Should be good now.
bors r=msiglreith

kvark:
Looks like the scripts need more love (building vulkan on OSX). I'm going to address those separately.

PR #1534, opened 2017-10-03
[ll] Makefile and stuff

kvark:
"Stuff":
  - all scripts are replaced by a single `Makefile`. Simply typing `make` will now test everything for your particular configuration without worrying about excludes and features. Same is used on CI.
  - all CI/build scripts are refactored if not rewritten
  - GL is now properly CI covered
  - examples are now properly CI covered and don't invoke unneeded dependencies, ignore warnings where irrelevant
  - ll/pre-ll branches are now CI built by default


kvark:
All set.
bors r=msiglreith

kvark:
bors r=msiglreith

PR #1543, opened 2017-10-04
OpenGL: backbuffer, FBO, buffer creation, mapping, and more

kvark:
Still a long way to go before the quad example is working, but some decent progress nevertheless.


src/backend/gl/src/device.rs
```diff
@@ -313,22 +372,97 @@ impl d::Device<B> for Device {
...
+            unsafe {
+                gl.BufferData(target,
+                    size as _,
+                    ptr::null(),
+                    flags,
+                );
+            }
+        }
+
+        unsafe {
```
msiglreith: nit: could be in a single line

src/backend/gl/src/device.rs
```diff
@@ -366,22 +522,44 @@ impl d::Device<B> for Device {
...
-        unimplemented!()
+        let gl = &self.share.context;
+        let access = match read {
+            Some(_) => gl::READ_ONLY,
+            None => gl::WRITE_ONLY,
+        };
+
+        let data = unsafe {
+            gl.BindBuffer(buffer.target, buffer.raw);
+            let ptr = gl.MapBuffer(buffer.target, access);
```
msiglreith: Could we add a safety check at the beginning? Not all GL versions support glMapBuffer, none for GL ES.

src/backend/gl/src/info.rs
```diff
@@ -237,19 +238,28 @@ impl Info {
...
 /// Load the information pertaining to the driver and the corresponding device
 /// capabilities.
-pub fn get(gl: &gl::Gl) -> (Info, Features, Limits, PrivateCaps) {
+pub fn query_all(gl: &gl::Gl) -> (Info, Features, Limits, PrivateCaps) {
     use self::Requirement::*;
     let info = Info::get(gl);
     let tessellation_supported =           info.is_supported(&[Core(4,0),
                                                                Ext("GL_ARB_tessellation_shader")]);
     let multi_viewports_supported =        info.is_supported(&[Core(4,1)]); // TODO: extension
-    let max_work_group = get_usize(gl, gl::MAX_COMPUTE_WORK_GROUP_COUNT);
```
msiglreith: I mixed up the group count and group size terms 😟 
What I wanted was the maximum number of workgroups launched per dispatch per dimension, which GL and Vk specify as `Count` but I called it `Size` (denoting size of each workgroup).
We need to add the other one as well as it's more important in practical use imo and current checks might need to use the correct value then.

src/core/src/window.rs
```diff
@@ -141,7 +142,7 @@ pub trait Surface<B: Backend> {
     fn build_swapchain<C>(&mut self,
         config: SwapchainConfig,
         present_queue: &CommandQueue<B, C>,
-    ) -> (B::Swapchain, Vec<Backbuffer<B>>);
+    ) -> (B::Swapchain, Backbuffer<B>);
```
msiglreith: FWIW my goal for the future is to only return the swapchain here and add a function for retrieving the images (with the hack for GL). The GL backend would additionally provide a function to retrieve the default framebuffer.

src/backend/gl/src/conv.rs
```diff
@@ -34,3 +33,13 @@ pub fn wrap_to_gl(w: i::WrapMode) -> t::GLenum {
...
     }
 }
+
+pub fn buffer_usage_to_gl_target(usage: buffer::Usage) -> Option<t::GLenum> {
+    match usage & (buffer::CONSTANT | buffer::INDEX | buffer::VERTEX | buffer::INDIRECT) {
+        buffer::CONSTANT => Some(gl::UNIFORM_BUFFER),
+        buffer::INDEX => Some(gl::ELEMENT_ARRAY_BUFFER),
+        buffer::VERTEX => Some(gl::ARRAY_BUFFER),
+        buffer::INDIRECT => unimplemented!(),
+        _ => None
```
msiglreith: Do we need to return `None` here? Feels a bit weird to return an error when we actually have a GL version which supports different targets but older versions would just use `ARRAY_BUFFER`.

src/backend/gl/src/device.rs
```diff
@@ -237,12 +261,47 @@ impl d::Device<B> for Device {
...
-        _: &[&n::RenderTargetView],
-        _: &[&n::DepthStencilView],
-        _: d::Extent,
+        pass: &n::RenderPass,
+        rtvs: &[&n::TargetView],
+        dsvs: &[&n::TargetView],
+        _extent: d::Extent,
     ) -> n::FrameBuffer {
-        unimplemented!()
+        let gl = &self.share.context;
```
msiglreith: Should check private caps for framebuffer support.

kvark:
All concerns are addressed now, please take another look @msiglreith
I'm especially happy with how playing with memory types exposed on GL allows to figure out mapping support and flags :D

src/backend/gl/src/conv.rs
kvark: Some GL platforms (ES, looking at you) don't allow changing the bind target from what was bound for the first time. Returning `None` here has the semantics of: no target matches all the required `Usage`. The calling side then decides if it needs to throw up, or use a default target (`ARRAY_BUFFER`).

examples/core/quad/main.rs
```diff
@@ -317,7 +331,8 @@ fn main() {
     let upload_type =
         memory_types.iter().find(|mem_type| {
             buffer_req.type_mask & (1 << mem_type.id) != 0 &&
-            mem_type.properties.contains(m::CPU_VISIBLE | m::COHERENT)
+            mem_type.properties.contains(m::CPU_VISIBLE) &&
+            mem_type.properties.intersects(m::CPU_CACHED | m::WRITE_COMBINED)
```
msiglreith: > There must be at least one memory type with both the VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT and
VK_MEMORY_PROPERTY_HOST_COHERENT_BIT bits set in its propertyFlags. There must be at least one
memory type with the VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT bit set in its propertyFlags.

We might end up without an upload heap here.

kvark: What do you propose? Use `COHERENT` here as a fallback?

kvark:
@msiglreith memory properties were incorrect, this is addressed now. PTAL

PR #1546, opened 2017-10-05
[ll] dx12: Implement shader module creation from SPIR-V

msiglreith:
* Implement shader module creation from SPIR-V binaries, including patching of descriptor sets for images and samplers (more patching needed for constant buffers, etc.) and invert y-axis to match Vulkan's coordinate system

* Port quad example over to SPIR-V based shader modules

* Use safer `clear_attachments` wrapper from ash

I'm not totally happy with all the error handling in `create_shader_module` but we don't know when SPIRV-Cross might throw an exception /:


examples/core/quad/main.rs
```diff
@@ -117,19 +117,11 @@ fn main() {
     // dx12 runtime shader compilation
     #[cfg(feature = "dx12")]
     let vs_module = device
-        .create_shader_module_from_source(
-            pso::Stage::Vertex,
-            "vs_main",
-            "main",
-            include_bytes!("shader/quad.hlsl"))
+        .create_shader_module(include_bytes!("data/vert.spv"))
```
kvark: Can we switch the vulkan path to also get SPIRV for consistency, merging them here?

src/backend/dx12/src/device.rs
```diff
@@ -761,10 +798,11 @@ impl d::Device<B> for Device {
 
             let shader_blob = Self::compile_shader(
                 stage,
-                compile_options.shader_model,
+                hlsl::ShaderModel::V5_1,
```
grovesNL: It's already hardcoded above?

msiglreith:
Thanks for the reviews! Addressed all points.
bors r=kvark

examples/core/quad/main.rs
```diff
@@ -114,35 +114,14 @@ fn main() {
     let (mut swap_chain, backbuffers) = surface.build_swapchain(swap_config, &queue);
 
     // Setup renderpass and pipeline
-    // dx12 runtime shader compilation
-    #[cfg(feature = "dx12")]
+    #[cfg(any(feature = "vulkan", feature = "dx12"))]
```
kvark: 🎉

kvark:
those lovely OSX CI timeouts...
feel free to merge once Travis is done

PR #1548, opened 2017-10-06
[ll] dx12: create surface from native HWND

davll:


src/backend/dx12/src/window.rs
```diff
@@ -3,23 +3,40 @@ use core::{self, image};
 use dxguid;
 use std::collections::VecDeque;
 use std::ptr;
+#[cfg(feature = "winit")]
 use winit;
```
kvark: nit: could do `use winit::Window;` right away

src/backend/dx12/src/window.rs
```diff
@@ -3,23 +3,40 @@ use core::{self, image};
...
             factory: self.factory.clone(),
-            wnd_handle: window.get_hwnd() as *mut _,
+            wnd_handle: hwnd as *mut _,
             width: width,
             height: height,
         }
     }
+
+    #[cfg(feature = "winit")]
+    pub fn create_surface(&self, window: &winit::Window) -> Surface {
```
kvark: oh man, I wish we could have it in the `Instance` trait. I suppose GL is the only blocker: you can't add a GL context to a window that is already created

davll: I don’t think it is a good idea to make ‘create_surface’ be a part of ‘Instance’, because it is implementation dependent, and some platforms (ex: game consoles) do not have winit support.

src/backend/dx12/src/window.rs
davll: Should we make other backends (previous pull requests) consistent with it?

src/backend/dx12/src/window.rs
davll: I'm imagining a scenario: porting gfx_hal to PS4 platform.
1. Create a new backend: gfx_backend_gnm
2. Winit is not supported, and therefore it does not provide default feature "winit"
3. PS4 system provides its own windowing API for surface creation, neithor HWND nor NSView.

In conclusion, surface creation is platform dependent and it is not suitable to become a method of Instance trait.

davll: If we make gfx_hal also has feature="winit", it will create another problem: rustc thinks gfx_hal with winit and gfx_hal without winit are different libraries, and therefore data types of one library are not compatible to the other. It will make users pay more caution on it.

davll:
OK, the PR is tested and ready to merge.

src/backend/dx12/src/window.rs
kvark: sure thing, although this is really minor

src/backend/dx12/src/window.rs
kvark: > I don’t think it is a good idea to make create_surface be a part of Instance, because it is implementation dependent, and some platforms (ex: game consoles) do not have winit support.

True, although if we do get our hands on consoles I'd prefer `winit` to be ported on them as well.

> In conclusion, surface creation is platform dependent and it is not suitable to become a method of Instance trait

One can still have platform-specific methods for surface creation, it's not conflicting with `create_surface` being hooked up to `winit`.

> If we make gfx_hal also has feature="winit", it will create another problem: rustc thinks gfx_hal with winit and gfx_hal without winit are different libraries, and therefore data types of one library are not compatible to the other. It will make users pay more caution on it.

I don't think this is correct. If you are talking about serialization, then `winit` doesn't matter, and binary compatibility is there. If you are talking about link time errors - Cargo treats features in an additive manner, so if we have `gfx_hal` being depended upon with and without `winit` feature within a single linking unit, Cargo will just link both to the one that has the feature. No conflicts there either.

kvark:
Thank you!
bors r+

src/backend/dx12/src/window.rs
davll: >> I don’t think it is a good idea to make create_surface be a part of Instance, because it is implementation dependent, and some platforms (ex: game consoles) do not have winit support.

> True, although if we do get our hands on consoles I'd prefer winit to be ported on them as well.

>> In conclusion, surface creation is platform dependent and it is not suitable to become a method of Instance trait

> One can still have platform-specific methods for surface creation, it's not conflicting with create_surface being hooked up to winit.

@kvark

Sure it's great to have `winit` support game console platforms to ease prototyping!

However, I'm wondering if game developers will prefer using `winit` for retail games, since **many game developers get used to prototype games on Desktop platforms**. So I'm not certain wether it is necessary to port `winit` to game console platforms.

Besides, anyone who want to access game console related code should become registered developers due to **NDA**. Thus, we have to fork some repositories to vendors' own repositories before porting, and `winit` should be forked as well.

>> If we make gfx_hal also has feature="winit", it will create another problem: rustc thinks gfx_hal with winit and gfx_hal without winit are different libraries, and therefore data types of one library are not compatible to the other. It will make users pay more caution on it.

> I don't think this is correct. If you are talking about serialization, then winit doesn't matter, and binary compatibility is there. If you are talking about link time errors - Cargo treats features in an additive manner, so if we have gfx_hal being depended upon with and without winit feature within a single linking unit, Cargo will just link both to the one that has the feature. No conflicts there either.

OK, despite feature gate problem. I got another issues we might have: forked `winit` codebase from game console vendors' repositories. Users and gfx_hal should depend on forked `winit` instead of original `winit` otherwise cargo will treat two `winit` as different libraries (and I had run into the error many times in other projects). I think the problem is minor since users should also be registered developers to access forked version libs.

PR #1549, opened 2017-10-06
[gl] 3D/Cube texture support for copies

kvark:


kvark:
All green, and review period expired. Merging.

PR #1550, opened 2017-10-06
[ll][render] nicer descriptor set usage

kvark:
Not 100% sure about this one, but it does make the ergonomics better.

### DIFF
--- a/src/render/src/pso.rs
+++ b/src/render/src/pso.rs
@@ -38,7 +38,7 @@ pub trait BindDesc {
 pub trait Bind<B: Backend>: BindDesc {
     type Handle: 'static + Clone;
 
-    fn write<'a>(&[&'a Self::Handle]) -> core::pso::DescriptorWrite<'a, B>;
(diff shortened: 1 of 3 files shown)


src/render/src/pso.rs
```diff
@@ -38,7 +38,7 @@ pub trait BindDesc {
 pub trait Bind<B: Backend>: BindDesc {
     type Handle: 'static + Clone;
 
-    fn write<'a>(&[&'a Self::Handle]) -> core::pso::DescriptorWrite<'a, B>;
+    fn write<'a, H: AsRef<Self::Handle>>(&[&'a H]) -> core::pso::DescriptorWrite<'a, B>;
```
Bastacyclop: What bothers me a little is that the whole array has to have the same `H` type. You won't be able to mix raw and typed handles for example. I see some consistency benefit in simply taking raw handles as input. Maybe it would be more useful to have a `write_single` method for example that would only write one element at offset 0 as well as provide the `AsRef` sugar.

kvark:
@Bastacyclop I'm worried about the same thing. Let's post-pone this.

PR #1551, opened 2017-10-06
[ll] consistent framebuffer naming

kvark:
This is to be consistent with Vulkan.
Note: `RenderPass` is like this in Vulkan, so no need to fix it.


kvark:
@Bastacyclop I'm not a linguist, but I think what happened is that words Frame and Buffer got so close it doesn't make sense to separate them any more.
bors r= @Bastacyclop

PR #1554, opened 2017-10-08
[WIP][ll] grand View refactor

kvark:
This is early WIP:
- [x] core
- [ ] core example
- [ ] render + example
- [ ] vulkan
- [ ] d3d12
- [ ] metal
- [ ] gl

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -1021,13 +1021,63 @@ impl d::Device<B> for Device {
         unimplemented!()
     }
 
-    fn view_image_as_render_target(&mut self,
+    fn view_image_as_render_target(
+        &mut self,
         image: &n::Image,
         format: format::Format,
-        _layers: image::SubresourceLayers,
+        (mip_level, layers): image::SubresourceLayers,
     ) -> Result<n::RenderTargetView, d::TargetViewError> {
         //TODO: use subresource range
         let handle = self.rtv_pool.lock().unwrap().alloc_handles(1).cpu;
+
+        if image.kind.get_dimensions().3 != image::AaMode::Single {
+            error!("No MSAA supported yet!");
+        }
+        if layers.start + 1 != layers.end { //TODO
+            return Err(d::TargetViewError::Layers(layers));
+        }
+
+        let mut desc = winapi::D3D12_RENDER_TARGET_VIEW_DESC {
+            Format: match conv::map_format(format, true) {
+                Some(format) => format,
+                None => return Err(d::TargetViewError::BadFormat)
+            },
+            .. unsafe { mem::zeroed() }
+        };
+
+        match image.kind {
+            image::Kind::D2(..) => {
+                desc.ViewDimension = winapi::D3D12_RTV_DIMENSION_TEXTURE2D;
+                *unsafe { desc.Texture2D_mut() } = winapi::D3D12_TEX2D_RTV {
+                    MipSlice: mip_level,
+                    PlaneSlice: layers.start,
+                };
+            },
+            _ => unimplemented!()
+        };
+
+        unsafe {
+            self.raw.CreateRenderTargetView(
+                image.resource,
+                &desc,
+                handle,
+            );
+        }
+
+        Ok(n::RenderTargetView {
+            resource: image.resource,
+            handle,
+        })
+    }
+
+    fn view_image_as_depth_stencil(
+        &mut self,
+        image: &n::Image,
+        format: format::Format,
+        (mip_level, layers): image::SubresourceLayers,
+    ) -> Result<n::DepthStencilView, d::TargetViewError> {
+        //TODO: use subresource range
+        let handle = self.dsv_pool.lock().unwrap().alloc_handles(1).cpu;
 
         if image.kind.get_dimensions().3 != image::AaMode::Single {
             error!("No MSAA supported yet!");
@@ -1045,22 +1095,22 @@ impl d::Device<B> for Device {
             image::Kind::D2(..) => {
                 desc.ViewDimension = winapi::D3D12_RTV_DIMENSION_TEXTURE2D;
                 *unsafe { desc.Texture2D_mut() } = winapi::D3D12_TEX2D_RTV {
-                    MipSlice: 0,
-                    PlaneSlice: 0,
+                    MipSlice: mip_level,
+                    PlaneSlice: layers.start,
                 };
             },
             _ => unimplemented!()
         };
 
         unsafe {
-            self.raw.CreateRenderTargetView(
+            self.raw.CreateDepthStencilView(
                 image.resource,
                 &desc,
                 handle,
             );
         }
 
-        Ok(n::RenderTargetView {
+        Ok(n::DepthStencilView {
             resource: image.resource,
             handle,
         })
--- a/src/core/src/image.rs
+++ b/src/core/src/image.rs
@@ -10,8 +10,10 @@
 use std::error::Error;
 use std::fmt;
 use std::ops::Range;
+
 use {format, state, target};
 pub use target::{Layer, Level};
+
 
 /// Maximum accessible mipmap level of a image.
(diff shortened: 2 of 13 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -1045,8 +1049,8 @@ impl d::Device<B> for Device {
             image::Kind::D2(..) => {
                 desc.ViewDimension = winapi::D3D12_RTV_DIMENSION_TEXTURE2D;
                 *unsafe { desc.Texture2D_mut() } = winapi::D3D12_TEX2D_RTV {
-                    MipSlice: 0,
-                    PlaneSlice: 0,
+                    MipSlice: mip_level,
+                    PlaneSlice: layers.start,
```
msiglreith: Plance slice is not the layer (https://msdn.microsoft.com/en-us/library/windows/desktop/dn705766(v=vs.85).aspx)

src/backend/dx12/src/device.rs
```diff
@@ -1066,6 +1070,52 @@ impl d::Device<B> for Device {
...
+                Some(format) => format,
+                None => return Err(d::TargetViewError::BadFormat)
+            },
+            .. unsafe { mem::zeroed() }
+        };
+
+        match image.kind {
+            image::Kind::D2(..) => {
+                desc.ViewDimension = winapi::D3D12_RTV_DIMENSION_TEXTURE2D;
+                *unsafe { desc.Texture2D_mut() } = winapi::D3D12_TEX2D_RTV {
```
msiglreith: Should be `D3D12_TEX2D_DSV` and not `D3D12_TEX2D_RTV`

src/core/src/image.rs
```diff
@@ -60,6 +62,59 @@ impl Error for CreationError {
...
+                "Selected array layer doesn't exist",
+            ViewError::BadFormat =>
+                "An incompatible format was requested for the view",
+            ViewError::Unsupported =>
+                "The backend refused for some reason",
+        }
+    }
+
+    fn cause(&self) -> Option<&Error> {
+        if let ViewError::Layer(ref e) = *self {
```
msiglreith: `match` nicer here maybe

src/backend/dx12/src/device.rs
kvark: whoa, great catch! TIL

src/backend/dx12/src/command.rs
```diff
@@ -368,25 +357,50 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
+        image: &n::Image,
         _: image::ImageLayout,
-        color: ClearColor,
+        range: image::SubresourceRange,
+        value: com::ClearColor,
     ) {
-        self.clear_render_target_view(rtv.handle, color, &[]);
+        assert!((image::ASPECT_COLOR).contains(range.aspects));
+        assert_eq!(range, image.to_subresource_range(range.aspects));
+        if range.aspects.contains(image::ASPECT_COLOR) {
```
msiglreith: No need for this `if` clause we already check for `ASPECT_COLOR` above.

src/backend/dx12/src/command.rs
```diff
@@ -368,25 +357,50 @@ impl command::RawCommandBuffer<Backend> for CommandBuffer {
...
         &mut self,
-        rtv: &n::RenderTargetView,
+        image: &n::Image,
         _: image::ImageLayout,
-        color: ClearColor,
+        range: image::SubresourceRange,
+        value: com::ClearColor,
     ) {
-        self.clear_render_target_view(rtv.handle, color, &[]);
+        assert!((image::ASPECT_COLOR).contains(range.aspects));
```
msiglreith: Can we use `assert_eq!` here as it **must** be `ASPECT_COLOR

src/core/src/command/transfer.rs
```diff
@@ -75,9 +75,7 @@ pub struct BufferImageCopy {
     ///
     pub buffer_slice_pitch: u32,
     ///
-    pub image_aspect: image::AspectFlags,
-    ///
-    pub image_subresource: image::SubresourceLayers,
+    pub image_range: image::SubresourceRange,
```
msiglreith: I'm not really sure about the general removal of `SubresourceLayers` as we often end up with ranges where we only use the `start` of the mip levels. Using `SubrsourceRanges` suggests that we could for example copy multiple mipmap layers here which is wrong, resulting in a run-time error.

src/core/src/pso/descriptor.rs
```diff
@@ -91,11 +91,11 @@ pub struct DescriptorSetWrite<'a, 'b, B: Backend> {
...
-    UniformTexelBuffer,
-    StorageTexelBuffer,
-    ConstantBuffer(Vec<&'a B::ConstantBufferView>),
-    StorageBuffer,
-    InputAttachment(Vec<(&'a B::ShaderResourceView, ImageLayout)>),
+    SampledImage(Vec<(&'a B::ImageView, ImageLayout)>),
+    StorageImage(Vec<(&'a B::ImageView, ImageLayout)>),
+    UniformTexelBuffer(Vec<&'a B::BufferView>),
+    StorageTexelBuffer(Vec<&'a B::BufferView>),
+    ConstantBuffer(Vec<&'a B::BufferView>),
```
msiglreith: Vulkan doesn't use a `BufferView` but a `Buffer` + `Range<u64>`.

davll: !!! Yeah, you are right. All buffer descriptors in Vulkan do not use `VkBufferView`. :/
Why does Vulkan provide `VkBufferView` handles?
[Spec](https://vulkan.lunarg.com/doc/view/1.0.37.0/linux/vkspec.chunked/ch11s02.html)

davll: I'm curious why `&Vec<&T>` instead of `&[&T]`?

src/backend/dx12/src/command.rs
kvark: I tried to carefully support the case where the mask is empty

src/backend/dx12/src/command.rs
kvark: the check is non-trivial if the aspect mask is empty

src/backend/dx12/src/command.rs
```diff
@@ -22,18 +20,8 @@ fn get_rect(rect: &target::Rect) -> winapi::D3D12_RECT {
 #[derive(Debug, Clone)]
 pub struct RenderPassCache {
     render_pass: n::RenderPass,
-    frame_buffer: n::FrameBuffer,
-    render_area: winapi::D3D12_RECT,
-    clear_values: Vec<ClearValue>,
```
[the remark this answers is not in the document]
kvark: Resolved now

src/core/src/command/transfer.rs
kvark: Resolved now

src/core/src/pso/descriptor.rs
kvark: Fixed the `BufferView` here. Left `Vec` for the original code, was wondering about `&[]` alternative myself too. @msiglreith do you recall why you used `Vec` here?

src/backend/dx12/src/command.rs
msiglreith: The aspect mask **must** not be empty for a subresource range.

src/backend/dx12/src/command.rs
msiglreith: (As above)

src/core/src/pso/descriptor.rs
msiglreith: I think I had some builder pattern back then for descriptor updates. `&[]` should be the way to go.

kvark:
Ok, let's proceed and address the aspect mask requirements in follow ups. The current code is behaving correctly, just not validating strictly enough.

kvark:
Oh well, it's as green as it could be. Gotta bump that timout soon.

PR #1556, opened 2017-10-10
[ll] Add compute descriptor set binding

msiglreith:
Add `bind_compute_descriptor_sets` command, which is essential for compute support!
Only vulkan and dx12 implementation.


src/backend/metal/src/command.rs
```diff
@@ -500,6 +500,15 @@ impl core::RawCommandBuffer<Backend> for CommandBuffer {
         unimplemented!()
     }
 
+    fn bind_graphics_descriptor_sets(
```
kvark: typo

msiglreith: ofc in the only backend I can't check locally :sweat_smile:
thanks for reviewing!

bors r=kvark

PR #1559, opened 2017-10-11
Going aLL the way down

kvark:
This PR is the second grand LL merging (first was #1391). It breaks examples on `master`, and `gfx_render` still needs more work, but we can't pretend that `master` makes any sense anyway.


msiglreith:
bors r+

Let's see if it needs to re-build else we can just merge it

PR #1560, opened 2017-10-11
Implement update_descriptor_sets with the new view system on Vulkan, plus corollary changes.

AlphaModder:
This PR updates `Device::update_descriptor_sets` to use the new system of views, and implements all currently supported descriptor types. 

Some related bits of code (match statements, enums) related to the changes have been reordered/renamed to be consistent. This can be undone if it's an inconvenience.

### DIFF
--- a/src/core/src/buffer.rs
+++ b/src/core/src/buffer.rs
@@ -54,13 +54,19 @@ bitflags!(
         ///
         const TRANSFER_DST = 0x2,
         ///
-        const CONSTANT = 0x4,
+        const UNIFORM = 0x4,
         ///
-        const INDEX = 0x8,
+        const STORAGE = 0x8,
         ///
-        const INDIRECT = 0x10,
+        const UNIFORM_TEXEL = 0x10,
         ///
-        const VERTEX = 0x20,
+        const STORAGE_TEXEL = 0x20,
+        ///
+        const INDEX = 0x40,
+        ///
+        const INDIRECT = 0x80,
+        ///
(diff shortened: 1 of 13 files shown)


src/core/src/buffer.rs
```diff
@@ -48,19 +48,25 @@ impl Error for ViewError {
 bitflags!(
     /// Buffer usage flags.
     #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
-    pub flags Usage: u16 {
+    pub flags Usage: u32 {
```
kvark: not sure why we'd need `u32` here just yet?

src/core/src/device.rs
```diff
@@ -183,7 +183,7 @@ pub trait Device<B: Backend>: Clone {
 
     ///
     fn create_buffer_view(
-        &mut self, &B::Buffer, Range<u64>
+        &mut self, &B::Buffer, Range<u64>, format::Format
```
kvark: for images, we do `image, format, subresource`. I think it would be consistent to follow the order for buffers:
```rust
&mut self, &B::Buffer, format::Format, Range<u64>
```

src/core/src/buffer.rs
AlphaModder: Whoops, I was tired when I was writing that and thought I was going from u8 to u16!

kvark:
Looks like we see a bors bug, it still thinks it's cancelled :(
Anyhow, I included the commit in #1561, so it's easier to CI that right away.

PR #1561, opened 2017-10-12
Core -> HAL rename

kvark:
Includes  #1560


kvark:
This change goes as planned with the team, no semantical changes. Better not delay it too much to minimize disruption.
bors r+

Issue #1564, opened 2017-10-12
[pre-ll][dx11] gamma example readback


kvark:
I've been testing the readback on dx11, and faced an unfortunate limitation: DX11 can't copy buffer<->texture, it's `CopySubresourceRegion` requires same types of resources on both sides. This seems like an unfortunate blocker, given that @zakorgy would want that for WR reftests.

I'll make a PR with what I have, but in the mean time we need to figure out possible solutions. One that I can think of is just providing a mapping ability for textures specifically in DX11 factory.

Bastacyclop:
Previously #1208. I guess compute shaders would be too much work to fix this ?

kvark:
Yes, especially if we can have an easier workaround.

Bastacyclop:
What if staging buffers had a DX11-specific flag to differenciate between buffer staging and texture staging ? For texture staging the staging buffer would actually be a mappable DX11 texture.

PR #1565, opened 2017-10-12
[WIP][pre-ll] better dx11 resize support, gamma port

kvark:
Findings on the way to #1564
RFC!
cc @Bastacyclop @zakorgy

### DIFF
--- a/examples/gamma/main.rs
+++ b/examples/gamma/main.rs
@@ -14,7 +14,12 @@
 
 #[macro_use]
 extern crate gfx;
+#[cfg(not(target_os = "windows"))]
 extern crate gfx_window_glutin;
+#[cfg(target_os = "windows")]
+extern crate gfx_window_dxgi;
+#[cfg(target_os = "windows")]
+extern crate gfx_device_dx11;
 extern crate glutin;
 extern crate image;
 
@@ -22,6 +27,7 @@ use gfx::traits::{Factory, FactoryExt};
 use gfx::Device;
 use gfx::memory::Typed;
 use gfx::format::{Formatted, SurfaceTyped};
+#[cfg(not(target_os = "windows"))]
 use glutin::GlContext;
 
 pub type ColorFormat = gfx::format::Rgba8;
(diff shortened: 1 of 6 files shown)


examples/gamma/main.rs
```diff
@@ -50,21 +56,46 @@ const CLEAR_COLOR: [f32; 4] = [0.5, 0.5, 0.5, 1.0];
 
 type SurfaceData = <<ColorFormat as Formatted>::Surface as SurfaceTyped>::DataType;
 
+enum Backend {
```
Bastacyclop: Why use a single variant enum instead of a tuple struct ?

examples/gamma/main.rs
```diff
@@ -50,21 +56,46 @@ const CLEAR_COLOR: [f32; 4] = [0.5, 0.5, 0.5, 1.0];
...
-    let (window, mut device, mut factory, main_color, mut main_depth) =
-        gfx_window_glutin::init::<ColorFormat, DepthFormat>(window_builder, context, &events_loop);
-    let mut encoder: gfx::Encoder<_, _> = factory.create_command_buffer().into();
-    let pso = factory.create_pipeline_simple(
-        include_bytes!("shader/quad_150.glslv"),
-        include_bytes!("shader/quad_150.glslf"),
-        pipe::new()
-    ).unwrap();
+
+    let (mut backend, mut device, mut factory, pso, main_color) = match () {
```
Bastacyclop: Why not use `{}` instead of `match () { _ => {} }` ?

src/lib.rs
```diff
@@ -242,18 +242,20 @@ A: Sized + ApplicationBase<gfx_device_dx11::Resources, D3D11CommandBuffer>
...
-                    let new_depth = factory.create_depth_stencil_view_only(width, height).unwrap();
-                    app.on_resize(&mut factory, WindowTargets {
-                        color: new_color,
-                        depth: new_depth,
-                        aspect_ratio: width as f32 / height as f32,
-                    });
-                },
-                Err(e) => error!("Resize failed: {}", e),
-            }
+            //TODO: clean up the old references first!
```
Bastacyclop: How do you plan to do that ? Some interior mutability might enable the DX11 backend to update the existing reference or to invalidate it. Another possibility could be to have some kind of generation id to avoid reference mutation while still invalidating it.

src/window/dxgi/src/lib.rs
```diff
@@ -190,20 +190,27 @@ impl DeviceExt for Device {
...
+    main_color: &mut h::RenderTargetView<Resources, Cf>,
+) where
+    Cf: format::RenderFormat + format::TextureFormat,
+    D: DeviceExt,
 {
-    
+    use core::Factory;
+    // first, replace the main view pointer with a dummy
+    // the code assumes this is the last standing reference, beware!
+    *main_color = factory.create_render_target(1, 1).unwrap().2;
```
Bastacyclop: I'm not sure what's the purpose of this dummy target ? Looking at #1122 makes me believe that it's meant to propagate it to the application resize handling, which is not possible at this crate level. You could drop the given handle and return a new one ?

examples/gamma/main.rs
kvark: that came from my original attempt to make it a run-time switch. Fixed now

examples/gamma/main.rs
```diff
@@ -50,21 +56,46 @@ const CLEAR_COLOR: [f32; 4] = [0.5, 0.5, 0.5, 1.0];
 
 type SurfaceData = <<ColorFormat as Formatted>::Surface as SurfaceTyped>::DataType;
 
+enum Backend {
+    #[cfg(not(target_os = "windows"))]
+    Gl(glutin::GlWindow, gfx::handle::DepthStencilView<gfx_window_glutin::Resources, DepthFormat>),
```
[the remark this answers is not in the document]
kvark: right, should be fixed now

examples/gamma/main.rs
kvark: `{}` in expressions is not allowed, and copying the whole starting line for both cases would be unfortunate

src/lib.rs
kvark: Sure, good ideas!
It's not a big priority for me now to make this work universally on Dx11. The client doesn't use `gfx_app`, they just need a way to read back textures.

src/window/dxgi/src/lib.rs
kvark: > You could drop the given handle and return a new one ?

dropping is fine, but then you'd have to do some magic to prevent a double drop when the new one needs to overwrite the old one.
I chose the simplest approach here - overwrite the handle with a dummy target temporarily, so that the old one is freed.

examples/gamma/main.rs
torkleyy: Maybe you could use `if cfg!` together with else?

torkleyy: Ah the code won't compile, sorry.

PR #1566, opened 2017-10-12
[ll] Fix missing methods in RenderPass Encoder

davll:
There are some commands allowed in RenderPass mode are missing.

- [X] set_descriptor_sets
- [x] set_viewports
- [x] set_scissors
- [x] set_blend_constants
- [x] set_stencil_reference
- ~set_line_width~ (move to another PR)
- ~set_depth_bounds~  (move to another PR)
- ~set_depth_bias~  (move to another PR)
- ~set_stencil_compare_mask~  (move to another PR)
- ~set_stencil_write_mask~  (move to another PR)

### DIFF
--- a/src/hal/src/command/renderpass.rs
+++ b/src/hal/src/command/renderpass.rs
@@ -1,5 +1,5 @@
 use std::ops::Range;
-use {pso, target, Backend, IndexCount, InstanceCount, VertexCount, VertexOffset};
+use {pso, target, Backend, IndexCount, InstanceCount, VertexCount, VertexOffset, Viewport};
 use buffer::IndexBufferView;
 use queue::{Supports, Graphics};
 use super::{AttachmentClear, ClearValue, CommandBuffer, RawCommandBuffer};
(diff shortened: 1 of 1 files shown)


src/hal/src/command/renderpass.rs
```diff
@@ -83,6 +83,31 @@ impl<'a, B: Backend> RenderPassInlineEncoder<'a, B> {
...
+    // TODO: set_line_width
+    // TODO: set_depth_bounds
+    // TODO: set_depth_bias
+    // TODO: set_blend_constants
+    // TODO: set_stencil_compare_mask
+    // TODO: set_stencil_reference
+    // TODO: set_stencil_write_mask
+
+    // TODO: push constants
+    // TODO: pipeline barrier
```
kvark: these ones are rather difficult, let's leave them off for now

msiglreith:
A large chunk of these are only dynamic states in vulkan but baked into the PSO on d3d12. So we need to expose dynamic PSO features and corresponding feature flags.

src/hal/src/command/renderpass.rs
davll: OK

davll:
@msiglreith 
OK, I’ll move these dynamic states to another PR for implementation. I think we can do pso caching in dx12 backend to emulate dynamic states.

kvark:
Thanks!
bors r+

PR #1569, opened 2017-10-13
Dxgi window patches

zakorgy:
@kvark These are the small changes we made and mentioned before.
- An `init_existing_raw` function for dxgi window, to use an already initialized window.
- We added reference counting for the inner (winit window) member of the dxgi window, because we need to use the winit window after `init_existing_raw` is done.

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1569)

### DIFF
--- a/src/window/dxgi/src/lib.rs
+++ b/src/window/dxgi/src/lib.rs
@@ -30,7 +30,7 @@ use device_dx11::{Device, Factory, Resources};
 
 
 pub struct Window {
-    inner: winit::Window,
+    pub inner: winit::Window,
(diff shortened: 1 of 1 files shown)


kvark:
> because we need to use the winit window after init_existing_raw is done.

Did you consider just exposing the `inner` window in public? I guess that would solve the need for `Rc`.

zakorgy:
Yes, that would be much simpler. I will check, how we can work with that, and update the PR for it.

zakorgy:
@kvark I have updated the PR

kvark:
Thanks @zakorgy !

Issue #1574, opened 2017-10-15
assertion failed: self.dimensions.map(|d| d == dim).unwrap_or(true) when rendering to off-screen targets


norru:
I have tracked this assertion down to a possible bug we discussed on the https://gitter.im/gfx-rs/gfx, I only need this as a place to track changes.

``thread 'main' panicked at 'assertion failed: self.dimensions.map(|d| d == dim).unwrap_or(true)', /u01/home/nico/Projects/3rdParty/gfx/src/core/src/pso.rs:297:8`

- set_dimensions is initially None
- then in `add_color` it gets set to Some(1024, 1024, 0, 1)
- then in `add_depth_stencil` is set to Some(1024, 1024, 1, 1)

It appears that the (width, height, depth) of the on-screen framebuffer is (w, h, 1) while it's set to (w, h, 0...) for off-screen targets. As the depth buffer is created with the initial window, it is consistent with the former but mismatched when used with another target,

This inconsistency appears to be introduced in `texture.rs`:
```rust
impl Kind {
    /// Get texture dimensions, with 0 values where not applicable.
    pub fn get_dimensions(&self) -> Dimensions {
        let s0 = AaMode::Single;
        match *self {
            Kind::D1(w) => (w, 0, 0, s0),
            Kind::D1Array(w, a) => (w, 0, a as Size, s0),
            Kind::D2(w, h, s) => (w, h, 0, s),
            Kind::D2Array(w, h, a, s) => (w, h, a as Size, s),
            Kind::D3(w, h, d) => (w, h, d, s0),
            Kind::Cube(w) => (w, w, 6, s0),
            Kind::CubeArray(w, a) => (w, w, 6 * (a as Size), s0)
        }
    }
   ...
```
Replacing `Kind::D2(w, h, s) => (w, h, 0, s)` with `Kind::D2(w, h, s) => (w, h, 1, s)` seems to fix the assert with no other ill effects. I'll put together a pull request once I've done with testing.

kvark:
> As the depth buffer is created with the initial window, it is consistent with the former but mismatched when used with another target,

Are you testing this on GL?

There is no need to be consistent between mian framebuffer's depth and user's off-screen target because these are never allowed to be used together anyway.

norru:
[...]

PR #1575, opened 2017-10-15
Fixed texture size mismatch https://github.com/gfx-rs/gfx/issues/1574

norru:
Details at https://github.com/gfx-rs/gfx/issues/1574

### DIFF
--- a/src/core/src/texture.rs
+++ b/src/core/src/texture.rs
@@ -265,13 +265,14 @@ pub enum Kind {
 }
 
 impl Kind {
-    /// Get texture dimensions, with 0 values where not applicable.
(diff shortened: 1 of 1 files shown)


norru:
This patch can (and should) be backported to branch v0.16 as well.

kvark:
Thanks for tracking this down!
Would you want to rebase  #1572 on `pre-ll` instead of merging this one?

norru:
@kvark @msiglreith: #1572 does include these changes, and does it better. Probably no need to track this further. I would rather take this change when it gets merged back to a release branch - 0.16.x or 0.17 when it's out.

kvark:
Ok, closing then.

Issue #1579, opened 2017-10-16
Clean up bitflags usage


kvark:
Upon `bitflags` update in #1568, we got a number of constants duplicating the bitflag consts. They should be removed in favor of using the bitflag types directly with their associated consts. This is an easy task for newcomers.

msiglreith:
Remaining parts cleaned up in https://github.com/gfx-rs/gfx/pull/1616

PR #1587, opened 2017-10-18
Update bitflags usage

HeroesGrave:
Started work on #1579 

Crates updated:

- [x] hal
- [x] render
- [x] backend/gl
- [x] backend/vulkan
- [x] backend/dx12 (haven't been able to check if it compiles)
- [x] backend/metal
- [ ] ~~backend/dx11~~
- [x] examples/hal/quad
- [x] examples/render/quad_render

I haven't changed hal yet because it will break the backends that I haven't been able to update (can't build metal/dx11/dx12 on my computer).

I've avoided renaming anything but there are a few cases that I think could be changed:

- `image::Access` -> `ImageAccess`
- `buffer::Access` -> `BufferAccess`
- `image::Usage` -> `ImageUsage`
- `buffer::Usage` -> `BufferUsage` (these first four are to avoid confusion and allow the imports of both the image and buffer variants in cases where they get used together. Alternatively, they can be left as-is and instead some files could use `use image::Access as ImageAccess;`, etc.)
- Remove `ASPECT_` prefix from `image::AspectFlags::*`
- Remove `STAGE_` prefix from `pso::ShaderStageFlags::*`

### DIFF
--- a/src/hal/src/image.rs
+++ b/src/hal/src/image.rs
@@ -559,20 +559,20 @@ bitflags!(
     #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
     pub struct AspectFlags: u8 {
         /// Color aspect.
-        const ASPECT_COLOR = 0x1;
+        const COLOR = 0x1;
         /// Depth aspect.
-        const ASPECT_DEPTH = 0x2;
+        const DEPTH = 0x2;
         /// Stencil aspect.
-        const ASPECT_STENCIL = 0x4;
+        const STENCIL = 0x4;
     }
 );
 
 /// Color aspect.
-pub const ASPECT_COLOR: AspectFlags = AspectFlags::ASPECT_COLOR;
+pub const ASPECT_COLOR: AspectFlags = AspectFlags::COLOR;
 /// Depth aspect.
-pub const ASPECT_DEPTH: AspectFlags = AspectFlags::ASPECT_DEPTH;
+pub const ASPECT_DEPTH: AspectFlags = AspectFlags::DEPTH;
 /// Stencil aspect.
-pub const ASPECT_STENCIL: AspectFlags = AspectFlags::ASPECT_STENCIL;
+pub const ASPECT_STENCIL: AspectFlags = AspectFlags::STENCIL;
 
 bitflags!(
     ///
--- a/src/hal/src/image.rs
+++ b/src/hal/src/image.rs
@@ -349,19 +349,6 @@ bitflags!(
     }
 );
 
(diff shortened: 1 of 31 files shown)


msiglreith:
Don't bother with dx11! It's outdated and its future is unclear.

HeroesGrave:
I don't have any way to build the dependencies for the metal backend so I can't just sift through compiler errors (as I had to do with DX12) to see if I've updated everything to the new bitflags correctly.

Because I haven't removed the old constants it should be fine to merge this after a rebase, and someone with the ability to build Metal could take care of that at some later time.

One last possible change is to deprecate the old constants.

kvark:
I’ll check out metal then

kvark:
@HeroesGrave I verified the current code works on Metal.
Could you rebase please? There is quite a bit of conflicts.

src/hal/src/image.rs
```diff
@@ -559,20 +559,20 @@ bitflags!(
...
-        const ASPECT_DEPTH = 0x2;
+        const DEPTH = 0x2;
         /// Stencil aspect.
-        const ASPECT_STENCIL = 0x4;
+        const STENCIL = 0x4;
     }
 );
 
 /// Color aspect.
-pub const ASPECT_COLOR: AspectFlags = AspectFlags::ASPECT_COLOR;
```
msiglreith: We should remove these imo, now or in a followup.

src/hal/src/pso/mod.rs
```diff
@@ -124,41 +124,41 @@ bitflags!(
...
+        const GRAPHICS = Self::VERTEX.bits | Self::HULL.bits |
+            Self::DOMAIN.bits | Self::GEOMETRY.bits | Self::FRAGMENT.bits;
         /// All shader stages.
-        const STAGE_ALL      = Self::STAGE_GRAPHICS.bits | Self::STAGE_COMPUTE.bits;
+        const ALL      = Self::GRAPHICS.bits | Self::COMPUTE.bits;
     }
 );
 
 ///
-pub const STAGE_VERTEX: ShaderStageFlags = ShaderStageFlags::STAGE_VERTEX;
```
msiglreith: Same as above.

kvark:
@HeroesGrave sorry, one more (small) rebase is needed
@msiglreith we will absolutely get rid of all the constant re-defines, in follow-ups

HeroesGrave:
I decided to bite the bullet and attempt to do the metal backend without the help of the compiler, which enabled me to remove the reexported constants. I have no idea if it will build.

Issue #1539, opened 2017-10-04
Place of GL backend in the new low-level world


kvark:
The GL support in LL is not an obvious question. There are multiple views on it:
  1. Drop it, focus on next-gen only. GL is outdated, poorly emulated via LL frontend, and would be a burden to support.
  2. Have GL as a backup backend, just to have an alternative on each platform. This is especially helpful on platforms where next-gen is not available, including older Android devices and WebGL.
  3. Have a fully featured GL as one of the primary targets. This would make sure that applications like Firefox/Servo keep running fast when WebRender updates from GL to gfx-rs LL.

I'm in favor of 3, given that I want WebRender to not suffer :)

kvark:
One of the pain points of GL comparing to next-gen is context handling. Currently, we have the `Surface` implementing `Instance` and used like this:
```rust
#[cfg(feature = "gl")]
    let (adapters, mut surface) = {
        let surface = back::Surface::from_window(window);
        let adapters = surface.enumerate_adapters();
        (adapters, surface)
    };
```
We expose a single device and a single queue, both associated with the same rendering context. Contrary to other backends, these are not sendable, which restricts what you can do with threading when you want to support it, also crippling our core interfaces a bit.

Here is my proposal for an alternative context handling, at first to be implemented behind a feature gate:
  - have `Instance` created out of thin air, exactly how other backends do it. GL backend would create a headless context for it.
  - surface creation would establish a new GL context created [with shared lists](https://docs.rs/glutin/0.10.0/glutin/struct.ContextBuilder.html#method.with_shared_lists) of the main headless one.
  - initially, the single adapter returned by `enumerate_adapters` would have a single queue family with a single queue, like now. The adapter's device would be constantly bound to the headless context.
  - with each surface created, the adapter would expose one extra family with a single queue. This family would be constantly associated with the GL context of that surface, and support presentation on it.
  - that would allow the users to create multiple surfaces and operate on multiple queues. I believe we could enforce the `Send` on core devices and queues even. There would still be an issue of the queue+device of the headless context going into different threads, which we'll probably just enforce at run-time, so that the compile time interfaces are nice.

kvark:
We don't need multiple issues for effectively the same thing. Let it be a dupe of #1619

PR #1589, opened 2017-10-18
[RFC] Warden test framework

### LINKED ISSUE #1410: Testing infrastructure
Having head-less tests (#583) on CI is great, but that would only test GL backend and only on a single implementation.

We need a set of tests that the devs/users will run on their local machines to ensure the correctness of all backends supported by their machines. Thus, the tests are to be written in a backend-independent way, preferably constructed by a simple readable/writable configuration (in RON format?..). The harness would just run them through all available backends on the local machine, using head-less mode where possible.

Having such tests is crucial for us to keep features working consistently on the way to a fully featured multi-backend implementation.

kvark:
This is a pre-alpha of the new test infrastructure. It's based on [RON](https://github.com/ron-rs/ron)-defined tests that are de-serialized and executed on each available native backend. See [readme](https://github.com/kvark/gfx/tree/warden/src/warden) for ~~devil in the~~ details.

For Unix machines, all you need is `make reftests`.
For Windows (without WSL): `(cd src/warden && cargo run --features vulkan,dx12,logger)`
Output:
```
Warding Vulkan:
        AdapterInfo { name: "Radeon(TM) RX 460 Graphics ", vendor: 4098, device: 26607, software_rendering: false }
        Loading scene 'basic':
                Test 'render-pass-clear' ...    ran: PASS
Warding DX12:
        AdapterInfo { name: "Radeon(TM) RX 460 Graphics", vendor: 4098, device: 26607, software_rendering: false }
        Loading scene 'basic':
                Test 'render-pass-clear' ...    ran: PASS
```

Fixes #1410
Note: it will really shine when:
  1. we get GL support - #1539 
  2. hook Warden up with headless glutin context - #1410 
  3. run on Travis CI

HAL changes:
  - `Instance` specifies `Backend` as an associated type
  - `create_renderpass` -> `create_render_pass` for naming consistency

The PR has a lot of unfinished bits (e.g. PSO creation) but it's functional and technically ready for merging. I'm not 100% confident that being data-driven is a good thing here, hence `RFC` in the subject:
-  (+) RON is more expressive (maps, enum values, etc)
-  (+) somewhat consistent with [Wrench](https://github.com/servo/webrender/tree/master/wrench/reftests)
-  (-) RON is another format/language (note: can do any `serde`-supported format, really)
-  (~) run-time checked instead of compile-time. This is not really a great point since we are supposed to run those tests everywhere all the time, so we shouldn't end up in a situation where someone built them but didn't run.

### DIFF
--- a/src/warden/src/gpu.rs
+++ b/src/warden/src/gpu.rs
@@ -43,7 +43,7 @@ pub struct Scene<B: hal::Backend> {
 }
 
 impl<B: hal::Backend> Scene<B> {
-    pub fn new(adapter: &B::Adapter, raw: &raw::Scene) -> Self {
+    pub fn new(adapter: &B::Adapter, raw: &raw::Scene, data_path: &str) -> Self {
         fn align(x: usize, y: usize) -> usize {
             if x > 0 && y > 0 {
                 ((x - 1) | (y - 1)) + 1
@@ -128,7 +128,7 @@ impl<B: hal::Backend> Scene<B> {
                                 .unwrap();
                             // write the data
                             {
-                                let mut file = File::open(&format!("../../reftests/data/{}", data))
+                                let mut file = File::open(&format!("{}/{}", data_path, data))
                                     .unwrap();
                                 let mut mapping = device.acquire_mapping_writer::<u8>(&upload_buffer, 0..upload_size)
                                     .unwrap();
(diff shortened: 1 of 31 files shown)


torkleyy:
This is awesome!

src/warden/src/gpu.rs
```diff
@@ -0,0 +1,592 @@
...
+            .unwrap();
+
+        let mut command_pool = self.queue.create_graphics_pool(
+            1,
+            hal::pool::CommandPoolCreateFlags::empty(),
+        );
+        let copy_submit = {
+            let mut cmd_buffer = command_pool.acquire_command_buffer();
+            let image_barrier = hal::memory::Barrier::Image {
+                states: image.stable_state .. (i::TRANSFER_READ, i::ImageLayout::TransferSrcOptimal),
```
msiglreith: Image needs to be created with `TRANSFER_SRC` usage.

kvark:
```rust
fn align(x: usize, y: usize) -> usize {
    if x > 0 && y > 0 {
        ((x - 1) | (y - 1)) + 1
    } else {
        x
    }
}
```
It aligns address `x` to alignment `y`.
If `x == 0` or `y == 0` then we return `x`.
else if `x % y == 0` then `(x-1) | (y-1)` would just be `x-1` (assuming `y` is pow2), and then we add `1` which gives us `x` again. Otherwise `(x-1) | (y-1) + 1` gives us the next number that is divisible by `y`.

It may be sub-optimal, or obscure, but it doesn't matter here as long as it's correct. Patches welcome ;)

> Image needs to be created with TRANSFER_SRC usage.

We currently use the exact flags provided by the user. We should probably either add that usage if we know it's needed, or at least check for its presense, as a follow-up.

Also, since there seem to be an agreement over the fact that we need this (:tada:), I'm releasing bors :shipit: 
bors r=msiglreith

kvark:
Hmm, bors didn't get the hint. Let me try again.

bors r=msiglreith

kvark:
Rebased now

bors r=msiglreith

PR #1590, opened 2017-10-19
Fix heap mapping in `Device::allocate_memory` for the dx12 backend

DaseinPhaos:
I came across a tiny nit when tinkering with the `hal/quad` example under dx12, thus this PR

The index used for `heap_properties` should be `mem_type.heap_index` instead of `mem_type.id`.

The fix should make the example workable with dx12!

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -393,8 +393,8 @@ impl d::Device<B> for Device {
 
         let properties = winapi::D3D12_HEAP_PROPERTIES {
             Type: winapi::D3D12_HEAP_TYPE_CUSTOM,
-            CPUPageProperty: self.heap_properties[mem_type.id].page_property,
-            MemoryPoolPreference: self.heap_properties[mem_type.id].memory_pool,
+            CPUPageProperty: self.heap_properties[mem_type.heap_index].page_property,
+            MemoryPoolPreference: self.heap_properties[mem_type.heap_index].memory_pool,
             CreationNodeMask: 0,
             VisibleNodeMask: 0,
         };


kvark:
Makes sense, thank you!
bors r+

msiglreith:
`heap_index` is either 0 or 1 depending on UMA or NUMA.
Heap properties depend on the specific heap types exposed not on the actual memory imo.
What errors are you facing with the dx12 examples?

bors r-

kvark:
nice save! :sweat:

DaseinPhaos:
@msiglreith index out of range
`heap_properties` has a length of 3 but `mem_type.id` goes up to 7 and 8, that's when I thought `heap_index` would be the right one to pass...

Can't access a windows machine to reproduce it atm but I will post some debug info as soon as I can

msiglreith:
I see, should be `mem_type.id % 4` or something like that then. I don't remember the details of the workaround for tier1 heaps atm.

DaseinPhaos:
@msiglreith Thanks for the info! Guess that'd make sense.

So how exactly should we interpret the ids? The code snippet in the memory type initialization part of the dx12 backend seems to be relevant, but I didn't find any documentation about it yet

```rust
        let memory_types = if device.private_caps.heterogeneous_resource_heaps {
            base_memory_types.to_vec()
        } else {
            // the bit pattern of ID becomes 0bTTII, where
            // TT=1 for buffers, TT=2 for images, and TT=3 for targets
            // TT=0 is reserved for future use, helps to avoid ambiguity
            // and II is the same `id` as the `base_memory_types` have
            let mut types = Vec::new();
            for &tt in &[1, 2, 3] {
                types.extend(base_memory_types
                    .iter()
                    .map(|&mt| core::MemoryType {
                        id: mt.id + (tt << 2),
                        .. mt
                    })
                );
            }
            types
        };
```

DaseinPhaos:
So the proper fix for this would be to change from `memory_type.id` into `memory_type.id % 4`.

Other reference to this `id` in the codebase includes:

- comparing this with the `memory::Requirements` type mask: `buffer_req.type_mask & (1 << mem_type.id) != 0`. Should this be changed as well?
- finding an entry in the `StackAllocator`: `let stack = inner.stacks.entry(memory_type.id).or_insert_with(|| ChunkStack::new(memory_type));`, guessing this one won't be affected.

msiglreith:
The issue is dx12 specific, using `id` is correct in the general case. The other two snippts linked are backend agnostic, I believe, so they should stay as is.

DaseinPhaos:
@msiglreith got it, thanks for the explanation! Definitely got a lot more to learn about gfx.

Would it be appropriate for me to file another PR fixing this nit?

msiglreith:
@DaseinPhaos Sure, I would appreciate it!

PR #1595, opened 2017-10-20
Add a example that runs a compute shader on a storage buffer

AlphaModder:
As a complement to our quad example, which demonstrates how to use gfx for graphics, this PR adds an example showcasing the compute functionality.

Specifically, this is a program that takes a bunch of numbers, and prints out their [Collatz conjecture][1] stopping times, performing the calculations on the GPU through a [compute shader][2]. If this is considered too obscure an application and a hindrance to the example, it can be switched out with something else easily.

[1]: https://en.wikipedia.org/wiki/Collatz_conjecture
[2]: https://github.com/AlphaModder/gfx/blob/ll-compute-example/examples/hal/compute/shader/collatz.comp


examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+#[cfg(feature = "vulkan")]
+extern crate gfx_backend_vulkan as back;
+#[cfg(feature = "metal")]
+extern crate gfx_backend_metal as back;
+#[cfg(feature = "gl")]
+extern crate gfx_backend_gl as back;
+#[cfg(feature = "gl")]
+extern crate glutin;
+#[cfg(feature = "gl")]
+use glutin::HeadlessRendererBuilder;
```
kvark: this is for OSMesa runs, it's not going to support compute. You can safely just ignore GL for now.

kvark:
Oh, also, please squash some commits!

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    #[cfg(feature = "metal")]
+    let instance = back::Instance::create();
+
+    let mut gpu = instance.enumerate_adapters().into_iter()
+        .find(|a| a.get_queue_families().into_iter().any(|&(_, t)| t.supports_compute()))
+        .expect("Failed to find a GPU with compute support!")
+        .open_with(|_, qtype| {
+            if qtype.supports_compute() {
+                (1, queue::QueueType::Compute)
+            }
```
msiglreith: nit: formatting:
`} else {`

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+                    ty: pso::DescriptorType::StorageBuffer,
+                    count: 1,
+                    stage_flags: pso::STAGE_COMPUTE,
+                }
+            ],
+        );
+
+        let pipeline_layout = gpu.device.create_pipeline_layout(&[&set_layout]);
+        let entry_point = pso::EntryPoint { entry: "main", module: &shader };
+        let pipeline = gpu.device.create_compute_pipelines(&[(entry_point, &pipeline_layout)]).remove(0).expect("Error creating compute pipeline!");
```
msiglreith: Could you split this over multiple lines?
```
let pipeline = gpu.device
    .create_compute_pipelines(..)
    .remove(0)
    .expect(..);
```

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+                    ty: pso::DescriptorType::StorageBuffer,
+                    count: 1,
+                },
+            ],
+        );
+        (pipeline_layout, pipeline, set_layout, desc_pool)
+    };
+
+    // For now this just panics if you didn't pass numbers. Could add proper error handling.
+    let numbers: Vec<u32> = std::env::args().skip(1).map(|s| u32::from_str(&s).unwrap()).collect();
```
msiglreith: Move this to the beginning imo to fail early. And give at least an error message on panic to let the user know how to run the program.

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    let stride = std::mem::size_of::<u32>() as u64;
+
+    let (staging_memory, staging_buffer) = create_buffer(
+        &mut gpu, 
+        memory::CPU_VISIBLE | memory::COHERENT,
+        buffer::TRANSFER_SRC,
+        stride,
+        numbers.len() as u64
```
msiglreith: nit: trailing comman

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+        writer.copy_from_slice(&numbers);
+        gpu.device.release_mapping_writer(writer);
+    }
+
+    let (device_memory, device_buffer) = create_buffer(
+        &mut gpu,
+        memory::DEVICE_LOCAL,
+        buffer::TRANSFER_DST,
+        stride,
+        numbers.len() as u64
```
msiglreith: (as above)

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+            set: &desc_set,
+            binding: 0,
+            array_offset: 0,
+            write: pso::DescriptorWrite::StorageBuffer(vec![(&device_buffer, 0..stride * numbers.len() as u64)])
+        }
+    ]);
+
+    let mut command_pool = gpu.compute_queues[0].create_compute_pool(16, pool::CommandPoolCreateFlags::empty());
+    let fence = gpu.device.create_fence(false);
+    let submission = queue::Submission::new().submit(&[{
```
msiglreith: heh, interesting!

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+        let mut command_buffer = command_pool.acquire_command_buffer();
+        command_buffer.copy_buffer(&staging_buffer, &device_buffer, &[command::BufferCopy { src: 0, dst: 0, size: stride * numbers.len() as u64}]);
+        command_buffer.bind_compute_pipeline(&pipeline);
```
msiglreith: Missing pipeline barriers from TRANSFER to SHADER_WRITE|SHADER_READ for the SSBO/UAV and the other way around after the dispatch imo.

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    gpu.device.destroy_compute_pipeline(pipeline);
+}
+
+fn create_buffer<B: Backend>(gpu: &mut Gpu<B>, properties: memory::Properties, usage: buffer::Usage, stride: u64, len: u64) -> (B::Memory, B::Buffer) {
+    let ty = (&gpu.memory_types).into_iter().find(|memory_type| {
+        memory_type.properties.contains(properties)
+    }).unwrap();
+
+    let memory = gpu.device.allocate_memory(&ty, stride * len).unwrap();
+    let buffer = gpu.device.create_buffer(stride * len, stride, usage).unwrap();
```
msiglreith: The created buffer has certain memory requirements, it might be larger then the size specified and only supports specific heaps. A more robust implementation would be for this example to:
- `create_buffer`
- request memory requirements
- `allocate_memory` with values from the requirements
- `bind_buffer_memory`

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    let buffer = gpu.device.bind_buffer_memory(&memory, 0, buffer).unwrap();
+
+    (memory, buffer)
+}
+
+#[cfg(not(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl")))]
+fn main() {
+    println!("You need to enable one of the API features in order to test the LL");
```
msiglreith: Maybe some thing like `..run this example` better than `test the LL`.

examples/hal/compute/shader/collatz.comp
```diff
@@ -0,0 +1,31 @@
+#version 450
+layout(set = 0, binding = 0) buffer PrimeIndicies {
+    uint[] indicies;
+}; // this is used as both input and output for convenience
+
+layout(local_size_x = 1) in;
```
msiglreith: (nit: I usually prefer such global information on the top after the version)

examples/hal/compute/shader/collatz.comp
```diff
@@ -0,0 +1,31 @@
...
+        }
+        i++;
+    }
+    return i;
+}
+
+void main() {
+    uint index = gl_GlobalInvocationID.x;
+    indicies[index] = collatz_iterations(indicies[index]);
+}
```
msiglreith: nit: newline

examples/hal/compute/shader/collatz.comp
```diff
@@ -0,0 +1,31 @@
+#version 450
+layout(set = 0, binding = 0) buffer PrimeIndicies {
+    uint[] indicies;
```
msiglreith: nit: typo: `indices` instead of `indicies`

examples/hal/compute/main.rs
AlphaModder: Are these the only pipeline barriers necessary?

examples/hal/compute/shader/collatz.comp
AlphaModder: Hm, for some reason that newline is getting eaten when I `git push`.

examples/hal/compute/shader/collatz.comp
```diff
@@ -0,0 +1,31 @@
...
+            n = (3 * n) + 1;
+        }
+        i++;
+    }
+    return i;
+}
+
+void main() {
+    uint index = gl_GlobalInvocationID.x;
+    indicies[index] = collatz_iterations(indicies[index]);
```
msiglreith: Still old spelling, probably won't build.

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+    ]);
+
+    let mut command_pool = gpu.compute_queues[0].create_compute_pool(16, pool::CommandPoolCreateFlags::empty());
+    let fence = gpu.device.create_fence(false);
+    let submission = queue::Submission::new().submit(&[{
+        let mut command_buffer = command_pool.acquire_command_buffer();
+        command_buffer.copy_buffer(&staging_buffer, &device_buffer, &[command::BufferCopy { src: 0, dst: 0, size: stride * numbers.len() as u64}]);
+        command_buffer.pipeline_barrier(
+            Range { start: pso::TOP_OF_PIPE, end: pso::COMPUTE_SHADER },
```
msiglreith: `TOP_OF_PIPE` should be `TRANSFER` as we do a transfer/copy command here.

examples/hal/compute/main.rs
```diff
@@ -0,0 +1,177 @@
...
+                        start: buffer::TRANSFER_WRITE, 
+                        end: buffer::SHADER_READ | buffer::SHADER_WRITE 
+                    },
+                    target: &device_buffer
+                }]);
+        command_buffer.bind_compute_pipeline(&pipeline);
+        command_buffer.bind_compute_descriptor_sets(&pipeline_layout, 0, &[&desc_set]);
+        command_buffer.dispatch(numbers.len() as u32, 1, 1);
+        command_buffer.pipeline_barrier(
+            Range { start: pso::COMPUTE_SHADER, end: pso::BOTTOM_OF_PIPE },
```
msiglreith: `BOTTOM_OF_PIPE` should be `TRANSFER`.

examples/hal/compute/shader/collatz.comp
AlphaModder: Whoops, nice catch!

PR #1596, opened 2017-10-20
Update metal-rs version

### LINKED ISSUE #1591: Use the new safer metal-rs bindings
See https://github.com/gfx-rs/metal-rs/pull/13
This will supposedly involve a lot of changes, but they should be mostly mechanical.
It would be great to confirm the Metal backend running before metal-rs-0.5 is published.
cc @JohnColanduoni

JohnColanduoni:
This brings the Metal backend up to date with the recent `metal-rs` changes in https://github.com/gfx-rs/metal-rs/pull/13. It requires the few changes in https://github.com/gfx-rs/metal-rs/pull/15 so it's WIP, but the quad example is working.

Fixes #1591


kvark:
Love this:
> +235 −317

I suggest we merge metal-rs changes, publish it, then update this PR and merge it.

examples/hal/quad/main.rs
```diff
@@ -63,6 +63,10 @@ const COLOR_RANGE: i::SubresourceRange = i::SubresourceRange {
 #[cfg(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl"))]
 fn main() {
     env_logger::init().unwrap();
+
+    #[cfg(feature = "metal")]
+    let mut autorelease_pool = unsafe { back::AutoreleasePool::new() };
```
kvark: I believe `hal/compute` and `warden` also need to be updated?
I wonder if we could make the `AutoreleasePool` a bit more convenient... say, do `release` in `drop()` for example

kvark: do we have any assertions/checks in place to detect if the autorelease pool is not active?

JohnColanduoni: `NSAutoreleasePool` doesn't provide any API to do so (it would have to be something like an epoch-based memory manager to provide anything helpful anyway). I had a similar issue when writing some bindings for V8 that I solved by passing a reference to the "autorelease pool" equivalent so that the references I returned were always correctly scoped, but I think that's a very heavy solution for this case (where it makes sense to have only one pool, reset per frame) because we'd have to somehow thread the pool through *every* call in the backend. It also doesn't completely solve the unsafety problem here because `NSAutoreleasePool` affects all ObjC objects, whether we allocate them or users do.

As long as there are no references (in the Rust sense) returned from the Metal API live when the pool is reset, we're safe. At the moment we don't use any unsafe code or reference hi-jinks to avoid retaining references that outlive their stack frame in the backend, and I don't think performance will ever demand we start doing that (in particular, modern ObjC compilers will do exactly what we do).

JohnColanduoni: It does release in `drop`, but since releasing the AutoreleasePool at the wrong time can introduce memory unsafety I opted for calling an unsafe `reset` method on the same `back::AutoreleasePool` (which swaps the `NSAutoreleasePool` internally) instead of relying on an implicit drop to determine when that gets called. The unsafety in the `Drop` implementation is the reason I made `AutoreleasePool::new` unsafe, so that at least  you can't shoot yourself in the foot without including *some* related unsafe code.

kvark: > NSAutoreleasePool doesn't provide any API to do so

Yes, but can we have something, given that we are wrapping it? Even if debug-only, would be very helpful.

How do we know right now that `warden` or `compute` example aren't going to crash right away because the pool was not set up?

JohnColanduoni: If you don't set up the pool, you won't crash but you will leak memory and eventually hang (after a few frames Metal will refuse to make more `CommandBuffers` and block). As it is, no use after free/double free is contingent on just a few things:

1. The signatures in the `metal-rs` crate are correct (i.e. return/accept `Foo` and `&FooRef` at the right places, never return/accept `&Foo`).
2. Do not allow any returned `&FooRef`s to outlive the stack frame without being converted to `Foo`s via `into_owned`. Rust makes this hard/impossible to mess up: the only place you can make a mistake without unsafe code is the `Foo::new` methods which return references. These references can technically be unbounded, but unless you assign them to an `&'static FooRef` or put an `&'a FooRef` in a structure they will function correctly and default to the narrowest scope. Even if this isn't followed they are only unsafe if you interact with `NSAutoreleasePool` before they leave the heap.
3. Do not interact with `NSAutoreleasePool` while any of the backend's code is on the stack. The backend doesn't take any closures or function pointers, so this is impossible without explicitly invoking it in the backend.

If you don't want leaks/deadlocks, you just need:

4. Create a `back::AutoreleasePool` before you start interacting with `gfx`, on each thread in which you interact with `gfx`. Call `reset` on it at some point every frame. This can be relaxed to every few frames if necessary, and you can almost certainly get away with not creating it until you start processing frames.

The touchiest one is (1), but I don't think there's any way around it; it's pretty much the best we can do for FFI bindings. At this point Rust is strictly enforcing most of the requirements that the ObjC compiler would, the only exceptions being the edge cases in (2) and mitigations if you break (3).

kvark:
I don't want this to stall any longer. It blocks subsequent work on the metal backend, such as - removing warnings :)

bors r+

PR #1601, opened 2017-10-21
Rename Adapter methods to match conventions

torkleyy:
Not sure if you care about this stuff.

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -102,7 +102,7 @@ fn main() {
     };
 
     for adapter in &adapters {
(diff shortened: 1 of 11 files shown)


kvark:
Thank you!
bors r+

torkleyy:
Should I do the same for other traits?

kvark:
Yes, but please also link to the convention.

torkleyy:
This is the RFC: https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis

PR #1603, opened 2017-10-21
Fix #1602

### LINKED ISSUE #1602: SDL throwing BadMatch with proprietary Nvidia driver on Linux
Requesting a specific OpenGL version when using SDL and the proprietary Nvidia driver on Linux causes this error to be thrown:
```
Could not make GL context current: BadMatch (invalid parameter attributes)
```
Removing the lines setting the GL version solves the problem with Nvidia, but causes Intel to not be able to understand 3.2 shaders. The problem seems to also be solved for Nvidia when removing the following line from `gfx_window_sdl/lib.rs`:
```rust
attr.set_alpha_size(cf.0.get_alpha_stencil_bits());
```
This however causes my Intel machine to display colors incorrectly.

magnonellie:
Setting GL attributes before creating the window fixes #1602. That did however require getting a reference to the video subsystem earlier, hence the new function signatures.

### DIFF
--- a/src/window/sdl/src/lib.rs
+++ b/src/window/sdl/src/lib.rs
@@ -90,28 +90,42 @@ pub type InitOk<Cf, Df> =
 ///
 /// fn main() {
 ///     let sdl = sdl2::init().unwrap();
+///     let video = sdl.video().unwrap();
 ///
-///     let builder = sdl.video().unwrap().window("Example", 800, 600);
+///     let builder = video.window("Example", 800, 600);
 ///     let (window, glcontext, device, factory, color_view, depth_view) =
-///         gfx_window_sdl::init::<Rgba8, DepthStencil>(builder).expect("gfx_window_sdl::init failed!");
+///         gfx_window_sdl::init::<Rgba8, DepthStencil>(&video, builder)
+///         .expect("gfx_window_sdl::init failed!");
 ///
 ///     // some code...
 /// }
 /// ```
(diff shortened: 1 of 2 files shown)


src/window/sdl/src/lib.rs
```diff
@@ -90,48 +90,52 @@ pub type InitOk<Cf, Df> =
...
-        format: sdl2_pixel_format_from_gfx(cf)
-                    .ok_or(InitError::PixelFormatUnsupportedError)?,
-        ..window.display_mode()?
-    };
-    window.set_display_mode((Some(display_mode)))?;
     {
         let depth_total_bits = df.0.get_total_bits();
         let stencil_bits = df.0.get_alpha_stencil_bits();
-        let attr = window.subsystem().gl_attr();
+        let attr = video_subsystem.gl_attr();
```
kvark: I don't understand. Isn't `window.subsystem()` exactly the same thing here as `video_subsystem`?

magnonellie: It is, but the problem is that if you get the subsystem through the window, you have to have already created the window before setting the attributes which is what caused the problem to begin with. :sweat:

kvark: Ah I see now. You moved the `let mut window = builder.opengl().build()?;` below.

kvark:
Thank you!
bors r+

PR #1605, opened 2017-10-22
Accept iterator in `Adapter::open`

torkleyy:
Okay I nearly solved it, but there's one problem in the Vulkan backend:

In order to create the queues, I need a device, but for the device I need queue information, which I can only get from the iterator. This means I have to iterate it twice :(


kvark:
Please hold off this a little bit.
I'm currently refactoring the `adapter.open` and queue family business, which conflicts with this PR and will essentially make it obsolete, if succeeded.
I'll ping back once I get something solid, or give up ;)

kvark:
See #1606 for what I have in mind.

PR #1606, opened 2017-10-23
Revised Adapter, QueueFamily, and CommandPool interaction

### LINKED ISSUE #1578: Clean up `core` usage in favor of `hal`
The leftover from our old code (after #1561): a bunch of `extern crate gfx_hal as core` and `use core::XXX`. We should fix it to use `hal` name. That's a good task for newcomers, if anyone wants to land a hand.

kvark:
Fixes #1578
Closes #1605
Also adds hal/compute to CI

This is an attempt to refactor our queue families so that we can create command pools from them as opposed to the queues. It refactors the way we discover adapters and queue families.

The PR introduces a concept of a `QueueGroup`, which is spawned from a particular queue family. There is `RawQueueGroup` and a stronger-typed `QueueGroup` that is capability-aware. Command pools can be made out of queue groups.

TODO:
- [x] hal
- [x] empty
- [x] vulkan
- [x] dx12
- [x] metal
- [x] gl
- [x] warden
- [x] hal examples
- [x] render + example

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -1,7 +1,7 @@
 #[macro_use]
 extern crate log;
 extern crate ash;
-extern crate gfx_hal as core;
+extern crate gfx_hal as hal;
 #[macro_use]
 extern crate lazy_static;
 extern crate smallvec;
@@ -22,8 +22,8 @@ use ash::{Entry, LoadingError};
 use ash::extensions as ext;
 use ash::version::{EntryV1_0, DeviceV1_0, InstanceV1_0, V1_0};
 use ash::vk;
-use core::memory;
-use core::{Features, Limits, PatchSize, QueueType};
+use hal::memory;
+use hal::{Features, Limits, PatchSize, QueueType};
 use std::{fmt, mem, ptr};
 use std::ffi::{CStr, CString};
 use std::sync::Arc;
@@ -235,7 +235,7 @@ impl Instance {
     }
 }
 
-impl core::Instance for Instance {
+impl hal::Instance for Instance {
     type Backend = Backend;
 
     fn enumerate_adapters(&self) -> Vec<Adapter> {
@@ -251,7 +251,7 @@ impl core::Instance for Instance {
                         .to_owned()
                 };
 
-                let info = core::AdapterInfo {
+                let info = hal::AdapterInfo {
                     name,
                     vendor: properties.vendor_id as usize,
                     device: properties.device_id as usize,
@@ -261,17 +261,14 @@ impl core::Instance for Instance {
                 let queue_families =
                     self.raw.0
                         .get_physical_device_queue_family_properties(device)
-                        .iter()
+                        .into_iter()
                         .enumerate()
-                        .map(|(i, queue_family)| {
-                        (
-                            QueueFamily {
-                                device: device,
-                                family_index: i as u32,
-                                queue_count: queue_family.queue_count,
-                            },
-                            map_queue_type(queue_family.queue_flags),
-                        )
+                        .map(|(i, properties)| {
+                            ProtoQueueFamily {
+                                properties,
+                                device,
+                                index: i as u32,
+                            }
                         }).collect();
 
                 Adapter {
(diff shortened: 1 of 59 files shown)


src/hal/src/lib.rs
```diff
@@ -327,3 +322,30 @@ pub struct Gpu<B: Backend> {
...
+    ) -> Result<QueueFamily<B, C>, ()> {
+        use queue::RawQueueFamily;
+
+        let pos_maybe = self.queue_families
+            .iter()
+            .position(|family| {
+                C::supported_by(family.queue_type()) &&
+                surface.supports_queue(family) &&
+                family.max_queues() != 0
+            });
```
torkleyy: nit: could be replaced with `map` and `ok_or(Err(()))`

src/hal/src/queue/mod.rs
```diff
@@ -31,32 +34,64 @@ pub enum QueueType {
...
-        *self == QueueType::General || *self == QueueType::Graphics
+/// `RawQueueFamily` denotes a group of command queues provided by the backend
+/// with the same properties/type.
+///
+/// *Note*: A backend can expose multiple queue families with the same properties.
+pub trait RawQueueFamily<B: Backend> {
+    /// Returns the type of queues.
+    fn queue_type(&self) -> QueueType;
+    /// Returns maximum number of queues created from this family.
+    fn max_queues(&self) -> usize;
```
torkleyy: Better naming, nice!

src/hal/src/lib.rs
```diff
@@ -327,3 +322,30 @@ pub struct Gpu<B: Backend> {
     /// Memory heaps with their size in bytes.
     pub memory_heaps: Vec<u64>,
 }
+
+impl<B: Backend> Gpu<B> {
+    /// Returns a strongly typed queue family matching the requirements:
+    ///   - supports the given `Capability`
+    ///   - allows presenting onto a given `Surface`
+    pub fn init_queue_family<C: queue::Capability>(
```
msiglreith: I guess we will need a function with a surface also.

kvark:
@msiglreith @torkleyy thank you for taking a quick look, much appreciated!
It turns out to be impractical on Vulkan :( so I had to re-think the architecture a bit. Just pushed the latest code that covers Vulkan backend as well as the hal/quad example. It's not as pretty as the previous solution, but arguably preferable to what we currently have in `master`. Suggestions are welcome!

src/backend/vulkan/src/lib.rs
```diff
@@ -472,31 +436,27 @@ impl core::Adapter<Backend> for Adapter {
...
 
-    fn info(&self) -> &core::AdapterInfo {
+    fn info(&self) -> &hal::AdapterInfo {
         &self.info
     }
 
-    fn queue_families(&self) -> &[(QueueFamily, QueueType)] {
-        &self.queue_families
+    fn list_queue_families(&mut self) -> Vec<ProtoQueueFamily> {
+        mem::replace(&mut self.queue_families, Vec::new())
```
msiglreith: This turned into some sort of 'one-shot' function. Do we want to preserve this or rather clone here?

src/hal/src/adapter.rs
```diff
@@ -2,110 +2,64 @@
...
     /// ```no_run
     /// # extern crate gfx_backend_empty as empty;
     /// # extern crate gfx_hal;
     /// # fn main() {
-    /// use gfx_hal::Adapter;
+    /// use gfx_hal::{Adapter};
     ///
     /// # let adapter: empty::Adapter = return;
-    /// println!("Adapter info: {:?}", adapter.info());
+    /// let gpu = adapter.open(vec![]);
```
msiglreith: IIRC empty number of queues is violating vulkan specs.

src/hal/src/queue/mod.rs
```diff
@@ -31,32 +34,66 @@ pub enum QueueType {
...
-///
-/// *Note*: A backend can expose multiple queue families with the same properties.
-pub trait QueueFamily: 'static {
-    /// Return the number of available queues of this family.
-    // TODO: some backends like d3d12 support infinite software queues (verify)
-    fn num_queues(&self) -> u32;
+impl<B: Backend, C: Capability> QueueFamily<B, C> {
+    /// Create a new strongly typed queue family from a raw one.
+    ///
+    /// *Note*: panics if the family doesn't expose required
```
msiglreith: I would prefer the
```
/// # Panics
///
/// Panics if the family doesn't expose required ..
```
doc style as proposed here (https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md), Fits better here than a note imo.

src/backend/vulkan/src/lib.rs
kvark: The idea is to have unique `ProtoQueueFamily` instances: user gets a list of them, passes some of them back when calling `open`, drops the other. The only non-fitting piece here is `create_command_pool`, which has to accept `&ProtoQueueFamily`, thus allowing the user to pass one of the prototypes not actually involved in the `Adapter::open`. Not sure how to fix that :( suggestions welcome!

PR #1609, opened 2017-10-24
Query surface formats

msiglreith:
Add functionality to query supported surface formats for swapchain creation.

- [x] hal
- [x] vulkan
- [x] dx12
- [x] metal
- [x] gl
- [x] render
- [x] hal/quad
- [x] render/example


src/backend/vulkan/src/conv.rs
```diff
@@ -148,6 +148,203 @@ pub fn map_format(surface: format::SurfaceType, chan: format::ChannelType) -> Op
     })
 }
 
+pub fn map_vk_format(
+    fmt: vk::Format
+) -> Option<(format::SurfaceType, format::ChannelType)> {
+    use ash::vk::Format::*;
+    use core::format::SurfaceType::*;
+    use core::format::ChannelType::*;
+    match fmt {
```
kvark: it's easier to write `Some(match fmt {` here, and then only `return None` in some of the branches

src/backend/vulkan/src/window.rs
```diff
@@ -7,6 +7,7 @@ use ash::vk;
 use ash::extensions as ext;
 
 use hal;
+use core::format::Format;
```
kvark: should be `hal`

src/hal/src/window.rs
```diff
@@ -218,7 +221,20 @@ impl SwapchainConfig {
...
+
+
+    /// Specify the color format for the backbuffer images.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    ///
+    /// ```
+    pub fn with_color_typed<Cf: format::RenderFormat>(mut self) -> Self {
```
kvark: nice, looks like we are establishing a convention
Now I wonder if mapping API should follow (instead of `_raw`)

msiglreith:
Thanks! Merged `capabilities` and `swapchain_formats` into `capabilities_and_formats`

msiglreith:
bors r=kvark

PR #1615, opened 2017-10-26
Reoder CI commands

kvark:
This cleans up our feature build ordering:
  1. no features are enabled, everything is built except `warden`
  2. examples are built with various backends
  3. `warden` is built with all the backends, which triggers a rebuild of `hal` due to required `serialize` feature
  4. test everything
  5. re-test `render` with `mint` feature

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,16 +43,16 @@ endif
 all: check test
 
 check:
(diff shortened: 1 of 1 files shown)


kvark:
This is roughly 20% increase on my laptop. Should be harmless.
bors r+

PR #1616, opened 2017-10-26
So long, draw_state, and thanks for all the fish!

### LINKED ISSUE #1509: The case for draw_state
Now that we are moving to LL the draw state becomes way more verbose. I can hardly think of Piston, for example, using the states we have in LL. Perhaps, Vulkano would want to share something?

Action item: check out Vulkano's states.

Options:
  1. move all LL states into draw_state
  2. detach from draw_state, keep all the states in `gfx::pso`
  3. wait till `gfx_render` is shaped up, perhaps it will need higher level states

kvark:
Closes #1509

TODO:
- [x] hal
- [x] render
- [x] examples
- backends
  - [x] vulkan
  - [x] gl
  - [x] dx12
  - [x] metal
- [ ] (not sure?) ~~rename `pso` module to `state`~~
- [x] rename `serialize` feature to `serde`
- moare states?
  - [ ] multisampling

### DIFF
--- a/src/hal/src/pso/output_merger.rs
+++ b/src/hal/src/pso/output_merger.rs
@@ -1,6 +1,6 @@
 //! Output Merger(OM) stage description.
 
-use {format, state as s};
+use command::StencilValue;
 
 /// A pixel-wise comparison function.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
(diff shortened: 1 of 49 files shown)


src/hal/src/pso/output_merger.rs
```diff
@@ -1,77 +1,288 @@
...
+    OneMinusSrc1Color,
+    Src1Alpha,
+    OneMinusSrc1Alpha,
+}
+
+/// Blending operation.
+#[allow(missing_docs)]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum BlendOp {
```
msiglreith: sweet

Issue #1610, opened 2017-10-25
Vulkan Compatibility


msiglreith:
Meta issue about the current missing features of gfx-rs compared to Vulkan (Core) API:

- Physical Device: 
  - [ ] Adapter limits (easy):
Expose all limits from Vulkan and improve implementation for other backends.

  - [ ] Adapter Info (easy):
Add missing fields from `PhysicalDeviceInfo`

  - [ ] Physical device limits (med):
Move `Limits` from `Device` into `AdapterInfo` following Vulkan.
Depending on adapter limits implementation.
OpenGL only exposes single adapter so we can just create a context here, depending on https://github.com/gfx-rs/gfx/issues/1539#issuecomment-334173505
DX12 might require temporarily create and destroy device again.

  - [ ] Queue families (easy):
Add missing queue family properties (timestampValidBits, minImageTransferGranularity)
minImageTransferGranularity arbitrary for transfer (0, 0, 0) for whole mipmaps
(1, 1, 1) for Compute or Graphics -> compute shader required for DX12 (metal?)

- [ ] Support queue priorities (easy)

- [ ] Support more formats (easy?)

- [ ] Query supported formats for device (easy-med)
Might depend a bit on how other physical device/device queries are implemented

- Command Buffers
  - [ ] Secondary command buffers (hard):
Expose buffer level
Inheritance info on buffer begin
cmd_execute_commands

  - [ ] Buffer usage flags:
OneTimeSubmit (Easy)
Simultaneous usage (Hard?)
RenderPassContinue (Easy)

- [ ] Blit command (hard)
Requires emulation via shaders

- Synchronization
  - [ ] (Memory barriers (hard))
Difficult for global memory barriers due to RESOURCE_STATES not changing, issue?
Queue ownership transfer

  - [ ] Query fence status (Easy?)

  - [ ] (Events)

  - [ ] Subpass self-dependencies (med)

  - [ ] Queue wait idle (med)

- Renderpass
  - [ ] Renderpass image resolves (med)

  - [ ] Renderpass implicit subpass dependencies:
Depends on global memory barriers (hard)

  - [ ] Query render area granularity (easy?)
Lacks a bit of information on this topic
Might require a bit of research in kind of exposed values on different devices (discrete, integrated, tilers)

- Pipeline
  - [ ] Geometry/Tessellation support (hard)

  - [ ] Early fragment test (med)

  - [ ] Pipeline caches (hard)

  - [ ] Pipeline derives (med)

[...]

PR #1618, opened 2017-10-28
hal: Custom queue priorities

msiglreith:
Small feature to allow setting queue priorities on device creation. Only used for vulkan and mainly for portability.

cc #1610


kvark:
Thank you!
bors r+

PR #1620, opened 2017-10-29
Pipeline deriving

msiglreith:
- Add support for deriving pipelines from other pipelines as supported in Vulkan.
- Additional pipeline flag (ie disable optimizations)
- Refactor pipeline descriptors a bit for compute and graphics

Only HAL and Vulkan implemented so far, others require a few adjustment due to the new API.

cc https://github.com/gfx-rs/gfx/issues/1610


msiglreith:
bors r=kvark (reviewed internally)

PR #1621, opened 2017-11-03
dx12: Implement buffer filling command

msiglreith:
Implement the `fill_buffer` command on d3d12.
We use `ClearUnorderedAccessViewUint` for the internal clearing, requiring us to create a UAV for the buffer. Atm we have the same restrictions as for other clear commands to only allow clearing of whole resources. Otherwise we need to dynamically create views on each command (cheap?).

Additional barriers required as D3D12 needs resources to be in unordered access state whereas vulkan needs transfer write. Probably due to implementation internal things (DMA unit vs compute).

(For future reference: `vkCmdFillBuffer` is also supported on transfer queues with the maintenance extension. D3D12 doesn't support it on Copy queues so far.)

Tested and seems to work. 🏁

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -712,11 +712,53 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
 
     fn fill_buffer(
         &mut self,
-        _buffer: &n::Buffer,
-        _range: Range<u64>,
-        _data: u32,
+        buffer: &n::Buffer,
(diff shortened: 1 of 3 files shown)


kvark:
>For future reference: vkCmdFillBuffer is also supported on transfer queues with the maintenance extension. D3D12 doesn't support it on Copy queues so far.

Ugh, that's an important note. Means we can't support the maintenance extension in VkPortability (in full).

PR #1622, opened 2017-11-04
Push constants!

AlphaModder:
This PR adds support for the use of push constants in shaders. 
It consists of three public api additions/changes: an extra parameter to `Device::create_pipeline_layout`, and the addition of `push_graphics_constants` and `push_compute_constants` to appropriate command buffers.

These can be easily implemented on Vulkan and ~~DX12~~ (nevermind, we need shader reflection it seems), I will see about other backends before this is merged, hopefully.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -601,6 +601,16 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     ) {
         unimplemented!()
     }
+    
+    fn push_graphics_constants(
+        &mut self,
+        layout: &n::PipelineLayout,
+        stages: pso::ShaderStageFlags,
+        offset: u32,
+        constants: &[u32],
+    ) {
+        unimplemented!()
+    }
 
     fn end_query(
         &mut self,
@@ -621,6 +631,15 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         &mut self,
         _: pso::PipelineStage,
         _: query::Query<Backend>,
+    ) {
+        unimplemented!()
+    }
+    
+    fn push_compute_constants(
+        &mut self,
+        layout: &n::PipelineLayout,
+        offset: u32,
+        constants: &[u32],
     ) {
         unimplemented!()
     }
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -843,7 +843,7 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
     ) {
         unimplemented!()
     }
-
+    
     fn end_query(
         &mut self,
         _query: Query<Backend>,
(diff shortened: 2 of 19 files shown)


src/backend/vulkan/src/command.rs
```diff
@@ -749,6 +749,42 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             )
         }
     }
+
+    fn push_compute_constants(
+        &mut self,
+        layout: &n::PipelineLayout,
+        stages: pso::ShaderStageFlags,
```
msiglreith: Should be always `COMPUTE` so we don't need this.

AlphaModder: Ah, right.

msiglreith:
Sup, what's the state with this one? Shall we proceed with fixing the build issues and merging the HAL + Vulkan changes so far? D3D12 turns out to be a bit trickier so maybe a separate PR would be good here.

AlphaModder:
@msilglreith Ah, sorry, I've been busy lately. I was hoping to get DX12 it into this one but that seems doubtful, so it's probably best just to merge in the Vulkan changes. I'll fix the issues asap,

msiglreith:
Thanks for the updates! `compute` example needs a small tweak
```
error[E0061]: this function takes 2 parameters but 1 parameter was supplied
  --> compute/main.rs:63:65
   |
63 |         let pipeline_layout = gpu.device.create_pipeline_layout(&[&set_layout]);
   |                                                                 ^^^^^^^^^^^^^^ expected 2 parameters
```
And could you squash a few commits? At least the merge commit shouldn't be gone.

kvark:
Needs a rebase and CI fixes.

AlphaModder:
Alright, this is ready to be merged.

src/backend/gl/src/command.rs
```diff
@@ -624,6 +634,15 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     ) {
         unimplemented!()
     }
+    
+    fn push_compute_constants(
```
kvark: I wonder if we should just make `unimplemented!()` a default implementation on most of the methods (at least, of `Device`). That would save a bit of code in the `empty` device and ease up maintenance during development

src/backend/metal/src/command.rs
```diff
@@ -843,7 +843,7 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
     ) {
         unimplemented!()
     }
-
+
```
kvark: adding spaces here?

src/warden/src/raw.rs
```diff
@@ -61,6 +61,7 @@ pub enum Resource {
     },
     PipelineLayout {
         set_layouts: Vec<String>,
+        push_constant_ranges: Vec<(hal::pso::ShaderStageFlags, Range<u32>)>
```
kvark: nit: missing trailing comma

PR #1626, opened 2017-11-11
Fix spurious error in creating graphics pipelines

keringar:
I came across this while I was working on the OpenGL backend. There's a missing brace causing create_graphics_pipelines to always return an error.

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -285,9 +285,9 @@ impl d::Device<B> for Device {
                     if status != 0 {
                         if !log.is_empty() {
                             warn!("\tLog: {}", log);
+                        } else {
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/device.rs
```diff
@@ -285,9 +285,9 @@ impl d::Device<B> for Device {
                     if status != 0 {
                         if !log.is_empty() {
                             warn!("\tLog: {}", log);
+                        } else {
```
kvark: this is incorrect: if status is success, there can still be error log that we want to print out the log, but this wouldn't mean an error. I believe the old code is correct. Would you mind posting your log and shader that you are trying to link?

kvark:
Closing due to inactivity

PR #1627, opened 2017-11-11
[mtl] Map entry point names if possible

grovesNL:
This removes the requirement to special case `main` entry points for Metal (which is probably going to be pretty common).

Ideally we'll upgrade the other backends to do this as well (i.e. if a reserved word is used as an entry point in GLSL/HLSL), but at the moment I believe SPIRV-Cross only handles renaming MSL entry points.

I'm not sure about the tuple being returned and the new `HashMap` attached to `ShaderModule`. Initially I performed the mapping directly in `compile_shader_library`. But then I thought we should keep the renamed entry points attached to `ShaderModule` to eventually support multiple entry points/caching shader modules.


src/backend/metal/src/device.rs
```diff
@@ -276,29 +303,40 @@ impl Device {
         pipeline.set_input_primitive_topology(primitive_class);
 
         // Shaders
-        let vs_lib = match pipeline_desc.shaders.vertex.module {
-            &n::ShaderModule::Compiled(ref lib) => lib.to_owned(),
+        let (vs_lib, vs_remapped_entries) = match pipeline_desc.shaders.vertex.module {
+            &n::ShaderModule::Compiled {ref library, ref remapped_entry_point_names} => (library.to_owned(), remapped_entry_point_names.to_owned()),
```
kvark: why do we need to clone the entry points here?

src/backend/metal/src/native.rs
```diff
@@ -17,7 +17,7 @@ use spirv_cross::msl;
 /// depend on pipeline layout, in which case the value would become `Compiled`.
 #[derive(Debug)]
 pub enum ShaderModule {
-    Compiled(metal::Library),
+    Compiled { library: metal::Library, remapped_entry_point_names: HashMap<String, String> },
```
kvark: nit: format in multiple lines, like a struct

src/backend/metal/src/device.rs
grovesNL: The lifetime of the hashmap returned in the `Raw` branch is too short, because I want to return a reference in both `match` branches. Take a look at my new commit to see my issue, this avoids the clone but it feels messy. Is there a way to do this without the temporary `_owned` variables?

src/backend/metal/src/native.rs
grovesNL: Fixed. Is there a plan to run `rustfmt` eventually? I tried to run it on this file but there were too many diffs 😃

src/backend/metal/src/device.rs
```diff
@@ -324,10 +330,13 @@ impl Device {
...
 
         let fs_lib = if let Some(fragment_entry) = pipeline_desc.shaders.fragment {
+            let fs_entries_owned;
             let (fs_lib, fs_remapped_entries) = match fragment_entry.module {
-                &n::ShaderModule::Compiled {ref library, ref remapped_entry_point_names} => (library.to_owned(), remapped_entry_point_names.to_owned()),
+                &n::ShaderModule::Compiled {ref library, ref remapped_entry_point_names} => (library.to_owned(), remapped_entry_point_names),
                 &n::ShaderModule::Raw(ref data) => {
-                    self.compile_shader_library(data, &pipeline_layout.res_overrides).unwrap()
+                    let raw = self.compile_shader_library(data, &pipeline_layout.res_overrides).unwrap();
+                    fs_entries_owned = raw.1;
```
kvark: ha, interesting trick

PR #1628, opened 2017-11-13
Implement Error and Display for buffer::CreationError

omni-viral:


src/hal/src/buffer.rs
```diff
@@ -15,6 +15,27 @@ pub enum CreationError {
...
+    }
+}
+
+impl Error for CreationError {
+    fn description(&self) -> &str {
+        match *self {
+            CreationError::Usage(_) =>
+                "Required `Usage` is not supported",
+            CreationError::Other =>
+                "Some other problem",
```
vitvakatu: Lol, I don't think it's appropriate descrtiption here.

omni-viral: I was thinking exacly same. But I just copied it from comment for variant.

PR #1630, opened 2017-11-16
[dx11] texture mapping for reads

kvark:
This is a little hacky, but it's fine for backend-specific stuff (especially in pre-ll).
@zakorgy would be great to get some feedback from you on it. And no rush!


zakorgy:
@kvark thanks for the patch. We managed to read the color data of the main rendertarget with these changes. Hopefully next week we can produce test results on D3D11 with wrench.

kvark:
bors r=msiglreith

Issue #1631, opened 2017-11-16
[RFC] Pipeline layout compatibility in D3D12


msiglreith:
### Recap
Pipeline layouts consist of multiple push constant ranges + descriptor set layouts. Descriptor set bindings and push constant updates are done for a specific layout. Pipeline layouts are compatible up to a certain descriptor set N if they **identical** push constant ranges and **identically defined** descriptor set layouts up to N.

This means when a new layout is used for setting user-data compared to the previous one, all data remains valid which is compatible with the previous layout.

### Example

Pipeline layout A: `[-- Push Constant Range 0 --][- Set 1 - ][--- Set 2 --- ][-- Set 3 --]`

Pipeline layout B: `[-- Push Constant Range 0 --][- Set 1 - ][-- Set 2 --][-- Set 3 --]`

Considering all data has been set for layout A. Now a new pipeline wants to be used which is created with layout B. Therefore, new data has been set with the specified layout.  A and B are compatible up to Set 2 (exclusive). Therefore, only Set 2 and Set 3 need to be updated, order doesn't matter here:

- *BindDescriptorSet(layout B, set 2)*: Push constant 0 and Set 1 remain due to compatibility with previous layout A, Set 2 will be updated, Set 3 is undefined
- *BindDescriptorSet(layout B, set 3)*: Set 3 will be updated, other parts remain valid as a pipeline layout is always compatible with itself.

### D3D12

The equivalent of pipeline layouts are root signatures, supporting more freedom but also are stricter in terms of security. Root signatures are explicitly set in the command buffer and updates apply to the current set root signature. Changing a root signature marks all data is stale and requires updating **every** part of the signature/layout.

## Issues

Emulating the Vulkan behavior requires checking for compatibility between in the root signatures in **Vulkan terms**. This means if two pipeline layouts in Vulkan are meant to be compatible this also means they should be compatible in D3D12. A few issues arrive here:

- Need a way to check if two root signatures use an identically defined descriptor set layout and the amount of compatiblity.
- Push constants in root signature have a stricter layout and specific binding points, along with no overlapping requirements. A pipeline layout could be compatible with two pipelines, which require **different** root signatures in D3D12 due to stricter root constant layouts.

## Proposal

The states set in command buffers are fully localized so tracking across command buffer boundaries are not required. Therefore, we can track the current user data and active root signature, which is also required to do push constant updates and descriptor table bounds independently.

[...]

PR #1632, opened 2017-11-17
d3d12: Refactor pipeline layouts to match Vulkan's compatibility rules.

msiglreith:
Switching root signatures requires rebinding all tables. Rebinding will now be done automatically and data tracked internally to allow support of two different pipeline bind points.

We now cache the user data inside the command buffer and rebind the required parts on draw/dispatch calls. Upcoming push constant support was considered during implementation. This mostly follows the compatibility aspect from the RFC #1631. Actual compatibility check is not required and the user is supposed to update the corresponding parts following vulkan specs. Therefore a valid program in Vulkan will run in D3D12, and an invalid application might still run correctly in D3D12. Corresponding parts of the Vulkan API specified the behavior as implementation specific.

Next step is to unify the `bind_graphics_descriptor_sets`/`bind_compute_descriptor_sets` and `bind_graphics_pipeline`/`bind_compute_pipeline` calls.

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -32,6 +32,98 @@ pub struct RenderPassCache {
     attachment_clears: Vec<AttachmentClear>,
 }
 
+/// Strongly-typed root signature element
+///
+/// Could be removed for an unsafer variant to occupy less memory
+#[derive(Debug, Copy, Clone)]
+enum RootElement {
+    /// Root constant in the signature
+    Constant(u32),
+    /// Descriptor table, storing table offset for the current descriptor heap
+    TableSrvCbvUav(u32),
+    /// Descriptor table, storing table offset for the current descriptor heap
+    TableSampler(u32),
+    /// Undefined value, implementation specific
+    Undefined,
+}
+
+/// Virtual data storage for the current root signature memory.
+#[derive(Clone)]
+struct UserData {
+    data: [RootElement; 64],
+    dirty_mask: u64,
+}
+
+impl UserData {
+    fn new() -> Self {
+        UserData {
+            data: [RootElement::Undefined; 64],
+            dirty_mask: 0,
+        }
+    }
+
+    /// Update root constant values. Changes are marked as dirty.
+    fn set_constants(&mut self, offset: usize, data: &[u32]) {
+        // Each root constant occupies one DWORD
+        for (i, val) in data.iter().enumerate() {
+            self.data[offset+i] = RootElement::Constant(*val);
+            self.dirty_mask |= 1u64 << (offset + i);
+        }
+    }
+
+    /// Update descriptor table. Changes are marked as dirty.
+    fn set_srv_cbv_uav_table(&mut self, offset: usize, table_start: u32) {
+        // A descriptor table occupies one DWORD
+        self.data[offset] = RootElement::TableSrvCbvUav(table_start);
+        self.dirty_mask |= 1u64 << offset;
+    }
+
+    /// Update descriptor table. Changes are marked as dirty.
+    fn set_sampler_table(&mut self, offset: usize, table_start: u32) {
+        // A descriptor table occupies one DWORD
+        self.data[offset] = RootElement::TableSampler(table_start);
+        self.dirty_mask |= 1u64 << offset;
+    }
+
+    /// Clear dirty flag.
+    fn clear(&mut self, i: usize) {
+        self.dirty_mask &= !(1 << i);
+    }
+}
+
+#[derive(Clone)]
+struct PipelineCache {
+    // Bound pipeline and root signature.
+    // Changed on bind pipeline calls.
+    pipeline: Option<(*mut winapi::ID3D12PipelineState, *mut winapi::ID3D12RootSignature)>,
+    // Paramter slots of the current root signature.
+    num_parameter_slots: usize,
+    // Virtualized root signature user data of the shaders
+    user_data: UserData,
+
+    // Descriptor heap gpu handle offsets
+    srv_cbv_uav_start: u64,
+    sampler_start: u64,
+}
+
+impl PipelineCache {
+    fn new() -> Self {
+        PipelineCache {
+            pipeline: None,
+            num_parameter_slots: 0,
+            user_data: UserData::new(),
+            srv_cbv_uav_start: 0,
+            sampler_start: 0,
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+enum BindPoint {
+    Compute,
+    Graphics,
+}
+
 #[derive(Clone)]
 pub struct CommandBuffer {
     raw: ComPtr<winapi::ID3D12GraphicsCommandList>,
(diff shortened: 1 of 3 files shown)


msiglreith:
> Next step is to unify the bind_graphics_descriptor_sets/bind_compute_descriptor_sets and bind_graphics_pipeline/bind_compute_pipeline calls.

Actually, I prefer the current way. Unification would require to introducte a `BindPoint` variable, which is no real improvement and we lose a bit of strong-type guarentees (e.g don't bind compute pipeline to graphics bind point).

src/backend/dx12/src/command.rs
```diff
@@ -198,19 +311,119 @@ impl CommandBuffer {
             );
         }
     }
+
+    fn set_gr_bind_point(&mut self) {
+        if self.active_bindpoint != BindPoint::Graphics {
+            // Switch to graphics bind point
+            assert!(self.gr_pipeline.pipeline.is_some(), "No graphics pipeline bound");
+            let (pipeline, _) = self.gr_pipeline.pipeline.unwrap();
```
kvark: just doing `expect` would be more idiomatic (than checking for `is_some()` in an `assert!`)

src/backend/dx12/src/command.rs
```diff
@@ -198,19 +311,119 @@ impl CommandBuffer {
...
+
+        // Flush descriptor tables
+        // Index in the user data array where tables are starting
+        let table_start = 0;
+        let table_slot_start = 0;
+        for i in 0..self.gr_pipeline.num_parameter_slots {
+            let table_index = i + table_start;
+            if ((user_data.dirty_mask >> table_index) & 1) == 1 {
+                let slot = (i + table_slot_start) as _;
+                match user_data.data[i] {
```
kvark: should probably just do `let ptr = match ... {... };`

src/backend/dx12/src/command.rs
```diff
@@ -32,6 +32,98 @@ pub struct RenderPassCache {
...
+
+    /// Update descriptor table. Changes are marked as dirty.
+    fn set_sampler_table(&mut self, offset: usize, table_start: u32) {
+        // A descriptor table occupies one DWORD
+        self.data[offset] = RootElement::TableSampler(table_start);
+        self.dirty_mask |= 1u64 << offset;
+    }
+
+    /// Clear dirty flag.
+    fn clear(&mut self, i: usize) {
```
kvark: should be renamed for clarity, maybe to `clear_dirty`

src/backend/dx12/src/command.rs
```diff
@@ -198,19 +311,119 @@ impl CommandBuffer {
...
+            assert!(self.comp_pipeline.pipeline.is_some(), "No compute pipeline bound");
+            let (pipeline, _) = self.comp_pipeline.pipeline.unwrap();
+            unsafe { self.raw.SetPipelineState(pipeline); }
+            self.active_bindpoint = BindPoint::Compute;
+        }
+
+        self.flush_compute_user_data();
+    }
+
+    fn flush_compute_user_data(&mut self) {
```
kvark: seems to be awfully similar to `flush_gr_user_data`. Would it make sense to share the implementations?

src/backend/dx12/src/command.rs
```diff
@@ -198,19 +311,119 @@ impl CommandBuffer {
...
+            assert!(self.gr_pipeline.pipeline.is_some(), "No graphics pipeline bound");
+            let (pipeline, _) = self.gr_pipeline.pipeline.unwrap();
+            unsafe { self.raw.SetPipelineState(pipeline); }
+            self.active_bindpoint = BindPoint::Graphics;
+        }
+
+        self.flush_gr_user_data();
+    }
+
+    fn flush_gr_user_data(&mut self) {
```
kvark: naming is a bit inconsistent ("gr" versus "compute")

src/backend/dx12/src/command.rs
```diff
@@ -630,23 +861,55 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+
                 table_id += 1;
             });
         }
     }
 
     fn bind_compute_pipeline(&mut self, pipeline: &n::ComputePipeline) {
         unsafe {
+            match self.comp_pipeline.pipeline {
+                Some((_, signature)) if signature == pipeline.signature => {
```
kvark: just to clarify: so if the current pipeline is graphics, but the last compute pipeline matches this one, we still don't need to rebind anything?

msiglreith: Pipelines only get bound on draw/dispatch calls to avoid redundant rebinds. We only need/know which pipeline to bind at these calls.
e.g
```
BindGraphicsPipeline() // the pipeline bind here would be useless
... // Fill graphics user data
BindComputePipeline()
... // Fill compute user data
dispatch()
dispatch()
draw()
```

msiglreith: I will add a comment as it's sensible question

src/backend/dx12/src/command.rs
msiglreith: hmm, I can try to unify them (I remember that we have something similar done prior).

src/backend/dx12/src/command.rs
msiglreith: Ok, now I think I understood the question correctly. As far as I can see from the dx12 docs, the graphics and compute root signature are independent so they shouldn't affect each other. The current pipeline state is independent from the root signature stuff afaik.

> If a root signature is changed on a command list, all previous root signature bindings become stale and all newly expected bindings must be set before Draw/Dispatch

msiglreith:
Should be ready 🚀 
Appveyor failure seems to be unrelated to this PR.

kvark:
thank you!
bors r+

msiglreith:
Looks like bors crashed
bors r=kvark

kvark:
Perhaps, we need to disable MSVC/nightly build for now...

kvark:
(or at least, allow it to fail)

msiglreith:
I would like to try again, giving that it built correctly on the `staging` branch. Due to new merges it might work now
bors r+

msiglreith:
bors keeps crashing 😕 
```
{{:badmatch, {:error, :push}},
 [{BorsNG.GitHub, :push!, 3,
   [file: 'lib/github.ex', line: 47]},
  {BorsNG.Worker.Batcher, :complete_batch, 3,
   [file: 'lib/batcher.ex', line: 375]},
  {BorsNG.Worker.Batcher, :maybe_complete_batch, 1,
   [file: 'lib/batcher.ex', line: 362]},
  {BorsNG.Worker.Batcher, :handle_cast, 2,
   [file: 'lib/batcher.ex', line: 71]},
  {:gen_server, :try_dispatch, 4,
   [file: 'gen_server.erl', line: 616]},
  {:gen_server, :handle_msg, 6,
   [file: 'gen_server.erl', line: 686]},
  {:proc_lib, :init_p_do_apply, 3,
   [file: 'proc_lib.erl', line: 247]}]}
```

Manually merging?

kvark:
we'll give a friendly heads up to @notriddle  and merge

PR #1634, opened 2017-11-20
Specialization constants

msiglreith:
Add support for specialization constants on pipeline creation for HAL, Vulkan and D3D12.
Specialization constants allow to change specially marked constants in the shader at compile time to allow some optimizations.

In contrast to Vulkan, we use a strong-typed API for defining the overrides for specialization constants as this is easier on the user side. Only scalar constants can be overriden adapting it for vulkan portability should be possible without API changes.

On D3D12, these are implemented by shifting the SPIR-V -> HLSL -> DXBC compilation process over to pipeline creation **if** we detect that the shader module contains specialization constants (or push constants as anticipation for future support). On pipeline creation the parsed SPIR-V module gets patched with the specialization constants and compiled.

`quad` example has been upated to use the new feature. The image/quad gets scaled by a specialization constant. Metal <-> Vulkan/D3D12 should now show slightly different sized quads as overridden and default values differ.

~~**WIP** as the branch sits on top of #1632 and spirv-cross changes are outstanding https://github.com/grovesNL/spirv_cross/pull/37~~


msiglreith:
Should be ready now, coming with latest spirv-cross master (thanks again @grovesNL !) including a fix to emit dx12 spaceN for samplers and support for `early_fragment_tests`.
A few steps closer to full Vulkan compatibility!

examples/hal/compute/main.rs
```diff
@@ -61,7 +61,7 @@ fn main() {
         );
 
         let pipeline_layout = gpu.device.create_pipeline_layout(&[&set_layout]);
-        let entry_point = pso::EntryPoint { entry: "main", module: &shader };
+        let entry_point = pso::EntryPoint { entry: "main", module: &shader, specialization: &[] };
```
kvark: great to see `EntryPoint` getting richer semantics!

examples/hal/quad/main.rs
```diff
@@ -206,14 +206,40 @@ fn main() {
...
+                ],
+            },
+            pso::EntryPoint::<back::Backend> {
+                entry: ENTRY_NAME,
+                module: &fs_module,
+                specialization: &[],
+            },
         );
 
         #[cfg(all(feature = "metal", feature = "metal_argument_buffer"))]
```
kvark: we can remove this path for now

src/backend/dx12/src/device.rs
```diff
@@ -122,6 +139,133 @@ impl Device {
...
+
+                for spec_constant in spec_constants {
+                    if let Some(constant) = source
+                        .specialization
+                        .iter()
+                        .find(|c| c.id == spec_constant.constant_id)
+                    {
+                        // Override specialization constant values
+                        unsafe {
+                            let value = match constant.value {
```
kvark: I don't think this `match` is unsafe

src/backend/dx12/src/device.rs
```diff
@@ -769,33 +913,25 @@ impl d::Device<B> for Device {
...
-                    }
-                    None => {
-                        winapi::D3D12_SHADER_BYTECODE {
-                            pShaderBytecode: ptr::null(),
-                            BytecodeLength: 0,
-                        }
-                    }
-                }
+            let build_shader = |source: Option<&pso::EntryPoint<'a, B>>| {
+                let source = if let Some(src) = source {
```
kvark: nit: match would look better

src/backend/vulkan/src/conv.rs
```diff
@@ -918,3 +921,30 @@ pub fn map_blend_op(
...
+pub fn map_specialization_constants(
+    specialization: &[pso::Specialization],
+    data: &mut SmallVec<[u8; 64]>,
+) -> Result<SmallVec<[vk::SpecializationMapEntry; 16]>, io::Error> {
+    specialization
+        .iter()
+        .map(|constant| {
+            let offset = data.len();
+            match constant.value {
+                pso::Constant::Bool(v) => { data.write_u32::<NativeEndian>(v as u32)?; }
```
kvark: nit: could move `?` around the `match`

src/backend/vulkan/src/device.rs
```diff
@@ -265,44 +265,64 @@ impl d::Device<B> for Device {
         let mut info_color_blend_states    = Vec::with_capacity(descs.len());
         let mut info_dynamic_states        = Vec::with_capacity(descs.len());
         let mut color_attachments          = Vec::with_capacity(descs.len());
+        let mut info_specializations       = Vec::with_capacity(descs.len() * 5);
```
kvark: where does `5` come from? let's have a constant or something

src/hal/src/pso/mod.rs
```diff
@@ -190,3 +186,31 @@ pub enum BasePipeline<'a, P: 'a> {
...
+pub struct Specialization {
+    /// Constant identifier in shader source.
+    pub id: u32,
+    /// Value to override specialization constant.
+    pub value: Constant,
+}
+
+/// Scalar specialization constant with value for overriding.
+#[derive(Debug, Clone)]
+pub enum Constant {
```
kvark: heh

examples/hal/quad/main.rs
msiglreith: I will comment them out.

src/backend/dx12/src/device.rs
msiglreith: Dereference pointers is unsafe

msiglreith:
bors r=kvark

PR #1636, opened 2017-11-22
[example][hal-quad] Destroy shader modules immediately after creating PSO

davll:
Actually, it's not an important PR, just for clarification that PSOs copy shader code from ShaderModules instead of borrowing.

It's OK to close it without merging ;)

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -265,6 +265,15 @@ fn main() {
 
     println!("pipelines: {:?}", pipelines);
 
+    // Cleanup unused shader modules
+    #[cfg(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl"))]
+    {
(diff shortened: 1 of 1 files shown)


msiglreith:
Not sure about this one, I prefer the old one because I think it's easier to follow the code when clearly separated into construction and destruction parts.
Would appreciate other opinions on this.

kvark:
I'm slightly more in favor of this PR. Would appreciate more opinions :)

msiglreith:
We could move shader module creation and destruction completely into the pipeline creation block. This would be a bit cleaner imo, also easier to reason about the lifetime of the shader modules.

kvark:
@davll would you want to finish the PR according to the last recommendation?

msiglreith:
Implemented in https://github.com/gfx-rs/gfx/pull/1776

PR #1637, opened 2017-11-22
[WIP][hal] Refine synchronization usage

davll:
# TODOs

- [x] HAL: add `wait_idle` to Device trait
- [x] ~HAL: add `wait_idle` to CommandQueue struct~
- [x] example/hal/quad: Use upload_semaphore for uploading sync
- [x] example/hal/quad: Use render_semaphore for buffer swapping sync
- [x] example/hal/quad: Remove frame_fence completely
- [x] Vulkan backend
- [x] Empty backend
- [x] Fix CI build errors

# Why Device::wait_idle?

1. It's required to make sure no command buffer is using before calling CommandPool::reset
2. GPU resources modification (ex: Texture deallocation) should not impact GPU execution
3. frame_fence is signaled when submission is completed but the command buffer is not released => CommandPool::reset will crash

# Why not Device::wait_idle?

1. Only Vulkan provides the function (and GL equivalent: glFinish)
2. Game engines don't like it since it hurts performance => too many CPU-wait GPU-idle

# Alternative

Double/Triple-buffered command pools and semaphores => ex: one command pool each frame and do buffer swapping

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -411,8 +411,11 @@ fn main() {
         depth: 0.0 .. 1.0,
     };
 
-    let mut frame_semaphore = device.create_semaphore();
-    let mut frame_fence = device.create_fence(false); // TODO: remove
+    let frame_semaphore = device.create_semaphore();
+    let frame_fence = device.create_fence(true);
+    let render_semaphore = device.create_semaphore();
+    let mut upload_semaphore = Some(device.create_semaphore());
+    let mut used_semaphores = Vec::new();
 
     // copy buffer to texture
     {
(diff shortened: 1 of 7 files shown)


msiglreith:
Thanks, we definitely want `wait_idle` to be feature complete with Vulkan.
The noted alternative sounds more like an approach for a higher level library (e.g render).
Could you keep the old fence based synchronization? I think it's a nicer starting point than `wait_idle` as you mentioned. That `// TODO` was unforunate, probably some leftover from the first versions of this example :D

davll:
OK, I'll revert the quad example back and move my version to new example

msiglreith:
Closing it as we have `wait_idle` now via https://github.com/gfx-rs/gfx/pull/1773

PR #1639, opened 2017-11-22
metal: Implement ConstantBuffers

scoopr:


src/backend/metal/src/command.rs
```diff
@@ -599,6 +599,17 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
                                         }
                                     }
                                 }
+                                ConstantBuffer(ref buffers) => {
+                                    if let EncoderState::Render(ref encoder) = inner.encoder_state {
```
kvark: looks like we'll miss a call to `add_buffers` if the encoder is not recording. We should call it before checking the encoder

src/backend/metal/src/command.rs
```diff
@@ -599,6 +599,17 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
...
+                                        for (i, ref bref) in buffers.iter().enumerate() {
+                                            if let Some(ref buffer) = **bref {
+                                                let offset = 0; // TODO: ?
```
kvark: should be stored in the `DescriptorSetBinding` alongside the buffer ID

src/backend/metal/src/command.rs
scoopr: Ah so thats why it was done like that above, I thought it wouldn't matter then.

kvark: I haven't double-checked if it matters, tbh, assumed it does

src/backend/metal/src/command.rs
scoopr: Oh right, in the ConstantBuffer variant itself? Alright then

kvark: Basically, store it besides the buffer. Buffer + offset is like a constant buffer view in D3D12

scoopr:
New attempt

PR #1641, opened 2017-11-23
[pre-ll][dx11] texture slice copy support

kvark:
cc @dati91


kvark:
Confirmed to sort of work :)
bors r+

PR #1643, opened 2017-11-24
Vulkanify swapchain creation

msiglreith:
Remove the requirement for specifying a present queue on swapchain creation. D3D12 requires to hide the present queue inside the device and have a single queue family supporting presentation. This changes aligns swapchain creation more with Vulkan for the sake of the portability layer.

I didn't touch the actual swapchain creation implementations only shift code a bit around.

Old API:
```rust
fn Surface::build_swapchain(&mut self, config: SwapchainConfig, present_queue: &CommandQueue<B, C>);
```
New API:
```rust
fn Device::create_swapchain(&mut self, surface: &mut Surface, config: SwapchainConfig);
```

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -127,7 +127,7 @@ fn main() {
     println!("{:?}", surface_format);
     let swap_config = SwapchainConfig::new()
         .with_color(surface_format);
-    let (mut swap_chain, backbuffer) = surface.build_swapchain(swap_config, &queue);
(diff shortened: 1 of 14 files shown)


kvark:
Somewhat unfortunate that all the `surface` module (in the backends) internals now have to be `pub(crate)` as opposed to just private. But that's minor.
Overall, the change looks good. r=me when ready

msiglreith:
yay, it builds!
bors r=kvark

Issue #1625, opened 2017-11-10
[RFC] Push constants in D3D12


msiglreith:
In Vulkan, push constants are decoupled from the concept of descriptor sets but still part of the pipeline layout. On pipeline layout creation push constants are declared via a range and shader stage flag. Updates of constant values is done separately for each range in the command buffer.

For D3D12 push constants are represented as constant buffers on the shader side and directly inlined into the root signature as _root constants_. Therefore they directly occupy space in the user data (in AMD terms), which is done under the hood in Vulkan.

#### Issues:

- **Mapping of push constant ranges to root constants:** SPIR-V to HLSL translation requires the shaders to annotate register binding and space. Without knowledge of the pipeline layout mapping is not directly possible.
- **Overlapping push constant ranges**: Vulkan allows, in theory for now at least, that push constants can be overlapping and re-interpreted(!) for different shader stages. (cc https://github.com/KhronosGroup/Vulkan-Docs/issues/609)  D3D12 doesn't allow overlapping root signatures.

#### Proposal
The mapping problem can be addressed by splitting the push constant block (can be only one per stage) on SPIR-V -> HLSL translation into its subfields. Each subfield recieves a root constant binding in the root signature. Register binding is determined by `offset / 4`, which is unique and maps well as root constants and push constants are multiple of `u32`. So we have a constant(!) mapping between ranges and bindings. `space` can be set to 0 for all as we don't want overlap (would require additional memory!).

Overlapping requires a few restrictions. We don't want to disallow overlapping at all as two shaders (e.g vertex and fragment) might shader the same push constant block but are split in two different SPIR-V modules. Therefore, we should require a structural similarity of the push constant blocks. This means that only *whole*  subfields can overlap and the must share the same memory representation:

*  Invalid because `a.y` and `b` overlap but not the whole subfield `a` is overlapping (and have different types):
```
layout(push_constant) uniform foo {
    layout(offset=0) vec2 a;
} u_foo;
```
```
layout(push_constant) uniform boo {
    layout(offset=4) int b;
} u_boo;
```

*  Valid because whole subfields are overlapping and same type.
```
layout(push_constant) uniform foo {
    layout(offset=0) vec2 a;
    layout(offset=8) vec2 b;
} u_foo;
```
```
layout(push_constant) uniform boo {
    layout(offset=8) vec2 c;
    layout(offset=12) vec2 d;
} u_boo;
```

[...]

PR #1647, opened 2017-11-25
dx12: Implement root constants

msiglreith:
Implement root constants for the d3d12 backend following https://github.com/gfx-rs/gfx/issues/1631
Push constants will be split into disjunct ranges, which can be mapped to root constants in the root signature. On pipeline creation, the generated HLSL shaders from spirv-cross will be patched to match the root signature.

- [x] Split push constants into disjunct ranges
- [x] Update root signature for root constants support
- [x] Patch generated descriptor set table register spaces
- [x] Patch push constant translation in spirv-cross

~~This PR is still in an early state. Especially, doing the spirv-cross changes will take a bit of time.~~

Closes https://github.com/gfx-rs/gfx/issues/1625 https://github.com/gfx-rs/gfx/issues/1631

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -9,6 +9,10 @@ use smallvec::SmallVec;
 use std::{mem, ptr};
 use std::ops::Range;
 
+// Fixed size of the root signature.
+// Limited by D3D12.
+const ROOT_SIGNATURE_SIZE: usize = 64;
+
 fn get_rect(rect: &com::Rect) -> winapi::D3D12_RECT {
     winapi::D3D12_RECT {
         left: rect.x as i32,
@@ -57,20 +61,21 @@ enum RootElement {
 /// Virtual data storage for the current root signature memory.
 #[derive(Clone)]
 struct UserData {
-    data: [RootElement; 64],
+    data: [RootElement; ROOT_SIGNATURE_SIZE],
     dirty_mask: u64,
 }
 
 impl UserData {
     fn new() -> Self {
         UserData {
-            data: [RootElement::Undefined; 64],
+            data: [RootElement::Undefined; ROOT_SIGNATURE_SIZE],
             dirty_mask: 0,
         }
     }
 
     /// Update root constant values. Changes are marked as dirty.
     fn set_constants(&mut self, offset: usize, data: &[u32]) {
+        assert!(offset + data.len() <= ROOT_SIGNATURE_SIZE);
         // Each root constant occupies one DWORD
         for (i, val) in data.iter().enumerate() {
             self.data[offset+i] = RootElement::Constant(*val);
@@ -80,6 +85,7 @@ impl UserData {
 
     /// Update descriptor table. Changes are marked as dirty.
     fn set_srv_cbv_uav_table(&mut self, offset: usize, table_start: u32) {
+        assert!(offset < ROOT_SIGNATURE_SIZE);
         // A descriptor table occupies one DWORD
         self.data[offset] = RootElement::TableSrvCbvUav(table_start);
         self.dirty_mask |= 1u64 << offset;
@@ -87,6 +93,7 @@ impl UserData {
 
     /// Update descriptor table. Changes are marked as dirty.
     fn set_sampler_table(&mut self, offset: usize, table_start: u32) {
+        assert!(offset < ROOT_SIGNATURE_SIZE);
         // A descriptor table occupies one DWORD
         self.data[offset] = RootElement::TableSampler(table_start);
         self.dirty_mask |= 1u64 << offset;
@@ -353,7 +360,9 @@ impl CommandBuffer {
                     0,
                 )
             },
-            |slot, gpu| unsafe { cmd_buffer.clone().SetGraphicsRootDescriptorTable(slot, gpu); },
+            |slot, gpu| unsafe {
+                cmd_buffer.clone().SetGraphicsRootDescriptorTable(slot, gpu);
+            },
         );
     }
 
@@ -377,8 +386,29 @@ impl CommandBuffer {
                     0,
                 )
             },
-            |slot, gpu| unsafe { cmd_buffer.clone().SetGraphicsRootDescriptorTable(slot, gpu); },
+            |slot, gpu| unsafe {
+                cmd_buffer.clone().SetGraphicsRootDescriptorTable(slot, gpu);
+            },
         );
+    }
+
+    fn push_constants(
+        user_data: &mut UserData,
+        layout: &n::PipelineLayout,
+        offset: u32,
+        constants: &[u32],
+    ) {
+        let num = constants.len() as u32;
+        for root_constant in &layout.root_constants {
+            assert!(root_constant.range.start <= root_constant.range.end);
+            if root_constant.range.start >= offset &&
+               root_constant.range.start < offset+num
+            {
+                let start = (root_constant.range.start-offset) as _;
+                let end = num.min(root_constant.range.end-offset) as _;
+                user_data.set_constants(offset as _, &constants[start..end]);
+            }
+        }
     }
 
     fn flush_user_data<F, G>(
@@ -397,8 +427,7 @@ impl CommandBuffer {
         let num_root_constant = pipeline.root_constants.len();
         let mut cur_index = 0;
         // TODO: opt: Only set dirty root constants?
-        for i in 0..num_root_constant {
-            let root_constant = &pipeline.root_constants[i];
+        for (i, root_constant) in pipeline.root_constants.iter().enumerate() {
             let num_constants = (root_constant.range.end-root_constant.range.start) as usize;
             let mut data = Vec::new();
             for c in cur_index..cur_index+num_constants {
@@ -430,14 +459,14 @@ impl CommandBuffer {
                         pipeline.srv_cbv_uav_start + offset as u64,
                     RootElement::TableSampler(offset) =>
                         pipeline.sampler_start + offset as u64,
-                    _ => {
-                        error!("Unexpected user data element in the root signature ({:?})", user_data.data[table_index]);
+                    other => {
+                        error!("Unexpected user data element in the root signature ({:?})", other);
                         continue
                     }
                 };
                 let gpu = winapi::D3D12_GPU_DESCRIPTOR_HANDLE { ptr };
                 table_update(i as _, gpu);
-                user_data.clear_dirty(i);
+                user_data.clear_dirty(table_index);
             }
         }
     }
@@ -1471,18 +1500,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         offset: u32,
         constants: &[u32],
     ) {
-        let num = constants.len() as u32;
-        for root_constant in &layout.root_constants {
-            if root_constant.range.start >= offset &&
-               root_constant.range.start < offset+num
-            {
-                let start = (root_constant.range.start-offset) as _;
-                let end = num.min(root_constant.range.end-offset) as _;
-                self.gr_pipeline
-                    .user_data
-                    .set_constants(offset as _, &constants[start..end]);
-            }
-        }
+        Self::push_constants(&mut self.gr_pipeline.user_data, layout, offset, constants);
     }
 
     fn push_compute_constants(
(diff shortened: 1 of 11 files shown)


msiglreith:
PR is should be ready, just waiting on a new spirv-cross release atm (I just openend the PR).

src/backend/dx12/src/command.rs
```diff
@@ -341,7 +345,15 @@ impl CommandBuffer {
...
-            |slot, gpu| unsafe { cmd_buffer.SetGraphicsRootDescriptorTable(slot, gpu); },
+            |slot, data| unsafe {
+                cmd_buffer.clone().SetGraphicsRoot32BitConstants(
+                    slot,
+                    data.len() as _,
+                    data.as_ptr() as *const _,
+                    0,
+                )
+            },
+            |slot, gpu| unsafe { cmd_buffer.clone().SetGraphicsRootDescriptorTable(slot, gpu); },
```
kvark: nit: let's have this expression on multiple lines

src/backend/dx12/src/command.rs
```diff
@@ -357,44 +369,75 @@ impl CommandBuffer {
...
         let user_data = &mut pipeline.user_data;
         if user_data.dirty_mask == 0 {
             return
         }
 
-        // TODO: root constant support
+        let num_root_constant = pipeline.root_constants.len();
+        let mut cur_index = 0;
+        // TODO: opt: Only set dirty root constants?
+        for i in 0..num_root_constant {
```
kvark: nit: use `iter().enumerate()`

src/backend/dx12/src/command.rs
```diff
@@ -357,44 +369,75 @@ impl CommandBuffer {
...
-                let slot = (i + table_slot_start) as _;
-                let ptr = match user_data.data[i] {
+                let ptr = match user_data.data[table_index] {
                     RootElement::TableSrvCbvUav(offset) =>
                         pipeline.srv_cbv_uav_start + offset as u64,
                     RootElement::TableSampler(offset) =>
                         pipeline.sampler_start + offset as u64,
                     _ => {
-                        error!("Unexpected user data element in the root signature ({:?})", user_data.data[i]);
+                        error!("Unexpected user data element in the root signature ({:?})", user_data.data[table_index]);
```
kvark: nit: match with `other` instead of `_` and use it here

src/backend/dx12/src/command.rs
```diff
@@ -357,44 +369,75 @@ impl CommandBuffer {
...
                         continue
                     }
                 };
                 let gpu = winapi::D3D12_GPU_DESCRIPTOR_HANDLE { ptr };
-                table_update(slot, gpu);
-                user_data.clear_dirty(table_index);
+                table_update(i as _, gpu);
+                user_data.clear_dirty(i);
```
kvark: should this use `table_index`?

src/backend/dx12/src/command.rs
```diff
@@ -1412,11 +1467,22 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
         constants: &[u32],
     ) {
-        unimplemented!()
+        let num = constants.len() as u32;
+        for root_constant in &layout.root_constants {
+            if root_constant.range.start >= offset &&
+               root_constant.range.start < offset+num
+            {
+                let start = (root_constant.range.start-offset) as _;
+                let end = num.min(root_constant.range.end-offset) as _;
```
kvark: shouldn't this assert/warn if the end is beyond the available range?

src/backend/dx12/src/command.rs
```diff
@@ -1425,7 +1491,18 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                self.comp_pipeline
```
kvark: would be nice to re-use this code

src/backend/dx12/src/device.rs
```diff
@@ -160,31 +163,66 @@ impl Device {
...
+        ast: &mut spirv::Ast<hlsl::Target>,
+        shader_model: hlsl::ShaderModel,
+        layout: &n::PipelineLayout,
+        stage: pso::Stage,
+    ) -> Result<String, d::ShaderError> {
         let mut compile_options = hlsl::CompilerOptions::default();
         compile_options.shader_model = shader_model;
         compile_options.vertex.invert_y = true;
 
+        let stage_flag = match stage {
```
kvark: we should probably implement `From<pso::Stage> for pso::ShaderStageFlags`, if it isn't already

src/backend/dx12/src/device.rs
```diff
@@ -798,17 +840,60 @@ impl d::Device<B> for Device {
...
+        push_constant_ranges: &[(pso::ShaderStageFlags, Range<u32>)],
+    ) -> n::PipelineLayout {
         // Pipeline layouts are implemented as RootSignature for D3D12.
         //
         // Each descriptor set layout will be one table entry of the root signature.
         // We have the additional restriction that SRV/CBV/UAV and samplers need to be
         // separated, so each set layout will actually occupy up to 2 entries!
+        //
+        // Root signature layout:
+        //     Root Constants: Register: Offest/4, Space: 0
```
kvark: 👍

src/backend/dx12/src/device.rs
```diff
@@ -798,17 +840,60 @@ impl d::Device<B> for Device {
...
+        //     DescriptorTable0: Space: 3 (+1) (Sampler)
+        //     DescriptorTable1: Space: 4 (+1) (SrvCbvUav)
+        //     ...
+
+        let root_constants = root_constants::split(push_constant_ranges)
+            .iter()
+            .map(|constant| {
+                RootConstant {
+                    stages: constant.stages,
+                    range: (constant.range.start/4)..(constant.range.end/4),
```
kvark: any checks needed to see if remainder by 4 is not zero?

src/backend/dx12/src/device.rs
```diff
@@ -798,17 +840,60 @@ impl d::Device<B> for Device {
...
+        for root_constant in root_constants.iter() {
+            let mut param = winapi::D3D12_ROOT_PARAMETER {
+                ParameterType: winapi::D3D12_ROOT_PARAMETER_TYPE_32BIT_CONSTANTS,
+                ShaderVisibility: winapi::D3D12_SHADER_VISIBILITY_ALL, //TODO
+                .. unsafe { mem::zeroed() }
+            };
+
+            *unsafe{ param.Constants_mut() } = winapi::D3D12_ROOT_CONSTANTS {
+                ShaderRegister: root_constant.range.start as _,
+                RegisterSpace: 0, // root constants will always be in space0.
```
kvark: nit: could make it a constant with a nice name at the top of the module

src/backend/dx12/src/device.rs
```diff
@@ -1143,31 +1242,7 @@ impl d::Device<B> for Device {
...
-                stage,
-                shader_model,
-                &entry_point.name,
-                shader_code.as_bytes(),
-            )?;
-
-            shader_map.insert(entry_point.name, shader_blob);
-        }
-        Ok(n::ShaderModule::Compiled(shader_map))
+        Ok(n::ShaderModule::Spirv(raw_data.into()))
```
kvark: can we detect here if any push constants are present and go to the `Compiled` fast path?

src/backend/dx12/src/root_constants.rs
```diff
@@ -0,0 +1,302 @@
...
+
+impl RootConstant {
+    fn is_empty(&self) -> bool {
+        self.range.end <= self.range.start
+    }
+
+    // Divide a root constant into two separate ranges depending on the overlap
+    // with another root constant.
+    fn divide(self, other: &RootConstant) -> (RootConstant, RootConstant) {
+        let left = RootConstant {
```
kvark: any checks for ordering? e.g. `self.range.start <= other.range.start` and such

src/backend/dx12/src/root_constants.rs
```diff
@@ -0,0 +1,302 @@
...
+            range: other.range.start..self.range.end,
+        };
+
+        (left, right)
+    }
+}
+
+impl PartialOrd for RootConstant {
+    fn partial_cmp(&self, other: &RootConstant) -> Option<Ordering> {
+        Some(self.range.start.cmp(&other.range.start)
```
kvark: technically if we reorder `stages` to go last, we could derive the `Ord` and `PartialOrd`, right?

src/backend/dx12/src/root_constants.rs
```diff
@@ -0,0 +1,302 @@
...
+}
+
+impl Ord for RootConstant {
+    fn cmp(&self, other: &RootConstant) -> Ordering {
+        self.partial_cmp(other).unwrap()
+    }
+}
+
+impl<'a> From<&'a (pso::ShaderStageFlags, Range<u32>)> for RootConstant {
+    fn from(&(stages, ref range): &'a (pso::ShaderStageFlags, Range<u32>)) -> Self {
```
kvark: where is this one needed?

src/backend/dx12/src/root_constants.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+}
+
+pub fn split(
+    ranges: &[(pso::ShaderStageFlags, Range<u32>)],
+) -> Vec<RootConstant> {
+    // Frontier of unexplored root constant ranges, sorted descending
+    // (less element shifting for Vec) regarding to the start of ranges.
+    let mut ranges = into_vec(ranges);
+    ranges.sort_by(|a, b| a.cmp(b).reverse());
```
kvark: wouldn't this just be `b.cmp(a)`?

src/backend/dx12/src/root_constants.rs
```diff
@@ -0,0 +1,302 @@
...
+                next.range.start = right.range.end;
+                ranges.push(right);
+                if !next.is_empty() {
+                    ranges.push(next);
+                }
+            }
+        } else {
+            disjunct.push(cur);
+        }
+        ranges.sort_by(|a, b| a.cmp(b).reverse());
```
kvark: similarly here

src/backend/dx12/src/root_constants.rs
```diff
@@ -0,0 +1,302 @@
...
+) -> Vec<RootConstant> {
+    // Frontier of unexplored root constant ranges, sorted descending
+    // (less element shifting for Vec) regarding to the start of ranges.
+    let mut ranges = into_vec(ranges);
+    ranges.sort_by(|a, b| a.cmp(b).reverse());
+
+    // Storing resulting disjunct root constant ranges.
+    let mut disjunct = Vec::with_capacity(ranges.len());
+
+    while !ranges.is_empty() {
```
kvark: could start here right away with `while let Some(cur) = ranges.pop()`

src/backend/dx12/src/root_constants.rs
```diff
@@ -0,0 +1,302 @@
...
+                    pso::ShaderStageFlags::VERTEX |
+                    pso::ShaderStageFlags::FRAGMENT,
+                range: 2..3,
+            },
+            RootConstant {
+                stages: pso::ShaderStageFlags::FRAGMENT,
+                range: 3..4,
+            },
+        ];
+        assert_eq!(reference, split(ranges));
```
kvark: neat!

src/backend/dx12/src/device.rs
msiglreith: Hm requires two things:
- Checking if it's allowed to not declare constants buffers from the root signature in the shaders
- Adjust register spaces of descriptor tables to enforce that space0 will always be reserved for root constants even if the pipeline layout doesn't have push constants.

src/backend/dx12/src/root_constants.rs
msiglreith: Not sure what you mean.

src/backend/dx12/src/root_constants.rs
msiglreith: It's used in `into_vec`.

src/backend/dx12/src/root_constants.rs
msiglreith: r-right... :facepalm:

src/backend/dx12/src/device.rs
msiglreith: Hm, I should just adjust the API to interpret the range as multiples of u32.

src/backend/dx12/src/root_constants.rs
kvark: Nvm, I see that `Range` doesn't implement `PartialOrd`. Was thinking about putting `#[derive()]` instead of manual implementation, but it's not possible.

src/backend/dx12/src/root_constants.rs
kvark: Let's move this code into the `into_vec` implementation then. I don't think this `From<>` is a good thing to have here.

msiglreith:
Thanks for the extensive review, I think I addressed all points.
Just noticed that push constant updates are bugged due to an issue in `ash`, but that's slightly out-of-scope for this PR. Requires a version bump in the future.

kvark:
I'm planning on getting the tests for this.
Thanks for addressing the issues!
bors r+

PR #1649, opened 2017-11-28
[pre-ll] [dx11] Add memory::TRANSFER_SRC to main target

zakorgy:
Added the `memory::TRANSFER_SRC` bind flag for the main_color. This is required, if we want to copy the main_color's content into a staging texture with `copy_texture_to_texture_raw`, and read the data from the staging buffer with `map_texture_read` introduced in #1630.

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1649)


kvark:
Thanks!
bors r+

PR #1650, opened 2017-11-29
Add an ability to downgrade CommandBuffer mutable reference.

omni-viral:


src/hal/src/command/mod.rs
```diff
@@ -47,6 +48,14 @@ impl<'a, B: Backend, C> CommandBuffer<'a, B, C> {
     pub fn finish(self) -> Submit<B, C> {
         unsafe { Submit::new(self) }
     }
+
+    /// Downgrade a command buffer to a lesser capability type.
+    pub fn downgrade<D>(&mut self) -> &mut CommandBuffer<'a, B, D>
```
kvark: where do you need this downgrade?
if you do, it is quite unsafe to do

msiglreith: I hoped for a variant without `transmute` or `unsafe` in general, for example taking only `self` or returning an actual CommandBuffer instead of a reference.

omni-viral: @kvark Why it is unsafe to do?
I need it to pass `&mut CommandBuffer` into object-safe method (no type parameters)

omni-viral: @msiglreith
> returning an actual CommandBuffer instead of a reference

It will allow you to call `finish` and get `Submit<B, D>` while it's incorrect.

kvark: > I thought it should be OK to use &mut CommandBuffer<B, General> as Graphics.

It's not OK, since the `General` type can contain compute commands.
If you know the lesser type in advance, maybe you could work with it from the start?

Alternatively, if you can't enforce the rule at the type level, you'd probably want to work with `B::CommandBuffer`, or at least pass it around instead of stronger-typed variant.

omni-viral: > It's not OK, since the General type can contain compute commands.

It can contain anything it wants. `&mut` allows only record more commands.

kvark:
shipit!
bors r+

PR #1652, opened 2017-11-30
Workaround for dx11 readpixels rowpitch issue

dati91:
The D3D11_MAPPED_SUBRESOURCE's RowPitch value can differ from the expected value (width * texel_size). Most probably it is a driver related thing.

This is just a half solution, because we get the whole pixel data from the texture, including empty pixels. And adding the correct rowpitch as a return value.
At the end it should be better to give back the filtered data, but I didn't had the time to do that.


kvark:
thanks!
bors r+

Issue #1604, opened 2017-10-21
[hal] Traits shouldn't accept slices


torkleyy:
E.g. `Adapter::open` accepts a slice, however we

1) first need to collect elements to
2) then create a slice from the `Vec` in order to
3) let the backend iterate over it and collect filtered and mapped versions of that iterator

Instead, maybe those methods should accept clone-able iterators?

kvark:
I think we briefly touched in with @msiglreith earlier and agreed to move to iterators. Slices served well for simplicity earlier in development, but these days we can't afford the user to collect data into vectors.

kvark:
I believe this is most important for command buffer API, which can be used many times a frame and generally expected to be fast. There is just a few places where we accept slices there, so doing this part should be trivial. `Device` API, on the other hand, is supposed to be heavy, so heap allocation is not as much critical there.

omni-viral:
As a workaround I collect iterator into `SmallVec` to pass values into current API as a slice.

msiglreith:
Done, see referenced commits

PR #1654, opened 2017-11-30
Replace Slices for Iterators in Command Buffer API

AIOOB:
An attempt at #1604, I haven't yet implemented the changes for metal, dx11 & dx12 as I'm on Linux and don't trust myself without a compiler. The formatting is also currently all over the place. This is my first time using rust so I don't know if this is even what you were intending, but I'm happy to have a go at fixing anything.

Are there any more areas where slices should be replaced with iterators? I think I have hit all traits but there are some plain impls that could also be subjected to this (e.g. most of render/src/encoder.rs).

(I may have introduced some bugs, I can't run the examples as I'm also having trouble with vulkan :confounded:)

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -6,6 +6,7 @@ use {conv, native as n, Backend, CmdSignatures};
 use root_constants::RootConstant;
 use smallvec::SmallVec;
 use std::{mem, ptr};
+use std::borrow::Borrow;
 use std::ops::Range;
 
 use winapi::um::d3d12;
@@ -23,6 +24,83 @@ fn get_rect(rect: &com::Rect) -> d3d12::D3D12_RECT {
         top: rect.y as i32,
         right: (rect.x + rect.w) as i32,
         bottom: (rect.y + rect.h) as i32,
+    }
+}
+
+fn bind_descriptor_sets<'a, T>(
+    raw: &ComPtr<d3d12::ID3D12GraphicsCommandList>,
+    pipeline: &mut PipelineCache,
+    layout: &n::PipelineLayout,
+    first_set: usize,
+    sets: T,
+) where
+    T: IntoIterator,
+    T::Item: Borrow<n::DescriptorSet>,
+{
+    let mut sets = sets.into_iter().peekable();
+    let (srv_cbv_uav_start, sampler_start) = if let Some(set_0) = sets.peek().map(Borrow::borrow) {
+        // Bind descriptor heaps
+        unsafe {
+            // TODO: Can we bind them always or only once?
+            //       Resize while recording?
+            let mut heaps = [
+                set_0.heap_srv_cbv_uav.as_mut() as *mut _,
+                set_0.heap_samplers.as_mut() as *mut _
+            ];
+            raw.SetDescriptorHeaps(2, heaps.as_mut_ptr())
+        }
+
+        (set_0.srv_cbv_uav_gpu_start().ptr, set_0.sampler_gpu_start().ptr)
+    } else {
+        return
+    };
+
+    pipeline.srv_cbv_uav_start = srv_cbv_uav_start;
+    pipeline.sampler_start = sampler_start;
+
+    let mut table_id = 0;
+    for table in &layout.tables[..first_set] {
+        if table.contains(n::SRV_CBV_UAV) {
+            table_id += 1;
+        }
+        if table.contains(n::SAMPLERS) {
+            table_id += 1;
+        }
+    }
+
+    let table_base_offset = layout
+        .root_constants
+        .iter()
+        .fold(0, |sum, c| sum + c.range.end - c.range.start);
+
+    for (set, table) in sets.zip(layout.tables[first_set..].iter()) {
+        let set = set.borrow();
+        set.first_gpu_view.map(|gpu| {
+            assert!(table.contains(n::SRV_CBV_UAV));
+
+            let root_offset = table_id + table_base_offset;
+            // Cast is safe as offset **must** be in u32 range. Unable to
+            // create heaps with more descriptors.
+            let table_offset = (gpu.ptr - srv_cbv_uav_start) as u32;
+            pipeline
+                .user_data
+                .set_srv_cbv_uav_table(root_offset as _, table_offset);
+
+            table_id += 1;
+        });
+        set.first_gpu_sampler.map(|gpu| {
+            assert!(table.contains(n::SAMPLERS));
+
+            let root_offset = table_id + table_base_offset;
+            // Cast is safe as offset **must** be in u32 range. Unable to
+            // create heaps with more descriptors.
+            let table_offset = (gpu.ptr - sampler_start) as u32;
+            pipeline
+                .user_data
+                .set_sampler_table(root_offset as _, table_offset);
+
+            table_id += 1;
+        });
     }
 }
 
@@ -506,14 +584,17 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         self.reset();
     }
 
-    fn begin_renderpass(
+    fn begin_renderpass<T>(
         &mut self,
         render_pass: &n::RenderPass,
         framebuffer: &n::Framebuffer,
         target_rect: com::Rect,
-        clear_values: &[com::ClearValue],
+        clear_values: T,
         _first_subpass: com::SubpassContents,
-    ) {
+    ) where
+        T: IntoIterator,
+        T::Item: Borrow<com::ClearValue>,
+    {
         assert_eq!(framebuffer.attachments.len(), render_pass.attachments.len());
         // Make sure that no subpass works with Present as intermediate layout.
         // This wouldn't make much sense, and proceeding with this constraint
@@ -527,18 +608,19 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                 any(|aref| aref.1 == image::ImageLayout::Present)
         }));
 
-        let mut clear_iter = clear_values.iter();
+        let mut clear_iter = clear_values.into_iter();
         let attachment_clears = render_pass.attachments.iter().enumerate().map(|(i, attachment)| {
             AttachmentClear {
                 subpass_id: render_pass.subpasses.iter().position(|sp| sp.is_using(i)),
                 value: if attachment.ops.load == pass::AttachmentLoadOp::Clear {
-                    Some(*clear_iter.next().unwrap())
+                    Some(*clear_iter.next().unwrap().borrow())
                 } else {
                     None
                 },
                 stencil_value: if attachment.stencil_ops.load == pass::AttachmentLoadOp::Clear {
-                    match clear_iter.next() {
-                        Some(&com::ClearValue::DepthStencil(value)) => Some(value.1),
+                    let clear = clear_iter.next().expect("Must provide an addition DepthStencil clear value");
+                    match *clear.borrow() {
+                        com::ClearValue::DepthStencil(value) => Some(value.1),
                         other => panic!("Unexpected clear value: {:?}", other),
                     }
                 } else {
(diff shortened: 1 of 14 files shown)


omni-viral:
> if this is even what you were intending

Exactly what you've done :smile:

msiglreith:
Hi and welcome \o
That's a really nice first contribution!

I can help you with updating the directx backends if needed. Regarding formatting could you use the style for where clauses as written [here](https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/items.md#where-clauses)?

> Are there any more areas where slices should be replaced with iterators? I think I have hit all traits but there are some plain impls that could also be subjected to this

No worry, don't need to catch every occurrence!

> I may have introduced some bugs, I can't run the examples as I'm also having trouble with vulkan

What issues are you encountering? Maybe we can help here.

Regarding the implementation: I would prefer `IntoIterator` instead of `Iterator` bounds as the examples should be mostly untouched if possible. Using `iter::once` doesn't look that nice imo compared to slices. Someone else has a strong opinion on this aspect?

omni-viral:
@msiglreith in case where `IntoIterator::Item` is not a reference you can't use slice fo it.

kvark:
I'm not quite sure that we need the device API changes here... see https://github.com/gfx-rs/gfx/issues/1604#issuecomment-344317995
We get complicated generic signatures where zero overhead is not as important.
@msiglreith do you strongly think we need the iterators everywhere?

omni-viral:
@kvark I think iterators would be helpful in submissions

msiglreith:
> @msiglreith do you strongly think we need the iterators everywhere?

Not necessarily, even though I can understand why it's desired (e.g https://github.com/amethyst/amethyst/pull/463/files#diff-91eb1c104ee09e411d82b2a0cd13cb3cR185)

AIOOB:
I have split this PR into two pieces as suggested by @kvark, one just dealing with the command buffer API (this one), and the other with the more debatable bits (#1665).

msiglreith:
It would like to rediscuss `Iterator` vs `IntoIterator`. Imo `Iterator` is less ergonomic for users and I don't see a issue with having references for `Item`.

[Example usage](https://play.rust-lang.org/?gist=17ae7501436feee2f95ab23e0c0d4cce&version=stable)

kvark:
@Gankro would you enlighten us about the cases where `Iterator` generic argument is better than `IntoIterator`?

AIOOB:
@msiglreith as far as I am aware there is no way to convert an iterator of values to an iterator over references to those values, which would be a problem as everywhere would need to accept references.

Gankro:
The standard library's policy is to always use IntoIterator because there's a blanket "impl<Iter: Iterator> IntoIterator for Iter", so it accepts strictly more that Iterator. The benefit of IntoIterator is that it allows you to pass a slice directly without having to do .iter() -- so it should make the API change completely transparent to clients, modulo inference issues.

I'm not aware of any reason to prefer Iterator, but I wouldn't be shocked if you had a concrete counter-example!

If you have an iterator of references (slice.into_iter()), you can convert it into an iterator of values with `.cloned()`, so an iterator of values isn't a significant burden (but does require slice.iter().cloned() for callers). If you have an iterator of values, there is indeed no way to convert it into an iterator of references (we ran into this a few times in webrender_api).

AIOOB:
I have changed all occurrences of Iterator to IntoIterator as there doesn't seem to be any downside however, it doesn't change very much due to slices giving iterators that iterate over references.

kvark:
Thank you!
Is this still WIP? What's left to do?

src/backend/metal/src/command.rs
```diff
@@ -446,12 +453,17 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
...
-    fn set_viewports(&mut self, vps: &[Viewport]) {
+    fn set_viewports<T>(&mut self, vps: T)
+    where
+        T: IntoIterator<Item=Viewport>,
+    {
+        let mut vps = vps.into_iter();
+        let vp = if let Some(first) = vps.next() {
+            first
+        } else {
+            return
```
AIOOB: This is a change in behaviour, should this panic or exit silently?

AIOOB:
Unless anyone has any objections, I think it is now ready to go!

src/backend/metal/src/command.rs
kvark: let's panic like it's 99th!

msiglreith:
Could we adjust the `Items` of the iterators so that it doesn't make a difference on the user-side using references? In particular for our current examples.

AIOOB:
Does that not go against the point of moving to iterators? It will cause unnecessary collection in the case they have an iterator of values.

msiglreith:
> Does that not go against the point of moving to iterators? It will cause unnecessary collection in the case they have an iterator of values.

All common datastructures (at least in std) implement IntoIterator for the actual datastructures with `Item=T` and for references with `Item=&'a T`. Slices only support Iterators with `Item=&'a T`, which is the motivation for moving towards reference items.

kvark:
Thanks @AIOOB, looks a bit cleaner now! Would you mind squashing before we proceed?

AIOOB:
Thanks @kvark , although not quite yet as I'm still not sure about the changes.
I have changed all the `IntoIterators` to accept references. It affects the implementations less than I thought it would, however, it does cause some extra collections in `render/encoder.rs` (possibly `init_resources` `require_state`, `copy_buffer`, `copy_buffer_to_image` & `require_clear_state`) although I'm not sure how much that would actually affect performance. 
It will also cause extra collection in user's code as in: https://github.com/amethyst/amethyst/pull/463/files#diff-6c10a131c86b189f11d30abfc43dfe79R186. This is the same example as above, `colors` and `depth_stencil` are iterators over values, so will need to be collected in order to iterate over references.

msiglreith:
Are these theoretical issues? I don't see the extra allocations required in the changes.

msiglreith:
I see, my initial assumption was that applying operators like `map` will keep references as items, but that's not always the case. The datastructures usually can be converted into iterators with references, which was the source of my confusion.
Regarding abstraction over references and values, we could introduce an additional generic parameter (e.g `T: Borrow<Item>`) but I'm unsure if this is desired for readability of the API.

AIOOB:
I think that should be most changes, I haven't added any single-element variants of any functions yet as I am unsure which functions it would be appropriate for. There are also probably quite a few available optimisations in `src/render/src/encoder.rs` but they would require some quite large changes to convince the compiler that disjoint struct fields are being mutably borrowed.

kvark:
Awesome, thanks @AIOOB !
> I haven't added any single-element variants of any functions yet as I am unsure which functions it would be appropriate for

As discussed, my proposal about single-element variants doesn't make as much sense for command buffer API (as it does for `Device`), since the elements come dependent on a context, so don't worry about it ;)

> There are also probably quite a few available optimisations in src/render/src/encoder.rs

This shouldn't be a big concern. gfx-render is too young before we may consider it as a use-case worth catering to.

kvark:
@Gankro hmm. We got the by-value iterators in the API now (in this version of the code... sorry about all the rewrite requests @AIOOB !).

> The benefit of IntoIterator is that it allows you to pass a slice directly without having to do .iter() -- so it should make the API change completely transparent to clients, modulo inference issues.

In most cases, we can't pass slices anyway, since they only provide iterators over references. So it appears to me that `Iterator` has the benefit in our case for being a little bit more straight with the semantics.

Please let me know what you think about 1) us getting iterators of values, and 2) using `Iterator<>` bound instead of `IntoIterator`.

Gankro:
It's been a while since I did this stuff, so I walked across the room to ask some of the lang/libs team people who have these tricks more in cache. One option you can consider is:

```
<I>(iter: I) 
  where I: IntoIter,
             I::Item: AsRef<T>,
{
```

`AsRef<T>` (which is in the prelude) is implemented for `T` and `&T`, so this basically makes you accept `Iterator<&T>` and `Iterator<T>` uniformly. Which is to say you can just pass in `&[T]`, `Vec<T>`, or `some_iter.map(make_a_T)`.

If you just want to go forward with by-value stuff, then the only benefit of IntoIter is like, accepting Vec<T> without .into_iter(). But also the downside is pretty low, considering IntoIter is a well-established std idiom.

AIOOB:
This is what it looks like using `std::borrow::Borrow` for the `IntoIterator` items, some cleanup is probably in order once it has been agreed on and squashed. I have used `Borrow` instead of `AsRef` as it is implemented as @Gankro  describes, `AsRef` is similar but as far as I can tell doesn't quite do what is needed, although I am happy to be corrected.

AIOOB:
I think that may be it, the only bit I'm not entirely sure of is https://github.com/gfx-rs/gfx/pull/1654/files#diff-444b87aecfd36d4ac7a3b2b9006537c2R30 as I made some quite large changes and there might be something about closures I'm missing or more likely the style is a bit off. Otherwise, it should be good to go!

src/backend/dx12/src/command.rs
```diff
@@ -26,6 +27,90 @@ fn get_rect(rect: &com::Rect) -> d3d12::D3D12_RECT {
...
+    pipeline: &mut PipelineCache,
+    layout: &n::PipelineLayout,
+    first_set: usize,
+    sets: T,
+) where
+    T: IntoIterator,
+    T::Item: Borrow<n::DescriptorSet>,
+{
+    let mut sets = sets.into_iter();
+    let set_0_borrowable = sets.next().expect("No descriptor sets provided");
```
msiglreith: Maybe something like
```rust
let mut sets = sets.into_iter().peekable();
let (srv_cbv_uav_base,  sampler_base) = if let Some(set0) = sets.peek().map(Borrow::borrow) {
     // bind heaps
    (set_0.srv_cbv_uav_gpu_start(), set_0.sampler_gpu_start())
} else {
    return
};

...
```

Then we wouldn't need the `except` call and the closure at the bottom imo.

src/backend/dx12/src/command.rs
```diff
@@ -527,26 +615,31 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
                     }
                 } else {
                     None
                 },
             }
         }).collect();
-        assert_eq!(clear_iter.next(), None);
+
+        match clear_iter.next() {
+            Some(_) => panic!("Too many clear values provided"),
```
msiglreith: `if let Some(_) = clear_iter.next() { .. }` a bit nicer here.

src/backend/dx12/src/command.rs
```diff
@@ -901,74 +1014,16 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-
-                table_id += 1;
-            });
-        }
+        sets: T,
+    ) where
+        T: IntoIterator,
+        T::Item: Borrow<n::DescriptorSet>,
+    {
+        bind_descriptor_sets(&self.raw, &mut self.gr_pipeline, layout, first_set, sets);
```
msiglreith: nice 👍

src/hal/src/command/renderpass.rs
```diff
@@ -48,7 +51,12 @@ impl<'a, B: Backend> RenderPassInlineEncoder<'a, B> {
     }
 
     ///
-    pub fn clear_attachments(&mut self, clears: &[AttachmentClear], rects: &[Rect]) {
+    pub fn clear_attachments<T, U>(&mut self, clears: T, rects: U)
+    where T: IntoIterator,
+        T::Item: Borrow<AttachmentClear>,
+          U: IntoIterator,
```
msiglreith: nit: indentation

src/render/src/encoder.rs
```diff
@@ -409,18 +410,28 @@ impl<'a, B: Backend, C> Encoder<'a, B, C>
             assert!(src_range.1 <= src.info().size, "out of source bounds");
             assert!(dst_range.1 <= dst.info().size, "out of destination bounds");
             // TODO: check if src == dst
+            self.require_state(
+                PipelineStage::TRANSFER,
+                &[(src, b::Access::TRANSFER_READ),
+                    (dst, b::Access::TRANSFER_WRITE)],
+                &[]);
```
msiglreith: nit: formatting
```rust
self.require_state(
    PipelineStage::TRANSFER,
    &[
        (src, b::Access::TRANSFER_READ),
        (dst, b::Access::TRANSFER_WRITE)
    ],
    &[],
);
```

msiglreith:
Seems to be good to go
bors r+

PR #1656, opened 2017-12-02
vk: android: Add surface creation from native window

msiglreith:


msiglreith:
> any visuals on Android so far?

nop, my smartphone supports android 7 (Vulkan support), but lacks actual drivers.. so I'm currently stuck with stubs
> no Vulkan HAL present, using stub HAL

PR #1659, opened 2017-12-04
[metal] Add specialization constants

grovesNL:
- Removed push constants from quad example so Metal can run again
- Increased minimum Metal version to 1.2 (necessary for function constants), at least for now
- Added support for specialization constants

WIP because it requires https://github.com/gfx-rs/metal-rs/pull/21 to be published first.

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -8,12 +8,10 @@ use std::path::Path;
 use std::sync::{Arc, Mutex};
 use std::{cmp, mem, ptr, slice};
 
-use hal::{self,
-        image, pass, format, mapping, memory, buffer, pso, query};
+use hal::{self, image, pass, format, mapping, memory, buffer, pso, query};
 use hal::device::{WaitFor, BindError, OutOfMemory, FramebufferError, ShaderError, Extent};
 use hal::pool::CommandPoolCreateFlags;
 use hal::pso::{DescriptorSetWrite, DescriptorType, DescriptorSetLayoutBinding, AttributeDesc};
-use hal::pass::{Subpass};
 
 use cocoa::foundation::{NSRange, NSUInteger};
 use metal::{self, MTLFeatureSet, MTLLanguageVersion, MTLArgumentAccess, MTLDataType, MTLPrimitiveType, MTLPrimitiveTopologyClass};
@@ -21,7 +19,6 @@ use metal::{MTLVertexStepFunction, MTLSamplerMinMagFilter, MTLSamplerMipFilter, 
 use foreign_types::ForeignType;
 use objc::runtime::Object as ObjcObject;
 use spirv_cross::{msl, spirv, ErrorCode as SpirvErrorCode};
-
 
 const RESOURCE_HEAP_SUPPORT: &[MTLFeatureSet] = &[
     MTLFeatureSet::iOS_GPUFamily1_v3,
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/device.rs
```diff
@@ -45,6 +42,24 @@ fn gen_parse_error(err: SpirvErrorCode) -> ShaderError {
     ShaderError::CompilationFailed(msg)
 }
 
+fn create_function_constants(specialization: &[pso::Specialization]) -> metal::FunctionConstantValues {
+    let constants_raw = metal::FunctionConstantValues::new();
+    use std::os::raw::c_void;
+    for constant in specialization {
+        unsafe {
+            let tv = match constant.value {
+                pso::Constant::Bool(ref v) => (MTLDataType::Bool, v as *const _ as *const c_void),
```
msiglreith: nit: `as *const _ as *const _` should also be enough

src/backend/metal/src/device.rs
```diff
@@ -45,6 +42,24 @@ fn gen_parse_error(err: SpirvErrorCode) -> ShaderError {
     ShaderError::CompilationFailed(msg)
 }
 
+fn create_function_constants(specialization: &[pso::Specialization]) -> metal::FunctionConstantValues {
+    let constants_raw = metal::FunctionConstantValues::new();
+    use std::os::raw::c_void;
+    for constant in specialization {
+        unsafe {
+            let tv = match constant.value {
```
msiglreith: Could unpack the tuple already here.

src/backend/metal/src/device.rs
```diff
@@ -254,7 +267,7 @@ impl Device {
         debug!("SPIRV-Cross generated shader:\n{}", shader_code);
 
         let options = metal::CompileOptions::new();
-        options.set_language_version(MTLLanguageVersion::V1_1);
+        options.set_language_version(MTLLanguageVersion::V1_2);
```
msiglreith: Does this affect support for older devices? Do we need a fallback?

src/backend/metal/src/device.rs
```diff
@@ -320,14 +333,22 @@ impl Device {
         } else {
             pipeline_desc.shaders.vertex.entry
         };
+
+        let vs_constants = if pipeline_desc.shaders.vertex.specialization.len() == 0 {
```
msiglreith: nit: `is_empty()` instead of `len() == 0`

src/backend/metal/src/device.rs
```diff
@@ -338,14 +359,22 @@ impl Device {
...
                 }
             };
+
             let fs_entry = if fs_remapped_entries.contains_key(fragment_entry.entry) {
                 fs_remapped_entries.get(fragment_entry.entry).unwrap()
             } else {
                 fragment_entry.entry
             };
+
+            let fs_constants = if fragment_entry.specialization.len() == 0 {
```
msiglreith: (same as above)

src/backend/metal/src/device.rs
grovesNL: I'm not sure, I've had trouble locating a comprehensive compatibility table for the Metal language version.

Eventually we'll probably have to request the highest version we can and use supported capability flags, like the OpenGL backend had. We could only request `V1_2` when a specialization constant value is provided, but I don't know if it's worthwhile at the moment.

grovesNL: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf is the table I've been looking at so far, but it lists OS versions only. I don't know much about device compatibility with the various iOS/tvOS/macOS versions, or whether we have a minimum target for OS version.

grovesNL:
Updated to address feedback, thanks!

src/backend/metal/src/device.rs
msiglreith: It seems to require a newer OS version at least:
https://developer.apple.com/documentation/metal/mtllanguageversion/1515361-version1_1
https://developer.apple.com/documentation/metal/mtllanguageversion/1639961-version1_2

kvark:
@grovesNL https://crates.io/crates/metal-rs/0.6.0
Also note `cocoa` updated to 0.13, so you might want to follow here as well.

grovesNL:
Updated `metal-rs` and `cocoa`, should be good to merge if we are okay with the version change for now.

kvark:
Oh, totally!
Thanks :)
bors r+

PR #1660, opened 2017-12-04
[pre-ll] Update glutin 0.11 + winit 0.9, remove trianglell

alexheretic:
- Update glutin, winit
- Remove trianglell example _(I can't run this myself, and I'm assuming it's a unfinished old version of what master is doing)_
- Update image, cgmath dev-dependencies


src/lib.rs
```diff
@@ -127,7 +127,7 @@ A: Sized + ApplicationBase<gfx_device_gl::Resources, gfx_device_gl::CommandBuffe
     let mut events_loop = glutin::EventsLoop::new();
     let (window, mut device, mut factory, main_color, main_depth) =
         gfx_window_glutin::init::<ColorFormat, DepthFormat>(window, context, &events_loop);
-    let (mut cur_width, mut cur_height) = window.get_inner_size_points().unwrap();
+    let (mut cur_width, mut cur_height) = window.get_inner_size().unwrap();
```
kvark: oh my, this inner size story is sad. First, this method was [deprecated](https://docs.rs/winit/0.8.0/winit/struct.Window.html#method.get_inner_size) in favor of the new ones. Now the new ones are deprecated...

alexheretic: Yep, winit is really at the hurting edge of multiplatform code though so I cut them more slack. Tbf in this case the gfx example code is pretty much unaffected.

alexheretic:
@msiglreith @kvark Thanks for reviewing. I'd like a v0.16 compatible release of the glutin/winit bits, would you accept a cherry-picked PR for `v0.16` branch + bumped minor versions for the affected crates?

kvark:
Unfortunately, bumping the minor versions of publicly used 0.x.y crates would be a breaking change... So we can't patch v0.16 unless those dependencies release patched updates (as opposed to minor) as well.

kvark:
We can hurry up and release v0.17 though ;)

alexheretic:
Well I meant; merge into v0.16 branch and bump, for example, `gfx_window_glutin` to `0.19`. The gfx crate itself hasn't changed, the new crates will be compatible with 0.16, while they have undergone a breaking lib change themselves.

kvark:
@alexheretic ah, yes, we should totally do that :) Would you be interested to provide a cherry-picked PR?

kvark:
:(
bors r+

kvark:
hmmm
bors r+

PR #1662, opened 2017-12-05
[metal] Add support for index buffers

omni-viral:


src/backend/metal/src/conversions.rs
```diff
@@ -173,3 +173,10 @@ pub fn map_texture_usage(usage: image::Usage) -> MTLTextureUsage {
     // TODO shader write
     texture_usage
 }
+
+pub fn map_index_type(index_type: IndexType) -> MTLIndexType {
+    match index_type {
+        IndexType::U16 => MTLIndexType::UInt16,
+        IndexType::U32 => MTLIndexType::UInt32,
+    }
+}
```
msiglreith: nit: newline

PR #1663, opened 2017-12-05
[v0.16] Update winit + glutin, bump versions

alexheretic:
Cherry pick of #1660 _without deleting trianglell_.

**winit/glutin updates**
gfx_app `0.7` -> `0.8`
~gfx_device_dx12ll `0.1.1` -> `0.2`~
~gfx_device_vulkanll `0.1` -> `0.2`~
gfx_window_dxgi `0.10` -> `0.11`
gfx_window_glutin `0.18` -> `0.19`
gfx_window_metal `0.2` -> `0.3`
gfx_window_vulkan `0.3` -> `0.4`

**nonbreaking dependency updates**
gfx_device_gl `0.14.5` -> `0.14.6` _(gfx_gl)_
~gfx_core `0.7.2` -> `0.7.3`~
gfx `0.16.1` -> `0.16.2` _(release 427ab42134de4852bb5958b5326792133699c698)_

### DIFF
--- a/src/render/Cargo.toml
+++ b/src/render/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.16.1"
+version = "0.16.2"
 description = "A high-performance, bindless graphics API"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx_app"
-version = "0.7.0"
+version = "0.8.0"
 description = "GFX example application framework"
 homepage = "https://github.com/gfx-rs/gfx"
 keywords = ["graphics", "gamedev"]
@@ -30,7 +30,7 @@ gfx_corell = { path = "src/corell", version = "0.1" }
 gfx = { path = "src/render", version = "0.16" }
 gfx_macros = { path = "src/macros", version = "0.2" }
 gfx_device_gl = { path = "src/backend/gl", version = "0.14" }
-gfx_window_glutin = { path = "src/window/glutin", version = "0.18" }
+gfx_window_glutin = { path = "src/window/glutin", version = "0.19" }
 
 [dependencies.gfx_device_vulkan]
 path = "src/backend/vulkan"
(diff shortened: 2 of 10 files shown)


kvark:
Thank you! A few notes:
1) Can we not include the cgmath bumps here? Even if optional, I'm not certain that this isn't technically a breaking change. I'd say for sure that something like an `unstable` feature is able to break things, but not other features.
2) Don't bother bumping the xxxLL crates versions. They aren't published anyway

alexheretic:
Sure. Can we still release gfx 0.16.2 just for 427ab42134de4852bb5958b5326792133699c698? It's super minor, just avoids a warning for some of my projects, but looks like v0.16 releases won't really be coming after this maybe.

kvark:
Yep, we will!

src/render/Cargo.toml
```diff
@@ -38,5 +38,5 @@ unstable = []
 cgmath = { version = "0.14", optional = true }
 derivative = "1.0"
 draw_state = "0.7"
-gfx_core = { path = "../core", version = "0.7.1" }
```
kvark: I think it specifically requires `0.7.1`, so please don't remove this

alexheretic: It doesn't have that effect though, only `=0.7.1` would have. `0.7.1` means error if not same or semver compatible newer. So it already locked at `0.7.2` and doesn't have much choice as that's what at that path.

I only changed it as specifying the min patch versions doesn't provide much and could be confusing.

kvark: I don't think this is correct.
> only =0.7.1 would have

I don't need exactly 0.7.1, I need stuff introduced in 0.7.1.

> So it already locked at 0.7.2 and doesn't have much choice as that's what at that path.

It's only locked by some of the locks, including the one at the root of the repo. However, if there is a side crate that depends on gfx_render, and it has a lock file with "gfx_core-0.7.0", your new dependency will be satisfied, but it will fail to build.

For this reason, `cargo` may soon start picking the *smallest* compatible version number when testing a crate before publishing instead of the highest.

alexheretic: Right, yeah I guess so! Thanks for explaining.

kvark:
There is a merge commit now at the end. Would you mind squashing or rebasing properly?

alexheretic:
Sorry,  the commit message just amended to a wrong one. The code was still fine. I did it as two commits, one cherry-pick, one version bump. Do you want them squashed?

kvark:
Oh, ok. Thanks!
bors r+

PR #1665, opened 2017-12-06
Replace iterators with slices (Non command buffer)

AIOOB:
A more debatable half of #1604, which APIs use slices and which should use iterators is more up for consideration. My implementation for the Device iterators frequently just collects them anyway which to me suggests there isn't much point (apart from some idea of API uniformity).

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1665)

### DIFF
--- a/src/backend/empty/src/lib.rs
+++ b/src/backend/empty/src/lib.rs
@@ -119,33 +119,35 @@ impl hal::Device<Backend> for Device {
         unimplemented!()
     }
 
-    fn create_render_pass(&self, _: &[pass::Attachment], _: &[pass::SubpassDesc], _: &[pass::SubpassDependency]) -> () {
+    fn create_render_pass<'a ,IA, IS, ID>(&self, _: IA, _: IS, _: ID) -> ()
+    where
+        IA: IntoIterator,
+        IA::Item: Borrow<pass::Attachment>,
+        IS: IntoIterator,
+        IS::Item: Borrow<pass::SubpassDesc<'a>>,
+        ID: IntoIterator,
+        ID::Item: Borrow<pass::SubpassDependency>,
+    {
         unimplemented!()
     }
 
-    fn create_pipeline_layout(&self, _: &[&()], _: &[(pso::ShaderStageFlags, Range<u32>)]) -> () {
+    fn create_pipeline_layout<IS, IR>(&self, _: IS, _: IR) -> ()
+    where
+        IS: IntoIterator,
+        IS::Item: Borrow<()>,
+        IR: IntoIterator,
+        IR::Item: Borrow<(pso::ShaderStageFlags, Range<u32>)>,
+    {
         unimplemented!()
     }
 
-    fn create_graphics_pipelines<'a>(
-        &self,
-        _: &[pso::GraphicsPipelineDesc<'a, Backend>],
-    ) -> Vec<Result<(), pso::CreationError>> {
-        unimplemented!()
-    }
-
-    fn create_compute_pipelines<'a>(
-        &self,
-        _: &[pso::ComputePipelineDesc<'a, Backend>],
-    ) -> Vec<Result<(), pso::CreationError>> {
-        unimplemented!()
-    }
-
-    fn create_framebuffer(
-        &self, _: &(),
-        _: &[&()],
-        _: device::Extent,
-    ) -> Result<(), device::FramebufferError> {
+    fn create_framebuffer<I>(
+        &self, _: &(), _: I, _: device::Extent
+    ) -> Result<(), device::FramebufferError>
+    where
+        I: IntoIterator,
+        I::Item: Borrow<()>,
+    {
         unimplemented!()
     }
 
@@ -196,14 +198,28 @@ impl hal::Device<Backend> for Device {
         unimplemented!()
     }
 
-    fn create_descriptor_pool(&self, _: usize, _: &[pso::DescriptorRangeDesc]) -> DescriptorPool {
-        unimplemented!()
-    }
-    fn create_descriptor_set_layout(&self, _: &[pso::DescriptorSetLayoutBinding]) -> () {
+    fn create_descriptor_pool<I>(&self, _: usize, _: I) -> DescriptorPool
+    where
+        I: IntoIterator,
+        I::Item: Borrow<pso::DescriptorRangeDesc>,
+    {
         unimplemented!()
     }
 
-    fn update_descriptor_sets<R: RangeArg<u64>>(&self, _: &[pso::DescriptorSetWrite<Backend, R>]) {
+    fn create_descriptor_set_layout<I>(&self, _: I) -> ()
+    where
+        I: IntoIterator,
+        I::Item: Borrow<pso::DescriptorSetLayoutBinding>,
+    {
+        unimplemented!()
+    }
+
+    fn update_descriptor_sets<'a, I, R>(&self, _: I)
+    where
+        I: IntoIterator,
+        I::Item: Borrow<pso::DescriptorSetWrite<'a, 'a, Backend, R>>,
+        R: RangeArg<u64>,
+    {
         unimplemented!()
     }
 
@@ -215,13 +231,6 @@ impl hal::Device<Backend> for Device {
         unimplemented!()
     }
 
-    fn reset_fences(&self, _: &[&()]) {
-        unimplemented!()
-    }
-
-    fn wait_for_fences(&self, _: &[&()], _: device::WaitFor, _: u32) -> bool {
-        unimplemented!()
-    }
     fn get_fence_status(&self, _: &()) -> bool {
         unimplemented!()
     }
@@ -497,10 +506,10 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         unimplemented!()
     }
 
-    fn bind_graphics_descriptor_sets<'a, T>(&mut self, _: &(), _: usize, _: T)
+    fn bind_graphics_descriptor_sets<I>(&mut self, _: &(), _: usize, _: I)
     where
-        T: IntoIterator,
-        T::Item: Borrow<()>,
+        I: IntoIterator,
+        I::Item: Borrow<()>,
     {
         unimplemented!()
     }
@@ -509,10 +518,10 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         unimplemented!()
     }
 
-    fn bind_compute_descriptor_sets<'a, T>(&mut self, _: &(), _: usize, _: T)
+    fn bind_compute_descriptor_sets<I>(&mut self, _: &(), _: usize, _: I)
     where
-        T: IntoIterator,
-        T::Item: Borrow<()>,
+        I: IntoIterator,
+        I::Item: Borrow<()>,
     {
         unimplemented!()
     }
@@ -669,10 +678,6 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
 #[derive(Debug)]
 pub struct DescriptorPool;
 impl hal::DescriptorPool<Backend> for DescriptorPool {
-    fn allocate_sets(&mut self, _: &[&()]) -> Vec<()> {
-        unimplemented!()
-    }
-
     fn reset(&mut self) {
         unimplemented!()
     }
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -311,18 +311,24 @@ impl d::Device<B> for Device {
         }
     }
 
-    fn create_render_pass(
-        &self,
-        attachments: &[pass::Attachment],
-        subpasses: &[pass::SubpassDesc],
-        _dependencies: &[pass::SubpassDependency],
-    ) -> n::RenderPass {
+    fn create_render_pass<'a, IA, IS, ID>(
+        &self, attachments: IA, subpasses: IS, _dependencies: ID
+    ) -> n::RenderPass
+    where
+        IA: IntoIterator,
+        IA::Item: Borrow<pass::Attachment>,
+        IS: IntoIterator,
+        IS::Item: Borrow<pass::SubpassDesc<'a>>,
+        ID: IntoIterator,
+        ID::Item: Borrow<pass::SubpassDependency>,
+    {
         let subpasses =
             subpasses
-                .iter()
+                .into_iter()
                 .map(|subpass| {
                     let color_attachments =
                         subpass
+                            .borrow()
                             .colors
                             .iter()
                             .map(|&(index, _)| index)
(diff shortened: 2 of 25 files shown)


src/backend/empty/src/lib.rs
```diff
@@ -97,33 +97,49 @@ impl hal::Device<Backend> for Device {
...
+        U::Item: Borrow<(pso::ShaderStageFlags, Range<u32>)>,
+    {
         unimplemented!()
     }
 
-    fn create_graphics_pipelines<'a>(
-        &self,
-        _: &[pso::GraphicsPipelineDesc<'a, Backend>],
-    ) -> Vec<Result<(), pso::CreationError>> {
+    fn create_graphics_pipeline<'a, T>(&self, _: T) -> Result<(), pso::CreationError>
```
kvark: can't we just have `&pso::GraphicsPipelineDesc<'a, Backend>` here?

src/backend/gl/src/device.rs
```diff
@@ -328,7 +350,8 @@ impl d::Device<B> for Device {
             gl.BindFramebuffer(target, 0);
         }
         if let Err(err) = self.share.check() {
-            panic!("Error creating FBO: {:?} for {:?} with attachments {:?}", err, pass, attachments);
+            //TODO: attachments has been consumed
```
kvark: should be "have been"

src/backend/empty/src/lib.rs
AIOOB: Yes, the generic is completely unnecessary and doesn't provide any benefit.

AIOOB:
I agree that the documentation would be a bit scary, but I'm not sure about API though; another pro is that this change means that the functions accept anything that looks vaguely like an iterator, which should simplify user's code.
However, I can't comment on how useful it actually is and it definitely does add some complexity.

src/backend/vulkan/src/device.rs
```diff
@@ -260,28 +287,34 @@ impl d::Device<B> for Device {
...
-        let mut info_viewport_states       = Vec::with_capacity(descs.len());
-        let mut info_rasterization_states  = Vec::with_capacity(descs.len());
-        let mut info_multisample_states    = Vec::with_capacity(descs.len());
-        let mut info_depth_stencil_states  = Vec::with_capacity(descs.len());
-        let mut info_color_blend_states    = Vec::with_capacity(descs.len());
-        let mut info_dynamic_states        = Vec::with_capacity(descs.len());
-        let mut color_attachments          = Vec::with_capacity(descs.len());
-        let mut info_specializations       = Vec::with_capacity(descs.len() * NUM_STAGES);
-        let mut specialization_data        = Vec::with_capacity(descs.len() * NUM_STAGES);
+        let mut info_stages                = Vec::with_capacity(desc_len_guess);
```
msiglreith: Before I forget it over the holidays.. it crucial in the current implementation that these vectors **don't** reallocate memory because we have pointers into the vector elements floating around.

Rhuagh:
This needs some serious documentation to make it obvious what types are accepted imho.

kvark:
Amazing work! I definitely appreciate the bidirectional implementations and iterators for methods accepting a single list. The `create_renderpass` and `create_pipeline_layout` are arguable.

---

Reviewed 27 of 27 files at r1.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[examples/hal/compute/main.rs, line 145 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CvJFp1miwDJB_r-Zc:-L4CvJFqNhAQ5TZlj06m:b9zfb5u) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/examples/hal/compute/main.rs#L145)):*
> ```Rust
>     }));
>     queue_group.queues[0].submit(submission, Some(&fence));
>     device.wait_for_fence(&fence, !0);
> ```

really nice to have this shortcut

---

*[examples/hal/quad/main.rs, line 164 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CvSGevvHU-WZ-U59N:-L4CvSGevvHU-WZ-U59O:b-kadvh7) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/examples/hal/quad/main.rs#L164)):*
> ```Rust
> 
>     let pipeline_layout = device.create_pipeline_layout(
>         Some(&set_layout),
> ```

nit: could probably also do `[set_layout]`

---

*[src/backend/dx12/src/device.rs, line 444 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CvyWfdnC8_Wy1X6FC:-L4CvyWiExZ3Pu09WCgP:b-bzce9s) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/backend/dx12/src/device.rs#L444)):*
> ```Rust
>     ) where
>         F: FnMut(&pso::DescriptorWrite<B, R>, &mut Vec<d3d12::D3D12_CPU_DESCRIPTOR_HANDLE>),
>         I: Iterator<Item=&'a pso::DescriptorSetWrite<'a, 'a, B, R>>,
> ```

should this be `I: IntoIterator` and `I::Item: Borrow<pso::DescriptorSetWrite>` or something?

---

*[src/backend/dx12/src/device.rs, line 787 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CwAoQwGHEDTKZKO-6:-L4CwAoQwGHEDTKZKO-7:b-dz2c0s) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/backend/dx12/src/device.rs#L787)):*
> ```Rust
>     }
> 
>     fn create_render_pass<'a, IA, IS, ID>(
> ```

maybe we could apply some fair judgement and not try to optimize at least this method? no one is creating render passes on the fly (I hope)

---

*[src/backend/dx12/src/device.rs, line 814 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CwnBewk7led7pkPT9:-L4CwnBfnlLmTMKXp-q7:b9yobug) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/backend/dx12/src/device.rs#L814)):*
> ```Rust
>         }
> 
>         let attachments = attachments.into_iter()
> ```

FTR, doesn't look like generic iterators are making any good for this specific implementation

---

*[src/backend/dx12/src/device.rs, line 939 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CwwEYtyBOQYFdIzFE:-L4CwwEYtyBOQYFdIzFF:buayp8k) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/backend/dx12/src/device.rs#L939)):*
> ```Rust
>     }
> 
>     fn create_pipeline_layout<IS, IR>(
> ```

similarly, I don't expect this to be called often

---

*[src/backend/vulkan/Cargo.toml, line 25 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4Cya3wXutn1NdZ_opx:-L4Cya3xn2pif37lXYOV:b3kl9rd) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/backend/vulkan/Cargo.toml#L25)):*
> ```TOML
> lazy_static = "0.2"
> shared_library = "0.1"
> ash = "=0.20.2"
> ```

this can be reverted now

---

*[src/hal/src/device.rs, line 203 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4Cz1iHo05hMxS3-qMc:-L4Cz1iHo05hMxS3-qMd:b-8xefr1) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L203)):*
> ```Rust
> 
>     /// Create graphics pipelines.
>     fn create_graphics_pipelines<'a, I>(
> ```

strange idea: maybe we can omit this from `gfx-backend-empty` now that we have default implementations? ;)

---

*[src/hal/src/device.rs, line 494 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CzXzvvq9I8IdmDdoO:-L4CzXzwlbn6a5JXx8Ih:b-bwr1rw) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L494)):*
> ```Rust
>             WaitFor::All => {
>                 for fence in fences {
>                     let elapsed_ms = as_ms(start.elapsed());
> ```

I'm a little concerned that his would prevent the expected behavior when the user passes in timeout of `0`. We should still be checking all the fences, but this implementation may bail out in the process

---

*[src/hal/src/device.rs, line 505 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CziooMqguuTf66j-k:-L4CziopWfoN3LPTlBKF:b-juzbfg) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L505)):*
> ```Rust
>             },
>             WaitFor::Any => {
>                 let fences: Vec<_> = fences.into_iter().collect();
> ```

why do we need to collect here?

---

*[src/render/src/device.rs, line 312 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CzuXJVP-erKAaU9T9:-L4CzuXJVP-erKAaU9TA:b-3ofgym) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/render/src/device.rs#L312)):*
> ```Rust
> 
>     #[doc(hidden)]
>     pub fn create_render_pass_raw<'a, IA, IS, ID>(
> ```

don't bother updating the `renderer` with iterators

---

*[src/render/src/pso.rs, line 179 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4D--yZVIpg-UvRE5BF:-L4D--yZVIpg-UvRE5BG:bsbw4ls) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/render/src/pso.rs#L179)):*
> ```Rust
>     }
> 
>     pub fn write<'b, T: Bind<B>, I>(
> ```

similarly, here. Current consensus is that `render` is going away

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-:-L4D-DQ_BIr8xtTAfLdo:b-7pp51a)*

src/hal/src/device.rs
```diff
@@ -411,11 +455,68 @@ pub trait Device<B: Backend> {
...
+                loop {
+                    for fence in &fences {
+                        if self.wait_for_fence(fence.borrow(), 0) {
+                            return true;
+                        }
+                    }
+                    if as_ms(start.elapsed()) >= timeout_ms {
+                        return false;
+                    }
+                    thread::sleep(time::Duration::from_millis(1));
```
msiglreith: I don't think we need issue a thread switch here, drivers will most likely do it internally anyways.

msiglreith:
@kvark I'm in favor of keeping the iterator interface everywhere.
1. Unified API interface
2. ptr -> collecting in `portability` -> reating a slice -> recollecting that slice in `gfx-backend`back into a ptr doesn't feel very nice. We can at least reduce the bottleneck here.
3. Once we have const generics (in some not so near future) it would be a nice alternative (syntax and logic) to use fixed-size arrays for passing parameters.

kvark:
Ok, I'm fine with going all-in for iterators. Let's rebase this, address my concerns, and proceed.

AIOOB:
Review status: 16 of 25 files reviewed at latest revision, 13 unresolved discussions.

---

*[examples/hal/quad/main.rs, line 164 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CvSGevvHU-WZ-U59N:-L4DXaq2ujWYAijHvNPE:b-nu5wsq) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/examples/hal/quad/main.rs#L164)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

nit: could probably also do `[set_layout]`
</blockquote></details>

Sadly no, set_layout is used later and `[set_layout]` moves it.

---

*[src/hal/src/device.rs, line 203 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4Cz1iHo05hMxS3-qMc:-L4DV3NeeL71jNdtZJOb:b-1u2ok8) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L203)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

strange idea: maybe we can omit this from `gfx-backend-empty` now that we have default implementations? ;)
</blockquote></details>

I'm not sure, not implementing either will probably do something funky at compile time, but `unimplemented!` panics at runtime. If you want I can but I feel like there is no harm leaving one there.

---

*[src/hal/src/device.rs, line 494 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CzXzvvq9I8IdmDdoO:-L4DfJ7ZWswKw_6ssV8m:bhbm9rg) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L494)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

I'm a little concerned that his would prevent the expected behavior when the user passes in timeout of `0`. We should still be checking all the fences, but this implementation may bail out in the process
</blockquote></details>

I think this should work as expected now.

---

*[src/hal/src/device.rs, line 505 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CziooMqguuTf66j-k:-L4DMiUMYCB4cW5WZxXC:b-i1dp8d) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L505)):*
<details><summary><i>Previously, kvark (Dzmitry Malyshau) wrote…</i></summary><blockquote>

why do we need to collect here?
</blockquote></details>

`fences` needs to be iterated over multiple times, the only other option would be to find a way of making the iterator Clone-able.

---

*[src/hal/src/device.rs, line 515 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4C70rck6oWWpyQLZKm-r1-515:-L4DfWAvjQGRFUnhOV_R:b-10ox9r) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L515)):*
<details><summary><i>Previously, msiglreith wrote…</i></summary><blockquote>

I don't think we need issue a thread switch here, drivers will most likely do it internally anyways.
</blockquote></details>

Could this not be a problem for the Mutex simulated fences in Metal, it will now just spin?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1665)*

msiglreith:
Review status: 16 of 25 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/hal/src/device.rs, line 515 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4C70rck6oWWpyQLZKm-r1-515:-L4GTI3D5c6tI4cn_t4H:b-s9725s) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L515)):*
<details><summary><i>Previously, AIOOB (AIOOB) wrote…</i></summary><blockquote>

Could this not be a problem for the Mutex simulated fences in Metal, it will now just spin?
</blockquote></details>

wait_on_fence handling is a bit out of scope atm, but we should rather use native primitives for synchronization (in particular semaphores here) than trying to create our own with locks, loops, etc. IMO

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1665)*

kvark:
Review status: 16 of 25 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/hal/src/device.rs, line 203 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4Cz1iHo05hMxS3-qMc:-L4GyDxbA3zwNlyx3M8C:b-ewhv2) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L203)):*
<details><summary><i>Previously, AIOOB (AIOOB) wrote…</i></summary><blockquote>

I'm not sure, not implementing either will probably do something funky at compile time, but `unimplemented!` panics at runtime. If you want I can but I feel like there is no harm leaving one there.
</blockquote></details>

It's easy to try out, right? Just go to the empty backend, comment out a few methods, and see if it builds. I expect it to compile fine but go to the stack overflow when run. And this would be fine to me, since I don't think there is any value in guaranteeing the empty backend to instantly panic: it's only there to compile-test against.

---

*[src/hal/src/device.rs, line 494 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CzXzvvq9I8IdmDdoO:-L4GzhEd59f2ig0uvxAo:b-wr3sth) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L494)):*
<details><summary><i>Previously, AIOOB (AIOOB) wrote…</i></summary><blockquote>

I think this should work as expected now.
</blockquote></details>

Well, technically, no. The `elapsed()` here can already be non-zero.

---

*[src/hal/src/device.rs, line 505 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4CziooMqguuTf66j-k:-L4GysO-DzRQ21FjXDhU:bp7jgim) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L505)):*
<details><summary><i>Previously, AIOOB (AIOOB) wrote…</i></summary><blockquote>

`fences` needs to be iterated over multiple times, the only other option would be to find a way of making the iterator Clone-able.
</blockquote></details>

Oh I see now.

---

*[src/hal/src/device.rs, line 515 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1665#-L4C70rck6oWWpyQLZKm-r1-515:-L4GzrnwdB2Ny8t2DHD3:b-7rtr2v) ([raw file](https://github.com/gfx-rs/gfx/blob/69fc5364ad68e8d38583d2181f3b5c46af695ea1/src/hal/src/device.rs#L515)):*
<details><summary><i>Previously, msiglreith wrote…</i></summary><blockquote>

wait_on_fence handling is a bit out of scope atm, but we should rather use native primitives for synchronization (in particular semaphores here) than trying to create our own with locks, loops, etc. IMO
</blockquote></details>

In general, either there is a simple working default implementation, or just defer this decision to the backends. In this case, there appear to be enough concerns to consider this a non-trivial implementaiton.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1665)*

AIOOB:
I think that its most things addressed. The only things that I can see that could be added are single variant versions of `update_descriptor_sets`, `bind_graphics_descriptor_sets` and `bind_compute_descriptor_sets`.

kvark:
@AIOOB alright, shipit!
bors r+

kvark:
There appears to be general agreement on the topic. We can nail down the details (descriptor sets) in follow ups, if needed.

PR #1670, opened 2017-12-11
[metal] uniforms

omni-viral:


src/backend/metal/src/device.rs
```diff
@@ -294,6 +294,11 @@ impl Device {
                 arg.set_data_type(MTLDataType::Texture);
                 arg.set_index(index as NSUInteger);
             }
+            DescriptorType::UniformBuffer | DescriptorType::StorageBuffer => {
+                arg.set_access(MTLArgumentAccess::ReadOnly);
```
msiglreith: I don't think storage buffers are read-only.

omni-viral: @msiglreith  Can you explain a bit the difference between `UniformBuffer` and `StorageBuffer`?

msiglreith: `UniformBuffer`: Read-only structured buffer
`StorageBuffer`: Read-Write structured buffer (+atomics)

src/backend/metal/src/device.rs
```diff
@@ -835,7 +848,21 @@ impl hal::Device<Backend> for Device {
                                 *old = Some(((new.0).0.clone(), new.1));
                             }
                         },
-                        (&Sampler(_), _) | (&SampledImage(_), _) => panic!("mismatched descriptor set type"),
+                        (&UniformBuffer(ref buffers), Some(&mut n::DescriptorSetBinding::ConstantBuffer(ref mut vec))) |
+                        (&StorageBuffer(ref buffers), Some(&mut n::DescriptorSetBinding::ConstantBuffer(ref mut vec))) => {
```
msiglreith: This should be a separate DescriptorSetBinding (`StorageBuffer`), which needs to be defined on set allocation afaik.

src/backend/metal/src/device.rs
```diff
@@ -752,11 +762,13 @@ impl hal::Device<Backend> for Device {
 
         let mut num_samplers = 0;
         let mut num_textures = 0;
+        let mut num_uniforms = 0;
```
msiglreith: I don't know what's the point of these, @kvark

kvark: To count resources for allocation in an argument buffer, see code below

src/backend/metal/src/device.rs
```diff
@@ -857,6 +884,26 @@ impl hal::Device<Backend> for Device {
...
+                            mtl_offsets.extend(buffers.iter().map(|buffer| buffer.1.clone()));
+
+                            let encoder: &metal::ArgumentEncoderRef = &encoder;
+
+                            let range = NSRange {
+                                location: offset,
+                                length: mtl_buffers.len() as NSUInteger,
+                            };
+                            unsafe {
+                                msg_send![encoder,
```
kvark: can we call it as `encoder.set_buffers` instead of sending a custom message?

omni-viral: Should I open PR in metal repo or just add function here?

kvark: eh, you are correct. Will leave this for follow up work.

PR #1672, opened 2017-12-12
[metal] Careful vertex buffer inputs remapping

### LINKED ISSUE #1673: [metal] Constant buffer and vertex buffer collision.
In cases where `ArgumentBuffer` is not used gfx-metal-backend tries to bind vertex buffers and constant buffers to the same indices.

kvark:
Fixes  #1673
This PR ensures that we map our vertex attributes after all the resource attributes of the vertex stage, so that they don't collide.
cc @omni-viral @msiglreith


src/backend/metal/src/command.rs
```diff
@@ -441,7 +443,8 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
         }
         if let EncoderState::Render(ref encoder) = inner.encoder_state {
             for (i, &(buffer, offset)) in buffer_set.0.iter().enumerate() {
-                encoder.set_vertex_buffer(i as _, offset as _, Some(&buffer.0));
+                let msl_buffer_index = inner.attribute_buffer_index as usize + i;
```
omni-viral: Probably this offset should be added few lines above too.

Why do we need to store buffers in `resources_vs` anyway?

PR #1675, opened 2017-12-15
Depth metal

omni-viral:
:skull:  :metal:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -522,7 +522,6 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
             command_buffer.encoder_state = EncoderState::None;
 
             // FIXME: subpasses
-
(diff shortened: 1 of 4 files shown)


src/backend/metal/src/command.rs
```diff
@@ -137,6 +138,9 @@ impl CommandBufferInner {
         if let Some(ref pipeline_state) = self.pipeline_state {
             encoder.set_render_pipeline_state(pipeline_state);
         }
+        if let Some(ref depth_stencil_state) = self.depth_stencil_state {
```
kvark: we also need to set the state if the encoder is already active in `bind_graphics_pipeline`

kvark:
@omni-viral ping?

omni-viral:
@kvark got a bit busy. I'll fix this soon.

src/backend/metal/src/command.rs
omni-viral: Done

PR #1676, opened 2017-12-16
quick fix for invalid vulkan api usage (via validation layers)

ForestCSharp:
The hal/quad demo had two vertex shader push constant ranges that the vulkan validation layers were complaining about. I combined them into one range.

ERROR:gfx_backend_vulkan: [DS] vkCreatePipelineLayout() Duplicate stage flags found in ranges 0 and 1. The spec valid usage text states 'Any two elements of pPushConstantRanges must not include the same stage in stageFlags'


kvark:
I wonder if it still works on Metal/D3D12 with this change. Our testing infra still needs more love.

PR #1677, opened 2017-12-17
allocating only as much memory as required for Buffers

ForestCSharp:
slightly altered the image staging buffer code to do the same (needed to defer the buffer bind so i could get the requirements).

At least for the Image Buffer, this suppresses another Vulkan validation layer error, as i guess these buffers technically needed a little more memory than their input data size?

Also changed the second image barrier to transition to the fragment stage instead of END_OF_PIPE

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -479,7 +479,7 @@ fn main() {
                 target: &image_logo,
                 range: COLOR_RANGE.clone(),
             };
-            cmd_buffer.pipeline_barrier(PipelineStage::TRANSFER .. PipelineStage::BOTTOM_OF_PIPE, &[image_barrier]);
+            cmd_buffer.pipeline_barrier(PipelineStage::TRANSFER .. PipelineStage::FRAGMENT_SHADER, &[image_barrier]);
 
             cmd_buffer.finish()
         };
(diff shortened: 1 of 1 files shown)


examples/hal/quad/main.rs
```diff
@@ -375,11 +375,10 @@ fn main() {
...
     println!("upload row pitch {}, total size {}", row_pitch, upload_size);
 
-    let image_upload_memory = device.allocate_memory(upload_type, upload_size).unwrap();
-    let image_upload_buffer = {
-        let buffer = device.create_buffer(upload_size, image_stride as u64, buffer::Usage::TRANSFER_SRC).unwrap();
-        device.bind_buffer_memory(&image_upload_memory, 0, buffer).unwrap()
-    };
+    let image_buffer_unbound = device.create_buffer(upload_size, image_stride as u64, buffer::Usage::TRANSFER_SRC).unwrap();
+    let image_mem_reqs = device.get_buffer_requirements(&image_buffer_unbound);
+    let image_upload_memory = device.allocate_memory(upload_type, image_mem_reqs.size).unwrap();
```
msiglreith: Just noticing that `upload_type` might not be supported by `image_buffer_unbound` as they were created with different usage flags, even though it probably works on 99% of the hardware out there.

kvark: Indeed... but this is only slightly touching on this PR

PR #1680, opened 2017-12-18
Remove stride parameter on buffer creation

msiglreith:
Stride was only needed for dx12 to use it for vertex buffer binding. The PSO stride and the stride on buffer creation needed to be equal, causing issues on the user side. Caching the vertex buffers and use the strides from the active PSO fixes this. Also, helps for better vulkan portability.

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -2,7 +2,7 @@ use wio::com::ComPtr;
 use hal::{command as com, image, memory, pass, pso, query};
 use hal::{IndexCount, IndexType, InstanceCount, VertexCount, VertexOffset};
 use hal::buffer::IndexBufferView;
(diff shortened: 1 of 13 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -360,6 +374,23 @@ impl CommandBuffer {
         }
 
         let cmd_buffer = &mut self.raw;
+
+        // Bind vertex buffers
+        // We currently don't support offsets for vertex buffer binding, therefore,
```
kvark: why not?

msiglreith: Just not implemented so far, first binding is always 0 atm.
https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkan/src/command.rs#L440

PR #1681, opened 2017-12-19
[pre-ll] Emscripten support with WebGL 2.0

### LINKED ISSUE #641: Emscripten compatibility
The way is clear now: http://bl.ocks.org/AerialX/26f81381b0b47c506959

kvark:
Fixes #641. It has more things, so we'll create extra issues as we go forward.

The support is limited, but it works. Unexpected limitations: `constant_buffer_supported` and `clear_buffer_supported` have to be OFF. Emscripten doesn't hook those GL functions, even though they are technically a part of GLES 3.0 standard. Also, the `glutin` dependency is switched to GitHub until the next version is published.

gfx-wasm-triangle

## Status of the examples
  - `triangle` works 🎉 
  - `flowmap` and `blend` produce ICE in the [fastcomp](https://github.com/kripken/emscripten-fastcomp) code
  - `instancing` fails because it wants to map a buffer
  - `mipmap`, `gamma`, and `performance` I haven't tried
  - others use constant buffers, which are not yet supported (see above)

## How to run and test yourself

You'll need Rust-1.22 or newer as well as Emscripten SDK installed. In Emscripten configuration, set `USE_WEBGL2=1`. I did it by just modifying "emscripten/src/settings.js" directly (from where it's installed). Hopefully, we'll be able to pass the linker parameters to `emcc` via `cargo` one day.

```bash
rustup target add wasm32-unknown-emscripten
cargo build --example triangle --feature wasm32-unknown-emscripten
ls target/wasm32-unknown-emscripten/debug/examples # find the "example-XXX" name
edit emscripten.html # put the example name in
firefox emscripten.html # open in the browser
```


emscripten.html
```diff
@@ -0,0 +1,31 @@
...
+        script.src = base + name + ".js";
+        script.onload = function() {
+          console.log("Emscripten boilerplate loaded.")
+        }
+        document.body.appendChild(script);
+      });
+    </script>
+    <canvas id="canvas" style="width:100%; height:100%"></canvas>
+  </body>
+</html>
```
msiglreith: nit: newline

PR #1687, opened 2017-12-20
Use memory type index for memory allocations

msiglreith:
Use the index of the memory type in the memory properties array instead of passing the actual memory type struct on memory allocation.

Pros:
* More robust compared to passing actual struct (can't pass bogus data)
* Easier for vulkan portability
* Remove redundant `id` field in `MemoryType`

Cons:
* User could cause out-of-boundary exceptions at runtime

On the API side it's probably the same for the user, I would say even a bit nicer with indices.


examples/hal/compute/main.rs
```diff
@@ -183,12 +183,16 @@ fn create_buffer<B: Backend>(
     let buffer = gpu.device.create_buffer(stride * len, usage).unwrap();
     let requirements = gpu.device.get_buffer_requirements(&buffer);
 
-    let ty = memory_types.into_iter().find(|memory_type| {
-        requirements.type_mask & (1 << memory_type.id) != 0 &&
-        memory_type.properties.contains(properties)
-    }).unwrap();
-
-    let memory = gpu.device.allocate_memory(&ty, requirements.size).unwrap();
+    let ty = memory_types
```
kvark: we could at least wrap that index in a newtype thing for cleaner code

src/backend/dx12/src/device.rs
```diff
@@ -10,20 +10,24 @@ use winapi::shared::{dxgi, dxgi1_2, dxgi1_4, dxgiformat, dxgitype, winerror};
...
-    Backend as B, Device, QueueFamily, MAX_VERTEX_BUFFERS,
+    Backend as B, Device, QueueFamily, MAX_VERTEX_BUFFERS, NUM_HEAP_PROPERTIES,
 };
 use pool::RawCommandPool;
 use root_constants::RootConstant;
 
 // Register space used for root constants.
 const ROOT_CONSTANT_SPACE: u32 = 0;
 
+const MEM_TYPE_MASK: u64 = 0x7;
```
kvark: nice!

src/backend/dx12/src/lib.rs
```diff
@@ -139,11 +141,10 @@ impl QueueFamily {
 
 pub struct PhysicalDevice {
     adapter: ComPtr<dxgi1_2::IDXGIAdapter2>,
-    factory: ComPtr<dxgi1_4::IDXGIFactory4>,
```
kvark: great :)

src/backend/dx12/src/lib.rs
```diff
@@ -650,19 +642,14 @@ impl hal::Instance for Instance {
             let memory_types = if heterogeneous_resource_heaps {
                 base_memory_types.to_vec()
             } else {
-                // the bit pattern of ID becomes 0bTTII, where
-                // TT=1 for buffers, TT=2 for images, and TT=3 for targets
-                // TT=0 is reserved for future use, helps to avoid ambiguity
-                // and II is the same `id` as the `base_memory_types` have
+                // We multiplicate the base memory types
+                //     0.. 3: Reserved for futures use
```
kvark: mention the constants here?

examples/hal/compute/main.rs
msiglreith: I tend to agree but at which API level (examples, render, hal) should this be added?

kvark: HAL is fine here. It doesn't add a ton of mental overhead, protects from ambiguation, and is still easy to escape from if the newtype contents are public.

msiglreith: So, something along
```rust
// hal
pub struct MemTypeId(pub usize);
impl Device {
    fn allocate_memory(&self, id: MemTypeId, size: u64);
}

// example
let ty = MemTypeId(
     mem_types.iter().position(|...|).unwrap()
);
```
?

kvark: yes, this will do.
The ergonomics of finding such a type ID is a little wordy, so we'll have a simple helper anyway

msiglreith:
Addressed all points
bors r=kvark

PR #1693, opened 2017-12-26
[gl] Switch `transmute` to `from_raw_parts`

grovesNL:
It seems like `transmute` does not perform the expected slice conversion here. Switching to `from_raw_parts` seems to work correctly, but I'm not sure what the performance trade-off will be (in the long-term).

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -7,7 +7,7 @@ use {native as n, Backend};
 use pool::{self, BufferMemory};
 
 use std::borrow::Borrow;
-use std::mem;
+use std::{mem, slice};
 use std::ops::Range;
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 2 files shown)


src/backend/gl/src/command.rs
```diff
@@ -226,8 +226,14 @@ impl RawCommandBuffer {
 
     /// Copy a given vector slice into the data buffer.
     fn add<T>(&mut self, data: &[T]) -> BufferSlice {
-        self.add_raw(unsafe { mem::transmute(data) })
+        self.add_raw(unsafe {
+            slice::from_raw_parts(
+                data.as_ptr() as *const _,
+                data.len() * mem::size_of::<T>() / mem::size_of::<u8>(),
```
msiglreith: Let's remove the `mem::size_of::<u8>()`. It's 1 by definition anyways

src/backend/gl/src/queue.rs
```diff
@@ -173,10 +173,15 @@ impl CommandQueue {
 
     /// Return a reference to a stored data object.
     fn get<T>(data: &[u8], ptr: com::BufferSlice) -> &[T] {
-        assert_eq!(ptr.size % mem::size_of::<T>() as u32, 0);
```
msiglreith: Does the following code cover this assertion? If not, could we keep it?

src/backend/gl/src/queue.rs
```diff
@@ -173,10 +173,15 @@ impl CommandQueue {
...
-        let raw_data = Self::get_raw(data, ptr);
-        unsafe { mem::transmute(raw_data) }
+        let raw = Self::get_raw(data, ptr);
+        unsafe {
+            slice::from_raw_parts(
+                raw.as_ptr() as *const _,
+                (raw.len() * mem::size_of::<u8>()) / mem::size_of::<T>(),
```
msiglreith: (same as above regarding `size_of::<u8>()`)

grovesNL:
Thanks! Updated to address comments.

PR #1694, opened 2017-12-26
[gl] Use primitive from graphics pipeline

grovesNL:
Use the primitive from the graphics pipeline for `Draw` and `DrawIndexed` commands.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -79,6 +79,7 @@ pub enum Command {
     BindFrameBuffer(FrameBufferTarget, n::FrameBuffer),
     BindTargetView(FrameBufferTarget, AttachmentPoint, n::ImageView),
     SetDrawColorBuffers(usize),
+    SetPatchSize(gl::types::GLint),
 }
 
 pub type FrameBufferTarget = gl::types::GLenum;
@@ -100,6 +101,8 @@ struct Cache {
     ///
     // Indicates that invalid commands have been recorded.
     error_state: bool,
+    // Vertices per patch for tessellation primitives (patches).
+    patch_size: Option<gl::types::GLint>,
 }
 
 impl Cache {
@@ -111,6 +114,7 @@ impl Cache {
             blend_color: None,
             framebuffer: None,
             error_state: false,
+            patch_size: None,
         }
     }
 }
@@ -478,8 +482,23 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         }
     }
 
-    fn bind_graphics_pipeline(&mut self, _pipeline: &n::GraphicsPipeline) {
-        unimplemented!()
+    fn bind_graphics_pipeline(&mut self, pipeline: &n::GraphicsPipeline) {
+        let &n::GraphicsPipeline {
+            primitive,
+            patch_size,
+            ..
+        } = pipeline;
+
+        if self.cache.primitive != Some(primitive) {
+            self.cache.primitive = Some(primitive);
+        }
+
+        if self.cache.patch_size != patch_size {
+            self.cache.patch_size = patch_size;
+            if let Some(size) = patch_size {
+                self.push_cmd(Command::SetPatchSize(size));
+            }
+        }
     }
 
     fn bind_graphics_descriptor_sets<'a, T>(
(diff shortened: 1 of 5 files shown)


msiglreith:
The change is motivated by carrying over the tessellation control point information over to the draw commands, right? I would prefer to set `gl.PatchParameteri` on pipeline change, if possible, as draw calls occur at a higher frequency than pipeline changes imho. Also, we should do the HAL -> GL conversion once at pipeline creation.

grovesNL:
Agreed, I wasn't sure where to put the tessellation control point information though. Should I add it to `GraphicsPipeline` as `Option<gl::types::GLint>`?

grovesNL:
I updated the commit to perform the conversion during pipeline creation, and store `patch_size` on `GraphicsPipeline` (updated during bind - didn't bother to cache it yet).

src/backend/gl/src/command.rs
```diff
@@ -478,8 +479,20 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
+            primitive,
+            patch_size,
+            ..
+        } = pipeline;
+
+        if self.cache.primitive != Some(primitive) {
+            self.cache.primitive = Some(primitive);
+        }
+
+        if let Some(size) = patch_size {
```
msiglreith: I think we can also cache this in the future, but not sure if worth it..

src/backend/gl/src/conv.rs
```diff
@@ -44,3 +44,18 @@ pub fn buffer_usage_to_gl_target(usage: buffer::Usage) -> Option<t::GLenum> {
...
+        Primitive::PointList => gl::POINTS,
+        Primitive::LineList => gl::LINES,
+        Primitive::LineStrip => gl::LINE_STRIP,
+        Primitive::TriangleList => gl::TRIANGLES,
+        Primitive::TriangleStrip => gl::TRIANGLE_STRIP,
+        Primitive::LineListAdjacency => gl::LINES_ADJACENCY,
+        Primitive::LineStripAdjacency => gl::LINE_STRIP_ADJACENCY,
+        Primitive::TriangleListAdjacency => gl::TRIANGLES_ADJACENCY,
+        Primitive::TriangleStripAdjacency => gl::TRIANGLE_STRIP_ADJACENCY,
+        Primitive::PatchList(_) => gl::PATCHES
```
msiglreith: nit: trailing comma

src/backend/gl/src/device.rs
```diff
@@ -347,8 +347,16 @@ impl d::Device<B> for Device {
                     name
                 };
 
+                let patch_size = if let Primitive::PatchList(size) = desc.input_assembler.primitive {
```
msiglreith: Let's do a `match` here, looks a bit better/more coherent(?)

src/backend/gl/src/queue.rs
```diff
@@ -463,6 +463,9 @@ impl CommandQueue {
             com::Command::SetDrawColorBuffers(num) => {
                 state::bind_draw_color_buffers(&self.share.context, num);
             }
+            com::Command::SetPatchSize(num) => unsafe {
+                &self.share.context.PatchParameteri(gl::PATCH_VERTICES, num);
```
msiglreith: neat :D

src/backend/gl/src/device.rs
```diff
@@ -347,8 +347,16 @@ impl d::Device<B> for Device {
...
+                    Some(size as _)
+                } else {
+                    None
+                };
+
                 Ok(n::GraphicsPipeline {
                     program,
+                    primitive: conv::primitive_to_gl_primitive(desc.input_assembler.primitive),
+                    patch_size
```
msiglreith: nit: trailing comma

grovesNL:
Alright, added to cache and fixed match/commas (it would be nice to use `rustfmt` at some point because I always miss these commas 😃).

src/backend/gl/src/command.rs
```diff
@@ -478,8 +482,22 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
+            primitive,
+            patch_size,
+            ..
+        } = pipeline;
+
+        if self.cache.primitive != Some(primitive) {
+            self.cache.primitive = Some(primitive);
+        }
+
+        if self.cache.patch_size != patch_size {
```
msiglreith: `self.cache.patch_size` never gets updated

grovesNL: Whoops :sweat_smile: Fixed

PR #1697, opened 2017-12-27
[gl] Set blend per attachment

grovesNL:
Sets blend per attachments. I haven't added caching yet because I wanted to verify whether this is the correct approach.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -1,7 +1,7 @@
 #![allow(missing_docs)]
 
 use gl;
-use hal::{self, command, image, memory, pso, query};
+use hal::{self, command, image, memory, pso, query, ColorSlot};
 use hal::buffer::IndexBufferView;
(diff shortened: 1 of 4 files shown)


grovesNL:
If blend isn't specified for an attachment, should we keep the previously bound blend state or revert to a default state (like `ColorBlendDesc::EMPTY` or the OpenGL default)? Or should we require that blend is specified for all attachments?

kvark:
@grovesNL IIRC, it has to be specified for all attachments

grovesNL:
Sounds good. We can add an assertion that blend is specified for all attachments later (if it doesn't already exist somewhere).

I updated my commit to add a cache (untested until the GL backend is slightly more usable) although it seems a bit ugly because of the nested `Option`. I guess we could share some of this logic when we cache more attachment state.

kvark:
thanks @grovesNL !
bors r=msiglreith

PR #1700, opened 2017-12-28
Refactor formats and tweak buffer-image copies

msiglreith:
Refactor the format definitions to support all core formats from Vulkan. We still support the old (SurfaceType, ChannelType) format, but it's a superset of the new `Format` enum.

We avoid that users uses formats which are not supported at all, additionally, we can reduce the conversion costs for VkPortability <-> HAL.

* Support compressed formats and adjust buffer-image copy accordingly
* Align `BufferImageCopy` fields with the Vulkan specification


kvark:
> We still support the old (SurfaceType, ChannelType) format, but it's a superset of the new Format enum.

Do you think it's still going to be useful? It doesn't strike me as obvious that we'd want to have two different ways of providing a format, even if one is a superset of the other.

msiglreith:
> Do you think it's still going to be useful? It doesn't strike me as obvious that we'd want to have two different ways of providing a format, even if one is a superset of the other.

The user can't specify or pass formats using the old (SurfaceType, ChannelType) format. It useful for getting information about the format like num bits of the surface types, dimension of blocks for compressed formats, supported aspects etc.

kvark:
I see now, makes good sense!

msiglreith:
Should be ready (+/- metal build errors)

src/backend/vulkan/src/conv.rs
```diff
@@ -4,347 +4,22 @@ use hal::{buffer, command, format, image, pass, pso, query};
...
-            _ => return None,
-        },
-        D32_S8 => match chan {
-            Float => vk::Format::D32SfloatS8Uint,
-            _ => return None,
-        },
-    })
+pub fn map_format(format: format::Format) -> vk::Format {
+    // Safe due to equivalence of HAL format values and Vulkan format values
+    unsafe { mem::transmute(format) }
```
kvark: neat! we should do this for some bit flags as well

src/hal/src/format.rs
```diff
@@ -341,237 +146,426 @@ bitflags!(
...
-            Vec4<$ty> = $channel R16_G16_B16_A16,
-        )*}
+impl Format {
+    /// Get base format.
+    ///
+    /// Returns `None` if format is `Undefined`.
+    pub fn base_format(self) -> Option<BaseFormat> {
+        assert!(NUM_FORMATS > self as usize);
+        match self {
+            Format::Undefined => None,
```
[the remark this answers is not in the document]
msiglreith: I'll look into it!

src/backend/dx12/src/conv.rs
```diff
@@ -16,7 +16,6 @@ pub fn map_format(format: Format) -> Option<DXGI_FORMAT> {
 
     // TODO: list may be incomplete or partially wrong..
     let format = match format {
-        Undefined => DXGI_FORMAT_UNKNOWN,
```
kvark: 🎉 🥇 👍

src/hal/src/adapter.rs
```diff
@@ -60,7 +60,7 @@ pub trait PhysicalDevice<B: Backend>: Sized {
     fn open(self, Vec<(B::QueueFamily, Vec<QueuePriority>)>) -> Gpu<B>;
 
     ///
-    fn format_properties(&self, format::Format) -> format::Properties;
+    fn format_properties(&self, Option<format::Format>) -> format::Properties;
```
kvark: does this function even make sense to call with `None`? I'd like to have the format passed here straight

src/hal/src/device.rs
```diff
@@ -247,7 +247,7 @@ pub trait Device<B: Backend> {
 
     ///
     fn create_buffer_view(
-        &self, &B::Buffer, format::Format, Range<u64>
+        &self, &B::Buffer, Option<format::Format>, Range<u64>
```
kvark: How does this work with `None`?

src/hal/src/pass.rs
```diff
@@ -66,7 +66,11 @@ impl AttachmentOps {
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Attachment {
     /// Attachment format
-    pub format: Format,
+    ///
+    /// In the most cases `format` is not `None`. It should be only used for
```
kvark: 👍 for comments!

src/hal/src/adapter.rs
msiglreith: Yes, uniform/storage texel buffers with `Format::Undefined` are allowed. (also called untyped buffers).
It's also handled inside the open AMD drivers.

src/hal/src/device.rs
msiglreith: Yes, uniform/storage texel buffers with Format::Undefined are allowed. (also called untyped buffers).
It's also handled inside the open AMD drivers.

src/hal/src/adapter.rs
msiglreith: I guess that it's required due to `create_buffer_view`. We need check the supported buffer features for `VK_FORMAT_UNDEFINED`.

msiglreith:
Fixed last metal build errors. Thanks the review again (:

kvark:
Alright, thanks! :ship: it
bors r+

PR #1698, opened 2017-12-27
travis: Disable incremental builds due to storage shortage

msiglreith:

PR #1701, opened 2017-12-28
[gl] Make hal/quad example work on GL backend

grovesNL:
This adds basic functionality to run the hal/quad example on the GL backend. Windows seems to work fine. OS X fails because the varying name doesn't match across vertex/fragment shaders, although I think this is supposed to work on 4.10.

~~WIP because there are a few format-related attribute arguments hardcoded (element type/size/stride) at the moment, but I'll probably wait for #1700 to fix them.~~

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -82,7 +82,7 @@ pub enum Command {
     SetPatchSize(gl::types::GLint),
     BindProgram(gl::types::GLuint),
     BindBlendSlot(ColorSlot, pso::ColorBlendDesc),
-    BindAttribute(n::AttributeDesc, gl::types::GLuint),
+    BindAttribute(n::AttributeDesc, gl::types::GLuint, gl::types::GLsizei),
     UnbindAttribute(n::AttributeDesc),
     CopyBufferToTexture(n::RawBuffer, gl::types::GLuint, command::BufferImageCopy),
 }
@@ -113,7 +113,9 @@ struct Cache {
     // Blend per attachment.
     blend_targets: Option<Vec<Option<pso::ColorBlendDesc>>>,
     // Maps bound vertex buffer offset (index) to handle.
-    vertex_buffers: Option<Vec<u32>>,
+    vertex_buffers: Option<Vec<gl::types::GLuint>>,
+    // Active vertex buffer descriptions.
+    vertex_buffer_descs: Option<Vec<pso::VertexBufferDesc>>,
     // Active attributes.
     attributes: Option<Vec<n::AttributeDesc>>,
 }
@@ -131,6 +133,7 @@ impl Cache {
             program: None,
             blend_targets: None,
             vertex_buffers: None,
+            vertex_buffer_descs: None,
             attributes: None,
         }
     }
@@ -387,19 +390,26 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
 
         if let Some(ref attributes) = self.cache.attributes {
             for attribute in attributes {
-                let handle = match self.cache.vertex_buffers {
-                    Some(ref vbs) => vbs[attribute.binding as usize],
-                    None => {
-                        error!("No vertex buffers are bound");
-                        0
-                    },
-                };
-                push_cmd_internal(
-                    &self.id,
-                    &mut self.memory,
-                    &mut self.buf,
-                    Command::BindAttribute(*attribute, handle)
-                );
+                if let Some(ref vbs) = self.cache.vertex_buffers {
+                    let binding = attribute.binding as usize;
+                    if let Some(ref descs) = self.cache.vertex_buffer_descs {
+                        let desc = &descs[binding];
+                        let handle = vbs[binding];
+
+                        assert_eq!(desc.rate, 0); // TODO: Input rate
+
+                        push_cmd_internal(
+                            &self.id,
+                            &mut self.memory,
+                            &mut self.buf,
+                            Command::BindAttribute(*attribute, handle, desc.stride as _)
+                        ); 
+                    } else {
+                        error!("No vertex buffer descriptions are bound");
+                    }
+                } else {
+                    error!("No vertex buffers are bound");
+                }
             }
         };
     }
@@ -488,7 +498,8 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         
         if let Some(ref mut cached) = self.cache.vertex_buffers {
             for vb in vbs.0 {
-                cached[vb.1] = vb.0.raw;
+                let buffer = vb.0;
+                cached[vb.1] = buffer.raw;
             }
         }
     }
@@ -581,6 +592,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             program,
             ref blend_targets,
             ref attributes,
+            ref vertex_buffers,
         } = pipeline;
 
         if self.cache.primitive != Some(primitive) {
@@ -600,6 +612,8 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         }
 
         self.cache.attributes = Some(attributes.clone());
+
+        self.cache.vertex_buffer_descs = Some(vertex_buffers.clone());
 
         self.update_blend_targets(blend_targets);
     }
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -82,6 +82,8 @@ pub enum Command {
     SetPatchSize(gl::types::GLint),
     BindProgram(gl::types::GLuint),
     BindBlendSlot(ColorSlot, pso::ColorBlendDesc),
+    BindAttribute(n::AttributeDesc),
+    UnbindAttribute(n::AttributeDesc),
 }
 
 pub type FrameBufferTarget = gl::types::GLenum;
@@ -556,7 +558,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             patch_size,
             program,
             ref blend_targets,
-            ..
+            ref attributes,
         } = pipeline;
 
         if self.cache.primitive != Some(primitive) {
@@ -573,6 +575,10 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         if self.cache.program != Some(program) {
             self.cache.program = Some(program);
             self.push_cmd(Command::BindProgram(program));
+        }
+
+        for attribute in attributes {
+            self.push_cmd(Command::BindAttribute(*attribute));
         }
 
         self.update_blend_targets(blend_targets);
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -224,25 +224,7 @@ impl RawCommandBuffer {
     }
 
     fn push_cmd(&mut self, cmd: Command) {
-        let slice = {
-            let mut memory = self
-                .memory
-                .try_lock()
-                .expect("Trying to record a command buffers, while memory is in-use.");
-
-            let cmd_buffer = match *memory {
-                BufferMemory::Linear(ref mut buffer) => &mut buffer.commands,
-                BufferMemory::Individual { ref mut storage, .. } => {
-                    &mut storage.get_mut(&self.id).unwrap().commands
-                }
-            };
-            cmd_buffer.push(cmd);
-            BufferSlice {
-                offset: cmd_buffer.len() as u32 - 1,
-                size: 1,
-            }
-        };
-        self.buf.append(slice);
+        push_cmd_internal(&self.id, &mut self.memory, &mut self.buf, cmd);
     }
 
     /// Copy a given vector slice into the data buffer.
@@ -403,33 +385,23 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             }
         }
 
-        let mut attribute_commands = None;
-
         if let Some(ref attributes) = self.cache.attributes {
-            attribute_commands = Some(attributes
-                .iter()
-                .map(|attribute| {
-                        let handle = match self.cache.vertex_buffers {
-                            Some(ref vbs) => vbs[attribute.binding as usize],
-                            None => {
-                                error!("No vertex buffers are bound");
-                                0
-                            },
-                        };
-                        Command::BindAttribute(*attribute, handle)
-                    })
-                .collect::<Vec<_>>()
-            );
-        };
-
-        match attribute_commands {
-            Some(cmds) => {
-                for cmd in cmds {
-                    self.push_cmd(cmd);
-                }
+            for attribute in attributes {
+                let handle = match self.cache.vertex_buffers {
+                    Some(ref vbs) => vbs[attribute.binding as usize],
+                    None => {
+                        error!("No vertex buffers are bound");
+                        0
+                    },
+                };
+                push_cmd_internal(
+                    &self.id,
+                    &mut self.memory,
+                    &mut self.buf,
+                    Command::BindAttribute(*attribute, handle)
+                );
             }
-            None => {}
-        }
+        };
     }
 
     fn next_subpass(&mut self, _contents: command::SubpassContents) {
@@ -853,6 +825,27 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     ) {
         unimplemented!()
     }
+}
+
+fn push_cmd_internal(id: &u64, memory: &mut Arc<Mutex<pool::BufferMemory>>, buffer: &mut BufferSlice, cmd: Command) {
+    let slice = {
+        let mut memory = memory
+            .try_lock()
+            .expect("Trying to record a command buffers, while memory is in-use.");
+
+        let cmd_buffer = match *memory {
+            BufferMemory::Linear(ref mut buffer) => &mut buffer.commands,
+            BufferMemory::Individual { ref mut storage, .. } => {
+                &mut storage.get_mut(id).unwrap().commands
+            }
+        };
+        cmd_buffer.push(cmd);
+        BufferSlice {
+            offset: cmd_buffer.len() as u32 - 1,
+            size: 1,
+        }
+    };
+    buffer.append(slice);
 }
 
 /// A subpass command buffer abstraction for OpenGL
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -361,7 +361,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         T: IntoIterator,
         T::Item: Borrow<memory::Barrier<'a, Backend>>,
     {
-        unimplemented!()
+        // TODO
     }
 
     fn fill_buffer(&mut self, _buffer: &n::Buffer, _range: Range<u64>, _data: u32) {
@@ -400,7 +400,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     }
 
     fn end_renderpass(&mut self) {
-        unimplemented!()
+        // TODO
     }
 
     fn clear_color_image(
@@ -587,7 +587,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         T: IntoIterator,
         T::Item: Borrow<n::DescriptorSet>,
     {
-        unimplemented!()
+        // TODO
     }
 
     fn bind_compute_pipeline(&mut self, _pipeline: &n::ComputePipeline) {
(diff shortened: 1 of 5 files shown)


src/backend/gl/src/device.rs
```diff
@@ -357,6 +362,16 @@ impl d::Device<B> for Device {
                     primitive: conv::primitive_to_gl_primitive(desc.input_assembler.primitive),
                     patch_size,
                     blend_targets: desc.blender.targets.clone(),
+                    attributes: desc.attributes
+                        .iter()
+                        .map(|&a| n::AttributeDesc {
+                            location: a.location,
+                            offset: a.element.offset,
+                            binding: a.binding + 1,
+                            size: 2, // TODO
```
kvark: heheh

grovesNL:
Yeah there are still lots of pieces missing at the moment 😆 Even the sampler just happens to work in this case. I'll try to clean it up a bit more and add some assertions before it's ready to merge.

src/backend/gl/src/command.rs
```diff
@@ -359,7 +364,6 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         T: IntoIterator,
         T::Item: Borrow<memory::Barrier<'a, Backend>>,
     {
-        unimplemented!()
```
msiglreith: Could you add `// TODO` at the places where `unimplemented!()` is removed?

src/backend/gl/src/command.rs
```diff
@@ -454,8 +466,10 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         self.push_cmd(Command::BindIndexBuffer(ibv.buffer.raw));
     }
 
-    fn bind_vertex_buffers(&mut self, _vbs: hal::pso::VertexBufferSet<Backend>) {
-        unimplemented!()
+    fn bind_vertex_buffers(&mut self, vbs: hal::pso::VertexBufferSet<Backend>) {
+        for vertex_buffer in vbs.0 {
+            self.push_cmd(Command::BindVertexBuffer(vertex_buffer.0.raw));
```
msiglreith: This would work if `BindVertexBuffer` would actually be translated to a `glBindVertexBuffer` call, but atm it only overrides the buffer bound at `ARRAY_BUFFER`.

src/backend/gl/src/device.rs
```diff
@@ -357,6 +362,16 @@ impl d::Device<B> for Device {
                     primitive: conv::primitive_to_gl_primitive(desc.input_assembler.primitive),
                     patch_size,
                     blend_targets: desc.blender.targets.clone(),
+                    attributes: desc.attributes
+                        .iter()
+                        .map(|&a| n::AttributeDesc {
+                            location: a.location,
+                            offset: a.element.offset,
+                            binding: a.binding + 1,
```
msiglreith: Where does this `+1` come from?

src/backend/gl/src/device.rs
```diff
@@ -205,7 +208,9 @@ impl d::Device<B> for Device {
     fn allocate_memory(
         &self, _mem_type: c::MemoryTypeId, _size: u64,
     ) -> Result<n::Memory, d::OutOfMemory> {
-        unimplemented!()
+        Ok(n::Memory {
+            properties: memory::Properties::CPU_VISIBLE | memory::Properties::CPU_CACHED
```
msiglreith: Is this a `TODO` ?

src/backend/gl/src/device.rs
```diff
@@ -580,18 +595,49 @@ impl d::Device<B> for Device {
...
 
-    fn create_image(&self, _: i::Kind, _: i::Level, _: Format, _: i::Usage)
+    fn create_image(&self, kind: i::Kind, _: i::Level, _: Format, _: i::Usage)
          -> Result<UnboundImage, i::CreationError>
     {
-        unimplemented!()
+        let gl = &self.share.context;
+
+        let raw = unsafe {
+            let mut raw = mem::uninitialized();
```
msiglreith: Let's use `0` instead of `mem::unitialized()`.

src/backend/gl/src/device.rs
```diff
@@ -580,18 +595,49 @@ impl d::Device<B> for Device {
...
+            gl.GenTextures(1, &mut raw);
+            raw
+        };
+
+        let width;
+        let height;
+
+        match kind {
```
msiglreith: `let (width, height) = match { .. }` looks nicer.

src/backend/gl/src/device.rs
```diff
@@ -580,18 +595,49 @@ impl d::Device<B> for Device {
...
+            i::Kind::D2(w, h, _aa) => unsafe {
+                width = w;
+                height = h;
+                gl.BindTexture(gl::TEXTURE_2D, raw);
+                gl.TexStorage2D(gl::TEXTURE_2D, 1, gl::SRGB8_ALPHA8, w as _, h as _);
+                gl.BindTexture(gl::TEXTURE_2D, 0);
```
msiglreith: I think we can omit the unbind.

src/backend/gl/src/queue.rs
```diff
@@ -256,7 +256,9 @@ impl CommandQueue {
                 self.state.index_buffer = Some(buffer);
                 unsafe { gl.BindBuffer(gl::ELEMENT_ARRAY_BUFFER, buffer) };
             }
-//          com::Command::BindVertexBuffers(_data_ptr) =>
+            com::Command::BindVertexBuffer(buffer) => unsafe {
```
msiglreith: I think we should roughly model the internal commands according the GL API:
```rust
BindBuffer => glBindBuffer
BindVertexBuffer => glBindVertexBuffer (4.3+, ES 3.1+)
BindVertexBuffers => glBindVertexBuffers (4.4+)
```
I guess it's easier to reason about the internals and moves computation from submit to recording (checking capabilities).

src/backend/gl/src/command.rs
grovesNL: True, maybe it's better to simply no-op this call for now until we add `glBindVertexBuffer`. At the moment it will be handled in `BindAttribute` anyway

src/backend/gl/src/device.rs
grovesNL: We should at least return flags which correspond to `mem_type` (from `PhysicalDevice::memory_properties`). Although I'm not sure how we plan to handle memory allocations in the GL backend in the long term for persistent mapping.

src/backend/gl/src/device.rs
grovesNL: Just a temporary shortcut, it needs to receive the buffer name instead

kvark:
@grovesNL what's currently left to do here?

grovesNL:
@kvark Just planning to clean up some hardcoded values using the updated formats from #1700

grovesNL:
@msiglreith @kvark Please take a look at the updated commits. I think this PR is ready now if we're fine with the remaining shortcuts (mostly handling textures and uniforms correctly).

src/backend/gl/src/command.rs
```diff
@@ -109,6 +112,12 @@ struct Cache {
     program: Option<gl::types::GLuint>,
     // Blend per attachment.
     blend_targets: Option<Vec<Option<pso::ColorBlendDesc>>>,
+    // Maps bound vertex buffer offset (index) to handle.
+    vertex_buffers: Option<Vec<gl::types::GLuint>>,
```
msiglreith: Is there a difference between an empty `Vec` and `None`?

src/backend/gl/src/command.rs
```diff
@@ -215,25 +227,7 @@ impl RawCommandBuffer {
...
-                }
-            };
-            cmd_buffer.push(cmd);
-            BufferSlice {
-                offset: cmd_buffer.len() as u32 - 1,
-                size: 1,
-            }
-        };
-        self.buf.append(slice);
+        push_cmd_internal(&self.id, &mut self.memory, &mut self.buf, cmd);
```
msiglreith: What's the purpose of this internal function?

src/backend/gl/src/command.rs
```diff
@@ -393,14 +387,39 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
                 }
             }
         }
+
+        if let Some(ref attributes) = self.cache.attributes {
```
msiglreith: vertex buffer binding can be done inside and outside a renderpass. The correct place would be before `draw`/`draw_indexed` in my opinion. See the dx12 backend for similar flushing concerning descriptor tables.

src/backend/gl/src/command.rs
```diff
@@ -465,8 +484,24 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         self.push_cmd(Command::BindIndexBuffer(ibv.buffer.raw));
     }
 
-    fn bind_vertex_buffers(&mut self, _vbs: hal::pso::VertexBufferSet<Backend>) {
-        unimplemented!()
+    fn bind_vertex_buffers(&mut self, vbs: hal::pso::VertexBufferSet<Backend>) {
+        if vbs.0.len() == 0 {
```
msiglreith: Let's assert here. Zero vertex buffers should be/is invalid.

src/backend/gl/src/command.rs
```diff
@@ -636,17 +676,28 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
+         T::Item: Borrow<command::BufferImageCopy>,
+     {
+        let image_raw = match dst {
+            &n::Image::Surface(s) => s,
+            &n::Image::Texture(t) => t
+        };
+
+        let regs = regions.into_iter().collect::<Vec<_>>();
+
+        assert!(regs.len() >= 1);
```
msiglreith: nit: `regs.is_empty()`

src/backend/gl/src/command.rs
```diff
@@ -636,17 +676,28 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
+
+        for reg in regs {
+            self.push_cmd(Command::CopyBufferToTexture(src.raw, image_raw, reg.borrow().clone()));
```
msiglreith: out-of-scope for this PR but we might should add functionality to push multiple commands via an iterator, saving the collect above and multiple lock acquistions.

src/backend/gl/src/queue.rs
```diff
@@ -472,8 +472,35 @@ impl CommandQueue {
...
             }
-            com::Command::BindBlendSlot(slot, blend) => {
-                state::bind_blend_slot(&self.share.context, slot, &blend);
+            com::Command::BindBlendSlot(slot, ref blend) => {
+                state::bind_blend_slot(&self.share.context, slot, blend);
+            }
+            com::Command::BindAttribute(ref attribute, handle, stride) => unsafe {
+                let gl = &self.share.context;
+                gl.BindBuffer(gl::ARRAY_BUFFER, handle);
+                gl.VertexAttribPointer(attribute.location, attribute.size, attribute.format, gl::FALSE,
```
msiglreith: There are several function variants of `VertexAttribPointer` (https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glVertexAttribPointer.xhtml)
Could you either fix it now or add a todo for later?

src/backend/gl/src/command.rs
grovesNL: I don't think so, I don't see any advantage currently. I could simplify all of the `Option<Vec<_>>` fields if we don't need to distinguish between these.

src/backend/gl/src/command.rs
grovesNL: It's needed to avoid a second mutable borrow of `self` in `begin_renderpass` (see [the commit where it was added](https://github.com/gfx-rs/gfx/pull/1701/commits/72b591d136320c72904812c0ede1092dc6b27b6f)) without copying. Basically it accomplishes a partial borrow of `self`.

I could temporarily copy or collect all of the commands instead (or even use `mem::swap` to move it temporarily or something), but I think we may hit this in a few places. I'm open to more ideas here 😃

src/backend/gl/src/command.rs
grovesNL: Good catch, thanks. I will move it to `draw`/`draw_indexed`

src/backend/gl/src/command.rs
grovesNL: Is this a limitation of any of the backends? With GL at least I thought it was fine to avoid VBOs entirely and generate vertices using geometry shaders, or use `gl_VertexID` (i.e. full-screen triangle without a VBO), etc.

src/backend/gl/src/queue.rs
grovesNL: Is it fine to simply assume `glVertexAttribIPointer` for all integer formats? Or is it possible in Vulkan to do something equivalent to calling `glVertexAttribPointer` with an integer format?

grovesNL:
Updated to address some of the review feedback, thanks!

src/backend/gl/src/command.rs
kvark: Having the function is good, but perhaps we could add a short explanation in comments/documentation?

src/backend/gl/src/command.rs
```diff
@@ -309,6 +303,34 @@ impl RawCommandBuffer {
...
+            ..
+        } = self.cache;
+
+        for attribute in attributes {
+            let binding = attribute.binding as usize;
+
+            assert!(vertex_buffer_descs.len() > binding);
+            assert!(vertex_buffers.len() > binding);
+
+            let desc = &vertex_buffer_descs[binding];
```
kvark: is there a point in `assert!` above if we are just indexing here and thus will get a trap even in release mode?

src/backend/gl/src/command.rs
kvark: @msiglreith 
> Zero vertex buffers should be/is invalid

where? why?

src/backend/gl/src/command.rs
```diff
@@ -636,17 +673,28 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
+     ) where
+         T: IntoIterator,
+         T::Item: Borrow<command::BufferImageCopy>,
+     {
+        let image_raw = match dst {
+            &n::Image::Surface(s) => s,
+            &n::Image::Texture(t) => t
+        };
+
+        let regs = regions.into_iter().collect::<Vec<_>>();
```
kvark: why do we need to collect here as opposed to just iterating?

src/backend/gl/src/command.rs
```diff
@@ -682,6 +730,8 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
                 self.cache.error_state = true;
             }
         }
+
+        self.bind_attributes();
```
kvark: shouldn't this come before the actual draw commands?

src/backend/gl/src/command.rs
```diff
@@ -717,6 +767,8 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
```
kvark: same here

src/backend/gl/src/command.rs
```diff
@@ -790,5 +842,26 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     }
 }
 
+fn push_cmd_internal(id: &u64, memory: &mut Arc<Mutex<pool::BufferMemory>>, buffer: &mut BufferSlice, cmd: Command) {
+    let slice = {
```
kvark: is this scope needed here?

src/backend/gl/src/device.rs
```diff
@@ -580,18 +601,44 @@ impl d::Device<B> for Device {
...
+        let raw = unsafe {
+            let mut raw = 0;
+            gl.GenTextures(1, &mut raw);
+            raw
+        };
+
+        let (width, height) = match kind {
+            i::Kind::D2(w, h, _aa) => unsafe {
+                gl.BindTexture(gl::TEXTURE_2D, raw);
+                gl.TexStorage2D(gl::TEXTURE_2D, 1, gl::SRGB8_ALPHA8, w as _, h as _);
```
kvark: let's assert on a format and sample count

src/backend/gl/src/queue.rs
```diff
@@ -472,8 +472,45 @@ impl CommandQueue {
...
+                use native::VertexAttribFunction::*;
+
+                let &native::AttributeDesc { location, size, format, offset, .. } = attribute;
+                let offset = offset as *const gl::types::GLvoid;
+                let gl = &self.share.context;
+
+                gl.BindBuffer(gl::ARRAY_BUFFER, handle);
+
+                match function_type {
+                    Float => gl.VertexAttribPointer(location, size, format, gl::FALSE, stride, offset),
```
kvark: no support for normalized values?

src/backend/gl/src/queue.rs
```diff
@@ -472,8 +472,45 @@ impl CommandQueue {
...
+                    Integer => gl.VertexAttribIPointer(location, size, format, stride, offset),
+                    Double => gl.VertexAttribLPointer(location, size, format, stride, offset),
+                }
+                
+                gl.EnableVertexAttribArray(attribute.location);
```
kvark: nit: just "location"

src/backend/gl/src/queue.rs
grovesNL: Not yet, I'll try to add a follow-up PR soon

src/backend/gl/src/command.rs
grovesNL: I just moved the body from `push_cmd`. I agree though, doesn't appear to be necessary so I'll remove it

src/backend/gl/src/command.rs
grovesNL: Yes, good catch. I missed the order when I moved this :sweat_smile:

src/backend/gl/src/command.rs
grovesNL: It's really only used for the assertion that the iterator isn't empty before submitting any commands. We could change it to assert at least one command was submitted instead, or something else?

src/backend/gl/src/command.rs
grovesNL: Agreed, it would probably be useful to change these to `error!` instead to let the user know when the vertex buffers or descriptions haven't been bound

grovesNL:
Thanks for all the feedback so far! Updated again

grovesNL:
Travis failure on Linux only: `oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"` I guess this is the incremental compilation issue (like #1698)?

src/backend/gl/src/command.rs
msiglreith: It's a common pattern in Vulkan that no-op calls are invalid.
In particular for bind_vertex_buffers: https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCmdBindVertexBuffers.html
```
bindingCount must be greater than 0
```

msiglreith: > With GL at least I thought it was fine to avoid VBOs entirely

That's totally fine, but then you wouldn't call `bind_vertex_buffers` at all.

src/backend/gl/src/queue.rs
msiglreith: > Is it fine to simply assume glVertexAttribIPointer for all integer formats

Yes, if you mean `Uint/Int` suffix for Formats.
For example, `Format::R8Uint` and `Format::R8Uscaled` only differ in the calling function but not in the parameters.

msiglreith:
> I guess this is the incremental compilation issue 

Hm looks like some travis internal error as compilation hasn't even started.

src/backend/gl/src/command.rs
kvark: do we have to assert on the empty list?

src/backend/gl/src/command.rs
kvark: @msiglreith I think we'll be fine with allowing that. The portability layer can assert if it has to, but gfx-hal doesn't have to do this.

grovesNL: Makes sense. I thought it was possible to explicitly unbind vertex buffers but it appears that’s not allowed

src/backend/gl/src/command.rs
grovesNL: I was going by [`regionCount must be greater than 0`](https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCmdCopyBufferToImage.html) but I don’t know if we should be enforcing it somewhere else?

kvark:
> The job exceeded the maximum time limit for jobs, and has been terminated.

unfortunate :(
bors r+

kvark:
Something is off with Travis + GCC on stable :(
bors r+

PR #1703, opened 2017-12-29
Fix handling of optional function constants in Metal backend

JohnColanduoni:
A shader with only optional function constants will be handled improperly by the current Metal backend, resulting in a crash. `MTLLibrary` distinguishes between creating a function without passing any specialization constants and creating a function with an empty list of specialization constants. If the function has only optional constants the former method produces a reflection-only function, while the latter produces a function suitable for inclusion in a PSO. This PR detects this edge case and handles it appropriately.

It also implements a few missing functions to get the `gfx_render` quad example working on Metal (the high level quad example also exercises the above edge case).

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -60,6 +60,41 @@ fn create_function_constants(specialization: &[pso::Specialization]) -> metal::F
     constants_raw
 }
 
+fn get_final_function(library: &metal::LibraryRef, entry: &str, specialization: &[pso::Specialization]) -> Result<metal::Function, ()> {
+    let initial_constants = if specialization.is_empty() {
+        None
+    } else {
+        Some(create_function_constants(specialization))
+    };
+
+    let mut mtl_function = library
+        .get_function(entry, initial_constants)
+        .map_err(|_| {
+            error!("Invalid vertex shader entry point");
+            ()
+        })?;
+    let has_more_function_constants = unsafe {
+        let dictionary: *mut ::objc::runtime::Object = msg_send![mtl_function, functionConstantsDictionary];
+        let count: NSUInteger = msg_send![dictionary, count];
+        count > 0
+    };
+    if has_more_function_constants {
+        // TODO: check that all remaining function constants are optional, otherwise return an error
+        if specialization.is_empty() {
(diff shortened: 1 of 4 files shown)


src/backend/metal/src/device.rs
```diff
@@ -60,6 +60,41 @@ fn create_function_constants(specialization: &[pso::Specialization]) -> metal::F
...
+    let initial_constants = if specialization.is_empty() {
+        None
+    } else {
+        Some(create_function_constants(specialization))
+    };
+
+    let mut mtl_function = library
+        .get_function(entry, initial_constants)
+        .map_err(|_| {
+            error!("Invalid vertex shader entry point");
```
grovesNL: nit: should probably just be `Invalid shader entry point`

src/backend/metal/src/device.rs
```diff
@@ -60,6 +60,41 @@ fn create_function_constants(specialization: &[pso::Specialization]) -> metal::F
...
+    };
+    if has_more_function_constants {
+        // TODO: check that all remaining function constants are optional, otherwise return an error
+        if specialization.is_empty() {
+            // These may be optional function constants, in which case we need to specialize the function with an empty set of constants
+            // or we'll get an error when we make the PSO
+            mtl_function = library
+                .get_function(entry, Some(create_function_constants(&[])))
+                .map_err(|_| {
+                    error!("Invalid vertex shader entry point");
```
grovesNL: nit: same as above

src/backend/metal/src/device.rs
```diff
@@ -396,16 +424,9 @@ impl Device {
...
-            } else {
-                Some(create_function_constants(fragment_entry.specialization))
-            };
-
-            let mtl_fragment_function = fs_lib
-                .get_function(fs_entry, fs_constants)
+            let mtl_fragment_function = get_final_function(&fs_lib, fs_entry, fragment_entry.specialization)
                 .map_err(|_| {
-                    error!("invalid pixel shader entry point");
+                    error!("Invalid vertex shader entry point");
```
grovesNL: nit: should be `fragment`

kvark:
@grovesNL thanks for taking a look (and spotting those issues)!

PR #1707, opened 2017-12-29
Add error handling for logical device creation

msiglreith:
Add error handling for `PhysicalDevice::open` using the Vulkan return codes.
Early feedback appreciated!

~~**[WIP]** Implement remaining backends~~

### DIFF
--- a/src/backend/vulkan/src/result.rs
+++ b/src/backend/vulkan/src/result.rs
@@ -1,0 +1,91 @@
+
+use ash::vk;
+
+use hal::adapter::DeviceCreationError;
+
+// Generic error codes from Vulkan
+#[derive(Debug)]
+pub(crate) enum Error {
+    OutOfHostMemory,
+    OutOfDeviceMemory,
+    InitializationFailed,
+    DeviceLost,
+    MemoryMapFailed,
+    LayerNotPresent,
+    ExtensionNotPresent,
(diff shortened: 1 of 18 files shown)


src/backend/vulkan/src/result.rs
```diff
@@ -0,0 +1,91 @@
...
+macro_rules! from_error {
+    { $($name:ident { $($base_error:ident => $err:ident,)* },)* } => {
+        $(
+            impl From<Error> for $name {
+                fn from(err: Error) -> Self {
+                    match err {
+                        $(
+                            Error::$base_error => $name::$err,
+                        )*
+                        _ => unreachable!("Unexpected error code ({:?}). Non specification conformant driver.", err),
```
msiglreith: This part is nagging on me a bit, as I don't have experience on how much drivers do things by the spec regarding return codes..

kvark: Agree. Can we have some sort of "unrecognized" error instead with logging the case to `error!`?

msiglreith: Addind an `Unrecognized` variant to all error structs?

kvark: Nah, on a second thought, we should be respecting the spec.

PR #1712, opened 2017-12-31
Reusable and Secondary command buffers

AlphaModder:
This PR introduces an API for multiple submit and secondary command buffers.
HAL support exists, and the backends at least compile, though the new functionality is not guaranteed to be implemented anywhere other than Vulkan. On Metal and GL, we will have to emulate secondary buffers, as these backends have no analogous concept. On DX12, it may be possible to at least partially map them to commmand 'bundles', but more research needs to be conducted on this front.

### Multi-Submit Command Buffers ###
Previously, we required the user to reset and rerecord all command buffers before resubmitting them. Now it is possible to specify that a command buffer may be submitted more than once, and furthermore what states (executable, pending) it can be in to do so.

Multi-submit buffers are implemented via the new `Shot` type parameter on `CommandBuffer`, which can be `OneShot` (default, previous behavior) or `MultiShot` (new behavior). These may not be the best names, but I can't see anything better. The use of a generic parameter here allows compile-time prevention of the resubmission of a `OneShot` buffer. By default, a `MultiShot` command buffer can only be submitted while it is in the **executable** state: that is, after it was recorded or after it has completed a previous execution. 

However, a flag on `CommandPool::acquire_command_buffer` and `acquire_secondary_command_buffer`, if enabled allows the command buffer to be submitted while it is in the **pending** state. This allows more than one instance of a command buffer to be executed at once. Note that the flag is not enforced at compile-time, because as far as I can tell, this would require a total restructuring of the `CommandBuffer` interface and lifecycle. I may look into this as an experiment, but for now it is out of scope, and the user is reponsible for not submitting a pending buffer for which the flag is not enabled.

### Secondary Command Buffers ###
We distinguish between two types of secondary command buffers. The first type is those that exist outside of a render pass, and ares represented by a `CommandBuffer` with the new `Level` type parameter set to `Secondary`. `Submit`s from these command buffers are passed to `command::CommandBuffer::execute_commands`, rather than `queue::Submission::submit`, and they are acquired via `CommandPool::acquire_secondary_command_buffer`. Otherwise, they function identically to a standard command buffer. 

The second type of secondary command buffer is those that exist **inside** a renderpass, or more specifically a subpass with `SubpassContents::SecondaryBuffers`. These are a separate type entirely from regular command buffers, `command::renderpass::SubpassCommandBuffer`, due to the fact that they exist somewhat outside of the ordinary capability system. Also, due to the fact a large part of their functionality is identical to `RenderPassInlineEncoder`'s , to avoid code duplication and desynchronization issues, I have created a struct `RenderSubpassCommon` that both types `Deref` to in order to hold this shared implementation.

Using the `begin_renderpass_inline` and `begin_renderpass_secondary` (better name?) methods on a graphics `CommandBuffer`, it is possible to obtain either a `RenderPassInlineEncoder` or its sibling `RenderPassSecondaryEncoder`, a variant which supports only `execute_commands` of `Submit`s from `SubpassCommandBuffer`s. In order to reuse the `Submit` mechanism here, it was unfortunately necessary to invent an otherwise useless value for `Capability`: `command::renderpass::Subpass` (an uninhabited enum). Both types of encoder possess `next_subpass_inline` and `next_subpass_secondary` methods, which take self by value and produce an encoder of the appropriate type. Note that `begin_renderpass_secondary` and `next_subpass_secondary` are only available on primary command buffers for obvious reasons.

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1712)


src/hal/src/command/mod.rs
```diff
@@ -15,53 +15,95 @@ pub use self::raw::RawCommandBuffer;
...
-
-impl<B: Backend, C> Submit<B, C> {
+/// Trait used to represent a single submission that can be added to a `Submission`.
+pub unsafe trait Submit<B: Backend, C>: Send {
+    ///
+    unsafe fn new(buffer: B::CommandBuffer) -> Self;
     ///
-    pub unsafe fn new(buffer: CommandBuffer<B, C>) -> Self {
-        Submit(buffer.raw.clone(), PhantomData)
+    unsafe fn buffer(self) -> B::CommandBuffer;
```
kvark: convention-wise, should be `into_buffer` or `unwrap`

src/hal/src/command/mod.rs
```diff
@@ -15,53 +15,95 @@ pub use self::raw::RawCommandBuffer;
...
+unsafe impl<B: Backend, C> Submit<B, C> for Resubmit<B, C> {
+    unsafe fn new(buffer: B::CommandBuffer) -> Self { 
+        Resubmit(buffer.clone(), PhantomData)
+    }
+
+    unsafe fn buffer(self) -> B::CommandBuffer { self.0 }
+}
+
+/// A dummy implementor of `Submit` used as a hack for downgraded command buffers.
+/// Since these cannot be submitted, an instance of this struct should never actually exist.
```
kvark: we should use `enum Something {}` for stuff that can't have instances

src/hal/src/queue/submission.rs
```diff
@@ -67,14 +67,16 @@ where
...
-    /// All submits for this call must be of the same capability.
+    /// All submits for this call must be of the same type.
     /// Submission will be automatically promoted to to the minimum required capability
     /// to hold all passed submits.
-    pub fn submit<S>(mut self, submits: &[Submit<B, S>]) -> Submission<'a, B, <(C, S) as Upper>::Result>
+    pub fn submit<I, S, K>(mut self, submits: I) -> Submission<'a, B, <(C, K) as Upper>::Result>
     where
-        (C, S): Upper
+        I: Iterator<Item=S>,
+        S: Submit<B, K>,
```
kvark: I don't think you actually need `S` parameter here, could just do `I:Item: Submit<B, K>` instead

src/hal/src/command/mod.rs
```diff
@@ -15,53 +15,95 @@ pub use self::raw::RawCommandBuffer;
 pub use self::renderpass::*;
 pub use self::transfer::*;
 
-/// Thread-safe finished command buffer for submission.
-pub struct Submit<B: Backend, C>(pub(crate) B::CommandBuffer, PhantomData<C>);
-unsafe impl<B: Backend, C> Send for Submit<B, C> {}
-
-impl<B: Backend, C> Submit<B, C> {
+/// Trait used to represent a single submission that can be added to a `Submission`.
+pub unsafe trait Submit<B: Backend, C>: Send {
```
kvark: There is no point to have a trait with methods if all the implementations do them in the same way. How about:
```rust
trait Shot {}
enum OneShot {}
enum MultiShot {}
impl Shot for OneShot {}
impl Shot for MultiShot {}

pub struct Submit<B: Backend, C, S>(pub(crate) B::CommandBuffer, PhantomData<(C, S)>);
impl<B: Backend, C, S: Shot> Submit<B, C> {
  ...
}
```

AlphaModder: Sure, that works. I still think `Shot` is kinda a strange name though...

src/hal/src/queue/submission.rs
AlphaModder: How can I specify that `I` is an iterator at all then?

AlphaModder: Nevermind.

AlphaModder:
Uh oh, seems for some reason `#[derive]` let me do something it shouldn't have. Will have to rethink how this works a bit.

src/hal/src/command/mod.rs
```diff
@@ -15,53 +15,72 @@ pub use self::raw::RawCommandBuffer;
...
     }
 
     /// Finish recording commands to the command buffers.
     ///
     /// The command buffer will be consumed and can't be modified further.
     /// The command pool must be reset to able to re-record commands.
-    pub fn finish(self) -> Submit<B, C> {
-        unsafe { Submit::new(self) }
+    pub fn finish(self) -> Submit<B, C, S> {
+        Submit(self.raw.clone(), PhantomData, PhantomData)
```
msiglreith: Why do we need to clone here?

src/hal/src/command/raw.rs
```diff
@@ -16,7 +16,7 @@ use super::{
 ///
 pub trait RawCommandBuffer<B: Backend>: Clone + Send {
     ///
-    fn begin(&mut self);
+    fn begin(&mut self, reusable: bool);
```
msiglreith: Could we use flags here as we will also want to support simultaneous usage and renderpass only command buffers.

src/hal/src/command/mod.rs
```diff
@@ -15,53 +15,72 @@ pub use self::raw::RawCommandBuffer;
...
+impl Shot for MultiShot { }
 
-impl<B: Backend, C> Submit<B, C> {
+/// A trait representing a command buffer that can be added to a `Submission`.
+pub unsafe trait Submittable<B: Backend, C> {
     ///
-    pub unsafe fn new(buffer: CommandBuffer<B, C>) -> Self {
-        Submit(buffer.raw.clone(), PhantomData)
-    }
+    unsafe fn into_buffer(self) -> B::CommandBuffer;
```
msiglreith: I would prefer returning a reference to a commandbuffer to avoid costly clones and adjust the queue submission interface accordingly.

src/hal/src/command/raw.rs
AlphaModder: Yeah, in my local repository I turned this into flags once I started working on secondary buffers.

src/hal/src/command/mod.rs
AlphaModder: In order to enforce `OneTime` command buffers, the queue submission interface must take all `Submittable`s by value. Returning a reference would require them to be kept alive somehow.

src/hal/src/command/mod.rs
AlphaModder: Whoops, we don't.

src/hal/src/command/mod.rs
msiglreith: I though about something like [this](https://play.rust-lang.org/?gist=f518fc0bbb9c19d4df7018368c01bb22&version=stable)
I don't remember the exact internals regarding queue submission, but this should be possible.

src/hal/src/command/mod.rs
AlphaModder: Oh wait, we do. It was there before, just hidden in `Submit::new`.

src/hal/src/command/mod.rs
AlphaModder: Oh, I never even realized that was possible! Cool!

AlphaModder: @msiglreith Hm, unfortunately the issue here is that the new structs would have to be generic, which would lead to really annoying signatures like `CommandBuffer<'a, B, C, L, SubmitOnce<B, C, L>>`.

src/hal/src/command/mod.rs
```diff
@@ -10,58 +10,123 @@ mod raw;
...
+/// A trait representing a command buffer that can be added to a `Submission`.
+pub unsafe trait Submittable<B: Backend, C, L: Level> {
     ///
-    pub unsafe fn new(buffer: CommandBuffer<B, C>) -> Self {
-        Submit(buffer.raw.clone(), PhantomData)
+    unsafe fn into_buffer(self) -> B::CommandBuffer;
+}
+
+/// Thread-safe finished command buffer for submission.
+pub struct Submit<B: Backend, C, S: Shot, L: Level> {
```
kvark: it's non-idiomatic to have struct generic bounds that are not required

src/hal/src/command/mod.rs
```diff
@@ -10,58 +10,123 @@ mod raw;
...
+    pub(crate) buffer:  B::CommandBuffer,
+    pub(crate) _capability: PhantomData<C>,
```
kvark: or we could have `_marker: PhantomData<(C, S, L)>`

src/hal/src/command/mod.rs
```diff
@@ -10,58 +10,123 @@ mod raw;
...
+    pub(crate) _shot: PhantomData<S>,
+    pub(crate) _level: PhantomData<L>
+}
+impl<B: Backend, C, S: Shot, L: Level> Submit<B, C, S, L> {
+    fn new(buffer: B::CommandBuffer) -> Self {
```
kvark: could be replaced by `impl From<B::CommandBuffer> for Submit`?

src/hal/src/command/mod.rs
```diff
@@ -10,58 +10,123 @@ mod raw;
...
     }
 }
 
+unsafe impl<B: Backend, C, L: Level> Submittable<B, C, L> for Submit<B, C, OneShot, L> {
+    unsafe fn into_buffer(self) -> B::CommandBuffer { self.buffer }
+}
+unsafe impl<'a, B: Backend, C, L: Level> Submittable<B, C, L> for &'a Submit<B, C, MultiShot, L> {
+    unsafe fn into_buffer(self) -> B::CommandBuffer { self.buffer.clone() }
+}
+unsafe impl<B: Backend, C, S: Shot, L: Level> Send for Submit<B, C, S, L> {}
```
kvark: and impls like this are going to be simpler if the struct doesn't have as many generic bounds

src/hal/src/command/mod.rs
```diff
@@ -10,58 +10,123 @@ mod raw;
...
+
+/// A convenience for not typing out the full signature of a secondary command buffer.
+pub type SecondaryCommandBuffer<'a, B: Backend, C, S: Shot = OneShot> = CommandBuffer<'a, B, C, S, Secondary>;
+
 /// Command buffer with compute, graphics and transfer functionality.
-pub struct CommandBuffer<'a, B: Backend, C> {
+pub struct CommandBuffer<'a, B: Backend, C, S: Shot = OneShot, L: Level = Primary> {
```
kvark: FYI, I'm not attached to `OneShot/MultiShot` terms either :) I just saw them in [ipc_channel](https://docs.rs/ipc-channel/0.9.1/ipc_channel/ipc/struct.IpcOneShotServer.html) crate and thought it sounds ok

src/hal/src/command/mod.rs
```diff
@@ -10,58 +10,123 @@ mod raw;
...
     }
 }
 
-impl<'a, B: Backend, C> Drop for CommandBuffer<'a, B, C> {
+impl<'a, B: Backend, C, S: Shot> CommandBuffer<'a, B, C, S, Primary> {
+    ///
+    pub fn execute_commands<I, Z, K>(&mut self, submits: I) 
+    where
+        I: Iterator<Item=Z>,
+        Z: Submittable<B, K, Secondary>,
```
kvark: could reduce the `fn` signature a bit by writing `I::Item: Submittable<..>`

src/hal/src/command/mod.rs
```diff
@@ -10,58 +10,123 @@ mod raw;
...
+pub struct Submit<B: Backend, C, S: Shot, L: Level> {
+    pub(crate) buffer:  B::CommandBuffer,
+    pub(crate) _capability: PhantomData<C>,
+    pub(crate) _shot: PhantomData<S>,
+    pub(crate) _level: PhantomData<L>
+}
+impl<B: Backend, C, S: Shot, L: Level> Submit<B, C, S, L> {
+    fn new(buffer: B::CommandBuffer) -> Self {
+        Submit {
+            buffer: buffer,
```
kvark: nit: just `buffer,`

src/hal/src/command/raw.rs
```diff
@@ -326,4 +341,10 @@ pub trait RawCommandBuffer<B: Backend>: Clone + Send {
         offset: u32,
         constants: &[u32],
     );
+
+    ///
+    fn execute_commands(
+        &mut self,
+        buffers: &[B::CommandBuffer],
```
kvark: should also use iterators, like the other command buffer methods do

src/hal/src/command/renderpass.rs
```diff
@@ -146,10 +117,150 @@ impl<'a, B: Backend> RenderPassInlineEncoder<'a, B> {
...
+    }
+
+    ///
+    pub fn next_subpass_inline(mut self) -> Self {
+        self.get_buffer().next_subpass(SubpassContents::Inline);
+        self
+    }
+
+    #[inline(always)]
+    fn get_buffer(&mut self) -> &mut B::CommandBuffer { self.0.as_mut().unwrap() }
```
kvark: I think it should be named `buffer_mut`

src/hal/src/command/renderpass.rs
```diff
@@ -17,39 +18,9 @@ pub enum SubpassContents {
...
-
-    ///
-    pub fn next_subpass_inline(self) -> Self {
-        self.0.next_subpass(SubpassContents::Inline);
-        self
-    }
-
+// This macro contains all methods for all commands submittable during a subpass.
+// It is used to implement the identical portions of RenderPassInlineEncoder and SubpassCommandBuffer's API.
+macro_rules! add_subpass_commands { ($buf:ident) => {
```
kvark: can we share the implementations without meta-programming (please?), maybe like this:
```rust
struct RenderSubpassCommon {
  // common methods
}
impl Deref<RenderSubpassCommon> for RenderPassInlineEncoder {..}
impl Deref<RenderSubpassCommon> for SubpassCommandBuffer {..}
```

src/hal/src/command/renderpass.rs
```diff
@@ -146,10 +117,150 @@ impl<'a, B: Backend> RenderPassInlineEncoder<'a, B> {
     // TODO: set_stencil_write_mask
     // TODO: pipeline barrier (postponed)
     // TODO: begin/end query
+}}
+
+///
+pub struct RenderPassInlineEncoder<'a, B: Backend, L: Level>(pub(crate) Option<&'a mut B::CommandBuffer>, PhantomData<L>)
```
kvark: why is there an `Option`?

src/hal/src/command/renderpass.rs
```diff
@@ -146,10 +117,150 @@ impl<'a, B: Backend> RenderPassInlineEncoder<'a, B> {
...
+        C: Supports<Graphics>,
+        T: IntoIterator,
+        T::Item: Borrow<ClearValue>,
+    {
+        cmd_buffer.raw.begin_renderpass(
+            render_pass,
+            frame_buffer,
+            render_area,
+            clear_values,
+            SubpassContents::SecondaryBuffers);
```
kvark: nit: pretty sure the closing bracket should go to a new line

src/hal/src/command/mod.rs
AlphaModder: So, just remove the `Level` trait entirely, then?

src/hal/src/command/mod.rs
AlphaModder: That's a much better idea.

src/hal/src/command/mod.rs
AlphaModder: Sure.

src/hal/src/command/renderpass.rs
AlphaModder: Would this work without requiring `use RenderSubpassCommon`?

src/hal/src/command/renderpass.rs
AlphaModder: So that it's possible to move the raw buffer out of the encoder in `next_subpass_inline`/`secondary`. You can't directly remove it, even in a `self` method because of the `Drop` impl, so I just `take` it and make the `Drop` impl a no-op when there's a `None`.

src/hal/src/command/renderpass.rs
AlphaModder: That works.

src/hal/src/command/mod.rs
AlphaModder: Actually, this seems to have some bizarre coherence issue. Maybe it somehow thinks `B::CommandBuffer` could be `Submit` and then it would overlap with `From<T> for T`? Seems odd, considering `Submit` doesn't impl `RawCommandBuffer`.

src/hal/src/command/renderpass.rs
kvark: Sure, since `RenderSubpassCommon` is a struct, no a trait

src/hal/src/command/renderpass.rs
kvark: Can the secondary encoder borrow the primary encoder or it's command buffer instead of moving it in and out?

src/hal/src/command/renderpass.rs
AlphaModder: Hm, not sure how to implement this when `RenderPassInlineEncoder` borrows its command buffer but `SubpassCommandBuffer` owns its.

src/hal/src/command/renderpass.rs
AlphaModder: Wouldn't that cause an issue where it would call `drop` up the tree of borrows, rather than once like it should?

kvark:
A few more things to address:

---

Reviewed 35 of 35 files at r1.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[reftests/suite.ron, line 3 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YYEVF6t9pJ7nXxFZD:-L3YYEVGOWBijtZXxj7x:boxcoar) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/reftests/suite.ron#L3)):*
> ```Markdown
> {
> 	"basic": {
> 		"render-pass-clear": (
> ```

finally!

---

*[reftests/suite.ron, line 12 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YYFg1xRcYCCu6WNhl:-L3YYFg2SQN_lkL5jf8I:b31deo) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/reftests/suite.ron#L12)):*
> ```Markdown
> 		),
> 	},
> }
> ```

nit: let's have a newline here

---

*[src/backend/dx12/src/command.rs, line 1567 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YYvfsP_wgXCSoCGb1:-L3YYvfuBsVfnTajqzDk:b-ciq117) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/backend/dx12/src/command.rs#L1567)):*
> ```Rust
>     ) where
>         I: IntoIterator,
>         I::Item: Borrow<CommandBuffer> 
> ```

nit: trailing comma is missing

---

*[src/backend/dx12/src/lib.rs, line 311 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YZ3Fns6jAMUZ0aXSq:-L3YZ3FoeKz7npU6bxv6:b-93tkdl) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/backend/dx12/src/lib.rs#L311)):*
> ```Rust
>     ) where 
>         IC: IntoIterator, 
>         IC::Item: Borrow<command::CommandBuffer> 
> ```

here and everywhere else: missing comma at the end
also, something is wrong with line endings? Reviewable shows red dots

---

*[src/backend/dx12/src/pool.rs, line 57 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YZEkJzVJLoLtq6gj6:-L3YZEkK4srQDJtpQZco:b30rxzi) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/backend/dx12/src/pool.rs#L57)):*
> ```Rust
>     }
> 
>     fn allocate(&mut self, num: usize, secondary: bool) -> Vec<CommandBuffer> { // TODO: Implement secondary buffers
> ```

let's switch this from `bool` to a clear `enum`

---

*[src/backend/metal/src/command.rs, line 228 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YZzzWLRT-nJ4ofaf_:-L3YZzzWLRT-nJ4ofafa:b-kp1204) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/backend/metal/src/command.rs#L228)):*
> ```Rust
>         };
> 
>         let buffers = submit.cmd_buffers.into_iter().collect::<Vec<_>>();
> ```

we should be able to avoid the collection here

---

*[src/backend/vulkan/src/command.rs, line 921 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3Y_7gRTr-1pAR2VGop:-L3Y_7gRTr-1pAR2VGoq:b-ceykcg) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/backend/vulkan/src/command.rs#L921)):*
> ```Rust
>         unsafe { self.device.0.cmd_execute_commands(self.raw, &command_buffers); }
>     }
> }
> ```

nit: newline missing

---

*[src/backend/vulkan/src/conv.rs, line 749 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3Y_Ba2U5JbwnJ_PqN6:-L3Y_Ba2U5JbwnJ_PqN7:b-4yr6dt) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/backend/vulkan/src/conv.rs#L749)):*
> ```Rust
>     }
>     usage
> }
> ```

newline

---

*[src/hal/src/command/mod.rs, line 42 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YfWI9UXjy4ETuGnIH:-L3YfWI9UXjy4ETuGnII:b-bgtwva) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/hal/src/command/mod.rs#L42)):*
> ```Rust
> 
> /// Thread-safe finished command buffer for submission.
> pub struct Submit<B: Backend, C, S: Shot, L: Level>(pub(crate) B::CommandBuffer, pub(crate) PhantomData<(C, S, L)>);
> ```

let's not have the generic bounds on the struct definition and implementations that don't need them (like `new()`)

---

*[src/hal/src/command/renderpass.rs, line 162 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YbIuKIDsXRJ5oQ3Mz:-L3YbIuKIDsXRJ5oQ3N-:b5op586) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/hal/src/command/renderpass.rs#L162)):*
> ```Rust
>     ///
>     pub fn next_subpass_secondary(mut self) -> RenderPassSecondaryEncoder<'a, B> {
>         self.0.as_mut().unwrap().0.next_subpass(SubpassContents::SecondaryBuffers);
> ```

we should just `take().unwrap()` straight away

---

*[src/hal/src/command/renderpass.rs, line 228 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3YcfH_dRt1Rae8DY4I:-L3YcfHaNFDFa-obvkxx:bn84lkl) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/hal/src/command/renderpass.rs#L228)):*
> ```Rust
>     ///
>     pub fn next_subpass_inline(mut self) -> RenderPassInlineEncoder<'a, B, Primary> {
>         self.0.as_mut().unwrap().next_subpass(SubpassContents::Inline);
> ```

similarly, let's just start by `take()`-ing it out

---

*[src/hal/src/queue/submission.rs, line 29 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-L3Yf8C5M0Cthw7TrgRs:-L3Yf8C6d-KS9Fme0RfS:b-79rbez) ([raw file](https://github.com/gfx-rs/gfx/blob/15479b03d874fdc1930320fe25bea1196b888f91/src/hal/src/queue/submission.rs#L29)):*
> ```Rust
> /// Submission information for a command queue.
> pub struct Submission<'a, B: Backend, C> {
>     cmd_buffers: SmallVec<[Box<Borrow<B::CommandBuffer> + 'a>; 16]>,
> ```

can we have this generic over `Borrow<..>` as opposed to boxing a trait?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1712#-:-L3YfbfbcUbFQ2NvqFoS:bz6smrz)*

AlphaModder:
@kvark I'm finding reviewable rather difficult to use, so I'm just gonna reply here. I just have two small issues.
- I tried a couple different ways that didn't involve using a `collect` but I couldn't find anything that would satisfy the borrow-checker and allow special treatment of the last element.
- Unfortunately `Submission::cmd_buffers` has to be boxed traits because the collection may be of heterogeneous type. This overhead is only present in the safe abstraction, `RawSubmission` is generic.
I think I've covered everything other than those two issues.

msiglreith:
bors r=kvark

kvark:
@AlphaModder please rebase

PR #1714, opened 2018-01-01
hal: Allow creating multiple logical devices from a physical device

msiglreith:
Following Vulkan specification and required to get portability running with dx12.
Currently, implementation clones the adapter, but that's a hack and creating multiple devices would result in a mess as all would be the same device under the hood on dx12.

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -32,7 +32,7 @@ use wio::com::ComPtr;
 use std::{mem, ptr};
 use std::os::windows::ffi::OsStringExt;
 use std::ffi::OsString;
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
 
 pub(crate) struct HeapProperties {
     pub page_property: d3d12::D3D12_CPU_PAGE_PROPERTY,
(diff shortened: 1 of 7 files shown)


src/backend/dx12/src/lib.rs
```diff
@@ -146,12 +146,21 @@ pub struct PhysicalDevice {
     private_caps: Capabilities,
     heap_properties: &'static [HeapProperties; NUM_HEAP_PROPERTIES],
     memory_properties: hal::MemoryProperties,
+    // Indicates that there is currently an active logical device.
+    // Opening the same adapter multiple times will return the same D3D12Device again.
+    is_open: Arc<Mutex<bool>>,
```
kvark: `AtomicBool`?

msiglreith: I want to lock the whole creation process and only set `is_open` to `true` once we finished creating the logical device.
Both approaches would work, but a mutex feels safe here.
pseudo code for atomics;
```
if !self.is_open { return Err(); }
self.is_open.set(true);
....
// check device error
if !SUCCESS { self.is_open.set(false); return Err(); }
...
// check queue creation error
if !SUCCESS { self.is_open.set(false); return Err(); }
...
Gpu { device, queues }
```

kvark:
sounds good, thank you!
bors r+

PR #1715, opened 2018-01-02
[v0.16] Hash for pipe::Meta

kvark:
Allows hashing and comparing pipeline states.

PR #1716, opened 2018-01-02
[pre-ll] Hash for pipe::Meta

kvark:
Brother of #1715


src/render/src/pso/resource.rs
```diff
@@ -69,7 +69,8 @@ pub struct Sampler(Option<(SamplerSlot, shade::Usage)>);
 ///
 /// - init: `&str` = name of the sampler/texture (assuming they match)
 /// - data: (`ShaderResourceView<T>`, `Sampler`)
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Derivative)]
+#[derivative(Clone, Debug, Eq, Hash(bound=""), PartialEq)]
```
vitvakatu: I wonder if we can do the same thing with `Eq`

PR #1718, opened 2018-01-03
dx12/vk: Implement vulkan based memory mapping

msiglreith:
Replace resource based mapping with device memory based mapping.
On dx12 this is implemented by a backing buffer resource covering the whole memory allocation. This currently results in a aliasing warnings in the debug layer. Requires further investigation about the safety of this approach.

The change would fix one of the large outstanding portability issues in my opinion. The branch is based on `portable` as it's experimental and lacks metal backend implementation, so most likely won't pass CI atm.


kvark:
Can we have metal compiling at least?

kvark:
waaait, why would we not want this in `master`?

msiglreith:
It would break all examples for an indefinite period of time 😄

kvark:
> It would break all examples for an indefinite period of time

You mean, all the 2.5 examples? :D shipit

msiglreith:
Updated the remaining crates to the point where it should pass CI.
Currently breaks all examples on metal (+gl) and quad_render won't be working correctly as the mapping part in `render` needs to be rewritten as it's heavily built around resource mapping.

I can redirect it towards the master branch if the mentioned points above are ok.

kvark:
I *think* we should put `unimplemented!()` where we are lazy to fill the gaps, merge this into `master` with CI passing, and then go from there.

msiglreith:
should be ready then!

src/backend/dx12/src/device.rs
```diff
@@ -1786,41 +1817,86 @@ impl d::Device<B> for Device {
...
+                &d3d12::D3D12_RANGE {
+                    Begin: 0,
+                    End: 0,
+                },
+            );
+        }
+    }
 
-        unsafe { (*buf.resource).Unmap(0, &written_range) };
+    fn flush_mapped_memory_ranges(&self, ranges: &[(&n::Memory, Range<u64>)]) {
```
kvark: use iterators?

src/hal/src/device.rs
```diff
@@ -316,82 +316,74 @@ pub trait Device<B: Backend> {
...
         T: Copy,
     {
-        let count = (range.end - range.start) as usize / mem::size_of::<T>();
-        self.acquire_mapping_raw(buffer, Some(range.clone()))
+        let len = range.end - range.start;
+        let count = len as usize / mem::size_of::<T>();
+        self.map_memory(memory, range.clone())
             .map(|ptr| unsafe {
-                let start_ptr = ptr.offset(range.start as isize) as *const _;
+                let start_ptr = ptr.offset(0) as *const _;
```
kvark: this becomes redundanty

src/hal/src/device.rs
```diff
@@ -316,82 +316,74 @@ pub trait Device<B: Backend> {
...
         -> Result<mapping::Writer<'a, B, T>, mapping::Error>
     where
         T: Copy,
     {
         let count = (range.end - range.start) as usize / mem::size_of::<T>();
-        self.acquire_mapping_raw(buffer, None)
+        self.map_memory(memory, range.clone())
             .map(|ptr| unsafe {
-                let start_ptr = ptr.offset(range.start as isize) as *mut _;
+                let start_ptr = ptr.offset(0) as *mut _;
```
kvark: same here

msiglreith:
done, thanks.
I should have updated the code after rebasing it. The codebase changes so fast 🎐

kvark:
Thank you!
bors r+

PR #1722, opened 2018-01-05
hal: Add unsafe resource memory binding methods for portability

msiglreith:


kvark:
Why is this needed for portability? Wouldn't the consumption of `UnboundImage` be easy to work around by just wrapping it in an `Option`-like enum?

msiglreith:
@kvark Sry, can't follow how this would look like. Would you mind elaborating?
As far as I understood it, the issue I had was due to requiring to consume the whole `Box<BufferEnum>`.

kvark:
Wouldn't something like that be sufficient in the portability layer?
```rust
enum Buffer {
  Temporary,
  Unbound(hal::UnboundBuffer),
  Bound(hal::Buffer),
}

extern fn vk_bind_buffer_memory(buf: Handle<Buffer>, mem: Memory) {
  let result = match mem::replace(&mut *buf, Buffer::Temporary) {
    Buffer::Unbound(ub) => hal_device.bind_buffer_memory(ub, mem),
    _ => panic!(""),
  };
  *buf = Buffer::Bound(result);
}
```

msiglreith:
Awesome! `mem::replace` seems to do the trick <3

PR #1728, opened 2018-01-07
vk: Fix shader module code length

msiglreith:
Casting from &[u8] to &[u32] requires adjusting the length.


kvark:
Why does the example work right now?

msiglreith:
Rechecked the specification and indeed, it's a u32 pointer but code size in bytes ..
Need to look again what's wrong with the lunarg sample

PR #1730, opened 2018-01-08
[RFC] Refactor queue creation

msiglreith:
Identify queue families by handle and use actual queue families only for device creation.
It hides the `RawQueueGroup` as implementation detail and gives the user a bit more contol over selecting the queue groups after creation.

Removes also the `open_with` function for now, as it's a bit awkward to use imo as we give up information about the order of queue families. I'm looking into replacing it with an alternative for single queue family selection using a filter, which is probably more common and sound.

Use-case scenario:
```rust
let adapter = instance.enumerate_adapters();
let (family_ids, families) = adapters.find_families(..);
let Gpu { device, queues } = adapter.open(families);
let main_queue = queues.take::<Graphics>(family_id[0]);
let compute_queue = queues.take::<Compute>(family_id[1]);

let graphics_pool = device.create_command_pool(main_queue.family());
```

Other motivation is implementation of portability layer to query queues from `(family_id, queue_id)`.

**WIP**: Backend implementations missing

### DIFF
--- a/src/hal/src/queue/family.rs
+++ b/src/hal/src/queue/family.rs
@@ -1,6 +1,7 @@
 //! Queue family and groups.
 
 use Backend;
+use backend::RawQueueGroup;
 use queue::{CommandQueue, QueueType};
 use queue::capability::{Capability, Graphics, Compute};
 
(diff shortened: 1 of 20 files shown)


src/hal/src/queue/family.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn supports_compute(&self) -> bool {
+        Compute::supported_by(self.queue_type())
+    }
+}
+
+/// Identifier for a queue family of a physical device.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct QueueFamilyId(pub usize);
+
+// Only needed for backend implementations.
```
kvark: Perhaps, we could have a special module `hal::backend` where we re-export all the things that are needed for backends only, so that we can see all this semi-legal public API in one place?

src/hal/src/queue/family.rs
```diff
@@ -0,0 +1,106 @@
...
+#[doc(hidden)]
+pub struct RawQueueGroup<B: Backend> {
+    pub family: B::QueueFamily,
+    pub queues: Vec<B::CommandQueue>,
+}
+
+impl<B: Backend> RawQueueGroup<B> {
+    #[doc(hidden)]
+    pub fn new(family: B::QueueFamily) -> Self {
```
kvark: would `impl From<B::QueueFamily>` fit here?

src/hal/src/queue/family.rs
```diff
@@ -0,0 +1,106 @@
...
+        Queues(queues)
+    }
+
+    /// Removes the queue family with the passed id from the queue list and
+    /// returns the queue group.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the family doesn't expose required queue capabilities.
+    pub fn take<C: Capability>(&mut self, id: QueueFamilyId) -> Option<QueueGroup<B, C>> {
```
kvark: looks good

src/hal/src/queue/family.rs
msiglreith: Sounds reasonable!

src/hal/src/queue/family.rs
msiglreith: yeah, would be applicable, but I have no clear preference between both approaches as it's backend only anyways. We want to hide the struct with it's function and it doesn't provide much benefit beside being a bit more ideomatic.

msiglreith:
bors r=kvark

PR #1731, opened 2018-01-08
[gl] Warden on CI, basic texture to buffer copies and memory mapping

kvark:
It's super annoying that GL doesn't have `GetTexSubImage`, which would make total sense in terms of API symmetry...

~~Next steps would be implementing the memory mapping.~~
Note:
We should probably use headless GL on CI. However, Travis Ubuntu 14.04 image only has an old "libosmesa6-dev" package that is not compatible with our `osmesa-sys`. Working around that doesn't appear to be trivial (but it is time consuming), and neither glutin or glium work on Linux CI either :/.
Fortunately, our headed GL context seems to be fine, so we are using that now.


kvark:
bors r=msiglreith

PR #1733, opened 2018-01-10
[RFC] Expose raw, union-based clearing API for portability

msiglreith:
The Vulkan API uses unions for [clear value structs](https://www.khronos.org/registry/vulkan/specs/1.0/man/html/VkClearValue.html), so we don't know which enum variant to choose when implementing the portability layer.

Therefore, I added new `raw` variants for a few clear based commands using union structs for the clear values, which should be memory layout compatible with the vulkan union structs. The raw functions are only exposed by the raw command buffer interface as they don't have much use for normal gfx-rs API users.

`clear_attachments` is excluded as I quite confident that we can infer the enum variants from the associated information when calling the function in the portability layer.

**Suggestions/feedback are welcome!**

~~Only defining the API layer for now, implementation will follow for empty/(gl)/dx12/vulkan (breaking metal again..)~~

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -642,26 +642,25 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         }));
 
         let mut clear_iter = clear_values.into_iter();
-        let attachment_clears = render_pass.attachments.iter().enumerate().map(|(i, attachment)| {
-            AttachmentClear {
-                subpass_id: render_pass.subpasses.iter().position(|sp| sp.is_using(i)),
-                value: if attachment.ops.load == pass::AttachmentLoadOp::Clear {
-                    Some(*clear_iter.next().unwrap().borrow())
-                } else {
-                    None
-                },
-                stencil_value: if attachment.stencil_ops.load == pass::AttachmentLoadOp::Clear {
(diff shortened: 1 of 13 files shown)


kvark:
Wouldn't we need to know those types anyway for D3D12, Metal, and GL? It appears to me as raw-ifying the clear values would only help Vulkan backend.

msiglreith:
> Wouldn't we need to know those types anyway for D3D12, Metal, and GL?

That's correct, we need extract the right union variant from the image/imageview, at least to some extent.

> It appears to me as raw-ifying the clear values would only help Vulkan backend.

It's needed to implement some commands (e.g `vkCmdBeginRenderPass`) in `portability`, without introducing overhead of guessing the information, which should be cheaper in the gfx-rs implementation (not needing to store a IR of the framebuffer again on the portability side).

I can implement it for the dx12 backend to see how much it affects the backend implementation.

msiglreith:
dx12 implementation: https://github.com/gfx-rs/gfx/pull/1733/commits/488d4ca417100e8bd373bb46ea1ce2919f87bd37
Actually comes with no overhead as we already know from the views which variant to choose.

msiglreith:
Should be ready know. Need to check if CI passes, but I'm positive!
Metal renderpass clearing is currently unimplemented.
GL received a bit more love: Handles now subpass color output locations and correctly clearing of the color and depth-stencil attachments on first use.

PR #1741, opened 2018-01-16
Implement new mapping on metal

omni-viral:


src/backend/metal/src/device.rs
```diff
@@ -800,28 +800,64 @@ impl hal::Device<Backend> for Device {
...
     where
         I: IntoIterator,
         I::Item: Borrow<(&'a n::Memory, Range<u64>)>,
     {
-        unimplemented!()
+        for item in iter.into_iter() {
+            let (memory, ref range) = *item.borrow();
+            let mut mapping = memory.mapping.lock().unwrap();
+            assert!(mapping.is_some());
+            let mapping = mapping.as_ref().unwrap();
```
kvark: the assert above is covered by this `unwrap`

omni-viral: Correct.

src/backend/metal/src/native.rs
```diff
@@ -226,7 +232,64 @@ pub enum DescriptorSetBinding {
 }
 
 #[derive(Debug)]
-pub enum Memory {
+pub struct Memory {
+    pub(crate) heap: MemoryHeap,
+    pub(crate) allocations: Arc<Mutex<MemoryAllocations>>,
+    pub(crate) mapping: Mutex<Option<MemoryMapping>>,
```
msiglreith: A mutex is not strictly necessary here I think as `memory` in `vkMapMemory` and `vkUnmapMemory` should be externally synch following Vulkan.

src/backend/metal/src/native.rs
```diff
@@ -226,7 +232,64 @@ pub enum DescriptorSetBinding {
...
+        buffers.extend(self.starts.range(range.clone()).map(|(&start, &(end, ref b))| (start .. end, b.clone())));
+        let range = (Bound::Excluded(range.start), Bound::Included(range.end));
+        buffers.extend(self.ends.range(range).map(|(&end, &(start, ref b))| (start .. end, b.clone())));
+        buffers.sort_unstable_by_key(|&(_, ref b)| b.as_ptr());
+        buffers.dedup_by_key(|&mut (_, ref b)| b.as_ptr());
+        buffers
+    }
+
+    pub(crate) fn insert(&mut self, range: Range<u64>, buffer: metal::Buffer) {
+        self.starts.insert(range.start, (range.end, buffer.clone()));
```
msiglreith: Depending on the cost of cloning that buffer an additional `Vec<metal::Buffer>` might be handy and only storing indices in the trees.

omni-viral: Cloning cost for buffer should be similar to `Arc`

src/backend/metal/src/native.rs
omni-viral: `UnsafeCell` then?

kvark: Actually, this is not a perf concern. We can leave a `//TODO` comment and proceed.

kvark:
Bors r+

kvark:
Bors r+

PR #1744, opened 2018-01-17
dx12: Convert remaining execution models

msiglreith:
These may or may not work with spirv-cross so far but doesn't hurt implementing them all for now.


kvark:
Bors r+

PR #1745, opened 2018-01-18
Add correct image extent (vulkan)

dati91:
Replace the original image extent with the region's extent.

Currently the image extent is ignored when mapping the region. This can cause unwanted memory overwrite.

cc @kvark 
(Also the image parameter is unused for now, i can remove it if you want)

### DIFF
--- a/src/backend/vulkan/src/command.rs
+++ b/src/backend/vulkan/src/command.rs
@@ -28,7 +28,7 @@ fn map_subpass_contents(contents: com::SubpassContents) -> vk::SubpassContents {
 }
 
 fn map_buffer_image_regions<T>(
-    image: &n::Image,
+    _image: &n::Image,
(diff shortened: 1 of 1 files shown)


src/backend/vulkan/src/command.rs
```diff
@@ -28,7 +28,7 @@ fn map_subpass_contents(contents: com::SubpassContents) -> vk::SubpassContents {
 }
 
 fn map_buffer_image_regions<T>(
-    image: &n::Image,
+    _image: &n::Image,
```
omni-viral: You should remove unused argument then.

kvark: Nah, it's fine to have it still, since we'll want to do some checks on the image (e.g. is the range contained, does it support mapping, etc)

PR #1760, opened 2018-01-23
dx12: Fix tier1 resource creation

### LINKED ISSUE #1756: hal: dx12 startup crash
When I try running the quad example using the dx12 backend on my Windows machine with a GTX560 Ti, it crashes during startup.

dx12 debug runtime log attached.

[quad-dx12-rhuagh.LOG](https://github.com/gfx-rs/gfx/files/1653359/quad-dx12-rhuagh.LOG)

msiglreith:
Tier1 devices expose 3 memory type groups (buffer only, image only and target only). We can only create a buffer for mapping in the first group. This PR fixes the exposed memory types and only creates buffers if possible (tier2 or tier1 buffer only and cpu visible).

The remaining issue are exposed image memory types, which should be only device local for tier1 (will be addressed in a followup). Vulkan requires us to expose at least one host-visible memory type for images supporting linear tiling. But we can circumvent this by not exposing any format supporting linear tiling at all for heap tier1 hardware..

Fixes #1756 
cc @Rhuagh @AlphaModder 

NOTE: This PR is written against the generic range type PR.

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -47,6 +47,19 @@ const MAX_VERTEX_BUFFERS: usize = 16;
 
 const NUM_HEAP_PROPERTIES: usize = 3;
 
+// Memory types are grouped according to the supported resources.
+// Grouping is done to circumvent the limitations of heap tier 1 devices.
+// Devices with Tier 1 will expose `BuffersOnl`, `ImageOnly` and `TargetOnly`.
+// Devices with Tier 2 or higher will only expose `Universal`.
+enum MemoryGroup {
+    Universal = 0,
+    BufferOnly,
+    ImageOnly,
+    TargetOnly,
+
+    NumGroups,
+}
+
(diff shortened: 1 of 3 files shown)


src/backend/dx12/src/lib.rs
```diff
@@ -709,8 +709,20 @@ impl hal::Instance for Instance {
...
-                    types.extend(base_memory_types.iter());
+                for i in 0..4 {
+                    types.extend(base_memory_types
+                        .iter()
+                        .map(|mem_type| {
+                            let mut ty = mem_type.clone();
+
+                            // Images and RenderTargets are not host visible as we can't create
+                            // a corresponding buffer for mapping
+                            if i >= 2 {
```
kvark: we should get rid of those magic constants eventually

Rhuagh:
LGTM!

msiglreith:
@Rhuagh Thanks for testing!
@kvark Addressed that magic values, should be cleaner now

just waiting on CI now (:

msiglreith:
@kvark done

PR #1761, opened 2018-01-24
hal: expose more physical device features

msiglreith:
Not the full vulkan list yet. Loosely grouped into `core`, `legacy` (GL related) and `extensions`. Extensions should probably be removed for the sake of actual `GFX_` extensions, but it's currently only vertex instance divisors.


kvark:
Confirmed this to build on Metal, merging.

PR #1571, opened 2017-10-14
Make the dx12 backend ocean ready 🌊 

msiglreith:
* Remove the `WRITE_COMBINE` memory property and adjust it towards the vulkan model
* Implement uniform and storage buffers for dx12 where we dynamically create view objects
* Fix exposed memory heaps and types (query available memory of the system)

PR #1762, opened 2018-01-24
Metal compute support

### LINKED ISSUE #1743: Compute on Metal
Are there any blockers now w.r.t. compute support?

https://github.com/KhronosGroup/SPIRV-Cross/pull/321 has been merged a while ago, plus we got https://github.com/gfx-rs/metal-rs/pull/22 . cc @grovesNL

kvark:
Fixes #1743
Depends on https://github.com/gfx-rs/metal-rs/pull/27
Includes part of #1751
r? @grovesNL

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -110,10 +110,12 @@ struct CommandBufferInner {
     encoder_state: EncoderState,
     viewport: Option<MTLViewport>,
     scissors: Option<MTLScissorRect>,
-    pipeline_state: Option<metal::RenderPipelineState>,
+    render_pso: Option<metal::RenderPipelineState>,
+    compute_pso: Option<metal::ComputePipelineState>,
     primitive_type: MTLPrimitiveType,
     resources_vs: StageResources,
     resources_fs: StageResources,
+    resources_cs: StageResources,
     index_buffer: Option<(metal::Buffer, u64, MTLIndexType)>,
     attribute_buffer_index: usize,
     depth_stencil_state: Option<metal::DepthStencilState>,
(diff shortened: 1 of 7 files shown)


src/backend/metal/src/command.rs
```diff
@@ -125,23 +129,39 @@ impl CommandBufferInner {
 
         self.resources_vs.clear();
         self.resources_fs.clear();
+        self.resources_cs.clear();
+    }
+
+    fn stop(&mut self) {
```
msiglreith: Could this be called `stop_encoding` or something similar?

src/backend/metal/src/device.rs
```diff
@@ -133,19 +112,47 @@ pub struct Device {
...
 impl PhysicalDevice {
+    pub(crate) fn new(raw: metal::Device) -> Self {
+        PhysicalDevice {
+            raw,
+            memory_types: [
+                hal::MemoryType {
+                    properties: Properties::CPU_VISIBLE | Properties::CPU_CACHED,
+                    heap_index: 0,
+                },
+                hal::MemoryType {
```
msiglreith: why is a second memory type here, exposing the same things as the first one?

kvark: this was just copied from the previous code. Perhaps @grovesNL remembers?

grovesNL: It was added in #1687 (https://github.com/gfx-rs/gfx/pull/1687/files#diff-97544b45a9d25f924c3a661a46b1cd0fR63). I guess it's not important unless there is some special meaning for these four indices?

grovesNL: It doesn't seem to be important unless those indices have special meaning elsewhere? These properties became the same in #1571 with the removal of the `memory::WRITE_COMBINED` (https://github.com/gfx-rs/gfx/pull/1571/files#diff-97544b45a9d25f924c3a661a46b1cd0fL120)

kvark:
Requested fixes are applied. Linux and Windows are covered by CI. OSX is covered by compiling locally.
Merge time!

PR #1765, opened 2018-01-26
dx12: Update to latest wio version

msiglreith:
~~~Currently fails due to new nullptr assert.~~~
Should only be merged once a new crates.io version has been published.

Fixes https://github.com/gfx-rs/gfx/issues/1754


kvark:
Doesn't look to be failing :)

msiglreith:
Restructured code a bit to work around that ComPtr assert, just waiting on publish right now.
cc @retep998

kvark:
Let's do it! https://crates.io/crates/wio/0.2.0

PR #1768, opened 2018-01-27
Convert HAL features to bitflags, move legacy out

kvark:
The main reasoning behind using bitflags is not as much an economy of space but the ability to check for intersections and do mathematical set-related operations for Warden. As such, I want to specify which features are affected by a test, so that the framework can skip backends that we know (at run-time) are not capable of running this test. This is way better than current crashing on, say, GL compute.

Requesting early feedback since converting all the backends is considerable amount of work.

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -795,12 +795,12 @@ impl hal::Instance for Instance {
                     max_patch_size: 0,
                     max_viewports: 0,
                     max_compute_group_count: [
-                        d3d12::D3D12_CS_THREAD_GROUP_MAX_X  as _,
-                        d3d12::D3D12_CS_THREAD_GROUP_MAX_Y  as _,
-                        d3d12::D3D12_CS_THREAD_GROUP_MAX_Z  as _,
+                        d3d12::D3D12_CS_THREAD_GROUP_MAX_X,
+                        d3d12::D3D12_CS_THREAD_GROUP_MAX_Y,
+                        d3d12::D3D12_CS_THREAD_GROUP_MAX_Z,
                     ],
                     max_compute_group_size: [
-                        d3d12::D3D12_CS_THREAD_GROUP_MAX_THREADS_PER_GROUP as _,
+                        d3d12::D3D12_CS_THREAD_GROUP_MAX_THREADS_PER_GROUP,
                         1, //TODO
                         1, //TODO
                     ],
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -319,8 +319,8 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
         &mut self,
         submission: hal::queue::RawSubmission<Backend, IC>,
         fence: Option<&native::Fence>,
-    ) where 
-        IC: IntoIterator, 
+    ) where
+        IC: IntoIterator,
         IC::Item: Borrow<command::CommandBuffer>,
     {
         // TODO: semaphores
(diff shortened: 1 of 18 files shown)


kvark:
@msiglreith this is ready for the review now

src/backend/gl/src/device.rs
```diff
@@ -214,11 +213,28 @@ impl Device {
...
+            (4, 10) => glsl::Version::V4_10,
+            (4, 00) => glsl::Version::V4_00,
+            (3, 30) => glsl::Version::V3_30,
+            (1, 50) => glsl::Version::V1_50,
+            (1, 40) => glsl::Version::V1_40,
+            (1, 30) => glsl::Version::V1_30,
+            (1, 20) => glsl::Version::V1_20,
+            (1, 10) => glsl::Version::V1_10,
+            other if other > (4, 60) => glsl::Version::V4_60,
+            other => panic!("GLSL version is note recognized: {:?}", other),
```
msiglreith: 'note'

src/backend/dx12/src/lib.rs
```diff
@@ -780,65 +780,27 @@ impl hal::Instance for Instance {
...
-                    sampler_lod_bias: true,
-                    sampler_border_color: true,
-                },
+                    Features::IMAGE_CUBE_ARRAY |
+                    Features::GEOMETRY_SHADER |
+                    Features::TESSELLATION_SHADER |
+                    //logic_op: false, // Optional on feature level 11_0
+                    Features::MULTI_DRAW_INDIRECT |
+                    Features::FORMAT_BC |
+                    Features::INSTANCE_RATE,
```
msiglreith: 👍

src/backend/gl/src/device.rs
```diff
@@ -435,9 +438,42 @@ impl d::Device<B> for Device {
...
+        let share = &self.share;
+        descs.iter()
+            .map(|desc| {
+                let program = {
+                    let name = unsafe { gl.CreateProgram() };
+
+                    let shader = self.compile_shader(&desc.shader, pso::Stage::Compute);
+                    unsafe { gl.AttachShader(name, shader) };
+
+                    unsafe { gl.LinkProgram(name) };
```
msiglreith: That part should probably be moved into a separate function

src/backend/gl/src/info.rs
```diff
@@ -190,6 +192,41 @@ pub struct Info {
     pub extensions: HashSet<&'static str>,
 }
 
+bitflags! {
+    pub struct LegacyFeatures: u16 {
```
msiglreith: Do you want to keep those in GL? If someone wants the target all backends, they need to have access to these flags. We might have dx11 specific flags also.

src/warden/src/bin/reftest.rs
```diff
@@ -185,5 +225,6 @@ fn main() {
         num_failures += harness.run(instance);
     }
     let _ = harness;
+    num_failures += 0;
```
msiglreith: `+= 0`?

kvark:
Bors r=msiglreith 
I'll follow up with answers

src/warden/src/bin/reftest.rs
kvark: This suppresses rust warning for unused `mut`

src/backend/gl/src/info.rs
kvark: I want to keep it in GL for now, then we'll think about expressing them in an extension to Vk portability, at which point we'll need to move them back into HAL.

kvark:
miscellaneous fixes added
bors r=msiglreith

PR #1777, opened 2018-01-30
Metal update buffer

omni-viral:
PR checklist:
  Wasn't working before also.
  - Metal. They wasn't using this method anyway.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -471,7 +471,7 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
         offset: u64,
         data: &[u8],
     ) {
-        let src = self.device().new_buffer_with_data(data as *const _ as _, data.len() as _, metal::MTLResourceOptions::StorageModePrivate);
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/command.rs
```diff
@@ -461,11 +467,22 @@ impl RawCommandBuffer<Backend> for CommandBuffer {
...
         &mut self,
-        _buffer: &native::Buffer,
-        _offset: u64,
-        _data: &[u8],
+        dst: &native::Buffer,
+        offset: u64,
+        data: &[u8],
     ) {
-        unimplemented!()
+        let src = self.device().new_buffer_with_data(data.as_ptr() as _, data.len() as _, metal::MTLResourceOptions::StorageModePrivate);
```
kvark: when is this buffer freed?

omni-viral: When the command buffer drops reference to it.

kvark: where is it keeping that reference?

PR #1778, opened 2018-01-30
[vk] XCB window support

kvark:
PR checklist:


kvark:
(required to run LunarG samples on Unix)

kvark:
Just didn't want to bring an extra dependency by default, otherwise no reason.
Let's hold off the merging until I actually get those samples running, and we'll see - maybe having it enabled will be needed after all.

kvark:
ok, the samples are linking, although I haven't got lucky to run them yet :/
let's merge this as a step forward regardless
bors r=msiglreith

PR #1781, opened 2018-01-31
Warden GL improvements

kvark:
PR checklist:

The reasoning behind always returning the compute queue type on GL:
  - no need to have all the clients suffer because GL is the only backend that can possibly have that limitation, and it's not 1st tier
  - the capability can still be checked in the `Limits` (compute group size/count), and that's how Warden filters it


kvark:
FIxed now.
bors r=msiglreith

kvark:
Oh wonderful. Ash broke everything.

kvark:
Fixed ash version for now.
bors r=msiglreith

PR #1786, opened 2018-01-31
Re-enable glsl-to-spirv by default in Warden

kvark:
PR checklist:

r? @grovesNL (trying your change, may still fail the CI)


kvark:
Alright this seems to work 🎉 
Next step would be enabling Warden testing on OSX, but I don't want to jinx it here.
bors r=grovesNL

kvark:
huh, weird. I thought it builds on OSX.

kvark:
(still in the queue for some reason?)
bors r-

PR #1792, opened 2018-02-04
Fix flushing and invalidating by helpers

omni-viral:
PR checklist:
  metal, gl

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1792)


kvark:
It can break some of the backends, but:
  1. it looks correct according to the Vulkan frontend logic
  2. this is not yet covered by Warden, so hard to keep regressions in place

bors r+

omni-viral:
@kvark it fixes metal backend though.

PR #1793, opened 2018-02-04
[metal] draw_indexed doesn't consume index buffer

omni-viral:
PR checklist:
   metal backend is broken as well as in master.

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1793)

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -394,6 +394,13 @@ impl CommandBuffer {
         }
     }
 
+    #[inline]
+    fn inner_ref(&self) -> &CommandBufferInner {
+        unsafe {
+            &*self.inner.get()
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


kvark:
Nice, thank you!
bors r+

scoopr:
For whatever reason my draw_indexed calls now segfault, and I pinpointed that to this commit.
Yet, I can't find anything obviously wrong. The segfault goes away when I add a line like `let buffer = buffer.clone();` before the `msg_send!`, but the indexes are garbled. But then it segfaults if it is just `buffer.clone();`, which I though in this case just returns itself? `println!("{:?}", buffer);` looks legit to me. Help!

kvark:
@scoopr thanks for investigation!
@JohnColanduoni could you take a peek please?

omni-viral:
@scoopr Could you provide the code that gets segfault?

omni-viral:
Now I'm getting segfault after `cargo update`.

PR #1795, opened 2018-02-05
dx12 : format properties

msiglreith:
Implementing format properties in dx12, mostly all formats are unsupported for now. Need to gradually implement more over the time..
Another point is to try merging format conversion and format properties in a follow up.

PR checklist:

### DIFF
--- a/src/backend/dx12/src/format.rs
+++ b/src/backend/dx12/src/format.rs
@@ -1,0 +1,1123 @@
+//! Format related queries for the backend.
+
+use hal::format::{BufferFeature, ImageFeature, Properties, NUM_FORMATS};
+
+///
+pub fn query_properties() -> [Properties; NUM_FORMATS] {
+    // TODO
+    let properties = [
+        // Undefined
+        Properties {
+            linear_tiling: ImageFeature::empty(),
+            optimal_tiling: ImageFeature::empty(),
+            buffer_features: BufferFeature::empty(),
+        },
+        // Rg4Unorm
+        Properties {
+            linear_tiling: ImageFeature::empty(),
+            optimal_tiling: ImageFeature::empty(),
+            buffer_features: BufferFeature::empty(),
+        },
+        // Rgba4Unorm
+        Properties {
+            linear_tiling: ImageFeature::empty(),
+            optimal_tiling: ImageFeature::empty(),
+            buffer_features: BufferFeature::empty(),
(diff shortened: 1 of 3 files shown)


src/backend/dx12/src/format.rs
```diff
@@ -0,0 +1,1123 @@
+//! Format related queries for the backend.
+
+use hal::format::{BufferFeature, ImageFeature, Properties, NUM_FORMATS};
+
+///
+pub fn query_properties() -> [Properties; NUM_FORMATS] {
```
kvark: This would be somewhat nice to just load from a RON or something, since it's pure data, rather verbose, and structurally simple. It's blocked by the bitflags support though.

src/backend/dx12/src/format.rs
```diff
@@ -0,0 +1,1123 @@
...
+    // TODO
+    let properties = [
+        // Undefined
+        Properties {
```
kvark: perhaps, we could use `default()` more aggressively, e.g.
```rust
Properties {
    optimal_tiling: ImageFeature::SAMPLED | ImageFeature::BLIT_SRC | ImageFeature::SAMPLED_LINEAR , 
    .. Properties::default()
}
```

Although, on a second thought, this would be harmful when we actually fill all of those up, at which point it will be clearer to see if some helper constructors would work best, so no point doing anything now.

src/hal/src/format.rs
```diff
@@ -215,8 +215,8 @@ macro_rules! surface_types {
 // ident { num_bits, aspects, dim, (color, alpha, ..) }
 surface_types! {
     R4_G4               {   8, COLOR, (1, 1), color: 8 },
-    R4_G4_B4_A4         {  32, COLOR, (1, 1), color: 24, alpha: 4 },
-    B4_G4_R4_A4         {  32, COLOR, (1, 1), color: 24, alpha: 4 },
+    R4_G4_B4_A4         {  16, COLOR, (1, 1), color: 12, alpha: 4 },
```
kvark: uh oh

src/backend/dx12/src/format.rs
msiglreith: to some extent yes, but we also need to manually check for optional format properties on physical device enumeration.

msiglreith:
bors r=kvark

PR #1798, opened 2018-02-07
Use ash reset_command_pool

kvark:
PR checklist:


kvark:
This is harmless
bors r+

PR #1800, opened 2018-02-09
[WIP] Starting to attempt to improve documentation.

icefoxen:
Mostly just some basic cleanup for the *few* bits I actually understand.  This is presented mainly as a place for me to put ongoing work and get comments/discussion.  I realize now that I actually understand very little of gfx-hal, so it's going to be slow going, but I'll do what I can as I figure it out.

My general plan:
 * Start with just filling out API docs where possible
 * Write up a guide explaining HAL structure and how to make the pieces fit together, like my LearningGfx tutorial does for pre-ll
 * Maybe essentially dissect the `gfx_ocean` program in detail as a case study, if I can figure out how the heck it works and I still have energy.  :-)

This isn't going to be super fast, but I'll do what I can!

PR checklist:
- [x] tested examples with the following backends: n/a

### DIFF
--- a/src/hal/src/adapter.rs
+++ b/src/hal/src/adapter.rs
@@ -1,6 +1,10 @@
-//! Physical devices and adapter.
+//! Physical devices and adapters.
 //!
-//! Physical devices are the main entry point for opening a [Device](../struct.Device).
+//! The `PhysicalDevice` trait specifies the API a backend must provide for dealing with
+//! and querying a physical device, such as a particular GPU.  An `Adapter` is a struct
+//! containing a `PhysicalDevice` and metadata for a particular GPU, generally created
+//! from an `Instance` of that backend.  From an `Adapter` you can get a `Device` via
+//! `adapter.open_with(...)`, which lets you specify properties of the `Device` you want.
 
 use {format, memory, Backend, Gpu, Features, Limits};
 use error::DeviceCreationError;
@@ -10,7 +14,7 @@ use queue::{Capability, QueueGroup};
 /// `1.0` (high).
 pub type QueuePriority = f32;
 
-///
+/// A strongly-typed index to a particular `MemoryType`.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct MemoryTypeId(pub usize);
@@ -21,11 +25,12 @@ impl From<usize> for MemoryTypeId {
     }
 }
 
-///
+/// A description for a single chunk of memory in a heap.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct MemoryType {
-    /// Properties of the associated memory.
+    /// Properties of the associated memory, such as  synchronization
+    /// properties or whether it's on the CPU or GPU.
     pub properties: memory::Properties,
     /// Index to the underlying memory heap in `Gpu::memory_heaps`
     pub heap_index: usize,
@@ -42,7 +47,7 @@ pub struct MemoryProperties {
     pub memory_heaps: Vec<u64>,
 }
 
-/// Represents a physical or virtual device, which is capable of running the backend.
+/// Represents a physical device (such as a GPU) capable of supporting the given backend.
 pub trait PhysicalDevice<B: Backend>: Sized {
     /// Create a new logical device.
     ///
@@ -65,21 +70,21 @@ pub trait PhysicalDevice<B: Backend>: Sized {
     /// ```
     fn open(&self, Vec<(B::QueueFamily, Vec<QueuePriority>)>) -> Result<Gpu<B>, DeviceCreationError>;
 
-    ///
+    /// Fetch details for a particular image format.
     fn format_properties(&self, Option<format::Format>) -> format::Properties;
 
-    ///
+    /// Fetch details for the memory regions provided by the device.
     fn memory_properties(&self) -> MemoryProperties;
 
     /// Returns the features of this `Device`. This usually depends on the graphics API being
     /// used.
     fn get_features(&self) -> Features;
 
-    /// Returns the limits of this `Device`.
+    /// Returns the resource limits of this `Device`.
     fn get_limits(&self) -> Limits;
 }
 
-/// Information about a backend adapter.
+/// Metadata about a backend adapter.
 #[derive(Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct AdapterInfo {
@@ -89,7 +94,7 @@ pub struct AdapterInfo {
     pub vendor: usize,
     /// PCI id of the adapter
     pub device: usize,
-    /// The device is based on a software rasterizer
+    /// Whether or not the device is based on a software rasterizer
     pub software_rendering: bool,
 }
 
@@ -103,7 +108,7 @@ pub struct Adapter<B: Backend> {
     pub info: AdapterInfo,
     /// Actual physical device.
     pub physical_device: B::PhysicalDevice,
-    /// Supported queue families information for this adapter.
+    /// Queue families supported by this adapter.
     pub queue_families: Vec<B::QueueFamily>,
 }
 
@@ -121,7 +126,7 @@ impl<B: Backend> Adapter<B> {
     /// # fn main() {
     ///
     /// # let adapter: hal::Adapter<empty::Backend> = return;
-    /// let gpu = adapter.open_with::<_, General>(1, |_| true);
+    /// let device = adapter.open_with::<_, General>(1, |_| true);
     /// # }
     /// ```
     ///
--- a/src/hal/src/adapter.rs
+++ b/src/hal/src/adapter.rs
@@ -3,8 +3,8 @@
 //! The `PhysicalDevice` trait specifies the API a backend must provide for dealing with
 //! and querying a physical device, such as a particular GPU.  An `Adapter` is a struct
 //! containing a `PhysicalDevice` and metadata for a particular GPU, generally created
-//! from an `Instance` of that backend.  From an `Adapter` you can get a `Device` via
-//! `adapter.open_with(...)`, which lets you specify properties of the `Device` you want.
+//! from an `Instance` of that backend.  `adapter.open_with(...)` will return a `Device`
+//! that has the properties specified.
 
 use {format, memory, Backend, Gpu, Features, Limits};
 use error::DeviceCreationError;
--- a/src/hal/src/adapter.rs
+++ b/src/hal/src/adapter.rs
@@ -126,7 +126,7 @@ impl<B: Backend> Adapter<B> {
     /// # fn main() {
     ///
     /// # let adapter: hal::Adapter<empty::Backend> = return;
-    /// let device = adapter.open_with::<_, General>(1, |_| true);
+    /// let (device, queues) = adapter.open_with::<_, General>(1, |_| true).unwrap();
     /// # }
     /// ```
     ///
--- a/src/hal/src/command/mod.rs
+++ b/src/hal/src/command/mod.rs
@@ -1,4 +1,4 @@
-//! DOC TODO
+//! DOC TODO: Need to understand the relationship between this and command pools.
 
 use Backend;
 use queue::capability::Supports;
--- a/src/hal/src/command/mod.rs
+++ b/src/hal/src/command/mod.rs
@@ -45,7 +45,7 @@ impl Level for Primary { }
 /// Vulkan describes a secondary command buffer as one which cannot be directly submitted
 /// to a queue, but can be executed by a primary command buffer.  This allows
 /// multiple secondary command buffers to be constructed which do specific 
-/// things, which can then be composed together into primary command buffers.
+/// things and can then be composed together into primary command buffers.
 pub enum Secondary { }
 impl Level for Secondary { }
 
--- a/src/hal/src/command/mod.rs
+++ b/src/hal/src/command/mod.rs
@@ -1,4 +1,4 @@
-//!
+//! DOC TODO
 
 use Backend;
 use queue::capability::Supports;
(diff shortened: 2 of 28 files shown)


src/hal/src/image.rs
```diff
@@ -246,10 +246,11 @@ pub const CUBE_FACES: [CubeFace; 6] = [
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Kind {
-    /// A single row of texels.
+    /// A single one-dimensional row of texels.
```
omni-viral: Tbh, I've never seen two-dimentional rows.

src/hal/src/format.rs
```diff
@@ -2,7 +2,11 @@
 //! Applicable to textures, views, and vertex buffers.
 
 bitflags!(
-    ///
+    /// Bitflags that describes what properties of a buffer
```
kvark: "describe"

kvark: "buffer" -> "image"

src/hal/src/image.rs
```diff
@@ -368,8 +370,10 @@ impl Usage {
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum WrapMode {
-    /// Tile the texture. That is, sample the coordinate modulo `1.0`. This is
-    /// the default.
+    /// Tile the texture, that is, sample the coordinate modulo `1.0`, so 
+    /// addressing the texture beyond an edge will "wrap" back from the
+    /// other edge.
+    /// This is the default.
```
kvark: I don't think there is a default any more

src/hal/src/image.rs
```diff
@@ -396,7 +400,7 @@ impl Into<f32> for Lod {
     }
 }
 
-/// A wrapper for the 8bpp RGBA color, encoded as u32.
+/// A wrapper for an RGBA color with 8 bits per pixel, encoded as a u32.
```
kvark: in the long run, we should tighten up the use of "pixel" versus "texel" versus "fragment".

src/hal/src/lib.rs
```diff
@@ -304,7 +305,9 @@ pub trait Instance {
     fn enumerate_adapters(&self) -> Vec<Adapter<Self::Backend>>;
 }
 
-/// Different types of a specific API.
+/// The `Backend` trait wraps together all the types needed
+/// to implement a graphics backend.  Each backend, such as OpenGL
```
kvark: the second sentence seems very repetit-ory. Perhaps, something like this would sound better?
> The `Backend` trait wraps together all the types needed to implement a graphics backend, turning into concrete types on OpenGL, Metal, etc.

src/hal/src/lib.rs
```diff
@@ -366,12 +369,13 @@ impl Error for SubmissionError {
 pub type SubmissionResult<T> = Result<T, SubmissionError>;
 
 
-/// Represents a handle to a physical device.
+/// Represents a handle to a physical device, such as a particular
```
kvark: This needs to be fixed. We have a separate type `PhysicalDevice` now, and the `Gpu` struct is not that but rather a combination of the logical device and the hardware queues it provides.

src/hal/src/window.rs
```diff
@@ -154,7 +154,20 @@ pub enum FrameSync<'a, B: Backend> {
     Fence(&'a B::Fence),
 }
 
-/// Allows you to configure a `Swapchain` for creation.
+/// Contains all the data necessary to create a new `SwapChain`:
```
kvark: "SwapChain" -> "Swapchain"

src/hal/src/image.rs
icefoxen: Which is it in this case?  Texel?

src/hal/src/format.rs
icefoxen: But these properties are applicable to vertex buffers as well, as per the module docs?

src/hal/src/image.rs
icefoxen: Yeah but the language used for that section in general is a little squirrelly and I'm trying to make it consistent (maybe unsuccessfully).  We have three "axes" this type can vary on: Dimension (1,2,3), array/not array, and cube/not cube, with dimension only valid for not-cube.

src/hal/src/image.rs
icefoxen: Is there a concrete default then?

src/hal/src/format.rs
kvark: Vertex buffers can't be COLOR/DEPTH/STENCIL, which is what `AspectFlags` is about

src/hal/src/lib.rs
icefoxen: Yeah, it could be phrased better.  This is one of the Important Bits though, so I feel like the repetition is useful to make sure the relationship between the trait, the backend modules, and the backend API's are clear.  I'll rework it.

src/hal/src/lib.rs
icefoxen: Ah, interesting, I need to look at this more.  Thank you.

src/hal/src/image.rs
kvark: AFAIK, this is the distinction:
  - "texel" refers to the actual contents of an image, which is the case here
  - "pixel" refers to the *result* on screen
  - "fragment" refers to the unit of work for the shader, multiple fragments can contribute to the same pixels

src/hal/src/image.rs
kvark: Not really. Neither `WrapMode` or `SamplerInfo` (which the `WrapMode` is a part of) implement `Default`, and the only constructor of the former gets `WrapMode` as an argument.

src/hal/src/command/mod.rs
```diff
@@ -19,22 +19,33 @@ use std::borrow::Borrow;
...
+///
+/// Vulkan describes a primary command buffer as one which can be directly submitted
+/// to a queue, and can execute `Secondary` command buffers.
 pub enum Primary { }
 impl Level for Primary { }
+
 /// Indicates a secondary command buffer.
+///
+/// Vulkan describes a secondary command buffer as one which cannot be directly submitted
+/// to a queue, but can be executed by a primary command buffer.  This allows you to
```
kvark: "This allows you" - I think needs to be rephrased to avoid "you". Also, there is an extra space before it.

src/hal/src/command/mod.rs
```diff
@@ -64,7 +75,7 @@ unsafe impl<'a, B: Backend, C, L: Level> Submittable<'a, B, C, L> for &'a Submit
     unsafe fn as_buffer(self) -> &'a B::CommandBuffer { &self.0 }
 }
 
-/// A convenience for not typing out the full signature of a secondary command buffer.
+/// A convenience type for not typing out the full signature of a secondary command buffer.
```
kvark: maybe `A convenience alias for not typing ...` so that we don't repeat `type`

src/hal/src/command/transfer.rs
```diff
@@ -9,63 +9,68 @@ use queue::capability::{Supports, Transfer};
...
     pub y: i32,
-    ///
+    /// Z offset, in texels.
     pub z: i32,
 }
 
-/// Region of two buffers for copying.
+/// Specifies a source region and a destination
+/// region in a buffer for copying.
+/// DOC TODO: What units for src, dst, size?
```
kvark: bytes for life!

src/hal/src/command/transfer.rs
```diff
@@ -9,63 +9,68 @@ use queue::capability::{Supports, Transfer};
...
+/// Bundles together all the data need to copy data from one `Image`
+/// to another.
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct ImageCopy {
-    ///
+    /// The aspect mask of what to copy: color, depth and/or stencil information.
     pub aspect_mask: format::AspectFlags,
-    ///
+    /// The image subresource to copy from
```
kvark: let's end comments with a dot consistently

src/hal/src/lib.rs
```diff
@@ -306,7 +306,7 @@ pub trait Instance {
 }
 
 /// The `Backend` trait wraps together all the types needed
-/// to implement a graphics backend.  Each backend, such as OpenGL
+/// for a graphics backend.  Each backend module, such as OpenGL
```
kvark: double space found

src/hal/src/pso/graphics.rs
```diff
@@ -6,44 +6,55 @@ use super::input_assembler::{AttributeDesc, InputAssemblerDesc, VertexBufferDesc
 use super::output_merger::{ColorBlendDesc, DepthStencilDesc};
 
 /// A complete set of shaders to build a graphics pipeline.
+///
+/// All except the vertex shader are optional.
+/// DOC TODO: What happens if, say, a fragment shader is not defined?
```
kvark: > If a fragment shader is omitted, the results of fragment processing are undefined. Specifically, any fragment color outputs are considered to have undefined values, and the fragment depth is considered to be unmodified. This can be useful for depth-only rendering.

src/hal/src/pso/graphics.rs
```diff
@@ -6,44 +6,55 @@ use super::input_assembler::{AttributeDesc, InputAssemblerDesc, VertexBufferDesc
...
+    /// be parameters for creating geometry) and produces one or more
+    /// output patches.
     pub hull: Option<EntryPoint<'a, B>>,
-    ///
+    /// A shader that takes in domains produced from a hull shader's output
+    /// patches and computes actual vertex positions.
     pub domain: Option<EntryPoint<'a, B>>,
-    ///
+    /// A shader that takes given input vertexes and outputs one
+    /// or more output vertexes.
```
kvark: can technically output 0 vertices as well, which is used for GPU-based culling

src/hal/src/pso/graphics.rs
```diff
@@ -6,44 +6,55 @@ use super::input_assembler::{AttributeDesc, InputAssemblerDesc, VertexBufferDesc
...
     pub geometry: Option<EntryPoint<'a, B>>,
-    ///
+    /// A shader that outputs a value for a texel.
```
kvark: for a fragment!

src/hal/src/pso/graphics.rs
```diff
@@ -104,15 +121,16 @@ pub enum FrontFace {
     CounterClockwise,
 }
 
-///
+/// A depth bias allows you to specify a hint for how to draw
```
kvark: I'd say "A depth bias allows changing the produced depth values for fragments slightly but consistently, allowing to draw multiple layers of the same geometry without Z-fighting"

src/hal/src/pso/graphics.rs
```diff
@@ -147,23 +165,23 @@ impl Rasterizer {
     };
 }
 
-///
+/// A description of an equation for how to blend transparent, overlapping texels.
```
kvark: fragments :)

src/hal/src/pso/mod.rs
```diff
@@ -206,6 +206,7 @@ pub struct Specialization {
 }
 
 /// Scalar specialization constant with value for overriding.
+/// DOC TODO, don't know what this means.
```
kvark: specialization constants allow for easy configuration of multiple similar pipelines, e.g. you may have a boolean exposed to the shader that switches the specularity on/off, and provide it as a specialization constant. That would make on/off to be separate PSOs but they share most of the internal stuff and are fast to produce. More importantly, they are fast to execute, since the driver can optimize out the branch on that other PSO creation.

src/hal/src/pso/output_merger.rs
```diff
@@ -98,36 +98,39 @@ pub enum BlendOp {
 }
 
 impl BlendOp {
-    ///
+    /// Replaces the destination value with the source.
```
kvark: "Replaces" -> "Replace" for consistency

src/hal/src/queue/submission.rs
```diff
@@ -24,7 +24,8 @@ where
     pub signal_semaphores: &'a [&'a B::Semaphore],
 }
 
-/// Submission information for a command queue.
+/// Submission information for a command queue, generic on a particular
```
kvark: "generic on" -> "generic over"

src/hal/src/queue/submission.rs
```diff
@@ -101,7 +102,8 @@ where
         }
     }
 
-    /// Promote a submission to a higher capability type.
+    /// Promote a submission to a higher capability type, so you can use this to
```
kvark: similarly, we should probably rephrase this to avoid "you"

src/hal/src/lib.rs
icefoxen: Unfortunately I was taught that you always put two spaces between sentences.  Markdown and HTML don't care but it's a hard habit to break.  I'll try to keep it consistent with your style.

src/hal/src/adapter.rs
```diff
@@ -121,7 +126,7 @@ impl<B: Backend> Adapter<B> {
     /// # fn main() {
     ///
     /// # let adapter: hal::Adapter<empty::Backend> = return;
-    /// let gpu = adapter.open_with::<_, General>(1, |_| true);
+    /// let device = adapter.open_with::<_, General>(1, |_| true);
```
kvark: `let (device, queues) = adapter.open_with::<_, General>(1, |_| true).unwrap();`

src/hal/src/command/mod.rs
```diff
@@ -43,9 +43,9 @@ impl Level for Primary { }
 /// Indicates a secondary command buffer.
 ///
 /// Vulkan describes a secondary command buffer as one which cannot be directly submitted
-/// to a queue, but can be executed by a primary command buffer.  This allows you to
-/// construct multiple secondary command buffers that do specific things, and compose
-/// them together into primary command buffers.
+/// to a queue, but can be executed by a primary command buffer.  This allows
+/// multiple secondary command buffers to be constructed which do specific
```
kvark: missing comma before `which`?

src/hal/src/device.rs
```diff
@@ -1,6 +1,15 @@
...
+//! This module exposes the `Device` trait, which provides methods for creating
+//! and managing 
+//!
+//! The `Adapter` and `Device` are very similar to the Vulkan concept of 
+//! "physical devices" vs. "logical devices"; an `Adapter` is single GPU 
+//! (or CPU) that implements a backend, a `Device` is a 
+//! handle to that GPU that fulfills the capabilities you request.
+
+
+//! used for creating and managing graphics resources, and
```
kvark: "Used"

src/hal/src/pso/mod.rs
```diff
@@ -206,21 +215,20 @@ pub struct Specialization {
 }
 
 /// Scalar specialization constant with value for overriding.
-/// DOC TODO, don't know what this means.
 #[derive(Debug, Clone)]
 pub enum Constant {
-    ///
+    /// `bool` value.
```
kvark: heh, might want to just add `missing_docs` here, it's fairly self-descriptive

src/hal/src/queue/mod.rs
```diff
@@ -1,12 +1,11 @@
-/*! Command queues.
-
-    Queues are the execution paths of the graphical processing units. These process
-    submitted commands buffers.
-
-    There are different types of queues, which can only handle associated command buffers.
-    `CommandQueue<B, C>` has the capability defined by `C` for backend `B`: graphics, 
-    compute and transfer.
-!*/
+//! Command queues.
```
kvark: is that a new standard - to use single-line comments for the module docs?

icefoxen: It's just what I've seen everywhere else...

src/hal/src/pso/mod.rs
icefoxen: 👍

kvark:
@icefoxen I assume this needs to go in fast in order to avoid collision with #1837 . Please rebase?

icefoxen:
Rebased but I'm really not sure it improved anything over a merge.  :roll_eyes: Sorry, I really do suck with git.

icefoxen:
Closing in favor of #1862

PR #1801, opened 2018-02-09
[metal] fix segfault

omni-viral:
PR checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1801)


kvark:
so why was it segfaulting?...

omni-viral:
Because the wrong type of arguments

Issue #1805, opened 2018-02-10
gfx-hal quad example segfaults on Android


torkleyy:
## Short info header:

- GFX version: current master (7eeffec9a37af5b8e4a74320eb84cbe9a193bf8f)
- OS: Android 8.1
- GPU: Adreno 418

## Debug output and printlns:

```
[gfx_backend_vulkan][WARN] Unable to find extension: VK_KHR_xlib_surface
[gfx_backend_vulkan][WARN] Unable to find extension: VK_KHR_xcb_surface
[gfx_backend_vulkan][WARN] Unable to find extension: VK_KHR_wayland_surface
[gfx_backend_vulkan][WARN] Unable to find extension: VK_KHR_mir_surface
[gfx_backend_vulkan][WARN] Unable to find extension: VK_KHR_win32_surface
[gfx_backend_vulkan][WARN] Unable to find layer: VK_LAYER_LUNARG_standard_validation
[quad][INFO] AdapterInfo { name: "Adreno (TM) 418", vendor: 20803, device: 67176448, software_rendering: false }
[quad][INFO] Surface format: Rgba8Srgb
[quad][INFO] Created swapchain!
[gfx_backend_vulkan::device][DEBUG] create_descriptor_set_layout [DescriptorSetLayoutBinding { binding: 0, descriptor_type: SampledImage, descriptor_count: 1, stage_flags: ShaderStageFlags(10000), p_immutable_samplers: 0x0 }, DescriptorSetLayoutBinding { binding: 1, descriptor_type: Sampler, descriptor_count: 1, stage_flags: ShaderStageFlags(10000), p_immutable_samplers: 0x0 }]
[quad][INFO] Created descriptor set layout
[gfx_backend_vulkan::device][DEBUG] create_pipeline_layout [0xf0abdf80]
[quad][INFO] Created render pass
[quad][INFO] Before pipeline creation..
[...]

PR #1807, opened 2018-02-10
Add more validation layers to vulkan

torkleyy:
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -44,6 +44,16 @@ mod window;
 const LAYERS: &'static [&'static str] = &[
     #[cfg(debug_assertions)]
     "VK_LAYER_LUNARG_standard_validation",
+    #[cfg(debug_assertions)]
+    "VK_LAYER_LUNARG_core_validation",
+    #[cfg(debug_assertions)]
+    "VK_LAYER_LUNARG_parameter_validation",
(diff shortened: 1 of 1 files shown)


msiglreith:
all the specified layers will be loaded automatically due to `VK_LAYER_LUNARG_standard_validation`

grovesNL:
Maybe it's failing to load the standard validation layer in #1805 because `VK_LAYER_LUNARG_standard_validation` doesn't exist yet in the Android NDK ([I don't see it listed in the official guide](https://developer.android.com/ndk/guides/graphics/validation-layer.html#ilp) but perhaps it just hasn't been updated in a while).

I guess we could specify all of the intended layers in the correct order (from the [Lunarg reference](https://vulkan.lunarg.com/doc/view/1.0.13.0/windows/layers.html)):

```
VK_LAYER_GOOGLE_threading
VK_LAYER_LUNARG_parameter_validation
VK_LAYER_LUNARG_device_limits
VK_LAYER_LUNARG_object_tracker
VK_LAYER_LUNARG_image
VK_LAYER_LUNARG_core_validation
VK_LAYER_LUNARG_swapchain
VK_LAYER_GOOGLE_unique_objects
```

torkleyy:
>  so this change actually enables the layers for you?

No, it doesn't. I didn't know that they were already included.

PR #1811, opened 2018-02-14
Tweak surface presentation support

omni-viral:
Make it possible to query whether queue supports presentation on surface after queues creation

PR checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1811)

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -186,7 +186,7 @@ pub struct PhysicalDevice {
 
 impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     fn open(
-        &self, families: Vec<(QueueFamily, Vec<hal::QueuePriority>)>
+        &self, families: Vec<(&QueueFamily, Vec<hal::QueuePriority>)>
     ) -> Result<hal::Gpu<Backend>, error::DeviceCreationError> {
         let lock = self.is_open.try_lock();
         let mut open_guard = match lock {
@@ -245,7 +245,7 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
 
         let queue_groups = families
             .into_iter()
-            .map(|(family, priorities)| {
+            .map(|(&family, priorities)| {
                 let mut group = hal::backend::RawQueueGroup::new(family);
 
                 let create_idle_event = || unsafe {
--- a/src/backend/empty/src/lib.rs
+++ b/src/backend/empty/src/lib.rs
@@ -56,7 +56,7 @@ impl hal::Backend for Backend {
 pub struct PhysicalDevice;
 impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     fn open(
-        &self, _: Vec<(QueueFamily, Vec<hal::QueuePriority>)>
+        &self, _: Vec<(&QueueFamily, Vec<hal::QueuePriority>)>
     ) -> Result<hal::Gpu<Backend>, error::DeviceCreationError> {
         unimplemented!()
     }
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -182,7 +182,7 @@ impl PhysicalDevice {
 
 impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     fn open(
-        &self, families: Vec<(QueueFamily, Vec<hal::QueuePriority>)>,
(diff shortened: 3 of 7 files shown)


kvark:
I don't understand. `Adapter` has a list of queue families, and you can already check if any of them are supported, even after the device is created if you don't consume `Adapter` with `open_with`.
With this PR, you still need a `PhysicalDevice` to call `supports_queue_family`, so you can't do it after `open_with` anyway, so what does it achieve?

omni-viral:
I don't use `Adapter::open_with` instead I use `PhysicalDevice::open` which doesn't consume `PhysicalDevice`. While `QueueFamily` are consumed during `QueueGroup`s creation so I can't call `supports_queue_family` after that.

kvark:
> While QueueFamily objects are consumed during QueueGroups creation

Where is this happening? I only see it in `open_with`, which you don't use

omni-viral:
`PhysicalDevice::open` receives `QueueFamily` by value.

kvark:
> PhysicalDevice::open receives QueueFamily by value.

Aha, true. Perhaps, it would be easier just to pass it by reference then.
@msiglreith do you have a strong opinion here?

msiglreith:
> Perhaps, it would be easier just to pass it by reference then.
probably this, it's a bit unsafe (passing same queue family multiple times) but shouldn't be much of an issue.

omni-viral:
How do you implement vulkan functions in portability layer?
In vulkan function implementation you have physical device and queue family id.

omni-viral:
> a bit unsafe (passing same queue family multiple times)

Should be totally safe. When you lost a device (not physical device) you have to do this anyway.

omni-viral:
So you want me to make `PhysicalDevice::open` to take `QueueFamily` by reference and revet `Surface::supports_queue_family`?

torkleyy:
Reviewed 12 of 12 files at r1, 4 of 7 files at r4.
Review status: 11 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/hal/src/adapter.rs, line 139 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1811#-L5Q3zUaVo5w50Xws3R-:-L5Q3zUaVo5w50Xws3R0:b80t97c) ([raw file](https://github.com/gfx-rs/gfx/blob/9aa51c3cc7b3b280904e548694241488a1c3a5a2/src/hal/src/adapter.rs#L139)):*
> ```Rust
>         use queue::QueueFamily;
> 
>         let ref physical_device = self.physical_device;
> ```

Unusual syntax, please use the explicit borrow.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1811)*

omni-viral:
Review status: 11 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/hal/src/adapter.rs, line 139 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1811#-L5Q3zUaVo5w50Xws3R-:-L5TeGECxFm3iw59q81u:b-dsqmh5) ([raw file](https://github.com/gfx-rs/gfx/blob/9aa51c3cc7b3b280904e548694241488a1c3a5a2/src/hal/src/adapter.rs#L139)):*
<details><summary><i>Previously, torkleyy (Thomas Schaller) wrote…</i></summary><blockquote>

Unusual syntax, please use the explicit borrow.
</blockquote></details>

Don't see how this binding syntax is less explicit and unusual.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1811)*

torkleyy:
Review status: 3 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/hal/src/adapter.rs, line 139 at r1](https://reviewable.io:443/reviews/gfx-rs/gfx/1811#-L5Q3zUaVo5w50Xws3R-:-L5TgEG8g8mn1zKsebhr:b-cjkyyt) ([raw file](https://github.com/gfx-rs/gfx/blob/9aa51c3cc7b3b280904e548694241488a1c3a5a2/src/hal/src/adapter.rs#L139)):*
<details><summary><i>Previously, omni-viral (Zakarum) wrote…</i></summary><blockquote>

Don't see how this binding syntax is less explicit and unusual.
</blockquote></details>

Bindings are usually used for patterns.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gfx-rs/gfx/1811)*

PR #1817, opened 2018-02-15
Warden temporary disabilities

kvark:
PR checklist:

This is better than trying to bail out on `macos` feature. Hopefully, the disability struct will not grow ;)


kvark:
this is fairly minor
bors r+

PR #1820, opened 2018-02-16
Make exposed objects send+sync.

omni-viral:
Place `Send + Sync` bound on everything in `Backend`.
Add `single-threaded` feature that removes `Send + Sync` bounds from some objects of `Backend`.
OpenGL backend requires `single-threaded` feature.

Some types from metal backend weren't `Send` or `Sync`.
Either their public methods takes `&mut self` and they don't share internal state.
Or metal docs says that they are internally synchronised.
The only exception is `RawCommandBuffer`. But it already was `Send` making use of it unsafe. Adding `Sync` can't make things to worse.

Let IC checks vulkan and dx12 as I can't compile them.

PR checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1820)

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -184,6 +184,9 @@ pub struct PhysicalDevice {
     is_open: Arc<Mutex<bool>>,
 }
 
+unsafe impl Send for PhysicalDevice { }
+unsafe impl Sync for PhysicalDevice { }
+
 impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     fn open(
         &self, families: Vec<(&QueueFamily, Vec<hal::QueuePriority>)>
@@ -334,7 +337,9 @@ pub struct CommandQueue {
     idle_fence: *mut d3d12::ID3D12Fence,
     idle_event: winnt::HANDLE,
 }
-unsafe impl Send for CommandQueue {} //blocked by ComPtr
+
+unsafe impl Send for CommandQueue {}
+unsafe impl Sync for CommandQueue {}
 
 impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
     unsafe fn submit_raw<IC>(
@@ -580,6 +585,9 @@ impl Drop for Device {
 pub struct Instance {
     pub(crate) factory: ComPtr<dxgi1_4::IDXGIFactory4>,
 }
+
+unsafe impl Send for Instance { }
+unsafe impl Sync for Instance { }
 
 impl Instance {
     pub fn create(_: &str, _: u32) -> Instance {
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -15,7 +15,10 @@ extern crate spirv_cross;
 pub extern crate glutin;
 
 use std::cell::Cell;
-use std::rc::Rc;
+use std::fmt;
+use std::sync::Arc;
+use std::ops::Deref;
+use std::thread::{self, ThreadId};
 
 use hal::error;
 use hal::queue::{Queues, QueueFamilyId};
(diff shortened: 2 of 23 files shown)


src/backend/dx12/Cargo.toml
```diff
@@ -13,6 +13,7 @@ workspace = "../../.."
 
 [features]
 default = ["winit"]
+single-threaded = ["gfx-hal/single-threaded"]
```
kvark: why do `dx12`/`empty`/other non-GL backends need to expose the feature?

src/hal/src/lib.rs
```diff
@@ -306,6 +306,7 @@ pub trait Instance {
 
 /// Different types of a specific API.
 #[allow(missing_docs)]
+#[cfg(feature = "single-threaded")]
```
kvark: It would be great to avoid duplication of this mega-trait.
I wonder if we could have
```rust
trait Threaded: Send + Sync {
}
trait PhysicalDevice<B>: Threaded {
...
}
```
and such, so that the only thing we need to replace for `single-threaded` is this `Threaded` trait

src/backend/dx12/Cargo.toml
omni-viral: Although uncommon but you may want to have all backends available (gl included). So you would need `gfx-hal` with `single-threaded` feature enabled. This will require all backends to use `single-threaded` feature.

kvark: Enabling a feature for a base crate (gfx-hal in this case) doesn't force any dependencies to specify that. Feature is supposed to be additive, so cargo assumes everything will still compile (otherwise your `single-threaded` feature in the backends wouldn't be helping anyway).

kvark:
@omni-viral ping

src/backend/dx12/src/lib.rs
```diff
@@ -347,6 +352,7 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
     {
         // Reset idle fence and event
         // That's safe here due to exclusive access to the queue
+        // Nope.
```
kvark: wut

src/backend/dx12/Cargo.toml
kvark: ... still there

src/backend/gl/Cargo.toml
```diff
@@ -21,7 +21,7 @@ default = ["glutin"]
 bitflags = "1"
 log = "0.4"
 gfx_gl = "0.3.1"
-gfx-hal = { path = "../../hal", version = "0.1" }
+gfx-hal = { path = "../../hal", version = "0.1", features = ["single-threaded"] }
```
kvark: feature still there

src/hal/src/lib.rs
kvark: ping again

src/backend/dx12/src/lib.rs
omni-viral: Uh. That was mark for me to notify someone that the statement above is incorrect.
It is not safe due to exclusive access to the queue because there is a clone in `Device`.
But later you told me that waiting and submitting must be externally synchronized. So it is indeed safe here 😄

src/backend/gl/src/lib.rs
```diff
@@ -127,7 +130,51 @@ impl Share {
...
+    }
+}
+
+unsafe impl<T: ?Sized> Send for Starc<T> {}
+unsafe impl<T: ?Sized> Sync for Starc<T> {}
+
+impl<T: ?Sized> Deref for Starc<T> {
+    type Target = T;
+    fn deref(&self) -> &T {
+        assert!(thread::current().id() == self.thread);
```
kvark: let's use `assert_eq!`

src/backend/gl/src/lib.rs
```diff
@@ -127,7 +130,51 @@ impl Share {
     }
 }
 
-pub struct PhysicalDevice(Rc<Share>);
+/// Single-threaded `Arc`.
+/// Wrapper for `Arc` that allows you to `Send` it even if `T: !Sync`.
+/// Yet internal data cannot be accessed in the thread where it was created.
+pub struct Starc<T: ?Sized> {
```
kvark: nice concept!

src/hal/src/adapter.rs
```diff
@@ -43,7 +45,7 @@ pub struct MemoryProperties {
 }
 
 /// Represents a physical or virtual device, which is capable of running the backend.
-pub trait PhysicalDevice<B: Backend>: Sized {
+pub trait PhysicalDevice<B: Backend>: Any + Send + Sync {
```
kvark: why do we need `Any` here?

omni-viral: Just for the case the user would want to handle `PhysicalDevice` from different backends differently. It is more convenient to cast `PhysicalDevice` instead of matching typeid of `Backend`.

src/backend/gl/src/lib.rs
omni-viral: Yeah. Except it should be `outside of the thread where it was created` 😄

src/backend/gl/src/lib.rs
omni-viral: :+1:

src/backend/gl/src/lib.rs
omni-viral: Except it should be
`Yet internal data cannot be accessed outside of the thread where it was created.`

src/hal/src/device.rs
```diff
@@ -118,7 +119,7 @@ pub struct FramebufferError;
 /// are not enforced at the HAL level due to OpenGL constraint (to be revised). Users can still
 /// benefit from the backends that support synchronization of the `Device`.
 ///
-pub trait Device<B: Backend> {
+pub trait Device<B: Backend>: Any + Send + Sync {
```
kvark: let's not have `Any` at least here

src/hal/src/lib.rs
```diff
@@ -322,7 +322,7 @@ pub trait Backend: 'static + Sized + Eq + Clone + Hash + Debug + Any {
     type RenderPass:          Debug + Any + Send + Sync;
```
kvark: wow, I didn't realize we have `Any` bounds. I don't think we need them.

src/hal/src/adapter.rs
kvark: I'm fine with physical devides having `Any` as they are somewhat top level (just above the instance) and useful without knowledge of specific backend. But I don't think `Any` helps us on other types down the line...

omni-viral:
I think we do. Until all backends work the same way.

src/hal/src/lib.rs
omni-viral: I think we do until all backends work the same way,

kvark: What do you mean? I don't think backends require `Any` to be operational. Searching for `use std::any` also shows up results from HAL, really.

omni-viral: It can be used by user of hal.

kvark:
Ok, let's leave `Any` out of scope for this PR and proceed.
bors r+

PR #1821, opened 2018-02-16
Descriptor set updates and copies

kvark:
PR checklist:

This PR avoids the need to allocate vectors just for descriptor updates. It also introduces a separate `Device` entry point for descriptor copies, implemented on Vulkan only at the moment.

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -431,18 +431,18 @@ fn main() {
         )
     );
 
-    device.update_descriptor_sets::<_,Range<_>>(&[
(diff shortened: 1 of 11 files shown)


examples/hal/quad/main.rs
```diff
@@ -431,18 +431,18 @@ fn main() {
         )
     );
 
-    device.update_descriptor_sets::<_,Range<_>>(&[
+    device.write_descriptor_sets::<_, Range<_>>(vec![
```
omni-viral: Don't need `vec!` here.

PR #1837, opened 2018-02-20
[WIP] Documentation improvement (ongoing)

adrianperreault:
Inspired by @icefoxen, this PR is an ongoing effort to improve documentation.

Ground to cover:
- [ ] command
    - [ ] mod
    - [ ] compute
    - [ ] graphics
    - [ ] raw
    - [ ] render_pass
    - [ ] transfer
- [ ] pso
    - [ ] mod
    - [ ] compute
    - [ ] descriptor
    - [ ] graphics
    - [ ] input_assembler
    - [ ] output_merger
- [ ] queue
    - [ ] mod
    - [ ] capability
    - [ ] family
    - [ ] submission
- [ ] adapter
- [ ] backend
- [ ] buffer
- [ ] device
- [ ] error
- [ ] format
- [ ] image
- [ ] lib
- [ ] mapping
- [ ] memory
- [ ] pass
- [ ] pool
- [ ] query
- [ ] queue
- [ ] range
- [ ] window

### DIFF
--- a/src/hal/src/image.rs
+++ b/src/hal/src/image.rs
@@ -511,27 +511,32 @@ impl From<RenderDesc> for DepthStencilDesc {
     }
 }
 
-///
+/// ImageLayouts specify how image subresource data is laid out in memory. Optimal performance
+/// may be obtained by choosing the most appropriate memory layout for any given image subresource.
+/// Images may have image subresources with differing layouts as long as depth and stencil aspects
+/// of a given image resource are in the same layout.
 #[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum ImageLayout {
-    ///
+    /// Supports all types of device access.
     General,
-    ///
+    /// Only valid for image subresources with COLOR_ATTACHMENT usage flag.
     ColorAttachmentOptimal,
-    ///
+    /// Only valid for image subresources with DEPTH_STENCIL_ATTACHMENT usage flag.
     DepthStencilAttachmentOptimal,
-    ///
+    /// Only valid for image subresources with DEPTH_STENCIL_ATTACHMENT usage flag.
     DepthStencilReadOnlyOptimal,
-    ///
+    /// Only valid for image subresources with SAMPLED or INPUT_ATTACHMENT usage flags.
     ShaderReadOnlyOptimal,
-    ///
+    /// Only valid for image subresources with TRANSFER usage flag.
     TransferSrcOptimal,
-    ///
+    /// Only valid for image subresources with TRANSFER usage flag.
     TransferDstOptimal,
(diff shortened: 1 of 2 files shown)


src/hal/src/image.rs
```diff
@@ -589,7 +589,14 @@ pub struct SubresourceLayers {
     pub layers: Range<Layer>,
 }
 
-/// A subset of resources contained within an image.
+/// When creating an image view, used for describing the view's purpose and which part
```
msiglreith: I'm not sure if this is the appropiate documentation for `SubresourceRange`. This rather fits for `create_image_view`.

src/hal/src/image.rs
```diff
@@ -589,7 +589,14 @@ pub struct SubresourceLayers {
...
+/// of the image should be accessed. Examples of purpose would be if the view should be
+/// used as a color, depth, or stencil target (or any combination). A range of mipmap levels
```
msiglreith: Yes, there is an aspect flag but it's not really indicating the usage purpose but rather with 'plane' of the image will be part of the subresource.

src/hal/src/pso/descriptor.rs
```diff
@@ -1,4 +1,16 @@
 //! Descriptor sets and layouts.
+//!
+//! Resource descriptors are used by shaders for accessing resources like buffers
+//! and images. The use of a descriptor has three stages:
+//!
+//! - Descriptor layout is specified during pipeline creation
+//! - Descriptor set is allocated from a descriptor pool
+//! - Descriptor set is bound during rendering
```
msiglreith: Descriptor sets need also be updated (copy or write)

src/hal/src/pso/descriptor.rs
```diff
@@ -1,4 +1,16 @@
...
+//!
+//! A Descriptor Set Layout describes the types of resources to be used by a pipeline.
+//!
+//! A Descriptor Set contains the actual resources to be bound/used: eg. buffer, images
```
msiglreith: Descriptor sets are actually views into resource, so an indirection. Basically a pointer to a memory slice and how it's supposed to be accessed by the shader.

src/hal/src/image.rs
adrianperreault: Ok! The wording "purpose" came from the [Vulkan tutorial](https://vulkan-tutorial.com/Drawing_a_triangle/Presentation/Image_views). So the aspect flag is for 'filtering' which aspect of the image will be a part of the subresource?

src/hal/src/image.rs
adrianperreault: Are SubresourceRanges used for anything other than describing details when creating image views?

adrianperreault:
@icefoxen Sounds good! ☺️

kvark:
Ping! Is this PR abandoned?

PR #1838, opened 2018-02-20
[ll] [WIP] Implement blitting for Metal

mjadczak:
This is a WIP. Currently only same source and destination size are supported. This is still untested so I'm not sure if it works properly.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1084,14 +1084,14 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             debug_assert!(is_offset_positive(&dst_start));
 
             let src_origin = MTLOrigin {
-                x: src_start.x as u64,
-                y: src_start.y as u64,
-                z: src_start.z as u64
+                x: src_start.x as _,
+                y: src_start.y as _,
+                z: src_start.z as _
             };
             let dst_origin = MTLOrigin {
-                x: dst_start.x as u64,
-                y: dst_start.y as u64,
-                z: dst_start.z as u64
+                x: dst_start.x as _,
+                y: dst_start.y as _,
+                z: dst_start.z as _
             };
             let (sx, sy, sz) = range_size(&region.dst_bounds);
             let src_size = MTLSize {
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1065,7 +1065,6 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             debug_assert!(is_offset_positive(&src_start));
             debug_assert!(is_offset_positive(&dst_start));
 
-
             let src_origin = MTLOrigin {
                 x: src_start.x as u64,
                 y: src_start.y as u64,
@@ -1076,14 +1075,12 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                 y: dst_start.y as u64,
                 z: dst_start.z as u64
             };
-
             let (sx, sy, sz) = range_size(&region.dst_bounds);
             let src_size = MTLSize {
                 width: sx,
                 height: sy,
                 depth: sz
             };
-
 
             debug_assert_eq!(region.src_subresource.layers.len(), region.dst_subresource.layers.len());
             //TODO aspect flags?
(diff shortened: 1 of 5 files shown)


src/backend/metal/src/command.rs
```diff
@@ -534,17 +535,101 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+            let src_size = MTLSize {
+                width: sx,
+                height: sy,
+                depth: sz
+            };
+
+
+            debug_assert_eq!(region.src_subresource.layers.len(), region.dst_subresource.layers.len());
+            //TODO aspect flags?
+            for (src_layer, dst_layer) in region.src_subresource.layers.clone().zip(region.dst_subresource.layers.clone()) {
```
kvark: let's have the `region` argument by move in this closure so that we can avoid `clone()` of the ranges

src/backend/metal/src/command.rs
```diff
@@ -534,17 +535,101 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                z: dst_start.z as u64
+            };
+
+            let (sx, sy, sz) = range_size(&region.dst_bounds);
+            let src_size = MTLSize {
+                width: sx,
+                height: sy,
+                depth: sz
+            };
+
```
kvark: really generous on empty lines here, this doesn't match the other code

src/backend/metal/src/command.rs
kvark: let's also assert on unsupported aspect flags and levels

mjadczak: So I haven't looked at how other code handles this, but in Metal as far as I know there are no aspect flags, supported or not—if you want multiple aspects of the same image you have to keep multiple `MTLTexture`s around, and externally keep track of which one has which aspect. If we do emulate Vulkan's multi-aspect images we'll use the aspect flags to do the right thing, right now the aspect flags are meaningless, not necessarily unsupported.

mjadczak: The region argument is borrowed by the whole function, since it takes `IntoIterator<Borrow<com::ImageBlit>>`

src/backend/metal/src/command.rs
mjadczak: Changed it to leave at most one empty line between blocks of code, and no empty lines within related blocks.

src/backend/metal/src/command.rs
mjadczak: As for levels, they are supported and are being used: `sourceLevel: region.src_subresource.level as NSUInteger` on line 607

mjadczak:
@kvark for the shader codepath, do you suggest also writing a GLSL shader and using Spir-V to cross-compile into MSL, or would a dedicated MSL shader be better (since it's internal to the library so we can make it follow any assumptions of the code)?

grovesNL:
IMO it's better just to use MSL for that shader (assuming it will always be static), and perhaps lazily compile/cache it.

kvark:
@mjadczak how is this going? got stuck somewhere?

mjadczak:
Not really, I did have to rewrite parts after command buffer re-use dropped, but mostly just some other stuff came up and got in the way. I pretty much just need to write the shader program for shader blitting, and then figure out how that integrates with the current structure of the command buffer (probably add some render-pass commands as well as blit commands for the regions that don't need a shader). I'll try to find some time to finish this off in the next week or so. As a bonus, this will also implement the `copy_image` command.

mjadczak:
I'm currently essentially trying to wrap my head around the sequence in which the `CommandBufferInner` is "normally" used, so that we can put in an extra render pass without messing with any pipeline / render state the user has set up themselves. There is also the issue of having to at the very least create a sampler which needs to be cleaned up at some point.

mjadczak:
Sorry for lack of progress on this. I've just entered quite a busy time period (Master's thesis due in a month or so) and so have not had the time to properly think through what actually needs to happen in the shader case, and how that interacts with anything else the user might be doing with the encoder to insert their own render passes. It might even be a month or more until I get time to return to this, but I do plan to do it—if, in the meantime, someone else has the motivation to do this or a better idea of how exactly that should work, by all means pick this up.

kvark:
@mjadczak I'm really looking forward to see this shipping. Currently, we have 11K image to image copy tests in the CTS that are not supported. Please keep us up to date on the status of your availability.

kvark:
Closing (due to inactivity) in favor of #1982

PR #1840, opened 2018-02-21
Add missing image usage flags

adrianperreault:
Add INPUT_ATTACHMENT & TRANSIENT_ATTACHMENT image usage flags


adrianperreault:
[Looked in to the equivalent flags for dx12](https://msdn.microsoft.com/en-us/library/windows/desktop/dn986742(v=vs.85).aspx) but nothing stuck out as directly mappable.

PR #1847, opened 2018-02-23
 [vk] careful depth clamping check 

kvark:
Follow-up to #1846

Fixes Vulkan validation warning:
> vkCreateGraphicsPipelines(): the depthClamp device feature is disabled: the depthClampEnable member of the VkPipelineRasterizationStateCreateInfo structure must be set to VK_FALSE. The spec valid usage text states 'If the depth clamping feature is not enabled, depthClampEnable must be VK_FALSE' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkPipelineRasterizationStateCreateInfo-depthClampEnable-00782)

Also enables framebuffer images to be used as clear/transfer/blit destinations.

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -406,7 +406,16 @@ impl d::Device<B> for Device {
                 s_type: vk::StructureType::PipelineRasterizationStateCreateInfo,
                 p_next: ptr::null(),
                 flags: vk::PipelineRasterizationStateCreateFlags::empty(),
-                depth_clamp_enable: if desc.rasterizer.depth_clamping { vk::VK_TRUE } else { vk::VK_FALSE },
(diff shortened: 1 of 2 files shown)


src/backend/vulkan/src/device.rs
```diff
@@ -1434,7 +1443,7 @@ impl d::Device<B> for Device {
                 height: surface.height,
             },
             image_array_layers: 1,
-            image_usage: vk::IMAGE_USAGE_COLOR_ATTACHMENT_BIT,
+            image_usage: vk::IMAGE_USAGE_COLOR_ATTACHMENT_BIT | vk::IMAGE_USAGE_TRANSFER_DST_BIT,
```
msiglreith: Should allow specifying usage flags for the swapchain images in the future

kvark:
bors r=msiglreith

PR #1852, opened 2018-02-26
[metal] Re-usable command buffers

### LINKED ISSUE #1848: Soft command buffers on Metal
#1712 introduced re-usable and secondary command buffers. There is no native support for it on Metal, so we need to record the incoming commands in "software" (without touching Metal API), to be able to record them live at any point of submission.
This would help Warden on OSX big time, as well as unlock more use-cases for HAL on Metal.

kvark:
Fixes #1848
PR checklist:
- [x] `make reftests` succeeds (on Metal, while GL has an unrelated error to be investigated)

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -5,12 +5,12 @@ use std::borrow::{Borrow, BorrowMut};
 use std::cell::UnsafeCell;
 use std::ops::{Deref, Range};
 use std::sync::{Arc};
-use std::mem;
+use std::{iter, mem};
 
-use hal::{error, memory, pool, pso};
+use hal::{command as com, error, memory, pool, pso};
 use hal::{VertexCount, VertexOffset, InstanceCount, IndexCount};
 use hal::buffer::{IndexBufferView};
-use hal::command as com;
+use hal::format::FormatDesc;
 use hal::image::{ImageLayout, SubresourceRange};
 use hal::query::{Query, QueryControl, QueryId};
 use hal::queue::{RawCommandQueue, RawSubmission};
@@ -19,12 +19,8 @@ use metal::{self, MTLViewport, MTLScissorRect, MTLPrimitiveType, MTLClearColor, 
 use cocoa::foundation::NSUInteger;
 use block::{ConcreteBlock};
 use conversions::map_index_type;
+use soft;
 
-
-fn div(a: u32, b: u32) -> u32 {
-    assert_eq!(a % b, 0);
-    a / b
-}
 
 pub struct CommandQueue(pub(crate) Arc<QueueInner>);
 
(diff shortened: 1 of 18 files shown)


kvark:
One of the things that really bothered me in the old Metal backend is the duplication of state-setting calls: on each state, we'd first modify the local state tracker and then check the current encoder, setting the state on it if available. But then, on starting a render/compute pass we'd go through the current state and set it (Vulkan has state inheritance, Metal does not), which used duplicate code, and it was growing pretty messy.

New implementation is straightforward - we compose a list of software commands, and then there are centralized functions that apply them either at the spot or at the beginning on the next relevant pass. Same commands are stored in the re-usable command lists. So, more use cases, but even less code paths :)

src/backend/metal/src/command.rs
```diff
@@ -113,128 +259,386 @@ struct CommandBufferInner {
...
+                drawPrimitives: primitive_type
+                vertexStart: vertices.start as NSUInteger
+                vertexCount: (vertices.end - vertices.start) as NSUInteger
+                instanceCount: (instances.end - instances.start) as NSUInteger
+                baseInstance: instances.start as NSUInteger
+            ];
+        }
+        Cmd::DrawIndexed { ref index, primitive_type, ref indices, base_vertex, ref instances } => {
+            let index_offset = indices.start as pso::BufferOffset * match index.index_type {
+                MTLIndexType::UInt16 => 2,
```
grovesNL: Not a regression, but I guess we should be careful about byte alignment somewhere here ([`indexBufferOffset` must be aligned to 4 bytes](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/1515520-drawindexedprimitives))

src/backend/metal/src/command.rs
```diff
@@ -113,128 +259,386 @@ struct CommandBufferInner {
...
+                        offset,
+                    }
+                })
+            }));
+            commands.extend(resources.textures
+                .iter()
+                .cloned()
+                .enumerate()
+                .filter(|&(_, ref resource)| resource.is_some())
+                .map(|(i, texture)| soft::RenderCommand::BindTexture {
```
grovesNL: Probably could do `filter_map` like the above for these cases (and cloning lazily), although this will change when we reuse storage anyway

src/backend/metal/src/command.rs
```diff
@@ -99,11 +95,161 @@ impl StageResources {
...
+    where
+        I: Iterator<Item = soft::RenderCommand>,
+    {
+        match *self {
+            CommandSink::Immediate { encoder_state: EncoderState::Render(ref encoder), .. } => {
+                for command in commands {
+                    exec_render(encoder, &command);
+                }
+            }
+            CommandSink::Deferred { ref mut passes, is_encoding: true } => {
```
grovesNL: Is there a valid case for `Deferred` where `is_encoding` would be `false`? Just wondering whether this should be an `assert` instead.

kvark:
bors r=msiglreith,grovesNL

Issue #1854, opened 2018-02-27
Swapchain resize support in HAL


kvark:
Vulkan allows these things that we currently don't provide:
  1. destroying swapchains
  2. providing the old swapchain in when creating a new one, which allows not only to re-use resources but also keep presenting the old image

hriundel:
For implementation requirements, are you thinking:
For 1), add `fn destroy(&self) -> ();` to `hal::window::Swapchain` trait and actually implement that functionality for all backends?
For 2), make `hal::window::SwapchainConfig` also accept an optional `Swapchain` and then modify existing `#create_swapchain()` implementations to return `SwapchainConfig`'s `Swapchain` if it exists?

msiglreith:
1. We are creating swapchains from device, therefore I would suggest adding `destroy_swapchain()` to the device trait
2. I would add a new argument to `create_swapchain`, less reference/lifetime hassle inside the `SwapchainConfig` struct imo.

hriundel:
Do you want the actual implementations of `destroy_swapchain()` and optional swapchain creation from old swapchain in `create_swapchain()` to be a part of this issue or should it be separate issues for each backend?

kvark:
@hriundel it's not a super huge thing, so I believe we can track it all here.

mpalomas:
out of curiosity may I know if this task is somehow related to bug [#1780](https://github.com/gfx-rs/gfx/issues/1780) ?

JohnDoneth:
It should be possible to close this issue as well due to resizing support in #2052 being merged. 👍 🎉

PR #1856, opened 2018-02-27
[WIP] Swapchain resize support

hriundel:
Working towards #1854 
- [ ] Add `destroy_swapchain` to hal::Device trait as well as function skeletons for backends that implement this trait
- [ ] Add optional `Swapchain` argument to `create_swapchain()` in hal::Device and update existing implementations with the new signature

- [ ] Implement `destroy_swapchain` for Vulkan
- [ ] Implement `destroy_swapchain` for Metal
- [ ] Implement `destroy_swapchain` for GL
- [ ] Implement `destroy_swapchain` for DX12

PR checklist:
- [x] `make` succeeds (on *nix) (Darwin)
- [ ] `make reftests` succeeds (GL fails on osx)


examples/hal/quad/main.rs
```diff
@@ -141,7 +141,7 @@ fn main() {
     println!("Surface format: {:?}", surface_format);
     let swap_config = SwapchainConfig::new()
         .with_color(surface_format);
-    let (mut swap_chain, backbuffer) = device.create_swapchain(&mut surface, swap_config);
+    let (mut swap_chain, backbuffer) = device.create_swapchain(&mut surface, swap_config, None);
```
hriundel: @msiglreith @kvark I know you said to just add a new argument to `create_swapchain`, but are you sure about that? Because now you have these random `None`s sprinkled around the codebase and it's not immediately clear what those are for?

src/backend/dx12/src/device.rs
```diff
@@ -2374,6 +2375,10 @@ impl d::Device<B> for Device {
         (swapchain, hal::Backbuffer::Images(images))
     }
 
+    fn destroy_swapchain(&self, swapchain: w::Swapchain) -> () {
+        // Just drop
```
hriundel: Mimicked the rest of `destroy_*` methods

src/backend/metal/src/device.rs
```diff
@@ -1403,10 +1403,15 @@ impl hal::Device<Backend> for Device {
         &self,
         surface: &mut Surface,
         config: hal::SwapchainConfig,
+        old_swapchain: Option<Swapchain>,
     ) -> (Swapchain, hal::Backbuffer<Backend>) {
         self.build_swapchain(surface, config)
     }
 
+    fn destroy_swapchain(&self, _swapchain: Swapchain) -> () {
```
hriundel: Looking at MoltenVK's [implementation](https://github.com/KhronosGroup/MoltenVK/blob/93c524d2f05d7a7f983354b8e53cd4e6812708c7/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L1006), they seem to just destroy it so followed their suit?

src/backend/vulkan/src/device.rs
```diff
@@ -1483,6 +1484,13 @@ impl d::Device<B> for Device {
         (swapchain, Backbuffer::Images(images))
     }
 
+    fn destroy_swapchain(&self, swapchain: w::Swapchain) -> () {
```
hriundel: I *hope* this is right, but I'm not sure - hard to test things for backends that you don't have :/ https://github.com/MaikKlein/ash/blob/master/ash/src/extensions/swapchain.rs#L37

examples/hal/quad/main.rs
mjadczak: Could wrap it in an enum.

`pub enum PreviousSwapchain {
  NoSwapchain,
  Swapchain(&Swapchain)
}
`

or something.

src/hal/src/device.rs
```diff
@@ -539,7 +539,8 @@ pub trait Device<B: Backend>: Any + Send + Sync {
     ///
     fn destroy_query_pool(&self, B::QueryPool);
 
-    /// Create a new swapchain from a surface and a queue family.
+    /// Create a new swapchain from a surface and a queue family, or, optionally, recreate it
+    /// from another swapchain.
```
mjadczak: > Create a new swapchain from a surface and a queue family, optionally providing the old swapchain to aid in resource reuse and rendering continuity.

examples/hal/quad/main.rs
kvark: @hriundel I'm fine with `None` there at the end. I agree this is not idea, but fine.

src/backend/dx12/src/device.rs
```diff
@@ -2259,6 +2259,7 @@ impl d::Device<B> for Device {
         &self,
         surface: &mut w::Surface,
         config: hal::SwapchainConfig,
+        old_swapchain: Option<w::Swapchain>,
```
kvark: this gives you a compile warning, please rename to `_old_swapchain` to omit it

src/backend/gl/src/device.rs
```diff
@@ -1060,10 +1060,13 @@ impl d::Device<B> for Device {
         &self,
         surface: &mut Surface,
         config: c::SwapchainConfig,
+        old_swapchain: Option<Swapchain>,
```
kvark: same here, please avoid adding warnings

src/backend/gl/src/device.rs
```diff
@@ -1060,10 +1060,13 @@ impl d::Device<B> for Device {
...
     ) -> (Swapchain, c::Backbuffer<B>) {
         self.create_swapchain_impl(surface, config)
     }
 
+    fn destroy_swapchain(&self, swapchain: Swapchain) -> () { unimplemented!() }
```
kvark: isn't it a no-op?

src/hal/src/device.rs
```diff
@@ -563,15 +564,19 @@ pub trait Device<B: Backend>: Any + Send + Sync {
...
     /// ```
     fn create_swapchain(
         &self,
         surface: &mut B::Surface,
         config: SwapchainConfig,
+        old_swapchain: Option<B::Swapchain>
     ) -> (B::Swapchain, Backbuffer<B>);
 
+    /// Destroys swapchain (e.g. after it's been used)
+    fn destroy_swapchain(&self, swapchain: B::Swapchain) -> ();
```
kvark: `-> ()` at the end is redundant

src/backend/dx12/src/device.rs
```diff
@@ -2374,6 +2375,10 @@ impl d::Device<B> for Device {
         (swapchain, hal::Backbuffer::Images(images))
     }
 
+    fn destroy_swapchain(&self, _swapchain: w::Swapchain) {
+        // Just drop
```
msiglreith: I don't think that's correct in the current implementation.
The swapchain gets drop due to ComPtr, I agree.
But as far as I know, the retrieved backbuffer images need to be also released.

A nice smoketest would probably to try implementing resize in quad.

hriundel: Do you mean modifying method's signature in `hal::Device` to accept both the swapchain and backbuffer, e.g. `fn destroy_swapchain(&self, swapchain: B::Swapchain);` -> `fn destroy_swapchain(&self, swapchain: B::Swapchain, backbuffer: Backbuffer<B>);`?

msiglreith: No, the signature is fine.
Swapchain has associated presentable images (GetBackbufferImage internally). We provide access to these images via the backbuffer. Presentable images are not supposed to be destroyed via `destroy_image` but freed alongside the swapchain on `destroy_swapchain`. In D3D12 we need to keep track of these presentable images inside the swapchain on creation and release them on destruction else they would be dangling. I think the best way would be to just store references inside the swapchain struct.

kvark:
@hriundel what's the status of this?

hriundel:
@kvark sorry I've been limited on time the last ~10 days and likely will continue to be for another 10-14 while I'm working on another project :/. All the `destroy_swapchain` methods are there, I think, but I haven't implemented window resizing. If anyone wants to take a crack at it in the meantime, please go ahead - I wouldn't want to keep blocking this.

JohnDoneth:
I am actively working on continuing @hriundel's work, I managed to fix an unrelated issue involving requested extensions not being checked against available extensions properly. 
(See https://github.com/JohnDoneth/gfx/commit/fe15621c6f54fb8133bf0fed0137abc741ccfd07 for details.)

I had that fix the extension checking in order for KHR_win32_surface to be enabled to be able to finish this Swapchain recreation work.

Now I am in the process of refactoring the quad example to use @hriundel's swapchain destruction. The problem is that a lot of objects depend on the Swapchain and need to be recreated as well. It might take quite some time to refactor it all cleanly.

Lymia:
Status update on this?

JohnDoneth:
@Lymia I've been attempting to work on this when I can. I underestimated how busy I would be following the work I had done and how difficult the refactoring would be.

Unfortunately, the way the example(s) was/are written is in one big function where every 'object' necessary is created up front instead of being separated into say 'create_swapchain', 'create_pipeline' functions which makes it difficult to recreate objects that are invalidated by the swapchain recreation like the renderpass, framebuffers, etc. I've refactored some of the functions but I have yet to finish them all.

I will attempt to make further progress today when I can. However if someone else wants to take a crack, in the meantime, be my guest. I don't want to hold this issue back.

kvark:
@JohnDoneth thanks for the update! is it on your roadmap for this week?

JohnDoneth:
@kvark I'd love to say yes, but at the same time I have to be realistic as I'm at college nearing the end of the semester so things are starting to pile up that I have to get done instead.

HybridEidolon:
So, I tried to update the example myself. The only things I can see that need to be recreated are the swapchain, frame images and framebuffers. I was able to extract that setup code into a separate struct+impl to manage that state. The render pass implicitly depends on the surface format, but that's not expected to change, so I didn't bother. The command buffer is recreated every frame so the framebuffer it uses is automatically up to date, provided the swapchain is replaced _before_ the command buffer is created.

Using code similar to this will immediately segfault the dx12 and vulkan backends on destroy_swapchain:

```rust
device.wait_idle().unwrap();
// destroy framebuffers and frame views
device.destroy_swapchain(swapchain);

let (new_swapchain, backbuffer) = device.create_swapchain(surface, swap_config, None);
```

There is also problem that the `acquire_frame` analog in Vulkan is able to return an error to indicate that the swapchain is no longer usable, but the hal version has no equivalent, so the API for that needs to be changed. Otherwise resizing in Vulkan will panic internally on ErrorVkOutOfDateKhr.

JohnDoneth:
@HybridEidolon do you think making `acquire_frame `and `queue_present_khr` both return a result with an OutOfDate error is the proper way to fix this directly? Handling VK_SUBOPTIMAL_KHR also comes to mind.

If Vulkan is the only API that requires error handling for these two functions it might make for an awkward API but if there's no other solution then that might be the way to go. I don't know enough about DX12 or Metal to know if there's a better solution here.

HybridEidolon:
`acquire_frame` and `queue_present` should both return Results, but the Ok type should _also_ be an enum that may represent both success and VK_SUBOPTIMAL_KHR, since that is also a successful return and doesn't need to be immediately handled. Their Err types can just be the normal runtime error, with an additional variant for OutOfDate, since those functions can return the OOM errors and such in Vulkan too.

Metal's versions aren't fallible, and I think only the present function is fallible in DX12 with no representation for suboptimal configuration.

PR #1858, opened 2018-02-28
API refactor (memory, offsets, wg counts)

kvark:
Improved API consistency.
PR checklist:
- [x] tested examples with the following backends: Vulkan,GL

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -1,8 +1,7 @@
 
-use hal::{command as com, image, memory, pass, pso, query};
-use hal::{IndexCount, IndexType, InstanceCount, VertexCount, VertexOffset};
-use hal::buffer::IndexBufferView;
-use hal::format::AspectFlags;
+use hal::{buffer, command as com, image, memory, pass, pso, query};
+use hal::{IndexCount, IndexType, InstanceCount, VertexCount, VertexOffset, WorkGroupCount};
+use hal::format::Aspects;
 
 use std::{mem, ptr};
 use std::borrow::Borrow;
@@ -696,6 +695,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
     fn pipeline_barrier<'a, T>(
         &mut self,
         _stages: Range<pso::PipelineStage>,
+        _dependencies: memory::Dependencies,
         barriers: T,
     ) where
         T: IntoIterator,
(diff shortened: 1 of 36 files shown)


src/backend/vulkan/src/command.rs
```diff
@@ -224,8 +241,8 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                 self.raw, // commandBuffer
                 conv::map_pipeline_stage(stages.start),
                 conv::map_pipeline_stage(stages.end),
-                vk::DependencyFlags::empty(), // dependencyFlags // TODO
-                &[],
+                mem::transmute(dependencies),
```
grovesNL: Could we use `dependencies.bits()` or does it not exist here?

kvark: It expects a bit-like opaque [structure](https://docs.rs/ash/0.21.1/ash/vk/types/struct.DependencyFlags.html) here that doesn't have `from_bits` (yet?). I'd honestly prefer ash just use the stock `bitflags` crate.

grovesNL: Oh ok, makes sense

src/backend/dx12/src/command.rs
```diff
@@ -705,8 +705,19 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-            match *barrier {
+            match *barrier.borrow() {
+                memory::Barrier::AllBuffers(_) |
+                memory::Barrier::AllImages(_) => {
+                    let mut bar = d3d12::D3D12_RESOURCE_BARRIER {
+                        Type: d3d12::D3D12_RESOURCE_BARRIER_TYPE_UAV,
+                        Flags: d3d12::D3D12_RESOURCE_BARRIER_FLAG_NONE,
+                        u: unsafe { mem::zeroed() },
+                    };
+                    *unsafe { bar.u.UAV_mut() } = d3d12::D3D12_RESOURCE_UAV_BARRIER {
```
msiglreith: So we have confirmation that it's fine with before-after state mismatches?

msiglreith: Could you also add a comment here that UAV barriers provide the necessary sync behaviour?

src/hal/src/buffer.rs
```diff
@@ -5,6 +5,10 @@ use std::fmt;
 
 use {IndexType, Backend};
 
+
+/// An offset inside a vertex buffer, in bytes.
```
msiglreith: Why vertex buffer only?

src/hal/src/memory.rs
```diff
@@ -62,9 +62,24 @@ bitflags!(
...
 );
 
+bitflags!(
+    /// Barrier dependency flags.
+    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+    pub struct Dependencies: u32 {
+        ///
+        const BY_REGION    = 0x1;
+        ///
+        const VIEW_LOCAL   = 0x2;
```
msiglreith: Should we indicate that these are part of extensions?

kvark:
(glutin breaking changes...)
bors r=msiglreith

kvark:
bors r=msiglreith

PR #1860, opened 2018-02-28
New descriptor update API

kvark:
PR checklist:
- [x] tested examples with the following backends: Vk,Metal,Dx12

The old approach to descriptor updates was not very ergonomic, required overhead, and didn't map well to any of the backend implementations (even though conceptually was close to Vulkan).
This PR attempts to revise the API, making it nicer and faster.

Edit: deleted disadvantages section :P

A side effect of this PR is much cleaner implementations on both DX12 and Metal, even supporting write spill-over \o/. Plus, nicer typing and a little bit of comments.


kvark:
Another rough alternative would be to keep all the storages monolitic:
```rust
fn write_descriptor_sets(
  &self,
  buffers: &[(&B::Buffer, Range<u64>)],
  buffer_views: &[&B::BufferView],
  images: &[(&B::Image, ImageLayout)],
  samplers: &[&B::Sampler],
  set_writes: I
) where
  I: IntoIterator,
  I::Item: Borrow<pso::DescriptorSetWrite<'a, B>>;

struct DescriptorWrite {
...
  ty: DescriptorType,
  range: Range<usize>, // to look up `buffers`, `buffer_views`, `images`, etc, based on the descriptor type
}
```

examples/hal/quad/main.rs
```diff
@@ -146,15 +145,13 @@ fn main() {
...
+                bindings: 0 .. 1,
                 ty: pso::DescriptorType::SampledImage,
-                count: 1,
                 stage_flags: ShaderStageFlags::FRAGMENT,
             },
             pso::DescriptorSetLayoutBinding {
-                binding: 1,
+                bindings: 1 .. 2,
                 ty: pso::DescriptorType::Sampler,
-                count: 1,
```
omni-viral: Aren't `binding` index and array `count` orthogonal?

src/hal/src/pso/descriptor.rs
```diff
@@ -95,33 +93,31 @@ pub trait DescriptorPool<B: Backend>: Send + Sync + fmt::Debug {
...
-    StorageBuffer(&'a [(&'a B::Buffer, R)]),
-    UniformTexelBuffer(&'a [&'a B::BufferView]),
-    StorageTexelBuffer(&'a [&'a B::BufferView]),
-    CombinedImageSampler(&'a [(&'a B::Sampler, &'a B::ImageView, ImageLayout)]),
+#[derive(Clone)]
+pub enum Descriptor<'a, B: Backend> {
+    Sampler(&'a B::Sampler),
+    Image(&'a B::ImageView, ImageLayout),
+    CombinedImageSampler(&'a B::ImageView, ImageLayout, &'a B::Sampler),
+    Buffer(&'a B::Buffer, Range<Offset>),
```
omni-viral: `Range<Offset>` can be replaced by `(Option<Offset>, Option<Offset>)`  in which any range-like type can be transformed by some convenient method of `RangeArg` trait.

examples/hal/quad/main.rs
kvark: I might have lost in the bindings, wish a better type level distinction there

src/backend/vulkan/src/native.rs
```diff
@@ -89,33 +91,39 @@ pub struct DescriptorPool {
...
         use std::ptr;
 
-        let layouts = layouts.into_iter().map(|layout| {
-            layout.borrow().raw
-        }).collect::<Vec<_>>();
+        let mut raw_layouts = Vec::new();
+        let mut layout_binginds = Vec::new();
+        for layout in layout_iter {
+            raw_layouts.push(layout.borrow().raw);
+            layout_binginds.push(layout.borrow().bindings.clone());
```
msiglreith: nit: bindings

kvark:
bors r=msiglreith

kvark:
Ok, let's try some more
Bors r=msiglreith

kvark:
Weird that it's still trying to build a merge commit...
bors r-

kvark:
Greeeen!
Bors r=msiglreith

Issue #1404, opened 2017-07-28
Support Min/Mag filter for samplers


msiglreith:
All backends should be able to support this. We only expose one filter method for both, minifaction and magnification. (https://docs.rs/gfx/0.16.1/gfx/texture/struct.SamplerInfo.html)

kvark:
The idea is to have a single enum as opposed to 3 different enums. It doesn't mean that all min/mag/mip produced filters are the same.

Keeping this issue open because now that we are going low level, this abstraction (of having a single enum) can go up to `gfx_render`.

lwiklendt:
For users coming from opengl, a way to get opengl-like nearest mag and bilinear min is to use FilterMethod::Mipmap with a mipmap.

PR #1871, opened 2018-03-08
WIP: Vulkanize the sampler

msiglreith:
Fixes https://github.com/gfx-rs/gfx/issues/1404
PR checklist:

### DIFF
--- a/src/hal/src/image.rs
+++ b/src/hal/src/image.rs
@@ -208,29 +208,29 @@ impl AaMode {
 
 /// How to [filter](https://en.wikipedia.org/wiki/Texture_filtering) the
 /// texture when sampling. They correspond to increasing levels of quality,
-/// but also cost. They "layer" on top of each other: it is not possible to
-/// have bilinear filtering without mipmapping, for example.
-///
-/// These names are somewhat poor, in that "bilinear" is really just doing
-/// linear filtering on each axis, and it is only bilinear in the case of 2D
-/// textures. Similarly for trilinear, it is really Quadralinear(?) for 3D
-/// textures. Alas, these names are simple, and match certain intuitions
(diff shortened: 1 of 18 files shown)


src/hal/src/image.rs
```diff
@@ -208,29 +208,29 @@ impl AaMode {
...
-///
-/// These names are somewhat poor, in that "bilinear" is really just doing
-/// linear filtering on each axis, and it is only bilinear in the case of 2D
-/// textures. Similarly for trilinear, it is really Quadralinear(?) for 3D
-/// textures. Alas, these names are simple, and match certain intuitions
-/// ingrained by many years of public use of inaccurate terminology.
+/// but also cost.
 #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum FilterMethod {
```
kvark: please note that there is a `BlitFilter` type that can now be removed as a duplicate

kvark: also, we might want to shorten it to just `Filter` as well as map the values directly to Vulkan counterparts for FFI

msiglreith: thanks, done

msiglreith:
Vulkan, DX12 and GL should be complete now. Could someone take care of the Metal backend?

kvark:
@msiglreith awesome! if no one else, I'll take care of Metal tonight.

kvark:
Finished all the sampler states in Metal and ported to the new `SamplerInfo`. Interestingly, I don't see setting LOD bias on the sampler descriptor in the API reference, but it seems to pass with no complaints.
Also verified the building/reftesting/examples work on Metal. Time to sail!
bors r+

PR #1876, opened 2018-03-08
[dx12] Full image kind support for view creation

### LINKED ISSUE #1842: Support more texture types in D3D12
Currently, only `Kind::D2` is supported by `view_image_as_shader_resource`. See https://github.com/gfx-rs/gfx/blob/8d0ef5ba96bb4a2ecc5671975677323f11773dc5/src/backend/dx12/src/device.rs#L609
cc @zakorgy

kvark:
Fixes #1842
PR checklist:

Wasn't able to test it on that machine, but nothing suspiciously wrong.

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -489,29 +489,78 @@ impl Device {
         format: dxgiformat::DXGI_FORMAT,
         range: &image::SubresourceRange,
     ) -> Result<d3d12::D3D12_CPU_DESCRIPTOR_HANDLE, image::ViewError> {
-        //TODO: use subresource range
-        let handle = self.dsv_pool.lock().unwrap().alloc_handles(1).cpu;
-
-        if kind.dimensions().3 != image::AaMode::Single {
-            error!("No MSAA supported yet!");
-        }
+        #![allow(non_snake_case)]
+        assert_eq!(range.levels.start + 1, range.levels.end);
 
         let mut desc = d3d12::D3D12_DEPTH_STENCIL_VIEW_DESC {
             Format: format,
-            .. unsafe { mem::zeroed() }
+            ViewDimension: 0,
+            Flags: 0,
+            u: unsafe { mem::zeroed() }
         };
+
+        let MipSlice = range.levels.start as _;
+        let FirstArraySlice = range.layers.start as _;
+        let ArraySize = (range.layers.end - range.layers.start) as _;
 
         match kind {
-            image::Kind::D2(..) => {
-                assert_eq!(range.levels.start + 1, range.levels.end);
+            image::Kind::D1(_) => {
+                desc.ViewDimension = d3d12::D3D12_DSV_DIMENSION_TEXTURE1D;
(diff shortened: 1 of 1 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -526,38 +575,110 @@ impl Device {
...
         };
 
+        let MostDetailedMip = range.levels.start as _;
+        let MipLevels = (range.levels.end - range.levels.start) as _;
+        let FirstArraySlice = range.layers.start as _;
+        let ArraySize = (range.layers.end - range.layers.start) as _;
+
         match kind {
+            image::Kind::D1(_) => {
+                assert_eq!(range.layers, 0..1);
```
msiglreith: Should switch to the array version in the case of non-zero base layer

msiglreith: Same for the other cases.

src/backend/dx12/src/device.rs
```diff
@@ -526,38 +575,110 @@ impl Device {
...
+            }
+            image::Kind::CubeArray(_, num_layers) => {
+                assert!(range.layers.end <= num_layers);
+                assert_eq!(0, ArraySize % 6);
+                desc.ViewDimension = d3d12::D3D12_SRV_DIMENSION_TEXTURECUBEARRAY;
+                *unsafe{ desc.u.TextureCubeArray_mut() } = d3d12::D3D12_TEXCUBE_ARRAY_SRV {
+                    MostDetailedMip,
+                    MipLevels,
+                    First2DArrayFace: FirstArraySlice,
+                    NumCubes: ArraySize / 6, //TODO: verify
```
msiglreith: I'm pretty sure that's correct

src/backend/dx12/src/device.rs
```diff
@@ -572,34 +693,76 @@ impl Device {
...
+            image::Kind::D2Array(..) => {
+                error!("MSAA images can't be viewed as UAV");
+                return Err(image::ViewError::Unsupported);
+            }
+            image::Kind::D3(_, _, _) => {
+                assert_eq!(range.layers, 0..1);
+                desc.ViewDimension = d3d12::D3D12_UAV_DIMENSION_TEXTURE3D;
+                *unsafe{ desc.u.Texture3D_mut() } = d3d12::D3D12_TEX3D_UAV {
+                    MipSlice,
+                    FirstWSlice: FirstArraySlice, //TODO: double-check
```
msiglreith: You already have set the assert correct :D 
FirstWSlice should be 0 and WSize can be -1

src/backend/dx12/src/device.rs
kvark: we can't switch to array version since `D1` kind has only a single layer

src/backend/dx12/src/device.rs
kvark: ok, I'll remove the TODO

src/backend/dx12/src/device.rs
kvark: I see, will fix

kvark:
@msiglreith PTAL

PR #1877, opened 2018-03-08
[WIP, dx12] careful temporary image placement for buffer copies

### LINKED ISSUE #1843: Fix subresource footprint calculation in D3D12
@zakorgy wrote:
> When creating the D3D12_SUBRESOURCE_FOOTPRINT the height variable should take the buffer offset in account. Because if the offset is not zero, the debug layer shows an error, that we want to write more bytes than the buffer size.  A possible solution which worked for us, is to replace this [line](https://github.com/gfx-rs/gfx/blob/8d0ef5ba96bb4a2ecc5671975677323f11773dc5/src/backend/dx12/src/command.rs#L1296) with
```rust
let height = height as u32 - (region.buffer_offset as u32 / row_pitch);
```

kvark:
Fixes #1843
PR checklist:

Basic idea is that we can't always map a virtual image at the buffer offset that the user provided, given that image offset in row/columns may be non-zero. Instead, we roll back the offset as much as needed. This is not ideal, given that it only works if the given buffer offset is big enough in the first place.

Note: I haven't yet tested this.

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -18,6 +18,7 @@ use {conv, native as n, Backend, CmdSignatures, MAX_VERTEX_BUFFERS};
 use root_constants::RootConstant;
 use smallvec::SmallVec;
 
+
 // Fixed size of the root signature.
 // Limited by D3D12.
 const ROOT_SIGNATURE_SIZE: usize = 64;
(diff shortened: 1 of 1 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -1283,61 +1285,82 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
 
+            let (footprint_height, height_offset) = if r.image_extent.depth <= 1 {
+                // when a single depth slice is copied, we can pretend to deal with a smaller image
+                (r.image_extent.height, 0)
+            } else {
+                (height as u32, r.image_offset.y as u32)
+            };
+
+            let src_box = d3d12::D3D12_BOX {
+                left: r.image_offset.x as _,
```
msiglreith: This looks incorrect, should be 0. image offset is specified as parameter on CopyTextureRegion.

src/backend/dx12/src/command.rs
```diff
@@ -1283,61 +1285,82 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                back: r.image_extent.depth as _,
+            };
+
+            // Copy each layer in the region
+            for layer in r.image_layers.layers.clone() {
                 // Advance buffer offset with each layer
                 *unsafe { src.u.PlacedFootprint_mut() } = d3d12::D3D12_PLACED_SUBRESOURCE_FOOTPRINT {
-                    Offset: region.buffer_offset as UINT64 + (layer as u32 * slice_pitch * depth) as UINT64,
+                    Offset: r.buffer_offset as UINT64 -
+                        (width_offset_bytes + height_offset * row_pitch) as UINT64 +
```
msiglreith: Looks suspicious, I don't believe that the resulting value will be aligned correctly

src/backend/dx12/src/command.rs
```diff
@@ -1283,61 +1285,82 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    r.image_extent.width != width as u32 || r.image_extent.height != height as u32 || r.image_extent.depth != depth as u32
+                {
+                    error!("MSAA and depth/stencil copies have to cover the whole subresource");
+                }
+            }
 
-                let row_pitch = div(buffer_width, image.block_dim.0 as _) * image.bytes_per_block as u32;
-                let slice_pitch = div(buffer_height, image.block_dim.1 as _) * row_pitch;
-                assert_eq!(row_pitch % d3d12::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT as u32, 0);
+            assert!(buffer_width >= width as u32);
```
msiglreith: We should definitely remove this assert as it's incorrect. There is no requirement that the buffer need be as 'wide' as the whole image

src/backend/dx12/src/command.rs
```diff
@@ -1283,61 +1285,82 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             Type: d3d12::D3D12_TEXTURE_COPY_TYPE_SUBRESOURCE_INDEX,
             u: unsafe { mem::zeroed() },
         };
-        let (width, height, depth, _) = image.kind.dimensions();
+        let (width, height, depth, aa) = image.kind.dimensions();
```
msiglreith: These should be dimensions of the mipmap

kvark:
@msiglreith thanks for your feedback! Closing in favor of #1877

PR #1878, opened 2018-03-08
[vk] image view type

### LINKED ISSUE #1874: vulkan: ImageView creation ignores ImageType
When creating an `ImageView` for a `2DArray` image (or any other than `2D`), it will ignore the `Array` part and will create a `2D`
 type `ImageView`.
The missing part is [here](https://github.com/gfx-rs/gfx/blob/4995da508d209a87dd48c88607f44b559c2427e7/src/backend/vulkan/src/device.rs#L1011).
(Further reading: https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#resources-image-views-compatibility)

kvark:
Fixes  #1874
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -900,45 +900,52 @@ impl d::Device<B> for Device {
             _ => vk::ImageCreateFlags::empty(),
         };
 
-        let (image_type, extent, array_layers, aa_mode) = match kind {
+        let (image_type, view_type, extent, array_layers, aa_mode) = match kind {
             D1(width) => (
                 vk::ImageType::Type1d,
+                vk::ImageViewType::Type1d,
                 vk::Extent3D { width: width as u32, height: 1, depth: 1 },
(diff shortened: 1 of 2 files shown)


kvark:
@msiglreith heh, right, I didn't think of that. If we are to follow Vulkan 1:1 then the view creation needs an extra argument, and the sad thing is - this argument is going to require a reduced version of our `Kind`, which in turn would trigger a sequence of events leading to destruction of `Kind` at all. May not be bad in the end, just not something I wanted to address right now.

Where is the view type used otherwise? Is, say, a view of `Type2dArray` not compatible with shader expecting a 2D view?

msiglreith:
I'm fine with a bandaid just wanted to be sure.

> Is, say, a view of Type2dArray not compatible with shader expecting a 2D view?

nope, SPIRV has an array flag for its image opcode

kvark:
Created #1879 for real view type support, as discussed here.

PR #1880, opened 2018-03-14
point documentation to https://docs.rs/gfx

kud1ing:
point documentation to https://docs.rs/gfx instead of to https://docs.rs/gfx-hal which says "crate does not exist"


kvark:
Sorry about the confusion! The link is correct, but the documentation is not published yet, so it points nowhere. A better fix would be to temporarily remove the link completely.
See also - https://github.com/gfx-rs/gfx/pull/1767#issuecomment-360883881

kud1ing:
Ah ok, thanks.

PR #1885, opened 2018-03-17
[Vulkan] Surface extension detection failure on newer Rust and 1.1 capable drivers

### LINKED ISSUE #1884: GFX Vulkan Backend does not run in Vulkan 1.1 compatible drivers
Short info header:
- GFX version: from github (latest)
- OS: Windows 10 Insiders (17120)
- GPU: AMD Radeon RX Vega 64 (Vulkan 1.1 driver)

Tried to run examples/hal/quad example with Vulkan feature. 
DX12 feature run successfully. 

```
Warning: Unrecognized CreateInstance->pCreateInfo->pApplicationInfo.apiVersion number -- (0x00000000) assuming VK_API_VERSION_1_1.

thread 'main' panicked at 'Vulkan driver does not support VK_KHR_WIN32_SURFACE', src\backend\vulkan\src\window.rs:195:13
error: process didn't exit successfully: `C:\Users\elvir\rust\gfx\target\debug\quad.exe` (exit code: 101)
The terminal process terminated with exit code: 1

```

elviras9t:
Fixes #1884
PR checklist:

Our fix for extension list check

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -162,8 +162,9 @@ impl Instance {
                 instance_extensions
                     .iter()
                     .find(|inst_ext| unsafe {
-                        CStr::from_ptr(inst_ext.extension_name.as_ptr()) ==
-                            CStr::from_ptr(ext.as_ptr() as *const _)
+                        let inst_ext_name = std::slice::from_raw_parts(inst_ext.extension_name.as_ptr(), ext.len());
+                        let surf_ext_name = std::slice::from_raw_parts(ext.as_ptr() as *const i8, ext.len());
+                        (surf_ext_name == inst_ext_name)
                     })
                     .map(|_| ext)
                     .or_else(|| {
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -1,5 +1,3 @@
-#![feature(libc)]
-
 #[macro_use]
 extern crate log;
 extern crate ash;
@@ -8,7 +6,6 @@ extern crate gfx_hal as hal;
 #[macro_use]
 extern crate lazy_static;
 extern crate smallvec;
-extern crate libc;
 
 #[cfg(windows)]
 extern crate winapi;
@@ -165,8 +162,8 @@ impl Instance {
                 instance_extensions
                     .iter()
                     .find(|inst_ext| unsafe {
-                        let _inst_ext = inst_ext.extension_name.as_ptr();
-                        std::slice::from_raw_parts(_inst_ext, libc::strlen(_inst_ext)) == std::slice::from_raw_parts(ext.as_ptr() as *const i8, ext.len())
+                        CStr::from_ptr(inst_ext.extension_name.as_ptr()) == 
+                        CStr::from_bytes_with_nul_unchecked(std::slice::from_raw_parts(ext.as_ptr(), ext.len()+1))
                     })
                     .map(|_| ext)
                     .or_else(|| {
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -1,3 +1,5 @@
+#![feature(libc)]
+
 #[macro_use]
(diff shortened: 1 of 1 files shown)


msiglreith:
Welcome!
Thanks for addressing this issue, could you add a quick description why the old code doesn't work?

elviras9t:
Naming of extensions joining with another, and not accounting length. Comparing goes wrong.

src/backend/vulkan/src/lib.rs
```diff
@@ -162,8 +162,9 @@ impl Instance {
                 instance_extensions
                     .iter()
                     .find(|inst_ext| unsafe {
-                        CStr::from_ptr(inst_ext.extension_name.as_ptr()) ==
-                            CStr::from_ptr(ext.as_ptr() as *const _)
+                        let inst_ext_name = std::slice::from_raw_parts(inst_ext.extension_name.as_ptr(), ext.len());
```
kvark: what's the guarantee that `inst_ext.extension_name` will have at least `ext.len()` characters?

elviras9t:
I badle know Rust language, so I have only dirty fix with instance length. 
```Rust        
                        // search intance length
                        let mut inst_len = 0usize; let mut acml = 1usize; 
                        inst_ext.extension_name.iter().for_each(|s_|{ if s_.clone()==0i8 {acml = 0usize}; inst_len += acml; });

                        let inst_ext_name = std::slice::from_raw_parts(inst_ext.extension_name.as_ptr(), inst_len);
                        let surf_ext_name = std::slice::from_raw_parts(ext.as_ptr() as *const i8, ext.len());
                        (surf_ext_name == inst_ext_name)
```

vitvakatu:
I can't understand why the old code was wrong and the new one is okay. The definition of `CStr::from_ptr`:
```rust
pub unsafe fn from_ptr<'a>(ptr: *const c_char) -> &'a CStr {
        let len = sys::strlen(ptr);
        let ptr = ptr as *const u8;
        CStr::from_bytes_with_nul_unchecked(slice::from_raw_parts(ptr, len as usize + 1))
    }
```

elviras9t:
Problem was been with `ext.as_ptr() as *const i8, ext.len()`
It striding with next extension name by pointers.

elviras9t:
Also, have alternate version 
```
.find(|inst_ext| unsafe {
                        let _inst_ext = inst_ext.extension_name.as_ptr();
                            CStr::from_bytes_with_nul_unchecked(std::slice::from_raw_parts(_inst_ext as *const u8, libc::strlen(_inst_ext))) == 
                            CStr::from_bytes_with_nul_unchecked(std::slice::from_raw_parts(ext.as_ptr() as *const u8, ext.len()))
                    })
```

elviras9t:
Corrected code:
```Rust
CStr::from_ptr(inst_ext.extension_name.as_ptr()) == 
CStr::from_bytes_with_nul_unchecked(std::slice::from_raw_parts(ext.as_ptr(), ext.len()+1))
```

Old code:
```Rust
CStr::from_ptr(inst_ext.extension_name.as_ptr()) ==
CStr::from_ptr(ext.as_ptr() as *const _)
```

msiglreith:
IMO this should be fixed in ash by null-terminating the string values.
https://github.com/MaikKlein/ash/blob/master/ash/src/vk.rs#L292-L316

Lymia:
This PR is really shady. I submitted what I think is a better fix in #1886. This construction of a `CStr` from `&str` relies on an undocumented implementation detail, that `CStr`s store the length of their input. It'd be better to call `to_str` or `to_bytes` on the CStr, and compare that with the `&str` rather than this hack.

In addition, the same problem exists still about 10 lines down, for extensions.

@msiglreith `ash` should have no reason to expect that other crates would try to use their `&str`s, a non-null terminated string type, as `CStr`s, a null terminated string type, for no good reason.

kvark:
@elviras9t Thank you for bringing our attention to the issue and proposing a fix! We ended up accepting an alternative in #1886, so I'm closing this one.

PR #1889, opened 2018-03-26
[dx12] image kind support for RTV creation

kvark:
Follow-up to #1876
PR checklist:
- [x] tested examples with the following backends: DX12

cc @zakorgy


kvark:
not sure why it got cancelled?
bors r=msiglreith

PR #1890, opened 2018-03-27
Image kinds re-design

### LINKED ISSUE #1879: Image view type
In Vulkan, creating an image view requires a kind of view to be specified. I.e. A 2D array texture can be seen with a 2D view. D3D12 has similar notation. We are currently deriving the view type from the image kind, which has a problem of potentially mismatching shader expectations. For instance, we are not able to use different-kind images to provide views for the same descriptor.

Solution: introduce a `ViewKind`/`ViewType` enum (matching Vulkan's 1:1 preferably) that is passed as an extra parameter to `create_image_view`.

kvark:
Fixes #1879 

~~Early WIP opened for feedback on the new HAL image API bits.~~ Ready for review now.

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -143,7 +143,7 @@ impl Instance {
             application_version: version,
             p_engine_name: b"gfx-rs\0".as_ptr() as *const _,
             engine_version: 1,
-            api_version: 0, //TODO
+            api_version: 0, //TODO: VK_API_VERSION_1_1
         };
 
         let instance_extensions = entry
--- a/src/hal/src/image.rs
+++ b/src/hal/src/image.rs
@@ -45,6 +45,11 @@ pub struct Offset {
     pub z: i32,
 }
 
+impl Offset {
+    /// Zero offset shortcut
+    pub const ZERO: Self = Offset { x: 0, y: 0, z: 0 };
+}
+
 /// Pure image object creation error.
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum CreationError {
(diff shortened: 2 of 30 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -143,7 +143,7 @@ impl Instance {
             application_version: version,
             p_engine_name: b"gfx-rs\0".as_ptr() as *const _,
             engine_version: 1,
-            api_version: 0, //TODO
+            api_version: 0, //TODO: VK_API_VERSION_1_1
```
msiglreith: heh

msiglreith: Actually, 0 is fine

src/hal/src/image.rs
```diff
@@ -254,98 +225,133 @@ pub const CUBE_FACES: [CubeFace; 6] = [
...
-    Cube(Size),
-    /// An array of Cube textures.
-    CubeArray(Size, Layer),
 }
 
 impl Kind {
-    /// Get texture dimensions
-    pub fn dimensions(&self) -> Dimensions {
-        let s0 = AaMode::Single;
+    /// Get the image extent.
```
msiglreith: Nit: annotation that these are the extent of a single image layer.

src/backend/vulkan/src/lib.rs
kvark: it spams a warning to console now (unrelated to the PR): 
> Warning: Unrecognized CreateInstance->pCreateInfo->pApplicationInfo.apiVersion number -- (0x00000000) assuming VK_API_VERSION_1_1.

msiglreith: Sounds like a validation layer bug

kvark:
bors r=msiglreith

PR #1891, opened 2018-03-28
Change Subresource and ImageCopy to match Vulkan API

mjadczak:
As discussed on Gitter, this changes the `Subresource` and `ImageCopy` structs to match the Vulkan API.

`Subresource` is `Copy` because it's just three ints, and because the previous `Subresource` type (a tuple of two ints), was also copy and some existing code assumes this.

I had to change the DirectX and Vulkan code blind as I can't compile them, so as well as the CI, perhaps @kvark and/or @msiglreith can take a look at that especially.

PR checklist:
- [ ] `make reftests` succeeds (no, but Metal reftests didn't work before anyway)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ help:
 	@echo "Supported backends: gl $(FEATURES_HAL) $(FEATURES_HAL2)"
 
 check:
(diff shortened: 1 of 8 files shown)


kvark:
Let's have CI passing and @msiglreith peeking at it.

PR #1893, opened 2018-03-28
[dx12] aligned buffer-image copies

### LINKED ISSUE #1843: Fix subresource footprint calculation in D3D12

kvark:
Replaces  #1877
Fixes #1843

PR checklist:

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -1244,31 +1244,31 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         };
 
         for region in regions {
-            let region = region.borrow();
-            debug_assert_eq!(region.src_subresource.layers.len(), region.dst_subresource.layers.len());
-            let num_layers = region.src_subresource.layers.len() as image::Layer;
-            let src_layer_start = region.src_subresource.layers.start;
-            let dst_layer_start = region.dst_subresource.layers.start;
+            let r = region.borrow();
+            debug_assert_eq!(r.src_subresource.layers.len(), r.dst_subresource.layers.len());
+            let num_layers = r.src_subresource.layers.len() as image::Layer;
+            let src_layer_start = r.src_subresource.layers.start;
+            let dst_layer_start = r.dst_subresource.layers.start;
+            let src_box = d3d12::D3D12_BOX {
+                left: r.src_offset.x as _,
+                top: r.src_offset.y as _,
+                right: (r.src_offset.x + r.extent.width as i32) as _,
+                bottom: (r.src_offset.y + r.extent.height as i32) as _,
+                front: r.src_offset.z as _,
+                back: (r.src_offset.z + r.extent.depth as i32) as _,
+            };
+
             for layer in 0..num_layers {
                 *unsafe { src_image.u.SubresourceIndex_mut() } =
-                    src.calc_subresource(region.src_subresource.level as _, (src_layer_start + layer) as _, 0);
+                    src.calc_subresource(r.src_subresource.level as _, (src_layer_start + layer) as _, 0);
                 *unsafe { dst_image.u.SubresourceIndex_mut() } =
-                    dst.calc_subresource(region.dst_subresource.level as _, (dst_layer_start + layer) as _, 0);
-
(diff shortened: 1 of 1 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -602,6 +616,148 @@ impl CommandBuffer {
...
+                            .. buf_offset
+                        },
+                        copy_extent: image::Extent {
+                            width: r.image_extent.width - half,
+                            .. r.image_extent
+                        },
+                    });
+                }
+            } else {
+                // worst case: row by row copy
```
msiglreith: Could you explain the idea here?
Imo the worst case scenario here is a image laid out in memory like this, considering row pitch alignment and data alignment:
```
-----------
----------x
xxxxxxxxxoo
ooooooooyyy
yyyyyyyzzzz
zzzz ...
```
The resulting image should be
```
xxxxxxxxxx
oooooooooo
yyyyyyyyyy
zzzzzzzzzz
...
```

src/backend/dx12/src/command.rs
```diff
@@ -602,6 +616,148 @@ impl CommandBuffer {
...
+                for z in 0 .. r.image_extent.depth {
+                    for y in 0 .. r.image_extent.height {
+                        let row_offset = layer_offset + z as u64 * slice_pitch as u64 + y as u64 * row_pitch as u64;
+                        let footprint_offset = row_offset & !(d3d12::D3D12_TEXTURE_DATA_PLACEMENT_ALIGNMENT as u64 - 1);
+                        let gap = (row_offset - footprint_offset) as u32;
+                        copies.push(Copy {
+                            footprint_offset,
+                            footprint: image::Extent {
+                                width: r.image_extent.width + gap,
+                                height: 1,
```
msiglreith: Should check block dimension (1, 1) here

kvark:
@msiglreith thanks for spotting the problems! should be addressed now, PTAL

src/backend/dx12/src/command.rs
```diff
@@ -602,6 +616,182 @@ impl CommandBuffer {
...
+            r.buffer_height
+        };
+        let row_pitch = div(buffer_width, image.block_dim.0 as _) * image.bytes_per_block as u32;
+        let slice_pitch = div(buffer_height, image.block_dim.1 as _) * row_pitch;
+        let is_pitch_aligned = row_pitch % d3d12::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT == 0;
+
+        for layer in r.image_layers.layers.clone() {
+            let img_subresource = image
+                .calc_subresource(r.image_layers.level as _, layer as _, 0);
+            let layer_offset = r.buffer_offset as u64 + (layer as u32 * slice_pitch * r.image_extent.depth) as u64;
```
msiglreith: `* r.image_extent.depth` probably not needed

src/backend/dx12/src/command.rs
```diff
@@ -602,6 +616,182 @@ impl CommandBuffer {
...
+                        });
+
+                        // and if it crosses a pitch alignment - we copy the rest separately
+                        if cut_width == r.image_extent.width {
+                            continue;
+                        }
+                        let leftover = r.image_extent.width - cut_width;
+
+                        copies.push(Copy {
+                            footprint_offset: next_aligned_offset,
```
msiglreith: This looks somewhat suspicious to me as `D3D12_TEXTURE_DATA_PLACEMENT_ALIGNMENT` is twice as large as the required row pitch.

kvark:
Noted. Thanks for the review!
Bors r=msiglreith

PR #1897, opened 2018-03-29
Baked pipeline states

### LINKED ISSUE #1736: Static viewports in graphics pipeline
Allow to specify the viewports/scissors statically on graphics pipeline creation. We enforce dynamic viewports for the users. The other backends can easily emulate this by adding the corresponding calls on pipeline binding.

kvark:
Fixes #1736

PR checklist:

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -456,8 +456,8 @@ fn main() {
     ]);
 
     // Rendering setup
-    let viewport = command::Viewport {
-        rect: command::Rect {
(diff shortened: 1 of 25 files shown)


kvark:
@msiglreith thanks for the review!

Multiple viewports/scissors are certainly on the TODO list, but adding them into this PR is not fully straightforward (some parts assume a single thing, and we'll need to pass down the counts in extra fields).
Let's get them in follow-ups.

bors r=msiglreith

PR #1904, opened 2018-04-04
Image format properties

### LINKED ISSUE #1859: Querying image format properties
We need a decently implemented analogue of `vkGetPhysicalDeviceImageFormatProperties`.

kvark:
Fixes #1859
PR checklist:
- [x] tested examples with the following backends: Vk

Side changes:
  - rename `ImageLayout` -> `Layout`
  - add `image::Tiling` parameter to `create_image`


src/backend/dx12/src/lib.rs
```diff
@@ -322,6 +322,46 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+                1 => d3d12::D3D12_REQ_TEXTURE1D_ARRAY_AXIS_DIMENSION as _,
+                2 => d3d12::D3D12_REQ_TEXTURE2D_ARRAY_AXIS_DIMENSION as _,
+                _ => return None,
+            },
+            sample_count_mask: if dimensions == 2 &&
+                (usage.contains(image::Usage::COLOR_ATTACHMENT) | usage.contains(image::Usage::DEPTH_STENCIL_ATTACHMENT))
+            {
+                0xFFFE //TODO: use D3D12_FEATURE_DATA_FORMAT_SUPPORT
+            } else {
+                0x2
```
msiglreith: Shouldn't this be `0x1`? Or do you define it as `1  << SampleCount` ?

src/backend/dx12/src/lib.rs
```diff
@@ -322,6 +322,46 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
         format::query_properties()[idx]
     }
 
+    fn image_format_properties(
+        &self, _format: f::Format, dimensions: u8, _tiling: image::Tiling,
```
msiglreith: Should handle tiling also in case of linear tiling:
- only 2d
- single mip and array layer
- single sample count

src/backend/vulkan/src/lib.rs
```diff
@@ -432,6 +430,44 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+        {
+            Ok(props) => Some(image::FormatProperties {
+                max_extent: image::Extent {
+                    width: props.max_extent.width,
+                    height: props.max_extent.height,
+                    depth: props.max_extent.depth,
+                },
+                max_levels: props.max_mip_levels as _,
+                max_layers: props.max_array_layers as _,
+                sample_count_mask: props.sample_counts.flags() as _,
```
msiglreith: This kind of mismatches with the sample bit mask used in dx12 and metal:
1-bit sample will return 1 instead of 2.

src/hal/src/adapter.rs
```diff
@@ -73,7 +73,15 @@ pub trait PhysicalDevice<B: Backend>: Any + Send + Sync {
     fn open(&self, families: Vec<(&B::QueueFamily, Vec<QueuePriority>)>) -> Result<Gpu<B>, DeviceCreationError>;
 
     /// Fetch details for a particular image format.
```
msiglreith: nit: we should probably remove the ìmage` part here

src/hal/src/image.rs
```diff
@@ -50,6 +50,18 @@ impl Offset {
...
 
+/// Image tiling modes.
+#[repr(u32)]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum Tiling {
+    /// Optimal tiling for GPU memory access. Implementation-dependent.
+    Optimal,
+    /// Optimal for CPU read/write. Texels are laid out in row-major order,
+    // possibly with some padding on each row.
```
msiglreith: nit : missing `/` at the beginning

kvark:
Nice review, thanks! Everything is addressed now, and the builds look green.
bors r=msiglreith

PR #1906, opened 2018-04-04
Add image usage to swapchain config

### LINKED ISSUE #1875: Add image usage flag for swapchain config
Allow to specify the image usage for swapchain presentable images on swapchain creation.
See https://github.com/gfx-rs/gfx/pull/1873

kichristensen:
Fixes #1875

PR checklist:
  - vulkan
  - dx12

### DIFF
--- a/src/hal/src/window.rs
+++ b/src/hal/src/window.rs
@@ -261,6 +261,18 @@ impl SwapchainConfig {
         self
     }
 
+    /// Specify the usage of backbuffer images.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    ///
+    /// ```
+    pub fn with_image_usage(mut self, usage: image::Usage) -> Self {
+        self.image_usage = usage;
+        self
+    }
+
     // TODO: depth-only, stencil-only, swapchain size, present modes, etc.
 }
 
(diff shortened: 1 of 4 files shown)


src/hal/src/window.rs
```diff
@@ -200,6 +200,8 @@ pub struct SwapchainConfig {
     pub depth_stencil_format: Option<Format>,
     /// Number of images in the swapchain.
     pub image_count: u32,
+    /// Image usage of the backbuffer images.
+    pub image_usage: Option<image::Usage>,
```
kvark: why is this optional?

kichristensen: You are correct, it actually isn't as "default" values are applied in create_swapchain in vulkan and dx12. What do you think the best way to handle this would be, just setting it to empty and force users to pass in the required flags, instead of setting the current "default" values in create_swapchain?

kvark: > "default" values are applied in create_swapchain in vulkan and dx12

where do you see the default usage value in Vulkan?

kichristensen: By the "default" value I mean the value that was supplied in the Vulkan backend, https://github.com/gfx-rs/gfx/pull/1906/files#diff-055df6f698963e935780372c2266b221L1421 and for the DX12 backend https://github.com/gfx-rs/gfx/pull/1906/files#diff-30c7631df6499e30d85955a6141462e4L2599

kvark: Let's just require the user to pass in the flags in full, no defaults needed

PR #1907, opened 2018-04-04
metal: Add debug capture scopes

scoopr:
Because of the way Swapchains are implemented, Xcode's frame capture
gets confused on where the frame starts and ends (which is the default
capture scope).

This commit creates a capture scope that tries to mimic the usual
behaviour. Enabled only when debug_assertations is on.

Fixes #issue
PR checklist:

I was just noodling around and this annoyed me a bit. The commit may not quite be ready for merge, perhaps those classes/methods should be exposed in metal-rs first?

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -20,6 +20,7 @@ use block::{ConcreteBlock};
 use conversions::{map_index_type};
 use soft;
 
+use objc::runtime::{Class, Object};
 
 pub struct CommandQueue(pub(crate) Arc<QueueInner>);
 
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/command.rs
```diff
@@ -775,6 +776,19 @@ impl RawCommandQueue<Backend> for CommandQueue {
                 msg_send![render_layer, setContents: io_surface.obj];
             }
         }
+
+        if cfg!(debug_assertions)
+        {
+            unsafe {
+                if let Some(mtl_capture_manager) = Class::get("MTLCaptureManager")
```
kvark: looks like we'll need some capture API in `metal-rs` to avoid raw obj-c handling here

scoopr:
Rebased, reformatted and added also a feature flag to enable it with release builds too.

PR #1912, opened 2018-04-06
Add destroy_swapchain method and update example

kichristensen:
This fixes the Vulkan object tracker error in the _hal/quad_ example:

```
ERROR 2018-04-06T18:20:46Z: gfx_backend_vulkan: [ObjectTracker] OBJ ERROR : For device 0x2a20eda1c50, SwapchainKHR object 0x3 has not been destroyed. The spec valid usage text states 'All child objects created on device must have been destroyed prior to destroying device' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-vkDestroyDevice-device-00378)
```

PR checklist:
  - DX12
  - GL
  - Vulkan

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -610,10 +610,11 @@ fn main() {
     for framebuffer in framebuffers {
         device.destroy_framebuffer(framebuffer);
     }
-    for (image, rtv) in frame_images {
+    for (_, rtv) in frame_images {
         device.destroy_image_view(rtv);
-        device.destroy_image(image);
     }
+
+    device.destroy_swapchain(swap_chain);
 }
 
 #[cfg(not(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl")))]
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1088,6 +1088,10 @@ impl d::Device<B> for Device {
         self.create_swapchain_impl(surface, config)
     }
 
+    fn destroy_swapchain(&self, _swapchain: Swapchain) {
+        unimplemented!()
+    }
+
     fn wait_idle(&self) -> Result<(), error::HostExecutionError> {
         unsafe { self.share.context.Finish(); }
         Ok(())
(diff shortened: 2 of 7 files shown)


examples/hal/quad/main.rs
```diff
@@ -612,8 +612,11 @@ fn main() {
     }
     for (image, rtv) in frame_images {
         device.destroy_image_view(rtv);
+        #[cfg(not(feature="vulkan"))]
```
kvark: why aren't we doing that for Vulkan? We shouldn't need any config flags here

kichristensen: Because in Vulkan when destorying the swapchain it also destroys images associated with the swapchain, if the images is destroyed before the swapchain the program crashes with a memory access violation

kvark: We should just follow Vulkan then and not require the user to destroy them.

kichristensen: You mean delete the line and config check? If so what about DX12 where the user should destroy the images?

kvark: In the API we have (even if you don't consider Vulkan Portability initiative), since the user doesn't explicitly create swapchain images, they shouldn't be explicitly destroying them. DX12 needs to have some logic to handle this automatically, and complain loudly if the user tries to destroy those images manually.

src/backend/gl/src/device.rs
```diff
@@ -1088,6 +1088,10 @@ impl d::Device<B> for Device {
         self.create_swapchain_impl(surface, config)
     }
 
+    fn destroy_swapchain(&self, _swapchain: Swapchain) {
+        unimplemented!()
```
kvark: let's keep the example running on GL

kichristensen: It is not possible to get the example to run on GL even before this, as almost all destroy functions for the GL implementation is unimplemented

kichristensen:
@kvark Commits is now squashed

kvark:
Thank you!
bors r+

PR #1913, opened 2018-04-07
[osx] virtualized memory mapping

### LINKED ISSUE #1911: Virtualized memory allocations/mapping on Metal
On Metal, we can only map resources, so the Vulkan client code that tries to allocate memory and map it (without binding it to a particular resource first) fails. In D3D12, we work around this by associating a memory chunk with a hidden buffer that we can map. We need to figure out what to do with Metal.

This is blocking Vulkan CTS runs, which (unfortunately) try to map memory very early and aggresively.

kvark:
Fixes #1911
PR checklist:

~~It doesn't work yet (for some reason)~~

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -796,7 +796,10 @@ impl RawCommandQueue<Backend> for CommandQueue {
     }
 
     fn wait_idle(&self) -> Result<(), error::HostExecutionError> {
-        unimplemented!()
+        let cmd_buffer = self.0.new_command_buffer_ref();
+        cmd_buffer.commit();
+        cmd_buffer.wait_until_completed();
+        Ok(())
     }
 }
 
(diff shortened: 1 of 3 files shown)


grovesNL:
If it's useful, I started to expose the capture manager for use with Xcode to [start the capture during queue creation](https://stackoverflow.com/questions/39254332/ios-metal-how-to-capture-gpu-frame-properly/46141675#46141675):
- [metal-rs changes](https://github.com/gfx-rs/metal-rs/pull/36)
- [using capture manager on queue creation in gfx](https://github.com/kvark/gfx/compare/osx-mapping...grovesNL:kvark-osx-mapping-with-capture)

However attaching the debugger appears to resolve this bug and everything renders fine in the hal/quad example... so maybe there is a synchronization issue.

kvark:
(As per gitter discussion)
Bors r=grovesNL

PR #1914, opened 2018-04-07
Make hal/quad runnable with GL

kichristensen:
The _hal/quad_ example would crash when running in GL mode, because a lot of the destroy methods where not implemented.

PR checklist:
  - GL

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1011,8 +1011,8 @@ impl d::Device<B> for Device {
         unimplemented!()
     }
 
-    fn free_memory(&self, _memory: n::Memory) {
-        // nothing to do
+    fn free_memory(&self, memory: n::Memory) {
+        // Nothing to do
     }
 
     fn create_query_pool(&self, _ty: query::QueryType, _count: u32) -> () {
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/device.rs
```diff
@@ -1011,8 +1011,9 @@ impl d::Device<B> for Device {
         unimplemented!()
     }
 
-    fn free_memory(&self, _memory: n::Memory) {
-        // nothing to do
+    fn free_memory(&self, memory: n::Memory) {
+        let gl = &self.share.context;
+        unsafe { gl.DeleteBuffers(1, &memory.first_bound_buffer.into_inner()) }
```
msiglreith: I believe that's a weak handle (no deletion here)

src/backend/gl/src/device.rs
```diff
@@ -1050,24 +1059,39 @@ impl d::Device<B> for Device {
...
+        }
     }
-    fn destroy_image_view(&self, _: n::ImageView) {
-        unimplemented!()
+
+    fn destroy_image_view(&self, image_view: n::ImageView) {
+        let gl = &self.share.context;
+        match image_view {
+            n::ImageView::Surface(rb) => unsafe { gl.DeleteRenderbuffers(1, &rb) },
+            n::ImageView::Texture(t, _) | n::ImageView::TextureLayer(t, _, _) => unsafe { gl.DeleteTextures(1, &t) },
```
msiglreith: Also weak handles here (until we use actual texture views)

kichristensen:
@msiglreith The commits have now been squashed

kvark:
thanks!
bors r=msiglreith

PR #1917, opened 2018-04-08
Set vulkan version

msiglreith:
Might help with https://github.com/msiglreith/gfx_ocean/issues/8
Considering the warnings/issues with the vulkan loader, we might want to consider using a specific value..

PR checklist:
- [x] tested quad examples with the following backends: vulkan


Lymia:
The TODO seems to imply the intent is to pass 1.1.0 and not 1.0.0?

PR #1924, opened 2018-04-11
Make allocate_set(s) failable

### LINKED ISSUE #1922: Make `allocate_set`/`allocate_sets` failable
The descriptor pool is limited, we can run out of descriptors.

This is kind of a blocker for VkPortability, since we are getting an out of descriptor error that's not handled properly by `ash`, and we need to forward it up instead of unwrapping. cc @MaikKlein

AIOOB:
Fixes #1922 
PR checklist:

### DIFF
--- a/src/backend/dx12/src/native.rs
+++ b/src/backend/dx12/src/native.rs
@@ -369,13 +369,16 @@ pub struct DescriptorHeapSlice {
 }
 
 impl DescriptorHeapSlice {
-    pub(crate) fn alloc_handles(&mut self, count: u64) -> DualHandle {
-        assert!(self.next + count <= self.range.end);
-        let index = self.next;
-        self.next += count;
-        DualHandle {
-            cpu: d3d12::D3D12_CPU_DESCRIPTOR_HANDLE { ptr: self.start.cpu.ptr + (self.handle_size * index) as usize },
-            gpu: d3d12::D3D12_GPU_DESCRIPTOR_HANDLE { ptr: self.start.gpu.ptr + (self.handle_size * index) as u64 },
+    pub(crate) fn alloc_handles(&mut self, count: u64) -> Option<DualHandle> {
+        if self.next + count <= self.range.end {
+            let index = self.next;
+            self.next += count;
+            Some(DualHandle {
+                cpu: d3d12::D3D12_CPU_DESCRIPTOR_HANDLE { ptr: self.start.cpu.ptr + (self.handle_size * index) as usize },
+                gpu: d3d12::D3D12_GPU_DESCRIPTOR_HANDLE { ptr: self.start.gpu.ptr + (self.handle_size * index) as u64 },
+            })
+        } else {
+            None
         }
     }
 }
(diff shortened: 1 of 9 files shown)


src/hal/src/pso/descriptor.rs
```diff
@@ -87,26 +87,42 @@ pub struct DescriptorRangeDesc {
...
+#[derive(Fail, Clone, Copy, Debug, PartialEq)]
+pub enum AllocationError {
+    /// Memory allocation on the host side failed.
+    /// This could be caused by a lack of memory or pool fragmentation.
+    #[fail(display = "Host memory allocation failed.")]
+    OutOfHostMemory,
+    /// Memory allocation on the host side failed.
+    /// This could be caused by a lack of memory or pool fragmentation.
+    #[fail(display = "Device memory allocation failed.")]
+    OutOfDeviceMemory,
```
AIOOB: The vulkan spec is unclear to me as to whether these errors mean something different to the fragmented pool error as I would assume it's possible to run out of memory without having a fragmented pool.

src/hal/src/pso/descriptor.rs
```diff
@@ -87,26 +87,42 @@ pub struct DescriptorRangeDesc {
...
     }
 
     /// Allocate one or multiple descriptor sets from the pool.
     ///
     /// Each descriptor set will be allocated from the pool according to the corresponding set layout.
-    /// The descriptor pool _must_ have enough space in to allocate the required descriptors.
     /// Descriptors will become invalid once the pool is reset. Usage of invalidated descriptor sets results
     /// in undefined behavior.
-    fn allocate_sets<I>(&mut self, layouts: I) -> Vec<B::DescriptorSet>
+    fn allocate_sets<I>(&mut self, layouts: I) -> Vec<Result<B::DescriptorSet, AllocationError>>
```
AIOOB: Vulkan provides a Result<Vec> but I think a Vec<Result> makes a bit more sense unless we want to promise that the length of the output and the length of the input will be equal.

src/backend/metal/src/native.rs
```diff
@@ -165,24 +165,24 @@ impl hal::DescriptorPool<Backend> for DescriptorPool {
...
             DescriptorPool::ArgumentBuffer { ref buffer, total_size, ref mut offset } => {
                 let (encoder, stage_flags) = match layout {
                     &DescriptorSetLayout::ArgumentBuffer(ref encoder, stages) => (encoder, stages),
-                    _ => panic!("Incompatible descriptor set layout type"),
+                    _ => return Err(pso::AllocationError::IncompatibleLayout),
                 };
 
                 let cur_offset = *offset;
                 *offset += encoder.encoded_length();
                 assert!(*offset <= total_size);
```
kvark: this is where we should return OOM error instead of asserting

src/backend/dx12/src/native.rs
```diff
@@ -447,13 +447,13 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
             };
         }
 
-        DescriptorSet {
+        Ok(DescriptorSet {
```
kvark: the OOM error would go from failing `alloc_handles`, although it's fine to do later on

AIOOB:
bors r=kvark

PR #1925, opened 2018-04-11
[vk] remove Drop for Device...

kvark:
Reverts #1921
PR checklist:


kvark:
somewhat urgent...
bors r+

PR #1931, opened 2018-04-12
[vk] disable dual-source blending on Intel/Windows

### LINKED ISSUE #1930: [vk] crash in dual-source blending on Windows + Intel
Short info header:
- GFX version: HAL master
- OS: Windows
- GPU: Intel HD Graphics 630

Linked issues:
  - https://software.intel.com/en-us/forums/graphics-driver-bug-reporting/topic/748226
  - https://github.com/szeged/webrender/issues/136
  - https://github.com/hrydgard/ppsspp/issues/10074
  - https://bugs.dolphin-emu.org/issues/10470

We need to:
  ~~1. confirm if it's the intel driver bug, check if a fix is coming (in which case we can wait)~~
  2. start black-listing some OS/driver/HW for specific features. This is unfortunate, but we'll get more exceptions like this in the long run, so we might as well start here.

kvark:
Fixes #1930
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -37,6 +37,7 @@ use std::sync::Arc;
 mod command;
 mod conv;
 mod device;
+mod info;
 mod native;
 mod pool;
 mod result;
@@ -511,6 +512,11 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     }
 
     fn features(&self) -> Features {
+        // see https://github.com/gfx-rs/gfx/issues/1930
+        let is_windows_intel_kaby = cfg!(windows) &&
+            self.properties.vendor_id == info::intel::VENDOR &&
+            self.properties.device_id & info::intel::DEVICE_KABY_LAKE_MASK == info::intel::DEVICE_KABY_LAKE_MASK;
+
         let features = self.instance.0.get_physical_device_features(self.handle);
         let mut bits = Features::empty();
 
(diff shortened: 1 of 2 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -511,6 +511,11 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     }
 
     fn features(&self) -> Features {
+        // see https://github.com/gfx-rs/gfx/issues/1930
+        let is_windows_intel_kaby = cfg!(windows) &&
+            self.properties.vendor_id == 0x8086 &&
+            self.properties.device_id >> 2 == 0x59;
```
jrmuizel: Do you know that this only applies Kabylake?

jrmuizel: Or to state more clearly, are there other Intel GPUs where this might be broken?

kvark: I based the filter on all the related issues of #1930 (reported in other places on the net). There might be more cases, we'll add them later anyway.

src/backend/vulkan/src/lib.rs
```diff
@@ -511,6 +511,11 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     }
 
     fn features(&self) -> Features {
+        // see https://github.com/gfx-rs/gfx/issues/1930
+        let is_windows_intel_kaby = cfg!(windows) &&
+            self.properties.vendor_id == 0x8086 &&
```
msiglreith: Could you declare them as constants somewhere? Might be helpful in case we need more of these exceptions.

src/backend/vulkan/src/lib.rs
msiglreith: How does is the Intel device ID built up? 2-bit shift compared with a hex value looks kind of suspicious.

kvark:
@msiglreith good review, thanks!

> Could you declare them as constants somewhere? Might be helpful in case we need more of these exceptions.

Was just too lazy, fixed now.

> How does is the Intel device ID built up? 2-bit shift compared with a hex value looks kind of suspicious.

That was the last moment change I screwed up. Was meant to shift by 2 hex digits, not bits.

PR #1932, opened 2018-04-12
Add inheritance info to begin command buffer

AIOOB:
Add `CommandBufferInheritanceInfo` which is nonnull for secondary command buffers. It is necessary for portability to continue the CTS.

PR checklist:
- [x] tested examples with the following backends: Vulkan, OpenGL

### DIFF
--- a/src/hal/src/command/raw.rs
+++ b/src/hal/src/command/raw.rs
@@ -2,11 +2,11 @@ use std::any::Any;
 use std::borrow::Borrow;
 use std::ops::Range;
 
-use {buffer, pso};
+use {buffer, pass, pso};
 use {Backend, IndexCount, InstanceCount, VertexCount, VertexOffset, WorkGroupCount};
 use image::{Filter, Layout, SubresourceRange};
 use memory::{Barrier, Dependencies};
-use query::{Query, QueryControl, QueryId};
+use query::{PipelineStatistic, Query, QueryControl, QueryId};
 use super::{
     AttachmentClear, BufferCopy, BufferImageCopy,
     ClearColor, ClearDepthStencil, ClearValue,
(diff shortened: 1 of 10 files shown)


src/hal/src/command/raw.rs
```diff
@@ -81,11 +81,20 @@ pub enum Level {
...
+    pub occlusion_query_flags: Option<QueryControl>,
+    pub pipeline_statistics: PipelineStatistic,
+}
+
 /// A trait that describes all the operations that must be
 /// provided by a `Backend`'s command buffer.
 pub trait RawCommandBuffer<B: Backend>: Clone + Any + Send + Sync {
     /// Begins recording commands to a command buffer.
-    fn begin(&mut self, flags: CommandBufferFlags);
+    fn begin(&mut self, flags: CommandBufferFlags, inheritance_info: Option<CommandBufferInheritanceInfo<B>>);
```
kvark: wouldn't it be easier to always pass the inheritance info (without `Option`) given a `Default` implementation on it that inherits nothing?

src/hal/src/command/raw.rs
```diff
@@ -81,11 +81,33 @@ pub enum Level {
...
+#[derive(Debug)]
+pub struct CommandBufferInheritanceInfo<'a, B: Backend> {
+    pub subpass: Option<pass::Subpass<'a, B>>,
+    pub framebuffer: Option<&'a B::Framebuffer>,
+    pub occlusion_query_enable: bool,
+    pub occlusion_query_flags: QueryControl,
+    pub pipeline_statistics: PipelineStatistic,
+}
+
+impl<'a, B: Backend> Default for CommandBufferInheritanceInfo<'a, B> {
```
kvark: we can get rid of this block if we put `#[derive(Default)]` on top of it (as well as `QueryControl` and `PipelineStatistic`).

PR #1936, opened 2018-04-13
[mtl] manually retain command buffer data

### LINKED ISSUE #1779: Use non-retaining command buffers on Metal
From https://developer.apple.com/documentation/metal/mtlcommandqueue/1508684-makecommandbufferwithunretainedr :
>Use this method only in extremely performance-critical applications where the cost of maintaining references to objects is significant, and references by some other part of the application can be guaranteed until the object is no longer needed for command execution.

So one important point is Apple did go this extra step to provide a method, meaning that the performance difference is expected. Since we are manually tracking the lifetimes on other backends, we should be using this code paths on Metal, minus some cases where we currently rely on tracking:
  - buffer updates of #1777

kvark:
Fixes #1779
PR checklist:

r? @JohnColanduoni @grovesNL 
As mentioned on gitter, this doesn't allow us to get rid of the autorelease pools unfortunately. We can see it as a different issue.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -33,7 +33,7 @@ unsafe impl Sync for QueueInner {}
 
 impl QueueInner {
     pub fn new_command_buffer_ref(&self) -> &metal::CommandBufferRef {
-        self.queue.new_command_buffer()
+        self.queue.new_command_buffer_with_unretained_references()
     }
 }
 
@@ -266,6 +266,7 @@ struct CommandBufferInner {
     // hopefully, this is temporary
     // currently needed for `update_buffer` only
     device: metal::Device,
+    retained_buffers: Vec<metal::Buffer>,
     //TODO: would be cleaner to move the cache into `CommandBuffer` iself
     // it doesn't have to be in `Inner`
     viewport: Option<MTLViewport>,
@@ -304,7 +305,9 @@ impl CommandBufferInner {
                 }
             }
         };
+
         self.reset_resources();
+        self.retained_buffers.clear();
     }
 
     fn stop_encoding(&mut self) {
@@ -715,7 +718,9 @@ impl CommandQueue {
 }
 
 impl RawCommandQueue<Backend> for CommandQueue {
-    unsafe fn submit_raw<IC>(&mut self, submit: RawSubmission<Backend, IC>, fence: Option<&native::Fence>)
+    unsafe fn submit_raw<IC>(
+        &mut self, submit: RawSubmission<Backend, IC>, fence: Option<&native::Fence>
+    )
     where
         IC: IntoIterator,
         IC::Item: Borrow<CommandBuffer>,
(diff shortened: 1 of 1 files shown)


kvark:
also r? @scoopr

src/backend/metal/src/command.rs
```diff
@@ -751,18 +754,25 @@ impl RawCommandQueue<Backend> for CommandQueue {
...
             }
             command_buffer.commit();
         }
+
+        if let Some(ref fence) = fence {
+            let command_buffer = self.0.new_command_buffer_ref();
+            let value_ptr = fence.0.clone();
+            let fence_block = ConcreteBlock::new(move |_cb: *mut ()| -> () {
+                *value_ptr.lock().unwrap() = true;
+            }).copy();
```
grovesNL: Not part of your change, but why is the `copy` needed here?

src/backend/metal/src/command.rs
```diff
@@ -751,18 +754,25 @@ impl RawCommandQueue<Backend> for CommandQueue {
...
-            // only append the fence handler to the last buffer
-            if i + 1 == num_buffers {
-                if let Some(ref fence) = fence {
-                    let value_ptr = fence.0.clone();
-                    let fence_block = ConcreteBlock::new(move |_cb: *mut ()| -> () {
-                        *value_ptr.lock().unwrap() = true;
-                    }).copy();
-                    msg_send![command_buffer, addCompletedHandler: fence_block.deref() as *const _];
-                }
+            if !inner.retained_buffers.is_empty() {
```
grovesNL: If it's deferred, doesn't this mean we will drop `retained_buffers` on first usage (i.e. application may not keep calling `update_buffer`)?

src/backend/metal/src/command.rs
kvark: uh, seems totally out of place

src/backend/metal/src/command.rs
kvark: Great eye! you spotted a bug :) lemme address it...

kvark:
@grovesNL the re-usable command buffers should be addressed now.
The `copy()` on `ConcreateBlock` appears to be needed though, it crushes without it, but I didn't investigate further. Perhaps @JohnColanduoni recalls?..
Anyhow, PTAL.

src/backend/metal/src/command.rs
```diff
@@ -736,13 +742,22 @@ impl RawCommandQueue<Backend> for CommandQueue {
...
+                        let retained_buffers = mem::replace(&mut inner.retained_buffers, Vec::new());
+                        let release_block = ConcreteBlock::new(move |_cb: *mut ()| -> () {
+                            let _ = retained_buffers; // move and auto-release
+                        }).copy();
+                        let cb_ref: &metal::CommandBufferRef = cmd_buffer;
+                        msg_send![cb_ref, addCompletedHandler: release_block.deref() as *const _];
+                    }
+                    cmd_buffer
+                }
+                CommandSink::Deferred { ref passes, .. } => {
```
grovesNL: Doesn't `update_buffer` still cause `retained_buffers` to become populated if the application is currently recording into `CommandSink::Deferred`? Or is this avoided somehow?

kvark: Correct, but for multi-use buffer (which is what `Deferred` is for) we don't want to set up the completion handler - retained buffer vectors is going to be cleaned up on `reset()`

grovesNL: Right, got it, thanks. Updated commit looks good to me then!

kvark:
Ok, thanks for a good review!
Bors r=grovesNL

PR #1937, opened 2018-04-14
added additional x11 and vulkan dev dependencies

rustfires:
compilation fails without libxcb1-dev and xorg-dev packages in ubuntu. so added them to list of packages to install


msiglreith:
bors r=kvark

PR #1939, opened 2018-04-15
[v0.17] Update glutin and winit

Rhuagh:
PR checklist:
- [x] tested examples with the following backends: OpenGL (mac)
No make on 0.17.


kvark:
Needs version bumps of the crates with affected public API

kvark:
Looks good now, thank you!
bors r+

PR #1944, opened 2018-04-17
[mtl] dependencies, dpi, more formats

kvark:
Improves the CTS support.
PR checklist:

~~A follow-up is planned to use https://github.com/gfx-rs/metal-rs/pull/37 , but it doesn't have to go here.~~

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -305,7 +305,7 @@ fn main() {
         pipeline
     };
 
-    println!("Pipeline: {:?}", pipeline);
+    //println!("Pipeline: {:?}", pipeline);
 
     // Descriptors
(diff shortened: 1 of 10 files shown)


kvark:
@grovesNL take another look please?

examples/hal/quad/main.rs
```diff
@@ -305,7 +305,7 @@ fn main() {
         pipeline
     };
 
-    println!("Pipeline: {:?}", pipeline);
+    //println!("Pipeline: {:?}", pipeline);
```
grovesNL: nit: was this left commented out on purpose? We could just remove it if it's not useful

kvark: it's really nit. The amount of stuff it prints on OSX (more than a page of text) is just to annoying. We can totally remove this for now.

PR #1946, opened 2018-04-17
[dx12] row pitch calculation fix for the worst case image-buffer copies

kvark:
Fixes part of https://github.com/szeged/webrender/issues/73#issuecomment-381637301
PR checklist:

r? @dati91 @msiglreith

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -742,8 +742,8 @@ impl CommandBuffer {
                         let cut_width = cmp::min(r.image_extent.width, cut_row_texels as image::Size);
                         let gap_texels = (row_offset - aligned_offset) as image::Size / image.bytes_per_block as image::Size;
                         // this is a conservative row pitch that should be compatible with both copies
-                        let max_unaligned_pitch = r.image_extent.width * image.bytes_per_block as u32;
-                        let row_pitch = (max_unaligned_pitch | d3d12::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT) + 1;
+                        let max_unaligned_pitch = (r.image_extent.width + gap_texels) * image.bytes_per_block as u32;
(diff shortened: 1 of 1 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -742,8 +742,8 @@ impl CommandBuffer {
                         let cut_width = cmp::min(r.image_extent.width, cut_row_texels as image::Size);
                         let gap_texels = (row_offset - aligned_offset) as image::Size / image.bytes_per_block as image::Size;
                         // this is a conservative row pitch that should be compatible with both copies
-                        let max_unaligned_pitch = r.image_extent.width * image.bytes_per_block as u32;
-                        let row_pitch = (max_unaligned_pitch | d3d12::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT) + 1;
+                        let max_unaligned_pitch = (r.image_extent.width + gap_texels) * image.bytes_per_block as u32;
```
msiglreith: What's the motivation behind this one?

kvark: The pitch is not really used here, since we copy row by row, but we need to make it at least as big as the rows we copy (to make DX12 happy). `r.image_extent.width + gap_texels` is the upper bound for those 2 copies to follow

dati91:
I tested it with the quad example (without the rowpitch_alignment so it went into the "wors case") and it works fine!

I tested it with our code as well, and it doesn't throw any dx12 error, which is great.
But there are some notable differences, i can't say for sure if it is related to this, because I didn't investigated it yet. But one thing to note, this is the exact same code which works fine with vulkan.

Blob example with dx12
dx12

A screenshot with renderdoc about the texture
dx12_renderdoc


Blob example with vulkan
vulkan

A screenshot with renderdoc about the texture
vulkan_renderdoc

It seems to me that this is only around 3/4 of the copies. And the green one shouldn't be in multiple pieces.

dati91:
@kvark without this patch, it doesn't even work (because a d3d12 error about the row_pitch), so i can't say its a regression :)
I'm just saying that the "worst case" needs further investigation (at least that is my first guess)

PR #1951, opened 2018-04-18
Adjust dx12 root descriptor set spaces

msiglreith:
Don't change the descriptor table space values for cbv/srv/uav and samplers. HLSL has additional signature parameter by indicating the descriptor type allowing to live two sets in harmony together.

Fixes combined image samplers for dx12!

PR checklist:
- [x] tested examples with the following backends: quad and some vulkan samples (dx12)


msiglreith:
@kvark 
https://msdn.microsoft.com/de-de/library/windows/desktop/dn859380(v=vs.85).aspx
`RangeType` matches with the `u/t/s/b` prefixes used in HLSL for the registers.
`s0, space0` and `t0, space0` seem to work fine together due to different range type.

PR #1953, opened 2018-04-19
Fix number of mipmaps calculation

msiglreith:
Should be `log2(x) + 1`

PR checklist:


src/hal/src/image.rs
```diff
@@ -306,7 +306,7 @@ impl Kind {
             _ => {
                 let extent = self.extent();
                 let dominant = max(max(extent.width, extent.height), extent.depth);
-                (1..).find(|level| dominant>>level <= 1).unwrap()
+                (1..).find(|level| dominant>>level <= 1).unwrap() + 1
```
kvark: I believe we should just do `dominant >> level == 0` instead

msiglreith: done

PR #1960, opened 2018-04-23
[mtl] buffer view support

### LINKED ISSUE #1935: Buffer view support on Metal
We have a few `unimplemented!()` blocks in the metal device related to buffer views. These are the current point of panic in OSX CTS, and we don't want to skip them (in general), so this is a blocker.

kvark:
Fixes #1935 
PR checklist:


kvark:
Closing if favor of #1966

PR #1964, opened 2018-04-23
[dx12] Add mapping between depth format and typeless format

callym:
This builds upon #c9b825 which automatically switches between D32-R32 depending on whether you're creating a depth or shader attachment. I'm not sure if this PR is actually needed, but I think the proper thing to do is create the image as `TYPELESS` format, and then the views are responsible for the types, so it just does this conversion in the background if you've asked for a depth format and also said you're going to be sampling from it.

PR checklist:

### DIFF
--- a/src/backend/dx12/src/conv.rs
+++ b/src/backend/dx12/src/conv.rs
@@ -99,6 +99,20 @@ pub fn map_format_dsv(surface: SurfaceType) -> Option<DXGI_FORMAT> {
     })
 }
 
+pub fn map_format_typeless(format: Format, usage: image::Usage) -> Option<DXGI_FORMAT> { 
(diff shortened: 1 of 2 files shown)


callym:
I'm not sure exactly why it's needed, but when I was looking into the issues I was having before c9b8254, searching the DX12 error message came up with creating a typeless image and use typed views to cast the values.

I'm not sure if this is actually technically needed though.

msiglreith:
It shouldn't be needed as the formats are compatible.

msiglreith:
AFAIK casting between formats of the same family is fine, therefore it's not mandatory to convert to the typeless format. So I'm closing, thanks nonetheless for the setting up the PR!

PR #1966, opened 2018-04-23
[mtl] Buffer views, combined image-samplers

### LINKED ISSUE #1935: Buffer view support on Metal

kvark:
Fixes #1935
PR checklist:


src/backend/metal/src/device.rs
```diff
@@ -1250,11 +1281,8 @@ impl hal::Device<Backend> for Device {
         if self.private_caps.resource_heaps {
             // We don't know what memory type the user will try to allocate the buffer with, so we test them
             // all get the most stringent ones. Note we don't check Shared because heaps can't use it
-            for &options in [
-                MTLResourceOptions::StorageModeManaged,
-                MTLResourceOptions::StorageModeManaged | MTLResourceOptions::CPUCacheModeWriteCombined,
-                MTLResourceOptions::StorageModePrivate,
-            ].iter() {
+            for mt in &self.memory_types {
+                let options = map_memory_properties_to_options(mt.properties);
```
grovesNL: What prevents `Shared` from being specified now? Or does above the comment no longer apply?

src/backend/metal/src/device.rs
```diff
@@ -1314,13 +1344,34 @@ impl hal::Device<Backend> for Device {
...
+            Some(fmt) => fmt,
+            None => return Err(buffer::ViewError::Unsupported),
+        };
+        let format_desc = format.surface_desc();
+        let block_count = (end_rough - start) * 8 / format_desc.bits as u64;
+        let size = block_count * (format_desc.bits as u64 / 8);
+        let mtl_format = map_format(format).ok_or(buffer::ViewError::Unsupported)?;
+
+        let descriptor = metal::TextureDescriptor::new();
+        descriptor.set_texture_type(MTLTextureType::D2);
```
grovesNL: Should we set `storageMode` here to match the buffer's storage mode also? Otherwise we get [the default](https://developer.apple.com/documentation/metal/mtltexturedescriptor/1516262-storagemode?language=objc#discussion):
> In iOS and tvOS, the default value is MTLStorageModeShared. In macOS, the default value is MTLStorageModeManaged.

but for the texture descriptor we pass to `new_texture_from_contents`, [it's required that](https://developer.apple.com/documentation/metal/mtlbuffer/1613852-newtexturewithdescriptor#discussion):
> storageMode can be MTLStorageModePrivate (all), MTLStorageModeShared (iOS or tvOS), or MTLStorageModeManaged (macOS). The storage mode of the buffer and the texture must match.

src/hal/src/format.rs
```diff
@@ -569,10 +569,15 @@ impl Format {
     ///
     /// Returns `None` if format is `Undefined`.
     pub fn base_format(self) -> BaseFormat {
-        assert!(NUM_FORMATS > self as usize);
+        assert!(self as usize != 0 && NUM_FORMATS > self as usize);
```
grovesNL: Is there still a way to do this now that `Undefined` has been removed?

src/backend/metal/src/device.rs
kvark: eh, right, hmmm

src/hal/src/format.rs
kvark: forwarding to @msiglreith

msiglreith: Might be. For portability a crash will most likely be fine but probably nice to have an additional check here.

kvark:
Concerns are addressed now. Thanks for the review!
bors r=grovesNL

PR #1969, opened 2018-04-23
[dx12] Check for WARP driver

bbrown683:
The code simply checks if the flag for software rendering is set for the adapter to determine if WARP was enabled.

Fixes #issue
PR checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/1969)

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -735,7 +735,7 @@ impl hal::Instance for Instance {
                 name: device_name,
                 vendor: desc.VendorId as usize,
                 device: desc.DeviceId as usize,
-                software_rendering: false, // TODO: check for WARP adapter (software rasterizer)?
+                software_rendering: (desc.Flags & dxgi::DXGI_ADAPTER_FLAG_SOFTWARE) != 0,
             };
 
             let mut features: d3d12::D3D12_FEATURE_DATA_D3D12_OPTIONS = unsafe { mem::zeroed() };


bbrown683:
Updated to use the appropriate flag for detecting if it software rendering is being enabled.

src/backend/dx12/src/lib.rs
```diff
@@ -731,11 +731,21 @@ impl hal::Instance for Instance {
                 name.to_string_lossy().into_owned()
             };
 
+            // Check for WARP driver to verify if software rendering is being used.
+            // Verified by checking the existence of the flag: DXGI_ADAPTER_FLAG_SOFTWARE.
+            let software_renderer = {
```
kvark: `if XXX {true} else {false}` is sort of a strange pattern :) please simplify

src/backend/dx12/src/lib.rs
```diff
@@ -731,11 +731,21 @@ impl hal::Instance for Instance {
...
+                    false
+                }
+            };
+
             let info = hal::AdapterInfo {
                 name: device_name,
                 vendor: desc.VendorId as usize,
                 device: desc.DeviceId as usize,
-                software_rendering: false, // TODO: check for WARP adapter (software rasterizer)?
+                software_rendering: software_renderer,
```
kvark: if you rename the variable, you could just do `software_rendering,` here. Alternative, move the expression here

bbrown683:
Cleaned up logic and packed into a single commit.

kvark:
bors r=msiglreith

Issue #1975, opened 2018-04-26
Metal buffer views of depth/stencil formats


kvark:
There is a (blocker) problem encountered in OSX CTS work, occurring in `create_buffer_view`. When the format requested is depth/stencil, metal validation error complains that those can't be linear (exact error to be provided here).
Given that the Vulkan spec only allows OUT_OF_MEMORY errors from this function, and issuing those stops any further CTS runs.

Possible ways to solve this:
  - try returning a different error for portability specifically
  - try convincing Metal that it's fiiine
  - implement a real workaround in Metal backend to support these cases

grovesNL:
We need to handle compressed textures somehow too, i.e. block compression (`BCx`):

>Test case 'dEQP-VK.api.buffer_view.create.suballocation.uniform.bc1_rgb_unorm_block'..
validateStrideTextureParameters:1165: failed assertion `Linear texture: cannot create compressed, depth, or stencil textures, pixelFormat (MTLPixelFormatBC1_RGBA) disallowed'

PR #1979, opened 2018-04-29
mtl: Add more format conversions

grovesNL:
Hopefully helps CTS slightly

PR checklist:


grovesNL:
Maybe I should keep all compressed formats commented out for now unless we can introduce a workaround easily (see https://github.com/gfx-rs/gfx/issues/1975#issuecomment-385215148)

Issue #1542, opened 2017-10-04
Multisampling support


msiglreith:
We currently don't have multisampling support and it's one of our few missing features for pipelines.
As always our aim is to stick as close as possible to Vulkan's API for this.

Information about the dx12 handling by @kvark:
> it's not a direct equivalent... but basically there are 3 (actually, 4) sample counts:
> 1. number of fragments in the render target = actual color values it stores
> 2. number of samples in the render target = "virtual color values" which are associated with specific locations inside a pixel but resolve to those fragments in the end. s2 >= s1
> 3. number of rasterized samples. If you have 2 fragments and 8 samples, you can rasterize at 4 samples in theory, each rasterizer sample would then map to 2 target virtual samples (I'm not aware of anyone actually doing that).
> 4. number of depth fragments, can be technically different from color ones
>
> so d3d12 sample count = s1. quality level = log2(s2 - s1), ForcedSampleCount = s3
actually, there is 5) number of shaded samples (how many PS invocations per pixel, basically)
so vulkans minSampleShading is s5 / s3

Further investigation is needed here how to exactly map the different APIs, also details for MTL are currently missing.

Additional resources on this topic:
* https://github.com/gpuweb/gpuweb/issues/26

PR #1981, opened 2018-04-30
Initial multisampling support

msiglreith:
* hal: Add resolve attachments, limits and multisampling desc to pso
* dx12: Handle sampling settings for images and pipeline states
* dx12: Add resolve operations for renderpasses
* dx12: Update resource barriers between subpasses

TODO:
- [x] dx12: Subpass transitions are not 100% correct yet
- [x] dx12: Select correct subresource on resolve
- [x] Implement for vulkan
- [x] Fix build errors on other backends (gl and metal)

Fixes https://github.com/gfx-rs/gfx/issues/1542
PR checklist:
- [x] reftests working on dx12
- [x] `multisampling` example on backends: dx12

### DIFF
--- a/examples/hal/quad/main.rs
+++ b/examples/hal/quad/main.rs
@@ -167,6 +167,7 @@ fn main() {
     let render_pass = {
         let attachment = pass::Attachment {
             format: Some(surface_format),
+            samples: 1,
             ops: pass::AttachmentOps::new(pass::AttachmentLoadOp::Clear, pass::AttachmentStoreOp::Store),
             stencil_ops: pass::AttachmentOps::DONT_CARE,
             layouts: i::Layout::Undefined .. i::Layout::Present,
@@ -176,6 +177,7 @@ fn main() {
             colors: &[(0, i::Layout::ColorAttachmentOptimal)],
             depth_stencil: None,
             inputs: &[],
+            resolves: &[],
             preserves: &[],
         };
 
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -293,6 +293,14 @@ pub struct CommandBuffer {
 unsafe impl Send for CommandBuffer { }
 unsafe impl Sync for CommandBuffer { }
 
+// Insetion point for subpasses.
+enum BarrierPoint {
+    // Pre barriers are inserted of the beginning when switching into a new subpass.
+    Pre,
+    // Post barriers are applied after exectuing the user defined commands.
+    Post,
+}
+
 impl CommandBuffer {
     pub(crate) fn new(
         raw: ComPtr<d3d12::ID3D12GraphicsCommandList>,
@@ -333,10 +341,13 @@ impl CommandBuffer {
         self.vertex_buffer_views = [NULL_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS];
     }
 
-    fn insert_subpass_barriers(&self) {
+    fn insert_subpass_barriers(&self, insertion: BarrierPoint) {
         let state = self.pass_cache.as_ref().unwrap();
-        let proto_barriers = match state.render_pass.subpasses.get(self.cur_subpass) {
-            Some(subpass) => &subpass.pre_barriers,
+        let proto_barriers =  match state.render_pass.subpasses.get(self.cur_subpass) {
+            Some(subpass) => match insertion {
+                BarrierPoint::Pre => &subpass.pre_barriers,
+                BarrierPoint::Post => &subpass.post_barriers,
+            },
             None => &state.render_pass.post_barriers,
         };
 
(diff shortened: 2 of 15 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -414,6 +425,29 @@ impl CommandBuffer {
...
+
+        for (i, resolve_attachment) in subpass.resolve_attachments.iter().enumerate() {
+            let (dst_attachment, _) = *resolve_attachment;
+            let (src_attachment, _) = subpass.color_attachments[i];
+
+            let resolve_src = state.framebuffer.attachments[src_attachment];
+            let resolve_dst = state.framebuffer.attachments[dst_attachment];
+
+            unsafe {
+            self.raw.ResolveSubresource(
```
kvark: nit: indentation is off

src/backend/dx12/src/lib.rs
```diff
@@ -917,6 +917,9 @@ impl hal::Instance for Instance {
                     min_buffer_copy_offset_alignment: d3d12::D3D12_TEXTURE_DATA_PLACEMENT_ALIGNMENT as _,
                     min_buffer_copy_pitch_alignment: d3d12::D3D12_TEXTURE_DATA_PITCH_ALIGNMENT as _,
                     min_uniform_buffer_offset_alignment: 256, // Required alignment for CBVs
+                    framebuffer_color_samples_count: 0b111,
```
kvark: any particular reason for 7 here?

msiglreith: Should be 0b101 (4_BIT | 1_BIT). Basically the sample counts supported for each render target.

src/backend/dx12/src/lib.rs
```diff
@@ -919,6 +919,9 @@ impl hal::Instance for Instance {
                     min_texel_buffer_offset_alignment: 1, // TODO
                     min_uniform_buffer_offset_alignment: 256, // Required alignment for CBVs
                     min_storage_buffer_offset_alignment: 1, // TODO
+                    framebuffer_color_samples_count: 0b101,
```
kvark: any reason this only supports 4xMSAA?

msiglreith: It's the minimum supported sample count for all render target formats. Every format must support at least these sample counts ( 1 and 4 ).

kvark: I see. Let's put a `//TODO` to expose more sample counts in the future?

examples/hal/quad/main.rs
```diff
@@ -176,6 +177,7 @@ fn main() {
             colors: &[(0, i::Layout::ColorAttachmentOptimal)],
             depth_stencil: None,
             inputs: &[],
+            resolves: &[],
```
kvark: We should probably have `Default` implemented for `pass::SubpassDesc`, so that cases like this can benefit from only supplying a few fields

src/backend/dx12/src/command.rs
```diff
@@ -414,6 +425,44 @@ impl CommandBuffer {
...
+                let subresource_src = resolve_src.calc_subresource(
+                    resolve_src.mip_levels.0 as _,
+                    (resolve_src.layers.0 + l) as _,
+                );
+                let subresource_dst = resolve_dst.calc_subresource(
+                    resolve_dst.mip_levels.0 as _,
+                    (resolve_dst.layers.0 + l) as _,
+                );
+
+                // TODO: take widht and height of render area into account.
```
kvark: typo: "widht"

src/backend/dx12/src/device.rs
```diff
@@ -1382,6 +1420,11 @@ impl d::Device<B> for Device {
             (rtvs, num_rtvs)
         };
 
+        let sample_desc = dxgitype::DXGI_SAMPLE_DESC {
+            Count: desc.multisampling.as_ref().map_or(1, |ms| ms.rasterization_samples as _),
```
kvark: nit: at some point, the chain of commands with closures becomes less readable than a simple `match`

src/backend/dx12/src/device.rs
```diff
@@ -1904,6 +1952,10 @@ impl d::Device<B> for Device {
             } else {
                 None
             },
+            dxgi_format: image.dxgi_format,
+            num_levels,
```
kvark: any reason these can't be stored as `Range` as opposed to the tuples?

src/backend/vulkan/src/device.rs
```diff
@@ -471,17 +472,38 @@ impl d::Device<B> for Device {
...
-            });
+                rasterization_samples: vk::SAMPLE_COUNT_1_BIT,
+                sample_shading_enable: vk::VK_FALSE,
+                min_sample_shading: 0.0,
+                p_sample_mask: ptr::null(),
+                alpha_to_coverage_enable: vk::VK_FALSE,
+                alpha_to_one_enable: vk::VK_FALSE,
+            };
+
+            if let Some(ref multisample) = desc.multisampling {
```
kvark: can we please do `let multisampling = match desc.multisampling {...}` instead of mutating it?

src/hal/src/lib.rs
```diff
@@ -247,7 +247,12 @@ pub struct Limits {
     pub min_uniform_buffer_offset_alignment: buffer::Offset,
     /// The alignment of the start of buffer used as a storage buffer, in bytes, non-zero.
     pub min_storage_buffer_offset_alignment: buffer::Offset,
-}
+    /// Number of samples supported for color attachments of framebuffers (floating/fixed point).
+    pub framebuffer_color_samples_count: image::NumSamples,
+    /// Number of samples supported for depth attachments of framebuffers.
+    pub framebuffer_depth_samples_count: image::NumSamples,
+    /// Number of samples supported for stencil attachments of framebuffers.
+    pub framebuffer_stencil_samples_count: image::NumSamples,}
```
kvark: weird closing bracket position

src/hal/src/pso/graphics.rs
```diff
@@ -269,3 +267,24 @@ pub enum LogicOp {
     Nand,
     Set,
 }
+
+///
+// TODO: Max 64 samples supported atm
```
kvark: 64 should be 'naff for everyone!

src/backend/dx12/src/device.rs
msiglreith: wanted to preserve the `Copy` derive

kvark: FYI, I'm not really a fan of getting `Copy` on everything that's copyable. Rust programs are already notorious for spending most of their run times copying bits around (as opposed to be doing actual work, lol), and making that cost implicit doesn't help. At least when you are forced to do a `clone()` that makes you consider using a reference here and there and minimize the copy costs.

PR #1982, opened 2018-05-02
Metal image copies and shader-based blits

kvark:
Replaces #1838 
~~Note: it only supports copying images of the same size so far, the actual shader usage is TODO~~
blit shader is included and even tested by Warden

PR checklist:

Unfinished parts left for follow-ups:
  - depth/stencil copies
  - selecting array/non-array shader versions
  - more tests


src/backend/metal/src/command.rs
```diff
@@ -644,49 +645,72 @@ pub(crate) fn exec_blit(encoder: &metal::BlitCommandEncoderRef, command: &soft::
...
                 region.dst as NSUInteger,
                 region.size as NSUInteger
             );
-        },
+        }
+        Cmd::CopyImage { ref src, ref dst, ref region } => {
+            let size = conv::map_extent(region.extent);
+            let src_offset = conv::map_offset(region.src_offset);
+            let dst_offset = conv::map_offset(region.dst_offset);
+            for lid in 0 .. {
```
msiglreith: Can we just do `end-start` instead of break condition inside the loop?

src/backend/metal/src/command.rs
```diff
@@ -1090,17 +1127,173 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+
+        self.sink().blit_commands(blit_commands.into_iter());
+
+        if !blit_vertices.is_empty() {
+            let inner = unsafe { &mut *self.inner.get() };
+
+            // Note: we don't bother to restore any render states here, since we are currently
+            // outside of a render pass, and the state will be reset automatically once
+            // we enter the next pass.
+            let mut pipes = self.shared.service_pipes
```
grovesNL: How do we enforce that we're outside of a renderpass (just curious)?

kvark: I guess there is a bit of ambiguity on render passes: there are Vulkan RPs in the API, and Metal RPs in the implementation.
On the API side, Vulkan specification guarantees that `vkCmdBlitImage` is only valid outside of a render pass.
In the implementation, we can just stop an active render pass at any point, technically, and start a new one. But I think we aren't going to have an active one outside of an API RP anyway.

grovesNL: Ok makes sense. So is there a reason we don't fail hard when `blit_image` is called inside the Vulkan/API RP within the Metal backend (since this can fail on the Vulkan backend)?

src/backend/metal/src/conversions.rs
```diff
@@ -291,16 +291,20 @@ pub fn resource_options_from_storage_and_cache(storage: MTLStorageMode, cache: M
...
+    // Note: for blitting, we do actual rendering, so we add more flags for TRANSFER_* usage
+    if usage.intersects(U::COLOR_ATTACHMENT | U::DEPTH_STENCIL_ATTACHMENT | U::TRANSFER_DST) {
         texture_usage |= MTLTextureUsage::MTLTextureUsageRenderTarget;
     }
-    if usage.contains(image::Usage::SAMPLED) {
+    if usage.intersects(U::SAMPLED | U::TRANSFER_SRC) {
         texture_usage |= MTLTextureUsage::MTLTextureUsageShaderRead;
     }
-    if usage.contains(image::Usage::STORAGE) {
+    if usage.intersects(U::STORAGE) {
```
grovesNL: I guess `contains` would still be fine here?

src/backend/metal/shaders/commands.metal
```diff
@@ -0,0 +1,35 @@
...
+
+fragment float4 ps_blit(
+    VertexData in [[stage_in]],
+    texture2d<float> tex2D [[ texture(0) ]],
+    sampler sampler2D [[ sampler(0) ]]
+) {
+  return tex2D.sample(sampler2D, in.uv.xy, level(in.uv.w));
+}
+
+fragment float4 ps_blit_array(
```
grovesNL: For future usage I assume?

reftests/local.ron
```diff
@@ -1,10 +1,30 @@
 {
```
grovesNL: Not a big deal, but does ron require tab spacing? Spaces would be a bit cleaner for these diffs if we could reformat the reftests at some point

kvark:
Will address those in follow-ups, thanks for the review!
Bors r=grovesNL

PR #1983, opened 2018-05-02
dx12: Block compressed format support for buffer-image copies

msiglreith:
Tested against a few of the vulkan samples, hit every path at least once iirc.

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -38,8 +38,11 @@ fn get_rect(rect: &pso::Rect) -> d3d12::D3D12_RECT {
 }
 
 fn div(a: u32, b: u32) -> u32 {
-    assert_eq!(a % b, 0);
-    a / b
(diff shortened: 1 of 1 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -38,8 +38,11 @@ fn get_rect(rect: &pso::Rect) -> d3d12::D3D12_RECT {
 }
 
 fn div(a: u32, b: u32) -> u32 {
-    assert_eq!(a % b, 0);
-    a / b
+    (a + b - 1) / b
```
kvark: oh wow, this would change the semantics completely, wouldn't it?

grovesNL: It's just `ceil` of `a / b`

kvark: Heh, I misread. Indeed fine is it. /yoda

PR #1985, opened 2018-05-04
mtl: Map more vertex formats

grovesNL:
Depends on gfx-rs/metal-rs#41 to be merged/published.

Not sure what to do with `Int1010102Normalized` and `UInt1010102Normalized` because the byte ordering for alpha appears to differ from Vulkan, so I'm leaving them out for now.

PR checklist:


kvark:
I published https://crates.io/crates/metal-rs/0.8.3 and added you to the owners of metal-rs.
The format issues are added to #1354
bors r+

PR #1986, opened 2018-05-04
[pre-ll] Update glutin 0.15, winit 0.13

alexheretic:
* Update winit & glutin
* Update image dev-dependency
* Up versions to v0.17 nexts

PR checklist:


kvark:
Oh god, these are steadily breaking once per month lately
Thank you for the PR though!
bors r+

PR #1987, opened 2018-05-04
dx12: Implement buffer views

msiglreith:
Untested so far, probably requires small changes to portability due to API breakage.

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -19,7 +19,7 @@ use hal::queue::{RawCommandQueue, QueueFamilyId};
 use hal::range::RangeArg;
 
 use {
-    conv, free_list, native as n, root_constants, window as w,
+    conv, format as fmt, free_list, native as n, root_constants, window as w,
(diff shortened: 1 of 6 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -2193,7 +2274,16 @@ impl d::Device<B> for Device {
                         dst_samplers.push(bind_info.sampler_range.as_ref().unwrap().at(offset));
                         num_samplers.push(1);
                     }
-                    pso::Descriptor::TexelBuffer(_) => unimplemented!()
+                    pso::Descriptor::UniformTexelBuffer(buffer_view) => {
+                        src_views.push(buffer_view.handle_srv);
```
kvark: given that the SRV here and UAV below can be NULL, can we at least `error!()` on it?

kvark:
Looks like Metal needs a few more corrections:
https://travis-ci.org/gfx-rs/gfx/jobs/374861471#L2659

msiglreith:
bors r=kvark

Issue #1684, opened 2017-12-19
First slot for vertex buffer binding


msiglreith:
First slot at `bind_vertex_buffers` is fixed at 0 atm. Users should be able to choose a different value.

msiglreith:
Fixed by https://github.com/gfx-rs/gfx/pull/1990

PR #1990, opened 2018-05-05
[ll] First slot for vertex buffer binding

jbatez:
Implements #1684

I want to start contributing to this project and this issue seemed like a good introduction. I apologize if I'm not doing something according to procedures.

Tested on Windows 10 with `dx12` and `vulkan` backends, Ubuntu 18.04 with `vulkan`, and macOS 10.13 with `metal`.

One thing I noticed while adding this that seemed a little bit awkward is that `GraphicsPipelineDesc::vertex_buffers` doesn't provide any way to skip around binding indices. The roughly equivalent part of Vulkan, `VkVertexInputBindingDescription`, allows you to specify the binding `index` explicitly along with the `stride` and `rate` instead of just making it implicit. Food for thought.

PR checklist:
- [X] tested examples with the following backends: `dx12`, `vulkan`, `metal`

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -282,6 +282,8 @@ pub struct CommandBuffer {
     // Cached vertex buffer views to bind.
     // `Stride` values are not known at `bind_vertex_buffers` time because they are only stored
     // inside the pipeline state.
+    vertex_buffers_start_slot: UINT,
+    vertex_buffers_num_views: UINT,
     vertex_buffer_views: [d3d12::D3D12_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS],
 
     // Re-using allocation for the image-buffer copies.
@@ -321,6 +323,8 @@ impl CommandBuffer {
             active_bindpoint: BindPoint::Graphics,
             occlusion_query: None,
             pipeline_stats_query: None,
+            vertex_buffers_start_slot: 0,
+            vertex_buffers_num_views: 0,
             vertex_buffer_views: [NULL_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS],
             copies: Vec::new(),
             viewport_cache: SmallVec::new(),
@@ -341,6 +345,8 @@ impl CommandBuffer {
         self.active_bindpoint = BindPoint::Graphics;
         self.occlusion_query = None;
         self.pipeline_stats_query = None;
+        self.vertex_buffers_start_slot = 0;
+        self.vertex_buffers_num_views = 0;
         self.vertex_buffer_views = [NULL_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS];
     }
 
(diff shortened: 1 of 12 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -531,16 +537,11 @@ impl CommandBuffer {
         // Bind vertex buffers
         // We currently don't support offsets for vertex buffer binding, therefore,
```
msiglreith: please remove the comment here, outdated

src/backend/dx12/src/command.rs
```diff
@@ -1243,9 +1244,12 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         }
     }
 
-    fn bind_vertex_buffers(&mut self, vbs: pso::VertexBufferSet<Backend>) {
+    fn bind_vertex_buffers(&mut self, first_binding: u32, vbs: pso::VertexBufferSet<Backend>) {
         // Only cache the vertex buffer views as we don't know the stride (PSO).
-        for (&(buffer, offset), view) in vbs.0.iter().zip(self.vertex_buffer_views.iter_mut()) {
+        self.vertex_buffers_start_slot = first_binding;
```
msiglreith: Multiple calls to bind_vertex_buffers will overwrite this. As the actual native call is deferred, we need to keep track of which vertex buffer are updated.

src/backend/dx12/src/command.rs
```diff
@@ -1243,9 +1244,12 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+        let clamped_start_slot = cmp::min(first_binding as _, MAX_VERTEX_BUFFERS);
```
msiglreith: Rather assert in case the function call violates boundaries. Easier to debug.

jbatez:
Done and done! I think I had somehow got it in my head that it was only valid to call `bind_vertex_buffers` once before resetting because I looked at the incomplete `gl` implementation first.

src/backend/dx12/src/command.rs
```diff
@@ -529,20 +529,28 @@ impl CommandBuffer {
...
-            .unwrap_or(MAX_VERTEX_BUFFERS);
-
-        unsafe {
-            cmd_buffer.IASetVertexBuffers(
-                0,
-                num_vbs as _,
-                self.vertex_buffer_views.as_ptr(),
-            );
+        let mut vertex_buffer_views = &self.vertex_buffer_views[..];
+        loop {
```
msiglreith: Logic looks sane, could you document it a bit if possible? A few lines would be totally enough!

src/backend/dx12/src/command.rs
```diff
@@ -1243,9 +1251,10 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         }
     }
 
-    fn bind_vertex_buffers(&mut self, vbs: pso::VertexBufferSet<Backend>) {
+    fn bind_vertex_buffers(&mut self, first_binding: u32, vbs: pso::VertexBufferSet<Backend>) {
         // Only cache the vertex buffer views as we don't know the stride (PSO).
-        for (&(buffer, offset), view) in vbs.0.iter().zip(self.vertex_buffer_views.iter_mut()) {
+        assert!(first_binding as usize <= MAX_VERTEX_BUFFERS);
```
msiglreith: `<` should be enough

src/backend/dx12/src/command.rs
msiglreith: (optionally: would be nice to avoid resetting all vertex buffers on each draw call. We didn't have it in the implementation so far.)

src/backend/dx12/src/command.rs
jbatez: Is it maybe enough to ignore range checks entirely and let the subslice indexing on the next line panic? I originally only added the range checking to keep the now-removed `vertex_buffers_first_slot` and `vertex_buffers_num_views` variables consistent with the `zip` behavior. If we're going to assert here, we should probably check `vbs.len()` as well which might be silently truncated (which was the behavior before this change).

kvark:
Thank you for addressing the issues! Please squash a few commits.

PR #1991, opened 2018-05-05
Add optional upper limit to subresources

AIOOB:
Thi is needed because Vulkan has a magic value to mean all remaining levels/layers in a SubresourceRange. This isn't a particularly nice solution, but especially since RangeBounds isn't stabilized yet, it was the only solution that I could think of. RangeBounds might be a sensible choice in the future, although to allow an easier transition to it an `(Option, Option)` might be better than what I have done here. It would probably have to involve dynamic dispatch but I'm not sure that would matter.
The other option is creating our own type with an optional end which we could then add a nice map operation to which would convert it to a Range.

PR checklist:

### DIFF
--- a/src/hal/src/range.rs
+++ b/src/hal/src/range.rs
@@ -57,3 +57,32 @@ impl<T> RangeArg<T> for (Option<T>, Option<T>) {
         self.1.as_ref()
     }
 }
+
+///
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct RangeOption<T> {
+    /// The optional lower bound of the range (inclusive)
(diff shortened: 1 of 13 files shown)


src/hal/src/range.rs
```diff
@@ -57,3 +57,32 @@ impl<T> RangeArg<T> for (Option<T>, Option<T>) {
         self.1.as_ref()
     }
 }
+
+///
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct RangeOption<T> {
```
kvark: how is this better than us just doing `Range<Option<T>>`?

AIOOB:
Closing in favour of gfx-rs/portability#73

Issue #1737, opened 2018-01-12
Modifying depth bounds


msiglreith:
Allow setting min/max depth bounds if supported the backend.

msiglreith:
yes, dx12 requires a newer update, but I dont see how CreatePipelineState would help here, There is a depth bounds command though. GL exposes an extension for it, se we can optionally expose the depth bounds feature. For the moment a no-op/error is enough for all backends expect vulkan.

jbatez:
It's not enough just to set the depth bounds; you need to enable the depth bounds test in the graphics pipeline state, and the pre-Creators Update `CreateGraphicsPipelineState` (as opposed to the newer, more flexible `CreatePipelineState`) doesn't provide the option. From the `OMSetDepthBounds` [documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/mt492658(v=vs.85).aspx):

> To use depth-bounds testing, the application must use the new CreatePipelineState method to enable depth-bounds testing on the PSO and then can use this command list method to change the depth-bounds dynamically.

msiglreith:
ugh, that sounds really annoying 😕 Apparently the new way of avoiding CreateGraphicsPipeline1/2/3/..
Thanks for mentioning, TIL.

kvark:
@jbatez are you saying that depth bounds are basically not supported in dx12 prior to the Creators Update? We may need to take that into account for the VkPI (by either introducing a flag to disable support, or requiring that Dx12 version).

kvark:
Sorry about hitting you with our jargon. VkPI stands short for Vulkan Portability Initiative, which drives our features/API requirements.
And you are correct, Vulkan does have it as a feature already (phew!) so we are good.

PR #1992, opened 2018-05-05
[ll] Depth bounds test

jbatez:
Stubs and DirectX 12 amd Vulkan implementations of #1737

PR checklist:
- [x] tested examples with the following backends: `dx12`, `vulkan`

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -112,6 +112,90 @@ pub struct UnboundImage {
     num_levels: image::Level,
 }
 
+#[repr(C)]
+struct PipelineStateSubobject<T> {
+    subobject_align: [usize; 0], // Subobjects must have the same alignment as pointers.
+    subobject_type: d3d12::D3D12_PIPELINE_STATE_SUBOBJECT_TYPE,
+    subobject: T,
+}
+
+impl<T> PipelineStateSubobject<T> {
+    fn new(subobject_type: d3d12::D3D12_PIPELINE_STATE_SUBOBJECT_TYPE, subobject: T) -> Self {
+        PipelineStateSubobject {
+            subobject_align: [],
+            subobject_type,
+            subobject,
+        }
+    }
+}
+
+#[repr(C)]
+struct GraphicsPipelineStateSubobjectStream {
+    root_signature: PipelineStateSubobject<*mut d3d12::ID3D12RootSignature>,
+    vs: PipelineStateSubobject<d3d12::D3D12_SHADER_BYTECODE>,
+    ps: PipelineStateSubobject<d3d12::D3D12_SHADER_BYTECODE>,
+    ds: PipelineStateSubobject<d3d12::D3D12_SHADER_BYTECODE>,
+    hs: PipelineStateSubobject<d3d12::D3D12_SHADER_BYTECODE>,
+    gs: PipelineStateSubobject<d3d12::D3D12_SHADER_BYTECODE>,
+    stream_output: PipelineStateSubobject<d3d12::D3D12_STREAM_OUTPUT_DESC>,
+    blend: PipelineStateSubobject<d3d12::D3D12_BLEND_DESC>,
+    sample_mask: PipelineStateSubobject<UINT>,
+    rasterizer: PipelineStateSubobject<d3d12::D3D12_RASTERIZER_DESC>,
+    depth_stencil: PipelineStateSubobject<d3d12::D3D12_DEPTH_STENCIL_DESC1>,
+    input_layout: PipelineStateSubobject<d3d12::D3D12_INPUT_LAYOUT_DESC>,
+    ib_strip_cut_value: PipelineStateSubobject<d3d12::D3D12_INDEX_BUFFER_STRIP_CUT_VALUE>,
+    primitive_topology: PipelineStateSubobject<d3d12::D3D12_PRIMITIVE_TOPOLOGY_TYPE>,
+    render_target_formats: PipelineStateSubobject<d3d12::D3D12_RT_FORMAT_ARRAY>,
+    depth_stencil_format: PipelineStateSubobject<dxgiformat::DXGI_FORMAT>,
+    sample_desc: PipelineStateSubobject<dxgitype::DXGI_SAMPLE_DESC>,
+    node_mask: PipelineStateSubobject<UINT>,
+    cached_pso: PipelineStateSubobject<d3d12::D3D12_CACHED_PIPELINE_STATE>,
+    flags: PipelineStateSubobject<d3d12::D3D12_PIPELINE_STATE_FLAGS>,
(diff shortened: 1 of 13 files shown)


jbatez:
Went ahead and added the `dx12` implementation as well. As discussed in #1737, I had to add a bunch of code to the pipeline state creation just to set `DepthBoundsTestEnable: TRUE`.

jbatez:
CreatePipelineState is returning E_INVALIDARG and I can't figure out why. :(

If anyone else wants to take a look it'd be much appreciated. :)

kvark:
@jbatez you'll get a hint by checking DX debug runtime output. I know 2 ways to get that:
  1. launch the app from visual studio (no need to have a project for sources, just the executable is needed)
  2. have [DebugView](https://docs.microsoft.com/en-us/sysinternals/downloads/debugview) running side by side and showing the messages

src/backend/dx12/src/lib.rs
```diff
@@ -752,6 +752,19 @@ impl hal::Instance for Instance {
                     mem::size_of::<d3d12::D3D12_FEATURE_DATA_ARCHITECTURE>() as _)
             });
 
+            let depth_bounds_test_supported = {
+                let mut features2: d3d12::D3D12_FEATURE_DATA_D3D12_OPTIONS2 = unsafe { mem::zeroed() };
+                if winerror::S_OK == unsafe {
```
kvark: let's move the function all out of the condition for clarity, we'll then be able to combine the check into `error == winerror::S_OK && features2.DepthBoundsTestSupported != 0`. Although, it depends on what other `winerror` things are expected here.

src/hal/src/command/raw.rs
```diff
@@ -292,6 +292,9 @@ pub trait RawCommandBuffer<B: Backend>: Clone + Any + Send + Sync {
     /// Set the blend constant values dynamically.
     fn set_blend_constants(&mut self, pso::ColorValue);
 
+    /// Set the depth bounds test values dynamically.
+    fn set_depth_bounds(&mut self, min: f32, max: f32);
```
kvark: how about `Range<f32>`?

jbatez:
@kvark Thanks for the DebugView tip! After hours of trial an error, I figured out the subobjects need to be aligned the same as pointers: a fact that's not documented anywhere as far as I can tell! I submitted MSDN documentation feedback 😛

src/backend/dx12/src/device.rs
```diff
@@ -1496,11 +1496,104 @@ impl d::Device<B> for Device {
 
         // Create PSO
         let mut pipeline = ptr::null_mut();
-        let hr = unsafe {
-            self.raw.clone().CreateGraphicsPipelineState(
-                &pso_desc,
-                &d3d12::IID_ID3D12PipelineState,
-                &mut pipeline as *mut *mut _ as *mut *mut _)
+        let hr = if desc.depth_stencil.unwrap_or(unsafe { mem::zeroed() }).depth_bounds {
```
kvark: nit: maps/unwrap_or is most useful when you have linear code. Once you need to process both branches, a simple `match` looks better (to me):
```rust
let hr = match desc.depth_stencil {
   Some(ds) if ds.depth_bounds => {...}
   _ => {...}
};
```

src/backend/dx12/src/device.rs
```diff
@@ -1496,11 +1496,104 @@ impl d::Device<B> for Device {
...
+            // The DepthBoundsTestEnable option isn't available in the original D3D12_GRAPHICS_PIPELINE_STATE_DESC struct.
+            // Instead, we must use the newer subobject stream method.
+            match self.raw.cast::<d3d12::ID3D12Device2>() {
+                Err(hr) => hr,
+                Ok(device2) => {
+                    #[repr(C)]
+                    struct Subobject<T> {
+                        subobject_align: [usize; 0], // Subobjects must have the same alignment as pointers.
```
kvark: great find! :clap:

src/backend/dx12/src/device.rs
```diff
@@ -1496,11 +1496,104 @@ impl d::Device<B> for Device {
...
+                        ib_strip_cut_value: Subobject<d3d12::D3D12_INDEX_BUFFER_STRIP_CUT_VALUE>,
+                        primitive_topology: Subobject<d3d12::D3D12_PRIMITIVE_TOPOLOGY_TYPE>,
+                        render_target_formats: Subobject<d3d12::D3D12_RT_FORMAT_ARRAY>,
+                        depth_stencil_format: Subobject<dxgiformat::DXGI_FORMAT>,
+                        sample_desc: Subobject<dxgitype::DXGI_SAMPLE_DESC>,
+                        node_mask: Subobject<UINT>,
+                        cached_pso: Subobject<d3d12::D3D12_CACHED_PIPELINE_STATE>,
+                        flags: Subobject<d3d12::D3D12_PIPELINE_STATE_FLAGS>,
+                    }
+                    let mut pss_stream = SubobjectStream {
```
kvark: this function body became quite heavy, with a lot of meat indented. Let's move the bulk of logic into a helper `create_pipeline_stream` method returning `SubobjectStream` (which also means moving the types out of the function body)

src/backend/dx12/src/device.rs
```diff
@@ -1512,14 +1605,19 @@ impl d::Device<B> for Device {
         if ds_destroy { destroy_shader(ds); }
 
         if winerror::SUCCEEDED(hr) {
+            let mut baked_states = desc.baked_states.clone();
+            if !desc.depth_stencil.unwrap_or(unsafe { mem::zeroed() }).depth_bounds {
```
kvark: similarly, let's avoid `unsafe` here, e.g. by doing `desc.depth_stencil.map(|ds| ds.depth_bounds).unwrap_or(false)`

src/backend/dx12/src/device.rs
jbatez: I thought about it but the `None` case is the same as the `false` case and I wasn't sure howto combine the two.

src/backend/dx12/src/device.rs
jbatez: Thanks! Still newish to Rust and didn't know about `Option::map`.

kvark:
Looks like all the concerns are addressed, thank you!
Let's make CI happy, remove the WIP part, and proceed.

jbatez:
@kvark Still waiting on new `ash` release.

kvark:
@MaikKlein can we have a new version published please?

MaikKlein:
@kvark https://crates.io/crates/ash/0.24.0

kvark:
Could you please merge the first and the last commit? That would make it so each commit is buildable on CI.

jbatez:
Commits squashed!

kvark:
Bors r+

PR #2001, opened 2018-05-08
[WIP] dx12 blitting support

msiglreith:
Working on dx12 blitting (2d color images atm only).
Loosely based on the metal implementation (using full screen triangle and viewport/scissors).

TODO:
- [x] Dynamically create descriptor heap
- [x] Push root constants
- [ ] Fix pre and post barriers
- [ ] Expose image feature flags
- [ ] Pass CTS

### DIFF
--- a/src/backend/dx12/shaders/blit.hlsl
+++ b/src/backend/dx12/shaders/blit.hlsl
@@ -1,0 +1,38 @@
+
+Texture2DArray blit_src : register(t0);
+SamplerState blit_sampler : register(s0);
+
+cbuffer region : register(b0) {
+    float2 src_start;
+    float2 src_end;
+    float z;
+};
+
+struct VsOutput {
+    float4 position : SV_POSITION;
+    float3 src_coords : TEXCOORD;
+};
+
+// Create a screen filling triangle
+VsOutput vs_blit_2d(uint id: SV_VertexID) {
+    VsOutput output;
+    switch (id) {
+        case 0:
+            output.position = float4(-1.0, -1.0, 0.0, 0.0);
+            output.src_coords = float3(src_start, z);
+            break;
+        case 1:
+            output.position = float4(0.0, 2.0, 0.0, 0.0);
+            output.src_coords = float3(src_start.x, 2 * src_end.y - src_start.y, z);
+            break;
+        default:
+            output.position = float4(2.0, 0.0, 0.0, 0.0);
+            output.src_coords = float3(2 * src_end.x - src_start.x, src_start.y, z);
+            break;
+    }
+    return output;
+}
+
+float4 ps_blit_2d(VsOutput input) : SV_TARGET0 {
+    return blit_src.Sample(blit_sampler, input.src_coords);
+}
--- a/src/backend/dx12/shaders/blit.hlsl
+++ b/src/backend/dx12/shaders/blit.hlsl
@@ -3,36 +3,27 @@ Texture2DArray blit_src : register(t0);
 SamplerState blit_sampler : register(s0);
 
 cbuffer region : register(b0) {
-    float2 src_start;
-    float2 src_end;
+    float2 offset;
+    float2 extent;
     float z;
+    float level;
 };
 
 struct VsOutput {
-    float4 position : SV_POSITION;
-    float3 src_coords : TEXCOORD;
+    float4 pos: SV_POSITION;
+    float4 uv: TEXCOORD0;
 };
 
 // Create a screen filling triangle
 VsOutput vs_blit_2d(uint id: SV_VertexID) {
-    VsOutput output;
-    switch (id) {
-        case 0:
-            output.position = float4(-1.0, -1.0, 0.0, 0.0);
-            output.src_coords = float3(src_start, z);
-            break;
-        case 1:
-            output.position = float4(0.0, 2.0, 0.0, 0.0);
-            output.src_coords = float3(src_start.x, 2 * src_end.y - src_start.y, z);
-            break;
-        default:
-            output.position = float4(2.0, 0.0, 0.0, 0.0);
-            output.src_coords = float3(2 * src_end.x - src_start.x, src_start.y, z);
-            break;
-    }
+    float2 coord = float2((id << 1) & 2, id & 2);
+    VsOutput output = {
+        float4(float2(-1.0, 1.0) + coord * float2(2.0, -2.0), 0.0, 1.0),
+        float4(offset + coord * extent, z, level)
+    };
     return output;
 }
 
-float4 ps_blit_2d(VsOutput input) : SV_TARGET0 {
-    return blit_src.Sample(blit_sampler, input.src_coords);
+float4 ps_blit_2d(VsOutput input) : SV_TARGET {
+    return blit_src.SampleLevel(blit_sampler, input.uv.xyz, input.uv.w);
 }
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -204,6 +204,11 @@ impl UserData {
     fn clear_dirty(&mut self, i: usize) {
         self.dirty_mask &= !(1 << i);
     }
+
+    /// Mark all entries as dirty.
+    fn dirty_all(&mut self) {
+        self.dirty_mask = !0;
+    }
 }
 
 #[derive(Clone)]
@@ -240,6 +245,8 @@ impl PipelineCache {
 enum BindPoint {
     Compute,
     Graphics,
+     // Internal pipelines used for blitting, copying, etc.
+    InternalGraphics,
 }
 
 #[derive(Clone)]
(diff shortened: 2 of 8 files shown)


kvark:
> Loosely based on the metal implementation (using full screen triangle and viewport/scissors).

FYI, Metal implementation doesn't use full screen triange with viewport/scissors for blitting. It issues actual rectangles.

src/backend/dx12/shaders/blit.hlsl
```diff
@@ -0,0 +1,38 @@
...
+    float3 src_coords : TEXCOORD;
+};
+
+// Create a screen filling triangle
+VsOutput vs_blit_2d(uint id: SV_VertexID) {
+    VsOutput output;
+    switch (id) {
+        case 0:
+            output.position = float4(-1.0, -1.0, 0.0, 0.0);
+            output.src_coords = float3(src_start, z);
```
kvark: that looks incorrect: your texture coordinates are going to be interpolated from the full-screen vertices, which is not desired

src/backend/dx12/shaders/blit.hlsl
```diff
@@ -0,0 +1,38 @@
...
+        default:
+            output.position = float4(2.0, 0.0, 0.0, 0.0);
+            output.src_coords = float3(2 * src_end.x - src_start.x, src_start.y, z);
+            break;
+    }
+    return output;
+}
+
+float4 ps_blit_2d(VsOutput input) : SV_TARGET0 {
+    return blit_src.Sample(blit_sampler, input.src_coords);
```
kvark: we should pick a fixed level here as well

src/backend/dx12/src/command.rs
```diff
@@ -344,6 +352,14 @@ impl CommandBuffer {
         self.vertex_buffer_views = [NULL_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS];
     }
 
+    // Indicates that the pipeline slot has been overriden with an internal pipeline.
+    //
+    // This only invalidates the slot and the user data!
+    fn set_internal_graphics_pipeline(&mut self) {
+        self.active_bindpoint = BindPoint::InternalGraphics;
+        self.gr_pipeline.user_data.dirty_all();
```
kvark: that's pretty harsh!

src/backend/dx12/src/command.rs
```diff
@@ -519,11 +535,23 @@ impl CommandBuffer {
...
-            self.active_bindpoint = BindPoint::Graphics;
+        match self.active_bindpoint {
+            BindPoint::Graphics => { }, // Nothing to do
+            BindPoint::Compute => {
+                // Switch to graphics bind point
+                let (pipeline, _) = self.gr_pipeline.pipeline.expect("No graphics pipeline bound");
+                unsafe { self.raw.SetPipelineState(pipeline); }
+                self.active_bindpoint = BindPoint::Graphics;
+            },
+            BindPoint::InternalGraphics => {
```
kvark: I wonder if we should just have a flag inside `BindPoint::Graphics { internal_signature: bool }` for merging it with the internal one

src/backend/dx12/src/command.rs
```diff
@@ -1213,17 +1244,128 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-        _filter: image::Filter,
-        _regions: T,
+        filter: image::Filter,
+        regions: T,
     ) where
         T: IntoIterator,
         T::Item: Borrow<com::ImageBlit>
     {
-        unimplemented!()
+        // TODO: only supporting 2D images
```
kvark: let's assert (or error + return) there for others

src/backend/dx12/src/command.rs
```diff
@@ -1213,17 +1244,128 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+
+        let rtv_pool = {
```
kvark: looks like it could be owned by the command pool?

src/backend/dx12/src/command.rs
```diff
@@ -1213,17 +1244,128 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+        };
+
+        for region in regions {
+            let r = region.borrow();
+
+            let first_layer = r.dst_subresource.layers.start;
+            let num_layers = r.dst_subresource.layers.end - first_layer;
+
+            let blit = match r.dst_subresource.aspects {
+                format::Aspects::COLOR => {
```
kvark: wow, I didn't know you can match the associated constants, TIL

src/backend/dx12/src/command.rs
```diff
@@ -1213,17 +1244,128 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    self.shared.service_pipes.get_blit_2d_color(dst.dxgi_format, filter)
+                },
+                _ => unimplemented!(),
+            };
+
+            self.set_internal_graphics_pipeline();
+
+            // TODO: support flipping?
+            let viewport = d3d12::D3D12_VIEWPORT {
+                TopLeftX: r.dst_bounds.start.x as _,
```
kvark: I would prefer us issuing real rectangles for blits (like we do in Metal), since it requires less state switching and can batch multiple blits together.
If you do want to use the viewport, then I don't think your shader even needs to know the destination bounds, it can write to -1 to 1 instead, and those will be scaled properly by the viewport you set here.

src/backend/dx12/src/command.rs
```diff
@@ -1213,17 +1244,128 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                TopLeftY: r.dst_bounds.start.y as _,
+                Width: (r.dst_bounds.end.x - r.dst_bounds.start.x) as _,
+                Height: (r.dst_bounds.end.y - r.dst_bounds.start.y) as _,
+                MinDepth: 0.0,
+                MaxDepth: 1.0,
+            };
+            let scissor = d3d12::D3D12_RECT {
```
kvark: ... and the scissor wouldn't be needed in this case

src/backend/dx12/src/command.rs
```diff
@@ -1213,17 +1244,128 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    num_layers as _,
+                    0,
+                    first_layer as _,
+                );
+            }
+        }
+
+        // Reset viewports and scissors
+        unsafe {
+            self.raw.RSSetViewports(
```
kvark: wouldn't this be covered by the dirty viewport/scissor flag we set already?

src/backend/dx12/src/command.rs
```diff
@@ -1213,17 +1244,128 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+
+            // Instanced full screen triangle
+            //
+            // Each instance corresponds to one blitting operation on one layer
+            unsafe {
+                self.raw.SetPipelineState(blit.pipeline.as_raw());
+                self.raw.SetGraphicsRootSignature(blit.signature.as_raw());
+                self.raw.RSSetViewports(1, &viewport);
+                self.raw.RSSetScissorRects(1, &scissor);
+                self.raw.DrawInstanced(
```
kvark: where do you select that each instance goes to a particular target layer?

src/backend/dx12/src/conv.rs
```diff
@@ -470,7 +470,8 @@ pub fn map_image_flags(usage: image::Usage) -> D3D12_RESOURCE_FLAGS {
     use self::image::Usage;
     let mut flags = D3D12_RESOURCE_FLAG_NONE;
 
-    if usage.contains(Usage::COLOR_ATTACHMENT) {
+    // Blit operations implemented via a graphics pipeline
```
kvark: actually, we should only do this if the format is renderable (yes, I'm missing this in Metal as well)

src/backend/dx12/src/internal.rs
```diff
@@ -0,0 +1,220 @@
...
+
+#[derive(Clone)]
+pub struct Blit {
+    pub pipeline: ComPtr<d3d12::ID3D12PipelineState>,
+    pub signature: ComPtr<d3d12::ID3D12RootSignature>,
+}
+
+pub struct ServicePipes {
+    device: ComPtr<d3d12::ID3D12Device>,
+    blits_2d_color: Mutex<HashMap<(dxgiformat::DXGI_FORMAT, d3d12::D3D12_FILTER), Blit>>,
```
kvark: nit: let's also define a typedef for this map key

src/backend/dx12/src/lib.rs
```diff
@@ -459,6 +460,13 @@ struct CmdSignatures {
     dispatch: ComPtr<d3d12::ID3D12CommandSignature>,
 }
 
+// Shared objects between command buffers, owned by the device.
+struct Shared {
+    pub signatures: CmdSignatures,
+    pub service_pipes: internal::ServicePipes,
+    pub device: *mut d3d12::ID3D12Device,
```
kvark: doesn't look like a device is needed here, given that `service_pipes` has one

src/backend/dx12/shaders/blit.hlsl
msiglreith: done

src/backend/dx12/src/command.rs
msiglreith: yes, temporarily owned. Waiting on renderdoc support to drop it immediately.

src/backend/dx12/src/command.rs
msiglreith: Need to rebase on top of it, currently based on an older revision

src/backend/dx12/src/command.rs
msiglreith: Fixed/reworked

src/backend/dx12/src/conv.rs
msiglreith: We don't expose the image feature if not supported.

kvark:
Closing in favor of #2028

PR #2002, opened 2018-05-08
[mtl] Copy buffer compute shader

kvark:
Fixes the case where size is not a multiple of 4
~~Based on (incomplete) #1995~~

PR checklist:

### DIFF
--- a/src/backend/metal/shaders/commands.metal
+++ b/src/backend/metal/shaders/commands.metal
@@ -61,14 +61,12 @@ fragment float4 ps_blit_3d(
 // -------------- Buffer Fill/Copy -------------- //
 
 kernel void cs_copy_buffer(
-    device uint *dest [[ buffer(0) ]],
-    device uint *source [[ buffer(1) ]],
+    device uchar *dest [[ buffer(0) ]],
+    device uchar *source [[ buffer(1) ]],
     constant uint &size [[ buffer(2) ]],
     uint index [[ thread_position_in_grid ]]
 ) {
     if (index < size) {
-        uint mask = index + 4 <= size ? ~0 : ((1 << (size - index) * 8) - 1);
-        uint wi = index / 4; // word index
-        dest[wi] = (source[wi] & mask) + (dest[wi] & ~mask);
+        dest[index] = source[index];
     }
 }
--- a/src/backend/metal/shaders/commands.metal
+++ b/src/backend/metal/shaders/commands.metal
@@ -1,25 +1,27 @@
 #include <metal_stdlib>
 using namespace metal;
 
+// -------------- Image Blits -------------- //
+
 typedef struct {
     float4 src_coords [[attribute(0)]];
     float4 dst_coords [[attribute(1)]];
-} TextureBlitAttributes;
+} BlitAttributes;
 
 typedef struct {
     float4 position [[position]];
     float4 uv;
     uint layer [[render_target_array_index]];
-} VertexData;
+} BlitVertexData;
 
-vertex VertexData vs_blit(TextureBlitAttributes in [[stage_in]]) {
+vertex BlitVertexData vs_blit(BlitAttributes in [[stage_in]]) {
     float4 pos = { 0.0, 0.0, 0.0f, 1.0f };
     pos.xy = in.dst_coords.xy * 2.0 - 1.0;
-    return VertexData { pos, in.src_coords, uint(in.dst_coords.z) };
+    return BlitVertexData { pos, in.src_coords, uint(in.dst_coords.z) };
 }
 
 fragment float4 ps_blit_1d(
-    VertexData in [[stage_in]],
+    BlitVertexData in [[stage_in]],
     texture1d<float> tex1D [[ texture(0) ]],
     sampler sampler2D [[ sampler(0) ]]
 ) {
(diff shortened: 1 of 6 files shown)


src/backend/metal/shaders/commands.metal
```diff
@@ -1,67 +1,84 @@
...
+kernel void cs_copy_buffer(
+    device uint *dest [[ buffer(0) ]],
+    device uint *source [[ buffer(1) ]],
+    constant uint &size [[ buffer(2) ]],
+    uint index [[ thread_position_in_grid ]]
+) {
+    if (index < size) {
+        uint mask = index + 4 <= size ? ~0 : ((1 << (size - index) * 8) - 1);
+        uint wi = index / 4; // word index
+        dest[wi] = (source[wi] & mask) + (dest[wi] & ~mask);
```
grovesNL: How do we know `dest[wi]` isn't out-of-bounds?

src/backend/metal/shaders/commands.metal
```diff
@@ -1,67 +1,84 @@
...
 }
+
+kernel void cs_copy_buffer(
+    device uint *dest [[ buffer(0) ]],
+    device uint *source [[ buffer(1) ]],
+    constant uint &size [[ buffer(2) ]],
+    uint index [[ thread_position_in_grid ]]
+) {
+    if (index < size) {
+        uint mask = index + 4 <= size ? ~0 : ((1 << (size - index) * 8) - 1);
```
grovesNL: What's the reason for the `- 1`?

src/backend/metal/src/command.rs
```diff
@@ -1963,14 +1963,78 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+            .unwrap()
+            .get_copy_buffer()
+            .to_owned();
+
+        let mut blit_commands = Vec::new();
+        let mut compute_commands = vec![
+            soft::ComputeCommand::BindPipeline(compute_pipe),
+        ];
+        let wg_size = MTLSize {
+            width: 64,
```
grovesNL: Could just use `self.shared.device.max_threads_per_threadgroup().width` for now I guess

src/backend/metal/src/command.rs
```diff
@@ -1963,14 +1963,78 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    src: src.raw.clone(),
+                    dst: dst.raw.clone(),
+                    region: r.clone(),
+                });
+            } else {
+                // not natively supported, going through compute shader
+                assert_eq!(0, r.size >> 32);
+                let word_count = (r.size + 3) / 4;
+                let copy_data = self.shared.device.new_buffer_with_data(
+                    &r.size as *const u64 as *const _,
```
grovesNL: Do we need to do `&(r.size as u32) as *const u32` here or can we assume this is safe?

src/backend/metal/shaders/commands.metal
kvark: if we need only 2 first bytes, the shift will give us `1<<16` and `-1` will produce `0x0000FFFF`

src/backend/metal/shaders/commands.metal
kvark: There is a size check two lines above. It can still write up to extra 3 bytes more than the size, but we at least guarantee to not change the contents. I don't know if we can do better here...

src/backend/metal/src/command.rs
kvark: We should get the optimal width from the [PSO itself](https://developer.apple.com/documentation/metal/mtlcomputepipelinestate/1414911-threadexecutionwidth?language=objc), but that's not critical

src/backend/metal/src/command.rs
kvark: I assume this is safe

grovesNL: Why don't we get a void pointer to a `u64`?

src/backend/metal/shaders/commands.metal
grovesNL: It seems like undefined behavior if we accidentally write outside the buffer, maybe the value wouldn't be written at all?

Is there a reason we don't want to copy per `uchar` instead of `uint`? If I understand correctly it also removes the need for the mask. We could always optimize this later if we're trying to reduce the amount of writes

src/backend/metal/shaders/commands.metal
grovesNL: I misread – makes sense

src/backend/metal/shaders/commands.metal
kvark: > It seems like undefined behavior if we accidentally write outside the buffer, maybe the value wouldn't be written at all?

Given that the offset has to be a multiple of 4, we should just round up the buffer sizes to the multiple of 4, which will guarantee that those copies never try to write OOB. I'll add this code.

> Is there a reason we don't want to copy per uchar instead of uint?

I looked it up previously and I don't think this is an option in MSL:
```
Ordinary Metal buffers may contain: 
  - Basic types such as float and int
  - Vector and matrix types
  - Arrays of buffer types
  - Structs of buffer types 
  - Unions of buffer types 
```

src/backend/metal/src/command.rs
kvark: let's do it, yeah

kvark:
@grovesNL concerns are addressed. Buffer length was already aligned to 256 (lol) so no overwrite is possible (but a new assert is in place).

src/backend/metal/src/command.rs
grovesNL: The buffer is supposed to be `u32` isn't it? I thought it would be:
```rust
&(r.size as u32) as *const u32 as *const _,
mem::size_of::<u32>() as _
```

kvark:
Note: I confirmed the buffer-to-buffer tests are all green in CTS :tada:

PR #2003, opened 2018-05-09
[ll] layer range for clear_attachments

fkaa:
Replaces the `Rect` in `clear_attachments` with `ClearRect` to allow for clearing layers in an attachment. Only implemented for vulkan & dx12 for now.

PR checklist:
- [x] tested examples with the following backends: vulkan, dx12

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -16,7 +16,7 @@ use winapi::shared::dxgiformat;
 
 use wio::com::ComPtr;
 
-use {conv, internal, native as n, Backend, Device, Shared, MAX_VERTEX_BUFFERS};
+use {conv, device, internal, native as n, Backend, Device, Shared, MAX_VERTEX_BUFFERS};
 use device::ViewInfo;
 use root_constants::RootConstant;
 use smallvec::SmallVec;
@@ -1204,15 +1204,25 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         T: IntoIterator,
         T::Item: Borrow<com::AttachmentClear>,
         U: IntoIterator,
-        U::Item: Borrow<pso::Rect>,
+        U::Item: Borrow<pso::ClearRect>,
     {
-        assert!(self.pass_cache.is_some(), "`clear_attachments` can only be called inside a renderpass");
-        let rects: SmallVec<[d3d12::D3D12_RECT; 16]> = rects.into_iter().map(|rect| get_rect(rect.borrow())).collect();
+        assert!(
+            self.pass_cache.is_some(),
+            "`clear_attachments` can only be called inside a renderpass"
+        );
+
+        let clear_rects: SmallVec<[pso::ClearRect; 16]> = rects
+            .into_iter()
+            .map(|rect| rect.borrow().clone())
+            .collect();
+
+        let mut device = self.shared.service_pipes.device.clone();
+
         for clear in clears {
             let clear = clear.borrow();
             match *clear {
                 com::AttachmentClear::Color(index, cv) => {
-                    let rtv = {
+                    let attachment = {
                         let pass_cache = self.pass_cache.as_ref().unwrap();
                         let rtv_id = pass_cache
                             .render_pass
@@ -1220,18 +1230,50 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                             .color_attachments[index]
                             .0;
 
-                        pass_cache
-                            .framebuffer
-                            .attachments[rtv_id]
-                            .handle_rtv
-                            .unwrap()
+                        pass_cache.framebuffer.attachments[rtv_id]
                     };
 
-                    self.clear_render_target_view(
-                        rtv,
-                        cv.into(),
-                        &rects,
-                    );
+                    let mut rtv_pool = n::DescriptorCpuPool {
+                        heap: Device::create_descriptor_heap_impl(
+                            &mut device.clone(),
+                            d3d12::D3D12_DESCRIPTOR_HEAP_TYPE_RTV,
+                            false,
+                            clear_rects.len()
+                        ),
+                        offset: 0,
+                        size: 0,
+                        max_size: clear_rects.len() as _
+                    };
+
+                    self.rtv_pools.push(rtv_pool.heap.raw.clone());
+
+                    for clear_rect in &clear_rects {
+                        let rect = [get_rect(&clear_rect.rect)];
+
+                        let view_info = device::ViewInfo {
+                            resource: attachment.resource,
+                            kind: attachment.kind,
+                            flags: image::StorageFlags::empty(),
+                            view_kind: image::ViewKind::D2Array,
+                            format: attachment.dxgi_format,
+                            range: image::SubresourceRange {
+                                aspects: Aspects::COLOR,
+                                levels: 0 .. 1,
+                                layers: clear_rect.layers.clone()
+                            }
+                        };
+                        let rtv = Device::view_image_as_render_target_impl(
+                            &mut device,
+                            &mut rtv_pool,
+                            view_info
+                        ).unwrap();
+
+                        self.clear_render_target_view(
+                            rtv,
+                            cv.into(),
+                            &rect,
+                        );
+                    }
                 }
                 _ => unimplemented!(),
             }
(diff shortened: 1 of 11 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -254,6 +254,7 @@ struct Copy {
 
 #[derive(Clone)]
 pub struct CommandBuffer {
+    device: ComPtr<d3d12::ID3D12Device>,
```
kvark: Note: this collides with #2001, which has a little more comprehensive approach to device pointer sharing
cc @msiglreith

src/backend/dx12/src/command.rs
```diff
@@ -1147,34 +1150,70 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
                             .subpasses[self.cur_subpass]
                             .color_attachments[index]
                             .0;
 
-                        pass_cache
-                            .framebuffer
-                            .attachments[rtv_id]
-                            .handle_rtv
-                            .unwrap()
+                        let attachment =
```
kvark: nit: no need for this assignment / temporary variable

src/backend/dx12/src/command.rs
```diff
@@ -1147,34 +1150,70 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+
+                        attachment
                     };
 
-                    self.clear_render_target_view(
-                        rtv,
-                        cv.into(),
-                        &rects,
-                    );
+                    let mut rtv_pool = n::DescriptorCpuPool {
```
kvark: where is it getting freed? how do we ensure it's alive for the lifetime of the command buffer?
(@msiglreith correct me if this is not required here)

src/backend/dx12/src/command.rs
```diff
@@ -1147,34 +1150,70 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                            view_kind: image::ViewKind::D2Array,
+                            format: attachment.dxgi_format,
+                            range: image::SubresourceRange {
+                                aspects: Aspects::COLOR,
+                                levels: 0 .. 1,
+                                layers: (clear_rect.base as _) .. (clear_rect.base +
+                                                                   clear_rect.layers) as _
+                            }
+                        };
+                        let rtv = Device::view_image_as_render_target_impl(&mut
```
kvark: nit: formatting/indentation

src/backend/dx12/src/device.rs
```diff
@@ -73,13 +73,13 @@ fn shader_bytecode(shader: *mut d3dcommon::ID3DBlob) -> d3d12::D3D12_SHADER_BYTE
...
 #[derive(Clone)]
-struct ViewInfo {
-    resource: *mut d3d12::ID3D12Resource,
-    kind: image::Kind,
-    flags: image::StorageFlags,
-    view_kind: image::ViewKind,
-    format: dxgiformat::DXGI_FORMAT,
-    range: image::SubresourceRange,
+pub(crate) struct ViewInfo {
+    pub(crate) resource: *mut d3d12::ID3D12Resource,
```
kvark: given that this is used in the command buffer module now, I feel that we more and more collide with that other PR. Need to see which parts work good, and how to avoid sharing too much between those modules

src/backend/dx12/src/command.rs
```diff
@@ -1147,34 +1150,70 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                        ),
+                        offset: 0,
+                        size: 0,
+                        max_size: 16
+                    };
+
+                    for clear_rect in &clear_rects {
+                        let rect = [get_rect(&clear_rect.rect)];
+
+                        let view_info = ::device::ViewInfo {
```
kvark: nit: let's not use global namespaces

src/backend/dx12/src/native.rs
```diff
@@ -215,6 +215,8 @@ pub struct ImageView {
     pub(crate) num_levels: image::Level,
     pub(crate) mip_levels: (image::Level, image::Level),
     pub(crate) layers: (image::Layer, image::Layer),
+
```
kvark: nit: empty line is a bit out of place here

src/hal/src/pso/graphics.rs
```diff
@@ -22,6 +22,18 @@ pub struct Rect {
...
 }
 
+/// A simple struct describing a rect with integer coordinates.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct ClearRect {
+    /// 2D region.
+    pub rect: Rect,
+    /// Base layer.
+    pub base: u32,
```
kvark: let's have `layers: Range<image::Layer>` here

src/backend/dx12/src/command.rs
msiglreith: > how do we ensure it's alive for the lifetime of the command buffer

AFAIK cpu descriptors don't need to live for the whole command buffer lifetime, just for encoding of the commands but don't quote me on this.

fkaa: Found this on msdn, but the current code seems to work for me at least
>  A handle itself can be copied freely or passed into APIs that operate on/use descriptors. There is no ref counting, so the application must ensure that it does not use a handle after the underlying descriptor heap has been deleted. 

https://msdn.microsoft.com/en-us/library/windows/desktop/dn859359(v=vs.85).aspx#Descriptor_Handles

kvark: So it's a matter of what "use the handle" means exactly. Does "use" last for the execution of the command buffer on GPU? Or does it end when we issue the API calls?

msiglreith: https://msdn.microsoft.com/en-us/library/windows/desktop/dn899199(v=vs.85).aspx

> After recording into the command list (with a call such as OMSetRenderTargets, for example) the memory used to hold the descriptors for this call is immediately available for re-use after the call.

src/backend/dx12/src/command.rs
```diff
@@ -1147,34 +1150,66 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    let mut rtv_pool = n::DescriptorCpuPool {
+                        heap: Device::create_descriptor_heap_impl(
+                            &mut self.device.clone(),
+                            d3d12::D3D12_DESCRIPTOR_HEAP_TYPE_RTV,
+                            false,
+                            16
+                        ),
+                        offset: 0,
+                        size: 0,
+                        max_size: 16
```
kvark: what happens if we have more than 16 rectangles to clear here?

src/backend/dx12/src/command.rs
```diff
@@ -307,11 +308,13 @@ enum BarrierPoint {
 
 impl CommandBuffer {
     pub(crate) fn new(
+        device: ComPtr<d3d12::ID3D12Device>,
```
kvark: Is COM API thread-safe? Is `ID3D12Device` thread-safe?

src/backend/dx12/src/command.rs
```diff
@@ -1204,34 +1204,68 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                            flags: image::StorageFlags::empty(),
+                            view_kind: image::ViewKind::D2Array,
+                            format: attachment.dxgi_format,
+                            range: image::SubresourceRange {
+                                aspects: Aspects::COLOR,
+                                levels: 0 .. 1,
+                                layers: clear_rect.layers.clone()
+                            }
+                        };
+                        let rtv = Device::view_image_as_render_target_impl(
```
kvark: do we need this one? images already have the `clear_rtv` arrays, which we could technically iterate

src/backend/dx12/src/command.rs
msiglreith: `self.rtv_pools` as they will/can be removed. renderdoc nightly works without storing those.
`temporary_gpu_heaps` are required to be linked with the lifetime of the command buffer.

kvark: sorry, I incorrectly recalled which one here, thanks for correction!

kvark:
bors delegate+

fkaa:
bors r=kvark

Issue #1984, opened 2018-05-03
Revise Metal thread safety


kvark:
In light of #1976, we need to go through all the `unsafe impl Send+Sync` entries and verify if they make sense. In general, ObjC doesn't guarantee thread safety, and Apple documentation is explicit about which Cocoa objects can be used from multiple threads: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html

As for Metal, the only documented thread-safe object appears to be `MTLCommandQueue` (so far). This is different from Vulkan, where `VkDevice` can be shared concurrently to create resources from multiple threads.

omni-viral:
To follow Vulkan API all `unsafe impl Send + Sync` must stay.
But implementations must be synchronized manually where needed.

PR #2004, opened 2018-05-10
[mtl] Improved blitting

### LINKED ISSUE #2006: Metal depth24stencil8 format
We currently disable this format due to a validation error. Need to investigate what's going wrong and re-enable it. Perhaps, it's just the copy image that doesn't like it.
See https://github.com/gfx-rs/gfx/pull/2004/files#r187233484

kvark:
Fixes #2006
(note: also disables d32s8 for now)

Brings the following goodies:
  - actually working blits. I was missing the Y-flip in rendering, which I encoded into the viewport now.
  - proper fast-path for copy_image. I screwed the src/dst arguments...
  - basic integer support for blit shaders (to be extended)
  - retained textures
  - non-equal formats support for copy_image (not verified by CTS yet)

Overall with this change, I was able to get through "api.copy_and_blit.core.blit_image.all_formats.color" section with the following stats:
  - 3538 successes
  - 29 failures (!)
  - 15 crashes

The failures/crashes don't appear systematic, may be caused by #1984

PR checklist:

### DIFF
--- a/src/backend/metal/shaders/commands.metal
+++ b/src/backend/metal/shaders/commands.metal
@@ -20,7 +20,7 @@ vertex BlitVertexData vs_blit(BlitAttributes in [[stage_in]]) {
     return BlitVertexData { pos, in.src_coords, uint(in.dst_coords.z) };
 }
 
-fragment float4 ps_blit_1d(
+fragment float4 ps_blit_1d_float(
     BlitVertexData in [[stage_in]],
     texture1d<float> tex1D [[ texture(0) ]],
     sampler sampler2D [[ sampler(0) ]]
@@ -28,7 +28,7 @@ fragment float4 ps_blit_1d(
   return tex1D.sample(sampler2D, in.uv.x);
 }
 
-fragment float4 ps_blit_1d_array(
+fragment float4 ps_blit_1d_array_float(
     BlitVertexData in [[stage_in]],
     texture1d_array<float> tex1DArray [[ texture(0) ]],
     sampler sampler2D [[ sampler(0) ]]
@@ -36,7 +36,7 @@ fragment float4 ps_blit_1d_array(
   return tex1DArray.sample(sampler2D, in.uv.x, uint(in.uv.z));
 }
 
-fragment float4 ps_blit_2d(
+fragment float4 ps_blit_2d_float(
     BlitVertexData in [[stage_in]],
     texture2d<float> tex2D [[ texture(0) ]],
     sampler sampler2D [[ sampler(0) ]]
@@ -44,15 +44,30 @@ fragment float4 ps_blit_2d(
   return tex2D.sample(sampler2D, in.uv.xy, level(in.uv.w));
 }
 
-fragment float4 ps_blit_2d_array(
+fragment uint4 ps_blit_2d_uint(
+    BlitVertexData in [[stage_in]],
+    texture2d<uint> tex2D [[ texture(0) ]],
+    sampler sampler2D [[ sampler(0) ]]
+) {
+  return tex2D.sample(sampler2D, in.uv.xy, level(in.uv.w));
+}
+
+fragment int4 ps_blit_2d_int(
+    BlitVertexData in [[stage_in]],
+    texture2d<int> tex2D [[ texture(0) ]],
+    sampler sampler2D [[ sampler(0) ]]
+) {
+  return tex2D.sample(sampler2D, in.uv.xy, level(in.uv.w));
+}
+
+fragment float4 ps_blit_2d_array_float(
     BlitVertexData in [[stage_in]],
     texture2d_array<float> tex2DArray [[ texture(0) ]],
     sampler sampler2D [[ sampler(0) ]]
 ) {
   return tex2DArray.sample(sampler2D, in.uv.xy, uint(in.uv.z), level(in.uv.w));
 }
-
-fragment float4 ps_blit_3d(
+fragment float4 ps_blit_3d_float(
     BlitVertexData in [[stage_in]],
     texture3d<float> tex3D [[ texture(0) ]],
     sampler sampler2D [[ sampler(0) ]]
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1356,8 +1356,9 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             let mut pipes = self.shared.service_pipes
                 .lock()
                 .unwrap();
+            let key = (dst.mtl_type, dst.mtl_format, dst.blit_channel);
             let pso = pipes
-                .get_blit_image(dst.mtl_type, dst.mtl_format, &self.shared.device)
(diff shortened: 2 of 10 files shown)


src/backend/metal/src/command.rs
```diff
@@ -1976,14 +1969,28 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         T: IntoIterator,
         T::Item: Borrow<com::ImageCopy>,
     {
+        let inner = unsafe { &mut *self.inner.get() };
```
grovesNL: Maybe we should just have a function to do this? We have this a few times now

src/backend/metal/shaders/commands.metal
```diff
@@ -20,38 +20,54 @@ vertex BlitVertexData vs_blit(BlitAttributes in [[stage_in]]) {
     return BlitVertexData { pos, in.src_coords, uint(in.dst_coords.z) };
 }
 
-fragment float4 ps_blit_1d(
+fragment float4 ps_blit_1d_float(
```
grovesNL: Looks like we are going to have a lot of shader combinations when we expand to 8-bits, 16-bits etc.. should we eventually attempt to generate these during build or something?

src/backend/metal/src/conversions.rs
```diff
@@ -52,7 +52,7 @@ pub fn map_format(format: Format) -> Option<MTLPixelFormat> {
         f::Rgba32Float       => RGBA32Float,
         f::D16Unorm          => Depth16Unorm,
         f::D32Float          => Depth32Float,
-        f::D24UnormS8Uint    => Depth24Unorm_Stencil8,
+        //f::D24UnormS8Uint    => Depth24Unorm_Stencil8, // !macOS
```
grovesNL: The [docs state macOS 10.11+](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatdepth24unorm_stencil8), is that wrong?

src/backend/metal/shaders/commands.metal
kvark: The more shaders the less incentive we have to load all of them at build time. Currently, all blit shaders are compiled at build time but the pipelines are lazily created. I think we are safe to continue doing so.

> when we expand to 8-bits, 16-bits etc.. 

This is not necessary. Bit sizes don't matter, only the texture dimensions and blit channels do matter. Plus the depth/stencil thing that is still TODO.

src/backend/metal/src/command.rs
kvark: uh, I approached this and now regret it - the changes are blowing out of control
perhaps, we can proceed and I'll follow up?

src/backend/metal/src/conversions.rs
kvark: Interesting. I'm getting the validation error: pixel format 255 is invalid. Will follow-up in #2006

src/backend/metal/shaders/commands.metal
grovesNL: Yeah, I mean to generate some of the repetitive actual MSL at build time and compile that, then lazily creating pipelines as we do already. Though it's probably overkill for now because we don't that many variations yet.

>This is not necessary. Bit sizes don't matter, only the texture dimensions and blit channels do matter. Plus the depth/stencil thing that is still TODO.

Don't we still have the issue of out-of-bounds writes if we only have `uint`, `int` and `float`? Because even with masking we still can't fill the last element correctly if it's not aligned to 4 bytes (in the existing shaders). For example, if we want to do a copy of 8-bit texture formats like R8, don't we need other shaders to sample `uchar`?

kvark:
@grovesNL please have a look at the last commit

src/backend/metal/src/conversions.rs
```diff
@@ -1,114 +1,118 @@
...
+impl PrivateCapabilities {
+    pub fn map_format(&self, format: Format) -> Option<MTLPixelFormat> {
+        use metal::MTLPixelFormat::*;
+        use hal::format::Format as f;
+        Some(match format {
+            f::B5g6r5Unorm   if self.format_b5 => B5G6R5Unorm,
+            f::B5g5r5a1Unorm if self.format_b5 => BGR5A1Unorm,
+            f::R8Srgb  if self.format_min_srgb_channels <= 1 => R8Unorm_sRGB,
+            f::Rg8Srgb if self.format_min_srgb_channels <= 2 => RG8Unorm_sRGB,
+            f::D24UnormS8Uint if self.format_depth24_stencil8 => Depth24Unorm_Stencil8,
```
grovesNL: I was trying to keep the order consistent with `format.rs` because it was a bit hard to spot mismatches

src/backend/metal/src/device.rs
```diff
@@ -135,12 +127,31 @@ impl Drop for Device {
...
+        let private_caps = PrivateCapabilities {
+            resource_heaps: Self::supports_any(&raw, RESOURCE_HEAP_SUPPORT),
+            argument_buffers: Self::supports_any(&raw, ARGUMENT_BUFFER_SUPPORT) && false, //TODO
+            format_depth24_stencil8: raw.d24_s8_supported(),
+            format_depth32_stencil8: false, //TODO: crashing the Metal validation layer upon copying from buffer
+            format_min_srgb_channels: if Self::is_mac(&raw) {4} else {1},
+            format_b5: !Self::is_mac(&raw),
+            max_buffers_per_stage: 31,
+            max_textures_per_stage: if Self::is_mac(&raw) {128} else {31},
+            max_samplers_per_stage: 31,
```
grovesNL: Should this be 16?

src/backend/metal/shaders/commands.metal
kvark: I agree there is a bit of redundancy, and it could technically be auto-generated. Let's see if it's going to be actually needed.

> For example, if we want to do a copy of 8-bit texture formats like R8, don't we need other shaders to sample uchar?

Well, for one, I have Vulkan CTS backing up this code, so we are doing fine there :)
My understanding is that the logic of converting from uchar to uint in this case lies on the hands of HW texture sampler. There is no need to generate a different pipeline/shader between u8 and u16 textures.

As for the render targets, the micro-code for converting from shader-output float4/uint4/int4 to the actual format is compiled at the pipeline creation, which knows about the target formats, and that's for we have a pipeline hash map here.

src/backend/metal/src/device.rs
kvark: nicely spotted! fixed now

kvark:
bors r=grovesNL

src/backend/metal/shaders/commands.metal
grovesNL: > Well, for one, I have Vulkan CTS backing up this code, so we are doing fine there :)
My understanding is that the logic of converting from uchar to uint in this case lies on the hands of HW texture sampler. There is no need to generate a different pipeline/shader between u8 and u16 textures.

Ok, interesting. It just seems strange to me that we need to distinguish `uint` and `int` but not between various sizes.

PR #2008, opened 2018-05-10
Remove UnsafeCell from Metal

kvark:
~~Only the last commit matters. Eventually to be rebased on #2004.~~

Greatly reduces the amount of unsafety we have in the backend, at the cost of a bit of extra verbosity.

The idea previously was that since the API isn't valid for accessing the same command buffer concurrently, we don't need to check for that at run-time. I believe it's a wishful thinking, and a footgun for possible unexpected issues.
Unless proven otherwise (by benchmarks), we should stick to safe Rust and be kept in warm and safety by the combination of type (where possible) and run-time checks.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -2,8 +2,8 @@ use {AutoreleasePool, Backend};
 use {native, window};
 use internal::{BlitVertex, ServicePipes};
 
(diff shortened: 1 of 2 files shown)


kvark:
bors r=grovesNL

PR #2010, opened 2018-05-11
[mtl] fixed borrowing and pooling of command buffers

kvark:
Enforces the queue pool counts.

Fixes the borrowed command buffer release strategy: instead of releasing the counts at the commission time, we are now automatically adding a competition handler to release the counter. The cases where we wait for GPU instantly are specifically handled in order to avoid hanging on the queue pool mutex.
Note: previously we could hang on trying to acquire a borrowed command buffer, since the code thinks they are instantly freed but the metal runtime only gets more available when they are done, so this was a rather dangerous mistake on our part.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -49,32 +49,51 @@ impl<'a> Deref for CommandBufferScope<'a> {
 
 impl QueuePool {
     fn find_queue(&mut self, device: &metal::DeviceRef) -> usize {
+        const POOL_SIZE: usize = 64;
         self.queues
             .iter()
             .position(|q| q.reserve.start != q.reserve.end)
             .unwrap_or_else(|| {
                 let queue = QueueInner {
-                    queue: device.new_command_queue(),
-                    reserve: 0 .. 64, //TODO
+                    queue: device.new_command_queue_with_max_command_buffer_count(POOL_SIZE as _),
+                    reserve: 0 .. POOL_SIZE,
                 };
                 self.queues.push(queue);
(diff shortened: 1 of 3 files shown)


kvark:
Note: it is tempting to abstract away the borrowed buffer scope that we know in advance needs to be waited upon (so that we avoid adding a handler and can use it for `wait_idle`). This is complicated by the use-case that our mapping invalidation does, which has a run-time condition on committing + waiting on that borrowed buffer. If we add a run-time flag inside `CommandBufferScope` we have no means to enforce the fact that user hasn't actually committed the command buffer manually prior to discarding it...
Thus, I had to through that idea out, and the responsibility to free the buffer is kept to the invalidation code, plus the `wait_idle` case are handled specifically.

src/backend/metal/src/command.rs
```diff
@@ -49,32 +49,51 @@ impl<'a> Deref for CommandBufferScope<'a> {
 
 impl QueuePool {
     fn find_queue(&mut self, device: &metal::DeviceRef) -> usize {
+        const POOL_SIZE: usize = 64;
```
grovesNL: I guess we could use a much higher value here if we'd like

kvark:
bors r=grovesNL

src/backend/metal/src/command.rs
grovesNL: Fair enough, at least in the future we should be able to safely lift it on macOS (any version) or iOS/tvOS >= 11.3 (11.3 contained the bug fix)

PR #2014, opened 2018-05-11
mtl: Temporarily remove misaligned fill buffer test

grovesNL:
Simply removes the misaligned fill buffer test until:

- we determine how to handle this on Metal (misaligned `copyFromBuffer`)
- we get clarification on whether this Vulkan restriction will be lifted (i.e. we know of misbehaving drivers already which appear to perform partial writes for misaligned buffers lengths)

PR checklist:


grovesNL:
Based on the changes in #2015 we shouldn't need to disable this anymore - closing

PR #2015, opened 2018-05-11
[mtl] Only invalidate/flush on memory map/unmap for coherent types, fix size alignment

kvark:
Fixes #2014

PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -278,6 +278,10 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
             framebuffer_color_samples_count: 0b101, // TODO
             framebuffer_depth_samples_count: 0b101, // TODO
             framebuffer_stencil_samples_count: 0b101, // TODO
+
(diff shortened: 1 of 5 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1001,14 +1015,20 @@ impl hal::Device<Backend> for Device {
...
                 command::exec_blit(encoder, &soft::BlitCommand::CopyBuffer {
                     src: cpu_buffer.clone(),
                     dst,
                     region: BufferCopy {
-                        src: start,
-                        dst: start - alloc_range.start,
-                        size: end - start,
+                        src: start & !ALIGNM_MASK,
+                        dst: other,
+                        size: size & !ALIGNM_MASK,
```
grovesNL: Minor but should probably prefer `end & !ALIGNM_MASK - start & !ALIGNM_MASK` here (and below) instead

kvark:
Thanks! fixed now
bors r=grovesNL

PR #2016, opened 2018-05-12
Out-of-pass image clears

### LINKED ISSUE #2007: Clear color/depth image API
We need to adjust the clear_color_image and clear_depth_stencil_image APIs to better match Vulkan side and actually implement those on the backends. The lack of it right now contribute to most Vulkan CTS panics.

kvark:
Fixes #2007
Depends on https://github.com/gfx-rs/metal-rs/pull/49
With this PR as well as the upcoming portability changes, we can go through the whole `api.image_clearing` section (on Metal) without crashes. 143 passes and 216 failures at the moment, out of 1027 tests in total.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1234,7 +1234,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
 
         let clear_color = unsafe {
             match image.shader_channel {
-                Channel::Float =>metal::MTLClearColor::new(
+                Channel::Float => metal::MTLClearColor::new(
                     color.float32[0] as _,
                     color.float32[1] as _,
                     color.float32[2] as _,
@@ -1271,7 +1271,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             for level in sub.levels.start .. end_level {
                 for layer in sub.layers.start .. end_layer {
                     let descriptor = metal::RenderPassDescriptor::new().to_owned();
-                    // descriptor.set_render_target_array_length(sub.layers.end as _); //TODO: fast patg
+                    // descriptor.set_render_target_array_length(sub.layers.end as _); //TODO: fast path
                     if sub.aspects.contains(Aspects::COLOR) {
                         let attachment = descriptor
                             .color_attachments()
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1257,9 +1257,19 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
 
         for subresource_range in subresource_ranges {
             let sub = subresource_range.borrow();
+            let end_level = if sub.levels.end == !0 {
+                image.raw.mipmap_level_count() as _
+            } else {
+                sub.levels.end
+            };
+            let end_layer = if sub.layers.end == !0 {
+                image.raw.array_length() as _
+            } else {
+                sub.layers.end
+            };
 
-            for level in sub.levels.clone() {
-                for layer in sub.layers.clone() {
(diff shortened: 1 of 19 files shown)


kvark:
This should be ready now :tada: 
The only thing I'm not 100% sure about is the clear shader/pipes on Metal that don't appear to be needed yet. I don't want to remove them, since I have a feeling we'll need them eventually.

src/backend/metal/src/command.rs
```diff
@@ -1191,24 +1218,101 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-        &mut self,
-        _image: &native::Image,
-        _layout: Layout,
-        _range: SubresourceRange,
-        _value: com::ClearDepthStencilRaw,
-    ) {
-        unimplemented!()
+        let clear_color = unsafe {
+            match image.shader_channel {
+                Channel::Float =>metal::MTLClearColor::new(
```
grovesNL: nit: space

src/backend/metal/src/internal.rs
```diff
@@ -6,26 +6,45 @@ use std::collections::HashMap;
...
-pub enum BlitChannel {
+pub enum Channel {
     Float,
     Int,
     Uint,
 }
 
-pub type BlitKey = (metal::MTLTextureType, metal::MTLPixelFormat, BlitChannel);
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub struct ClearMask {
```
grovesNL: Should these (and others) be `pub(crate)`?

src/backend/metal/src/command.rs
```diff
@@ -1191,24 +1218,101 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+            let end_layer = if sub.layers.end == !0 {
+                image.raw.array_length() as _
+            } else {
+                sub.layers.end
+            };
+
+            for level in sub.levels.start .. end_level {
+                for layer in sub.layers.start .. end_layer {
+                    let descriptor = metal::RenderPassDescriptor::new().to_owned();
+                    // descriptor.set_render_target_array_length(sub.layers.end as _); //TODO: fast patg
```
grovesNL: nit: typo

src/backend/metal/src/command.rs
```diff
@@ -1191,24 +1218,101 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+            };
+            let end_layer = if sub.layers.end == !0 {
+                image.raw.array_length() as _
+            } else {
+                sub.layers.end
+            };
+
+            for level in sub.levels.start .. end_level {
+                for layer in sub.layers.start .. end_layer {
+                    let descriptor = metal::RenderPassDescriptor::new().to_owned();
```
grovesNL: IMO we may want to remove this function from metal-rs – it seems a bit sketchy because it creates [`MTLRenderPassDescriptorInternal`](https://github.com/gfx-rs/metal-rs/blob/master/src/renderpass.rs#L292) directly. There is a [`renderPassDescriptor`](https://developer.apple.com/documentation/metal/mtlrenderpassdescriptor/1437979-renderpassdescriptor) convenience function that should set up the default attachments that we use here though.

kvark: good point! filed https://github.com/gfx-rs/metal-rs/issues/50

src/backend/metal/src/internal.rs
kvark: I often see this over-used and noisy. In our case, the whole module is private, so unless `lib.rs` re-exports it explicitly, it stays hidden within the crate.

PR #2017, opened 2018-05-13
[mtl] force Store mode on depth/stencil clears

kvark:
Apparently the store mode is DontCare *by default* on depth/stencil. This change doesn't let us pass more tests though - something is broken up in multiple spots (no sure who to blame yet):
  - clearing multiple ranges doesn't work with UINT/INT (go figure what it has to do with multiple ranges...)
  - clearing 1D texture doesn't work with any formats above 16 bits (e.g. RGBA8, RG16, etc), except for sRGB ones (we do have a hack in place to treat 1D textures as 2D if mipmaps are requested, but still unclear how is this related with the observed limitations)

Note: CTS has a lot of magic in it, and maybe it's simply broken (that already happened a few times), or there is something wonky going on around transfers/mapping that are required to check the results (even if clears are fine).

PR checklist:
- [ ] `make reftests` succeeds (Nope, it actually hangs :( right after the beta 5 update)
- [] tested examples with the following backends:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -496,27 +496,27 @@ impl CommandSink {
 
     fn quick_render_pass<I>(
         &mut self,
-        descriptor: metal::RenderPassDescriptor,
+        descriptor: &metal::RenderPassDescriptorRef,
         commands: I,
     ) where
         I: IntoIterator<Item = soft::RenderCommand>,
     {
+        self.stop_encoding();
+
         match *self {
-            CommandSink::Immediate { ref cmd_buffer, ref mut encoder_state, .. } => {
-                match *encoder_state {
-                    EncoderState::None => (),
-                    _ => panic!("Must be outside of encoding"),
-                };
+            CommandSink::Immediate { ref cmd_buffer, .. } => {
                 let _ap = AutoreleasePool::new();
-                let encoder = cmd_buffer.new_render_command_encoder(&descriptor);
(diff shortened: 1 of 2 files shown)


src/backend/metal/src/command.rs
```diff
@@ -1689,7 +1695,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
     {
         let descriptor = unsafe {
             // FIXME: subpasses
-            let pass_descriptor: metal::RenderPassDescriptor = msg_send![frame_buffer.0, copy];
+            let pass_descriptor = frame_buffer.0.clone();
```
grovesNL: Is this change intentional? AFAIK [clone will send `retain` instead of `copy`](https://github.com/gfx-rs/metal-rs/blob/master/src/lib.rs#L388)

kvark: Sorry, my bad here.
A strong reason to revise the ref-counting model we do for Metal. If we go with the plan expressed in https://github.com/gfx-rs/metal-rs/issues/40 , we can make it so `Clone` is actually cloning the thing (rather than the pointer).

kvark:
@grovesNL 
> I agree that it seems unusual for this to be the default considering it's an optimization... glad we caught this early.

I'm still confused though, how did we even pass the simple depth/stencil clear tests if the default is DontCare? Perhaps, pure luck.

kvark:
bors r=grovesNL

PR #2023, opened 2018-05-14
Metal image views

### LINKED ISSUE #2019: `make reftests` sometimes hangs on macOS
- GFX version: f10a8fb
- OS: macOS
- GPU: Intel HD Graphics 6000

Sometimes (about 10-20% of the time) running `make reftests` will hang on the output `Scene 'transfer':`.

Also running the reftests through Xcode with Metal API validation enabled sometimes displays messages such as:
> 2018-05-12 23:48:33.330609-0600 reftest[8278:727374] Execution of the command buffer was aborted due to an error during execution. Invalid Resource (IOAF code 9)

which may be related to this issue (for example, resources accidentally being freed out-of-order are causing the command buffer completion to hang somehow).

We need to:

- Investigate why this hang occurs (perhaps add more debug logging so we can determine the exact set of calls into the Metal API to reproduce this more reliably)
- Investigate whether the the Xcode messages are related or whether better feedback is possible (maybe there is a way to get more detail about these failures, or newer versions of Xcode or macOS might provide more detail)

kvark:
Fixes #2019
(finally had a good reason to learn "git bisect", at last)

The image views should be robust now. Unfortunately, related CTS tests fail because one of the buffers used in their compute somehow ends up with offset of 5 being bound, which is not supported by Metal. Need to investigate more (later).

PR checklist:
- [x] `make reftests` succeeds (yay!)

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -1050,8 +1050,10 @@ impl hal::Device<Backend> for Device {
 
         // temporary command buffer to copy the contents from
         // the given buffers into the allocated CPU-visible buffers
-        let mut pool = self.command_shared.queue_pool.lock().unwrap();
-        let (queue_id, cmd_buffer) = pool.make_command_buffer(&self.device);
+        let (queue_id, cmd_buffer) = self.command_shared.queue_pool
+            .lock()
+            .unwrap()
+            .make_command_buffer(&self.device);
         let encoder = cmd_buffer.new_blit_command_encoder();
         let mut num_copies = 0;
 
@@ -1090,7 +1092,10 @@ impl hal::Device<Backend> for Device {
         }
 
         encoder.end_encoding();
-        pool.release_command_buffer(queue_id);
+        self.command_shared.queue_pool
+            .lock()
+            .unwrap()
+            .release_command_buffer(queue_id);
         if num_copies != 0 {
             debug!("\twaiting...");
             cmd_buffer.commit();
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -1633,13 +1633,11 @@ impl hal::Device<Backend> for Device {
     fn create_image_view(
         &self,
         image: &n::Image,
-        _kind: image::ViewKind,
+        kind: image::ViewKind,
         format: format::Format,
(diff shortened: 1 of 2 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1050,8 +1050,10 @@ impl hal::Device<Backend> for Device {
 
         // temporary command buffer to copy the contents from
         // the given buffers into the allocated CPU-visible buffers
-        let mut pool = self.command_shared.queue_pool.lock().unwrap();
-        let (queue_id, cmd_buffer) = pool.make_command_buffer(&self.device);
+        let (queue_id, cmd_buffer) = self.command_shared.queue_pool
```
grovesNL: Ouch, is there a way we can improve this to avoid being bitten by this mutex again? It seems like it would be really easy to make this mistake – basically anytime we keep a lock active while calling other functions (which may or may not happen to try to acquire a lock somewhere down their callstack).

Maybe (internally) we could pass `queue_pool` through some parts call stack explicitly, so the caller can decide how the locking should work.

src/backend/metal/src/device.rs
```diff
@@ -1643,7 +1646,25 @@ impl hal::Device<Backend> for Device {
             },
         };
 
-        Ok(n::ImageView(image.raw.new_texture_view(mtl_format)))
+        if swizzle != format::Swizzle::NO {
```
grovesNL: Do we have any ideas about how to handle this already?

src/backend/metal/src/device.rs
kvark: > Maybe (internally) we could pass queue_pool through some parts call stack explicitly, so the caller can decide how the locking should work.

Could you elaborate on that?

I was thinking more in line of separating acquisition from release. The latter can just be `Arc<AtomicUint>`, which will be flushed on acquisition.

src/backend/metal/src/device.rs
kvark: Actually, no. I think we'll have this as a flag for VkPI: swizzling is not allowed (on Metal)

kvark:
Interestingly, this regressed our 1D image clears. The failure seems legitimate and related to the hack we have for treating a 1D image as 2D if it has mipmaps.

PR #2028, opened 2018-05-14
Blitting for dx12

kvark:
Revives work of #2001

PR checklist:

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -243,9 +243,10 @@ impl PipelineCache {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 enum BindPoint {
     Compute,
-    Graphics,
-     // Internal pipelines used for blitting, copying, etc.
-    InternalGraphics,
+    Graphics {
+        /// Internal pipelines used for blitting, copying, etc.
+        internal: bool,
+    }
 }
 
 #[derive(Clone)]
@@ -338,7 +339,7 @@ impl CommandBuffer {
             gr_pipeline: PipelineCache::new(),
             primitive_topology: d3dcommon::D3D_PRIMITIVE_TOPOLOGY_UNDEFINED,
             comp_pipeline: PipelineCache::new(),
-            active_bindpoint: BindPoint::Graphics,
+            active_bindpoint: BindPoint::Graphics { internal: false },
             active_descriptor_heaps: [ptr::null_mut(); 2],
             occlusion_query: None,
             pipeline_stats_query: None,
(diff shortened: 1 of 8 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -1231,17 +1293,231 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+            unsafe {
+                self.raw.RSSetViewports(1, &viewport);
+                self.raw.RSSetScissorRects(1, &scissor);
+                self.raw.IASetPrimitiveTopology(d3dcommon::D3D_PRIMITIVE_TOPOLOGY_TRIANGLELIST);
+                self.raw.SetPipelineState(blit.pipeline.as_raw());
+                self.raw.SetGraphicsRootSignature(blit.signature.as_raw());
+                self.raw.SetGraphicsRootDescriptorTable(0, srv_heap_gpu);
+            }
+
+            let mut barriers = (0 .. num_layers)
```
msiglreith: The barriers should be moved out of the region loops.
It might be possible that two regions access the same subresource, messing with the resource states. Therefore barriers should be set once on begin of the blitcommand and once at the end.
(The worst case would be if src and dst image are the same but not sure if this is allowed..)

src/backend/dx12/src/command.rs
```diff
@@ -1231,17 +1293,231 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                        }
+                    }
+
+                    self.shared.service_pipes.get_blit_2d_color(dst.dxgi_format, filter)
+                },
+                _ => unimplemented!(),
+            };
+
+            self.set_internal_graphics_pipeline();
+            // TODO: reset descriptor heaps
```
msiglreith: nit: remove comment

src/backend/dx12/src/command.rs
```diff
@@ -1231,17 +1293,231 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+        // TODO: depth or stencil images not supported so far
+
+        // TODO: only supporting 2D images
+        match (src.kind, dst.kind) {
+            (image::Kind::D2(..), image::Kind::D2(..)) => {},
+            _ => unimplemented!(),
+        }
+
+        // Descriptor heap for the current blit, only storing the src image
+        let srv_heap = {
```
msiglreith: As I've just seen it I think we can use `Device::create_descriptor_heap_impl` here also.

src/backend/dx12/src/command.rs
kvark: That's ideal but more problematic to actually implement...

src/backend/dx12/src/command.rs
```diff
@@ -1231,17 +1294,232 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                        *unsafe { desc.u.Texture2DArray_mut() } = d3d12::D3D12_TEX2D_ARRAY_RTV {
+                            MipSlice: r.dst_subresource.level as _,
+                            FirstArraySlice: (i + first_layer) as u32,
+                            ArraySize: 1,
+                            PlaneSlice: 0, // TODO
+                        };
+
+                        let view = rtv_pool.at(i as _).cpu;
+                        unsafe {
+                            device.CreateRenderTargetView(dst.resource, &desc, view);
```
msiglreith: It's probably not easy to see (sorry), but the RTVs only live for the lifetime of one regions. The handles get overriden at each region call (not required anymore after calling OMSetRenderTargets in the previous code).
ATM it will set the wrong RTVs in case of multiple regions

msiglreith: Confused myself with my initial implementation.

PR #2029, opened 2018-05-15
Reducing number of allocated buffers for small payload commands.

Thomspoon:
Fixes #issue
PR checklist:
- [✅] `make` succeeds (on *nix)
- [❌] `make reftests` succeeds
- [✅] tested examples with the following backends:
**metal**

There is currently a segmentation fault with the `make reftests` command. Most likely due to invalid use of raw pointers with the BindBufferData addition.

This PR should contain a 75% of the additions of the setBytes-genre functions that get rid of buffer allocations in certain render and compute pipelines.


src/backend/metal/src/command.rs
```diff
@@ -1175,10 +1179,11 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                 buffer: Some(buffer.raw.clone()),
                 offset: start,
             },
-            soft::ComputeCommand::BindBuffer {
+            soft::ComputeCommand::BindBufferData {
                 index: 1,
-                buffer: Some(buf_value_and_length),
-                offset: 0,
+                bytes: unsafe { 
+                    slice::from_raw_parts(&value_and_length as *const u32 as _, mem::size_of::<u32>() * value_and_length.len()).to_owned()
```
kvark: let's use the same formatting and `as_ptr()` call as with the other place

grovesNL: Could we just pass `vec![data, length as _]` here?

src/backend/metal/src/command.rs
```diff
@@ -1543,11 +1539,15 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-                    soft::RenderCommand::BindBuffer {
+                    soft::RenderCommand::BindBufferData {
                         stage: pso::Stage::Vertex,
                         index: 0,
-                        buffer: Some(vertex_buffer),
-                        offset: 0,
+                        bytes: unsafe { 
+                            slice::from_raw_parts(
+                                list.as_ptr() as *const u8, 
+                                list.len() * mem::size_of::<BlitVertex>()
```
grovesNL: I think this size is too big, seems like it should be `list.len() * mem::size_of::<BlitVertex>() / mem::size_of::<u32>()` if I understand correctly

grovesNL:
Deleted my comments because they were only relevant when it was still `Vec<u32>` 😅 oops

Issue #2032, opened 2018-05-16
Missing D3D formats


kvark:
There is a few formats that are required by Vulkan but appears missing on D3D11/12:
  - a1r5g5b5
  - a8b8g8r8

We need to think if they are possible to work around somehow, e.g. with swizzling.

PR #2033, opened 2018-05-16
[dx12] proper querying for format properties

kvark:
Brings our failure count for format properties to 7 (out of 220), all of which correspond to #2032

PR checklist:
- [x] tested examples with the following backends: d3d12


src/backend/dx12/src/lib.rs
```diff
@@ -320,14 +320,30 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
     }
 
     fn image_format_properties(
         &self, format: f::Format, dimensions: u8, tiling: image::Tiling,
         usage: image::Usage, storage_flags: image::StorageFlags,
     ) -> Option<image::FormatProperties> {
         conv::map_format(format)?; //filter out unknown formats
+        let format_desc = format.surface_desc();
+        // transfers are tricky, here is a list of exclusions
+        if usage.intersects(image::Usage::TRANSFER_SRC | image::Usage::TRANSFER_DST) {
```
msiglreith: Couldn't we just check here if the requested flags are supported properties of the format?

src/backend/dx12/src/lib.rs
```diff
@@ -781,6 +805,79 @@ impl hal::Instance for Instance {
...
+                let can_buffer = 0 != data.Support1 & d3d12::D3D12_FORMAT_SUPPORT1_BUFFER;
+                let can_image = 0 != data.Support1 & (
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURE1D |
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURE2D |
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURE3D |
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURECUBE
+                );
+                if can_image {
+                    props.optimal_tiling |= f::ImageFeature::SAMPLED;
+                    // enable blit flags on non-compressed images only
```
msiglreith: BC compressed must support blit src.
blit dst is implemented via rendertargets so should be set along with the color attachment flag. For later also for depth formats.

src/backend/dx12/src/lib.rs
```diff
@@ -781,6 +806,79 @@ impl hal::Instance for Instance {
...
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURE1D |
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURE2D |
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURE3D |
+                    d3d12::D3D12_FORMAT_SUPPORT1_TEXTURECUBE
+                );
+                if can_image {
+                    props.optimal_tiling |= f::ImageFeature::SAMPLED | f::ImageFeature::BLIT_SRC;
+                    // enable blit destination flag on non-compressed images only
+                    if !format.surface_desc().is_compressed() {
+                        props.optimal_tiling |= f::ImageFeature::BLIT_DST;
```
msiglreith: still need to check for `D3D12_FORMAT_SUPPORT1_RENDER_TARGET` format support. As BC formats don't support RT anyway no need for this additional check. Applying `f::ImageFeature::BLIT_DST` together with `f::ImageFeature::COLOR_ATTACHMENT` should be sufficient.

kvark:
(as per gitter discussion - will follow up with more fixes)
bors r=msiglreith

kvark:
hmm, the compressed formats as well as some blits are now screwed up, need more polish here
bors r-

src/backend/dx12/src/device.rs
```diff
@@ -2049,8 +2053,15 @@ impl d::Device<B> for Device {
...
         // if the format supports being rendered into, allowing us to create clear_Xv
-        let can_clear_color = image.usage.intersects(Usage::TRANSFER_DST | Usage::COLOR_ATTACHMENT);
-        let can_clear_depth = image.usage.intersects(Usage::TRANSFER_DST | Usage::DEPTH_STENCIL_ATTACHMENT);
+        let format_properties = &self.format_properties[image.format as usize];
+        let props = match image.tiling {
+            image::Tiling::Optimal => format_properties.optimal_tiling,
+            image::Tiling::Linear => format_properties.linear_tiling,
+        };
+        let can_clear_color = image.usage.intersects(Usage::TRANSFER_DST | Usage::COLOR_ATTACHMENT) &&
+            props.contains(format::ImageFeature::COLOR_ATTACHMENT);
```
msiglreith: I don't think that's required, but haven't looked into it.

Issue #1945, opened 2018-04-17
[dx12] Increase max RTV allocations (or make it configurable).


dati91:
https://github.com/gfx-rs/gfx/blob/281e79864181973a530538b557e59de1db540122/src/backend/dx12/src/lib.rs#L497
To run our test suite, we need more rtv allocated. Currently it is (a constant) 256.

msiglreith:
We need a way to handle these cpu descriptors in the longer term:

1. (Currently used in rostkatze): Having a vector of fixed size cpu descriptor heads which will be filled up.
Pro: Easy to implement
Cons: Requires synchronization, additional work required to quickly check for free spaces

2. Dynamically allocate cpu descriptor heaps on demand, motivated by the fact that cpu descriptor heaps are basically just designed to mirror the gpu heaps (nicer design). In particular, interesting for RTV and DSV as they are only needed for framebuffers, so have a fixed scope and can be batched together in case of multiple attachments. Only other use case for RTV and DSV would be depth clearing images outside of renderpasses, where we only specify an image and subresource range. This requires creating a RTV/DSV on the fly, but one might argue that these functions are slow-path.

zakorgy:
Could we temporary bump the `max_rtvs` value to 2048 and `max_dsv` to 128, until the proper solution is shipped? These are the minimum limits we need to run all the wrench tests without fail.

kvark:
@zakorgy we certainly should, yeah

msiglreith:
`dEQP-VK.texture.filtering.cube.formats.a1r5g5b5_unorm.nearest_mipmap_linear` fails due to running out of samplers.

PR #2037, opened 2018-05-17
Format packing semantics, dx12 handling

kvark:
Fixes all of the image related copy/blit CTS tests (>11K passes!). There are 0 failures in this category now and ~5 crashes due to #1945.

PR checklist:

### DIFF
--- a/src/hal/src/format.rs
+++ b/src/hal/src/format.rs
@@ -40,6 +40,11 @@ pub struct FormatDesc {
     pub bits: u16,
     /// Dimensions (width, height) of the texel blocks.
     pub dim: (u8, u8),
+    /// The format representation depends on the endianness of the platform.
+    ///
+    /// * On little-endian systems, the actual oreder of components is reverse of what
+    ///   a surface type specifies.
+    pub packed: bool,
     /// Format aspects
     pub aspects: Aspects,
 }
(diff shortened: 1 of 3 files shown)


src/hal/src/format.rs
```diff
@@ -229,13 +234,14 @@ macro_rules! surface_types {
...
-            ///
+            /// Return the format descriptor.
             pub fn desc(&self) -> FormatDesc {
                 match *self {
                     $( SurfaceType::$name => FormatDesc {
-                        bits: $total,
-                        aspects: $(Aspects::$aspect)|*,
+                        bits: $total.min(!$total),
                         dim: $dim,
+                        packed: $total > 0x1000,
```
msiglreith: Quite a bit of magic going on here but looks ok for now :p

src/hal/src/format.rs
```diff
@@ -40,6 +40,11 @@ pub struct FormatDesc {
     pub bits: u16,
     /// Dimensions (width, height) of the texel blocks.
     pub dim: (u8, u8),
+    /// The format representation depends on the endianness of the platform.
+    ///
+    /// * On little-endian systems, the actual oreder of components is reverse of what
```
msiglreith: nit: oreder/order

kvark:
Thanks for the review!
Can't wait to make that portability pr, so let's proceed and follow up later with nit fixes.
Bors r=msiglreith

PR #2040, opened 2018-05-17
[dx11] Start work on dx11 backend

fkaa:
Adds dx11 as a new backend for hal. Not currently hooked up to anything else since it's quite full of `unimplemented!()` and currently only capable of a basic run loop (swapchain `Present()`).

PR checklist:


src/backend/dx11/src/conv.rs
```diff
@@ -0,0 +1,82 @@
+use hal::format::{Format};
+
+use winapi::um::d3d11::*;
+use winapi::um::d3dcommon::*;
+use winapi::shared::dxgiformat::*;
+
+// TODO: stolen from d3d12 backend, maybe share function somehow?
```
kvark: speaking of sharing, which of the things here are from `d3d11`?

src/backend/dx11/src/lib.rs
```diff
@@ -93,408 +127,1163 @@ impl Instance {
...
+            &mut feature_level as *mut _,
+            ptr::null_mut()
+        )
+    };
+
+    if !winerror::SUCCEEDED(hr) {
+        // if there is no 11.1 runtime installed, requesting
+        // `D3D_FEATURE_LEVEL_11_1` will return E_INVALIDARG so we just retry
+        // without that
+        if hr == winerror::E_INVALIDARG {
```
kvark: what about different errors?

src/backend/dx11/src/lib.rs
```diff
@@ -93,408 +127,1163 @@ impl Instance {
...
-                    QueueType::Compute => {
-                        gpu.compute_queues.push(core::ComputeQueue::new(raw_queue()));
-                    }
-                    QueueType::Transfer => {
-                        gpu.transfer_queues.push(core::TransferQueue::new(raw_queue()));
-                    }
-                }
+            let format_desc = config
+                .color_format
+                .base_format()
```
kvark: nit: there is `surface_desc()` now

src/backend/dx11/src/conv.rs
fkaa: You mean which formats are compatible with dx11?

src/backend/dx11/src/lib.rs
fkaa: According to [msdn](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476082(v=vs.85).aspx) that's the only relevant return code here (we dont request debug)

src/backend/dx11/src/conv.rs
kvark: I mean, what part of `winapi::um::d3d11` is used here?

src/backend/dx11/src/lib.rs
kvark: let's assert on that knowledge then ;)

src/backend/dx11/src/conv.rs
fkaa: Ah, good catch 😄

fkaa:
The Makefile already builds with --all (can also see it in appveyor logs), so I only had to add it to the exclude list for linux.

kvark:
@fkaa in the future, we'd need to have either `FEATURES_HAL3=dx11` somewhere, or refactor the extra features semantics in `Makefile` to just be a list. But this is not required for now, as long as the basics are compiled at least.
bors r+

kvark:
Needs a rebase.

fkaa:
@kvark rebased

PR #2042, opened 2018-05-17
[hal] export a prelude  #1287

### LINKED ISSUE #1287: Have prelude module
It can just be renamed from `traits`, plus maybe a few things added.

dannyfritz:
Fixes #1287

### DIFF
--- a/src/hal/src/lib.rs
+++ b/src/hal/src/lib.rs
@@ -23,6 +23,8 @@ use std::hash::Hash;
 
 //TODO: reconsider what is publicly exported
 
+pub mod prelude;
+
 pub use self::adapter::{
     Adapter, AdapterInfo, MemoryProperties, MemoryType, MemoryTypeId,
     PhysicalDevice, QueuePriority,
--- a/src/hal/src/prelude.rs
+++ b/src/hal/src/prelude.rs
@@ -1,0 +1,2 @@
+#[doc(no_inline)] pub use ::Instance;
+#[doc(no_inline)] pub use ::Backend;
--- a/src/hal/src/prelude.rs
+++ b/src/hal/src/prelude.rs
@@ -1,2 +1,23 @@
-#[doc(no_inline)] pub use ::Instance;
-#[doc(no_inline)] pub use ::Backend;
+//! The HAL prelude
+//!
+//! The `prelude` module specifies common traits, mods, structs, enums, and fn's to be included
+//! into the current namespace using a wildcard.
+//! 
+//! To use this prelude:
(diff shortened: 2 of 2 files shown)


dannyfritz:
Currently only reexports the one `hal` trait.

https://github.com/gfx-rs/gfx/blob/master/examples/render/quad_render/main.rs#L15-L18
I was looking at the quad example and was wondering if these are worth including in the prelude:
* `hal::command`
* `hal::image`
* `hal::format`
* `hal::pso`

kvark:
Note that `src/render` is no longer developed. Focus your changes on `src/hal` please.

dannyfritz:
I think there is an argument to allow this or another prelude to pollute the namespace with more than just the traits.

https://github.com/gfx-rs/gfx/blob/master/examples/hal/quad/main.rs#L20-L29
What from this example would exist in every `hal` program?

kvark:
> I think there is an argument to allow this or another prelude to pollute the namespace with more than just the traits.

I haven't given this too much thought, but I'd like to hear that argument.
Subjectively, I'd like to draw a hard line between what needs to go into prelude and what not, if we even need one. Drawing the line between traits and non-traits makes sense to me.

dannyfritz:
A base prelude I think can stay with just the traits as those are annoying to bring in when you aren't directly using them usually.

But I don't think there is a reason there can't be a `hal::prelude::render::*` that brings in stuff that is common for rendering, or a `hal::prelude::compute::*` that brings in stuff more common to compute programs, etc.

The thinking is that when someone is writing an application using the `hal`, it is a core part of their project and they are okay with the somewhat implicit namespace pollution. The namespace pollution is completely opt-in after all.

kvark:
> The thinking is that when someone is writing an application using the hal, it is a core part of their project and they are okay with the somewhat implicit namespace pollution. The namespace pollution is completely opt-in after all.

True, but there is a different thinking here as well - one about the fact HAL is not quite meant to be consumed directly by the users, but rather - via higher level libraries and/or engines. With this in mind, we shouldn't try too hard to provide better ergonomics, given that function and performance are higher priorities.

kvark:
Sorry about not giving it a proper thought in the first place!
Closing as there doesn't seem to be any disagreement with my last statement.
Hopefully, we can get an easy introductory bug for you that is more useful.

dannyfritz:
No problem! I have been using `gfx` a bit since this PR and am gaining more understanding of how `gfx` works. I'll sift through the issues again and see if there is another one I'd like to tackle.

PR #2043, opened 2018-05-18
Implement `set_line_width`

grovesNL:
Just implementing stubs for `set_line_width` to help with portability crashes. No-ops on DX12 and Metal because it doesn't seem we have any way to expose it there (at least at first glance), so we simply don't expose `wideLines` as a feature and force `1.0` as the line width.

(I haven't actually tried to build the Vulkan and DX12 backends yet, so CI may fail here – I'll fix later if it does)

PR checklist:

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -1155,3 +1155,10 @@ impl hal::Backend for Backend {
     type Semaphore = native::Semaphore;
     type QueryPool = native::QueryPool;
 }
+
+fn validate_line_width(width: f32) {
+    // Note from the Vulkan spec:
+    // > If the wide lines feature is not enabled, lineWidth must be 1.0
+    // Simply assert and no-op because DX12 never exposes `Features::LINE_WIDTH` 
+    assert_eq!(width, 1.0);
+}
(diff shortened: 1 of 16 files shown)


grovesNL:
It turns out ash doesn't expose `cmd_set_line_width` yet – opened MaikKlein/ash#63

src/backend/dx12/src/lib.rs
```diff
@@ -1150,3 +1150,10 @@ impl hal::Backend for Backend {
     type Semaphore = native::Semaphore;
     type QueryPool = native::QueryPool;
 }
+
+fn validate_line_width(width: f32) {
+    // Note from the Vulkan spec:
+    // > If the wide lines feature is not enabled, lineWidth must be 1.0
+    // Simply assert and no-op because Metal never exposes `Features::LINE_WIDTH`
```
msiglreith: Nit: dx12 not metal

src/backend/dx12/src/device.rs
```diff
@@ -1605,6 +1605,11 @@ impl d::Device<B> for Device {
 
         let topology = conv::map_topology(desc.input_assembler.primitive);
 
+        match pipeline_desc.rasterizer.polygon_mode {
```
kvark: nit: use `if let`

src/backend/metal/src/device.rs
```diff
@@ -835,6 +835,11 @@ impl hal::Device<Backend> for Device {
 
         pipeline.set_vertex_descriptor(Some(&vertex_descriptor));
 
+        match pipeline_desc.rasterizer.polygon_mode {
```
kvark: similarly, let's go with `if let`

grovesNL:
Thanks for the review! Addressed comments and squashed

grovesNL:
Rebased for dx11

grovesNL:
@kvark sorry just noticed an error and need to rebase again, could we cancel this until I fix it?

kvark:
sure
bors r-

grovesNL:
@kvark should be fine now

kvark:
Bors r+

PR #2044, opened 2018-05-18
[mtl] lock Device with Mutex. Force Private memory for tiled textures.

kvark:
Addresses a major part of #1984 but doesn't resolve it completely.
PR checklist:


kvark:
@grovesNL yeah, I'm not a fan of those locks either, but that's the reality today we live in. Let's keep an eye on them. Thanks for the review!
bors r=grovesNL

Issue #1961, opened 2018-04-23
[ll] Swapchain acquire_frame and present need to return Result


HybridEidolon:
The Vulkan analog for the two trait functions `acquire_frame` and `present` on hal::Swapchain can result in errors. See [the vulkan spec for the WSI swapchain extension.](http://vulkan-spec-chunked.ahcox.com/ch29s06.html)

The return codes VK_ERROR_OUT_OF_DATE_KHR and VK_SUBOPTIMAL are critical for recognizing when a swapchain needs to be recreated; there can be a data race between the calls for `acquire_frame` and `present` in user code that would result in OUT_OF_DATE being returned on the latter, even if a window resize event from the host window system were handled to recreate the swapchain just before `acquire_frame`. _(On my GTX 1080 on Windows, it does not appear that nvidia's driver will ever yield VK_SUBOPTIMAL and gracefully fall back to scaling the swap image like DXGI does)._

User code both needs to be exposed to these error conditions and also handle them for full Vulkan transparency.

This is probably necessary for Vulkan Portability. In DXGI, Present can return a comparable error, but only when the device itself is lost; it silently handles swapchains of the wrong size (which is why you could resize the quad demo freely right now). In Metal, there are no errors at all from `present`. In GL, well, it depends on the windowing implementation.

kvark:
Acquire part is addressed in #2345, present still needs to be fixed.

kvark:
Present got addressed in #2502

Issue #1780, opened 2018-01-31
vulkan: master branch 'quad' example crashes while resizing the window


ccll:
Short info header:
'quad' example crashes with an error 'Unable to acquire a swapchain image: ErrorOutOfDateKhr' while resizing the window.

- GFX version: master branch - 48b3ef4
- OS: Ubuntu 17.10
- GPU: Intel(R) HD Graphics 530 (Skylake GT2)
- Backtrace:
```
AdapterInfo { name: "Intel(R) HD Graphics 530 (Skylake GT2)", vendor: 32902, device: 6418, software_rendering: false }
Surface format: Bgra8Srgb
Pipelines: [Ok(GraphicsPipeline(0x7f5a0132ae00))]
Memory types: [MemoryType { properties: DEVICE_LOCAL | COHERENT | CPU_VISIBLE | CPU_CACHED, heap_index: 0 }, MemoryType { properties: DEVICE_LOCAL | COHERENT | CPU_VISIBLE | CPU_CACHED, heap_index: 1 }]
thread 'main' panicked at 'Unable to acquire a swapchain image: ErrorOutOfDateKhr', /checkout/src/libcore/result.rs:906:4
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:68
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:57
             at /checkout/src/libstd/panicking.rs:381
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:397
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:577
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:538
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:522
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:498
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:71
   9: core::result::unwrap_failed
             at /checkout/src/libcore/macros.rs:23
  10: <core::result::Result<T, E>>::expect
             at /checkout/src/libcore/result.rs:799
[...]

PR #2052, opened 2018-05-19
Swapchain resize support

ZeGentzy:
Continuation of #1856 
When complete, this pull request will solve issues #1961, #1780 and #1854

Todo:
 - [x] Make quad example use resize

PR checklist:
- [x] tested examples with the following backends: vulkan and gl (both tested on ArchLinux)


ZeGentzy:
@kvark Can you test this pull request on windows with DX12 and on mac with Metal? I suspect that this patch will stop those backends from compiling properly, but I don't have access to those OSes so I can't test.

Can you also review this pull request?

Also, for some reason, the image stretches with the window with the vulkan backend but clips on the edges in the gl backend. I assume that it's necessary for both back ends to perform identically, however I'm not sure how I'd go onto doing this.

To demonstrate what I mean, here is the gl backend (left) and vulkan backend (right) side by side:

kvark:
I'll only be back on Tuesday night. In the meantime, our CI will tell you if there are any build issues on d3d12/metal.

fkaa:
@ZeGentzy AFAICT stretched image is the desired behavior, since the vertices for the image are specified in NDC space. You could check out the GL backend with renderdoc, might be something related to the viewport?

ZeGentzy:
The issue was that I wasn't updating the viewport in the gl backend. Fixed in next commit.

Now the gl backend is not respecting the quad's position and is stretching it to fill the window.

ZeGentzy:
I just realized that we need separate semaphores for each image in the swapchain. I'll implement this tomorrow.

ZeGentzy:
I'll look into refactoring and simplifying the code.

Also, why would you consider #1961 not being resolved by this? Do you want separate return codes for OUT_OF_DATE and SUBOPTIMAL?

src/backend/gl/src/window/glutin.rs
```diff
@@ -97,45 +96,57 @@ impl Surface {
...
         match (color_bits, alpha_bits, srgb) {
-            (24, 8, true) => vec![
-                f::Format::Rgba8Srgb,
-                f::Format::Bgra8Srgb,
-            ],
-            (24, 8, false) => vec![
-                f::Format::Rgba8Unorm,
-                f::Format::Bgra8Unorm,
-            ],
+            (24, 8, true) => vec![f::Format::Rgba8Srgb, f::Format::Bgra8Srgb],
```
kvark: this is an arguable rustfmt change

examples/hal/quad/main.rs
```diff
@@ -4,613 +4,1357 @@
...
+extern crate glsl_to_spirv_macros;
+#[macro_use]
+extern crate glsl_to_spirv_macros_impl;
 
-const QUAD: [Vertex; 6] = [
-    Vertex { a_Pos: [ -0.5, 0.33 ], a_Uv: [0.0, 1.0] },
-    Vertex { a_Pos: [  0.5, 0.33 ], a_Uv: [1.0, 1.0] },
-    Vertex { a_Pos: [  0.5,-0.33 ], a_Uv: [1.0, 0.0] },
+#[cfg(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl"))]
+mod renderer {
```
kvark: I agree with @msiglreith here that the increased complexity of the quad example is not a good thing. The examples need to be straightforward and readable, also the less code the better. Could you make this simpler?

examples/hal/quad/main.rs
```diff
@@ -4,613 +4,1357 @@
...
-const ENTRY_NAME: &str = "main";
-
-#[derive(Debug, Clone, Copy)]
-#[allow(non_snake_case)]
-struct Vertex {
-    a_Pos: [f32; 2],
-    a_Uv: [f32; 2],
-}
+#[macro_use]
+extern crate glsl_to_spirv_macros;
```
kvark: why do we need those macros? I'd prefer us to just do the stuff at run time if it makes us depend on less externals

examples/hal/quad/main.rs
```diff
@@ -4,613 +4,1357 @@
...
 extern crate gfx_backend_gl as back;
-
-extern crate winit;
+#[cfg(feature = "metal")]
+extern crate gfx_backend_metal as back;
+#[cfg(feature = "vulkan")]
+extern crate gfx_backend_vulkan as back;
+extern crate gfx_hal as hal;
 extern crate image;
+extern crate takeable_option;
```
kvark: can we get away without using this?

examples/hal/quad/main.rs
ZeGentzy: I wrapped many things in `#[cfg]`s to stop unused variable warnings, I can get rid of them if your ok with warnings on some/all builds.

examples/hal/quad/main.rs
ZeGentzy: Lots gfx structs need to be moved into a destroy function like `destroy_descriptor_pool` when being destroyed. We can't move things out of structs (even in the Drop function) without putting something back in.

Now, if we had some empty default objects we could do std::mem::replace(old, default). Alternatively, we could use Option and plaster the code with unwraps (which I nor @msiglreith like) or we can use my thin wrapper around Option which has the Deref and DerefMut traits (which do the unwrapping for us). 

Of course, there might be other ways, but I can't think of them.

examples/hal/quad/main.rs
kvark: Do you mean the warnings shown when no backends are selected? We should automatically use the `empty` backend in this case.

examples/hal/quad/main.rs
kvark: Isn't this only a problem when you try to wrap things inside a struct that lives longer? The old example code just did
```rust
fn main() {
   let descriptor_pool = device.create_descriptor_pool(...);
   // do stuff
   device.destroy_descriptor_pool(descriptor_pool);
}
```

ZeGentzy: Yes, only when they are in a structure. I want to avoid having one big function doing everything, and if we don't wrap related fields in structures then we might have to pass half a dozen to a dozen fields to everything.

examples/hal/quad/main.rs
ZeGentzy: Yes, but then we'd have to put certain functions behind an feature gate, because they require certain functions not present in the empty backend. I've done so in the next commit and I can revert it if you feel it's more complex.

examples/hal/quad/main.rs
kvark: > I want to avoid having one big function doing everything

This is fine for an example, especially if it simplifies destruction.

msiglreith:
> Also, why would you consider #1961 not being resolved by this? Do you want separate return codes for OUT_OF_DATE and SUBOPTIMAL?

Yes, it's only partially resolving it. To some extent this requires also changes to ash as we currently have 'everything is fine' -> Ok, 'else' -> Err. We should handle it a bit better imo.

examples/hal/quad/main.rs
```diff
@@ -3,614 +3,1380 @@
...
+extern crate gfx_backend_vulkan as back;
+#[cfg(not(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl")))]
+extern crate gfx_backend_empty as back;
 
-extern crate winit;
+extern crate env_logger;
+extern crate gfx_hal as hal;
+extern crate glsl_to_spirv;
 extern crate image;
+extern crate takeable_option;
```
kvark: do we still need that? I thought we had a consensus that laying down everything into a function body is fine for the example

examples/hal/quad/main.rs
```diff
@@ -3,614 +3,1380 @@
...
-    levels: 0 .. 1,
-    layers: 0 .. 1,
+    levels: 0..1,
+    layers: 0..1,
 };
 
-#[cfg(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl"))]
-fn main() {
-    env_logger::init();
+struct RendererState<B: Backend> {
```
kvark: ah, ok. That's definitely a bit more complex than we'd like for this example. Maybe for another one, a more general/complex one, we'll need a struct, but this one is supposed to be as simple as possible.

ZeGentzy:
BTW, quads is WIP

grovesNL:
There seems to be quite a lot of changes in this PR now, so it would be good if we could integrate the swapchain resize changes first IMO (maybe as a separate PR?).

examples/hal/quad/main.rs
```diff
@@ -499,62 +349,186 @@ fn main() {
...
-        let submission = Submission::new()
-            .submit(Some(submit));
-        queue.submit(submission, Some(&mut frame_fence));
+        let submission = Submission::new().submit(Some(submit));
+        queue_group.queues[0].submit(submission, Some(&mut frame_fence));
 
         device.wait_for_fence(&frame_fence, !0);
     }
 
+    #[cfg(feature = "vulkan")]
```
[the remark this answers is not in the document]
ZeGentzy: Only vulkan requires recreating the swapchain, the rest only need updating the viewport. It would be a waste of performance if we recreated it in other backends.

examples/hal/quad/main.rs
```diff
@@ -499,62 +349,186 @@ fn main() {
...
                     _ => (),
                 }
             }
         });
+        first = false;
+
+        if recreate_swapchain {
+            device.wait_idle().unwrap();
+            unsafe { ManuallyDrop::drop(&mut viewport) }
+            #[cfg(feature = "vulkan")]
```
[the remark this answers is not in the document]
ZeGentzy: See above

examples/hal/quad/main.rs
```diff
@@ -499,62 +349,186 @@ fn main() {
...
+                    }
                     _ => (),
                 }
             }
         });
+        first = false;
+
+        if recreate_swapchain {
+            device.wait_idle().unwrap();
+            unsafe { ManuallyDrop::drop(&mut viewport) }
```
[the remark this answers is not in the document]
ZeGentzy: I think we need to destroy the old viewport (because it depends on the swapchain) before making  a new one. Non-vulkan backends of course don't need this, but I made it manuallydrop anyways to avoid having two separate let statements.

examples/hal/quad/main.rs
```diff
@@ -617,3 +598,218 @@ fn main() {
...
+                        .create_framebuffer(&render_pass, Some(rtv), extent)
+                        .unwrap()
+                })
+                .collect();
+            (pairs, fbos)
+        }
+        Backbuffer::Framebuffer(fbo) => (Vec::new(), vec![fbo]),
+    };
+
+    let pipeline_layout =
```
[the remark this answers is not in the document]
ZeGentzy: The render pass needs to be recreated because it depends on the format of the swap chain images, which can, however unlikely, change.. The pipeline depends on that renderpass.

src/hal/src/adapter.rs
```diff
@@ -50,7 +50,7 @@ pub struct MemoryProperties {
 }
 
 /// Represents a physical device (such as a GPU) capable of supporting the given backend.
-pub trait PhysicalDevice<B: Backend>: Any + Send + Sync {
+pub trait PhysicalDevice<B: Backend>: Any + Send + Sync + Clone {
```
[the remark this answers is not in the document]
ZeGentzy: We need the physical device when updating the surface's dissensions (for the vulkan backend), sadly adapter.open_with consumes the adapter (and the physical_device attached with it). That's why we clone the physical for use later.

examples/hal/compute/main.rs
```diff
@@ -46,7 +51,19 @@ fn main() {
         .open_with::<_, Compute>(1, |_family| true)
         .unwrap();
 
-    let shader = device.create_shader_module(include_bytes!("shader/collatz.spv")).unwrap();
+    let mut glsl = String::new();
+    File::open("compute/shader/collatz.comp")
+        .unwrap()
+        .read_to_string(&mut glsl)
```
kvark: let's use a shorter variant here: https://doc.rust-lang.org/stable/std/fs/fn.read_to_string.html

examples/hal/compute/main.rs
```diff
@@ -46,7 +51,19 @@ fn main() {
...
+        .unwrap();
+    let mut spirv: Vec<u8> = vec![];
+    for byte in glsl_to_spirv::compile(&glsl, glsl_to_spirv::ShaderType::Compute)
+        .unwrap()
+        .bytes()
+    {
+        spirv.push(byte.unwrap())
```
kvark: this shouldn't be as complex. Maybe just `unwrap().bytes().collect::<Vec<_>>()`?

examples/hal/Cargo.toml
```diff
@@ -25,7 +25,9 @@ env_logger = "0.5"
 image = "0.18"
 log = "0.4"
 winit = "0.13"
+glsl-to-spirv = "0.1.4"
 gfx-hal = { path = "../../src/hal", version = "0.1" }
+gfx-backend-empty = { path = "../../src/backend/empty", version = "0.1" }
```
kvark: where is it used?

examples/hal/Cargo.toml
```diff
@@ -25,7 +25,9 @@ env_logger = "0.5"
 image = "0.18"
 log = "0.4"
 winit = "0.13"
+glsl-to-spirv = "0.1.4"
```
kvark: this reminds me: we have custom glsl-to-spirv support in some of the backends, like Vulkan.
It doesn't make any sense now, since all the backends accept SPIR-V, so any integration with glsl should be done outside of the backends.

examples/hal/quad/main.rs
```diff
@@ -4,31 +4,33 @@
...
-use hal::{
-    DescriptorPool, FrameSync, Primitive,
-    Backbuffer, SwapchainConfig,
-};
 use hal::format::{AsFormat, ChannelType, Rgba8Srgb as ColorFormat, Swizzle};
 use hal::pass::Subpass;
 use hal::pso::{PipelineStage, ShaderStageFlags, Specialization};
 use hal::queue::Submission;
+use hal::{buffer, command, format as f, image as i, memory as m, pass, pool, pso};
+use hal::{Backbuffer, DescriptorPool, FrameSync, Primitive, SwapchainConfig};
```
kvark: I prefer the old formatting. Any reason we need to change it here? Or is is just a rustfmt pass?
We don't want to mix formatting changes with functional one...

examples/hal/quad/main.rs
```diff
@@ -40,19 +42,36 @@ struct Vertex {
 }
 
 const QUAD: [Vertex; 6] = [
-    Vertex { a_Pos: [ -0.5, 0.33 ], a_Uv: [0.0, 1.0] },
```
kvark: similarly, the old formatting was cleaner

examples/hal/quad/main.rs
```diff
@@ -105,40 +120,31 @@ fn main() {
...
+    let limits = adapter.physical_device.limits();
 
-    let memory_types = adapter
-        .physical_device
-        .memory_properties()
-        .memory_types;
-    let limits = adapter
-        .physical_device
-        .limits();
+    let physical_device = adapter.physical_device.clone();
```
kvark: why do we need to clone it?

examples/hal/quad/main.rs
```diff
@@ -105,40 +120,31 @@ fn main() {
...
+    let mut command_pool =
+        device.create_command_pool_typed(&queue_group, pool::CommandPoolCreateFlags::empty(), 16);
 
-    println!("Surface format: {:?}", surface_format);
-    let swap_config = SwapchainConfig::new()
-        .with_color(surface_format)
-        .with_image_usage(i::Usage::COLOR_ATTACHMENT);
-    let (mut swap_chain, backbuffer) = device.create_swapchain(&mut surface, swap_config);
+    let (width, height) =
+        surface.update_dimensions(pixel_width as _, pixel_height as _, &physical_device);
```
kvark: oh, what happens here? why do we get different width/height? why do we need to update the dimensions, even?

examples/hal/quad/main.rs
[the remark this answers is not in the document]
kvark: Interesting. Aren't *we* dictating the format of swapchain images? Why/when would it change?

src/backend/dx12/src/window.rs
```diff
@@ -98,10 +103,13 @@ pub struct Swapchain {
     pub(crate) frame_queue: VecDeque<usize>,
     #[allow(dead_code)]
     pub(crate) rtv_heap: n::DescriptorHeap,
+    // need to associate raw image pointers with the swapchain so they can be properly released
+    // when the swapchain is destroyed
+    pub(crate) resources: Vec<*mut d3d12::ID3D12Resource>,
```
kvark: shouldn't this be using `ComPtr` for releasing?

src/backend/vulkan/src/device.rs
```diff
@@ -1461,6 +1462,11 @@ impl d::Device<B> for Device {
         // TODO: handle depth stencil
         let format = config.color_format;
 
+        let mut old_swapchain = vk::SwapchainKHR::null();
+        if let Some(osc) = provided_old_swapchain {
```
kvark: nit: let's do `let old_swapchain = match provided_old_swapchain {...}`

src/hal/src/adapter.rs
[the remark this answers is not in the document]
kvark: I don't quite like the clone-ability of things... we need to consider alternatives, e.g. not consume the physical device in the first place

examples/hal/quad/main.rs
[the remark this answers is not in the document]
kvark: but viewport is just plain data, it doesn't interact with the swapchain in any hidden way. There should be no need to hack around `drop` for the plain data

examples/hal/Cargo.toml
ZeGentzy: No where currently, used to be used. Removed.

examples/hal/Cargo.toml
ZeGentzy: If I understand correctly, you want to strip out glsl-to-spirv from the backends? I think that's outside this PR's job.

examples/hal/quad/main.rs
ZeGentzy: Just a rustfmt pass, reverting.

src/backend/dx12/src/window.rs
ZeGentzy: I inherited this from hriundel's changes. The correct usage is beyond me, you should ask someone with d3d12 experience.

ZeGentzy:
The formats supported by the surface can change between the time of creation of the first swapchain and the new one. We should be querying the surface for the info again, fixed.

ZeGentzy:
So here are the changes bellow, the commit doesn't actually work. Still trying to figure out why.

src/backend/dx12/src/window.rs
msiglreith: yes should be a ComPtr

examples/hal/quad/main.rs
```diff
@@ -578,13 +480,19 @@ fn main() {
...
             .submit(Some(submit));
-        queue.submit(submission, Some(&mut frame_fence));
+        queue_group.queues[0].submit(submission, Some(&mut frame_fence));
 
         // TODO: replace with semaphore
         device.wait_for_fence(&frame_fence, !0);
 
         // present frame
-        swap_chain.present(&mut queue, &[]);
+        match swap_chain.present(&mut queue_group.queues[0], &[]) {
```
msiglreith:
```rust
if let Err(_) = swap_chain.present(..) {
..
}
```

examples/hal/a
```diff
@@ -0,0 +1,45 @@
+diff --git a/examples/hal/quad/main.rs b/examples/hal/quad/main.rs
```
msiglreith: something went wrong here

src/backend/vulkan/src/window.rs
```diff
@@ -111,26 +182,15 @@ impl Instance {
...
+        self.create_surface_from_vk_surface_khr(surface, 1, Connection::Xcb(connection, window))
     }
 
     #[cfg(all(unix, not(target_os = "android")))]
     pub fn create_surface_from_wayland(
-        &self, display: *mut c_void, surface: *mut c_void, width: Size, height: Size
+        &self,
+        display: *mut c_void,
+        surface: *mut c_void,
+        window: &Arc<winit::Window>,
```
msiglreith: We can't just require the user to provide a reference to a winit window as this interface is supposed to work with other crates or even C application as well.

src/backend/vulkan/src/window.rs
```diff
@@ -286,13 +348,18 @@ impl Instance {
...
+            samples,
+            connection: Arc::new(Mutex::new(connection)),
+        }
     }
 }
 
 impl hal::Surface<Backend> for Surface {
     fn kind(&self) -> hal::image::Kind {
-        hal::image::Kind::D2(self.width, self.height, 1, self.samples)
+        hal::image::Kind::D2(self.current_extent.unwrap().width, self.current_extent.unwrap().height, 1, self.samples)
```
msiglreith: I feel like most of the effort went into implementing this function but I don't believe we actually need it.

ZeGentzy: We need to update the extent for swapchain recreation (so the swapchain images are the right size). Caching it for the kind function adds like ~10 lines.

examples/hal/Cargo.toml
kvark: Yes, hence the part "this reminds me" :) No action expected from this PR

src/backend/dx11/src/lib.rs
```diff
@@ -1053,7 +1054,9 @@ impl hal::Device<Backend> for Device {
         &self,
         surface: &mut Surface,
         config: hal::SwapchainConfig,
-    ) -> (Swapchain, hal::Backbuffer<Backend>) {
+        old_swapchain: Option<Swapchain>,
+        _physical_device: &PhysicalDevice,
```
kvark: I agree with @msiglreith that it doesn't seem right to accept a `PhysicalDevice` in a `Device` method.

src/backend/vulkan/src/window.rs
```diff
@@ -16,13 +16,102 @@ use winit;
 use conv;
 use {VK_ENTRY, Backend, Instance, PhysicalDevice, QueueFamily, RawInstance};
 
+pub trait GetExtentable {
+    fn get_extent(&self) -> Option<(u32, u32)>;
```
kvark: any reason this is not `hal::window::Extent2D`?

src/backend/vulkan/src/window.rs
```diff
@@ -16,13 +16,102 @@ use winit;
...
+    }
+}
+
+enum Connection {
+    #[cfg(all(unix, not(target_os = "android")))]
+    Xlib(*mut vk::Display, vk::Window),
+    #[cfg(all(unix, not(target_os = "android")))]
+    Xcb(*mut vk::xcb_connection_t, vk::xcb_window_t),
+    #[cfg(all(unix, not(target_os = "android")))]
+    Wayland(Arc<GetExtentable>),
```
kvark: why are these trait objects instead of concrete types like the other variants?

src/backend/vulkan/src/window.rs
```diff
@@ -16,13 +16,102 @@ use winit;
...
+}
 
 pub struct Surface {
     // Vk (EXT) specs [29.2.7 Platform-Independent Information]
     // For vkDestroySurfaceKHR: Host access to surface must be externally synchronized
     pub(crate) raw: Arc<RawSurface>,
-    pub(crate) width: Size,
-    pub(crate) height: Size,
+    pub(crate) current_extent: Option<hal::window::Extent2D>,
+    connection: Arc<Mutex<Connection>>,
```
kvark: who is this shared with?

src/backend/vulkan/src/window.rs
```diff
@@ -286,13 +359,18 @@ impl Instance {
...
+            samples,
+            connection: Arc::new(Mutex::new(connection)),
+        }
     }
 }
 
 impl hal::Surface<Backend> for Surface {
     fn kind(&self) -> hal::image::Kind {
-        hal::image::Kind::D2(self.width, self.height, 1, self.samples)
+        hal::image::Kind::D2(self.current_extent.unwrap().width, self.current_extent.unwrap().height, 1, self.samples)
```
kvark: let's make it cleaner by having `let ext = self.current_extent.expect("");`

src/backend/vulkan/src/window.rs
```diff
@@ -368,6 +449,14 @@ impl hal::Surface<Backend> for Surface {
     }
 }
 
+impl Surface {
+    pub(crate) fn update_extent(&mut self, physical_device: &PhysicalDevice) {
+        use hal::Surface;
+        let (caps, _) = self.capabilities_and_formats(physical_device);
```
kvark: is there a reason we can't generate `caps` during initialization and just keep it instead of re-generating here?

src/hal/src/adapter.rs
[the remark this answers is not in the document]
kvark: This is still a problem

src/backend/vulkan/src/window.rs
ZeGentzy: So we can support non-winit.

src/backend/vulkan/src/window.rs
ZeGentzy: Nobody as far as I know, the mutex is only used by me insure Sync & Send get implemented. Or do you mean the Arc? That's so that we don't have to use lifetimes which can't be used due to lack of type ascription.

src/backend/vulkan/src/window.rs
ZeGentzy: Because the extent changes when the window resizes? How else do I query for the new size?

src/backend/vulkan/src/window.rs
ZeGentzy: No

src/backend/vulkan/src/window.rs
msiglreith: Correct me if I'm wrong why do we need to cache those at all? Users should just call createswapchain with the old swapchain and the new size gathered from the resize event of the native window or winit. For additional safety checks usere can requery the surface capabilties the check the current extent (optional per platform) and check if the extents are inside the boundaries. Updating these are done by the vulkan drivers already.

src/backend/vulkan/src/window.rs
kvark: heh, I was reading it over email thinking, hmmm... did I ask "can we use `Extend2D`?" and you just replied straight "No" :rofl:

src/backend/vulkan/src/window.rs
kvark: I mean, could this just be `connection: Connection`?

examples/hal/compute/main.rs
```diff
@@ -46,7 +51,13 @@ fn main() {
         .open_with::<_, Compute>(1, |_family| true)
         .unwrap();
 
-    let shader = device.create_shader_module(include_bytes!("shader/collatz.spv")).unwrap();
+    let glsl = fs::read_to_string("compute/shader/collatz.comp").unwrap();
+    let spirv: Vec<u8> = glsl_to_spirv::compile(&glsl, glsl_to_spirv::ShaderType::Compute)
```
kvark: nit: no need to specify the `Vec` type both here and in `collect()`

examples/hal/quad/main.rs
```diff
@@ -621,3 +519,206 @@ fn main() {
...
+    std::vec::Vec<(
+        <back::Backend as hal::Backend>::Image,
+        <back::Backend as hal::Backend>::ImageView,
+    )>,
+    <back::Backend as hal::Backend>::GraphicsPipeline,
+    <back::Backend as hal::Backend>::PipelineLayout,
+    Extent2D,
+) {
+    let surface_format = surface
+        .capabilities_and_formats(device.get_physical_device())
```
kvark: Let's follow the same logic as a Vulkan application:
  - pass in the physical device here that is held by the user
  - don't consume it on `open`

PR #2053, opened 2018-05-19
Fix metal allocation tracking

kvark:
The old logic was a bit too complex and incorrect: it only worked with cases where either start or the end of tracked ranges are within the current one.

Another thing here is more limits exposed.

I was hoped to get a chunk of `texture` CTS category tests, but most of them still fail at the moment. The problem seems to be that tests try to use the attribute offsets above the buffer strides... Metal doesn't like that. Need to investigate more and file another issue here.

PR checklist:


src/backend/metal/src/command.rs
```diff
@@ -2200,7 +2200,9 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         }
 
         let sink = inner.sink();
-        sink.blit_commands(blit_commands.into_iter());
+        if !blit_commands.is_empty() {
+            sink.blit_commands(blit_commands.into_iter());
```
grovesNL: Should we simply move this check to the beginning of `blit_commands` instead? It seems like there are already some other places where an empty iterator could be passed

src/backend/metal/src/device.rs
```diff
@@ -1048,7 +1049,7 @@ impl hal::Device<Backend> for Device {
                 }
                 command::exec_blit(encoder, &soft::BlitCommand::CopyBuffer {
                     src: cpu_buffer.clone(),
-                    dst,
+                    dst: dst.to_owned(),
```
grovesNL: Why is this needed now? Or is this simply preparing for future changes on the metal-rs side?

src/backend/metal/src/native.rs
```diff
@@ -284,30 +280,30 @@ unsafe impl Sync for Memory {}
...
+        MemoryAllocations {
+            buffers: HashMap::new(),
+        }
+    }
+
+    /// Get all unique buffers that intersects specified range
+    pub fn find<'a>(&'a self, range: &'a Range<u64>) -> impl Iterator<Item=(Range<u64>, &'a metal::BufferRef)> {
+        self.buffers
+            .iter()
+            .filter(move |&(ref r, _)| r.start < range.end && r.end > range.start)
```
grovesNL: Nice and clean 👍

src/backend/metal/src/command.rs
kvark: It doesn't seem convenient, since we have an iterator, so we'd have to mess with some lazy initialization or fixed-length iterator traits, etc. Doing it here is straightforward.

src/backend/metal/src/device.rs
kvark: Because we iterate over `&BufferRef` now. Previously this was done within `find`, so no performance/semantics change.

kvark:
Thanks for the review!
bors r=grovesNL

PR #2057, opened 2018-05-21
Adds constant specialization to gl

### LINKED ISSUE #2056: OpenGL doesn't support constant specialization
The vulkan backend is rendering images too small. To make this more obvious, I've set the scaling factor in the vertex shader to 1 (no scaling) and changed the quad info to this:

POS X | POS Y | UV X | UV Y
-|-|-|-
-1 | 1 | 0 | 1
1 | 1 | 1 | 1
1 | -1 | 1 | 0
-1 | 1 | 0 | 1
1 | -1 | 1 | 0
-1 | -1 | 0 | 0

We should then expect to see the image touching the window's borders, and we do in the OpenGL backend, but in the vulkan one we don't. 

OpenGL on left, Vulkan on right.

Short info header:
- GFX version:  #2052 and master
- OS: Linux
- GPU: 00:01.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Kaveri [Radeon R7 Graphics]

ZeGentzy:

Fixes #2056 

PR checklist:
- [x] tested examples with the following backends: gl on archlinux

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -213,6 +213,39 @@ impl Device {
             })
     }
 
+    fn specialize_ast(
+        &self,
+        ast: &mut spirv::Ast<glsl::Target>,
+        specializations: &[pso::Specialization],
+    ) -> Result<(), d::ShaderError> {
+        let spec_constants = ast
+            .get_specialization_constants()
+            .map_err(gen_unexpected_error)?;
+
+        for spec_constant in spec_constants {
+            if let Some(constant) = specializations
+                .iter()
+                .find(|c| c.id == spec_constant.constant_id)
+            {
+                // Override specialization constant values
+                unsafe {
+                    let value = match constant.value {
+                        pso::Constant::Bool(v) => v as u64,
+                        pso::Constant::U32(v) => v as u64,
+                        pso::Constant::U64(v) => v,
+                        pso::Constant::I32(v) => *(&v as *const _ as *const u64),
+                        pso::Constant::I64(v) => *(&v as *const _ as *const u64),
+                        pso::Constant::F32(v) => *(&v as *const _ as *const u64),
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/device.rs
```diff
@@ -213,6 +213,34 @@ impl Device {
...
+        specializations: &[pso::Specialization],
+    ) {
+        let spec_cons = ast.get_specialization_constants().unwrap();
+        for spec in specializations {
+            let id = spec_cons.clone().iter()
+                .filter(|tspec| tspec.constant_id == spec.id)
+                .next().unwrap().id;
+
+            unsafe {
+                let value = match spec.value {
```
grovesNL: Could we just use the same logic from https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12/src/device.rs#L403 here?

src/backend/gl/src/device.rs
```diff
@@ -213,6 +213,32 @@ impl Device {
             })
     }
 
+    fn specialize_ast(
+        &self,
+        ast: &mut spirv::Ast<glsl::Target>,
+        specializations: &[pso::Specialization],
+    ) {
+        let spec_cons = ast.get_specialization_constants().unwrap();
+        for spec in specializations {
```
grovesNL: I guess this part could be the same as [the DX12 logic](https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12/src/device.rs#L396) as well

grovesNL:
Looks great, thanks! Please squash the commits and then it should be ready to merge 👍

grovesNL:
@kvark this should be fine to merge when you're back online

kvark:
@ZeGentzy @msiglreith wait, was this not merged?

ZeGentzy:
It was merged but it didn't close the PR, see https://github.com/gfx-rs/gfx/commit/909b13cffd0d548bae41e785db255537094c6c0f

PR #2062, opened 2018-05-23
[dx11] add dxgi helper module and work on command buffers

fkaa:
Current plan is to represent command buffers as deferred contexts. Not entirely sure how to do Memory/heap, but [this presentation](https://developer.nvidia.com/sites/default/files/akamai/gamedev/files/gdc12/Efficient_Buffer_Management_McDonald.pdf) (`MAP_NO_OVERWRITE`) has some interesting stuff.

PR checklist:
- [ ] tested examples with the following backends: dx11


src/backend/dx11/src/conv.rs
```diff
@@ -79,3 +82,23 @@ pub(crate) fn map_format(format: Format) -> Option<DXGI_FORMAT> {
 
     Some(format)
 }
+
+pub(crate) fn map_viewport(viewport: &Viewport) -> D3D11_VIEWPORT {
```
msiglreith: why `pub(crate)` instead of just `pub`? `conv` should be a private module anyway imo.

src/backend/dx11/src/lib.rs
```diff
@@ -331,10 +311,18 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+                    properties: Properties::CPU_VISIBLE | Properties::COHERENT,
+                    heap_index: 0,
+                },
+                hal::MemoryType { // download
+                    properties: Properties::CPU_VISIBLE | Properties::COHERENT | Properties::CPU_CACHED,
+                    heap_index: 0,
+                },
             ],
-            memory_heaps: vec![!0]
+            memory_heaps: vec![!0, !0, !0]
```
msiglreith: 3 heaps?

src/backend/dx11/src/lib.rs
```diff
@@ -454,20 +449,86 @@ impl hal::Device<Backend> for Device {
...
+        assert_eq!(0, offset);
+        // TODO: structured buffers
+        assert_eq!(0, unbound_buffer.bind & d3d11::D3D11_BIND_SHADER_RESOURCE);
+        // TODO: change memory to be capable of more than one buffer?
+        // assert_eq!(None, memory.buffer);
+
+        let (usage, cpu) = match (memory.can_download(), memory.can_upload()) {
+            (true,  true)  => (d3d11::D3D11_USAGE_DYNAMIC, d3d11::D3D11_CPU_ACCESS_WRITE),
+            (true,  false) => (d3d11::D3D11_USAGE_STAGING, d3d11::D3D11_CPU_ACCESS_READ | d3d11::D3D11_CPU_ACCESS_WRITE),
+            (false, true)  => (d3d11::D3D11_USAGE_DEFAULT, 0),
```
msiglreith: Could you document the decision for these a bit? I guess default is chosen due to the way data gets flushed via subresource updates during mapping?

src/backend/dx11/src/lib.rs
```diff
@@ -454,20 +449,86 @@ impl hal::Device<Backend> for Device {
...
-        unimplemented!()
+        // TODO: offset
+        assert_eq!(0, offset);
+        // TODO: structured buffers
+        assert_eq!(0, unbound_buffer.bind & d3d11::D3D11_BIND_SHADER_RESOURCE);
+        // TODO: change memory to be capable of more than one buffer?
+        // assert_eq!(None, memory.buffer);
+
+        let (usage, cpu) = match (memory.can_download(), memory.can_upload()) {
+            (true,  true)  => (d3d11::D3D11_USAGE_DYNAMIC, d3d11::D3D11_CPU_ACCESS_WRITE),
```
msiglreith: A possible d3d11 limitation? Not being able to read from the upload buffer (even if not encouraged).

src/backend/dx11/src/lib.rs
fkaa: I mostly took reasoning from GL backend to get something running, but it doesn't look very correct ("can download" = `CPU_VISIBLE | CPU_CACHED`, "upload" = `CPU_VISIBLE`).

I think there are some memory types which dont apply. For example, there doesn't seem to be non-coherent memory. Same for lazily allocated (though this doesnt seem to hard to emulate), and CPU cached memory (?) AFAICT (unless this is implied by `USAGE_STAGING`)..

Would this be more appropriate? (Assuming `CPU_VISIBLE` means read & write access)
```rust
DEVICE_LOCAL               => USAGE_DEFAULT, None
DEVICE_LOCAL | CPU_VISIBLE => USAGE_DYNAMIC, Read/Write
CPU_VISIBLE  | CPU_CACHED  => USAGE_STAGING, Read/Write
```

Not sure where to fit in `USAGE_IMMUTABLE` in to that though

PR #2064, opened 2018-05-24
Subresource footprint

### LINKED ISSUE #2051: Image layout queries and initialization
Vulkan allows to specify the initial layout on image creation, which we don't have. There is also `vkGetImageSubresourceLayout` that we need to provide. It appears to be blocking the tests:
"dEQP-VK.draw.simple_draw.*

kvark:
Fixes #2051
PR checklist:

### DIFF
--- a/src/backend/metal/src/native.rs
+++ b/src/backend/metal/src/native.rs
@@ -81,11 +81,27 @@ unsafe impl Sync for ComputePipeline {}
 #[derive(Debug)]
 pub struct Image {
     pub(crate) raw: metal::Texture,
-    pub(crate) extent: hal::image::Extent,
+    pub(crate) extent: image::Extent,
+    pub(crate) num_layers: Option<image::Layer>,
     pub(crate) format_desc: hal::format::FormatDesc,
     pub(crate) shader_channel: Channel,
     pub(crate) mtl_format: metal::MTLPixelFormat,
     pub(crate) mtl_type: metal::MTLTextureType,
+}
+
+impl Image {
+    pub(crate) fn pitches_impl(
+        level: image::Level, extent: image::Extent, format_desc: hal::format::FormatDesc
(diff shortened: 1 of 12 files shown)


src/backend/metal/src/native.rs
```diff
@@ -81,13 +82,30 @@ unsafe impl Sync for ComputePipeline {}
...
 
+impl Image {
+    pub(crate) fn pitches_impl(
+        extent: image::Extent, format_desc: hal::format::FormatDesc
+    ) -> [hal::buffer::Offset; 3] {
+        let bytes_per_texel = format_desc.bits as image::Size >> 3;
+        let row_pitch = extent.width * bytes_per_texel;
+        let depth_pitch = extent.height * row_pitch;
+        let array_pitch = extent.depth * depth_pitch;
+        [row_pitch as _, depth_pitch as _, array_pitch as _]
```
grovesNL: We could just use a tuple here so we can destructure with `let (.., array_pitch)` in the callers (instead of `pitches[2]`)

src/hal/src/image.rs
```diff
@@ -33,6 +34,17 @@ pub struct Extent {
     pub depth: Size,
 }
 
+impl Extent {
+    /// Get the extent at a particular mipmap level.
+    pub fn at_level(&self, level: Level) -> Self {
+        Extent {
+            width: 1.max(self.width >> level),
```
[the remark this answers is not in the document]
kvark: I was also wondering about this, and apparently the spec (for one of the APIs, needs a link) says it's always flooring

src/backend/metal/src/native.rs
kvark: we could, yeah

kvark:
I got too much stuff crumbled on top, and there are no major concerns here, so proceeding. Thanks for all the reviews!
bors r=grovesNL,msiglreith

PR #2070, opened 2018-05-25
Metal indirect push constants

### LINKED ISSUE #2065: Push constants on Metal
Currently blocking a large portion of CTS draw tests: "dEQP-VK.draw.instanced.*"

### LINKED ISSUE #2066: Indirect draw calls on Metal
Currently blocking a large portion of CTS draw tests, including everything in "dEQP-VK.draw.indirect_draw.*.*.*"

kvark:
Fixes #2066, Fixes #2065 
~~Includes #2064 (to be rebased)~~

Includes many cool things for Metal:
- [x] indirect draw support
- [x] push constant support
- [x] TODO: don't require push constants if the shader doesn't use them
- [x] TODO: support pipeline inheritance
- [x] workaround for an issue occurring when the PSO is bound after the vertex buffers 
- [x] metal-rs flags update
- [x] point_size built-in usage by SPIRV-Cross
- [x] new `DrawCount` typedef in HAL
- [x] depth bounds

The CTS category of "dEQP-VK.draw" now shows the following stats:
  - 170 Succeeded :tada: 
  - 0 Crashed
  - 25 Failed
    - mostly due to missing TriangleFan topology
    - depth ranges don't work probably, Apple upstream bug is filed

PR checklist:


src/backend/metal/src/command.rs
```diff
@@ -1067,6 +1122,50 @@ impl CommandBuffer {
         self.state.blend_color = Some(*color);
         soft::RenderCommand::SetBlendColor(*color)
     }
+
+    fn push_vs_constants(
+        &mut self, layout_id: native::UniquePipelineLayoutId, buffer_id: Option<u32>
+    ) -> Option<soft::RenderCommand> {
+        self.state.resources_vs.push_constants_buffer_id = buffer_id;
```
grovesNL: We should be able to collapse these functions into one and `match` on stage instead

src/backend/metal/src/command.rs
```diff
@@ -1738,14 +1837,19 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
 
     fn bind_graphics_pipeline(&mut self, pipeline: &native::GraphicsPipeline) {
         let pipeline_state = pipeline.raw.to_owned();
+        println!("{:?}", pipeline.rasterizer_state);
```
grovesNL: We should remove this

src/backend/metal/src/command.rs
```diff
@@ -2334,22 +2511,43 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+        offset: buffer::Offset,
+        count: DrawCount,
+        stride: u32,
     ) {
-        unimplemented!()
+        let commands = (0 .. count)
+            .map(|i| soft::RenderCommand::DrawIndirect {
+                primitive_type: self.state.primitive_type,
+                buffer: buffer.raw.clone(),
+                offset: offset + (i * stride) as buffer::Offset,
```
grovesNL: Should we assert here (and below)?
> Must be a multiple of 4 bytes.

src/backend/metal/src/command.rs
```diff
@@ -2385,21 +2583,69 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-        _offset: u32,
-        _constants: &[u32],
+        layout: &native::PipelineLayout,
+        stages: pso::ShaderStageFlags,
+        offset: u32,
+        constants: &[u32],
     ) {
-        unimplemented!()
+        {
+            assert_eq!(offset & 3, 0);
```
grovesNL: Maybe `ALIGN_MASK` like we did elsewhere instead of `3`

src/backend/metal/src/device.rs
```diff
@@ -363,6 +372,8 @@ impl Device {
 
         // compile with options
         let mut compiler_options = msl::CompilerOptions::default();
+        compiler_options.enable_point_size_builtin = primitive_class == MTLPrimitiveTopologyClass::Point;
+        compiler_options.resolve_specialized_array_lengths = true;
```
grovesNL: Is it fine for portability to have this requirement? Basically this means we can't allow specialization constants to define array length.

src/backend/metal/src/command.rs
```diff
@@ -159,14 +159,19 @@ struct State {
     resources_cs: StageResources,
     index_buffer: Option<IndexBuffer>,
     attribute_buffer_index: usize,
+    rasterizer_state: Option<native::RasterizerState>,
     depth_stencil_state: Option<metal::DepthStencilState>,
+    push_constant_data: HashMap<native::UniquePipelineLayoutId, Vec<u32>>,
```
msiglreith: I'm a bit skeptical of this design here:
* How does it handle the case where a user uses the **same** layout for compute and graphics, but pushes different constant values?
* I don't see the point to store data per layout. One Vec per bind point should be enough ideally, which should also make pipeline inheritance easier (not handled so far as I can see)

src/backend/metal/src/command.rs
```diff
@@ -2302,6 +2471,10 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         vertices: Range<VertexCount>,
         instances: Range<InstanceCount>,
     ) {
+        if instances.start >= instances.end {
```
msiglreith: assert/debug_assert here?

src/backend/metal/src/command.rs
kvark: I love that kind of critique :)

> How does it handle the case where a user uses the same layout for compute and graphics, but pushes different constant values?

I don't know how this is needed to be handled, hence was asking you on gitter :) Are push constants associated with the (pipeline layout, bind point) pair? I thought it's just (pipeline layout), given that `vkCmdPushConstants` doesn't specify the bind point.

So in my understanding (and the current implementation) if the user pushes different constant values for compute, they will get them for graphics as well. Please tell me if it's supposed to work differently.

> I don't see the point to store data per layout. One Vec per bind point should be enough ideally, which should also make pipeline inheritance easier (not handled so far as I can see)

The idea was to handle the following use case:
  1. user binds a pipeline using layout A
  2. user pushes constants to layout B
  3. user binds a pipeline using layout B

If we only store data per bind point, we'd not be able to keep track of data associated with layouts not being bound.

> make pipeline inheritance easier (not handled so far as I can see)

Yeah, I haven't looked into inheritance really, yet. It should be fine to follow-up.

src/backend/metal/src/command.rs
kvark: I'm a big DRY proponent, and I wasn't able to convince myself that this is a case to merge. Many things are different:
  - `stage` (this is most trivial)
  - `resources_XX` field
  - `soft::XxxCommand` command, which also happens to be the return type

So we could match by stage, but we couldn't return a different type based on that anyway.

src/backend/metal/src/command.rs
kvark: oops! :sweat_smile:

src/backend/metal/src/command.rs
kvark: This is valid from Vulkan point of view (to have 0 instances). Or do you suggest catching the case of `instances.start > instances.end` with an assert? Given the amount of ranges we use, adding asserts to all of them looks like an overkill.

src/backend/metal/src/command.rs
kvark: yep, will do

src/backend/metal/src/command.rs
kvark: sure!

src/backend/metal/src/device.rs
kvark: This is only based on the defaults, so no functional change. I haven't looked into semantics, and we should definitely take advantage of it in later changes.

src/backend/metal/src/command.rs
msiglreith: I see, I though this was some sort of error handling. Didn't know that the spec allows that.

src/backend/metal/src/command.rs
msiglreith: >  if the user pushes different constant values for compute, they will get them for graphics as well. Please tell me if it's supposed to work differently.

Double checked the spec on this and this appears to be correct in case of push constants as the user needs to specify all shader flags, which means both will be set anyway at the same time.

> If we only store data per bind point, we'd not be able to keep track of data associated with layouts not being bound.

I think I should have explained it a bit more in detail. Up to my understand binding descriptor sets  works like this:
We have a small contiguous range of memory which stores push constant data or descriptor table pointers. This memory range is hidden from us and duplicated for each of the bind points (one graphics and one compute). On an action command (draw/dispatch) the shaders will use the layout of the current pipeline for interpreting and fetching the user data (constants and descriptor table pointers) from the associated memory range. When binding a descriptor set we need the layout that the driver can decide at which point in the memory range the tables should be written. Changing the layout during bind calls has no real influence beside that your data might be messed up/invalidates. Pipeline layout compatibility specifies which bindings are still valid. Therefore a pipeline layout should just be seen of an interpretation scheme (and abstraction layer).

In an example:
```
    user binds descriptors to layout A
    user binds descriptors to layout B
    user binds a pipeline using layout B
    draw
    user binds a pipeline using layout A
    draw
```
The 2nd call will overwrite (usually) the binding of layout A. Which means the users need to rebind the descriptor sets to layout A before the draw call due to invalidation. If A and B are compatible (up to the descriptor sets by A) then no rebinding is required.

src/backend/metal/src/command.rs
grovesNL: Hmm yeah fair enough. I was thinking we could pass `resources_XX` as an argument, and perform the push to `commands` directly in these functions (and not return anything at all), but that has some downsides too.

src/backend/metal/src/command.rs
kvark: In hindsight, we should just check for equality here. Negative ranges are going to panic automatically on our computation of the count (in debug, at least).

kvark:
(merging as per gitter chat confirmation)
bors r=msiglreith

PR #2071, opened 2018-05-28
Remaping descriptor sets in the gl backend

ZeGentzy:
I'll rebase off master when I'm done.

Uniforms in gl only have a bindings field, not a set one. This means that for shaders that use multiple sets to work, we must change where we are binding them.

See page 14 for what I mean: https://www.khronos.org/assets/uploads/developers/library/2016-vulkan-devday-uk/4-Using-spir-v-with-spirv-cross.pdf

PR checklist:

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -3,7 +3,7 @@ use std::cell::Cell;
 use std::iter::repeat;
 use std::ops::Range;
 use std::{ptr, mem, slice};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, RwLock};
 
 use gl;
 use gl::types::{GLint, GLenum, GLfloat};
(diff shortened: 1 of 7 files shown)


ZeGentzy:
Breaks the vulkan example and I got no clue why. Otherwise, I think this pull request is ready

kvark:
> Merge branch 'master' into desc-remaping-gl 

please rebase instead of merging

examples/hal/quad/main.rs
```diff
@@ -449,15 +438,7 @@ fn main() {
...
-                pso::Descriptor::Image(&image_srv, i::Layout::Undefined)
-            ),
-        },
-        pso::DescriptorSetWrite {
-            set: &desc_set,
-            binding: 1,
-            array_offset: 0,
-            descriptors: Some(
-                pso::Descriptor::Sampler(&sampler)
+                pso::Descriptor::CombinedImageSampler(&image_srv, i::Layout::ShaderReadOnlyOptimal, &sampler)
```
kvark: looks like we had a bug in here by specifying `Undefined`?

src/backend/gl/src/native.rs
```diff
@@ -15,13 +17,16 @@ pub type FrameBuffer = gl::types::GLuint;
 pub type Surface     = gl::types::GLuint;
 pub type Texture     = gl::types::GLuint;
 pub type Sampler     = gl::types::GLuint;
+pub(crate) type SetID     = gl::types::GLuint;
```
kvark: the whole module is private to the crate, so no need to have `pub(crate)` for all the things here, just `pub` should be good

src/backend/gl/src/native.rs
```diff
@@ -15,13 +17,16 @@ pub type FrameBuffer = gl::types::GLuint;
...
+pub(crate) type BindingID     = gl::types::GLuint;
```
kvark: note that we already have `hal::pso::DescriptorSetIndex` and `hal::pso::DescriptorBinding`. Maybe we could use them here instead of defining new types?

src/backend/gl/src/native.rs
```diff
@@ -38,6 +43,79 @@ impl Fence {
     }
 }
 
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+pub(crate) enum BindingTypes {
```
kvark: sounds like we could just use `hal::pso::DecriptorType` instead of this enum

src/backend/gl/src/native.rs
```diff
@@ -38,6 +43,79 @@ impl Fence {
...
+    StorageBuffers,
+    StorageImages,
+    SampledImages,
+    SeparateImages,
+    SeparateSamplers,
+    SubpassInputs,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct DescRemapData {
```
kvark: please add comments specifying what this is needed for

src/backend/gl/src/native.rs
```diff
@@ -38,6 +43,79 @@ impl Fence {
...
+    bindings: HashMap<BindingTypes, HashMap<SetID, HashMap<BindingID, BindingID>>>,
```
kvark: this could benefit from squashing all 3 hashmaps into a single one:
```rust
bindings: HashMap<(pso::DescriptorType, pso::DescriptorSetIndex, pso::DescriptorBinding), pso::DescriptorBinding>,
```

src/backend/gl/src/native.rs
```diff
@@ -38,6 +43,79 @@ impl Fence {
...
+impl DescRemapData {
+    pub(crate) fn new() -> Self {
+        DescRemapData {
+            bindings: HashMap::new(),
+            names: HashMap::new(),
+            next_binding: 0,
+        }
+    }
+
+    pub(crate) fn get_binding_insert(&mut self, btype: BindingTypes, set: SetID, binding: BindingID) -> BindingID {
```
kvark: let's rename the function to something clearer

src/backend/gl/src/native.rs
```diff
@@ -38,6 +43,79 @@ impl Fence {
...
+        if !self.bindings.contains_key(&btype) {
```
kvark: let's use the [Entry API](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry)

src/backend/gl/src/native.rs
```diff
@@ -38,6 +43,79 @@ impl Fence {
...
+
+        *set_map.get(&binding).unwrap()
+    }
+
+    pub(crate) fn get_binding(&self, btype: BindingTypes, set: SetID, binding: BindingID) -> Option<&BindingID> {
+        if !self.bindings.contains_key(&btype) {
+            return None
+        }
+
+        let bindings = self.bindings.get(&btype).unwrap();
```
kvark: similarly, instead of checking for `contains_key` followed to unwrapping we could just chain all the things:
```rust
self.bindings
  .get(&btype)?
  .get(&set)?
  .get(&binding)
```

src/backend/gl/src/native.rs
```diff
@@ -82,10 +160,17 @@ pub enum ImageView {
 }
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
-pub struct DescriptorSetLayout;
+pub(crate) enum DescSetBindings {
+    Buffer(BindingTypes, BindingID, RawBuffer, gl::types::GLintptr, gl::types::GLsizeiptr),
```
kvark: the last parameters are not obvious here, so better have them named, e.g.
```rust
Buffer {
  ty: ...,
  id: ...,
  buffer: ...,
}
```

src/backend/gl/src/device.rs
```diff
@@ -386,11 +451,13 @@ impl d::Device<B> for Device {
...
     where
         IS: IntoIterator,
-        IS::Item: Borrow<n::DescriptorSetLayout>,
+        IS::Item: Borrow<Vec<pso::DescriptorSetLayoutBinding>>,
         IR: IntoIterator,
         IR::Item: Borrow<(pso::ShaderStageFlags, Range<u32>)>,
     {
-        n::PipelineLayout
+        n::PipelineLayout {
+            desc_remap_data: Arc::new(RwLock::new(n::DescRemapData::new())),
```
kvark: If I understand correctly, you chose to have `Arc+RwLock` here because creating a pipeline needs to modify those mappings. This is concerning, actually. Can't we have the remapping table constantly associated with the set of descriptor set layouts? In this case, it would be immutable and filled up right here.

src/backend/gl/src/native.rs
```diff
@@ -82,10 +160,17 @@ pub enum ImageView {
...
+    Texture(BindingID, Texture),
+    Sampler(BindingID, Sampler),
+}
 
-#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
-pub struct DescriptorSet;
+#[derive(Clone, Debug)]
+pub struct DescriptorSet {
+    layout: Vec<pso::DescriptorSetLayoutBinding>,
+    pub(crate) bindings: Arc<RwLock<Vec<DescSetBindings>>>,
```
kvark: to recap, if I understand correctly you want `Arc+RwLock` here because this is an internally mutable state that's changed by `write_descriptor_sets`, correct? I believe we could go for simpler `RefCell` here instead, given that Vulkan spec requires that `write_descriptor_sets` doesn't conflict with any descriptor usage (in flight) or another write/copy of descriptor sets.

src/backend/gl/src/device.rs
```diff
@@ -989,9 +1059,55 @@ impl d::Device<B> for Device {
...
+            let mut offset = write.array_offset as _;
+
+            for descriptor in write.descriptors {
+                match descriptor.borrow() {
+                    pso::Descriptor::Buffer(buffer, ref range) => {
+                        let start = range.start.unwrap_or(0);
+                        let end = range.end.unwrap_or(buffer.size);
+                        let size = (end - start) as _;
+
+                        set.bindings
```
kvark: can we do this locking just once per set (higher up)?

examples/hal/quad/main.rs
ZeGentzy: I only changed it because `ShaderReadOnlyOptimal` is more optimal. No clue if it's a bug.

src/backend/gl/src/native.rs
ZeGentzy: `hal::pso::DecriptorType` has more types then we need. Once more of the binding types get implemented (I've only tested `SampledImages`) I expect a lot of them to get merged together (ex: `SampledImages` & `SeperateSamplers`).

src/backend/gl/src/device.rs
ZeGentzy: I'll look into this when I get back.

ZeGentzy: Ok, I just realized something. The mapping should be owned by the pipeline and not shared between pipelines. If we have two pipelines sharing the same layout but with different shaders then we will get a suboptimal remapping.

The issue with making the pipelines own the remapping is that the pipeline doesn't get passed to `bind_graphics_descriptor_sets,` the pipeline layout does. The pipeline is never passed in the creation of the descriptors sets or their layout either. To store the remap data in it's correct place we'd either have to keep track of the current bound pipeline or pass in the pipeline to `bind_graphics_descriptor_sets`.

We can't store them with the the descriptor set layouts because the layouts are completely unrelated. The remapping is dependent on the binding points and sets set in the shaders of a single pipeline, and is only used when we bind the related descriptor sets after binding that pipeline.

src/backend/gl/src/native.rs
ZeGentzy: Yes, we could if rust wouldn't error with the following message:
```
error[E0277]: `std::cell::RefCell<std::vec::Vec<native::DescSetBindings>>` cannot be shared between threads safely
  --> src/backend/gl/src/lib.rs:44:6
   |
44 | impl hal::Backend for Backend {
   |      ^^^^^^^^^^^^ `std::cell::RefCell<std::vec::Vec<native::DescSetBindings>>` cannot be shared between threads safely
   |
   = help: within `native::DescriptorSet`, the trait `std::marker::Sync` is not implemented for `std::cell::RefCell<std::vec::Vec<native::DescSetBindings>>`
   = note: required because it appears within the type `native::DescriptorSet`
```

Should I just manually implement the Sync trait?

kvark:
No, just wrap it in Arc

ZeGentzy:
Uhh, arc doesn't fix it.

src/backend/gl/src/native.rs
kvark: Ah, sorry, you are correct... `RefCell` is `!Sync`. Let's go with `Arc<Mutex>` then, it should be lighter than `RwLock`

src/backend/gl/src/device.rs
kvark: > If we have two pipelines sharing the same layout but with different shaders then we will get a suboptimal remapping.

This is fine by me. Pipeline layout *is* that mapping, semantically speaking. We shouldn't go out of our way to make it more efficient than the API suggests.

ZeGentzy:
Changed RefCell to mutex. 
Added `_` before the unused variables at src/backend/gl/src/device.rs:1086 to 1104.

src/backend/gl/src/lib.rs
```diff
@@ -70,7 +70,7 @@ impl hal::Backend for Backend {
     type ComputePipeline = native::ComputePipeline;
     type GraphicsPipeline = native::GraphicsPipeline;
     type PipelineLayout = native::PipelineLayout;
-    type DescriptorSetLayout = native::DescriptorSetLayout;
+    type DescriptorSetLayout = Vec<pso::DescriptorSetLayoutBinding>;
```
kvark: if you do need it to just be a vec, let's do `pub type DescriptorSetLayout = Vec<pso::...>;` in `native` module and use here and elsewhere consistently

ZeGentzy:
Was hoping to get seperate image and samplers implemented too, but it would be ok to do them in a seperate PR.

kvark:
@ZeGentzy please make commits more informative (e.g. not just "Requested changes") and address the CI complains to proceed.

src/backend/gl/src/info.rs
```diff
@@ -226,6 +226,8 @@ bitflags! {
         const SAMPLER_LOD_BIAS = 0x2000;
         /// Support setting border texel colors.
         const SAMPLER_BORDER_COLOR = 0x4000;
+        /// No explicit layouts in shader support
+        const NO_EXPLICIT_LAYOUTS_IN_SHADER = 0x8000;
```
kvark: let's not use negative flags - you can see that everything else is positive

src/backend/gl/src/device.rs
kvark: I think this is still not the way I imagine it should be. The pipeline layout should not have internal mutability. When it's created, it has all the information required to map from descriptor binding/sets to GL resource slots, and that's the mapping it should carry. All the actual pipeline construction should *use* that mapping but not modify it.

ZeGentzy:
A new issue has come to my attention, when build_combined_image_shaders() is called internally the binding number gets reset, so this will have to change until I add support seperate samplers and images.

ZeGentzy:
@kvark While this implementation of remaping and binding samplers isn't the most optimized (including the lack of the 1 sampler to image perf limitation discussed on gitter), it works. Since we don't change the exposed api or break any old behavior, please consider merging this as it is currently and leaving the perf optimizations latter.

src/backend/gl/src/device.rs
```diff
@@ -423,7 +588,7 @@ impl d::Device<B> for Device {
                 .iter()
                 .filter_map(|&(stage, point_maybe)| {
                     point_maybe.map(|point| {
-                        let shader_name = self.compile_shader(point, stage);
+                        let shader_name = self.compile_shader(point, stage, &mut *desc.layout.desc_remap_data.write().unwrap());
```
kvark: we shouldn't need those `&mut *` constructs here

grovesNL:
Needs rebasing to get fast hash map changes from #2150:
```
329 |         let mut id_map = HashMap::<u32, (pso::DescriptorSetIndex, pso::DescriptorBinding)>::new();
    |                          ^^^^^^^ Use of undeclared type or module `HashMap`
```

You should be able to simply replace these with `FastHashMap` (from `hal::backend::FastHashMap`)

ZeGentzy:
Sorry for the delay, I had to fix my computer. Two changes above.

kvark:
It's good to go once https://github.com/rust-lang/rust/issues/51699 is resolved...

ZeGentzy:
@kvark Should be working now.

PR #2076, opened 2018-05-28
Metal revised memory types

### LINKED ISSUE #2069: Coherent memory on Metal
Here is the strategy (discussed with VkPI):
  - have one shared buffer backing a coherent memory allocation
  - represent buffers in coherent memory as just a range
  - represent textures as aliases on that memory buffer. The restrictions put in place for the mappable textures (linear tiling, 1D, no mipmaps) effectively make it compatible with buffer reinterpretation.

kvark:
Fixes #2069
~~Includes #2070 (please only review the last 2 commits)~~

The PR effectively removes all the complicated memory allocation tracking from the Metal backend, instead relying on mapping the available storage types (with CPU caching policy) more accurately to what can be represented in Vulkan API.

With this in, we slice through the whole `memory` section without crashes. The only issue I failed to fix/workaround is filed as #2077

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -36,18 +36,6 @@ pub struct QueuePool {
     queues: Vec<QueueInner>,
 }
 
-pub struct CommandBufferScope<'a> {
-    inner: &'a metal::CommandBufferRef,
-    _pool: AutoreleasePool,
-}
-
-impl<'a> Deref for CommandBufferScope<'a> {
-    type Target = metal::CommandBufferRef;
-    fn deref(&self) -> &Self::Target {
-        self.inner
-    }
-}
-
 impl QueuePool {
     fn find_queue(&mut self, device: &Mutex<metal::Device>) -> usize {
         const POOL_SIZE: usize = 64;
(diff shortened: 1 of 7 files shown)


src/backend/metal/src/lib.rs
```diff
@@ -65,6 +66,7 @@ impl Shared {
     pub(crate) fn new(device: metal::Device) -> Self {
         Shared {
             queue_pool: Mutex::new(command::QueuePool::default()),
+            aux_queue: Mutex::new(device.new_command_queue()),
```
grovesNL: What happens if we exceed `64` temporary command buffers?

kvark: I don't think this is a problem, since we commit those command buffers right after acquiring them. So in the worst case there is going to be a bit of a delay for GPU to finish.

kvark:
Thanks for taking a look! There appear to be no major concerns.
bors r=grovesNL

src/backend/metal/src/command.rs
```diff
@@ -1103,7 +1103,7 @@ impl CommandBuffer {
         while data.len() < offset + constants.len() {
```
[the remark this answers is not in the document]
grovesNL: 👍 I didn't notice the clear in `reset_resources`

src/backend/metal/src/device.rs
```diff
@@ -1857,10 +1857,15 @@ impl hal::Device<Backend> for Device {
     }
 
     fn wait_idle(&self) -> Result<(), error::HostExecutionError> {
-        self.shared.queue_pool
-            .lock()
-            .unwrap()
-            .wait_idle(&self.shared.device);
+        debug!("waiting for idle");
```
grovesNL: Any reason not to share the `wait_idle` implementation across device and command?

src/backend/metal/src/lib.rs
grovesNL: I guess it depends on how many active deferred commands buffers we actually expect in real use cases, i.e. if it becomes hundreds in practice it would be unfortunate to hit that delay often IMO

kvark: Hmm. Theoretically, it can become an issue, but it's too early to worry about this right now

src/backend/metal/src/device.rs
kvark: it could be shared, yes

PR #2078, opened 2018-05-29
[mtl] vertex buffer remapping

### LINKED ISSUE #2054: Support vertex attribute offsets higher than the strides in Metal
There is a lot of test in CTS (e.g. "dEQP-VK.texture.filter.**") that try to use an attribute offset higher than the stride. For example, an offset of 64 on a buffer with stride 8. It is technically valid in Vulkan, but Metal doesn't like it.

One possible solution would be to detect this at pipeline creation *before* linking the shader, then assign additional buffer inputs to be used in the shader by those attributes... adjusting the binding model accordingly.

kvark:
Fixes #2054
PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -3,7 +3,7 @@ use {conversions as conv, command, native as n};
 use internal::Channel;
 
 use std::borrow::Borrow;
-use std::collections::HashMap;
+use std::collections::hash_map::{Entry, HashMap};
 use std::ops::Range;
 use std::path::Path;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/device.rs
```diff
@@ -887,61 +886,93 @@ impl hal::Device<Backend> for Device {
...
+        for (i, &pso::AttributeDesc { binding, element, ..}) in pipeline_desc.attributes.iter().enumerate() {
+            let original = pipeline_desc.vertex_buffers
+                .iter()
+                .find(|vb| vb.binding == binding)
+                .expect("no associated vertex buffer found");
+            // handle wrapping offsets
+            let elem_size = element.format.surface_desc().bits as pso::ElemOffset >> 3;
+            let (cut_offset, base_offset) = if element.offset < original.stride {
+                (element.offset, 0)
+            } else if element.offset + elem_size <= original.stride {
```
grovesNL: This case isn't clear to me, shouldn't it be greater than the stride if it's wrapping?

src/backend/metal/src/device.rs
```diff
@@ -887,61 +886,93 @@ impl hal::Device<Backend> for Device {
...
+                let remainder = element.offset % original.stride;
+                (remainder, element.offset - remainder)
+            } else {
+                (0, element.offset)
+            };
+            let mtl_buffer_index = match vertex_buffer_map.entry((binding, base_offset)) {
+                Entry::Vacant(_) if next_buffer_index == self.shared.push_constants_buffer_id => {
+                    error!("Attribute offset {} exceeds the stride {}, and there is no room for replacement.",
```
grovesNL: Feels a bit weird to fail here for push constants here IMO, isn't it our fault if we can't remap with enough room?

kvark:
bors r=grovesNL

kvark:
Bors retry

PR #2079, opened 2018-05-29
Immutable samplers

### LINKED ISSUE #1723: Immutable samplers
Vulkan allows defining immutable samplers on descriptor set layout creation.
Need to investigate how to map these to other backends.

kvark:
Fixes #1723
- [x] hal/empty/render/warden
- [x] metal
- [ ] d3d12
- [x] vulkan

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -1094,13 +1094,21 @@ impl d::Device<B> for Device {
         }
     }
 
-    fn create_descriptor_set_layout<T>(
-        &self, binding_iter: T
-    )-> n::DescriptorSetLayout
+    fn create_descriptor_set_layout<I, J>(
+        &self, binding_iter: I, immutable_sampler_iter: J
+    ) -> n::DescriptorSetLayout
     where
-        T: IntoIterator,
-        T::Item: Borrow<pso::DescriptorSetLayoutBinding>,
+        I: IntoIterator,
+        I::Item: Borrow<pso::DescriptorSetLayoutBinding>,
+        J: IntoIterator,
+        J::Item: Borrow<n::Sampler>,
     {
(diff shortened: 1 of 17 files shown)


kvark:
I don't want to ship the DX11/12 support with this PR, can go as follow uo

src/hal/src/pso/descriptor.rs
```diff
@@ -74,7 +74,8 @@ pub struct DescriptorSetLayoutBinding {
     pub count: DescriptorArrayIndex,
     /// Valid shader stages.
     pub stage_flags: ShaderStageFlags,
-    // TODO: immutable samplers?
+    /// Use the associated list of immutable samplers.
+    pub immutable_samplers: bool,
```
msiglreith: where is this needed?

msiglreith: Resolved in internal discussion.
Vulkan takes a null ptr or array of samplers. Works basically like a Option, but combining the samplers into one iterator.

src/backend/vulkan/src/device.rs
```diff
@@ -1113,7 +1121,13 @@ impl d::Device<B> for Device {
                 descriptor_type: conv::map_descriptor_type(b.ty),
                 descriptor_count: b.count as _,
                 stage_flags: conv::map_stage_flags(b.stage_flags),
-                p_immutable_samplers: ptr::null(), // TODO
+                p_immutable_samplers: if b.immutable_samplers {
+                    let slice = &immutable_samplers[sampler_offset..];
```
grovesNL: Why don't we need `sampler_offset + b.count` here to match the others?

kvark:
Because all we care about here is a pointer

kvark:
Thanks for the reviews!
Appears to be no blockers.
Bors r=msiglreith,grovesNL

PR #2084, opened 2018-05-30
Add free descriptor sets API

### LINKED ISSUE #2045: Free descriptor sets API
Our `DescriptorPool` trait in HAL is missing the `free` function, analogous to `vkFreeDescriptorSets`.

Xaeroxe:
Fixes #2045 

Very untested.  Do not merge until I've had time to test this on all backends.

PR checklist:
  - [x] Windows Vulkan
  - [x] Windows OpenGL
  - [x] DirectX 12
  - [ ] Metal
  - [x] OSX OpenGL
  - [x] Linux Vulkan
  - [x] Linux OpenGL


Xaeroxe:
Doesn't compile on metal.  Guess that's what happens when you try to write code for OSX on Windows haha, I'll fix it when I get to an OSX device.

src/backend/dx12/src/native.rs
```diff
@@ -401,22 +404,96 @@ pub struct DescriptorHeapSlice {
...
+                        if self.empty_ranges[i + 1].start == self.empty_ranges[i].end {
+                            self.empty_ranges[i].end += self.empty_ranges[i + 1].end - self.empty_ranges[i + 1].start;
+                            self.empty_ranges.remove(i + 1);
+                        }
+                    }
+                    break;
+                }
+
+                // We're inbetween two handles
+                if i + 1 != self.empty_ranges.len() {
```
kvark: what happens if the loop finishes and we haven't returned yet?

src/backend/dx12/src/native.rs
```diff
@@ -497,8 +574,51 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
         })
     }
 
+    fn free_sets(&mut self, descriptor_sets: &[DescriptorSet]) {
+        for descriptor_set in descriptor_sets {
+            for binding_info in &descriptor_set.binding_infos {
+                if let Some(ref view_range) = binding_info.view_range {
+                    let has_view = match view_range.ty {
```
kvark: let's have a helper method returning, say, 2 booleans over the descriptor type?

src/backend/dx12/src/native.rs
```diff
@@ -622,6 +600,25 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
     }
 }
 
+/// Returns DescriptorType properties for DX12.
+/// 
+/// Return value is in the form (has_view, has_sampler, is_uav)
+fn get_heap_properties(desc_type: pso::DescriptorType) -> (bool, bool, bool) {
```
kvark: nit: let's return a struct with 3 named booleans

src/backend/dx12/src/native.rs
```diff
@@ -513,19 +525,7 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
...
-                pso::DescriptorType::SampledImage |
-                pso::DescriptorType::UniformTexelBuffer |
-                pso::DescriptorType::UniformBuffer => (true, false, false),
-                pso::DescriptorType::StorageImage |
-                pso::DescriptorType::StorageTexelBuffer |
-                pso::DescriptorType::StorageBuffer => (true, false, true),
-                _ => unimplemented!()
-            };
-
+            let (has_view, has_sampler, is_uav) = get_heap_properties(binding.ty);
```
kvark: perhaps, make it to be `binding.heap_properties()`?

Xaeroxe:
@kvark how is this?

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+            }
+            match best_range_tuple {
+                Some((ref mut index, ref mut best_range)) => {
+                    // We want to find the best fit for this allocation in order to reduce 
+                    // memory fragmentation, so we'll look for it here.
+                    if range.1.end - range.1.start < best_range.end - best_range.start {
+                        *index = range.0;
+                        *best_range = range.1.clone();
+                        // Found a perfect fit, so stop looking.
+                        if range.1.end - range.1.start == 0 {
```
kvark: should this check for `==count` instead of `0` ?

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+                            break;
+                        }
+                    }
+                },
+                None => { 
+                    best_range_tuple = Some(range);
```
kvark: could also check for it being a perfect fit

src/backend/dx12/src/native.rs
```diff
@@ -370,6 +372,7 @@ impl DescriptorHeap {
         DualHandle {
             cpu: d3d12::D3D12_CPU_DESCRIPTOR_HANDLE { ptr: self.start.cpu.ptr + (self.handle_size * index) as usize },
             gpu: d3d12::D3D12_GPU_DESCRIPTOR_HANDLE { ptr: self.start.gpu.ptr + self.handle_size * index },
+            size: 0,
```
kvark: let's pass `size` as an argument here

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+            }
         }
+        best_range_tuple.map(|range| {
+            if range.1.end - range.1.start == count {
+                self.empty_ranges.remove(range.0);
+            } else {
+                self.empty_ranges[range.0].start += count;
+            }
+            DualHandle {
+                cpu: d3d12::D3D12_CPU_DESCRIPTOR_HANDLE { ptr: self.start.cpu.ptr + (self.handle_size * range.1.start) as usize },
```
kvark: let's use `self.at(range.1.start, count)` here

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+                self.empty_ranges[i].start -= handle_range.end - handle_range.start;
+
+                // Merge this into an adjacent range to the left if necessary.
+                if i > 0 {
+                    if self.empty_ranges[i - 1].end == self.empty_ranges[i].start {
+                        self.empty_ranges[i - 1].end += self.empty_ranges[i].end - self.empty_ranges[i].start;
+                        self.empty_ranges.remove(i);
+                    }
+                }
+                freed = true;
```
kvark: can't we just return instead of having the boolean + `break` combo?

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+            // Handle is immediately to the right of an existing empty range.
+            if handle_range.start == self.empty_ranges[i].end {
+                // Extend this range
+                self.empty_ranges[i].end += handle_range.end - handle_range.start;
+
+                // Merge this into an adjacent range to the right if necessary.
+                if i + 1 != self.empty_ranges.len() {
+                    if self.empty_ranges[i + 1].start == self.empty_ranges[i].end {
+                        self.empty_ranges[i].end += self.empty_ranges[i + 1].end - self.empty_ranges[i + 1].start;
+                        self.empty_ranges.remove(i + 1);
```
kvark: let's combine this into
```rust
let r = self.empty_ranges.remove(i + 1);
self.empty_ranges[i].end += r.end - r.start;
```

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+            // Handle is immediately to the left of an existing empty range.
+            if handle_range.end == self.empty_ranges[i].start {
+                // Extend this range
+                self.empty_ranges[i].start -= handle_range.end - handle_range.start;
+
+                // Merge this into an adjacent range to the left if necessary.
+                if i > 0 {
+                    if self.empty_ranges[i - 1].end == self.empty_ranges[i].start {
+                        self.empty_ranges[i - 1].end += self.empty_ranges[i].end - self.empty_ranges[i].start;
+                        self.empty_ranges.remove(i);
```
kvark: similarly, can piggy-back on `remove` returning the value here

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+    /// Free handles previously given out by this `DescriptorHeapSlice`.  Do not use this with handles not given out by this `DescriptorHeapSlice`.
+    pub(crate) fn free_handles(&mut self, handle: DualHandle) {
+        let start = handle.gpu.ptr - self.start.gpu.ptr / self.handle_size;
+        let handle_range = start..start + handle.size as u64;
+        let mut freed = false;
+        for i in 0..self.empty_ranges.len() {
+            // Handle is immediately to the left of an existing empty range.
+            if handle_range.end == self.empty_ranges[i].start {
+                // Extend this range
+                self.empty_ranges[i].start -= handle_range.end - handle_range.start;
```
kvark: shouldn't this just be `self.empty_ranges[i].start = handle_range.start;`?

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+
+                // Merge this into an adjacent range to the left if necessary.
+                if i > 0 {
+                    if self.empty_ranges[i - 1].end == self.empty_ranges[i].start {
+                        self.empty_ranges[i - 1].end += self.empty_ranges[i].end - self.empty_ranges[i].start;
```
kvark: similarly, this could just be `self.empty_ranges[i - 1].end = self.empty_ranges[i].end;`

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+                    }
+                }
+                freed = true;
+                break;
+            }
+
+            // Handle is immediately to the right of an existing empty range.
+            if handle_range.start == self.empty_ranges[i].end {
+                // Extend this range
+                self.empty_ranges[i].end += handle_range.end - handle_range.start;
```
kvark: `self.empty_ranges[i].end = handle_range.end;`

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+
+                // Merge this into an adjacent range to the right if necessary.
+                if i + 1 != self.empty_ranges.len() {
+                    if self.empty_ranges[i + 1].start == self.empty_ranges[i].end {
+                        self.empty_ranges[i].end += self.empty_ranges[i + 1].end - self.empty_ranges[i + 1].start;
```
kvark: `self.empty_ranges[i].end = self.empty_ranges[i+1].end;`

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+        })
+    }
+
+    /// Free handles previously given out by this `DescriptorHeapSlice`.  Do not use this with handles not given out by this `DescriptorHeapSlice`.
+    pub(crate) fn free_handles(&mut self, handle: DualHandle) {
+        let start = handle.gpu.ptr - self.start.gpu.ptr / self.handle_size;
+        let handle_range = start..start + handle.size as u64;
+        let mut freed = false;
+        for i in 0..self.empty_ranges.len() {
+            // Handle is immediately to the left of an existing empty range.
```
kvark: assert/debug_assert that we don't actually intersect with any existing empty ranges?

src/backend/dx12/src/native.rs
```diff
@@ -401,21 +404,107 @@ pub struct DescriptorHeapSlice {
...
+                    }
+                }
+                freed = true;
+                break;
+            }
+
+            // We're inbetween two handles
+            if i + 1 != self.empty_ranges.len() {
+                // Handle is inbetween two empty ranges.
+                if handle_range.start > self.empty_ranges[i].end && handle_range.end < self.empty_ranges[i + 1].start {
```
kvark: could combine with the wrapping condition (that checks for `i+1`)

src/backend/dx12/src/native.rs
```diff
@@ -497,8 +574,68 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
         })
     }
 
+    fn free_sets(&mut self, descriptor_sets: &[DescriptorSet]) {
+        for descriptor_set in descriptor_sets {
+            for binding_info in &descriptor_set.binding_infos {
+                if let Some(ref view_range) = binding_info.view_range {
+                    let HeapProperties { has_view, .. } = view_range.ty.heap_properties();
+                    if has_view {
```
kvark: nit: let's just do `if view_range.ty.heap_properties().has_view {..}`
the destructuring code is not needed here, just adds noise

src/backend/dx12/src/native.rs
```diff
@@ -497,8 +574,68 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
...
+                        self.heap_srv_cbv_uav.free_handles(view_range.handle);
+                    }
+                    
+                }
+                if let Some(ref sampler_range) = binding_info.sampler_range {
+                    let HeapProperties { has_sampler, .. } = sampler_range.ty.heap_properties();
+                    if has_sampler {
```
kvark: same here

src/backend/dx12/src/native.rs
```diff
@@ -497,8 +574,68 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
...
+    }
+}
+
+struct HeapProperties {
+    has_view: bool,
+    has_sampler: bool,
+    is_uav: bool,
+}
+
+trait HasHeapProperties {
```
kvark: sorry, I didn't expect us to come up with an internal trait here (wasn't sure if the types are from this crate - they are not).
mind if we go back to just having a method then? Perhaps, `HeapProperties::from(DescriptorType)`?

src/backend/metal/src/native.rs
```diff
@@ -219,30 +222,130 @@ impl hal::DescriptorPool<Backend> for DescriptorPool {
...
                 let (encoder, stage_flags) = match layout {
                     &DescriptorSetLayout::ArgumentBuffer(ref encoder, stages) => (encoder, stages),
                     _ => return Err(pso::AllocationError::IncompatibleLayout),
                 };
-
-                let cur_offset = *offset;
-                *offset += encoder.encoded_length();
-                if *offset <= total_size {
-                    Ok(DescriptorSet::ArgumentBuffer {
+                let mut best_range_tuple: Option<(usize, Range<NSUInteger>)> = None;
```
kvark: This is an unfortunately large code duplication between backends. We should find a way to share the code, also take that opportunity to add unit tests for it. I think the easiest way would be, right now, to have `src/backend/aux` *module* (not crate) that is included in both backends by doing
```rust
#[path = "../aux/range_pool.rs"]
mod range_pool;
```

src/backend/dx12/src/native.rs
Xaeroxe: The `at` function doesn't exist for `DescriptorHeapSlice`, it only exists for `DescriptorHeap`.  Should I add an `at` function?

kvark: damnit, you are correct. No need to move it out as long as that's the only spot it'd be used in

src/backend/metal/src/native.rs
Xaeroxe: Phew the unit tests were a good idea to say the least.

Xaeroxe:
Squashed

Xaeroxe:
@kvark ready for re-review

src/backend/dx12/src/native.rs
```diff
@@ -497,8 +496,60 @@ impl HalDescriptorPool<Backend> for DescriptorPool {
...
+            pso::DescriptorType::Sampler => HeapProperties::new(false, true, false),
+            pso::DescriptorType::CombinedImageSampler => HeapProperties::new(true, true, false),
+            pso::DescriptorType::InputAttachment |
+            pso::DescriptorType::SampledImage |
+            pso::DescriptorType::UniformTexelBuffer |
+            pso::DescriptorType::UniformBuffer => HeapProperties::new(true, false, false),
+            pso::DescriptorType::StorageImage |
+            pso::DescriptorType::StorageTexelBuffer |
+            pso::DescriptorType::StorageBuffer => HeapProperties::new(true, false, true),
+            _ => unimplemented!()
```
kvark: nit: let's list those explicitly here

msiglreith:
Main concern is why not use the existing free list implementation in the dx12 backend?

Xaeroxe:
@msiglreith can you point me to it?

msiglreith:
Sure: https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12/src/free_list.rs

Xaeroxe:
Ah... that seems obvious in retrospect :P I'll have a look

Xaeroxe:
@msiglreith Well the free list implementation shown isn't a proper free list because it uses a separate `LinkedList` structure, rather than storing the list in the buffer being allocated, so we lost that benefit.  Meaning the only major distinction between that implementation and mine is that I get better cache usage out of a Vec, and I generally have to do fewer memory allocations.

Small differences, but differences nonetheless.  Honestly I'm wondering if I ought to replace the uses of this free_list with uses of the `RangeAllocator` instead.

kvark:
Thinking a bit some more, I got the following concerns:
  1. Correctness: according to the [spec](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkAllocateDescriptorSets.html), allocation may fail due to exceeding either the maximum set count or the sum of descriptor counts for a type. Fragmentation is *not* a valid reason for a failed allocation.
  2. Conflict with existing freelist implementation in d3d12 backend. We need to carefully evaluate which one works best.
  3. Sharing: technically speaking, Metal can get away (and probably should) without this allocator. Only the argument buffer path currently uses it, and we can rewrite it to have a buffer per descriptor set as opposed to - per descriptor pool, efficiently treating each descriptor set separately (and never failing an allocation).

Xaeroxe:
In response to

1. VK_ERROR_FRAGMENTED_POOL is a code that can be returned by that function.  It's not really possible to prevent this type of error unless we want to start using smart pointers as handles and modifying the internal values during allocation.  Avoiding this kind of error means we need the ability to shift around existing allocations.

2. Sure, I'll benchmark these

3. Sounds reasonable, although it does sort of defy the notion of a "Pool" being memory you don't have to request from the OS.

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
+use std::ops::{Add, AddAssign, Range, Sub};
+use std::fmt::Debug;
+
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub struct RangeAllocator<T> {
+    /// The range this allocator covers.
+    range: Range<T>,
```
msiglreith: Maybe something like 'initial_range'  or ' space' to avoid repeating 'range' so often?

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// A Vec of ranges in this heap which are unused.
+    /// Must be ordered with ascending range start to permit short circuiting allocation.
+    /// No two ranges in this vec may overlap.
+    empty_ranges: Vec<Range<T>>,
```
msiglreith: 'unused' or 'free'  instead of empty?

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+impl<T> RangeAllocator<T>
+    where T: Clone + Copy + Add<Output=T> + AddAssign + Sub<Output=T> + Eq + PartialOrd + Debug
```
msiglreith: nit formatting:
```Rust
impl Struct
where
    T: ...
{
```

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+{
+    pub fn new(range: Range<T>) -> Self {
+        RangeAllocator {
+            range: range.clone(),
+            empty_ranges: vec![range.clone()],
+        }
+    }
+
+    pub fn range(&self) -> Range<T> {
```
msiglreith: Same as above, I guess a better name should be chosen here

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.range.clone()
+    }
+
+    pub fn allocate_range(&mut self, length: T) -> Option<Range<T>> {
+        let mut best_range_tuple: Option<(usize, Range<T>)> = None;
```
msiglreith: `best_fit`?

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+        for range in self.empty_ranges.iter().cloned().enumerate() {
```
msiglreith: unpack range already into `(index, range)` to avoid all these `range.0/.1`s

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+            if range.1.end - range.1.start < length {
```
msiglreith: nit: start + length < end might be easier to read

msiglreith: Or calculate the size of the current range once and store it in a variable. Seem to be recalculated multiple times below

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+            }
+            match best_range_tuple {
+                Some((ref mut index, ref mut best_range)) => {
+                    // We want to find the best fit for this allocation in order to reduce 
+                    // memory fragmentation, so we'll look for it here.
+                    if range.1.end - range.1.start < best_range.end - best_range.start {
+                        *index = range.0;
+                        *best_range = range.1.clone();
+                        // Found a perfect fit, so stop looking.
+                        if range.1.end - range.1.start == length {
```
msiglreith: Should be able to move this out of the match and early out

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    pub fn allocate_range(&mut self, length: T) -> Option<Range<T>> {
+        let mut best_range_tuple: Option<(usize, Range<T>)> = None;
+        for range in self.empty_ranges.iter().cloned().enumerate() {
+            if range.1.end - range.1.start < length {
+                continue;
+            }
+            match best_range_tuple {
+                Some((ref mut index, ref mut best_range)) => {
+                    // We want to find the best fit for this allocation in order to reduce
```
msiglreith: `// Find best fit for this allocation to reduce fragmentation` should be enough

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+                self.empty_ranges.remove(range.0);
+            } else {
+                self.empty_ranges[range.0].start += length;
+            }
+            range.1.start..(range.1.start + length)
+        })
+    }
+
+    pub fn free_range(&mut self, range: Range<T>) -> bool {
+        assert!(range.start >= self.range.start && range.end <= self.range.end);
```
msiglreith: `range.start >= self.range.start` swapping them will give a nice inclusion check

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+        assert!(range.start <= range.end);
```
msiglreith: `<`, empty ranges shouldn't be allowed

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+        if self.empty_ranges.len() == 0 {
```
msiglreith: either a match over len() or if { .. } else { .. }

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+            if range.1.end - range.1.start == length {
+                self.empty_ranges.remove(range.0);
+            } else {
+                self.empty_ranges[range.0].start += length;
+            }
+            range.1.start..(range.1.start + length)
+        })
+    }
+
+    pub fn free_range(&mut self, range: Range<T>) -> bool {
```
msiglreith: Why return a bool here?

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+        assert!(range.start >= self.range.start && range.end <= self.range.end);
+        assert!(range.start <= range.end);
+        if self.empty_ranges.len() == 0 {
+            self.empty_ranges.push(range);
+            return true;
+        }
+        // Input is within range, but before any empty ranges
+        if self.empty_ranges.len() > 0 {
```
msiglreith: Same case is done below in the iteration, not really needed.

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+            if self.empty_ranges[0].start == range.end {
+                self.empty_ranges[0].start = range.start;
+                return true;
+            } else if self.empty_ranges[0].start > range.end {
+                self.empty_ranges.insert(0, range);
+                return true;
+            }
+        }
+        // Input is within range, but after all empty ranges
+        if let Some(last) = self.empty_ranges.last().cloned() {
```
msiglreith: Actually I don't think all these early out special cases are not really needed. Basically hoping the users follow some sort of pattern when freeing but I don't believe that's the case.

src/backend/auxil/range_alloc.rs
msiglreith: note: see below

src/backend/auxil/range_alloc.rs
msiglreith: note: see below

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+                // Merge this into an adjacent range to the right if necessary.
+                if i + 1 != self.empty_ranges.len() && self.empty_ranges[i + 1].start == self.empty_ranges[i].end {
+                    let r = self.empty_ranges.remove(i + 1);
+                    self.empty_ranges[i].end = r.end;
+                }
+                return true;
+            }
+
+            // Input is inbetween two empty ranges.
+            if i + 1 != self.empty_ranges.len() && range.start > self.empty_ranges[i].end && range.end < self.empty_ranges[i + 1].start {
```
msiglreith: just check for `empty[i].start < start` based on the sorting constraint.
Also do this check on the beginning of the loop to just find the position where to insert the range.
Ideally it would look like this:
```rust
idx = self.empty_ranges.iter().positon(|r| r.start < range.start)
// check left merge
// check right merge
// insert

File: src/backend/auxil/range_alloc.rs [hunk 39]
```diff
@@ -0,0 +1,169 @@
...
+            if range.end == self.empty_ranges[i].start {
+                // Extend this range
+                self.empty_ranges[i].start = range.start;
+                // Merge this into an adjacent range to the left if necessary.
+                if i > 0 && self.empty_ranges[i - 1].end == self.empty_ranges[i].start {
+                    let r = self.empty_ranges.remove(i);
+                    self.empty_ranges[i - 1].end = r.end;
+                    
+                }
+                return true;
```
msiglreith: Returning too early, might find inbetween two ranges and we could merge with both neighboring ranges

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,169 @@
...
+        assert!(alloc.free_range(3..6));
+        assert_eq!(alloc.empty_ranges, vec![3..6, 9..10]);
+        // Now request space that the middle block can fill, but the end one can't.
+        assert_eq!(alloc.allocate_range(3), Some(3..6));
+
+        // Clear the allocations and make sure this worked.
+        alloc.reset();
+        assert_eq!(alloc.empty_ranges, vec![0..10]);
+    }
+}
```
msiglreith: nit: line ending

src/backend/auxil/range_alloc.rs
Xaeroxe: This code already does this.  First it extends one range, then if it's touching its neighbor due to extension then it's merged with its neighbor.

src/backend/auxil/range_alloc.rs
Xaeroxe: This comment is confusing because of a few double negatives.  Can you reword it?

src/backend/auxil/range_alloc.rs
msiglreith: why return anything at all?

src/backend/auxil/range_alloc.rs
msiglreith: sry I meant 'early out special cases are not really needed'.
Everything is handled in the iteration case as far as I can see.

src/backend/auxil/range_alloc.rs
msiglreith: Ah right! I got somehow confused by reading 'left' twice

src/backend/auxil/range_alloc.rs
Xaeroxe: I thought so too until I added some unit tests.  I added these because the iteration wasn't accounting for these cases.

src/backend/auxil/range_alloc.rs
Xaeroxe: Because the free could fail and it's important to communicate that.

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    pub fn initial_range(&self) -> Range<T> {
+        self.initial_range.clone()
+    }
+
+    pub fn allocate_range(&mut self, length: T) -> Option<Range<T>> {
+        let mut best_fit: Option<(usize, Range<T>)> = None;
+        for range in self.free_ranges.iter().cloned().enumerate() {
+            let (index, range) = range;
```
msiglreith: `for (index, range) in ... ` should work

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,190 @@
...
+            let range_length = range.end - range.start;
+            if range_length < length {
+                continue;
+            } else if range_length == length {
+                // Found a perfect fit, so stop looking.
+                best_fit = Some((index, range));
+                break;
+            }
+            match best_fit {
```
msiglreith: Can we do something like
```rust
best_fit = Some(match best_fit { .. })
```
here? might be a bit nicer

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,190 @@
...
+                    }
+                }
+                None => {
+                    best_fit = Some((index, range.clone()));
+                }
+            }
+        }
+        best_fit.map(|range| {
+            let (index, range) = range;
+            if range.end - range.start == length {
```
msiglreith: `range.end - range.start` can be replaced by `range_length`

src/backend/auxil/range_alloc.rs
```diff
@@ -0,0 +1,190 @@
...
+                        *best_range = range.clone();
+                    }
+                }
+                None => {
+                    best_fit = Some((index, range.clone()));
+                }
+            }
+        }
+        best_fit.map(|range| {
+            let (index, range) = range;
```
msiglreith: Similar inplace unpacking as above

Xaeroxe:
```
Benching range_alloc...
Time to allocate 500,000,000 items: Duration { secs: 3, nanos: 635872168 }
Time to de-allocate 500,000,000 items: Duration { secs: 2, nanos: 713068272 }
Benching free_list...
Time to allocate 500,000,000 items: Duration { secs: 32, nanos: 371233908 }
Time to de-allocate 500,000,000 items: Duration { secs: 29, nanos: 60538157 }
```

# Meta

 - rustc 1.26.1 (827013a31 2018-05-25)
 - Windows 10
 - Intel(R) Core(TM) i5-6500 CPU @ 3.20 GHz (4 CPUs)
 - Run in release mode

Bench program:

```rust
use std::ops::Range;
use std::time::Instant;

mod free_list;
mod range_alloc;

use free_list::Allocator as FreeList;
use range_alloc::RangeAllocator;

trait Allocator {
    fn alloc(&mut self, size: u64) -> Option<Range<u64>>;

    fn dealloc(&mut self, range: Range<u64>);
}

impl Allocator for FreeList {
    fn alloc(&mut self, size: u64) -> Option<Range<u64>> {
        self.allocate(size)
    }

    fn dealloc(&mut self, range: Range<u64>) {
        self.deallocate(range);
    }
}

impl Allocator for RangeAllocator<u64> {
    fn alloc(&mut self, size: u64) -> Option<Range<u64>> {
        self.allocate_range(size)
    }

    fn dealloc(&mut self, range: Range<u64>) {
        self.free_range(range);
    }
}

fn main() {
    let mut free_list = FreeList::new(1_000_000_000);
    let mut range_alloc = RangeAllocator::new(0..1_000_000_000);
    let allocs: &mut[(&mut Allocator, &str)] = &mut [(&mut range_alloc, "range_alloc"), (&mut free_list, "free_list")];
    for (alloc, name) in allocs {
        println!("Benching {}...", name);
        let start = Instant::now();
        for _ in 0..500_000_000 {
            assert!(alloc.alloc(2).is_some());
        }
        let duration = start.elapsed();
        println!("Time to allocate 500,000,000 items: {:?}", duration);
        let start = Instant::now();
        for i in 0..500_000_000 {
            alloc.dealloc(i*2..(i*2 + 2));
        }
        let duration = start.elapsed();
        println!("Time to de-allocate 500,000,000 items: {:?}", duration);
    }
}

```

@kvark Any criticisms of this benchmark?

src/backend/auxil/range_alloc.rs
msiglreith: But indicates that something is wrong imo.

src/backend/auxil/range_alloc.rs
msiglreith: The current use cases seem to all ignore the return value.

src/backend/auxil/range_alloc.rs
msiglreith: I probably haven't explained my motivation properly behind this. The idea was to shrinking the loop to a minium, only searching for the insertion position. All the merge/insertion/removal operations would then only operate with the returned index.

(edit: my condition inside position(..) might be incorrect..)

src/backend/auxil/range_alloc.rs
Xaeroxe: No, the borrow checker disallows this.  If you try to fill in that match statement it gets convoluted pretty quickly.

src/backend/auxil/range_alloc.rs
Xaeroxe: It can't, because `range_length` is out of scope here.

src/backend/auxil/range_alloc.rs
Xaeroxe: Actually disregard that, I figured it out.

Xaeroxe:
@msiglreith @kvark ready for re-review

src/backend/auxil/range_alloc.rs
Xaeroxe: Uh yeah now that I look at that condition again I'm not sure it's possible to create correct code with the `position` function, so I'm going to undo this change.

kvark:
@Xaeroxe yeah, let's generate a random sequence of allocations and frees, and run both choices through it.

Xaeroxe:
@kvark new bench

```
Benching range_alloc...
Time to allocate 200002938 items: Duration { secs: 1, nanos: 591619302 }
Time to de-allocate 200002938 items: Duration { secs: 1, nanos: 167790382 }
Benching free_list...
Time to allocate 200002938 items: Duration { secs: 13, nanos: 522406580 }
Time to de-allocate 200002938 items: Duration { secs: 12, nanos: 148544505 }
```

program:

```rust
extern crate rand;

use rand::prelude::*;

use std::ops::Range;
use std::time::Instant;

mod free_list;
mod range_alloc;

use free_list::Allocator as FreeList;
use range_alloc::RangeAllocator;

trait Allocator {
    fn alloc(&mut self, size: u64) -> Option<Range<u64>>;

    fn dealloc(&mut self, range: Range<u64>);
}

impl Allocator for FreeList {
    fn alloc(&mut self, size: u64) -> Option<Range<u64>> {
        self.allocate(size)
    }

    fn dealloc(&mut self, range: Range<u64>) {
        self.deallocate(range);
    }
}

impl Allocator for RangeAllocator<u64> {
    fn alloc(&mut self, size: u64) -> Option<Range<u64>> {
        self.allocate_range(size)
    }

    fn dealloc(&mut self, range: Range<u64>) {
        self.free_range(range);
    }
}

fn main() {
    let range_max = 10_000_000_000;
    let mut free_list = FreeList::new(range_max);
    let mut range_alloc = RangeAllocator::new(0..range_max);
    let allocs: &mut[(&mut Allocator, &str)] = &mut [(&mut range_alloc, "range_alloc"), (&mut free_list, "free_list")];
    let mut sizes = Vec::new();
    let mut sum = 0;
    let mut rng = rand::thread_rng();
    while sum < range_max {
        let rand_num = rng.gen_range(1, 100).min(range_max - sum);
        sum += rand_num;
        sizes.push(rand_num);
    }
    for (alloc, name) in allocs {
        println!("Benching {}...", name);
        let start = Instant::now();
        for i in 0..sizes.len() {
            assert!(alloc.alloc(sizes[i]).is_some());
        }
        let duration = start.elapsed();
        println!("Time to allocate {} items: {:?}", sizes.len(), duration);
        let start = Instant::now();
        let mut sum = 0;
        for i in 0..sizes.len() {
            alloc.dealloc(sum..(sum+sizes[i]));
            sum += sizes[i];
        }
        let duration = start.elapsed();
        println!("Time to de-allocate {} items: {:?}", sizes.len(), duration);

    }
}
```

kvark:
Thank you! Looks pretty convincing.
Would you want to squash some commits before proceeding?
Also, let's see if @msiglreith got any concerns left.

Xaeroxe:
Well some testing on Linux and OSX would be nice.  I'll squash it.

PR #2087, opened 2018-05-31
[mtl] Lazier vertex buffer binding

kvark:
This is a follow-up to #2078, which fixes an issue where a vertex buffer wouldn't be bound upon switching the PSO after binding vertices.

PR checklist:


kvark:
Thanks!
Bors r=grovesNL

PR #2088, opened 2018-05-31
[dx11] add graphics pipeline creation & shader module creation

fkaa:
Adds basic pipeline creation and some plumbing for creating/converting from hal to dx11 state. The shader stuff is mostly from the dx12 backend without the root signature stuff, need to investigate how much patching needs to be done on the output (SM 5.0 vs 5.1)

PR checklist:

### DIFF
--- a/src/backend/dx11/src/conv.rs
+++ b/src/backend/dx11/src/conv.rs
@@ -1,10 +1,20 @@
 use hal::format::{Format};
-use hal::pso::{Rect, Viewport};
+use hal::pso::{
+    BlendDesc, BlendOp, BlendState, ColorBlendDesc, Comparison, CullFace, DepthStencilDesc,
+    DepthTest, Factor, PolygonMode, Rasterizer, Rect, StencilFace, StencilOp, StencilTest,
+    Viewport, Stage
+};
+use hal::Primitive;
+
+use spirv_cross::spirv;
+
+use winapi::shared::dxgiformat::*;
+use winapi::shared::minwindef::{FALSE, INT, TRUE};
 
 use winapi::um::d3dcommon::*;
-use winapi::shared::dxgiformat::*;
+use winapi::um::d3d11::*;
 
-use winapi::um::d3d11::{D3D11_RECT, D3D11_VIEWPORT};
+use std::mem;
 
 // TODO: stolen from d3d12 backend, maybe share function somehow?
 pub fn map_format(format: Format) -> Option<DXGI_FORMAT> {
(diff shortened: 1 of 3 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -358,6 +360,178 @@ impl Device {
             context
         }
     }
+
+    fn create_rasterizer_state(&self, rasterizer_desc: &pso::Rasterizer) -> Result<ComPtr<d3d11::ID3D11RasterizerState>, pso::CreationError> {
```
kvark: let's stop extending the device in "lib.rs" and do it in "device.rs" like all other backends do

src/backend/dx11/src/conv.rs
```diff
@@ -102,3 +112,210 @@ pub fn map_rect(rect: &Rect) -> D3D11_RECT {
         bottom: rect.h as _,
     }
 }
+
+pub fn map_topology(primitive: Primitive) -> D3D11_PRIMITIVE_TOPOLOGY {
```
kvark: another thing to share with dx12? assuming `D3D11_PRIMITIVE_TOPOLOGY == D3D_PRIMITIVE_TOPOLOGY`

src/backend/dx11/src/lib.rs
fkaa: I should probably get rid of the files which are leftover from before too.

Do you mind if I do this on the next PR? I have some commits which are based off of this branch already and can see the merge conflicts happening already in my mind..

src/backend/dx11/src/conv.rs
fkaa: Yup

src/backend/dx11/src/lib.rs
kvark: sure

src/backend/dx11/src/lib.rs
```diff
@@ -358,6 +360,178 @@ impl Device {
...
+        let desc = conv::map_rasterizer_desc(rasterizer_desc);
+
+        let hr = unsafe {
+            self.device.CreateRasterizerState(
+                &desc,
+                &mut rasterizer as *mut *mut _ as *mut *mut _
+            )
+        };
+
+        if !winerror::SUCCEEDED(hr) {
```
msiglreith: Swap those
```
if SUCCEED { } else { }
```

src/backend/dx11/src/lib.rs
```diff
@@ -358,6 +360,178 @@ impl Device {
...
+        let desc = conv::map_blend_desc(blend_desc);
+
+        let hr = unsafe {
+            self.device.CreateBlendState(
+                &desc,
+                &mut blend as *mut *mut _ as *mut *mut _
+            )
+        };
+
+        if !winerror::SUCCEEDED(hr) {
```
msiglreith: same here and further down

src/backend/dx11/src/lib.rs
```diff
@@ -415,14 +591,58 @@ impl hal::Device<Backend> for Device {
...
+        let rasterizer_state = self.create_rasterizer_state(&desc.rasterizer)?;
+        let blend_state = self.create_blend_state(&desc.blender)?;
+        let depth_stencil_state = if let Some(desc) = desc.depth_stencil {
+            Some(self.create_depth_stencil_state(&desc)?)
+        } else {
+            None
+        };
+
+        let vs = self.create_vertex_shader(vs)?;
+        let ps = if let Some(blob) = ps {
```
msiglreith: `ps.map(..)`?

fkaa: IIRC compiler complained about `?` inside closure or something similar, i'll give it another try

fkaa:
@msiglreith Updated the error handling, but I couldn't get `map()` to work (`.map()` on `Option<T>` has to return `Option<_>`, but the `create_pixel_shader` returns `Result<T, _>` and so on).

I'll revisit this later but I think at least for now it reads more clear than having a series of transform operations on the return value and nested `?` in closures.

PR #2093, opened 2018-06-01
dx12: remap vertex bindings

msiglreith:
Map input attribute id to input binding to work around issues where attribute offsets are larger than the vertex stride. Current mapping strategy is naive, mapping one attribute to one binding. Better packing would be useful but only provides performance improvements.

Tested against several samples and the CTS texture reftests look reasonable but fail due to lack of freeDescriptorSets implementation.

PR checklist:

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -297,7 +297,7 @@ pub struct CommandBuffer {
     // Cached vertex buffer views to bind.
     // `Stride` values are not known at `bind_vertex_buffers` time because they are only stored
     // inside the pipeline state.
-    vertex_buffer_views_needs_bind: [bool; MAX_VERTEX_BUFFERS],
+    vertex_bindings_remap: [Option<n::VertexBinding>; MAX_VERTEX_BUFFERS],
     vertex_buffer_views: [d3d12::D3D12_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS],
 
     // Re-using allocation for the image-buffer copies.
@@ -346,7 +346,7 @@ impl CommandBuffer {
             active_descriptor_heaps: [ptr::null_mut(); 2],
             occlusion_query: None,
             pipeline_stats_query: None,
-            vertex_buffer_views_needs_bind: [false; MAX_VERTEX_BUFFERS],
+            vertex_bindings_remap: [None; MAX_VERTEX_BUFFERS],
             vertex_buffer_views: [NULL_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS],
             copies: Vec::new(),
             viewport_cache: SmallVec::new(),
@@ -372,7 +372,7 @@ impl CommandBuffer {
         self.active_descriptor_heaps = [ptr::null_mut(); 2];
         self.occlusion_query = None;
         self.pipeline_stats_query = None;
-        self.vertex_buffer_views_needs_bind = [false; MAX_VERTEX_BUFFERS];
+        self.vertex_bindings_remap = [None; MAX_VERTEX_BUFFERS];
         self.vertex_buffer_views = [NULL_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS];
         self.rtv_pools.clear();
         self.temporary_gpu_heaps.clear();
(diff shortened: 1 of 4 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -591,25 +591,36 @@ impl CommandBuffer {
...
                     Some(start_offset) => {
                         let start_slot = last_end_slot + start_offset;
-                        let num_views = vbs_needs_bind[start_slot..]
+                        let num_views = vbs_remap[start_slot..]
                             .iter()
-                            .position(|needs_bind| !*needs_bind)
-                            .unwrap_or(vbs_needs_bind.len());
+                            .position(|remap| remap.is_none())
+                            .unwrap_or(vbs_remap.len());
+                        let buffers = (start_slot..start_slot+num_views).map(|i| {
```
kvark: nit: iterate slice of `vbs_remap[start_slot .. ]` and break on `None` instad?

src/backend/dx12/src/command.rs
```diff
@@ -591,25 +591,36 @@ impl CommandBuffer {
...
+                            let remapping = vbs_remap[i].unwrap();
+                            let view = vbs[remapping.mapped_binding];
+
+                            d3d12::D3D12_VERTEX_BUFFER_VIEW {
+                                BufferLocation: view.BufferLocation + remapping.offset as u64,
+                                SizeInBytes: view.SizeInBytes - remapping.offset,
+                                StrideInBytes: remapping.stride,
+                            }
+                        }).collect::<Vec<_>>();
```
kvark: could we re-use the vector allocation at least within this function?

src/backend/dx12/src/command.rs
```diff
@@ -591,25 +591,36 @@ impl CommandBuffer {
...
+                            .position(|remap| remap.is_none())
+                            .unwrap_or(vbs_remap.len());
+                        let buffers = (start_slot..start_slot+num_views).map(|i| {
+                            let remapping = vbs_remap[i].unwrap();
+                            let view = vbs[remapping.mapped_binding];
+
+                            d3d12::D3D12_VERTEX_BUFFER_VIEW {
+                                BufferLocation: view.BufferLocation + remapping.offset as u64,
+                                SizeInBytes: view.SizeInBytes - remapping.offset,
+                                StrideInBytes: remapping.stride,
```
kvark: doesn't it cause d3d12 validation warnings, given that offset + N * stride is outside the view range?

src/backend/dx12/src/device.rs
```diff
@@ -1505,8 +1507,8 @@ impl d::Device<B> for Device {
                             return Some(Err(pso::CreationError::Other));
                         }
                     },
-                    InputSlot: attrib.binding as _,
-                    AlignedByteOffset: attrib.element.offset,
+                    InputSlot: i as _,
+                    AlignedByteOffset: 0, // Rebased during the vertex binding remapping
```
kvark: Wait, so you are force remapping everything to have zero attribute offset? That is quite rough and may cause us to run out of the vertex buffer could.

msiglreith: Never happened so far

src/backend/dx12/src/command.rs
msiglreith: nop

src/backend/dx12/src/command.rs
msiglreith: done but involved quite a bit of iterator origami

src/backend/dx12/src/command.rs
msiglreith: Done via SmallVec avoiding allocations (I hope)

src/backend/dx12/src/command.rs
```diff
@@ -601,25 +601,26 @@ impl CommandBuffer {
...
+                            .map(|mapping| {
+                                let view = vbs[mapping.mapped_binding];
+
+                                d3d12::D3D12_VERTEX_BUFFER_VIEW {
+                                    BufferLocation: view.BufferLocation + mapping.offset as u64,
+                                    SizeInBytes: view.SizeInBytes - mapping.offset,
+                                    StrideInBytes: mapping.stride,
+                                }
+                            })
+                            .collect::<SmallVec<[d3d12::D3D12_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS]>>();
```
kvark: I don't think you need that actual type in there, could just be `SmallVec<_>`

src/backend/dx12/src/command.rs
```diff
@@ -601,25 +601,26 @@ impl CommandBuffer {
...
-                            let remapping = vbs_remap[i].unwrap();
-                            let view = vbs[remapping.mapped_binding];
-
-                            d3d12::D3D12_VERTEX_BUFFER_VIEW {
-                                BufferLocation: view.BufferLocation + remapping.offset as u64,
-                                SizeInBytes: view.SizeInBytes - remapping.offset,
-                                StrideInBytes: remapping.stride,
-                            }
-                        }).collect::<Vec<_>>();
+                            .take_while(|x| x.is_some())
```
kvark: nice iterator usage :+1:

src/backend/dx12/src/device.rs
```diff
@@ -1497,6 +1515,30 @@ impl d::Device<B> for Device {
                 };
                 let format = attrib.element.format;
 
+                // Check if we need to add a new remapping in-case the offset is
+                // higher than the vertex stride.
+                // In this case we rebase the attribute to zero offset.
+                let binding = attrib.binding as usize;
+                let stride = vertex_strides[binding];
+                let offset = attrib.element.offset;
+                let (input_slot, offset) = if stride <= offset {
```
kvark: how does it handle a case when `offset < stride` but `offset + element_size > stride`?

msiglreith: Works fine so far, added warden test for this.

src/backend/dx12/src/command.rs
msiglreith: I could remove the array type at least but not the size

PR #2096, opened 2018-06-01
Metal mega-queue, coherent memory, and other goodies

### LINKED ISSUE #2069: Coherent memory on Metal

### LINKED ISSUE #2092: Broken depth ranges on Metal
All tests here are failing: "dEQP-VK.draw.inverted_depth_ranges". It doesn't look like Metal respects the depth viewport range as it should. I submitted a bug report, will keep this open for tracking.

### LINKED ISSUE #2094: Queue synchronization in Metal
Metal has a limit on the number of command buffers a queue can maintain in flight (default is 64). In order to support unlimited number of command buffers, we maintain a pool of queues, from which the command buffers are allocated. The question is now - do we need to have any extra synchronization between command buffers spawned by different queues?..

Example:
  - cmdbuf A is allocated from Queue 1
  - cmdbuf B is allocated from Queue 2
  - users commits A followed by B
  - user has a semaphore/fence associated with B, and upon reaching that they assume that A is done as well. Will this guaranteed to be the case on Metal?

kvark:
Fixes #2069
Fixes #2094
Fixes #2092 

This PR accumulated quite a few seemingly unrelated changes, but all rather tasty ones:
  - one mega-queue instead of a queue pool
  - condvars for fences
  - enabled coherent memory
  - inverted depth range workaround
  - depth blits

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -922,9 +922,10 @@ impl RawCommandQueue<Backend> for CommandQueue {
 
         if let Some(ref fence) = fence {
             let command_buffer = queue.new_command_buffer_with_unretained_references();
-            let value_ptr = fence.0.clone();
+            let fence = Arc::clone(fence);
             let fence_block = ConcreteBlock::new(move |_cb: *mut ()| -> () {
-                *value_ptr.lock().unwrap() = true;
+                *fence.mutex.lock().unwrap() = true;
+                fence.condvar.notify_all();
             }).copy();
             msg_send![command_buffer, addCompletedHandler: fence_block.deref() as *const _];
             command_buffer.commit();
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -27,50 +27,56 @@ use {conversions as conv, soft};
 
 const WORD_ALIGNMENT: u64 = 4;
 
-pub(crate) struct QueueInner {
-    queue: metal::CommandQueue,
+pub struct QueueInner {
+    raw: metal::CommandQueue,
     reserve: Range<usize>,
 }
 
-#[derive(Default)]
-pub struct QueuePool {
-    queues: Vec<QueueInner>,
+pub struct Token {
+    active: bool,
 }
 
-impl QueuePool {
-    fn find_queue(&mut self, device: &Mutex<metal::Device>) -> usize {
-        const POOL_SIZE: usize = 64;
-        self.queues
-            .iter()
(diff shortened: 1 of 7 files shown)


src/backend/metal/src/command.rs
```diff
@@ -27,50 +27,58 @@ use {conversions as conv, soft};
...
+        token.active = false;
+        self.reserve.start -= 1;
+    }
+
+    /// Block until GPU is idle.
+    pub(crate) fn wait_idle(queue: &Mutex<Self>) {
+        debug!("waiting for idle");
+        let _pool = AutoreleasePool::new();
+        // note: we deliberately don't hold the Mutex lock while waiting,
+        // since the completition handlers need to access it.
```
msiglreith: typo: completion

src/backend/metal/src/command.rs
```diff
@@ -27,50 +27,58 @@ use {conversions as conv, soft};
...
-        self.queues[index].reserve.start -= 1;
+    /// Returns a command buffer to a virtual pool.
+    pub(crate) fn release(&mut self, mut token: Token) {
+        token.active = false;
+        self.reserve.start -= 1;
+    }
+
+    /// Block until GPU is idle.
+    pub(crate) fn wait_idle(queue: &Mutex<Self>) {
+        debug!("waiting for idle");
```
msiglreith: intended debug message?

src/backend/metal/src/command.rs
```diff
@@ -1166,15 +1168,13 @@ impl CommandBuffer {
 
 impl com::RawCommandBuffer<Backend> for CommandBuffer {
     fn begin(&mut self, flags: com::CommandBufferFlags, _info: com::CommandBufferInheritanceInfo<Backend>) {
+        self.reset(false);
```
grovesNL: Why didn't this fail before?

kvark:
Thanks for the reviews!
It didn't fail because we were leaking command buffers...

Bors r=msiglreith,grovesNL

src/backend/metal/src/command.rs
kvark: yeah, it's been there before as well

kvark:
Thanks!
Bors r=grovesNL

Issue #2097, opened 2018-06-02
dx12: quad fails to run


grovesNL:
Quad fails to run on DX12. Maybe related to #1978?

- GFX version: master/ a0fc1b7
- OS: Windows 10
- GPU: GTX 970

```
$ RUST_BACKTRACE=1 cargo run --bin quad --features dx12
    Finished dev [unoptimized + debuginfo] target(s) in 0.18 secs
     Running `C:\Projects\rust\gfx\gfx\target\debug\quad.exe`
AdapterInfo { name: "NVIDIA GeForce GTX 970", vendor: 4318, device: 5058, software_rendering: false }
AdapterInfo { name: "Microsoft Basic Render Driver", vendor: 5140, device: 140, software_rendering: true }
Surface format: Bgra8Srgb
Memory types: [MemoryType { properties: DEVICE_LOCAL, heap_index: 0 }, MemoryType { properties: COHERENT | CPU_VISIBLE, heap_index: 1 }, MemoryType { properties: COHERENT | CPU_VISIBLE | CPU_CACHED, heap_index: 1 }, MemoryType { properties: DEVICE_LOCAL, heap_index: 0 }, MemoryType { properties: COHERENT | CPU_VISIBLE, heap_index: 1 }, MemoryType { properties: COHERENT | CPU_VISIBLE | CPU_CACHED, heap_index: 1 }, MemoryType { properties: DEVICE_LOCAL, heap_index: 0 }, MemoryType { properties: COHERENT, heap_index: 1 }, MemoryType { properties: COHERENT | CPU_CACHED, heap_index: 1 }, MemoryType { properties: DEVICE_LOCAL, heap_index: 0 }, MemoryType { properties: COHERENT, heap_index: 1 }, MemoryType { properties: COHERENT | CPU_CACHED, heap_index: 1 }]
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `-2147024809`', src\backend\dx12\src\device.rs:2070:9
```

fkaa:
Works for me, also on master. What does the debug layer say? (launch through visual studio or open Dbgview.exe)

```rust
C:\Dev\gfx\examples\hal>cargo run --bin quad --features=dx12
    Finished dev [unoptimized + debuginfo] target(s) in 0.36 secs
     Running `C:\Dev\gfx\target\debug\quad.exe`
[...]

PR #2098, opened 2018-06-02
Improve static/dynamic stencil support

### LINKED ISSUE #2067: Switch cull mode into bitflags
We need to support culling both front and back on the API side.
Test: "dEQP-VK.draw.negative_viewport_height.front_ccw_cull_both"

grovesNL:
Fixes #2067.

- Map static stencil values for Metal
- Implement dynamic stencil reference values for Metal
- Modify signature of `set_stencil_reference` to match Vulkan (debatable, although otherwise we don't have a way to set single faces..)
- Add a new `State` type (open to better name ideas) to try out the idea of static/dynamic fields instead of moving them all into `baked_states`

I haven't had a chance to run any test cases for this yet, so I'm hoping the state setup/binds at least make sense logically.

I also haven't added dynamic read/write masks yet, though maybe I can do that in a followup PR instead.

PR checklist:
- [ ] `make reftests` succeeds (having an issue with it after rebasing, not sure if it's related to my change?.. Will look into it more)
- [X] tested examples with the following backends: Metal, Vulkan, DX12 (crash at #2097), GL

### DIFF
--- a/src/backend/dx12/src/conv.rs
+++ b/src/backend/dx12/src/conv.rs
@@ -148,7 +148,6 @@ pub fn map_topology(primitive: Primitive) -> D3D12_PRIMITIVE_TOPOLOGY {
 
 pub fn map_rasterizer(rasterizer: &pso::Rasterizer) -> D3D12_RASTERIZER_DESC {
     use hal::pso::PolygonMode::*;
-    use hal::pso::CullFace::*;
     use hal::pso::FrontFace::*;
 
     D3D12_RASTERIZER_DESC {
@@ -164,9 +163,10 @@ pub fn map_rasterizer(rasterizer: &pso::Rasterizer) -> D3D12_RASTERIZER_DESC {
             Fill => D3D12_FILL_MODE_SOLID,
         },
         CullMode: match rasterizer.cull_face {
-            None => D3D12_CULL_MODE_NONE,
-            Some(Front) => D3D12_CULL_MODE_FRONT,
-            Some(Back) => D3D12_CULL_MODE_BACK,
+            Some(cf) if cf == pso::Face::FRONT => D3D12_CULL_MODE_FRONT,
+            Some(cf) if cf == pso::Face::BACK => D3D12_CULL_MODE_BACK,
+            Some(_) => panic!("Culling both front and back faces is not supported"),
+            _ => D3D12_CULL_MODE_NONE,
         },
         FrontCounterClockwise: match rasterizer.front_face {
             Clockwise => FALSE,
@@ -278,8 +278,14 @@ pub fn map_depth_stencil(dsi: &pso::DepthStencilDesc) -> D3D12_DEPTH_STENCIL_DES
         DepthWriteMask: if depth_write {D3D12_DEPTH_WRITE_MASK_ALL} else {D3D12_DEPTH_WRITE_MASK_ZERO},
         DepthFunc: depth_func,
         StencilEnable: stencil_on,
-        StencilReadMask: read_mask as _,
-        StencilWriteMask: write_mask as _,
+        StencilReadMask: match read_mask {
+            pso::State::Static(rm) => rm as _,
+            pso::State::Dynamic => !0,
+        },
+        StencilWriteMask: match write_mask {
+            pso::State::Static(wm) => wm as _,
+            pso::State::Dynamic => !0,
+        },
         FrontFace: front,
         BackFace: back,
     }
--- a/src/backend/gl/src/state.rs
+++ b/src/backend/gl/src/state.rs
@@ -31,7 +31,6 @@ pub fn bind_polygon_mode(gl: &gl::Gl, mode: pso::PolygonMode, bias: Option<pso::
 }
 
 pub fn bind_rasterizer(gl: &gl::Gl, r: &pso::Rasterizer, is_embedded: bool) {
-    use hal::pso::CullFace::*;
     use hal::pso::FrontFace::*;
 
     unsafe {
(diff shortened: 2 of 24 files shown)


src/backend/dx12/src/conv.rs
```diff
@@ -278,8 +278,14 @@ pub fn map_depth_stencil(dsi: &pso::DepthStencilDesc) -> D3D12_DEPTH_STENCIL_DES
         DepthWriteMask: if depth_write {D3D12_DEPTH_WRITE_MASK_ALL} else {D3D12_DEPTH_WRITE_MASK_ZERO},
         DepthFunc: depth_func,
         StencilEnable: stencil_on,
-        StencilReadMask: read_mask as _,
-        StencilWriteMask: write_mask as _,
+        StencilReadMask: match read_mask {
+            pso::State::Static(rm) => rm as _,
+            pso::State::Dynamic => 1,
```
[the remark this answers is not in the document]
grovesNL: Yeah `1` here is incorrect – it was just a placeholder. Should we prefer `0` or `!0` for the read/write mask and reference value as defaults?

src/hal/src/pso/mod.rs
```diff
@@ -229,3 +229,13 @@ pub enum Constant {
     F32(f32),
     F64(f64),
 }
+
+/// Pipeline state which may be static or dynamic.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum State<T> {
```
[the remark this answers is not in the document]
grovesNL: Yep, if this `State` approach looks good, I'd prefer to transition all of the baked states over too, otherwise I'll update this PR to move these values into `BakedStates`.

src/backend/dx11/src/state.rs
```diff
@@ -19,10 +19,12 @@ pub fn make_rasterizer(device: &mut ComPtr<ID3D11Device>, rast: &state::Rasteriz
             },
             state::RasterMethod::Fill => D3D11_FILL_SOLID,
         },
-        CullMode: match rast.cull_face {
-            state::CullFace::Nothing => D3D11_CULL_NONE,
-            state::CullFace::Front => D3D11_CULL_FRONT,
-            state::CullFace::Back => D3D11_CULL_BACK,
+        CullMode: if cf.contains(pso::Face::FRONT) {
```
kvark: nit: needs to check if both are not set

src/backend/dx12/src/conv.rs
```diff
@@ -164,9 +163,9 @@ pub fn map_rasterizer(rasterizer: &pso::Rasterizer) -> D3D12_RASTERIZER_DESC {
             Fill => D3D12_FILL_MODE_SOLID,
         },
         CullMode: match rasterizer.cull_face {
-            None => D3D12_CULL_MODE_NONE,
-            Some(Front) => D3D12_CULL_MODE_FRONT,
-            Some(Back) => D3D12_CULL_MODE_BACK,
+            Some(cf) if cf.contains(pso::Face::FRONT) => D3D12_CULL_MODE_FRONT,
```
kvark: same here - needs a check for culling everything

src/backend/gl/src/state.rs
```diff
@@ -42,16 +41,17 @@ pub fn bind_rasterizer(gl: &gl::Gl, r: &pso::Rasterizer, is_embedded: bool) {
     };
 
     match r.cull_face {
-        Some(face) => unsafe {
+        Some(cf) if !cf.is_empty() => unsafe {
             gl.Enable(gl::CULL_FACE);
-            gl.CullFace(match face {
-                Front => gl::FRONT,
-                Back => gl::BACK,
+            gl.CullFace(if cf.contains(pso::Face::FRONT) {
```
kvark: same here - needs to use `gl::FRONT_ANT_BACK` for all faces

src/backend/gl/src/state.rs
```diff
@@ -114,21 +114,33 @@ pub fn bind_stencil(
     gl: &gl::Gl,
     stencil: &pso::StencilTest,
     refs: (pso::StencilValue, pso::StencilValue),
```
kvark: nit: we could do `(ref_front, ref_back): (pso::StencilValue, pso::StencilValue),`

src/backend/vulkan/src/conv.rs
```diff
@@ -265,10 +265,13 @@ pub fn map_polygon_mode(rm: pso::PolygonMode) -> (vk::PolygonMode, f32) {
     }
 }
 
-pub fn map_cull_face(cf: pso::CullFace) -> vk::CullModeFlags {
-    match cf {
-        pso::CullFace::Front   => vk::CULL_MODE_FRONT_BIT,
-        pso::CullFace::Back    => vk::CULL_MODE_BACK_BIT,
+pub fn map_cull_face(cf: pso::Face) -> vk::CullModeFlags {
+    if cf.contains(pso::Face::FRONT) {
```
kvark: same here - needs to handle `FRONT | BACK`

src/hal/src/pso/mod.rs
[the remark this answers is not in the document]
kvark: Hmm, so this is the deciding point... I'll think about it till the end of the day.
There is just a few things to address here before we can merge, it seems.

src/backend/vulkan/src/conv.rs
grovesNL: Sure 👍 I was trying to avoid changing the behavior too much, but I think we should fix this for #2067

src/backend/gl/src/state.rs
grovesNL: 🐜 A new kind of GL face? 😆

src/hal/src/pso/mod.rs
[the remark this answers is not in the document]
grovesNL: We could change them over in later PRs too.

IMO it does seem more elegant to group related states together (i.e. instead of spreading stencil flags across two separate fields), and more consistent with Vulkan I guess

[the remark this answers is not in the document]
kvark: There are following choices:
  A. let all static states (that could be dynamic) to go into `BakedStates`
  B. have them as `Option` within bigger state blocks (e.g. stencil ref inside stencil face)
  C. have them as `State` within bigger state blocks

I think that 1) having states be more local helps (so, B and C are slightly better), and 2) this is a somewhat special mechanics of baking states in the pipeline, so it makes sense to have a separate enum type (`State`). Thus, I tend to agree with the direction this PR is taking.

However, @msiglreith please check this out and let us know in general if you have concerns.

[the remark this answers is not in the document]
msiglreith: No concerns from my side. I personally prefer option C.

kvark:
Cool. Please proceed when ready.
Bors delegate+

grovesNL:
bors r=kvark

PR #2103, opened 2018-06-03
Metal clear image improvements

### LINKED ISSUE #2050: Support mipmapped 1D images on Metal
Metal doesn't support mipmaps on 1D textures, and Vulkan CTS appears to use that quite heavily.
We currently have a quick hack in https://github.com/gfx-rs/gfx/blob/8d2b93a8dfd9c99b7cbec703924a628d56cc9c73/src/backend/metal/src/device.rs#L1487 for treating those as 2D, but it is not supported consistently, and we end up getting tons and tons of Metal validation errors/panics on any tests for 1D textures.

We need to come up with a good plan for handling those. Perhaps, just go through the code and make sure that there is no assumption about a 1D HAL image being backend by 1D Metal texture.

kvark:
Fixes #2050 (yay! `image_format_properties()` saves the day)
Fixes 3D and 2D multi-layer clears in CTS.

Known issues:
  - for some reason, RGBA32* formats fail for multi-layered clears. Need to investigate further and/or make a repro case for Apple
  - depth multi-layered clears work for the most part, but the initial clear that touches all the layers is (for some reason) ignoring layer=0... Similarly, this needs to be investigated, reproduced, and tracked separately.
  - I implemented a fast path that clears all the layers at once. We clear by making a dummy render pass, and I'm creating an extra texture view (to have the layers start at 0) and make it a layered attachment. This doesn't appear to work, even though layered attachments work fine in general (when CTS explicitly wants them).

PR checklist:
- [ ] `make reftests` succeeds (apparently, this got broken recently - "entity not found", cc @msiglreith )

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -20,12 +20,16 @@ use hal::range::RangeArg;
 
 use foreign_types::ForeignType;
 use metal::{self, MTLViewport, MTLScissorRect, MTLPrimitiveType, MTLIndexType, MTLSize, CaptureManager};
-use cocoa::foundation::{NSUInteger, NSInteger};
+use cocoa::foundation::{NSUInteger, NSInteger, NSRange};
 use block::{ConcreteBlock};
 use {conversions as conv, soft};
 
 
 const WORD_ALIGNMENT: u64 = 4;
+/// Enable an optimization to have multi-layered render passed
+/// with clear operations set up to implement our `clear_image`
+/// Note: currently doesn't work, needs a repro case for Apple
+const CLEAR_IMAGE_ARRAY: bool = false;
 
 pub struct QueueInner {
     raw: metal::CommandQueue,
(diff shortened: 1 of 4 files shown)


msiglreith:
See https://github.com/gfx-rs/gfx/pull/2104 regarding reftests

src/backend/metal/src/device.rs
```diff
@@ -290,13 +287,21 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+            width: max_dimension,
+            height: if dimensions >= 2 { max_dimension } else { 1 },
+            depth: if dimensions >= 3 { max_dimension } else { 1 },
+        };
 
         self.private_caps.map_format(format).map(|_| image::FormatProperties {
-            max_extent: image::Extent { width, height, depth },
-            max_levels: max_dimension.log2().ceil() as u8 + 1,
+            max_extent,
+            max_levels: if dimensions == 1 { 1 } else { 12 },
```
grovesNL: Why `12`?

src/backend/metal/src/command.rs
```diff
@@ -1335,47 +1335,18 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
 
         for subresource_range in subresource_ranges {
             let sub = subresource_range.borrow();
-            let end_level = if sub.levels.end == !0 {
-                image.raw.mipmap_level_count() as _
-            } else {
-                sub.levels.end
-            };
-            let end_layer = if sub.layers.end == !0 {
-                image.raw.array_length() as _
```
grovesNL: Why don't we have this case anymore?

src/backend/metal/src/command.rs
```diff
@@ -2242,14 +2282,19 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
         let new_src = if src.mtl_format == dst.mtl_format {
-            src.raw.clone()
+            &src.raw
         } else {
             assert_eq!(src.format_desc.bits, dst.format_desc.bits);
             let tex = src.raw.new_texture_view(dst.mtl_format);
-            inner.retained_textures.push(tex.clone());
-            tex
+            retained_textures.push(tex);
+            retained_textures.last().unwrap()
```
grovesNL: Seems a bit indirect to pull `tex` back out like this, is there another way?

src/backend/metal/src/command.rs
kvark: it's handled in `portability` now, no special values here any more :tada:

src/backend/metal/src/command.rs
kvark: I don't think so - we happen to use a similar patter a bit in WebRender. I think it's just a lack of Entry-like API for vectors.

src/backend/metal/src/device.rs
kvark: based on `max_dimension`. I should make it cleaner...

grovesNL: Oh ok, fair enough 👍 If we query this somehow we'll need to move back to the calculation though, or precompute all of these.

PR #2104, opened 2018-06-03
reftests: Add missing vertex-offset files

msiglreith:
PR checklist:


grovesNL:
I think you overwrote the remote branch with the changes for #2105

msiglreith:
ugh that was close..
thanks for mentioning

PR #2105, opened 2018-06-03
auxil: Simplify allocator range freeing

msiglreith:
dx12: Allow destroying descriptor pools before freeing all sets

Removing all the special case handling and also returning an error. In case we hit an actual error case in the former implementation it would indicate that the user tried to free the same descriptor set multiple times or we did something wrong internally. The first case shouldn't be possible due to move semantics and in the second case an assert should be fine enough also. This version catches the same issues as the previous one.

PR checklist:

cc @Xaeroxe

### DIFF
--- a/src/backend/auxil/range_alloc.rs
+++ b/src/backend/auxil/range_alloc.rs
@@ -71,33 +71,35 @@ where
         // Before: |left|-(range)-|right|
         if i > 0 && range.start == self.free_ranges[i - 1].end {
             // Merge with |left|.
-            self.free_ranges[i - 1].end = range.end;
-            // Check for possibly merge with |right| also.
-            if i < self.free_ranges.len()
-                && self.free_ranges[i - 1].end == self.free_ranges[i].start
-            {
-                let right = self.free_ranges.remove(i);
-                self.free_ranges[i - 1].end = right.end;
-            }
+            self.free_ranges[i - 1].end =
+                if i < self.free_ranges.len() && range.end == self.free_ranges[i].start {
+                    // Check for possible merge with |left| and |right|.
+                    let right = self.free_ranges.remove(i);
+                    right.end
+                } else {
+                    range.end
+                };
+
             return;
         } else if i < self.free_ranges.len() && range.end == self.free_ranges[i].start {
             // Merge with |right|.
-            self.free_ranges[i].start = range.start;
-            // Check for possibly merge with |left| also.
-            if i > 0 && self.free_ranges[i].start == self.free_ranges[i - 1].end {
-                let left = self.free_ranges.remove(i - 1);
-                self.free_ranges[i].start = left.start;
-            }
+            self.free_ranges[i].start =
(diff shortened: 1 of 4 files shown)


src/backend/auxil/range_alloc.rs
```diff
@@ -62,68 +58,48 @@ where
...
+            .position(|r| r.start > range.start)
+            .unwrap_or(self.free_ranges.len());
+
+        // Try merging with neighboring ranges in the free list.
+        // Before: |left|-(range)-|right|
+        if i > 0 && range.start == self.free_ranges[i - 1].end {
+            // Merge with |left|.
+            self.free_ranges[i - 1].end = range.end;
+            // Check for possibly merge with |right| also.
+            if i < self.free_ranges.len()
```
kvark: nit: could do
```rust
self.free_ranges[i-1].end = if i < self.free_ranges.len() && ranges.end == self.free_ranges[i].start {
  self.free_ranges.remove(i).end
} else {
  range.end
};
```

src/backend/auxil/range_alloc.rs
```diff
@@ -62,68 +58,48 @@ where
...
-                // Merge this into an adjacent range to the right if necessary.
-                if i + 1 != self.free_ranges.len()
-                    && self.free_ranges[i + 1].start == self.free_ranges[i].end
-                {
-                    let r = self.free_ranges.remove(i + 1);
-                    self.free_ranges[i].end = r.end;
-                }
-                return Ok(());
+            return;
+        } else if i < self.free_ranges.len() && range.end == self.free_ranges[i].start {
```
kvark: if we are always to check both sides in both branches, couldn't we do this separately without repeating ourselves?

src/backend/auxil/range_alloc.rs
```diff
@@ -62,68 +58,48 @@ where
...
-            if i + 1 != self.free_ranges.len()
-                && range.start > self.free_ranges[i].end
-                && range.end < self.free_ranges[i + 1].start
-            {
-                self.free_ranges.insert(i + 1, range);
-                return Ok(());
-            }
+        // Debug checks
+        if i > 0 {
+            assert!(self.free_ranges[i - 1].end < range.start);
```
kvark: any reason not to put `i>0` into the assert expression?

src/backend/auxil/range_alloc.rs
msiglreith: As far as I understand it's an optimization to avoid two removes. I dont see an easy way atm to simplify it more.

src/backend/auxil/range_alloc.rs
msiglreith: So `i==0 || ..`? Yeah that could work.

kvark:
Ok, please "r=me,Xaeroexe" when ready

msiglreith:
bors r=kvark,Xaeroexe

PR #2107, opened 2018-06-04
Metal attachment clears

kvark:
It's great to have that shader-based clear code finally coming handy :)
Known issues: all multi-layered clears fail. Seems to be very similar to #2103. Supposedly, layered rendering in Metal gets totally off the rails when used on an image view, e.g. a texture that has a parent texture with non-zero layer offset. I'll start making a repro case for Apple to address, which will hopefully let us pass all the tests and do it efficiently.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1,6 +1,6 @@
 use {AutoreleasePool, Backend, Shared, validate_line_width};
 use {native, window};
-use internal::BlitVertex;
+use internal::{BlitVertex, ClearKey, ClearVertex};
 
 use std::borrow::{self, Borrow};
 use std::cell::RefCell;
@@ -120,6 +120,7 @@ struct State {
     depth_stencil_state: Option<metal::DepthStencilState>,
     push_constants: Vec<u32>,
     vertex_buffers: Vec<Option<(metal::Buffer, u64)>>,
+    framebuffer_inner: native::FramebufferInner,
 }
 
 impl State {
@@ -1017,6 +1018,7 @@ impl pool::RawCommandPool<Backend> for CommandPool {
                 depth_stencil_state: None,
                 push_constants: Vec::new(),
                 vertex_buffers: Vec::new(),
+                framebuffer_inner: native::FramebufferInner::default(),
             },
         }).collect();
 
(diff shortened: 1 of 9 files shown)


src/backend/metal/src/command.rs
```diff
@@ -1435,15 +1437,179 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
         T: IntoIterator,
         T::Item: Borrow<com::AttachmentClear>,
         U: IntoIterator,
         U::Item: Borrow<pso::ClearRect>,
     {
-        unimplemented!()
+        // gather vertices/polygons
+        let de = self.state.framebuffer_inner.extent;
+        let mut vertices = Vec::new();
+        println!("xx");
```
grovesNL: Should remove this

src/backend/metal/src/internal.rs
```diff
@@ -116,43 +111,63 @@ impl ServicePipes {
...
+    //TODO: return `Option<metal::DepthStencilState>` instead?
+    pub fn get_depth_stencil(&self, depth: bool, stencil: bool) -> &metal::DepthStencilStateRef {
+        if depth && stencil {
+            &self.ds_write_all_state
+        } else if depth {
+            &self.ds_write_depth_state
+        } else if stencil {
+            &self.ds_write_stencil_state
+        } else {
+            panic!("Can't write no thing!")
```
grovesNL: nit: `nothing`

src/backend/metal/src/command.rs
```diff
@@ -1435,15 +1437,179 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    let raw_value = com::ClearColorRaw::from(value);
+                    commands.push(soft::RenderCommand::BindBufferData {
+                        stage: pso::Stage::Fragment,
+                        index: 0,
+                        bytes: unsafe {
+                            slice::from_raw_parts(raw_value.float32.as_ptr() as *const u8, 16)
+                        }.to_owned(),
+                    });
+                    ClearKey {
+                        color: Some((format, index as u8, channel.unwrap())),
```
grovesNL: What happens when `index > 1` here? Won't `let ps_name = format!("ps_clear{}_{}", index, s_channel);` fail to find the appropriate shader?

kvark: Yep, it feels like this either needs a constant-based RT index selection that comes in Metal 2.0, or mamy many repeated shaders, or some sort of a generator for those (TODO).

kvark:
Thanks for the review! Comments are addressed now, merging.
bors r=grovesNL

src/backend/metal/src/command.rs
grovesNL: Ok, yeah, maybe a simple generator wouldn't be too hard

PR #2108, opened 2018-06-04
[dx11] buffer image copies and more pipeline things

fkaa:
Adds buffer to (2d) image copies and (very) basic version descriptor sets. Also adds more pipeline things like index buffer binding, depth stencil creation, binding & clearing. Also removes old files in the backend directory.

Made a small example on the side to test everything so far with the backend (vertex/index buffer uploads, image copies, pipeline binding, buffer updates etc.):

Fixes #issue
PR checklist:


msiglreith:
`+1,625 −3,186`
what happened with the 1.5k LOC ?

fkaa:
Moved `Device` struct and impl into device.rs

fkaa:
Just remembered, this PR is blocked by https://github.com/retep998/winapi-rs/pull/618 (next winapi release which is soon). I'll update the PR when it's out.

src/backend/dx11/src/conv.rs
```diff
@@ -16,6 +17,100 @@ use winapi::um::d3d11::*;
...
 
+pub fn map_index_type(ty: IndexType) -> DXGI_FORMAT {
+    match ty {
+        IndexType::U16 => DXGI_FORMAT_R16_UINT,
+        IndexType::U32 => DXGI_FORMAT_R32_UINT,
+    }
+}
+
+pub fn typeless_format(format: DXGI_FORMAT) -> Option<DXGI_FORMAT> {
+    let format = match format {
```
kvark: nit: can do just `Some(match format {..})`

src/backend/dx11/src/conv.rs
```diff
@@ -319,3 +414,46 @@ pub fn map_stage(stage: Stage) -> spirv::ExecutionModel {
...
+        WrapMode::Mirror => D3D11_TEXTURE_ADDRESS_MIRROR,
+        WrapMode::Clamp  => D3D11_TEXTURE_ADDRESS_CLAMP,
+        WrapMode::Border => D3D11_TEXTURE_ADDRESS_BORDER,
+    }
+}
+
+pub fn map_anisotropic(anisotropic: Anisotropic) -> D3D11_FILTER {
+    match anisotropic {
+        Anisotropic::On(_) => D3D11_FILTER_ANISOTROPIC,
+        Anisotropic::Off => 0,
```
kvark: 0 feels awkward here. Perhaps we could make this cleaner?

src/backend/dx11/src/device.rs
```diff
@@ -1,983 +1,1322 @@
...
-    share: Arc<Share>,
-    frame_handles: h::Manager<R>,
-    vs_cache: Map<u64, Vec<u8>>,
-    /// Create typed surface formats for the textures. This is useful for debugging
-    /// with PIX, since it doesn't understand typeless formats. This may also prevent
-    /// some valid views to be created because the typed formats can't be reinterpret.
-    use_texture_format_hint: bool,
-    sub_data_array: Vec<winapi::D3D11_SUBRESOURCE_DATA>,
-    feature_level: winapi::D3D_FEATURE_LEVEL,
+    pub(crate) device: ComPtr<d3d11::ID3D11Device>,
```
kvark: why does this all need to be `pub(crate)`?

kvark: nit: I typically name the internal things as `raw` in order to avoid usage like `device.device.xxx`

src/backend/dx11/src/device.rs
```diff
@@ -1,983 +1,1322 @@
...
+    pub(crate) context: ComPtr<d3d11::ID3D11DeviceContext>,
+    pub(crate) memory_properties: hal::MemoryProperties,
+    pub(crate) internal: Arc<RefCell<internal::BufferImageCopy>>
```
kvark: I'm surprised to see `Arc<RefCell<..>>`. Here, `RefCell` is `!Sync`, which makes `Arc<>` to be `!Sync`, so what do you get out of `Arc` here?

src/backend/dx11/src/device.rs
```diff
@@ -1,983 +1,1322 @@
...
+    pub fn new(device: ComPtr<d3d11::ID3D11Device>, context: ComPtr<d3d11::ID3D11DeviceContext>, memory_properties: hal::MemoryProperties) -> Self {
         Device {
-            device: device,
-            share: share,
-            frame_handles: h::Manager::new(),
-            vs_cache: Map::new(),
-            use_texture_format_hint: false,
-            sub_data_array: Vec::new(),
-            feature_level: feature_level,
+            device: device.clone(),
```
kvark: nit: could get away without a second `clone()`, or just re-order the fields to have plain `device,`

src/backend/dx11/src/device.rs
```diff
@@ -1,983 +1,1322 @@
...
+        
+        if usage.contains(Usage::TRANSFER_SRC) ||
+           usage.contains(Usage::SAMPLED) ||
+           usage.contains(Usage::STORAGE) { bind |= d3d11::D3D11_BIND_SHADER_RESOURCE; }
+
+        if usage.contains(Usage::COLOR_ATTACHMENT) { bind |= d3d11::D3D11_BIND_RENDER_TARGET; }
+        if usage.contains(Usage::DEPTH_STENCIL_ATTACHMENT) { bind |= d3d11::D3D11_BIND_DEPTH_STENCIL; }
+
+        // TODO: how to do buffer copies
+        if usage.contains(Usage::TRANSFER_DST) { bind |= d3d11::D3D11_BIND_UNORDERED_ACCESS; }
```
kvark: note: in the future we'd also need to add bind options to support blitting

src/backend/dx11/src/device.rs
```diff
@@ -1,983 +1,1322 @@
...
+
+        println!("{:b}", bind);
```
kvark: use `info!` or `debug!`

src/backend/dx11/src/device.rs
```diff
@@ -1,983 +1,1322 @@
...
-                (winapi::D3D11_RTV_DIMENSION_TEXTURE2D, [level, 0, 0]),
-            (Kind::D2(_, _, _), None) if level == 0 =>
-                (winapi::D3D11_RTV_DIMENSION_TEXTURE2DMS, [0, 0, 0]),
-            (Kind::D2Array(_, _, nlayers, AaMode::Single), None) =>
-                (winapi::D3D11_RTV_DIMENSION_TEXTURE2DARRAY, [level, 0, nlayers as UINT]),
-            (Kind::D2Array(_, _, nlayers, AaMode::Single), Some(lid)) if lid < nlayers =>
-                (winapi::D3D11_RTV_DIMENSION_TEXTURE2DARRAY, [level, lid as UINT, 1+lid as UINT]),
-            (Kind::D2Array(_, _, nlayers, _), None) if level == 0 =>
-                (winapi::D3D11_RTV_DIMENSION_TEXTURE2DMSARRAY, [0, nlayers as UINT, 0]),
-            (Kind::D2Array(_, _, nlayers, _), Some(lid)) if level == 0 && lid
```
kvark: similarly, please use `log` crate macros

kvark: should put `unimplemented!()` in those arms?

kvark: let's assert/panic at least in those cases, not ignore

src/backend/dx11/src/internal.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        unsafe {
+            context.Unmap(
+                self.copy_info.as_raw() as _,
+                0,
+            );
+        }
+    }
+
+    pub fn copy_2d(&mut self,
```
kvark: wow, amazing stuff!
btw, don't you need to clean up/reset the state after this operation?

src/backend/dx11/src/lib.rs
```diff
@@ -120,7 +133,7 @@ impl hal::Instance for Instance {
                         heap_index: 0,
                     },
                     hal::MemoryType {
-                        properties: Properties::DEVICE_LOCAL | Properties::CPU_VISIBLE | Properties::CPU_CACHED,
+                        properties: Properties::DEVICE_LOCAL, //| Properties::CPU_VISIBLE | Properties::CPU_CACHED,
```
kvark: this is fully matching the first type now?

src/backend/dx11/src/lib.rs
```diff
@@ -1730,21 +961,56 @@ pub struct UnboundBuffer {
     requirements: memory::Requirements,
 }
 
+#[derive(Clone)]
+pub enum InternalBuffer {
+    Coherent(ComPtr<d3d11::ID3D11Buffer>),
+    NonCoherent {
+        device: ComPtr<d3d11::ID3D11Buffer>,
```
kvark: why do we need a device local counterpart for the non-coherent buffer type here?

src/backend/dx11/src/lib.rs
```diff
@@ -1769,14 +1037,27 @@ unsafe impl Sync for Image { }
 #[derivative(Debug)]
 pub struct ImageView {
     #[derivative(Debug="ignore")]
-    rtv: ComPtr<d3d11::ID3D11RenderTargetView>
+    rtv_handle: Option<ComPtr<d3d11::ID3D11RenderTargetView>>,
+    #[derivative(Debug="ignore")]
```
kvark: I don't think `derive(Debug)` makes any sense at all here given that all the fields are ignored :)

src/backend/dx11/src/lib.rs
```diff
@@ -1836,8 +1116,30 @@ impl hal::DescriptorPool<Backend> for DescriptorPool {
     }
 }
 
-#[derive(Debug)]
-pub struct DescriptorSet;
+#[derive(Derivative)]
+#[derivative(Debug)]
```
kvark: same here

retep998:
`winapi` 0.3.5 is published so this PR is no longer blocked on that.

src/backend/dx11/src/device.rs
fkaa: Did you comment on the right place? this section is from the old stuff in the file which got nuked, same for two comments below

src/backend/dx11/src/lib.rs
fkaa: Required by the backend trait:
```rust
type ImageView: fmt::Debug + Any + Send + Sync;
```

Supposedly you can do
```rust
#[derivative(Debug="ignore")]
struct ImageView {
    ...
}
```
on top of the struct, but then you start getting fun errors like this:

```rust
error[E0308]: mismatched types
    --> C:\Dev\gfx\src\backend\dx11\src\lib.rs:1036:10
     |
1036 | #[derive(Derivative, Clone)]
     |          ^^^^^^^^^^ expected enum `std::result::Result`, found ()
     |
     = note: expected type `std::result::Result<(), std::fmt::Error>`
                found type `()`
```

src/backend/dx11/src/lib.rs
fkaa: Still messing around with the memory types a bit. Comment was to make it implicitly ignore a certain heap IIRC (not the best way I know).

For the memory types I think it's better to start conservatively then increase support afterwards.

Right now I have:
* `DEVICE_LOCAL` which works as intended (maps to `USAGE_DEFAULT` and should work for both buffers and images)
* `DEVICE_LOCAL | CPU_VISIBLE | CPU_CACHED` which is a bit more flakey. This is backed by one resource with `USAGE_STAGING` and another one with `USAGE_DEFAULT`. Reasoning here is that only `STAGING` resources in dx11 have CPU read & write, while at the same time can't be bound to any stage in the pipeline (hence another resource for device local memory)

I'm not entirely sure how solid the last method is, but I'm not too worried about this area.

src/backend/dx11/src/device.rs
fkaa: I think my reasoning was that I wanted internal mutability (and I was fumbling a bit with the types..) for the struct but I don't think it actually needs it, at least for now.

src/backend/dx11/src/conv.rs
fkaa: This function is only used to construct the `D3D11_FILTER` enum through bit twiddling so I think 0 is fine here (see https://msdn.microsoft.com/en-us/library/windows/desktop/dn770367(v=vs.85).aspx#code-snippet-2).

Though considering it doesn't actually map to an enum like the other functions do in the module, might be good to rename it.

src/backend/dx11/src/internal.rs
fkaa: Yeah, strange that I didn't get an error from this before (image bound as UAV on CS, then SRV on PS), or maybe the rules are a bit more lax

src/backend/dx11/src/conv.rs
kvark: this function shouldn't know the only way it's used, technically :) if it can't produce a `D3D11_FILTER` under some circumstances, it should return `Option`

src/backend/dx11/src/lib.rs
kvark: Please check out the heap mapping in Metal: https://github.com/gfx-rs/gfx/blob/447a0da3ad43310b124e4632a532b4fcba1f27a8/src/backend/metal/src/device.rs#L138

It's not ideal, but at least there isn't as much bit twiddling going on in the user code. Maybe we could have those memory flags in D3D11 as well? Not exactly the same, of course, just in terms of organization.

kvark: I don't see why we'd want to provide `DEVICE_LOCAL` *and* `CPU_VISIBLE` if D3D11 doesn't have that locally. Can we just expose `CPU_VISIBLE` for staging resources and leave the copy to/from device local memory to the user?

fkaa: Problem is `USAGE_STAGING` only allows for CopyResource/CopySubresourceRegion, which requires resources to be of the same type, so no buffer -> image copies for example

src/backend/dx11/src/conv.rs
fkaa: It is only used for the bit stuff, so I'd consider just renaming it to `map_anisotropic_mask` or something similar

kvark:
Time to ship!
bors r+

PR #2110, opened 2018-06-05
dx12: Add dynamically sized CPU descriptor heaps

### LINKED ISSUE #1915: Color space issues with SRGB framebuffer in OpenGL
Short info header:
- GFX version: 0.17.1
- OS: macOS 10.13.4
- GPU: AMD Radeon R9 M370X 2048 MB graphics / Intel Iris Pro 1536 MB graphics

Today I started updating the `gfx` dependency in [`imgui-rs`](https://github.com/Gekkio/imgui-rs). One of the things I noticed is that the `imgui-gfx-renderer` looks washed out compared to `imgui-glium-renderer`. It took me all day to find the cause, which turned out to be a color space mismatch between the (linear space) colors provided by `imgui` in the vertex attribute buffer, and the `gfx_device_gl` backend [forcefully enabling SRGB on the framebuffer](https://github.com/gfx-rs/gfx/blob/v0.17/src/backend/gl/src/lib.rs#L281-L285).

I was able to reproduce it by patching the `gamma` example to use the [purple title bar color](https://github.com/ocornut/imgui/blob/v1.52/imgui.cpp#L744) from imgui instead of the grayscale. Here's the patch:

```diff
diff --git a/examples/gamma/main.rs b/examples/gamma/main.rs
index af29c5b4..b95a8c82 100644
--- a/examples/gamma/main.rs
+++ b/examples/gamma/main.rs
@@ -31,7 +31,7 @@ pub type DepthFormat = gfx::format::DepthStencil;
 gfx_defines!{
     vertex Vertex {
         pos: [f32; 2] = "a_Pos",
-        color: [f32; 3] = "a_Color",
+        color: [f32; 4] = "a_Color",
     }

     pipeline pipe {
@@ -41,13 +41,14 @@ gfx_defines!{
 }

 const QUAD: [Vertex; 4] = [
-    Vertex { pos: [ -0.5,  0.5 ], color: [0.0, 0.0, 0.0] },
-    Vertex { pos: [ -0.5, -0.5 ], color: [0.0, 0.0, 0.0] },
-    Vertex { pos: [  0.5, -0.5 ], color: [1.0, 1.0, 1.0] },
-    Vertex { pos: [  0.5,  0.5 ], color: [1.0, 1.0, 1.0] }
+    // Use imgui TitleBg color: https://github.com/ocornut/imgui/blob/v1.52/imgui.cpp#L744
+    Vertex { pos: [ -0.5,  0.5 ], color: [0.27, 0.27, 0.54, 0.83] },
+    Vertex { pos: [ -0.5, -0.5 ], color: [0.27, 0.27, 0.54, 0.83] },
+    Vertex { pos: [  0.5, -0.5 ], color: [0.27, 0.27, 0.54, 0.83] },
+    Vertex { pos: [  0.5,  0.5 ], color: [0.27, 0.27, 0.54, 0.83] }
 ];

-const CLEAR_COLOR: [f32; 4] = [0.5, 0.5, 0.5, 1.0];
+const CLEAR_COLOR: [f32; 4] = [1.0, 1.0, 1.0, 1.0];

 type SurfaceData = <<ColorFormat as Formatted>::Surface as SurfaceTyped>::DataType;

diff --git a/examples/gamma/shader/quad_120.glslv b/examples/gamma/shader/quad_120.glslv
index a64fe208..44a270ed 100644
--- a/examples/gamma/shader/quad_120.glslv
+++ b/examples/gamma/shader/quad_120.glslv
@@ -1,10 +1,10 @@
 #version 120

 attribute vec2 a_Pos;
-attribute vec3 a_Color;
+attribute vec4 a_Color;
 varying vec4 v_Color;

 void main() {
-    v_Color = vec4(a_Color, 1.0);
+    v_Color = a_Color;
     gl_Position = vec4(a_Pos, 0.0, 1.0);
 }
diff --git a/examples/gamma/shader/quad_150_core.glslv b/examples/gamma/shader/quad_150_core.glslv
index 2ce6461b..c438e2ba 100644
--- a/examples/gamma/shader/quad_150_core.glslv
+++ b/examples/gamma/shader/quad_150_core.glslv
@@ -1,10 +1,10 @@
 #version 150 core

 in vec2 a_Pos;
-in vec3 a_Color;
+in vec4 a_Color;
 out vec4 v_Color;

 void main() {
-    v_Color = vec4(a_Color, 1.0);
+    v_Color = a_Color;
     gl_Position = vec4(a_Pos, 0.0, 1.0);
 }
diff --git a/examples/gamma/shader/quad_300_es.glslv b/examples/gamma/shader/quad_300_es.glslv
index f0f59952..3df5b91e 100644
--- a/examples/gamma/shader/quad_300_es.glslv
+++ b/examples/gamma/shader/quad_300_es.glslv
@@ -1,10 +1,10 @@
 #version 300 es

 in vec2 a_Pos;
-in vec3 a_Color;
+in vec4 a_Color;
 out vec4 v_Color;

 void main() {
-    v_Color = vec4(a_Color, 1.0);
+    v_Color = a_Color;
     gl_Position = vec4(a_Pos, 0.0, 1.0);
 }
```

And here's a screenshot, showing the color rendered incorrectly:

screen shot 2018-04-07 at 11 53 31 pm

When I disable the SRGB frame buffer (using unsafe code), the rendering works as expected:

```rust
// XXX: Fix incorrect render results
unsafe {
    device.with_gl(|gl| {
        gl.Disable(gfx_gl::FRAMEBUFFER_SRGB);
    });
}
```

screen shot 2018-04-07 at 11 56 27 pm

As a final data point, here are some screenshots of the `imgui-rs` hello_gfx example; the latter screenshot has the SRGB disable workaround:

screen shot 2018-04-07 at 11 04 03 pm

screen shot 2018-04-07 at 11 04 20 pm

May be related to: #992 #997

### LINKED ISSUE #1945: [dx12] Increase max RTV allocations (or make it configurable).

msiglreith:
Dynamic allocator for CPU descriptor heaps. We currently have 2 ways of allocating CPU descriptors: from a device global heap and command local linear allocators. This PR cleans up the two paths a bit and fixes the current size limitation for the first kind by dynamically creating new small fixed size CPU heaps.

Freeing currently not support but that's not an regression as we didn't support it before either.
Small 'regression' concerning `fill_buffer` which was implemented incorrectly as far as I can see. Probably would work only on AMD. This breaks the fill reftests.

Fixes #1945 
PR checklist:
- [x] tested quad, relevant CTS and a few vulkan samples with the following backends: dx12

### DIFF
--- a/src/backend/dx12/src/descriptors_cpu.rs
+++ b/src/backend/dx12/src/descriptors_cpu.rs
@@ -49,7 +49,7 @@ impl HeapLinear {
     }
 
     pub fn alloc_handle(&mut self) -> d3d12::D3D12_CPU_DESCRIPTOR_HANDLE {
-        assert!(!self.full());
+        assert!(!self.is_full());
 
         let slot = self.num;
         self.num += 1;
@@ -59,7 +59,7 @@ impl HeapLinear {
         }
     }
 
-    pub fn full(&self) -> bool {
+    pub fn is_full(&self) -> bool {
         self.num >= self.size
     }
 
@@ -68,15 +68,15 @@ impl HeapLinear {
     }
 }
 
-const HEAP_SIZE_FIXED: u32 = 64;
+const HEAP_SIZE_FIXED: usize = 64;
 
 // Fixed-size free-list allocator for CPU descriptors.
 struct Heap {
-    // Bit flag representation of occupied handles in the heap.
+    // Bit flag representation of available handles in the heap.
     //
     //  0 - Occupied
     //  1 - free
-    occupancy: u64,
+    availability: u64,
     handle_size: usize,
     start: d3d12::D3D12_CPU_DESCRIPTOR_HANDLE,
     _raw: ComPtr<d3d12::ID3D12DescriptorHeap>,
@@ -86,7 +86,7 @@ impl Heap {
     pub fn new(device: &ComPtr<d3d12::ID3D12Device>, ty: d3d12::D3D12_DESCRIPTOR_HEAP_TYPE) -> Self {
         let desc = d3d12::D3D12_DESCRIPTOR_HEAP_DESC {
             Type: ty,
-            NumDescriptors: HEAP_SIZE_FIXED,
+            NumDescriptors: HEAP_SIZE_FIXED as _,
             Flags: d3d12::D3D12_DESCRIPTOR_HEAP_FLAG_NONE,
             NodeMask: 0,
         };
@@ -104,27 +104,26 @@ impl Heap {
 
         Heap {
             handle_size,
-            occupancy: !0, // all free!
+            availability: !0, // all free!
             start,
             _raw: unsafe { ComPtr::from_raw(heap) },
         }
     }
 
     pub fn alloc_handle(&mut self) -> d3d12::D3D12_CPU_DESCRIPTOR_HANDLE {
-        assert!(!self.full());
-
         // Find first free slot.
-        let slot = self.occupancy.trailing_zeros() as usize;
+        let slot = self.availability.trailing_zeros() as usize;
+        assert!(slot < HEAP_SIZE_FIXED);
         // Set the slot as occupied.
-        self.occupancy ^= 1 << slot;
+        self.availability ^= 1 << slot;
 
         d3d12::D3D12_CPU_DESCRIPTOR_HANDLE {
             ptr: self.start.ptr + self.handle_size * slot,
         }
     }
 
-    pub fn full(&self) -> bool {
-        self.occupancy == 0
+    pub fn is_full(&self) -> bool {
+        self.availability == 0
     }
 }
 
@@ -161,7 +160,7 @@ impl DescriptorCpuPool {
 
         let heap = &mut self.heaps[heap_id];
         let handle = heap.alloc_handle();
-        if heap.full() {
+        if heap.is_full() {
             self.free_list.remove(&heap_id);
         }
 
--- a/src/backend/dx12/src/descriptors_cpu.rs
+++ b/src/backend/dx12/src/descriptors_cpu.rs
@@ -11,7 +11,7 @@ pub struct HeapLinear {
     num: usize,
     size: usize,
     start: d3d12::D3D12_CPU_DESCRIPTOR_HANDLE,
-    raw: ComPtr<d3d12::ID3D12DescriptorHeap>,
+    _raw: ComPtr<d3d12::ID3D12DescriptorHeap>,
 }
 
 impl HeapLinear {
@@ -44,7 +44,7 @@ impl HeapLinear {
             num: 0,
             size,
             start,
-            raw: unsafe { ComPtr::from_raw(heap) },
+            _raw: unsafe { ComPtr::from_raw(heap) },
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/backend/dx12/src/descriptors_cpu.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn full(&self) -> bool {
+        self.num >= self.size
+    }
+
+    pub fn clear(&mut self) {
+        self.num = 0;
+    }
+}
+
+// Fixed-size (64) free-list allocator for CPU descriptors.
```
kvark: let's have this defined as a constant here

src/backend/dx12/src/descriptors_cpu.rs
```diff
@@ -0,0 +1,165 @@
...
+            handle_size,
+            occupancy: 0,
+            start,
+            raw: unsafe { ComPtr::from_raw(heap) },
+        }
+    }
+
+    pub fn alloc_handle(&mut self) -> d3d12::D3D12_CPU_DESCRIPTOR_HANDLE {
+        // Find first free slot
+        let slot = (0..64)
```
kvark: please add TODO to use `https://doc.rust-lang.org/std/intrinsics/fn.ctlz.html (or actually do it behind the `nightly` feature ;) )

src/backend/dx12/src/descriptors_cpu.rs
```diff
@@ -0,0 +1,172 @@
...
+
+const HEAP_SIZE_FIXED: u32 = 64;
+
+// Fixed-size free-list allocator for CPU descriptors.
+struct Heap {
+    // Bit flag representation of occupied handles in the heap.
+    //
+    //  0 - Occupied
+    //  1 - free
+    occupancy: u64,
```
kvark: should be renamed into `availability` to avoid confusion

src/backend/dx12/src/descriptors_cpu.rs
```diff
@@ -0,0 +1,172 @@
...
+        let slot = self.occupancy.trailing_zeros() as usize;
+        // Set the slot as occupied.
+        self.occupancy ^= 1 << slot;
+
+        d3d12::D3D12_CPU_DESCRIPTOR_HANDLE {
+            ptr: self.start.ptr + self.handle_size * slot,
+        }
+    }
+
+    pub fn full(&self) -> bool {
```
kvark: given that this is not an accessor, I don't think it's idiomatic to omit the verb here. Should be `is_full()` to mimic `is_empty()` in standard collections

src/backend/dx12/src/descriptors_cpu.rs
```diff
@@ -0,0 +1,172 @@
...
+            start,
+            _raw: unsafe { ComPtr::from_raw(heap) },
+        }
+    }
+
+    pub fn alloc_handle(&mut self) -> d3d12::D3D12_CPU_DESCRIPTOR_HANDLE {
+        assert!(!self.full());
+
+        // Find first free slot.
+        let slot = self.occupancy.trailing_zeros() as usize;
```
kvark: super nice! What's the result if we were running it on `!0`? should we perhaps move the assert down here checking for `slot < HEAP_SIZE_FIXED`?

msiglreith: !0 returns 0.
yes, that assert check looks a bit nicer.

kvark:
Greatness here!
bors r+

Issue #1949, opened 2018-04-18
Dynamic buffer descriptors


msiglreith:
Add support for dynamic uniform buffers and dynamic storage buffers.

Implementation:

* Vulkan: obvious
* DX12: Descriptor + Root constant for offset maybe? Separate root constant space for all dynamic descriptor offset and updated on BindDescriptorSet. Further limits the available place in the root signature (32 root constants, 8 descriptor sets (2 slots due to sampler separation),  8 + 4 root constants for dynamic, 3 root constants for workgroup size, 1 spare for now). Other ideas?
* Metal: No idea

Edit: Update dx12 root signature split up

kvark:
The current status is - the feature is implemented and working well on Vulkan and Metal.
DX12/11 needs this covered as well, as it's blocking [rx](https://www.reddit.com/r/rust/comments/dauizc/rx_v020_released_a_modern_pixel_editor/f21psn6/) and potentially Amethyst from using the backend. See [code](https://github.com/gfx-rs/gfx/blob/543b521773a17f5f9d58a9ee3cc6a22b6442d1d7/src/backend/dx12/src/command.rs#L213).

msiglreith:
Note: Using root descriptors for implementing it in DX12 should be the easier approach compared to the one outlined. It should occupy the same space in the root signature.

msiglreith:
With the implementation of uniform buffers as root descriptors we end up with a root signature layout (64slots):
- 32 root constants -> 128bytes (Vulkan minimum)
- 8 descriptor tables for 4 descriptor sets (CBV/SRV/UAV and Sampler tables split)
- 8 root descriptors for dynamic uniform buffers -> 2x8 slots (Vulkan Minimum)

This leafs us with 8 slots for dynamic storage buffers, https://github.com/gfx-rs/gfx/issues/2604/ and https://github.com/gfx-rs/gfx/issues/2589

Root descriptors for SRV and UAV are different which would require us to know which resource type is used on the shader side. We don't have this information in the host side API (see https://github.com/gfx-rs/gfx/issues/2933).

In case we assume 4 slots for the https://github.com/gfx-rs/gfx/issues/2604/ and https://github.com/gfx-rs/gfx/issues/2589, we end up with 4 slots -> 2 root descriptors for storage buffers. As we would have to bind both SRV and UAV if we follow the approach used for descriptor tables, it would end up with 1 storage buffer max which would not fit the vulkan minimum.

[...]

PR #2111, opened 2018-06-05
Dynamic buffers

msiglreith:
Start defining the basic API for dynamic buffer descriptors. Atm based upon vulkan, not optimal but provides the best performance. Other possbiel API design could use `(set, Option<Offset>)` for descriptor set binding, but involves more work on the portability layer and might be more typing for users.

Feedback appreciated.
More backends to follow ..

Addresses https://github.com/gfx-rs/gfx/issues/1949
PR checklist:

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -166,15 +166,18 @@ impl PipelineCache {
         }
     }
 
-    fn bind_descriptor_sets<'a, T>(
+    fn bind_descriptor_sets<'a, I, J>(
         &mut self,
         layout: &n::PipelineLayout,
         first_set: usize,
-        sets: T,
+        sets: I,
+        offsets: J,
     ) -> [*mut d3d12::ID3D12DescriptorHeap; 2]
     where
-        T: IntoIterator,
-        T::Item: Borrow<n::DescriptorSet>,
+        I: IntoIterator,
+        I::Item: Borrow<n::DescriptorSet>,
+        J: IntoIterator,
+        J::Item: Borrow<com::DescriptorSetOffset>,
     {
         let mut sets = sets.into_iter().peekable();
         let (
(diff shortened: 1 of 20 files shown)


src/hal/src/command/compute.rs
```diff
@@ -4,6 +4,7 @@ use std::borrow::Borrow;
 
 use {Backend, WorkGroupCount};
 use buffer::Offset;
+use command::raw::DescriptorSetOffset;
 use queue::capability::{Compute, Supports};
 use super::{CommandBuffer, RawCommandBuffer, Shot, Level};
```
kvark: should probably include the `DescriptorSetOffset` here

src/hal/src/command/graphics.rs
```diff
@@ -5,6 +5,7 @@ use std::ops::Range;
 use Backend;
 use {image, pso};
 use buffer::IndexBufferView;
+use command::raw::DescriptorSetOffset;
```
kvark: similarly, let's move that down into `use super::{...}`

msiglreith:
> So, to clarify: the length of this extra iterator should be equal to the number of dynamic buffers in the given descriptor sets, right?

y, that's correct.

src/backend/dx12/src/command.rs
```diff
@@ -166,15 +166,18 @@ impl PipelineCache {
...
+        sets: I,
+        offsets: J,
     ) -> [*mut d3d12::ID3D12DescriptorHeap; 2]
     where
-        T: IntoIterator,
-        T::Item: Borrow<n::DescriptorSet>,
+        I: IntoIterator,
+        I::Item: Borrow<n::DescriptorSet>,
+        J: IntoIterator,
+        J::Item: Borrow<com::DescriptorSetOffset>,
```
kvark: Actually, I don't think there is much value in going with `Borrow<>` here, since the offsets are easily copyable. Mind changing it to just `IntoIterator<Item = com::DescriptorSetOffset>`?

msiglreith: sure, beside that I'm done with the PR from my side as far as I can see.

kvark: great! i'm also ready with Metal side, just blocked on that `IntoIterator` change

msiglreith: hold up, actually not that much a fan of the `intoIterator` change, makes using slices difficult/impossible as far as I can see.

kvark: aha, good point

src/backend/dx12/src/command.rs
```diff
@@ -166,15 +166,18 @@ impl PipelineCache {
...
     }
 
-    fn bind_descriptor_sets<'a, T>(
+    fn bind_descriptor_sets<'a, I, J>(
         &mut self,
         layout: &n::PipelineLayout,
         first_set: usize,
-        sets: T,
+        sets: I,
+        offsets: J,
```
grovesNL: Where is `offsets` used? Or should it be `_offsets`?

kvark:
Bors r+
Dx12/11 impls to follow separately

PR #2115, opened 2018-06-06
[mtl] basic stencil support

kvark:
PR checklist:


src/backend/metal/src/device.rs
```diff
@@ -179,7 +179,7 @@ impl PhysicalDevice {
                 argument_buffers: Self::supports_any(device, ARGUMENT_BUFFER_SUPPORT) && false, //TODO
                 shared_textures: !Self::is_mac(device),
                 format_depth24_stencil8: device.d24_s8_supported(),
-                format_depth32_stencil8: false, //TODO: crashing the Metal validation layer upon copying from buffer
+                format_depth32_stencil8: true, //TODO: crashing the Metal validation layer upon copying from buffer
```
grovesNL: So this works now? 😃

src/backend/metal/src/device.rs
```diff
@@ -613,13 +613,25 @@ impl hal::Device<Backend> for Device {
             .map(|attachment| attachment.borrow().clone())
             .collect::<Vec<_>>();
         let mut color_channels = Vec::new();
+        let mut depth_stencil_aspects = format::Aspects::empty();
+
         for attachment in &attachments {
             let is_depth = match attachment.format {
                 Some(f) => f.is_depth(),
```
grovesNL: Probably better to check `|| f.is_stencil()` too (i.e. `is_depth_or_stencil` for `S8Uint` we don't have that in Metal yet though)

src/backend/metal/src/native.rs
```diff
@@ -34,6 +34,7 @@ pub struct RenderPass {
     pub(crate) desc: metal::RenderPassDescriptor,
     pub(crate) attachments: Vec<pass::Attachment>,
     pub(crate) color_channels: Vec<Option<Channel>>,
+    pub(crate) depth_stencil_aspects: hal::format::Aspects,
```
grovesNL: What's the reason we need to track this separately outside of `desc`? As an optimization for state switches later?

src/backend/metal/src/device.rs
kvark: not really, but Dota uses that, and we should just file an actual Apple bugreport if this doesn't work instead of avoiding it

src/backend/metal/src/device.rs
kvark: yep

src/backend/metal/src/native.rs
kvark: In `create_framebuffer` we need to know which attachments to assign the depth/stencil texture. Would you think that checking for `descriptor.xxx_attachment().is_some()` would achieve the same thing cleaner?

grovesNL: IMO less state tracking is preferable if there's no significant disadvantage

kvark:
Thank you!
Fixed now.
Bors r=grovesNL

PR #2116, opened 2018-06-06
Better MTLPixelFormat and Metal Feature Set Support

### LINKED ISSUE #2049: Query image format capabilties on Metal
A lot of CTS tests are filtered based on what we return in `format_properties()` and `image_format_properties`. The implementations in Metal backend are barebones, unlike those in D3D12. Ideally, we'll need to encode the last section of the official [feature tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) in the code.

Thomspoon:
Fixes #2049 
PR checklist:
- [x] Add remaining formats
- [x] Add private capabilities for compression formats
- [x] Map format_properties to device capabilities

This WIP is meant to add support for all [MTLPixelFormats](https://developer.apple.com/documentation/metal/mtlpixelformat) and enable the [feature sets](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) accordingly.


src/hal/src/format.rs
```diff
@@ -255,11 +255,14 @@ surface_types! {
     R4_G4               {  !8, COLOR, (1, 1), color: 8 },
     R4_G4_B4_A4         { !16, COLOR, (1, 1), color: 12, alpha: 4 },
     B4_G4_R4_A4         { !16, COLOR, (1, 1), color: 12, alpha: 4 },
+    A4_B4_G4_R4         { !16, COLOR, (1, 1), color: 12, alpha: 4 },
```
kvark: hold on, we don't want to introduce any new HAL formats. Those are 1:1 mapped to Vulkan for the sake of portability initiative. If something can't be mapped to Metal, that becomes a problem of the Metal backend, not HAL.

src/backend/metal/src/conversions.rs
```diff
@@ -155,6 +155,114 @@ impl PrivateCapabilities {
             }
         }
     }
+
+    pub fn map_format_properties(&self, format: MTLPixelFormat) -> Properties {
+        use hal::format::ImageFeature as feat;
```
kvark: let's use the naming conventions - should start with a capital letter

src/backend/metal/src/conversions.rs
```diff
@@ -1,7 +1,7 @@
 use PrivateCapabilities;
 
 use hal::{pass, image, pso, IndexType};
-use hal::format::{Format, Swizzle};
+use hal::format::{Format, Swizzle, ImageFeature, Properties, BufferFeature};
```
kvark: at this point, just add `format` to the list above

src/backend/metal/src/conversions.rs
```diff
@@ -155,6 +155,114 @@ impl PrivateCapabilities {
...
+
+    pub fn map_format_properties(&self, format: MTLPixelFormat) -> Properties {
+        use hal::format::ImageFeature as feat;
+        use metal::MTLPixelFormat::*;
+        // use ImageFeature::SAMPLED_LINEAR as filter;
+        // use ImageFeature::STORAGE as write;
+        // use ImageFeature::COLOR_ATTACHMENT as color;
+        // use ImageFeature::COLOR_ATTACHMENT_BLEND as blend;
+
+        macro_rules! props {
```
kvark: at the end of the day, we'll have more variation between formats than just a few bits of `optimal_tiling`, which this macro implies. Thus, the macro would have to grow and eventually become useless. Maybe we should start on the proper path right away, and do something like:
```rust
let defaults = Properties {
                    linear_tiling: If::empty(),
                    optimal_tiling: If::SAMPLED | If::BLIT_SRC | If::BLIT_DST,
                    buffer_features: BufferFeature::all(),
};
match format {
  some_format => Properties {
    optimal_tiling: If::SAMPLED_LINEAR | defaults.optimal_tiling,
    .. defaults
  },
}
```

src/backend/metal/src/device.rs
```diff
@@ -283,6 +319,15 @@ impl PhysicalDevice {
                 format_bc: Self::is_mac(device),
                 format_eac_etc: !Self::is_mac(device),
                 format_astc: Self::supports_any(device, ASTC_PIXEL_FORMAT_FEATURES),
+                format_r8unorm_srgb_all: Self::supports_any(device, R8UNORM_SRGB_ALL),
```
kvark: could this fit into the old "format_min_srgb_channels" property?

Thomspoon: @kvark Not exactly, there are some versions of iOS and tvOS that don't support it.

src/backend/metal/src/conversions.rs
```diff
@@ -132,6 +153,661 @@ impl PrivateCapabilities {
...
+
+        let defaults = Properties {
+            linear_tiling: If::empty(),
+            optimal_tiling: If::SAMPLED | If::BLIT_SRC | If::BLIT_DST,
+            buffer_features: BufferFeature::all(),
+        };
+
+        match format {
+            A8Unorm => Properties {
+                optimal_tiling: If::SAMPLED_LINEAR,
```
kvark: I believe what you actually want (here and below) is:
```rust
optimal_tiling: defaults.optimal_tiling
  | If::EXTRA_1
  | If::EXTRA_2,
```

src/backend/metal/src/conversions.rs
```diff
@@ -132,6 +153,661 @@ impl PrivateCapabilities {
...
+                ..defaults
+            },
+            RG11B10Float if self.format_rg11b10_all => Properties {
+                optimal_tiling: If::SAMPLED_LINEAR
+                    | If::STORAGE
+                    | If::COLOR_ATTACHMENT
+                    | If::COLOR_ATTACHMENT_BLEND,
+                ..defaults
+            },
+            RG11B10Float if self.format_rg11b10_no_write => Properties {
```
kvark: are `xx_no_write` and `xxx_all` excluding? In this case, we can omit the second `if` here, it will just drop through

kvark: even better, since the difference is only one flag, we should move the `if` in there:
```rust
...
  | if self.format_xxx_all { If::STORAGE } else { If::empty() }
```

src/backend/metal/src/conversions.rs
```diff
@@ -132,6 +153,661 @@ impl PrivateCapabilities {
...
+                    | If::COLOR_ATTACHMENT_BLEND,
+                ..defaults
+            },
+            BGR10A2Unorm if self.format_bgr10a2_no_write => Properties {
+                optimal_tiling: If::SAMPLED_LINEAR
+                    | If::COLOR_ATTACHMENT
+                    | If::COLOR_ATTACHMENT_BLEND,
+                ..defaults
+            },
+            _ => defaults,
```
kvark: are those the compressed formats being excluded?

src/backend/metal/src/conversions.rs
```diff
@@ -202,75 +878,91 @@ pub fn map_blend_op(operation: &pso::BlendOp) -> (MTLBlendOperation, MTLBlendFac
...
-        f::Rg16Int     => Short2,
-        f::Rg16Float   => Half2,
-        f::Rgb16Unorm  => UShort3Normalized,
-        f::Rgb16Inorm  => Short3Normalized,
-        f::Rgb16Uint   => UShort3,
-        f::Rgb16Int    => Short3,
-        f::Rgb16Float  => Half3,
+    use metal::MTLVertexFormat::*;
+    Some(match format {
+        f::R8Unorm => UCharNormalized,
```
kvark: please don't mix formatting changes with functional ones, unless it's too late by now :)

src/backend/metal/src/device.rs
```diff
@@ -44,6 +44,119 @@ const ARGUMENT_BUFFER_SUPPORT: &[MTLFeatureSet] = &[
     MTLFeatureSet::macOS_GPUFamily1_v3,
 ];
 
+const ASTC_PIXEL_FORMAT_FEATURES: &[MTLFeatureSet] = &[
+    MTLFeatureSet::iOS_GPUFamily2_v1,
```
kvark: I don't think we need multiple versions of each family - only the base version (where support appeared) needs to be checked

src/backend/metal/src/device.rs
```diff
@@ -262,9 +375,51 @@ impl PhysicalDevice {
...
-                format_depth32_stencil8: true, //TODO: crashing the Metal validation layer upon copying from buffer
+                format_depth32_stencil8_filter: Self::is_mac(device), 
+                format_depth32_stencil8_none: !Self::is_mac(device),
                 format_min_srgb_channels: if Self::is_mac(&*device) {4} else {1},
                 format_b5: !Self::is_mac(device),
+                format_bc: Self::is_mac(device),
+                format_eac_etc: !Self::is_mac(device),
+                format_astc: Self::supports_any(device, ASTC_PIXEL_FORMAT_FEATURES),
+                format_r8unorm_srgb_all: Self::supports_any(device, R8UNORM_SRGB_ALL),
+                format_r8unorm_srgb_no_write: !Self::supports_any(device, R8UNORM_SRGB_ALL) && !Self::is_mac(device),
```
kvark: frankly, it does sound easier to just assign the flags right here instead of storing booleans only to transform them to flags later on. The image format flags are more expressive than a pair of `xxx_all` and `xxx_no_write` booleans.

kvark:
@Thomspoon do you have the capacity to revive and land this? Other PRs start to get blocked by the changes.
Note https://github.com/gfx-rs/metal-rs/pull/62 that exposes the APIs you could use now.

Thomspoon:
@kvark I'm not sure I'm going to be able to finish the work, it sounds like you guys want to revive it immediately, and i'm only able to dedicate a couple hours to it over the next couple weeks. Sorry...

src/backend/metal/src/conversions.rs
Thomspoon: Should be addressed.

src/backend/metal/src/conversions.rs
Thomspoon: Not sure I follow... If I remember right, there wasn't a pattern of feature-sets supporting certain flags, and it was rather arbitrary which is why I ended up having to have if-elseif-elseif-else in this case.

src/backend/metal/src/conversions.rs
Thomspoon: I believe so.

kvark:
Thank you for updating it!
Would you mind rebasing properly instead of introducing the merge commit?
Also, @AIOOB since you are going to be using the API, let us know if you want to have a look as well. It would be great!

kvark:
@Thomspoon I'm ok with a single commit, but it doesn't seem to have the author assigned to it? Also, CI is not happy, and a rebase is needed (conflicting files with master)>

Thomspoon:
@kvark I think I have it... Had to bring out the git-book this time haha.

grovesNL:
bors r=kvark

kvark:
There is a problem: one of the formats needed by Dota is now not exposed, so we aren't running it any more.

Thomspoon:
@kvark: What is the format? There could be a potential problem with an if-guard, or Dota mis-using a format?

kvark:
Still trying to figure this out... dota queries a bunch of formats and then just passes UNDEFINED for stuff it wasn't able to find :(

PR #2117, opened 2018-06-06
Attachment clears API refinement, DX12 depth/stencil support

kvark:
Fixes https://github.com/szeged/webrender/pull/183#issuecomment-395087515
PR checklist:

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -179,6 +179,8 @@ impl PipelineCache {
         J: IntoIterator,
         J::Item: Borrow<com::DescriptorSetOffset>,
     {
+        assert!(offsets.into_iter().next().is_none()); //TODO
+
         let mut sets = sets.into_iter().peekable();
         let (
             srv_cbv_uav_start, sampler_start,
@@ -1225,10 +1227,11 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         U: IntoIterator,
         U::Item: Borrow<pso::ClearRect>,
     {
-        assert!(
-            self.pass_cache.is_some(),
-            "`clear_attachments` can only be called inside a renderpass"
-        );
+        let pass_cache = match self.pass_cache {
+            Some(ref cache) => cache,
+            None => panic!("`clear_attachments` can only be called inside a renderpass")
+        };
+        let sub_pass = &pass_cache.render_pass.subpasses[self.cur_subpass];
 
         let clear_rects: SmallVec<[pso::ClearRect; 16]> = rects
             .into_iter()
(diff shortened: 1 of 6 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -653,17 +653,17 @@ impl Device {
 
     fn view_image_as_render_target(
         &self, info: ViewInfo
-    ) -> Result<d3d12::D3D12_CPU_DESCRIPTOR_HANDLE, image::ViewError> {
-        let mut pool = self.rtv_pool.lock().unwrap();
-        let handle = pool.alloc_handle();
-
+    ) -> Result<(d3d12::D3D12_CPU_DESCRIPTOR_HANDLE), image::ViewError> {
```
fkaa: Single element tuple? Does this still compile?

src/backend/dx12/src/command.rs
```diff
@@ -1282,12 +1278,54 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                        let view_info = device::ViewInfo {
+                            resource: attachment.resource,
+                            kind: attachment.kind,
+                            flags: image::StorageFlags::empty(),
+                            view_kind: image::ViewKind::D2Array,
+                            format: attachment.dxgi_format,
+                            range: image::SubresourceRange {
+                                aspects: if depth.is_some()  { Aspects::DEPTH } else { Aspects::empty() } |
+                                    if stencil.is_some() { Aspects::STENCIL } else {Aspects::empty() },
+                                levels: attachment.mip_levels.0 .. attachment.mip_levels.1,
```
fkaa: AFAICT dx12 only supports having one mip level per view, (`view_render_target_impl` and the dsv version only uses the start of the range)

fkaa: Though still an improvement over hardcoded 0!

fkaa: The same is true for vulkan though, so I think the correct thing to do here is to get the mip level that the attachment is using?

kvark: The check there, just happening a bit later in `view_image_as_*_impl`.
Technically we wouldn't be here if the user wanted to create a renderpass looking into multiple levels, so an assert is fine.

src/backend/dx12/src/device.rs
kvark: this is actually the old code. fixed now regardless :)

kvark:
Thanks for the review!
bors r=fkaa

src/backend/dx12/src/command.rs
kvark: We do assert on that later inside `view_image_as_*_impl`, and technically we shouldn't even come that far if the user wanted to render to multiple levels at once. They'd fail a framebuffer creation.

PR #2120, opened 2018-06-06
[mtl] diversity and inclusion of stencils, better binding code

### LINKED ISSUE #2119: Metal scissor rectangles
Metal doesn't like the scissor rect to be outside of the render area. Thus, we need to clamp the values all the time (when starting a render pass, then within a pass upon changing the scissor).
(marked as blocker for Dota2)

kvark:
Fixes #2119
PR checklist:


src/backend/metal/src/device.rs
```diff
@@ -1910,7 +1915,7 @@ impl hal::Device<Backend> for Device {
 
         if swizzle != format::Swizzle::NO {
             error!("swizzling not supported");
-            return Err(image::ViewError::Unsupported);
+            //return Err(image::ViewError::Unsupported);
```
grovesNL: Do we need to leave this out?

src/backend/metal/src/device.rs
```diff
@@ -971,11 +970,17 @@ impl hal::Device<Backend> for Device {
                 .expect("no associated vertex buffer found");
             // handle wrapping offsets
             let elem_size = element.format.surface_desc().bits as pso::ElemOffset / 8;
-            let (cut_offset, base_offset) = if element.offset + elem_size <= original.stride {
+            let stride = if original.stride != 0 {
+                original.stride
+            } else {
+                warn!("Zero sized vertex stride found!");
```
grovesNL: Is there a Vulkan reference for what's supposed to happen in this case? I don't see it explicitly disallowed

src/backend/metal/src/command.rs
```diff
@@ -2160,117 +2172,50 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-                                            index: start + i,
-                                            texture: texture.as_ref().map(|&(ref texture, _)| texture.clone()),
+                                        None => (None, 0),
+                                    };
+                                    for &mut (ref loc, ref mut resources) in &mut bind_stages {
+                                        let start = layout.res_overrides[loc].buffer_id as usize;
+                                        if let Some(buffer) = buffer {
+                                            resources.add_buffer(start + i, buffer, offset as _);
                                         }
-                                    }));
```
grovesNL: What happened to all the previous `commands.extend`/`push` here? Are they not actually required?

kvark: The old code was doing exactly the same thing for vertex and fragment stages. The problem came from a case where a dynamic buffer is used in both, so the code would try to advance the iterator twice for the same buffer...

The new code just builds a `bind_stages` vector (of size 2 at max) beforehand, and iterates it for each descriptor set type, thus bringing the amount of code duplication to a reasonable minimum as well as resolving the dynamic buffer issue.

src/backend/metal/src/device.rs
kvark: Molten just considers the stride of `float4` in this case, so I thought it doesn't matter that much.
But thinking about this more (as well as checking the documentation), the 0 stride has a very clear use case: repeat the attribute. What we should be doing in Metal then is switching it to be per-instance and setting the instance rate to the max.
Perhaps, in a follow-up?

src/backend/metal/src/device.rs
```diff
@@ -1131,8 +1139,8 @@ impl hal::Device<Backend> for Device {
                     .set_texture(Some(&at.raw));
             }
             if aspects.contains(format::Aspects::STENCIL) {
-                if let Some(mtl_format) = inner.depth_stencil {
-                    assert_eq!(mtl_format, at.mtl_format);
+                if let Some(old_format) = inner.depth_stencil {
```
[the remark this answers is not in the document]
kvark: Good catch!

src/backend/metal/src/device.rs
kvark: I don't think swizzling should be the end of the world. We've yet to figure out how to properly limit this with VkPI group, so just commenting the return out seems fine for now.

kvark:
Great review, thank you!
bors r=grovesNL

Issue #2121, opened 2018-06-06
Metal framebuffer compatibility with graphics pipelines


kvark:
We are hitting this error currently:
> -[MTLDebugRenderCommandEncoder validateFramebufferWithRenderPipelineState:]:1232: failed assertion `For depth attachment, the render pipeline's pixelFormat (MTLPixelFormatInvalid) does not match the framebuffer's pixelFormat (MTLPixelFormatDepth32Float_Stencil8).'

kvark:
I got an idea on why this may be happening. We keep track of the actively bound pipeline state, when outside of a render pass. And we do bind it at the start of the render pass. Now imagine if the application:
  1. starts render pass A with depth
  2. binds graphics pipeline 1
  3. draws some stuff
  4. ends rp A
  5. starts rb B without depth
  6. binds graphics pipeline 2
  7. draws some more
  8. ends rp B

Our backend currently will try to bind pipeline 1 at start of pass B, since it doesn't know if other pipelines are going to be bound later. But there is no guarantee that the pipeline is compatible with pass B, hence our Metal validation error.

PR #2124, opened 2018-06-07
Dynamic stencil masks

grovesNL:
Implemented only on Vulkan and Metal for now.

As mentioned I think the Metal implementation could be improved a lot by reworking depth stencil descriptors/states to make it more obvious how these states are being passed around and bound.

Fails in Dota 2 apparently due to the same reason as #2121: `self.state.depth_stencil_desc` does not exist at the time we try to update the stencil masks, so the assertion fails – not sure if I can attempt a local fix here..

PR checklist:
- [X] tested examples with the following backends: all

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -117,9 +117,9 @@ struct State {
     resources_cs: StageResources,
     index_buffer: Option<IndexBuffer>,
     rasterizer_state: Option<native::RasterizerState>,
-    depth_stencil_desc: Option<metal::DepthStencilState>,
-    stencil_front_reference: pso::StencilValue,
-    stencil_back_reference: pso::StencilValue,
+    pipeline_depth_stencil: Option<(pso::DepthStencilDesc, metal::DepthStencilState)>,
+    dynamic_depth_stencil: Option<metal::DepthStencilDescriptor>,
+    stencil: native::StencilState<pso::StencilValue>,
     push_constants: Vec<u32>,
     vertex_buffers: Vec<Option<(metal::Buffer, u64)>>,
     framebuffer_inner: native::FramebufferInner,
@@ -1041,9 +1041,16 @@ impl pool::RawCommandPool<Backend> for CommandPool {
                 resources_cs: StageResources::new(),
                 index_buffer: None,
                 rasterizer_state: None,
-                depth_stencil_desc: None,
-                stencil_front_reference: 0,
-                stencil_back_reference: 0,
+                pipeline_depth_stencil: None,
+                dynamic_depth_stencil: None,
+                stencil: native::StencilState::<pso::StencilValue> {
+                    front_reference: 0,
+                    back_reference: 0,
+                    front_read_mask: !0,
+                    back_read_mask: !0,
+                    front_write_mask: !0,
+                    back_write_mask: !0,
+                },
                 push_constants: Vec::new(),
                 vertex_buffers: Vec::new(),
                 framebuffer_inner: native::FramebufferInner {
@@ -1211,10 +1218,11 @@ impl CommandBuffer {
 
     fn set_depth_stencil_desc(
         &mut self,
-        depth_stencil_desc: &metal::DepthStencilState,
+        depth_stencil_desc: &pso::DepthStencilDesc,
+        depth_stencil_raw: &metal::DepthStencilState,
     ) -> soft::RenderCommand {
-        self.state.depth_stencil_desc = Some(depth_stencil_desc.clone());
-        soft::RenderCommand::SetDepthStencilDesc(depth_stencil_desc.clone())
+        self.state.pipeline_depth_stencil = Some((depth_stencil_desc.clone(), depth_stencil_raw.clone()));
+        soft::RenderCommand::SetDepthStencilDesc(depth_stencil_raw.clone())
     }
 
     fn set_stencil_reference_values(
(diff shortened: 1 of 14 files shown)


src/backend/metal/src/command.rs
```diff
@@ -1226,6 +1239,57 @@ impl CommandBuffer {
         self.state.stencil_back_reference = back;
         soft::RenderCommand::SetStencilReferenceValues(front, back)
     }
+
+    fn set_stencil_mask_values(
+        &mut self,
+        front_back_read_masks: Option<(pso::StencilValue, pso::StencilValue)>,
```
kvark: I'm confused about the semantics of this `Option`. Does it mean that the read masks are not used, or that they are not provided (but could still be used)?

src/backend/metal/src/command.rs
```diff
@@ -1969,7 +2033,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         let (front, back) = match faces {
             pso::Face::FRONT => (value, self.state.stencil_back_reference),
             pso::Face::BACK => (self.state.stencil_front_reference, value),
-            cf if cf == pso::Face::FRONT | pso::Face::BACK => (value, value),
+            f if f == pso::Face::FRONT | pso::Face::BACK => (value, value),
```
kvark: and here

src/backend/metal/src/command.rs
```diff
@@ -2069,31 +2177,86 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+            }
+            pso::State::Dynamic => self.state.stencil_back_read_mask,
+        };
+        if any_static_read_mask {
+            let com = self.set_stencil_mask_values(
+                Some((front, back)),
+                None,
+                ds.depth_stencil_desc
+            );
+            if let Some(c) = com {
```
kvark: can we do `commands.extend(...)` instead?

src/backend/metal/src/command.rs
```diff
@@ -1,5 +1,6 @@
 use {AutoreleasePool, Backend, Shared, validate_line_width};
 use {conversions as conv, native, soft, window};
+use device::{create_depth_stencil_state};
```
kvark: perhaps, we could have it as a static method on the device?

src/backend/metal/src/device.rs
```diff
@@ -113,6 +113,60 @@ fn get_final_function(library: &metal::LibraryRef, entry: &str, specialization:
...
 }
 
+pub fn create_depth_stencil_state(
+    device: &metal::DeviceRef,
+    desc: &pso::DepthStencilDesc,
+) -> native::DepthStencilState {
+    let raw = metal::DepthStencilDescriptor::new();
+
+    match desc.depth {
+        pso::DepthTest::On { fun, write } => {
```
kvark: makes me think that maybe the On/Off types don't work that well, and having an `Option<SomeStruct>` would end up being simpler and more convenient

src/backend/metal/src/command.rs
grovesNL: Just not provided – the current callsites are basically providing either read or write masks. Originally I had two functions `set_stencil_read_mask_values` and `set_stencil_write_mask_values` but I wanted to collapse them together to reduce the boilerplate. Is there a way to make this cleaner?

src/backend/metal/src/command.rs
```diff
@@ -117,9 +118,14 @@ struct State {
     resources_cs: StageResources,
     index_buffer: Option<IndexBuffer>,
     rasterizer_state: Option<native::RasterizerState>,
-    depth_stencil_desc: Option<metal::DepthStencilState>,
+    depth_stencil_desc: Option<pso::DepthStencilDesc>,
+    depth_stencil_raw: Option<metal::DepthStencilState>,
```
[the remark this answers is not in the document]
grovesNL: Just tried this but it seems like there are a couple places where we only use one, or it causes new issues with the borrow checker

src/backend/metal/src/device.rs
grovesNL: It's not too bad IMO, we could improve it with a helper function for these kinds of cases

grovesNL:
@kvark thanks for the review! PTAL at the updated commit

src/backend/metal/src/command.rs
```diff
@@ -1211,21 +1227,73 @@ impl CommandBuffer {
...
+                    (pso::StencilFace::default(), pso::StencilFace::default())
+                }
+            };
+
+            if let Some((f, b)) = front_back_read_masks {
+                front_face.mask_read = pso::State::Static(f);
+                back_face.mask_read = pso::State::Static(b);
+            }
+
+            if let Some((f, b)) = front_back_write_masks {
```
kvark: so I'm concerned about the following case:
  1. user calls function (Some, None). We update `self.state.stencil` fields for A, then create a new ds descriptor and produce a `SetDepthStencilDesc`
  2. user calls function (None, Some). We update fields for B (all good so far), then create a new ds descriptor that only has the values of B, where we need it to have both changes from the first and second calls by this time

grovesNL:
@kvark here's the new approach:
- Always hold onto a Metal dynamic depth stencil descriptor which receives any dynamic masks.
- Whenever a dynamic depth stencil command is submitted, we update the Metal dynamic depth stencil descriptor, create a new Metal depth stencil state from it, and set it for use through the Metal encoder.
- When a pipeline is bound, if it has a depth descriptor, clone it, and use it as the new Metal dynamic depth stencil descriptor.
- If a pipeline has no dynamic depth stencil state (i.e. all static), create a Metal depth stencil state immediately, and always use that instead of the dynamic depth stencil descriptor/state.

src/backend/metal/src/command.rs
```diff
@@ -24,7 +24,6 @@ use cocoa::foundation::{NSUInteger, NSInteger, NSRange};
 use block::{ConcreteBlock};
 use smallvec::SmallVec;
 
-
```
kvark: fwiw, I like leaving 2 empty lines after the `use` section to separate "headers" from "body"

src/backend/metal/src/command.rs
```diff
@@ -1211,21 +1227,82 @@ impl CommandBuffer {
...
     fn set_depth_stencil_desc(
         &mut self,
-        depth_stencil_desc: &metal::DepthStencilState,
+        depth_stencil_desc: &pso::DepthStencilDesc,
+        depth_stencil_raw: &metal::DepthStencilState,
     ) -> soft::RenderCommand {
-        self.state.depth_stencil_desc = Some(depth_stencil_desc.clone());
-        soft::RenderCommand::SetDepthStencilDesc(depth_stencil_desc.clone())
+        self.state.pipeline_depth_stencil = Some((depth_stencil_desc.clone(), depth_stencil_raw.clone()));
+        soft::RenderCommand::SetDepthStencilDesc(depth_stencil_raw.clone())
```
kvark: I wonder how expensive those clones (of metal stuff) are

src/backend/metal/src/command.rs
```diff
@@ -2069,29 +2177,53 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-
-        // If static stencil reference values were provided, update them here
-        // Otherwise, leave any dynamic stencil reference values bound
-        let mut any_static = false;
-        let front = match ds.stencil_front_reference {
-            pso::State::Static(f) => {
-                any_static = true;
-                f
+        if let Some(desc) = ds.depth_stencil_desc {
+            if let Some(ref raw) = ds.depth_stencil_static {
```
kvark: nit: `let command = match ds.depth_stencil_static {...}; commands.push(command);`

src/backend/metal/src/command.rs
```diff
@@ -2069,29 +2177,53 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                commands.push(self.set_depth_stencil_desc(&desc, raw));
+            } else {
+                let front_r = match ds.stencil_front_read_mask {
+                    pso::State::Static(f) => f,
```
kvark: perhaps, add a helper `static_or`?

src/backend/metal/src/device.rs
```diff
@@ -113,6 +113,90 @@ fn get_final_function(library: &metal::LibraryRef, entry: &str, specialization:
...
+    match desc.stencil {
+        pso::StencilTest::On { ref front, ref back } => {
+            dss.stencil_front_reference = front.reference;
+            dss.stencil_back_reference = back.reference;
+
+            let front_desc = metal::StencilDescriptor::new();
+            front_desc.set_stencil_compare_function(conv::map_compare_function(front.fun));
+
+            dss.stencil_front_read_mask = front.mask_read;
+            if let pso::State::Static(mr) = front.mask_read {
```
kvark: nit: let's just `match` instead of `if let + else`. I know this is opinionated, and I'm open to discuss it. We haven't come to a strong consensus in WebRender project, but maybe we could achieve it here? :)

src/backend/metal/src/native.rs
```diff
@@ -81,17 +81,29 @@ impl Default for RasterizerState {
 
 #[derive(Clone, Debug)]
 pub struct DepthStencilState {
-    pub depth_stencil_desc: Option<metal::DepthStencilState>,
+    pub depth_stencil_desc: Option<pso::DepthStencilDesc>,
+    pub depth_stencil_desc_raw: Option<metal::DepthStencilDescriptor>,
+    pub depth_stencil_static: Option<metal::DepthStencilState>,
     pub stencil_front_reference: pso::State<pso::StencilValue>,
```
kvark: Let's have our `Stencil` struct defined as generic over T being the actual stencil value, and thus we'd be able to re-use it here with `T == pso::State<pso::StencilValue>`

grovesNL: I just tried this and it doesn't give us much benefit IMO, except to reduce the repeated `pso::State<pso::StencilValue>` here (but I have to add it a few other places instead). What do you think?

src/backend/metal/src/device.rs
grovesNL: This actually started as an `if` originally, then I changed to `match`, then back to `if`, so I'm undecided too 😄

src/backend/metal/src/command.rs
grovesNL: Yeah, I don't know the answer to this either. We could possible optimize some of these, or perhaps pool these somewhere to reuse them...

kvark:
bors retry

PR #2125, opened 2018-06-07
[dx11] add format_properties and image_format_properties

fkaa:
Based on the dx12 counterpart. Fixes all of the crashes in `dEQP-VK.info.format_properties`

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -14,29 +14,27 @@ extern crate winapi;
 extern crate winit;
 extern crate wio;
 
-use hal::{buffer, command, error, format, image, memory, mapping, query, pool, pso, pass, Features, Limits, QueueType};
+use hal::{buffer, command, error, format, image, memory, query, pso, Features, Limits, QueueType};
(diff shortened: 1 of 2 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -334,12 +477,97 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
 
+    fn image_format_properties(&self, format: format::Format, dimensions: u8, tiling: image::Tiling, usage: image::Usage, storage_flags: image::StorageFlags) -> Option<image::FormatProperties> {
+        conv::map_format(format)?; //filter out unknown formats
 
-    fn image_format_properties(&self, _format: format::Format, dimensions: u8, tiling: image::Tiling, usage: image::Usage, storage_flags: image::StorageFlags) -> Option<image::FormatProperties> {
-        unimplemented!()
+        let supported_usage = {
+            use hal::image::Usage as U;
+            let format_props = &self.format_properties[format as usize];
+            let props = match tiling {
```
kvark: we should just `impl Index<Tiling> for `format::Properties` and use it in everywhere

PR #2128, opened 2018-06-08
[dx11] fix some dxgi calls not being present on windows 7

fkaa:
In particular the `CreateDXGIFactory2` was used to create the factory,
but it is not present on older dxgi versions. Other function calls
shouldn't be affected since the factory creation infers the dxgi version
and tip-toes around the newer functions.

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -1246,16 +1246,18 @@ impl hal::Device<Backend> for Device {
             unsafe { ComPtr::from_raw(swapchain) }
         };
 
-        let images = (0..config.image_count).map(|i| {
+        // TODO: for now we clamp to 1 buffer..
+        let images = (0..config.image_count.min(1)).map(|i| {
             let mut resource: *mut d3d11::ID3D11Resource = ptr::null_mut();
 
-            unsafe {
+            let hr = unsafe {
                 swapchain.GetBuffer(
                     i as _,
                     &d3d11::IID_ID3D11Resource,
                     &mut resource as *mut *mut _ as *mut *mut _
-                );
+                )
             };
+            assert_eq!(hr, winerror::S_OK);
 
             let mut desc: d3d11::D3D11_RENDER_TARGET_VIEW_DESC = unsafe { mem::zeroed() };
             desc.Format = format;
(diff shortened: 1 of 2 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -1246,16 +1246,18 @@ impl hal::Device<Backend> for Device {
             unsafe { ComPtr::from_raw(swapchain) }
         };
 
-        let images = (0..config.image_count).map(|i| {
+        // TODO: for now we clamp to 1 buffer..
```
kvark: why only one?

src/backend/dx11/src/dxgi.rs
```diff
@@ -105,15 +105,16 @@ fn create_dxgi_factory1(guid: &GUID) -> Result<ComPtr<dxgi::IDXGIFactory>, winer
 pub(crate) fn get_dxgi_factory() -> Result<(ComPtr<dxgi::IDXGIFactory>, DxgiVersion), winerror::HRESULT> {
     let mut factory: *mut IUnknown = ptr::null_mut();
 
-    if let Ok(factory) = create_dxgi_factory2(&dxgi1_5::IDXGIFactory5::uuidof()) {
+    // TODO: do we even need `create_dxgi_factory2`?
```
kvark: let's double check the PR that started using factory2

src/backend/dx11/src/device.rs
fkaa: Right now the swapchain is hardcoded to only use the `DXGI_SWAP_EFFECT_DISCARD` which as I understand it can only handle 1 backbuffer resource. All the others require w7+ which can be added in later: https://msdn.microsoft.com/en-us/library/windows/desktop/bb173077(v=vs.85).aspx

src/backend/dx11/src/dxgi.rs
fkaa: The only thing different with `CreateDXGIFactory2` is that you can specify `DXGI_CREATE_FACTORY_DEBUG`, which gets set by the runtime anyway: https://msdn.microsoft.com/en-us/library/windows/desktop/dn268307(v=vs.85).aspx

Using `CreateDXGIFactory1` will still return the interfaces from later versions.

fkaa:
@kvark appveyor CI missing?

PR #2129, opened 2018-06-08
[mtl] fix renderpass clear operations

kvark:
Our big problem was unexpected clearing of the render passes, happened because:
  1. the descriptor was created in `create_render_pass` with the clear ops setup
  2. then copied over in `create_framebuffer`, set with exact texture references
  3. *then* we'd copy that one again on `begin_render_pass`, and set the actual clear values..

The PR eliminates the first instance of the descriptor, which greatly simplifies the code.

PR checklist:


src/backend/metal/src/window.rs
```diff
@@ -16,7 +16,7 @@ use core_foundation::base::TCFType;
 use core_foundation::string::CFString;
 use core_foundation::dictionary::CFDictionary;
 use core_foundation::number::CFNumber;
-use core_graphics::base::CGFloat;
+//use core_graphics::base::CGFloat;
```
grovesNL: Are we keeping these in gfx for now? Does it break hal/quad on your Retina display?

src/backend/metal/src/command.rs
```diff
@@ -1092,6 +1102,13 @@ impl pool::RawCommandPool<Backend> for CommandPool {
     }
 }
 
+// Sets up the load/store operations. Returns `true` if the clear color needs to be set.
+fn set_operations(attachment: &metal::RenderPassAttachmentDescriptorRef, ops: AttachmentOps) -> bool {
+    attachment.set_load_action(conv::map_load_operation(ops.load));
+    attachment.set_store_action(conv::map_store_operation(ops.store));
+    ops.load == AttachmentLoadOp::Clear
```
grovesNL: nit: seems a bit confusing to do this here

src/backend/metal/src/window.rs
kvark: Excellent point! They are broken. How about we only apply the scale when initialized via `winit` and ignore it otherwise?

src/backend/metal/src/command.rs
kvark: yeah, I agree. It is sooo tempting though. Would it be cleaner if I return something like:
```rust
OperationFollowup {
  clear: bool,
}
```

src/backend/metal/src/command.rs
```diff
@@ -1109,11 +1109,11 @@ impl pool::RawCommandPool<Backend> for CommandPool {
     }
 }
 
-// Sets up the load/store operations. Returns `true` if the clear color needs to be set.
-fn set_operations(attachment: &metal::RenderPassAttachmentDescriptorRef, ops: AttachmentOps) -> bool {
+/// Sets up the load/store operations. Returns `true` if the clear color needs to be set.
```
grovesNL: nit: this comment is incorrect now

kvark:
bors r=grovesNL

Issue #1427, opened 2017-08-13
cargo format


bjadamson:
I found during my [implementation of this pull-request](https://github.com/bjadamson/gfx/pull/1), that a lot of the gfx project doesn't have cargo fmt run on it. My editor uses it by default on .rs files, can we run cargo format on the whole project, or do the developers of gfx prefer not using cargo-fmt?

msiglreith:
Personally, I use it more often on my `ll` branch but usually only on a per file basis and doubling check the result for some corner cases. On a side note, the used formatting style looks outdated imo.

I would like to enforce rustfmt for the project but in the linked PR is still see some occurences where our currenty formatting looks better imo:
- [Removal of intentional whitespaces](https://github.com/bjadamson/gfx/pull/1/files#diff-a73b73a5aa78db6d9acd6aa48e08c337L284)
- [Transform multi-line to single line comments for WIP code](https://github.com/bjadamson/gfx/blob/12e8192fd8c7fb4636e5e7058c9f4e6989bdb3d0/backend/vulkan/src/pool.rs#L127-L193)
- [Some corner cases here?](https://github.com/bjadamson/gfx/blob/12e8192fd8c7fb4636e5e7058c9f4e6989bdb3d0/backend/vulkan/src/device.rs#L109-L113)
- [Indentation issue with macros](https://github.com/bjadamson/gfx/blob/12e8192fd8c7fb4636e5e7058c9f4e6989bdb3d0/gfx_core/core/src/memory.rs#L147-L173)

nrc:
It looks like you might be using a very old (or customised) version of rustfmt. You'll get the best results by using the latest version of rustfmt-nightly.

AIOOB:
It might be sensible to look into this again, rustfmt has come a long way, there have been lots of new contributions (woo!) since this was opened and the style is getting very inconsistent in places. It might also be sensible looking into adding it as a CI test so that it can easily be enforced. #2052 sparked my curiosity in this as it makes a load of (possibly sensible I haven't really looked) formatting changes and it gets in the way of the actual changes.

nrc:
I'd definitely recommend using rustfmt now! It is a lot more stable and there are not too many bugs remaining. I would not recommend using it on CI just yet, there are still a few too many changes happening for that. I think it should be ready for CI use in a couple of months.

kvark:
Consider this fixed by #2131

PR #2131, opened 2018-06-08
Add Rustfmt to workflow

AIOOB:
This is not designed to be merged as-is but is here for discussion over whether this is a good idea/what options should be used. This covers most of @msiglreith's concerns on #1427 apart from the fact that the whitespace in the `Backend` trait gets removed. I've made comments on the few config options that it might make sense to tweak.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,1 @@
+newline_style = "Native"
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,3 +3,4 @@ PR checklist:
 - [ ] `make` succeeds (on *nix)
 - [ ] `make reftests` succeeds
 - [ ] tested examples with the following backends:
+- [ ] `rustfmt` run on changed code


rustfmt.toml
```diff
@@ -0,0 +1,55 @@
+indent_style = "Block"
+use_small_heuristics = true
+binop_separator = "Front"
+combine_control_expr = true
+comment_width = 80
```
AIOOB: This could be made large, clippy uses 100.

rustfmt.toml
```diff
@@ -0,0 +1,55 @@
...
+struct_field_align_threshold = 0
+spaces_around_ranges = false
+struct_lit_single_line = true
+tab_spaces = 4
+trailing_comma = "Vertical"
+trailing_semicolon = true
+type_punctuation_density = "Wide"
+use_field_init_shorthand = false
+use_try_shorthand = false
+wrap_comments = false
```
AIOOB: This could be enabled.

examples/hal/compute/main.rs
```diff
@@ -4,27 +4,27 @@
...
+extern crate gfx_backend_vulkan as back;
+extern crate gfx_hal as hal;
 
 use std::str::FromStr;
 
-use hal::{
-    Backend, Compute, Device, DescriptorPool, Instance, PhysicalDevice, QueueFamily,
-};
-use hal::{queue, pso, memory, buffer, pool, command};
+use hal::{buffer, command, memory, pool, pso, queue};
```
kvark: is it possible to have `rustfmt` preserve the old style here?

examples/hal/compute/main.rs
```diff
@@ -34,48 +34,54 @@ fn main() {
...
 
         let desc_pool = device.create_descriptor_pool(
             1,
-            &[
-                pso::DescriptorRangeDesc {
-                    ty: pso::DescriptorType::StorageBuffer,
-                    count: 1,
-                },
-            ],
+            &[pso::DescriptorRangeDesc {
```
kvark: I wonder how this will look like with multiple items in the array

rustfmt.toml
kvark: does it make a difference, given that comments are not wrapped? is it just for warnings?

rustfmt.toml
```diff
@@ -0,0 +1,55 @@
+indent_style = "Block"
+use_small_heuristics = true
+binop_separator = "Front"
+combine_control_expr = true
+comment_width = 80
+condense_wildcard_suffixes = false
+control_brace_style = "AlwaysSameLine"
+disable_all_formatting = false
```
kvark: are these all non-default? I think it would be sane to only specify non-default switches here, to avoid the temptation to deviate from those defaults

rustfmt.toml
```diff
@@ -0,0 +1,55 @@
...
+format_strings = false
+hard_tabs = false
+imports_indent = "Visual"
+imports_layout = "Mixed"
+merge_imports = false
+match_block_trailing_comma = false
+max_width = 100
+merge_derives = true
+force_multiline_blocks = false
+newline_style = "Unix"
```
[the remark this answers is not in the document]
kvark: yes, I think Native would make sense. Git for windows by default stores code with native line endings anyhow (and converts back and forth on checkout/commit)

rustfmt.toml
kvark: nope, we are good :)

src/backend/dx11/src/conv.rs
```diff
@@ -26,60 +24,60 @@ pub fn map_index_type(ty: IndexType) -> DXGI_FORMAT {
...
+            Some(DXGI_FORMAT_R8_TYPELESS)
+        }
+
+        DXGI_FORMAT_R8G8_UNORM
+        | DXGI_FORMAT_R8G8_SNORM
+        | DXGI_FORMAT_R8G8_UINT
+        | DXGI_FORMAT_R8G8_SINT => Some(DXGI_FORMAT_R8G8_TYPELESS),
+
+        DXGI_FORMAT_R16_UNORM
+        | DXGI_FORMAT_R16_SNORM
```
kvark: I find this controversial. It doesn't align with the first item, and it's not obvious to me what this improves

examples/hal/compute/main.rs
AIOOB: What is the old style, exactly?

rustfmt.toml
AIOOB: These are all the defaults, I have just included them here so that you can see all the available options. I'll remove them when everything has been decided on.

examples/hal/compute/main.rs
AIOOB: It indents it and puts it on a newline, e.g.:
```rust
&[
    struct {
        ..
    },
    struct {
        ..
    },
],
```

src/backend/dx11/src/conv.rs
AIOOB: While I agree it is not clearly better, there is no specific config option for this, `binop_seperator` which can be `"Front"` or `"Back"` however that effects all binary operators. This exact problem led to the allowance of a pipe at the start of a match clause, but I don't think that actually looks any better.

examples/hal/compute/main.rs
kvark: use xxx::{
    YY, ZZZ, WW,
    AA, BB,
};

kvark:
Sounds like my concerns are addressed for the most part :+1:

examples/hal/compute/main.rs
AIOOB: It kind of does that apart from preserving new lines:
https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#imports_layout

kvark: oh, I see. It split into multiple `use` statements here because one is for modules and another is for types. Otherwise, the "default" makes total sense to me (and I'm pleasantly surprised it is the default).

kvark:
bors retry

Issue #2134, opened 2018-06-09
mtl: Investigate failing CTS tests for dynamic_states


grovesNL:
Currently (as of https://github.com/gfx-rs/gfx/commit/68d1539f57d1b363c72c76ca46c98e7a687ba7ad and https://github.com/gfx-rs/portability/commit/2a03ef2acd90bc0f9960531457622222f6496e24) all of the CTS `dynamic_states` tests  return `Pass` or `NotSupported` on Metal, except a few depth and stencil tests:

* dEQP-VK.dynamic_state.rs_state.depth_bias
* dEQP-VK.dynamic_state.rs_state.depth_bias_clamp
* dEQP-VK.dynamic_state.ds_state.stencil_params_basic_1
* dEQP-VK.dynamic_state.ds_state.stencil_params_advanced

Let's look into this further and understand what goes wrong in these cases.

grovesNL:
For the `depth_bias` test we receive the following message from validation:
> validateRenderPassDescriptor:634: failed assertion `When depth and stencil are used together, the texture bound to the depth and stencil render pass attachments must be the same depth and stencil texture.'

grovesNL:
The validation error is caused during an image clear.

Currently, if we previously had depth and stencil attachments pointing some texture, if the subresource range targets a specific aspect like depth only, we will update the depth attachment to point to a new texture, but leave the stencil attachment pointing to the old texture. But both attachments are still active, so we can't ignore the stencil attachment here, and the stencil attachment needs to be updated as well.

So we can check the `image.format_desc.aspects` to determine whether the attachment needs to be updated. But the problem then becomes: how do we ensure that the stencil isn't cleared in this case? We could try simply setting `MTLLoadAction` to `Load` for the stencil and setting it to `Clear` for the depth, but because they're backed by the same texture internally – this (probably) produces undefined behavior or always chooses one of these actions.

See https://github.com/grovesNL/gfx/commit/774b247350aedc3d345500174644c6f16679a51e as an example of what this looks like. This fixes the validation errors, but doesn't help the CTS tests at all (may or may not be related).

I guess we should add a slow path for this?

kvark:
> We could try simply setting MTLLoadAction to Load for the stencil and setting it to Clear for the depth, but because they're backed by the same texture internally – this (probably) produces undefined behavior or always chooses one of these actions.

[...]

PR #2135, opened 2018-06-10
mtl: Set depth stencil state in `begin_render_pass`

grovesNL:
Fixes depth issue noticed by @scoopr (unfortunately doesn't fix the CTS issues in #2134)

Mistakenly removed during the `BindPipeline` changes in https://github.com/gfx-rs/gfx/commit/40d3647a32bb8d84c40179cf63857ceab57c4afa#diff-abc29e1de3205709a964ff5e3b0e6333L148

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -123,7 +123,8 @@ struct State {
     index_buffer: Option<IndexBuffer>,
     rasterizer_state: Option<native::RasterizerState>,
     pipeline_depth_stencil: Option<(pso::DepthStencilDesc, metal::DepthStencilState)>,
-    dynamic_depth_stencil: Option<metal::DepthStencilDescriptor>,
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/command.rs
```diff
@@ -123,7 +123,8 @@ struct State {
     index_buffer: Option<IndexBuffer>,
     rasterizer_state: Option<native::RasterizerState>,
     pipeline_depth_stencil: Option<(pso::DepthStencilDesc, metal::DepthStencilState)>,
-    dynamic_depth_stencil: Option<metal::DepthStencilDescriptor>,
+    dynamic_depth_stencil_desc: Option<metal::DepthStencilDescriptor>,
```
grovesNL: I tried collapsing these into one field but it was surprisingly difficult to make it work with `set_stencil_mask_values`, maybe somebody else has an idea?

PR #2136, opened 2018-06-10
[WIP] Merge window creation between backends

ZeGentzy:

Fixes #issue
PR checklist:
- [ ] Make the gl backend work
- [ ] Make the types better (I don't like wrapping everything in Arcs and Mutexes)
- [ ] Make other backends compile
- [ ] Insure it still compiles when winit and/or glutin are disabled
- [x] tested examples with the following backends: OpenGL and Vulkan back ends in ArchLinux


src/hal/src/lib.rs
```diff
@@ -375,6 +375,8 @@ pub trait Backend: 'static + Sized + Eq + Clone + Hash + fmt::Debug + Any + Send
     type Fence:               fmt::Debug + Any + Send + Sync;
     type Semaphore:           fmt::Debug + Any + Send + Sync;
     type QueryPool:           fmt::Debug + Any + Send + Sync;
+
+    type Window:                           Any + Send + Sync;
```
kvark: This shouldn't be here, ideally.
Could you explain the thought process that leaded you to this decision? GL should not affect our HAL API.

ZeGentzy:
Latest version was tested against https://github.com/tomaka/winit/commit/c873c2db1508bb420e7fa0518e6727c31154e1b8 and https://github.com/ZeGentzy/glutin/commit/cdad4a945486e49db5be1b2899d46e3bce5d3093 FYI

src/backend/gl/src/command.rs
```diff
@@ -886,9 +886,13 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             for new_binding in &*desc_set.bindings.lock().unwrap() {
                 match new_binding {
                     n::DescSetBindings::Buffer {ty: btype, binding, buffer, offset, size} => {
+                        let btype = match btype {
```
ZeGentzy: This is not strictly necessary and could be separated into a separate PR.

ZeGentzy:
A couple notes:
 - While the changes here haven't been tested with multiwindow setups w/ the opengl backend, it should make supporting them easier.
 - `Cargo.toml`s will be returned to there prior state
 - The changes here limit users of gfx to one events loop per instance (and given they will likely only have one instance, that's basically one events loop per program). I think this is 100% OK because after talking with Franseska I discovered that they don't officially support more than 1 event loop per application. Nor do they support event loops not on the main thread on macOS.
 - The gl code isn't the most simplistic and I'd be up for ideas at simplifying it.
 - Insuring we are always on the main context (unless when we need not) is a possible source of bugs which might be hard to track down. Up for ideas on how to eliminate them. (Maybe accessing a share should make it's context current? What if they try to access multiple contexts simultaneously?)
 - Now respecting the user's swap chain config is a couple changes away. I opted to not include it in this PR because I consider it out of this PRs scope (and more importantly, I'm lazy.)
 - We do open separate devices per window, but those are "handled" internally.)
 - Currently only works on linux (every backend) & windows (wgl), see https://github.com/tomaka/glutin/issues/899 and https://github.com/tomaka/glutin/pull/1031
- Once this is in, hopefully we can do the portability cts stuff with the opengl backend (yay!)
- Some changes here are to support glutin's (or was it winit?) new hidpi support model. (Don't remember which changes tho :( )

Anyways, I quick summary of this PR off the top of my head is that we:
 - Have instances make a windowed headless context.
 - We delay window making till swapchain creation.
 - When recreating a swapchain we don't recreate the window (comment in code for why)
 - When creating a swapchain the first time we make an vao, some shaders, a vbo with stuff for rendering a quad which we share between all instances.
- When creating a swapchain, we open them a new share + window + context + image. This is stuff we can't share
- When recreating  a swapchain, we only recreate the image.
- When the surface's compatibilities are queried for and the window has been already made, we give it the window's compatibilities (cause we can't change them). If it hasn't we just give them everything because it doesn't matter since we ignore it anyways.
- When doing rendering stuff, we render to the image belonging to the swapchain they wan't to render to.
- When presenting, we switch to that swapchain's context, bind the texture & then render a quad (with that texture) to that window's framebuffer.

examples/quad/main.rs
```diff
@@ -62,33 +66,24 @@ const COLOR_RANGE: i::SubresourceRange = i::SubresourceRange {
...
+            width: DIMS.width as _,
+            height: DIMS.height as _,
+        }, 1.0))
         .with_title("quad".to_string());
     // instantiate backend
-    #[cfg(not(feature = "gl"))]
     let (window, _instance, mut adapters, mut surface) = {
-        let window = wb.build(&events_loop).unwrap();
-        let instance = back::Instance::create("gfx-rs quad", 1);
+        let instance = back::Instance::create("gfx-rs quad", 1, Arc::clone(&events_loop));
```
kvark: how would this map to gfx-portability? We don't have the event loop when we are creating an instance there.

src/backend/gl/src/device.rs
```diff
@@ -424,6 +428,74 @@ impl Device {
             }
         }
     }
+
+    pub fn create_shader_program<'a>(
```
kvark: are these changes a part of the windowing PR? or just refactoring accidentally got mixed in?

src/backend/gl/src/lib.rs
```diff
@@ -37,15 +39,91 @@ mod state;
...
 
 #[cfg(feature = "glutin")]
-pub use window::glutin::{config_context, Headless, Surface, Swapchain};
+pub use window::glutin::{Surface, Swapchain, Window};
+
+#[cfg(feature = "glutin")]
+pub struct Instance {
+    el: Arc<Mutex<glutin::EventsLoop>>,
+    context: glutin::Context,
+    pub(crate) quadvbo: Arc<Mutex<Option<u32>>>,
```
kvark: nit: you can insert `_` between words

src/backend/gl/src/lib.rs
```diff
@@ -37,15 +39,91 @@ mod state;
...
+#[cfg(feature = "glutin")]
+impl Instance {
+    pub fn create(_name: &str, _version: u32, el: Arc<Mutex<glutin::EventsLoop>>) -> Arc<Self> {
+        let cb = glutin::ContextBuilder::new();
+        let context = glutin::Context::new(cb, &el.lock().unwrap(), true).unwrap();
+
+        Arc::new(Instance {
+            el,
+            context,
+            quadvbo: Arc::new(Mutex::new(None)),
```
kvark: why aren't we initializing those here?

src/backend/gl/src/lib.rs
```diff
@@ -37,15 +39,91 @@ mod state;
...
+    pub fn create_window(&self, wb: glutin::WindowBuilder) -> Arc<Mutex<Window>> {
+        Window::new(wb)
+    }
+}
+
+#[cfg(feature = "glutin")]
+impl Drop for Instance {
+    fn drop(&mut self) {
+        unsafe {
+            let gl = gl::Gl::load_with(|s| self.context.get_proc_address(s) as *const _);
```
kvark: why are we loading GL pointers here?

examples/quad/main.rs
ZeGentzy: I have no clue, I'd have to look into it.

src/backend/gl/src/device.rs
ZeGentzy: I wanted to avoid re implementing create_shader_programs when making the shader for the swapchain quad stuff.

src/backend/gl/src/lib.rs
ZeGentzy: We haven't loaded the gl pointers yet nor done the stuff which is done when opening the device. I'm not actually sure what the opening the device stuff is done for so I avoided changing that code and just made the stuff get initialized when making a swap chain the first time.

src/backend/gl/src/lib.rs
ZeGentzy: Because the instance doesn't have a copy of the share or the gl ptrs, so it's easier (and still correct) to just load them in again.

src/backend/gl/src/lib.rs
ZeGentzy: I guess we could make the physical device & open the device one in the instance then just return those. I think that's what I'll change it to do, and then we can avoid reloading the gl ptrs in drop.

kvark:
@ZeGentzy what do you think about my earlier plan to go with the "pure" GL context (for the Instance) and sharing for all the others (created per window/swapchain)? Would it still be compatible with your approach if we get https://github.com/tomaka/glutin/pull/942 merged eventually?

ZeGentzy:
I don't see why it wouldn't work, instead of sharing everything with the instance's context, we share them with each other. Then we need to make current the context for the window we are using. 

Which method is preferable is up to debate, and I personally don't have an opinion on this.

ZeGentzy:
We now all the shared stuff in the instance. We now automatically change contexts only when necessary.
Tested against tomaka/winit@c873c2d and ZeGentzy/glutin@cdad4a9 FYI

ZeGentzy:
We just blit fbo's now as discussed on gitter with @kvark .
Tested against tomaka/winit@c873c2d and ZeGentzy/glutin@cdad4a9 FYI

PR #2137, opened 2018-06-11
mtl: Fix depth stencil issues

### LINKED ISSUE #2134: mtl: Investigate failing CTS tests for dynamic_states

grovesNL:
Fixes #2134

PR checklist:

Probably easier to read these commits one at a time because they all fix separate issues:

- Always set load/store actions for attachments which aren't being cleared. We need this to preserve values in shared textures, i.e. between two separate depth and stencil attachments with the same backing texture.
- During the changes in #2129, we stopped mapping load/store actions during framebuffer creation. As a result we always used "don't care" which led to unexpected images at the beginning of render passes. For example, if image clears occurred prior to beginning the render pass, we had no guarantee that the image was still cleared at the beginning of the render pass.
- The stencil write mask updates passed the wrong argument, so they were misinterpreted as read mask updates.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -2158,8 +2158,12 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         };
         let mut num_colors = 0;
 
-        for (rat, clear_value) in render_pass.attachments.iter().zip(clear_values) {
-            let value = *clear_value.borrow();
+        let clear_values_iter = clear_values
+            .into_iter()
+            .map(|c| Some(*c.borrow()))
+            .chain(iter::repeat(None));
+
+        for (rat, clear_value) in render_pass.attachments.iter().zip(clear_values_iter) {
             let (aspects, channel) = match rat.format {
                 Some(format) => (format.surface_desc().aspects, Channel::from(format.base_format().1)),
                 None => continue,
@@ -2171,7 +2175,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                     .unwrap();
                 if set_operations(color_desc, rat.ops) == AttachmentLoadOp::Clear {
                     let mtl_color = channel
-                        .interpret(unsafe { value.color });
+                        .interpret(unsafe { clear_value.unwrap().color });
                     color_desc.set_clear_color(mtl_color);
                 }
                 num_colors += 1;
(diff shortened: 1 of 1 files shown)


kvark:
> As a result we always used "don't care" which led to unexpected images at the beginning of render passes.

My assumption was that the RP descriptor in the framebuffer is not used anywhere but for `begin_render_pass`, where we are already setting all the load ops. Could you explain a bit more the use case that is broken?

grovesNL:
During `begin_render_pass` [`clear_values` is empty](https://github.com/gfx-rs/gfx/blob/master/src/backend/metal/src/command.rs#L2147), so we don't map the ops correctly. [`clear_values` doesn't seem to be required to be the same length as attachments](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkRenderPassBeginInfo.html) if `clear` isn't used as as the load op.

Maybe there is a way to simplify this further though (by changing this `.zip)` so we don't repeat the mapping?

kvark:
@grovesNL it says:
> The array is indexed by attachment number.

So I assumed there's got to be the same number of clear values. More precisely, there is at least as many provided as the highest attachment number that is cleared, + 1. So how about we do `clear_values.chain(repeat(_whatever))` for that `zip` iterator?

grovesNL:
Sure, seems reasonable to me

grovesNL:
@kvark or @gfx-rs/metallists could somebody take a look at the updated commit? We extend the iterator by `repeat(None)` now, so my earlier fix to map load/store ops during framebuffer creation isn't needed anymore.

kvark:
Bors delegate+
Please proceed after somebody reviews this.

src/backend/metal/src/command.rs
```diff
@@ -2142,8 +2158,12 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         };
         let mut num_colors = 0;
 
-        for (rat, clear_value) in render_pass.attachments.iter().zip(clear_values) {
-            let value = *clear_value.borrow();
+        let clear_values_iter = clear_values
+            .into_iter()
+            .map(|c| Some(*c.borrow()))
+            .chain(iter::repeat(None));
```
kvark: I don't think the `Option` really helps here. The extra elements that we add are not any worse than the ones provided but not used, there is no need to try taking extra care of them. Instead, could just go with a default clear value or something.

grovesNL: Fair enough, I was trying to avoid creating extra clear values but maybe it's not a big deal.

PR #2138, opened 2018-06-11
CAMetalLayer presentation

kvark:
Fixes/rewrites the presentation mechanics in Metal, improves the API side elsewhere.
Old swapchain was represented by `CALayer` with custom created `IOSurface` frames that we attempted to present. This is nowhere near the recommended presentation path, which is implemented here. Honestly, I don't think the old path ever worked properly:
  - our example (`quad`) doesn't do any animation, and waits after each frame
  - LunarG samples weren't really tested on MacOS
  - gfx_ocean (as well as Dota) showed presentation lags/artifacts IIRC

The new logic creates a `CAMetalLayer`, set's up its format and parameters, and acquires drawables with `nextDrawable`. There is one but important complication with this method: frame images are temporary. Thus, while we can get `MTLTexture` from a drawable, it's not guaranteed to be the same one every time (if I understand correctly - please double-check). Therefore, I had to rewrite all texture handling logic to resolve the actual `MTLTexture` only when live encoding a command buffer. Until then, the regular textures are kept unchanged, but framebuffer ones are stored as a pointer to swapchain plus the frame index. I'd really like us *not* to do this, so please tell me if all of that was unnecessary :)

PR checklist:

For an icing on the cake, Dota2 now plays well:
dota-gfx-metal2

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -2,7 +2,7 @@ use {AutoreleasePool, Backend, Shared, validate_line_width};
 use {conversions as conv, native, soft, window};
 use internal::{BlitVertex, Channel, ClearKey, ClearVertex};
 
-use std::borrow::{self, Borrow};
+use std::borrow::Borrow;
 use std::cell::RefCell;
 use std::collections::HashMap;
 use std::ops::{Deref, Range};
@@ -11,7 +11,7 @@ use std::{iter, mem};
 use std::slice;
 
 use hal::{buffer, command as com, error, memory, pool, pso};
-use hal::{DrawCount, VertexCount, VertexOffset, InstanceCount, IndexCount, WorkGroupCount};
+use hal::{DrawCount, FrameImage, VertexCount, VertexOffset, InstanceCount, IndexCount, WorkGroupCount};
 use hal::format::{Aspects, Format, FormatDesc};
 use hal::image::{Extent, Filter, Layout, SubresourceRange};
 use hal::pass::{AttachmentLoadOp, AttachmentOps};
@@ -584,7 +584,6 @@ pub struct CommandBufferInner {
     sink: Option<CommandSink>,
     retained_buffers: Vec<metal::Buffer>,
     retained_textures: Vec<metal::Texture>,
-    present_frames: Vec<native::Frame>,
 }
 
 impl Drop for CommandBufferInner {
@@ -606,7 +605,6 @@ impl CommandBufferInner {
         }
         self.retained_buffers.clear();
         self.retained_textures.clear();
-        self.present_frames.clear();
     }
 
     fn sink(&mut self) -> &mut CommandSink {
@@ -971,14 +969,12 @@ unsafe impl Sync for CommandBuffer {}
 
 pub struct CommandQueue {
     shared: Arc<Shared>,
-    present_frames: Vec<native::Frame>,
 }
 
 impl CommandQueue {
     pub(crate) fn new(shared: Arc<Shared>) -> Self {
         CommandQueue {
             shared,
-            present_frames: Vec::new(),
         }
     }
 }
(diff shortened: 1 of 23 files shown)


JohnColanduoni:
I don't think there's an alternative to the indirection if we want to use `CAMetalLayer`. It's worth noting that `CAMetalLayer` actually uses swapping of `IOSurface`s internally. The old method *could* be brought up to speed with all the additional handling `CAMetalLayer` does, but that would be a fair amount of work. One advantage of this route would be the ability to provide different Vulkan swap chain modes; CAMetalLayer's default mode corresponds to FIFO. However this will be mostly disappearing in the next OS X version (we can disable `displaySyncEnabled` to implement Vulkan's immediate mode), with the exception being a lack of mailbox support.

kvark:
Thanks for the info @JohnColanduoni !

> It's worth noting that CAMetalLayer actually uses swapping of IOSurfaces internally

Out of curiosity, how do you know that? From my super limited understanding, `IOSurface` is meant for interop between processes, so I'd expect it to not be as efficient as something directly aimed at presentation, like `CADrawable`.

Also, do you have an idea on why the old method didn't work correctly? We still have an option to leave it as an alternative run/compile time path.

JohnColanduoni:
Anytime!

> Out of curiosity, how do you know that? From my super limited understanding, IOSurface is meant for interop between processes, so I'd expect it to not be as efficient as something directly aimed at presentation, like CADrawable.

I've looked at the disassembly ;). `IOSurface` is used internally on OS X/iOS for general cross-API GPU-resident buffer management, even within processes. In fact, `CADrawable`s are actually backed by `IOSurface`s. Their is some overhead involved in IOSurface creation and I suspect they take up slightly more memory than an `MTLTexture`, but once that's done (once at swapchain creation) there isn't really any additional overhead.

> Also, do you have an idea on why the old method didn't work correctly? We still have an option to leave it as an alternative run/compile time path.

It was basically the minimum implementation that could get something to display on the screen without  adding the texture indirection. In particular there was no handling of timing or explicit display refresh requests. Chromium uses `IOSurface` -> `CALayer` for all of their rendering (including WebGL) so I'm sure it can be made robust, I just didn't take the time to implement that.

Now that the new version of OS X will let you turn off vsync, I don't think there's much of a reason to get the old implementation up to speed since you've already done the work for adding the indirection. That limitation was the main reason I bothered with the `IOSurface`-based implementation.

src/backend/metal/src/command.rs
```diff
@@ -996,32 +1063,22 @@ impl RawCommandQueue<Backend> for CommandQueue {
...
+            swapchain.borrow().present(index);
         }
 
-        if cfg!(feature = "auto-capture") {
-            let shared_capture_manager = CaptureManager::shared();
-            if let Some(default_capture_scope) = shared_capture_manager.default_capture_scope() {
-                default_capture_scope.end_scope();
-                default_capture_scope.begin_scope();
-            }
+        let shared_capture_manager = CaptureManager::shared();
```
grovesNL: Why remove the feature here?

src/backend/metal/src/command.rs
```diff
@@ -1376,12 +1433,14 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
         _stages: Range<pso::PipelineStage>,
         _dependencies: memory::Dependencies,
-        _barriers: T,
+        barriers: T,
     ) where
         T: IntoIterator,
         T::Item: Borrow<memory::Barrier<'a, Backend>>,
     {
         // TODO: MTLRenderCommandEncoder.textureBarrier on macOS?
+        for _barrier in barriers {
```
grovesNL: Why is this iteration needed?

src/backend/metal/src/lib.rs
```diff
@@ -126,12 +125,14 @@ impl Instance {
                 panic!("window does not have a valid contentView");
             }
 
-            msg_send![view, setWantsLayer: YES];
-            let render_layer: *mut Object = msg_send![Class::get("CALayer").unwrap(), new]; // Returns retained
+            let class = Class::get("CAMetalLayer").unwrap();
+            let render_layer: *mut Object = msg_send![class, new];
+            msg_send![view, setLayer: render_layer];
+            //let render_layer: *mut Object = msg_send![view, layer];
```
grovesNL: Let's remove these comments

src/backend/metal/src/command.rs
kvark: Because we want to stop the capture here unconditionally. The feature allows us to capture workloads that don't involve presenting a frame, e.g. compute/transfers/tests. Here though in most cases I could think of stopping the capture is the right solution.

kvark:
Alright, comments are addressed. Thanks for the review!
Time to sail :)
bors r=grovesNL

PR #2142, opened 2018-06-12
[WIP] mtl: Disable depth/stencil when off

grovesNL:
Possibly fixes the issue noticed by @kvark?

WIP as I'm not sure whether this is actually useful or necessary – just a guess based on code inspection.

PR checklist:


kvark:
This doesn't fix the issue, although looks reasonable. Here is what I think is going on:
  - user has a pso bound with the depth test (enabled from a previous pass or something)
  - user starts a new render pass with no depth attachment
  - BAM
  - user wants to bind a different PSO with no depth test

Interestingly, the issue doesn't happen when I run from XCode :/

Anyhow, a workaround for this specific thing is - provide `make_render_pass` with `Aspects` representing what the render pass has. If there is no depth/stencil, don't try to set the current one.
My worry is that it's just one of the things that may go wrong with trying to reset the PSO state there...

grovesNL:
Interesting, let's keep looking into that solution.

(After looking into this descriptor more, these should match all the default states anyway - so we shouldn't need this PR.)

PR #2149, opened 2018-06-14
[dx11] add memory flush/invalidate & image/buffer copies

fkaa:
Main changes are adding more robust implementation of `Memory`, adding flush/invalidate and adding image/buffer copies. Implements `Memory` like the following (for `HOST_VISIBLE` memory):

```
    0.........................size
    +----------------------------+
    |          Memory            |
    +----------------------------+
    A..B  C.....D         E...F

    1 fixed-size `STAGING` buffer which gets used for reading back from
      resources.(and should be used to copy from/to on flush/invalidate):
      (0..size, ComPtr<Buffer>)

    1 `Vec<u8>` which covers the whole memory range (0..size). This is
      pointer we hand out to users. flush/invalidate moves the affected
      regions into our `STAGING` buffer to get read/uploaded.

    *N* Resources:
      (A..B, ComPtr<Resource>),
      (C..D, ComPtr<Resource>),
      (E..F, ComPtr<Resource>),
```

Implements copying between images and buffers. Image<->Image copies are mostly handled by `CopySubresourceRegion` but some formats, while same size, cant be copied with this method:

> Cannot invoke CopySubresourceRegion when the Formats of each Resource are not the same or at least castable to each other, unless one format is compressed (DXGI_FORMAT_R9G9B9E5_SHAREDEXP, or DXGI_FORMAT_BC[1,2,3,4,5]_* ) and the source format is similar to the dest according to: BC[1|4] ~= R16G16B16A16|R32G32, BC[2|3|5] ~= R32G32B32A32, R9G9B9E5_SHAREDEXP ~= R32. [ RESOURCE_MANIPULATION ERROR #281: ]

These has to be done through compute shaders instead. Image->Buffer & Buffer->Image copies also have to be done through compute shaders, as `CopySubresourceRegion` can only copy between resources of same type (Image<->Image, Buffer<->Buffer). The following formats have Buffer->Image and Image->Buffer copies implemented with these changes:

* `R8`
* `Rg8`
* `R16`
* `Rg16`
* `R32`

Gets about 400 tests passed and equal amount failed in `dEQP-VK.api.copy_and_blit.core.image_to_image.all_formats.color` (mostly because `CopySubresourceRegion` failing to copy between some formats as mentioned above)

Fixes #issue
PR checklist:


src/backend/dx11/shaders/copy.hlsl
```diff
@@ -1,15 +1,200 @@
...
+}
+
+uint GetBufferSrc(uint3 dispatch_thread_id)
+{
+    return BufferImageCopies.BufferVars.x + dispatch_thread_id.x + dispatch_thread_id.y * BufferImageCopies.BufferVars.y;
+}
+
+uint Uint4ToUint(uint4 data)
+{
+    data.x = min(data.x, 0x000000ff);
```
kvark: nit: could probably vectorize this better, e.g.
```hlsl
return dot(min(data, 0xFF), 1 << uint4(0, 8, 16, 24));
```

src/backend/dx11/shaders/copy.hlsl
```diff
@@ -1,15 +1,200 @@
...
+                  (data.y << 8)  |
+                  (data.z << 16) |
+                  (data.w << 24));
+
+    return output;
+}
+
+uint4 UintToUint4(uint data)
+{
+    return uint4((data & 0xff000000) >> 24, (data & 0xff0000) >> 16, (data & 0xff00) >> 8, data & 0xff);
```
kvark: maybe even simpler:
```hlsl
return ((data >> uint4(24, 16, 8, 0)) & 0xFF;

File: src/backend/dx11/shaders/copy.hlsl [hunk 3]
```diff
@@ -1,15 +1,200 @@
...
+
+    return output;
+}
+
+// Buffers are always R32-aligned
+StructuredBuffer<uint> BufferCopySrc : register(t0);
+RWBuffer<uint>         BufferCopyDst: register(u0);
+
+// R32
+Texture2D<uint>    ImageCopySrcR32 : register(t0);
```
kvark: we should totally share these definitions between different bit sizes (R16/R32/R8/etc)

src/backend/dx11/src/conv.rs
```diff
@@ -24,62 +24,84 @@ pub fn map_index_type(ty: IndexType) -> DXGI_FORMAT {
     }
 }
 
-pub fn typeless_format(format: DXGI_FORMAT) -> Option<DXGI_FORMAT> {
+pub fn typeless_format(format: DXGI_FORMAT) -> Option<(DXGI_FORMAT, DXGI_FORMAT)> {
```
kvark: what's the return value semantics now?

src/backend/dx11/src/device.rs
```diff
@@ -1069,7 +1089,14 @@ impl hal::Device<Backend> for Device {
         I::Item: Borrow<(&'a Memory, R)>,
         R: RangeArg<u64>,
     {
-        // TODO: flush?
+
+        // go through every range we wrote to
+        for range in ranges.into_iter() {
+            let &(memory, ref range) = range.borrow();
+            let range = *range.start().unwrap_or(&0)..*range.end().unwrap_or(&memory.size);
```
kvark: we need a helper `memory.resolve(range)` function for ranges, like we do in Metal.

src/backend/dx11/src/internal.rs
```diff
@@ -10,36 +11,71 @@ use std::{mem, ptr};
...
-                    bytecode.GetBufferPointer(),
-                    bytecode.GetBufferSize(),
-                    ptr::null_mut(),
-                    &mut shader as *mut *mut _ as *mut *mut _
-                )
-            };
-            assert_eq!(true, winerror::SUCCEEDED(hr));
+fn compile(device: ComPtr<d3d11::ID3D11Device>, entrypoint: &str) -> ComPtr<d3d11::ID3D11ComputeShader> {
+    let shader_src = include_bytes!("../shaders/copy.hlsl");
+    let bytecode = unsafe { ComPtr::from_raw(shader::compile_hlsl_shader(Stage::Compute, spirv_cross::hlsl::ShaderModel::V5_0, entrypoint, shader_src).unwrap()) };
```
kvark: nit: let's split into multiple lines

src/backend/dx11/src/lib.rs
```diff
@@ -1363,7 +1530,7 @@ pub struct DescriptorSet {
     #[derivative(Debug="ignore")]
     srv_handles: RefCell<Vec<(u32, ComPtr<d3d11::ID3D11ShaderResourceView>)>>,
     #[derivative(Debug="ignore")]
-    cbv_handles: RefCell<Vec<(u32, ComPtr<d3d11::ID3D11Buffer>)>>,
+    cbv_handles: RefCell<Vec<(u32, *mut d3d11::ID3D11Buffer)>>,
```
kvark: why the raw pointer here?

src/backend/dx11/shaders/copy.hlsl
```diff
@@ -1,15 +1,200 @@
...
+    BufferCopyDst[dst_idx] = Uint2ToUint(ImageCopySrcR16G16[src_idx].yx);
+}
+
+// R16
+Texture2D<uint>   ImageCopySrcR16 : register(t0);
+RWTexture2D<uint> ImageCopyDstR16 : register(u0);
+
+[numthreads(1, 1, 1)]
+void cs_copy_buffer_image2d_r16(uint3 dispatch_thread_id : SV_DispatchThreadID) {
+    uint src_idx = BufferImageCopies.BufferVars.x + dispatch_thread_id.x + dispatch_thread_id.y * BufferImageCopies.BufferVars.y / 2;
```
kvark: why isn't `GetBufferSrc` used here?

src/backend/dx11/shaders/copy.hlsl
```diff
@@ -1,15 +1,200 @@
...
+
+    uint2 data = UintToUint2(BufferCopySrc[src_idx]);
+
+    ImageCopyDstR16[GetImageDst(uint3(2, 1, 0) * dispatch_thread_id + uint3(0, 0, 0))] = data.y;
```
kvark: not sure I understand this. The image is `R16`, why do we need to write into 2 different texels?

fkaa: `R32_UINT` is the only format which has guaranteed UAV typed load support[0], so I was aiming for 1 buffer (source) access per thread, so in the case of `R16` one thread would have to do 2 writes. The dispatch call is also scaled depending on the format.

0: https://msdn.microsoft.com/en-us/library/windows/desktop/mt427455(v=vs.85).aspx

src/backend/dx11/src/lib.rs
fkaa: It was to avoid having a `Drop` impl for the `Buffer` struct, since portability [creates a zeroed struct](https://github.com/gfx-rs/portability/blob/master/libportability-gfx/src/impls.rs#L1145-L1147) which gets dropped (and tries to free null ComPtr)

kvark: looks like a bug in `portability` than that we need to fix

src/backend/dx11/shaders/copy.hlsl
fkaa: Beautiful! 😄

fkaa:
Updated with some changes, most of the other comments are addressed in the blitting stuff I have in another branch

PR #2150, opened 2018-06-15
Replace hash maps with fast hash maps

### LINKED ISSUE #2148: Switch to a faster hasher for the maps
Ideally, we shouldn't be using hash maps if arrays do. But if we have to, let's avoid the default hasher and instead use a faster one. Here is some code from WebRender for inspiration:
```rust
use fxhash::FxHasher;
pub type FastHashMap<K, V> = HashMap<K, V, BuildHasherDefault<FxHasher>>;
pub type FastHashSet<K> = HashSet<K, BuildHasherDefault<FxHasher>>;
```

grovesNL:
Fixes #2148
PR checklist:
- [X] tested examples with the following backends: Metal/GL/DX12/Vulkan

Changes:
- Replaced `HashMap` with `FastHashMap` type (uses `FxHash`)
- Changed `Queues` to use a `Vec` instead of `HashMap`. It doesn't seem like a `HashMap` is necessary here, and the family ID is already present within the queue group, if I understand correctly. But this seems trivial overall because this shouldn't be a hot path ever.
- mtl: Changed bindings to use a `Vec` (holey, filled with potentially a lot of `None`) indexed by binding. Not sure if this is the best approach but it should provide fast iteration and lookup by binding

Overall this seems to give a pretty good improvement in Dota 2 as far as I can tell from profiling. I don't have precise benchmarks.


fintelia:
Using [`VecMap<T>`](https://crates.io/crates/vec_map) might be preferable to using `Vec<Option<T>>`  directly. I'm not sure whether or not the slightly nicer interface would justify an additional dependency though.

src/backend/metal/src/device.rs
```diff
@@ -6,17 +6,19 @@ use {conversions as conv, command, native as n};
 use native;
 
 use std::borrow::Borrow;
-use std::collections::hash_map::{Entry, HashMap};
+use std::collections::HashMap;
```
kvark: why this change? it's the same as the old code but more verbose

src/hal/src/lib.rs
```diff
@@ -55,6 +56,9 @@ pub mod queue;
 pub mod range;
 pub mod window;
 
+#[doc(hidden)]
+pub mod auxil;
```
kvark: actually, it appears to me that `backend` has a very similar goal. Should we move the fast hashmap into `backend`?

src/backend/metal/src/lib.rs
```diff
@@ -12,6 +12,7 @@ extern crate core_graphics;
 extern crate block;
 extern crate smallvec;
 extern crate spirv_cross;
+extern crate fxhash;
```
kvark: why does Metal need this explicitly? I'd assume it to just use the types exported from `hal`

grovesNL: You're right, just left it there accidentally

src/backend/metal/src/device.rs
grovesNL: Yeah I need to combine these again. Originally `HashMap` was removed entirely then I realized we still need it for the spirv_cross overrides

grovesNL:
CI failure seems unrelated, I'll rebase now

kvark:
thanks!
bors r+

PR #2152, opened 2018-06-15
[mtl] Filter the render pass states

kvark:
Fixes a Metal validation error on us setting up the depth/stencil state upon entering a render pass that has no depth/stencil.
PR checklist:


kvark:
I'd assume the default to be fast, but we should double check with Molten.
Bors r=grovesNL

PR #2153, opened 2018-06-15
Present modes

kvark:
Happens to include #2152
PR checklist:


kvark:
Bors r=msiglreith

PR #2154, opened 2018-06-16
hal: Improve buffer documentation and cleanup error handling

msiglreith:
Fixes #issue
PR checklist:

### DIFF
--- a/src/hal/src/buffer.rs
+++ b/src/hal/src/buffer.rs
@@ -1,72 +1,58 @@
-//! Memory buffers
+//! Memory buffers.
+//!
+//! # Buffer
+//!
+//! Buffers interpret memory slices as linear continguous data array.
+//! They can be used as shader resources, vertex buffers, index buffers or for
+//! specifying the action commands for indirect exection.
 
-use std::error::Error;
-use std::fmt;
-
-use {IndexType, Backend};
+use {format, IndexType, Backend};
 
 
 /// An offset inside a buffer, in bytes.
 pub type Offset = u64;
 
+/// Buffer state.
+pub type State = Access;
+
 /// Error creating a buffer.
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Fail, Debug, Clone, PartialEq, Eq)]
 pub enum CreationError {
-    /// Required `Usage` is not supported.
-    Usage(Usage),
-    /// Some other problem.
-    Other,
+    /// Memory allocation on the host side failed.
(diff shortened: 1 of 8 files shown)


src/hal/src/buffer.rs
```diff
@@ -132,11 +124,10 @@ bitflags!(
...
 
-/// Buffer state
-pub type State = Access;
-
-/// Index buffer view for `bind_index_buffer`, slightly
-/// analogous to an index table into an array.
+/// Index buffer view for `bind_index_buffer`.
+///
+/// Defines a buffer slice used for acquiring the indicies on draw commands.
+/// Indices are used to lookup the vertex indices in the vertex buffers.
```
kvark: "to lookup vertices" perhaps?

src/hal/src/buffer.rs
```diff
@@ -1,72 +1,58 @@
...
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        let description = self.description();
-        match *self {
-            ViewError::Usage(usage) => write!(f, "{}: {:?}", description, usage),
-            _ => write!(f, "{}", description)
-        }
-    }
+/// Error creating a buffer.
+#[derive(Fail, Debug, Clone, PartialEq, Eq)]
+pub enum BufferCreationError {
```
kvark: what's the motivation behind having the `Buffer*` prefix for those errors? If we are exposing `buffer` as a module, I find it elegant to keep it `buffer::CreationError`. If we are using the fully qualified names, there is no need to expose the modules.

msiglreith: I'm not sure which direction we should go for the errors.
Currently having a `error` module containing `DeviceCreationError` and a few more. On the other side we could also move these errors into the different modules as done here with possible name shortening.

kvark: I'm (non-strongly) in favor of having short error names in respective modules (as opposed to putting them into a dedicated `error` module)

msiglreith: yeah I think this will be more clear. changed.

msiglreith:
bors r=kvark

kvark:
bors retry

msiglreith:
bors retry

msiglreith:
bors r=kvark

kvark:
bors retry

kvark:
bors retry

kvark:
Bors retry

kvark:
Bors r+

PR #2160, opened 2018-06-20
Numbers the examples.

ZeGentzy:
Numbers the examples so that when we add more examples, people can have a general order to view them in.

PR checklist:
- [x] tested examples with the following backends: vulkan and opengl


kvark:
Thank you for the PR!
I don't think we should number the examples. In pre-ll we had quite a few examples without any numbering, and they worked well for feature discovery. There doesn't have to be a specific order, outside maybe a tutorial, which we don't have.

ZeGentzy:
In that case, closing.

Issue #2161, opened 2018-06-20
[meta] Dota performance issues on Metal


kvark:
Analyzing the Instruments profiles reveals quite a few things we could do better.

### Native API calls
  - [x] Inefficient ObjectiveC selector lookups - #2145
  - [x] ~~Too many ObjectiveC messages. Roughly, our profile shows 5-7% dedicated to `objc_msgSend` versus 0.3-0.5% in MoltenVK. The difference in retain/release calls is of a similar scale.~~
  - [x] Too many `retain`/`release` calls - #2175
  - [ ] Retains & releases could be faster if not done via messaging  - https://github.com/gfx-rs/metal-rs/issues/58 .
  - [x] ~~Attribute access goes through messages - #2167~~
  - [x] `Class::get` calls are super slow - https://github.com/SSheldon/rust-objc/issues/65 . We spend roughly 3% of our library's time doing those calls.
  - [x] nsAutoreleasePool could be created faster - #2267, https://github.com/SSheldon/rust-objc/pull/67

### Metal backend issues
  - [x] Too much heap (re)allocation - #2185. We should avoid any allocations at run-time by re-using the storage and/or using the iterators more aggressively.
  - [x] Too many command buffers - #2180. The instrumented memory profile doesn't show command buffers being re-used. I suppose the queue just creates a new one every time until it reaches the capacity, and then starts trying to re-use the completed ones. This is undesired, and we may attempt to address it (temporarily) by playing with the capacity limit.
  - [x] Dynamic creation of a render pass - #2178. We create one every time a render pass is started by copying one from the framebuffer and filling out all the clear values/operations. Apparently, this is super heavy, and we should probably avoid creating any Metal objects at normal run time.
  - [x] Dynamic creation of depth-stencil states - #2195.
  - [x] Binding descriptor sets is visibly slow - #2183. We should avoid doing work for repeated bindings, and need to explore ways to share heap-allocated data between sets of the same group.
  - [x] Non-existent semaphore frame synchronization  - #2143
  - [x] Too much locking going on - #2229
  - [x] Too many command buffer callbacks - #2224
  - [x] Copying the render pass descriptor (which we do at each start of an RP) takes 7.4% of the time in our library
  - [ ] Binding descriptor sets still involves a few hot loops that could be faster. In particular, we set resources in batches, and we provide a closure, which checks for the current pre-render status. All of that can be simplified if we are outside of pass.

### Portability issues
[...]

PR #2164, opened 2018-06-21
[mtl] Borrowed commands

kvark:
PR checklist:

r? @gfx-rs/metallists 

This PR attempts to have lightweight software commands that don't take any heap space or own ObjC objects. In most cases, where a command list is live-recorded and executed once, this should reduce the amount of work we do per command, which is especially important if those commands are thrown away (e.g. because we are not inside a render pass).

My expectation would be to see an improvement in #2161 due to us doing less work. The actual results are somewhat shocking: with v-sync enabled I'm getting the same 59-60 fps as usual. With v-sync OFF, I'm getting between 25 and 50 fps now (which is lower than the previous 50-70). Not sure what's going on, the instrumental profile doesn't give a clue. Please check out the code.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1,6 +1,6 @@
-use {AutoreleasePool, Backend, Shared, validate_line_width};
+use {AutoreleasePool, Backend, PrivateDisabilities, Shared, validate_line_width};
 use {conversions as conv, native, soft, window};
-use internal::{BlitVertex, Channel, ClearKey, ClearVertex};
+use internal::{BlitVertex, Channel, ClearKey, ClearVertex, ServicePipes};
 
 use std::borrow::Borrow;
 use std::cell::RefCell;
@@ -13,7 +13,7 @@ use hal::{buffer, command as com, error, memory, pool, pso};
 use hal::{DrawCount, FrameImage, VertexCount, VertexOffset, InstanceCount, IndexCount, WorkGroupCount};
 use hal::backend::FastHashMap;
 use hal::format::{Aspects, Format, FormatDesc};
-use hal::image::{Extent, Filter, Layout, SubresourceRange};
+use hal::image::{Extent, Filter, Layout, Level, SubresourceRange};
 use hal::pass::{AttachmentLoadOp, AttachmentOps};
 use hal::query::{Query, QueryControl, QueryId};
 use hal::queue::{RawCommandQueue, RawSubmission};
@@ -26,7 +26,8 @@ use block::{ConcreteBlock};
 use smallvec::SmallVec;
 
 
-const WORD_ALIGNMENT: u64 = 4;
+const WORD_SIZE: usize = 4;
+const WORD_ALIGNMENT: u64 = WORD_SIZE as _;
 /// Enable an optimization to have multi-layered render passed
 /// with clear operations set up to implement our `clear_image`
 /// Note: currently doesn't work, needs a repro case for Apple
@@ -103,6 +104,13 @@ pub struct CommandBuffer {
     inner: CommandBufferInnerPtr,
     shared: Arc<Shared>,
     state: State,
+    temp: Temp,
+}
+
+#[derive(Clone)]
+struct Temp {
+    clear_vertices: Vec<ClearVertex>,
+    blit_vertices: FastHashMap<(Aspects, Level), Vec<BlitVertex>>,
 }
 
 #[derive(Clone)]
@@ -120,7 +128,7 @@ struct State {
     resources_vs: StageResources,
     resources_fs: StageResources,
     resources_cs: StageResources,
-    index_buffer: Option<IndexBuffer>,
+    index_buffer: Option<IndexBuffer<metal::Buffer>>,
     rasterizer_state: Option<native::RasterizerState>,
     pipeline_depth_stencil: Option<(pso::DepthStencilDesc, metal::DepthStencilState)>,
     dynamic_depth_stencil_desc: Option<metal::DepthStencilDescriptor>,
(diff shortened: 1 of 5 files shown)


kvark:
It might be the power management playing tricks with me. When plugged in (as opposed to 50% battery), I see stable 60-70fps. Just in case, I double-checked the version before my PR, and it doesn't appear faster. It's roughly the same (which is also kinda concerning), maybe a few frames per second less. With this in mind, removing the WIP tag, open for reviews.

src/backend/metal/src/command.rs
```diff
@@ -698,19 +964,19 @@ fn exec_render(encoder: &metal::RenderCommandEncoderRef, command: &soft::RenderC
                 _ => unimplemented!()
             }
         }
-        Cmd::BindBufferData { stage, ref bytes, index } => {
+        Cmd::BindBufferData { stage, index, words } => {
             match stage {
                 pso::Stage::Vertex =>
-                    encoder.set_vertex_bytes(index as _, bytes.len() as _, bytes.as_ptr() as _),
+                    encoder.set_vertex_bytes(index as _, words.len() as u64 * 4, words.as_ptr() as _),
```
grovesNL: nit: `4` -> `WORD_SIZE` or similar? We use this a few times

src/backend/metal/src/command.rs
```diff
@@ -1495,29 +1555,24 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                words: &value_and_length[..],
             },
             soft::ComputeCommand::Dispatch {
                 wg_size,
                 wg_count,
             },
         ];
 
-        inner.sink().begin_compute_pass(commands);
+        inner.sink().begin_compute_pass(commands.iter().cloned());
```
grovesNL: Seems like we should be able to avoid `cloned` here?

src/backend/metal/src/soft.rs
```diff
@@ -1,49 +1,75 @@
...
         index: usize,
-        buffer: Option<metal::Buffer>,
+        buffer: Option<R::Buffer>,
         offset: hal::buffer::Offset,
     },
     BindBufferData {
         stage: hal::pso::Stage,
         index: usize,
-        bytes: Vec<u8>,
+        words: R::Data,
```
grovesNL: In general I don't think 4-byte alignment is necessary in most places for iOS and tvOS, so we may want to reconsider eventually, or set a `WORD_SIZE` based on the OS

src/backend/metal/src/command.rs
```diff
@@ -2764,21 +2880,21 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
                 compute_commands.push(soft::ComputeCommand::BindBufferData {
                     index: 2,
-                    bytes: unsafe {
+                    words: unsafe {
                         slice::from_raw_parts(
-                            &(r.size as u32) as *const u32 as _,
-                            mem::size_of::<u32>()
-                        ).to_owned()
+                            &(r.size as u32) as *const u32,
+                            1,
```
grovesNL: nit: `mem::size_of::<u32>() / WORD_SIZE` if we ever change this (as mentioned in other comment)

src/backend/metal/src/command.rs
```diff
@@ -3015,28 +3137,27 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
-            } else {
-                None
-            };
-            let commands = com_vs.into_iter().chain(com_ps);
-
-            self.inner
-                .borrow_mut()
-                .sink()
-                .pre_render_commands(commands);
+            if stages.contains(pso::ShaderStageFlags::VERTEX) {
```
grovesNL: The old approach avoids two calls to `pre_render_commands` here – not sure how the `Option` compares to this trade-off

src/backend/metal/src/internal.rs
```diff
@@ -304,6 +312,43 @@ impl ServicePipes {
 
         device.new_render_pipeline_state(&pipeline).unwrap()
     }
+}
+
+
+//#[derive(Clone)]
```
grovesNL: nit: comment and extra newline

src/backend/metal/src/native.rs
```diff
@@ -171,29 +171,53 @@ pub enum ImageRoot {
...
+        }
+    }
+}
+
+pub enum ImageGuard<'a> {
     Texture(&'a metal::TextureRef),
     Frame {
         swapchain: RwLockReadGuard<'a, SwapchainInner>,
-        index: hal::FrameImage,
+        index: hal::FrameImage
```
grovesNL: nit: comma removed

src/backend/metal/src/command.rs
kvark: I had to go through quite a few hoops with the borrow checker...

src/backend/metal/src/command.rs
```diff
@@ -140,141 +149,398 @@ impl State {
...
+    fn set_viewport<'a>(
+        &mut self, vp: &'a pso::Viewport, disabilities: &PrivateDisabilities
+    ) -> soft::RenderCommand<&'a soft::Own> {
+        let viewport = MTLViewport {
+            originX: vp.rect.x as _,
+            originY: vp.rect.y as _,
+            width: vp.rect.w as _,
+            height: vp.rect.h as _,
+            znear: vp.depth.start as _,
+            zfar: if disabilities.broken_viewport_near_depth {
```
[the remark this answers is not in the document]
kvark: you can check it locally, I suppose, since this flag is only enabled on Intel, and you got AMD IIRC

src/backend/metal/src/command.rs
kvark: Actually, no. The iterator of a slice is over `&Command` while we expect commands by move there. Cloning is lightweight, since this are commands over references.

kvark:
Thank you for a detailed review! :heart: 
Concerns are addressed, let's proceed.
bors r=grovesNL

src/backend/metal/src/command.rs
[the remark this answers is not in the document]
grovesNL: I'm on Intel HD Graphics 6000 1536 MB

kvark:
Bors retry

kvark:
bors retry

kvark:
bors r=grovesNL

kvark:
Bors r=grovesNL

kvark:
Bors retry

kvark:
bors retry

kvark:
bors r=grovesNL

PR #2169, opened 2018-06-21
Persistent Metal swapchain

### LINKED ISSUE #2143: Frame synchronization in Metal
After https://github.com/gfx-rs/gfx/pull/2138 got in, we got a problem with exclusive full-screen mode of Dota, where a semaphore is provided for the frame, and there is nothing instantly available. So `nextDrawable == nil`, and we don't get a `CADrawable`, and we don't get the associated `MTLTexture`... which is needed for immediate command buffer recording.

One possible solution would be to switch to deferred recording once we encounter a texture that will be available later (from an acquired frame). We'd only be able to record the rest when we get that texture.

### LINKED ISSUE #2168: Present via command buffers on Metal
Currently, presenting on a queue simply results in `CAMetalDrawable::present` message being sent. Apple documentation advises against this approach because the presentation can only wait for operations (it depends on) that have been scheduled. So if we committed a command buffer prior to presentation that hasn't reached the "scheduling" phase, then there is going to be a corruption or some data race for presentation that we definitely do not what to see.

The solution would be to have a temporary command buffer created at `submit` which presents all the things and is getting committed. We can make it a (default) option, but I'm leaning towards just having this as the only true path.

kvark:
~~Includes #2164~~
Fixes #2168
Fixes #2143
Removes deferred image resolution (yay!), and also renames some of the outdated concepts.
We have no flickering, and proper frame sync now.

PR checklist:

### DIFF
--- a/src/backend/metal/src/window.rs
+++ b/src/backend/metal/src/window.rs
@@ -3,8 +3,8 @@ use internal::Channel;
 use native;
 use device::{Device, PhysicalDevice};
 
-use std::{fmt, ops};
-use std::sync::{Arc, Mutex, RwLock};
+use std::cell::Cell;
+use std::sync::{Arc, Mutex};
 
 use hal::{self, format, image};
 use hal::{Backbuffer, SwapchainConfig};
@@ -26,6 +26,7 @@ pub struct Surface {
     pub(crate) apply_pixel_scale: bool,
 }
 
+//TODO: double-check who needs it shared
 pub(crate) struct SurfaceInner {
     pub(crate) nsview: *mut Object,
     pub(crate) render_layer: Mutex<CAMetalLayer>,
@@ -42,44 +43,22 @@ impl Drop for SurfaceInner {
 
 #[derive(Debug)]
 struct Frame {
-    drawable: Option<CADrawable>,
+    drawable: Cell<Option<CADrawable>>,
     texture: metal::Texture,
 }
 
-pub struct SwapchainInner {
-    frames: Vec<Frame>,
-}
-
-impl ops::Index<hal::SwapImageIndex> for SwapchainInner {
-    type Output = metal::TextureRef;
-    fn index(&self, index: hal::SwapImageIndex) -> &Self::Output {
-        &self.frames[index as usize].texture
-    }
-}
-
-unsafe impl Send for SwapchainInner {}
-unsafe impl Sync for SwapchainInner {}
-
-impl fmt::Debug for SwapchainInner {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Swapchain with {} image", self.frames.len())
-    }
-}
-
-impl Drop for SwapchainInner {
+impl Drop for Frame {
     fn drop(&mut self) {
-        for mut frame in self.frames.drain(..) {
-            if let Some(drawable) = frame.drawable.take() {
-                unsafe {
-                    msg_send![drawable, release];
-                }
+        if let Some(drawable) = self.drawable.get() {
+            unsafe {
+                msg_send![drawable, release];
             }
         }
     }
 }
 
 pub struct Swapchain {
-    inner: Arc<RwLock<SwapchainInner>>,
+    frames: Vec<Frame>,
     surface: Arc<SurfaceInner>,
     _size_pixels: (u64, u64),
 }
@@ -89,12 +68,8 @@ unsafe impl Sync for Swapchain {}
 
 impl Swapchain {
     pub(crate) fn present(&self, index: hal::SwapImageIndex) {
-        let drawable = self.inner
-            .write()
-            .unwrap()
-            .frames[index as usize]
-            .drawable
-            .take()
+        let drawable = self.frames[index as usize].drawable
+            .replace(None)
             .unwrap();
         unsafe {
             msg_send![drawable, present];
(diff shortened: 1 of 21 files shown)


kvark:
I've been looking at the resize issue with this approach, and haven't got a working solution yet. It appears that `CAMetalLayer` does a cleanup routine on each `nextDrawable` call, which involves freeing any `MTLIGAccelTexture` and `IOSurface` that are not needed:
```
   0 libobjc.A.dylib -[NSObject release]
   1 QuartzCore release_image(_CAImageQueue*, unsigned long, CAImageQueueReleased*, bool)
   2 QuartzCore CAImageQueueCollect_
   3 QuartzCore -[CAMetalLayer nextDrawable]
   4 quad _$LT$$LP$$RP$$u20$as$u20$objc..message..MessageArguments$GT$::invoke::h1b3182d6d03268e8 /Users/dmalyshau/.cargo/registry/src/github.com-1ecc6299db9ec823/objc-0.2.2/src/message/mod.rs:123
   5 quad objc::message::send_unverified::h3d8bba873fb684c6 /Users/dmalyshau/.cargo/registry/src/github.com-1ecc6299db9ec823/objc-0.2.2/src/message/mod.rs:187
   6 quad objc::message::send_message::h70a5434b096fefd8 src/backend/metal/src/window.rs:197
   7 quad gfx_backend_metal::window::SurfaceInner::next_frame::h03e08cb6c11f5ce5 src/backend/metal/src/window.rs:48
   8 quad gfx_backend_metal::window::SwapchainImage::wait_until_ready::h7046cc58479c4f64 src/backend/metal/src/window.rs:170
   9 quad gfx_backend_metal::command::CommandQueue::wait::h3fa7880b032ce033 /Users/dmalyshau/Code/gfx/src/backend/metal/src/command.rs:1212
  10 quad _$LT$gfx_backend_metal..command..CommandQueue$u20$as$u20$gfx_hal..queue..RawCommandQueue$LT$gfx_backend_metal..Backend$GT$$GT$::submit_raw::h5e304cc3bfc623b7 /Users/dmalyshau/Code/gfx/src/backend/metal/src/command.rs:1228
  11 quad _$LT$gfx_hal..queue..CommandQueue$LT$B$C$$u20$C$GT$$GT$::submit::h8b44ace069ee5932 /Users/dmalyshau/Code/gfx/src/hal/src/queue/mod.rs:115
  12 quad quad::main::hd8ef08905678179a examples/quad/main.rs:504
  13 quad std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::he6e7766661d087f6 /Users/travis/build/rust-lang/rust/src/libstd/rt.rs:74
  14 quad std::rt::lang_start_internal::_$u7b$$u7b$closure$u7d$$u7d$::h119e0490912ce91a libstd/panicking.rs:59
  15 quad std::panicking::try::do_call::hd1b51a136314709e libstd/panicking.rs:306
  16 quad __rust_maybe_catch_panic libpanic_unwind/lib.rs:102
  17 quad std::rt::lang_start_internal::hb7855285a8133f7a libstd/rt.rs:285
  18 quad std::rt::lang_start::h2db61e66a5bb226a /Users/travis/build/rust-lang/rust/src/libstd/rt.rs:74
```
... except that it decides to *not* clean up those images we use persistently. I did confirm that the refcount on them (meaning `MTLIGAccelTexture`) is `1` by the point of swapchain re-creation, and it matches the old code behavior (which works). So the problem lies something else... something prevents `CAMetalLayer` to not free those surfaces.

src/backend/metal/src/window.rs
```diff
@@ -39,84 +39,144 @@ impl Drop for SurfaceInner {
...
+        // check the target frame first
+        {
+            let frame = self.frames[self.index as usize].inner.lock().unwrap();
+            assert!(!frame.available);
+            if frame.drawable.is_some() {
+                return
+            }
+        }
+        // wait for new frames to come until we meet the chosen one
+        while self.surface.next_frame(&self.frames).0 != self.index as usize {
```
grovesNL: Is there a low-cost alternative to spinning here? I don't know how long these waits commonly last, but it seems like there's potential to run faster than presents occur sometimes

src/backend/metal/src/window.rs
```diff
@@ -177,8 +240,10 @@ impl Device {
             msg_send![render_layer, setDevice: device_raw];
             msg_send![render_layer, setPixelFormat: mtl_format];
             msg_send![render_layer, setFramebufferOnly: framebuffer_only];
+            msg_send![render_layer, setMaximumDrawableCount: config.image_count as u64];
             //TODO: only set it where supported
             msg_send![render_layer, setDisplaySyncEnabled: display_sync];
+            //msg_send![render_layer, setPresentsWithTransaction: true];
```
grovesNL: nit: comment

src/backend/metal/src/window.rs
```diff
@@ -198,28 +263,41 @@ impl Device {
...
+                    let drawable: &metal::DrawableRef = msg_send![render_layer, nextDrawable];
+                    assert!(!drawable.as_ptr().is_null());
+                    let texture: &metal::TextureRef = msg_send![drawable, texture];
+                    (drawable, texture)
+                };
+                if index == 0 {
+                    // when resizing, this trick frees up the currently shown frame
+                    drawable.present();
+                }
+                //info!("\tframe[{}] = {:?}", index, texture);
```
grovesNL: nit: comment

kvark:
Thanks for the review! Comments are addressed now, let's ship this hack :)
bors r=grovesNL

PR #2174, opened 2018-06-24
metal: Hack to fix the first frame of a swapchain

scoopr:
The freeing up of the drawable now checks if there is anything to free
from the has_swapchain boolean.

The boolean is totally a hack, and might have some edge-cases where it
is wrong, but this happens to fix warnings and red-frame flashes
(present of not drawn frame) from the quad example, with the resizing
still working.

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -144,6 +144,7 @@ impl Instance {
         window::Surface {
             inner: Arc::new(self.create_from_nsview(nsview)),
             apply_pixel_scale: false,
+            has_swapchain: false
         }
     }
 
@@ -153,6 +154,7 @@ impl Instance {
         window::Surface {
             inner: Arc::new(self.create_from_nsview(window.get_nsview())),
             apply_pixel_scale: true,
+            has_swapchain: false,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/backend/metal/src/lib.rs
```diff
@@ -143,6 +143,7 @@ impl Instance {
         window::Surface {
             inner: Arc::new(self.create_from_nsview(nsview)),
             apply_pixel_scale: false,
+            has_swapchain: false
```
kvark: ok, now I realize the ugliness of the hack: in Metal we can't really create multiple swapchains, since each one will think it's a sole owner of surface's drawables. Unless.... we somehow make a `CAMetalLayer` per swapchain.

src/backend/metal/src/window.rs
```diff
@@ -274,18 +275,30 @@ impl Device {
                     let texture: &metal::TextureRef = msg_send![drawable, texture];
                     (drawable, texture)
                 };
-                if index == 0 {
+                trace!("\tframe[{}] = {:?}", index, texture);
+
+                if index == 0 && surface.has_swapchain {
```
kvark: let's condense the two arms with something like
```rust
let drawable = if index == 0 && surface.has_swapchain {
  drawable.present();
  None
} else {
   Some(drawable.to_owned())
};
```

src/backend/metal/src/window.rs
```diff
@@ -274,14 +275,20 @@ impl Device {
...
-                }
-                trace!("\tframe[{}] = {:?}", index, texture);
+                    None
+                } else {
+                    Some(drawable.to_owned())
+                };
                 Frame {
                     inner: Mutex::new(FrameInner {
-                        drawable: Some(drawable.to_owned()),
+                        drawable: drawable,
```
kvark: nit: just `drawable,`

PR #2175, opened 2018-06-24
mtl: Use pointers for temporary state

### LINKED ISSUE #2161: [meta] Dota performance issues on Metal

grovesNL:
Partially fix to #2161
PR checklist:

This replaces buffer/texture/sampler object usage with pointers instead. This allows us to bypass Objective-C's reference counting (i.e. retain/release). The idea is to create the same objects as before, but purposely leak them and acquire a pointer to them instead. When we're finished with the objects, we take back ownership and release as usual. We should be able to do this either through `drop` on the owning struct, or through relevant `destroy_*` functions.

Still heavily WIP (but works with quad/Dota 2) because there are a few places that will cause leaks or use-after-free in its current state (need to fix cases where the same image is returned for an image view, verify `drop` exists on all the correct owners, etc.)

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1,4 +1,7 @@
-use {AutoreleasePool, Backend, PrivateDisabilities, Shared, validate_line_width};
+use {
+    AutoreleasePool, Backend, PrivateDisabilities, Shared, validate_line_width,
+    BufferPtr, TexturePtr, SamplerPtr,
+};
 use {conversions as conv, native, soft, window};
 use internal::{BlitVertex, Channel, ClearKey, ClearVertex, ServicePipes};
 
@@ -19,7 +22,7 @@ use hal::query::{Query, QueryControl, QueryId};
 use hal::queue::{RawCommandQueue, RawSubmission};
 use hal::range::RangeArg;
 
-use foreign_types::ForeignType;
+use foreign_types::{ForeignType, ForeignTypeRef};
 use metal::{self, MTLViewport, MTLScissorRect, MTLPrimitiveType, MTLIndexType, MTLSize, CaptureManager};
 use cocoa::foundation::{NSUInteger, NSInteger, NSRange};
 use block::{ConcreteBlock};
@@ -128,14 +131,14 @@ struct State {
     resources_vs: StageResources,
     resources_fs: StageResources,
     resources_cs: StageResources,
-    index_buffer: Option<IndexBuffer<metal::Buffer>>,
+    index_buffer: Option<IndexBuffer<BufferPtr>>,
     rasterizer_state: Option<native::RasterizerState>,
     pipeline_depth_stencil: Option<(pso::DepthStencilDesc, metal::DepthStencilState)>,
     dynamic_depth_stencil_desc: Option<metal::DepthStencilDescriptor>,
     dynamic_depth_stencil_state: Option<metal::DepthStencilState>,
     stencil: native::StencilState<pso::StencilValue>,
     push_constants: Vec<u32>,
-    vertex_buffers: Vec<Option<(metal::Buffer, u64)>>,
+    vertex_buffers: Vec<Option<(BufferPtr, u64)>>,
     framebuffer_inner: native::FramebufferInner,
 }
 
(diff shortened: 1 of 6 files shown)


jrmuizel:
How does this approach ensure safety? It seems like it would be error prone.

grovesNL:
@jrmuizel It would be nice to rely on Objective-C to reference count these correctly, but we shouldn't really need it. The objects we expose externally have well-defined lifetimes matching the Vulkan spec. That's why we could decide to use the `destroy_*` functions from Vulkan (which no-op currently) instead of `drop`.

We only have to be really careful while handling temporary objects that we create internally. We already have to work around this issue in some other places, for example we have `retained_buffers` and `retained_textures` to hold onto our own temporary buffers/textures until the owning command buffer has been reset or dropped.

grovesNL:
I'm going to think about this some more and try to restrict pointer usage to high frequency retains/releases as @kvark mentioned on Gitter

src/backend/metal/src/command.rs
```diff
@@ -938,11 +921,12 @@ fn exec_render<'a>(encoder: &metal::RenderCommandEncoderRef, command: soft::Rend
             encoder.set_stencil_front_back_reference_value(front, back);
         }
         Cmd::BindBuffer { stage, index, buffer, offset } => {
+            let native = buffer.as_ref().map(|b| b.as_native());
```
grovesNL: @kvark without `as_ref`, the borrow checker complains because `b` is dropped in these cases

grovesNL:
Updated the commit – slightly safer now with less pointer usage for low frequency calls

src/backend/metal/src/internal.rs
```diff
@@ -71,8 +76,8 @@ impl Channel {
 
 
 pub struct SamplerStates {
-    nearest: metal::SamplerState,
-    linear: metal::SamplerState,
+    nearest: SamplerPtr,
```
kvark: I don't think we should have weak pointers here (in this module)

src/backend/metal/src/native.rs
```diff
@@ -260,7 +261,7 @@ impl hal::DescriptorPool<Backend> for DescriptorPool {
                                 sampler_offset += layout.count;
                                 slice
                                     .iter()
-                                    .map(|s| Some(s.clone()))
+                                    .map(|s| Some(SamplerPtr(s.as_ptr())))
```
kvark: isn't this just a copy? e.g. ca't we just do `cloned()` instead of `map`

grovesNL: It's still `DescriptorSetLayout::Emulated(Vec<pso::DescriptorSetLayoutBinding>, Vec<metal::SamplerState>)`. Do we want `Vec<SamplerPtr>` here instead?

grovesNL: I assume we do want `SamplerPtr`, but even then we still need to `map` to get `Some` here

kvark: oh, nvm then. I thought it's `SamplerPtr`, but it doesn't have to be

grovesNL: Let's leave `metal::SamplerState` for now I guess – we appear to rely on it being retained here (not sure if its intentional or if there is supposed to be another owner)

grovesNL:
@kvark thanks, updated to remove `SamplerPtr` changes to internal

kvark:
Bors r+

PR #2178, opened 2018-06-25
[mtl] minimize creation of render passes

kvark:
PR checklist:

The logic is changed in the following way:
  - a framebuffer still keeps a hold of a render pass descriptor, but now behind a mutex
  - starting a render pass mutates that locked descriptor in place
  - if a command buffer is deferred, only then we copy the whole descriptor out
  - careful treatment of image blits/clears is done to re-use the RP descriptor more often

Note: doesn't seem to affect Dota framerate considerably

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -774,7 +774,11 @@ impl CommandSink {
             CommandSink::Deferred { ref mut passes, ref mut is_encoding } => {
                 *is_encoding = keep_open;
                 passes.push(soft::Pass::Render {
-                    desc: descriptor.to_owned(),
+                    //Note: the original descriptor belongs to the framebuffer,
+                    // and will me mutated afterwards.
+                    desc: unsafe {
+                        msg_send![descriptor, copy]
+                    },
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/command.rs
```diff
@@ -774,7 +774,11 @@ impl CommandSink {
             CommandSink::Deferred { ref mut passes, ref mut is_encoding } => {
                 *is_encoding = keep_open;
                 passes.push(soft::Pass::Render {
-                    desc: descriptor.to_owned(),
+                    //Note: the original descriptor belongs to the framebuffer,
+                    // and will me mutated afterwards.
+                    desc: unsafe {
+                        msg_send![descriptor, copy]
```
grovesNL: Could we somehow specify whether the `copy` is actually needed to avoid this when it hasn't changed?

kvark:
The copy is needed there anyway in most cases: for copies and blits we know the descriptor is different, otherwise we wouldn't have multiple passes. For real render passes the framebuffer still needs to own one to mutate later, so retaining a strong reference here is not an option.

kvark:
bors r=grovesNL

PR #2185, opened 2018-06-28
Various Metal performance optimizations

kvark:
Helps #2161 . I'm now getting 80-85 fps on the test run.
Aside from Metal, also changes HAL to avoid heap allocation for vertex buffer binding.
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -821,11 +821,11 @@ impl CommandSink {
     ) where
         I: Iterator<Item = soft::RenderCommand<&'a soft::Own>>,
     {
+        //assert!(AutoReleasePool::is_active());
         self.stop_encoding();
 
         match *self {
             CommandSink::Immediate { ref cmd_buffer, ref mut encoder_state, .. } => {
-                let _ap = AutoreleasePool::new();
                 let encoder = cmd_buffer.new_render_command_encoder(descriptor);
                 for command in init_commands {
                     exec_render(encoder, command);
@@ -1446,7 +1446,7 @@ impl pool::RawCommandPool<Backend> for CommandPool {
                 framebuffer_inner: native::FramebufferInner {
                     extent: Extent::default(),
                     aspects: Aspects::empty(),
-                    colors: Vec::new(),
+                    colors: SmallVec::new(),
                     depth_stencil: None,
                 }
             },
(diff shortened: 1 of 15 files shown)


src/backend/metal/src/command.rs
```diff
@@ -821,11 +821,11 @@ impl CommandSink {
     ) where
         I: Iterator<Item = soft::RenderCommand<&'a soft::Own>>,
     {
+        //assert!(AutoReleasePool::is_active());
```
grovesNL: nit: remove this?

src/backend/metal/src/command.rs
```diff
@@ -1653,6 +1653,8 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         T: IntoIterator,
         T::Item: Borrow<SubresourceRange>,
     {
+        let _ap = AutoreleasePool::new();
```
grovesNL: Does this fix the leak?

src/backend/metal/src/command.rs
kvark: this is intentional - we need to make sure there is an active autorelease pool set up by the caller here, but I don't know how to code it yet

src/backend/metal/src/command.rs
kvark: possible! my motivation was to avoid 2 autorelease pools during a regular `begin_render_pass`, so now I removed the inner one and made sure all the callers set up one.

src/backend/metal/src/command.rs
```diff
@@ -2233,13 +2233,26 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+        if self.state.vertex_buffers.len() <= first_binding as usize {
+            self.state.vertex_buffers.resize(first_binding as usize + 1, None);
+        }
+        for (i, (buffer, offset)) in buffers.into_iter().enumerate() {
+            let b = buffer.borrow();
+            let buffer_ptr = BufferPtr(b.raw.as_ptr());
+            let index = first_binding as usize + i;
+            let value = Some((buffer_ptr, b.range.start + offset));
+            if index >= self.state.vertex_buffers.len() {
+                debug_assert_eq!(index, self.state.vertex_buffers.len());
```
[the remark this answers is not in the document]
kvark: it shouldn't I'm just being overly careful here (hence, it's a debug assert and not a usual one)

kvark:
CI looks green
bors r=grovesNL

PR #2188, opened 2018-06-28
[pre-ll] Update winit -> 0.16

alexheretic:
* gfx_window_glutin -> `0.25`
* gfx_window_dxgi -> `0.16`
* Use physical dimensions in window projects.
* Use `poll_events` in examples instead of `run_forever` which can only handle one event per frame.
* Use WINIT_HIDPI_FACTOR override in performance example to avoid non-1 dpi x11 issues as this example can't handle resizing.

PR checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,13 +25,13 @@ name = "gfx_app"
 [dependencies]
 log = "0.4"
 env_logger = "0.5"
-glutin = "0.16"
-winit = "0.15"
+glutin = "0.17"
+winit = "0.16"
 gfx_core = { path = "src/core", version = "0.8" }
 gfx = { path = "src/render", version = "0.17" }
 gfx_macros = { path = "src/macros", version = "0.2" }
 gfx_device_gl = { path = "src/backend/gl", version = "0.16" }
-gfx_window_glutin = { path = "src/window/glutin", version = "0.24" }
+gfx_window_glutin = { path = "src/window/glutin", version = "0.25" }
 gfx_window_glfw = { path = "src/window/glfw", version = "0.17", optional = true }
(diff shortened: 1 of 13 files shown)


francesca64:
> Use WINIT_HIDPI_FACTOR override in performance example to avoid non-1 dpi x11 issues as this example can't handle resizing.

Note that non-1.0 scale factors will also trigger a resize on Windows.

alexheretic:
> Note that non-1.0 scale factors will also trigger a resize on Windows.

Yep I figured this out by frowning for a while. The _performance_ example doesn't handle resizing & I didn't really want to add it.

It is a little unexpected the window size changes shortly after creation, or that the initial window size is wrong until a resize event.

francesca64:
> It is a little unexpected the window size changes shortly after creation, or that the initial window size is wrong until a resize event.

The problem is that half of the platforms deal natively in logical pixels, whereas the other half deal natively in physical pixels. Without knowing which monitor a window will be on in advance, this quirk is unavoidable.

An application that doesn't handle resizes is a broken application, as there are *many* ways to trigger a resize event. However, if you wish to avoid the initial resize on Windows, you can simply disable DPI awareness on that platform: https://github.com/tomaka/winit/blob/089816d9bab498c0ec002d7e560ad940b78d476e/src/os/windows.rs#L15

Doing so results in Windows pretending that logical pixels are physical pixels, so while the render might be a bit blurry, it won't cause any issues.

alexheretic:
> An application that doesn't handle resizes is a broken application

Yeah I think that's reasonable. It was really only a problem with the performance example, which isn't real code.

In the main new winit changes make logical vs physical much clearer, so thanks!

PR #2189, opened 2018-06-28
[v0.17] Update winit -> 0.16

alexheretic:
* gfx_window_glutin -> `0.25`
* gfx_window_dxgi -> `0.16`
* Use physical dimensions in window projects.
* Use `poll_events` in examples instead of `run_forever` which can only handle one event per frame. _Included in this branch as otherwise testing is a pain_.
* Use WINIT_HIDPI_FACTOR override in performance example to avoid non-1 dpi x11 issues as this example can't handle resizing.

PR checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,13 +25,13 @@ name = "gfx_app"
 [dependencies]
 log = "0.4"
 env_logger = "0.5"
-glutin = "0.16"
-winit = "0.15"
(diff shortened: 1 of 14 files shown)


alexheretic:
@kvark gfx_window_glutin release?

kvark:
https://crates.io/crates/gfx_window_glutin/0.25.0

PR #2195, opened 2018-06-29
[mtl] cache depth/stencil states

kvark:
Fixes the dynamic allocation of depth/stencil states in Metal.
PR checklist:

The new logic goes as follows:
  - there is a central storage for depth-stencil states in our service pipes
  - a command buffer keeps a hold of just 2 things: the DS state that came from the last graphics pipeline and the set of stencil masks & reference values
  - a command buffer then assembles a full DS state from those things in the following sync points:
    - binding a render pipeline
    - setting stencil masks
    - starting a render pass

I find this logic easier to understand than the previous solution. It allows us to avoid re-creating Metal depth-stencil states. As a small bonus, it becomes possible to work around Dota's bug of trying to use a depth state without a proper attachment (included in the PR).


grovesNL:
Confirmed the CTS failure seems unrelated to this PR, created #2199

kvark:
Thank you!
With pattern matching, we try to avoid more generic patterns if we know a more specific one.
Bors r=grovesNL

PR #2200, opened 2018-06-30
[mtl] Retain copied render pass descriptor

### LINKED ISSUE #2199: Investigate Metal CTS regression
It appears that a CTS test category started to be flaky with commit d6b8b934b for some reason.

When running the `dEQP-VK.dynamic_state.ds_state` category (at once, not independently) on commits earlier to d6b8b934b, we always receive three `Pass` and two `NotSupported`.

However when running with commit d6b8b934b or beyond, sometimes (usually 1 out of every 3 or 4 runs) the `dEQP-VK.dynamic_state.ds_state.stencil_params_advanced` test fails. It's not clear why this commit would have changed this behavior.

grovesNL:
Appears to fix #2199 

PR checklist:

With this change, `dEQP-VK.dynamic_state.ds_state` seems to pass consistently again. The reason I looked into this `Deferred` branch is because the autorelease pool wouldn't have covered it previously. So moving it up the callstack caused the copied render pass descriptor to be allocated correctly but freed before its used.

I haven't tested this at all (outside of the CTS), so this may leak in other places, but the reasoning seems to make sense at least.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -842,7 +842,9 @@ impl CommandSink {
                     //Note: the original descriptor belongs to the framebuffer,
                     // and will me mutated afterwards.
                     desc: unsafe {
-                        msg_send![descriptor, copy]
+                        let desc: metal::RenderPassDescriptor = msg_send![descriptor, copy];
+                        msg_send![desc.as_ptr(), retain];
+                        desc
                     },
                     commands: init_commands.map(soft::RenderCommand::own).collect(),
                 });


src/backend/metal/src/command.rs
```diff
@@ -842,7 +842,9 @@ impl CommandSink {
                     //Note: the original descriptor belongs to the framebuffer,
                     // and will me mutated afterwards.
                     desc: unsafe {
-                        msg_send![descriptor, copy]
+                        let desc: metal::RenderPassDescriptor = msg_send![descriptor, copy];
+                        msg_send![desc.as_ptr(), retain];
```
grovesNL: Alternatively I suppose we could have `retained_descriptor` or similar (to be consistent)

kvark:
Oh, great!
Bors r+

PR #2202, opened 2018-07-02
A bunch of Metal goodies (better image clears and debugging)

kvark:
Fixes a large number of CTS's out-of-pass image clearing tests.

PR checklist:

### DIFF
--- a/src/backend/auxil/range_alloc.rs
+++ b/src/backend/auxil/range_alloc.rs
@@ -1,8 +1,7 @@
 use std::fmt::Debug;
 use std::ops::{Add, AddAssign, Range, Sub};
 
-#[derive(Derivative)]
-#[derivative(Debug)]
+#[derive(Debug)]
 pub struct RangeAllocator<T> {
     /// The range this allocator covers.
     initial_range: Range<T>,
@@ -10,6 +9,11 @@ pub struct RangeAllocator<T> {
     /// Must be ordered with ascending range start to permit short circuiting allocation.
     /// No two ranges in this vec may overlap.
     free_ranges: Vec<Range<T>>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct RangeAllocationError<T> {
+    pub fragmented_free_length: T,
 }
 
 impl<T> RangeAllocator<T>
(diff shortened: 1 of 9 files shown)


src/backend/auxil/range_alloc.rs
```diff
@@ -23,11 +27,13 @@ where
         }
     }
 
-    pub fn allocate_range(&mut self, length: T) -> Option<Range<T>> {
+    pub fn allocate_range(&mut self, length: T) -> Result<Range<T>, RangeAllocationError<T>> {
         assert_ne!(length + length, length);
         let mut best_fit: Option<(usize, Range<T>)> = None;
+        let mut fragmented_free_length = length - length;
```
Xaeroxe: This looks like code that could be made more clear by using traits from the `num` crate.  Specifically [this](https://docs.rs/num/0.2.0/num/trait.Zero.html) would be of use here.

src/backend/auxil/range_alloc.rs
```diff
@@ -49,14 +55,19 @@ where
...
+                    self.free_ranges.remove(index);
+                } else {
+                    self.free_ranges[index].start += length;
+                }
+                Ok(range.start..(range.start + length))
             }
-            range.start..(range.start + length)
-        })
+            None => Err(RangeAllocationError {
+                fragmented_free_length,
```
Xaeroxe: Rather than using a `match` here you could just call `ok_or` on the output from the original map function.

src/backend/auxil/range_alloc.rs
Xaeroxe: `num-traits` is also a good crate if you don't want the other goodies it comes with.

kvark:
Thanks @Xaeroxe for taking a look!

I do agree that the zero checks/initialization with generic `T` is wonky as hell, but I don't want to hook up an extra dependency just to make it nicer. Perhaps, we could refactor it without `num`?..

I don't feel strongly about `match` versus those `map/ok_or` things. It's getting into the conflicted territory here between the preference of simplest language constructs (`match` is simpler than passing a closure somewhere) and avoiding some minor duplication (`Ok/Err` are mentioned twice).

Given no blocking concerns, let's proceed!
bors r=Xaeroxe

Xaeroxe:
FWIW `num` is depended on by 368 other crates, several of which are graphics or audio related, so there's a pretty good chance the end user is compiling it anyways.

kvark:
bors retry

kvark:
bors retry

kvark:
bors r=Xaeroxe

PR #2203, opened 2018-07-02
[dx11] rework descriptors, add UAV/RTV for subresources

fkaa:
Descriptor pools now holds a fixed size `Vec` containing d3d11 interface pointers, managed by `RangeAlloc`. Slices of the `Vec` are given out to descriptor sets .

A descriptor set layout contains information on what binding number maps to what index in the descriptor set's slice of handles. This is copied to descriptor sets on set allocation.

PipelineLayout has optimized ranges (less api calls) of descriptors for each relevant shader stage.

Images now create & store subresource views (only UAV & RTV for now) on image binding.

Also fixes some minor stuff like invalidating memory ranges bigger than `1 << 15` vertex buffer strides & offsets.

r? @kvark

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -26,6 +26,14 @@ use {
 
 use {conv, internal, shader};
 
+struct InputLayout {
+    raw: ComPtr<d3d11::ID3D11InputLayout>,
+    required_bindings: u32,
+    max_vertex_bindings: u32,
+    topology: d3d11::D3D11_PRIMITIVE_TOPOLOGY,
+    vertex_strides: Vec<u32>,
+}
+
 pub struct Device {
     raw: ComPtr<d3d11::ID3D11Device>,
     pub(crate) context: ComPtr<d3d11::ID3D11DeviceContext>,
@@ -104,11 +112,21 @@ impl Device {
         }
     }
 
-    fn create_input_layout(&self, vs: ComPtr<d3dcommon::ID3DBlob>, vertex_buffers: &[pso::VertexBufferDesc], attributes: &[pso::AttributeDesc], input_assembler: &pso::InputAssemblerDesc) -> Result<([u32; d3d11::D3D11_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT as usize], d3d11::D3D11_PRIMITIVE_TOPOLOGY, ComPtr<d3d11::ID3D11InputLayout>), pso::CreationError> {
+
+    fn create_input_layout(&self, vs: ComPtr<d3dcommon::ID3DBlob>, vertex_buffers: &[pso::VertexBufferDesc], attributes: &[pso::AttributeDesc], input_assembler: &pso::InputAssemblerDesc) -> Result<InputLayout, pso::CreationError> {
         let mut layout = ptr::null_mut();
 
-        let mut vertex_strides = [0u32; d3d11::D3D11_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT as usize];
+        let mut vertex_strides = Vec::new();
+        let mut required_bindings = 0u32;
+        let mut max_vertex_bindings = 0u32;
         for buffer in vertex_buffers {
+            required_bindings |= 1 << buffer.binding as u32;
+            max_vertex_bindings = max_vertex_bindings.max(1u32 + buffer.binding as u32);
+
+            while vertex_strides.len() <= buffer.binding as usize {
+                vertex_strides.push(0);
+            }
+
             vertex_strides[buffer.binding as usize] = buffer.stride;
         }
 
@@ -121,6 +139,7 @@ impl Device {
                     Some(buffer_desc) => buffer_desc,
                     None => {
                         // TODO:
+                        // L
                         // error!("Couldn't find associated vertex buffer description {:?}", attrib.binding);
                         return Some(Err(pso::CreationError::Other));
                     }
@@ -164,7 +183,13 @@ impl Device {
         if winerror::SUCCEEDED(hr) {
             let topology = conv::map_topology(input_assembler.primitive);
 
-            Ok((vertex_strides, topology, unsafe { ComPtr::from_raw(layout) }))
+            Ok(InputLayout {
+                raw: unsafe { ComPtr::from_raw(layout) },
+                required_bindings,
+                max_vertex_bindings,
+                topology,
+                vertex_strides,
+            })
         } else {
             Err(pso::CreationError::Other)
         }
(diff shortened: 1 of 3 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -1004,31 +1123,71 @@ impl hal::Device<Backend> for Device {
...
-        _descriptor_pools: I,
+        max_sets: usize,
+        ranges: I,
     ) -> DescriptorPool
     where
         I: IntoIterator,
         I::Item: Borrow<pso::DescriptorRangeDesc>
     {
-        // TODO: descriptor pool
+        let count = ranges.into_iter().map(|r| r.borrow().count).sum::<usize>() * max_sets;
```
kvark: I don't think we need to multiply that total count in ranges with the max sets. Those are independent limits.

src/backend/dx11/src/device.rs
```diff
@@ -1004,31 +1123,71 @@ impl hal::Device<Backend> for Device {
...
+                stage: binding.stage_flags,
+                ty: binding.ty,
+                binding_range: binding.binding..(binding.binding + 1),
+                handle_offset: 0
+            });
+        }
+
+        // we sort the internal descriptor's handle (the actual dx interface) by some categories to
+        // make it easier to group api calls together
+        bindings.sort_unstable_by(|a, b| {
```
kvark: nit: could be easier to just do sort by key being `(a.ty, a.binding_range.star, a.stage)`.

src/backend/dx11/src/lib.rs
```diff
@@ -709,7 +711,10 @@ pub struct CommandBuffer {
     #[derivative(Debug="ignore")]
     context: ComPtr<d3d11::ID3D11DeviceContext>,
     #[derivative(Debug="ignore")]
-    list: Option<ComPtr<d3d11::ID3D11CommandList>>
+    list: Option<ComPtr<d3d11::ID3D11CommandList>>,
+    vertex_buffers: [*mut d3d11::ID3D11Buffer; 16],
```
kvark: similarly, is there any reason to not have a single array of structures here (as opposed to structure of arrays)?

kvark: I think I already asked somewhere but forgot the answer - why aren't we using `ComPtr` here and in descriptors? Is this to save the clocks of bumping the refcounts? Doesn't appear too important for DX11 in this case (but still makes sense for DX12 if applicable).

src/backend/dx11/src/lib.rs
```diff
@@ -950,6 +1015,9 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
     }
 
     fn bind_graphics_pipeline(&mut self, pipeline: &GraphicsPipeline) {
+        self.vertex_strides = pipeline.strides;
+        self.vertex_offsets = pipeline.strides;
```
kvark: offsets = strides?

src/backend/dx11/src/lib.rs
```diff
@@ -1098,6 +1161,8 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
     }
 
     fn draw(&mut self, vertices: Range<VertexCount>, instances: Range<InstanceCount>) {
+        self.set_graphics_bind_point();
```
kvark: that is something I strongly feel against - doing any kind of work on per-draw basis. This is the highest frequency we got, better try to rework this to happen at the PSO binding or something.

src/backend/dx11/src/lib.rs
```diff
@@ -1270,38 +1330,53 @@ impl Memory {
...
+            self.unmap(&context);
+        }
+
+    }
+
     pub fn invalidate(&self, context: &ComPtr<d3d11::ID3D11DeviceContext>, range: Range<u64>) {
         for &(ref buffer_range, ref buffer) in self.local_buffers.borrow().iter() {
             if let Some(range) = intersection(&range, &buffer_range) {
-                unsafe {
+                let stride = 1 << 15;
```
kvark: what is `15` here?

src/backend/dx11/src/device.rs
fkaa: Maybe my understanding of descriptor pools are wrong. You supply arbitrary number of descriptor types and lengths which it hands out for allocation. Is the amount of sets you provide to pool creation not related at all to how many descriptors you can allocate?

src/backend/dx11/src/lib.rs
fkaa: Compare
```rust
self.context.IASetVertexBuffers(
    0,
    d3d11::D3D11_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT,
    self.vertex_buffers.as_ptr(),
    self.vertex_strides.as_ptr(),
    self.vertex_offsets.as_ptr(),
);
```
to
```rust
for (i, (buffer, stride, offset)) in self.vertex_buffer_aos.iter().enumerate() {
    self.context.IASetVertexBuffers(
        i,
        1,
        buffer,
        stride,
        offset,
    );
}
```

fkaa: Mostly to avoid having to make an API call to set every resource individually when the functions take start, end, ptr arguments

src/backend/dx11/src/lib.rs
fkaa: Problem comes from DX11 pipeline states not holding vertex buffer strides and offsets and instead has to bind them on every `IASetVertexBuffer` call. This means we cant directly set the vertex buffers on a `set_vertex_buffer` call, but we have to wait until *after* our pipeline is bound and the only reliable place to set this is unfortunately at draw call. This can probably be improved quite a bit with some `dirty` bool flags, but I don't see a way around it.

src/backend/dx11/src/device.rs
```diff
@@ -409,9 +445,90 @@ impl hal::Device<Backend> for Device {
...
+            // PS/VS/CSSetXX()
+            for &stage in &stages {
+                let mut current_type = None;
+                let mut current_range = None;
+                let mut current_offset = 0;
+
+                for binding in bindings {
+                    match (current_type, current_range.clone()) {
+                        (None, None) => {
+                            if binding.stage.contains(stage) {
```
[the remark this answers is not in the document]
fkaa: The intended behavior is:
```
For every individual shader stage:
    For every descriptor binding:
        If binding belongs to stage, add to current range
```
So I think the above is correct, but looking at the code again I see that I dont check for the shader stage in the other match arm at all.

src/backend/dx11/src/device.rs
kvark: I'm open to consider a different interpretation, but I see it as a different limit. E.g. you can create up to as many of each of the descriptors AND up to this many of descriptor sets in total. How exactly each of the sets is using the descriptors is not important.
Let me invite @msiglreith to clarify, if they have a clearer understanding.

src/backend/dx11/src/lib.rs
kvark: yep, makes sense

src/backend/dx11/src/lib.rs
kvark: We actually have a similar problem in Metal (dependency of vertex buffers to the pipeline state). It is somewhat solved by trying to re-bind the vertex buffer on *BOTH* pipeline state setup and vertex buffer setup. There are cases where the combined state doesn't make sense (and we strongly expect the user to follow up with binding some other PSO/vertex buffer before issuing actual draw calls), and we should try to detect it here are avoid setting it up with D3D12 (bonus points for asserting if the user *does* try to render with the incomplete state).

src/backend/dx11/src/device.rs
msiglreith: @kvark is correct ([Slide 22](http://32ipi028l5q82yhj72224m8j-wpengine.netdna-ssl.com/wp-content/uploads/2016/05/Most-common-mistakes-in-Vulkan-apps.pdf) for IHV response on this)

src/backend/dx11/src/lib.rs
fkaa: Would having a conditional jump to check for dirty state and set them appropriately be *that* bad to put in `draw_*`? There's a bit more state that needs to be separated as well (eg. scissors, viewports dont allow for providing an offset) and I can imagine it could get a bit messy. I'm thinking of something like the following:
```rust
fn set_state(&mut self) {
    if self.state.contains(DirtyState::VERTEX_BUFFERS) {
        self.bind_vertex_buffers();
    }

    ...

    self.state = 0;
}

fn draw(&mut self, ...) {
    if self.dirty != 0 {
        self.set_state()
    }

    unsafe {
        self.context.DrawInstanced(
            ...
        );
    }
}
```

kvark: This dirty mechanics is what Angle and MoltenVK do. If you think of an app doing 10K-50K draw calls *per frame* (which is 1M+ per second), those little switches start matter. Not just in terms of the branch cost, but more about the code inlining and the instruction cache hits.
We could just give up on that sort of constraint in D3D11 and make it a truly 2nd tier backend, and ultimately it's your call to make, since you are the owner of D3D11 backend, but I feel responsible to provide you with the overall vision of an "ideal" backend that I myself at least try to reach in Metal/D3D12 backends.

kvark: And yes, this may sound like a premature optimization. We may very well recognize later that the overhead of the driver is so high that our switches don't matter. It's crucial to measure this overhead, but we don't have any real-world benchmarks (other than Dota, perhaps). Changing this decision later along the way may also be expensive.

fkaa: You have me convinced :-)

PR #2205, opened 2018-07-04
[mtl] don't stop the capture at present

### LINKED ISSUE #2204: Metal capturing is broken
As of a few weeks, we aren't able to capture some of the workloads:
  - dota2 capturing fails
  - WR capturing fails - https://github.com/szeged/webrender/issues/189#issuecomment-402176347
  - Quad example - unconfirmed
  - CTS tests are still captured just fine

We basically see a `stopCapture` called somewhere from the *user* code (not us) for some reason ([example call stack](https://user-images.githubusercontent.com/2149430/42225356-ab2255ec-7edc-11e8-8595-c26174c45488.png))... This hinders our ability to debug issues, given no alternative GPU debugging solution on OSX, and needs to be investigated promptly.

kvark:
Fixes #2204
PR checklist:

Apparently, capture is stopping automatically at present(), and we abort if if call `stop_capture` explicitly...

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -23,7 +23,7 @@ use hal::queue::{RawCommandQueue, RawSubmission};
 use hal::range::RangeArg;
 
 use foreign_types::{ForeignType, ForeignTypeRef};
-use metal::{self, MTLViewport, MTLScissorRect, MTLPrimitiveType, MTLIndexType, MTLSize, CaptureManager};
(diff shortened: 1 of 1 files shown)


kvark:
I confirmed this to allow the quad example to be captured. IIRC, it still didn't help Dota, but I think it will unblock WebRender at least. Merging as trivial and harmless.
bors r=common sense

kvark:
Common sense missed the unused import warning...
bors r=common sense

PR #2207, opened 2018-07-05
[mtl] debug option to retain references

kvark:
Allows us to quickly test if the reference retaining saves the day.
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -39,6 +39,7 @@ const CLEAR_IMAGE_ARRAY: bool = false;
 pub struct QueueInner {
     raw: metal::CommandQueue,
     reserve: Range<usize>,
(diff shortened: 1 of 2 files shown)


src/backend/metal/src/command.rs
```diff
@@ -1246,6 +1260,7 @@ impl RawCommandQueue<Backend> for CommandQueue {
         IC: IntoIterator,
         IC::Item: Borrow<CommandBuffer>,
     {
+        let _ap = AutoreleasePool::new();
```
fkaa: Should this also be behind a `if self.debug_retain_references { .. }`?

kvark:
Thanks!
This autorelease pool is for command buffers themselves, not their used resources, so it equally applies.
Bors r=fkaa

PR #2210, opened 2018-07-07
[mtl] cull face support

kvark:
PR checklist:

r? anyone

Note: the viewport change is needed because that's what CTS wants to use (with MAINTENANCE1 extension) for the cull tests... although they still fail. It looks like Metal changes the winding order based on the viewport, while Vulkan does not (or the other way around...).


kvark:
Implemented the front winding now, and it resolved the CTS issues (yay!). Culling both sides is still not implemented though - needs some shader magic (not for this PR):
> RasterizationEnabled is false but the vertex shader's return type is not void

fkaa:
👍 LGTM, surprised the type change in `Rect` didn't cause other backends to fail

kvark:
Thanks! Every backend just casts it anyway ;)
Bors r=fkaa

kvark:
Bors retry

PR #2212, opened 2018-07-08
[dx11] change buffer UAV/SRVs to raw views and add compute pipelines

fkaa:
r? @kvark

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -815,6 +815,28 @@ impl CommandBuffer {
         }
     }
 
+    unsafe fn bind_compute_descriptor(&self, context: &ComPtr<d3d11::ID3D11DeviceContext>, binding: &PipelineBinding, handles: *mut Descriptor) {
+        use pso::DescriptorType::*;
+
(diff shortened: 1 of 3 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -815,6 +815,28 @@ impl CommandBuffer {
...
+            SampledImage => context.CSSetShaderResources(start, len, handles as *const *mut _ as *const *mut _),
+            CombinedImageSampler => {
+                context.CSSetShaderResources(start, len, handles as *const *mut _ as *const *mut _);
+                context.CSSetSamplers(start, len, handles.offset(1) as *const *mut _ as *const *mut _);
+            },
+            UniformBuffer |
+            UniformBufferDynamic => context.CSSetConstantBuffers(start, len, handles as *const *mut _ as *const *mut _),
+            StorageImage |
+            StorageBuffer => context.CSSetUnorderedAccessViews(start, len, handles as *const *mut _ as *const *mut _, ptr::null_mut()),
+            _ => unimplemented!()
```
kvark: nit: let's call those out explicitly

PR #2216, opened 2018-07-10
[dx11] add more view support and memory type enum

fkaa:
Adds support for creating views of 2d arrays, 3d textures, cube maps and cube map arrays.

Changes the views that image binding creates for copying/blitting etc. from `D2` to `D2Array`

Adds a memory type enum which replaces the previous hardcoded conditionals.

### DIFF
--- a/src/backend/dx11/shaders/blit.hlsl
+++ b/src/backend/dx11/shaders/blit.hlsl
@@ -22,29 +22,29 @@ VsOutput vs_blit_2d(uint id: SV_VertexID) {
 
 SamplerState BlitSampler : register(s0);
 
-Texture2D<uint4>  BlitSrc_Uint  : register(t0);
-Texture2D<int4>   BlitSrc_Sint  : register(t0);
-Texture2D<float4> BlitSrc_Float : register(t0);
+Texture2DArray<uint4>  BlitSrc_Uint  : register(t0);
+Texture2DArray<int4>   BlitSrc_Sint  : register(t0);
+Texture2DArray<float4> BlitSrc_Float : register(t0);
 
 // TODO: get rid of GetDimensions call
 uint4 Nearest_Uint(float4 uv)
 {
-    float2 size;
-    BlitSrc_Uint.GetDimensions(size.x, size.y);
+    float4 size;
+    BlitSrc_Uint.GetDimensions(0, size.x, size.y, size.z, size.w);
 
-    float2 pix = uv.xy * size;
+    float2 pix = uv.xy * size.xy;
 
-    return BlitSrc_Uint[int2(pix)].xyzw;
+    return BlitSrc_Uint.Load(int4(int2(pix), uv.zw)).xyzw;
 }
 
 int4 Nearest_Sint(float4 uv)
 {
-    float2 size;
-    BlitSrc_Sint.GetDimensions(size.x, size.y);
+    float4 size;
+    BlitSrc_Sint.GetDimensions(0, size.x, size.y, size.z, size.w);
 
-    float2 pix = uv.xy * size;
+    float2 pix = uv.xy * size.xy;
 
-    return BlitSrc_Sint[int2(pix)].xyzw;
+    return BlitSrc_Sint.Load(int4(int2(pix), uv.zw)).xyzw;
 }
 
 uint4 ps_blit_2d_uint(VsOutput input) : SV_Target
@@ -59,5 +59,5 @@ int4 ps_blit_2d_int(VsOutput input) : SV_Target
 
 float4 ps_blit_2d_float(VsOutput input) : SV_Target
 {
-    return BlitSrc_Float.Sample(BlitSampler, input.uv);
+    return BlitSrc_Float.SampleLevel(BlitSampler, input.uv.xyz, input.uv.w);
 }
--- a/src/backend/dx11/shaders/blit.hlsl
+++ b/src/backend/dx11/shaders/blit.hlsl
@@ -34,7 +34,7 @@ uint4 Nearest_Uint(float4 uv)
 
     float2 pix = uv.xy * size.xy;
 
-    return BlitSrc_Uint.Load(int4(int2(pix), uv.zw)).xyzw;
+    return BlitSrc_Uint.Load(int4(int2(pix), uv.zw));
 }
 
 int4 Nearest_Sint(float4 uv)
@@ -44,7 +44,7 @@ int4 Nearest_Sint(float4 uv)
 
     float2 pix = uv.xy * size.xy;
 
-    return BlitSrc_Sint.Load(int4(int2(pix), uv.zw)).xyzw;
+    return BlitSrc_Sint.Load(int4(int2(pix), uv.zw));
 }
 
 uint4 ps_blit_2d_uint(VsOutput input) : SV_Target
--- a/src/backend/dx11/src/conv.rs
+++ b/src/backend/dx11/src/conv.rs
@@ -230,8 +230,8 @@ pub fn map_rect(rect: &Rect) -> D3D11_RECT {
     D3D11_RECT {
         left: rect.x as _,
         top: rect.y as _,
-        right: rect.w as _,
-        bottom: rect.h as _,
+        right: (rect.x + rect.w) as _,
+        bottom: (rect.y + rect.h) as _,
     }
 }
 
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -21,7 +21,7 @@ use {
     Backend, Buffer, BufferView, CommandPool, ComputePipeline, DescriptorPool, DescriptorSetLayout,
     Fence, Framebuffer, GraphicsPipeline, Image, ImageView, InternalBuffer, InternalImage, Memory,
     PipelineLayout, QueryPool, RenderPass, Sampler, Semaphore, ShaderModule, Surface, Swapchain,
-    UnboundBuffer, UnboundImage, ViewInfo, PipelineBinding, Descriptor, MemoryHeapType
+    UnboundBuffer, UnboundImage, ViewInfo, PipelineBinding, Descriptor, MemoryHeapFlags
 };
 
 use {conv, internal, shader};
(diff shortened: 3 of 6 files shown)


src/backend/dx11/shaders/blit.hlsl
```diff
@@ -22,29 +22,29 @@ VsOutput vs_blit_2d(uint id: SV_VertexID) {
...
-    float2 size;
-    BlitSrc_Uint.GetDimensions(size.x, size.y);
+    float4 size;
+    BlitSrc_Uint.GetDimensions(0, size.x, size.y, size.z, size.w);
 
-    float2 pix = uv.xy * size;
+    float2 pix = uv.xy * size.xy;
 
-    return BlitSrc_Uint[int2(pix)].xyzw;
+    return BlitSrc_Uint.Load(int4(int2(pix), uv.zw)).xyzw;
```
kvark: those `xyzw` at the end are redundant

src/backend/dx11/src/device.rs
```diff
@@ -405,36 +488,44 @@ impl hal::Device<Backend> for Device {
...
-                    &desc,
-                    ptr::null_mut(),
-                    &mut working_buffer as *mut *mut _ as *mut *mut _
-                )
-            };
+        let (working_buffer, working_buffer_size) = match heap_ty {
+            MemoryHeapType::DEVICE_LOCAL => {
+                (None, 0)
+            },
+            MemoryHeapType::HOST_NONCOHERENT | MemoryHeapType::HOST_COHERENT => {
```
kvark: this works as a pattern? O_O

src/backend/dx11/src/device.rs
```diff
@@ -992,7 +1081,14 @@ impl hal::Device<Backend> for Device {
                     Usage: usage,
                     BindFlags: bind,
                     CPUAccessFlags: cpu,
-                    MiscFlags: 0
+                    // unfortunately no better way of doing this, since creating a cube view
```
kvark: actually, we do know that from `image.flags`

src/backend/dx11/src/device.rs
```diff
@@ -1391,6 +1492,10 @@ impl hal::Device<Backend> for Device {
             let ptr = host_visible.borrow_mut().as_mut_ptr();
             memory.mapped_ptr.replace(Some(ptr));
 
+            if memory.ty == MemoryHeapType::HOST_COHERENT {
+                memory.invalidate(&self.context, memory.resolve(&(None, None)));
```
kvark: This is not actually correct. AFAIK, the operation of mapping/unmapping memory doesn't involve any synchronization in Vulkan *regardless* of the memory coherency:
  - non-coherent memory has to still be explicitly sync-ed with flush/invalidate
  - coherent memory still has to affect all subsequent device and queue based commands (e.g. submissions), even if not unmapped.

src/backend/dx11/src/lib.rs
```diff
@@ -1312,6 +1316,25 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
     }
 }
 
+bitflags! {
+    struct MemoryHeapType: u64 {
```
kvark: semantic issue here: `XxxType` implies an enum, but you have bit flags. Should be `XxxFlags` or `XxxProperties`

src/backend/dx11/src/lib.rs
```diff
@@ -1312,6 +1316,25 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
...
+    }
+}
+
+impl MemoryHeapType {
+    pub fn from_raw(index: usize) -> Option<Self> {
+        match Self::from_bits(1 << index).unwrap() {
+            Self::DEVICE_LOCAL => Some(Self::DEVICE_LOCAL),
+            Self::HOST_NONCOHERENT => Some(Self::HOST_NONCOHERENT),
+            Self::HOST_COHERENT => Some(Self::HOST_COHERENT),
+            _ => None
```
kvark: how is this possible?

src/backend/dx11/src/device.rs
fkaa: Yeah, left side only allows patterns, so eg. `1 + 2 => { ... }` wont compile

src/backend/dx11/src/device.rs
fkaa: Must've missed it!

src/backend/dx11/src/device.rs
fkaa: Ouch!

Do we need to put in checks on submissions etc. and flush mapped coherent memory then? Sounds a bit messy but not impossible..

src/backend/dx11/src/device.rs
kvark: oh, I get it now, this is not a match over the bitflag combo, this is a match over one of the two flags

src/backend/dx11/src/device.rs
kvark: Well, putting the sync in map/unmap is simply incorrect. It's not even a subset of the proper logic, as I explained. And if you just remove it, then coherent memory wouldn't be synced at all... I know, this sucks. And ultimately the main client of your backend is WebRender, which can just use non-coherent memory. So if you can think of a quick and easy way to provide at least some coherency guarantees (that are a subset of the end goal), go for it. Otherwise, let's try implementing the proper logic in place.

src/backend/dx11/src/lib.rs
fkaa: Invalid input? I suppose that's not really my job though :p. Also noticed there is a `from_bits_truncate` in bitflags which is more suitable.

src/backend/dx11/src/device.rs
kvark: Not trivially. The copies to/from are command buffer operations. It's not correct to synchronize at command buffer *recording*, you'll have to defer this to submission time.

fkaa:
@kvark quite a big commit, but I think I got the coherency problem nailed down, and it seems to work on multiple vulkan examples that use coherent memory

src/backend/dx11/src/conv.rs
```diff
@@ -226,6 +226,15 @@ pub fn map_viewport(viewport: &Viewport) -> D3D11_VIEWPORT {
     }
 }
 
+pub fn map_scissor(rect: &Rect) -> D3D11_RECT {
+    D3D11_RECT {
+        left: rect.x as _,
+        top: rect.y as _,
+        right: (rect.x + rect.w) as _,
```
kvark: wait a sec, why are we keeping `map_rect` incorrect instead of fixing and reusing it?

src/backend/dx11/src/device.rs
```diff
@@ -933,7 +1001,7 @@ impl hal::Device<Backend> for Device {
             requirements: memory::Requirements {
                 size: size,
                 alignment: 1,
-                type_mask: 0x1,
+                type_mask: MemoryHeapFlags::DEVICE_LOCAL.bits(),
```
kvark: FYI, technically, a Vulkan implementation has to also support trivial 2D images that are CPU visible. I don't think it's worth pursuing for you though.

src/backend/dx11/src/device.rs
```diff
@@ -992,7 +1058,11 @@ impl hal::Device<Backend> for Device {
                     Usage: usage,
                     BindFlags: bind,
                     CPUAccessFlags: cpu,
-                    MiscFlags: 0
+                    MiscFlags: if image.flags == image::StorageFlags::CUBE_VIEW {
```
kvark: should check for `contains` instead of equality

src/backend/dx11/src/device.rs
```diff
@@ -1319,9 +1394,22 @@ impl hal::Device<Backend> for Device {
                     pso::Descriptor::Buffer(buffer, ref _range) => {
                         match ty {
                             pso::DescriptorType::UniformBuffer => {
+                                if buffer.ty == MemoryHeapFlags::HOST_COHERENT {
+                                    let old_buffer = unsafe { (*handle).0 } as *mut _;
+
+                                    write.set.flush_coherent(old_buffer, buffer);
```
kvark: do we have to flush at `write_descriptor_sets`?

src/backend/dx11/src/lib.rs
```diff
@@ -678,7 +682,14 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
     {
         for cmd_buf in submission.cmd_buffers.into_iter() {
```
kvark: note: `into_iter()` is called automatically by `for`, so you can remove it from here

src/backend/dx11/src/lib.rs
```diff
@@ -697,7 +708,8 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
     }
 
     fn wait_idle(&self) -> Result<(), error::HostExecutionError> {
-        unimplemented!()
+        // unimplemented!()
```
kvark: maybe `error!`?

src/backend/dx11/src/lib.rs
```diff
@@ -872,7 +919,7 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
     }
 
     fn reset(&mut self, _release_resources: bool) {
-        unimplemented!()
+        // unimplemented!()
```
kvark: is this working correctly now?

src/backend/dx11/src/lib.rs
```diff
@@ -1358,31 +1436,70 @@ fn intersection(a: &Range<u64>, b: &Range<u64>) -> Option<Range<u64>> {
     }
 }
 
-// TODO: implement flush/invalidate for bound images as well
-impl Memory {
-    pub fn resolve<R: hal::range::RangeArg<u64>>(&self, range: &R) -> Range<u64> {
-        *range.start().unwrap_or(&0) .. *range.end().unwrap_or(&self.size)
+impl MemorySync {
+    fn flush(&self, context: &ComPtr<d3d11::ID3D11DeviceContext>) {
+        if let Some(range) = intersection(&self.sync_range, &self.bound_range) {
```
kvark: should it be here in the first place if `sync_range` and `bound_range` don't intersect?

src/backend/dx11/src/device.rs
fkaa: Mhm, does seem like it could be a bit of a pain to add. Is there a list of minimum requirements or is it scattered over the spec?

src/backend/dx11/src/device.rs
fkaa: It only marks the descriptor set as having descriptor(s) referring to coherent buffers. On `bind_graphics_descriptor_sets` they are then added to the command buffer.

PR #2220, opened 2018-07-10
[mtl] defer the depth-stencil checks to the sync point

### LINKED ISSUE #2209: Stale depth-stencil state in Metal
See https://github.com/szeged/webrender/issues/189#issuecomment-403082557
Looks like we aren't always updating the d/s state in Metal still. Need to run WebRender's Wrench/Metal on `reftests/filters/filter-saturate-red-1-ref.yaml` and see what comes in versus what comes out.

kvark:
Fixes #2209
PR checklist:


kvark:
Bors r=grovesNL

PR #2221, opened 2018-07-10
Support using RTLD_Default

ZeGentzy:

Depends on: https://github.com/MaikKlein/ash/pull/77#discussion_r201234586
PR checklist:
- [x] tested examples with the following backends: vulkan on archlinux


src/backend/vulkan/Cargo.toml
```diff
@@ -21,8 +22,9 @@ name = "gfx_backend_vulkan"
 byteorder = "1"
 log = "0.4"
 lazy_static = "1"
-shared_library = "0.1"
-ash = "0.24.3"
+shared_library = { version = "0.1.9", optional = true }
+#ash = "0.24.4"
```
ZeGentzy: Will fix when 0.24.4 in released.

src/backend/vulkan/Cargo.toml
```diff
@@ -13,6 +13,7 @@ workspace = "../../.."
 
 [features]
 default = ["winit"]
+use-rtld-default = ["shared_library"]
```
kvark: should we perhaps just gate it on `shared_library` itself?

ZeGentzy: We could, but then they'd have to look at the source code to know what we're using `shared_library` for. I think having the `use-rtld-default` feature makes it clearer what the purpose is.

ZeGentzy: Enabling it by default would require that all linux apps preload the system's `libvulkan.so.1`, AFAIK.

I guess we could name it something like 'use-default-vulkan-function-symbols` or something. Truth is, 'RTLD_{default,next}` are obscure features which will probably only be used by someone looking to wrap something. I'm guessing that anyone who is looking to wrap a library probably googled something along the lines of (wrap library calls linux ld_preload)[https://duckduckgo.com/?q=wrap+library+calls+linux+ld_preload&t=ffab&ia=qa] and read numerous blogs/websites using either `RTLD_DEFAULT` or `RTLD_NEXT`.

kvark: > Enabling it by default would require that all linux apps preload the system's libvulkan.so.1, AFAIK.

Could you explain why?

As for the name, you convinced me to keep it.

ZeGentzy: Ok, so I've tested this out, and you get the following error if you don't use `LD_PRELOAD`:
```
thread 'main' panicked at 'Unable to load Vulkan entry points: StaticLoadError(["\"vkGetInstanceProcAddr\""])', libcore/result.rs:945:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:511
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:426
   6: rust_begin_unwind
             at libstd/panicking.rs:337
   7: core::panicking::panic_fmt
             at libcore/panicking.rs:92
   8: core::result::unwrap_failed
             at /checkout/src/libcore/macros.rs:26
   9: <core::result::Result<T, E>>::expect
             at /checkout/src/libcore/result.rs:809
  10: gfx_backend_vulkan::Instance::create
             at src/backend/vulkan/src/lib.rs:160
  11: quad::main
             at examples/quad/main.rs:81
  12: std::rt::lang_start::{{closure}}
             at /checkout/src/libstd/rt.rs:74
  13: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:310
  14: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:105
  15: std::rt::lang_start_internal
             at libstd/panicking.rs:289
             at libstd/panic.rs:397
             at libstd/rt.rs:58
  16: std::rt::lang_start
             at /checkout/src/libstd/rt.rs:74
  17: main
  18: __libc_start_main
  19: _start
```

The way I understand it is that it searches for the first occurrence of the symbol (sorted by the default library search order) in the opened libraries. `LD_PRELOAD` opens your library in the front. If you don't use `LD_PRELOAD` to opens the system library then there will be no `libvulkan.so.1` opened for it to find.

kvark:
Is this still blocked?

ZeGentzy:
@kvark Yes, till @MaikKlein releases 0.24.4.

MaikKlein:
@ZeGentzy Should be live now

ZeGentzy:
@kvark Try now.

kvark:
would be good to test it on CI as well

PR #2223, opened 2018-07-11
[mtl] debug output for descriptor allocation and frame wait times

kvark:
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -9,8 +9,7 @@ use std::borrow::Borrow;
 use std::cell::RefCell;
 use std::ops::{Deref, Range};
 use std::sync::{Arc, Mutex};
-use std::{iter, mem};
-use std::slice;
+use std::{iter, mem, slice, time};
 
 use hal::{buffer, command as com, error, memory, pool, pso};
 use hal::{DrawCount, SwapImageIndex, VertexCount, VertexOffset, InstanceCount, IndexCount, WorkGroupCount};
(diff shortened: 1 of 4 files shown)


src/backend/metal/src/device.rs
```diff
@@ -125,6 +125,7 @@ pub struct Device {
     pub(crate) shared: Arc<Shared>,
     pub(crate) private_caps: PrivateCapabilities,
     memory_types: [hal::MemoryType; 4],
+    debug_min_descriptor_count: usize,
```
grovesNL: Was this left in this PR intentionally?

src/backend/metal/src/command.rs
```diff
@@ -1373,6 +1393,18 @@ impl RawCommandQueue<Backend> for CommandQueue {
 
         command_buffer.commit();
 
+        if let Some(ref mut wait) = self.frame_wait {
+            wait.frame += 1;
+            if wait.frame >= FRAME_WAIT_REPORT_WINDOW {
+                let time = wait.duration / wait.frame as u32;
+                println!("Frame wait time={}ms count={}",
```
grovesNL: The `println` could be undesirable for some applications if they're ok with this

kvark:
The printlns are currently blocked by a compile time constant, so shouldn't be a problem for now.

Bors r=grovesNL

src/backend/metal/src/device.rs
kvark: oh yes, this is an unfortunate workaround we have to get until Dota is fixed

PR #2224, opened 2018-07-12
[mtl] dynamic depth bias, callback coalescence, and more stats

### LINKED ISSUE #2170: Fix Z-fighting issues in Dota
Short info header:
- GFX version: 842b9313d9391dc99dc215bc2dd01f9e4daffcda
- OS: OSX 10.13
- GPU: Intel Iris Graphics 550

There is a visible Z-fighting on the ground, should not be too difficult to fix if you look at a GPU capture.

kvark:
Fixes #2170

Expands our Metal stats to report the number of active and temporary command buffers. Coalesces the completion handlers, and finally fixes the depth bias issue we've been seeing (HAL breaking change).

PR checklist:

### DIFF
--- a/src/backend/dx12/src/conv.rs
+++ b/src/backend/dx12/src/conv.rs
@@ -150,6 +150,11 @@ pub fn map_rasterizer(rasterizer: &pso::Rasterizer) -> D3D12_RASTERIZER_DESC {
     use hal::pso::PolygonMode::*;
     use hal::pso::FrontFace::*;
 
+    let bias = match rasterizer.depth_bias { //TODO: support dynamic depth bias
+        Some(pso::State::Static(db)) => db,
+        Some(_) | None => pso::DepthBias::default(),
+    };
+
     D3D12_RASTERIZER_DESC {
         FillMode: match rasterizer.polygon_mode {
             Point => {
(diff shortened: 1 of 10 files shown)


src/backend/dx12/src/conv.rs
```diff
@@ -150,6 +150,11 @@ pub fn map_rasterizer(rasterizer: &pso::Rasterizer) -> D3D12_RASTERIZER_DESC {
     use hal::pso::PolygonMode::*;
     use hal::pso::FrontFace::*;
 
+    let bias = match rasterizer.depth_bias { //TODO
```
grovesNL: What's the `TODO` for?

src/backend/gl/src/state.rs
```diff
@@ -20,11 +20,11 @@ pub fn bind_polygon_mode(gl: &gl::Gl, mode: pso::PolygonMode, bias: Option<pso::
     unsafe { gl.PolygonMode(gl::FRONT_AND_BACK, gl_draw) };
 
     match bias {
-        Some(bias) => unsafe {
+        Some(pso::State::Static(bias)) => unsafe {
```
grovesNL: Isn't it possible for a dynamic depth bias to have been set at this point?

Otherwise, where would the dynamic depth bias be set?

kvark: Yeah, the problem is that we haven't had a dynamic depth bias exposed at all, and I want to limit the scope of this PR to providing this in API and implementing in Metal and Vulkan only. Other backends are not regressed, they will just need to catch up.

kvark:
@grovesNL I improved those TODO comments now

kvark:
Retained stuff of a queue is only appended in &mut self methods, so there is no send/sync concern.

The condvar signalling was just missing afaik.

kvark:
I believe the concerns are addressed, and we can revisit the code if new ones rise up.
Let's not hold the PR for too long, it has too many goodies for us.
bors r=grovesNL

PR #2231, opened 2018-07-13
Add runtime check for `setDisplaySyncEnabled`

Michael-Lfx:
`setDisplaySyncEnabled` only supported on macOS 10.13+ currently, in order to run iOS, add runtime check for it.

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/metal/src/window.rs
+++ b/src/backend/metal/src/window.rs
@@ -10,7 +10,7 @@ use hal::{Backbuffer, SwapchainConfig};
 use hal::window::Extent2D;
 
 use metal;
-use objc::runtime::Object;
+use objc::runtime::{Object, YES, BOOL};
 use core_graphics::base::CGFloat;
 use core_graphics::geometry::CGRect;
 use cocoa::foundation::{NSRect};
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/window.rs
```diff
@@ -242,8 +242,13 @@ impl Device {
             msg_send![render_layer, setPixelFormat: mtl_format];
             msg_send![render_layer, setFramebufferOnly: framebuffer_only];
             msg_send![render_layer, setMaximumDrawableCount: config.image_count as u64];
-            //TODO: only set it where supported
-            msg_send![render_layer, setDisplaySyncEnabled: display_sync];
+            // determines whether the Metal layer
+            // and it's drawables are synchronized with the display's refresh rate.
+            let can_responds_to_display_sync_enabled: BOOL = msg_send![render_layer, respondsToSelector:sel!(setDisplaySyncEnabled:)];
```
kvark: this should be handled much earlier - in [Surface::capability](https://github.com/gfx-rs/gfx/blob/d07a03b4df5a0a49de319dcb8401c7d90014285d/src/backend/metal/src/window.rs#L191). We shouldn't report `Immediate` presentation mode as available on systems that don't support `setDisplaySyncEnabled` in the first place.

kvark: not exactly. We don't want to call `setDisplaySyncEnable` when querying for capability, but we should not return `Immediate` in the list if it's not available.

kvark:
ping!

PR #2234, opened 2018-07-14
Fix zero allocated range DX12

Napokue:
Fixes #issue
PR checklist:
- [x] `make` succeeds (on *nix) - Windows
- [x] tested examples with the following backends: Vulkan, DX12, GL (quad only)

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -2333,6 +2333,7 @@ impl d::Device<B> for Device {
         let descriptor_pools = descriptor_pools.into_iter()
                                                .map(|desc| *desc.borrow())
                                                .collect::<Vec<_>>();
+        
         for desc in &descriptor_pools {
             match desc.ty {
                 pso::DescriptorType::Sampler => {
@@ -2354,10 +2355,14 @@ impl d::Device<B> for Device {
                 .lock()
                 .unwrap();
 
-            let range = heap_srv_cbv_uav
-                .range_allocator
-                .allocate_range(num_srv_cbv_uav as _)
-                .unwrap(); // TODO: error/resize
+            let range = match num_srv_cbv_uav {
+                0 => 0..0,
+                _ => heap_srv_cbv_uav
+                        .range_allocator
+                        .allocate_range(num_srv_cbv_uav as _)
+                        .unwrap(), // TODO: error/resize
+            };
+
             n::DescriptorHeapSlice {
                 heap: heap_srv_cbv_uav.raw.clone(),
                 handle_size: heap_srv_cbv_uav.handle_size as _,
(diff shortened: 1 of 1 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -2334,18 +2334,8 @@ impl d::Device<B> for Device {
                                                .map(|desc| *desc.borrow())
                                                .collect::<Vec<_>>();
         for desc in &descriptor_pools {
```
Napokue: Based on commit: https://github.com/gfx-rs/gfx/commit/1f9b40e4 both `num_samplers` and `num_srv_cbv_uav` must be incremented, regardless on what type of DescriptorType is passed in.

ZeGentzy:
Tested in a VM and it fixes the Dx12 backend. Good Job!

src/backend/dx12/src/device.rs
kvark: Not sure how got went from the commit to the idea. The commit is for Metal backend specifically, and in general I don't see why we'd want to increase the sampler count for descriptors that have nothing to do with samplers.

kvark:
@Napokue ping! needs a squash, plus my comment addressed/answered

src/backend/dx12/src/device.rs
Napokue: Yes sorry for the late reaction. We spoke about this in Gitter and what I think the best solution would be is: check if `num_samplers` is equal to 0, if that is so I want to return a None, else I want to return the heap_sampler in the Some. 

Same goes for `num_srv_cbv_uav`.

Is this a correct solution?

Napokue: So about the solution, maybe something like this (bear with me, written from my head and without testing):

`fn get_descriptor_sampler(number: Option<?>) -> Option<?> {
    match numer {
        0 => None,
        _ => {
                    let mut heap_sampler = self
                       .heap_sampler
                       .lock()
                       .unwrap();

                    let range = heap_sampler
                       .range_allocator
                       .allocate_range(num_samplers as _)
                       .unwrap(); // TODO: error/resize

                    Some(n::DescriptorHeapSlice {
                        heap: heap_sampler.raw.clone(),
                        handle_size: heap_sampler.handle_size as _,
                        range_allocator: RangeAllocator::new(range),
                        start: heap_sampler.start,
                    })
        }
    }
}`

I am not sure about the return type of the option.

kvark: So you'd want to have both `heap_sampler` and `heap_srv_cbv_uav` to be stored as `Option<>`, based on whether or not the appropriate count was 0.

Napokue: Yes, that is right.

kvark:
That is slightly hacky (calling `RangeAllocator::new(0 .. 0)`) but is an acceptable workaround. Please squash your commits into one before proceeding.

Issue #2229, opened 2018-07-12
[meta] Locking on Metal


kvark:
This is a meta-issue to track all the locking we do in Metal backend. We need to check if they are appropriate, assume the worst use case, and try to optimize them.

- [x] service pipes - #2227 . They are locked quite a bit in various places.
- [ ] pass descriptor. Locked in `begin_render_pass`, so that we can mutate it and start a real pass. The problem is that an app can do a dozen of render passes from multiple threads on the same framebuffer. And starting one is accidentally the single heaviest operation of the backend...
- [ ] shared queue. It is accessed on submission as well as any time we need to spawn a new command buffer. These shouldn't take long, so I'd not expect a congestion, but it does show up in the profile.
- [ ] shared device. Locked whenever we need to create a resource, pipeline, or a new state. The cost should largely go away as the application warms up, but we need to double-check.
- [x] command buffer inner - #2226 . I don't think `RefCell` has any locking overhead.
- [x] descriptor pool inner -  #2235
- [x] fences - #2252
- [ ] swapchain frames. It doesn't look to matter so far. Changing the `acquire` to be blocking doesn't nearly affect the framerate.

kvark:
Major issues are fixed, I believe.

kvark:
Reviving this, since apparently our performance doesn't scale well enough from 2 to 4 cores. Not sure what else, if not locking, would affect it.

PR #2235, opened 2018-07-16
Metal pools improved

kvark:
Addresses part of  #2229 by ~~removing the descriptor pool lock entirely - it's meant to be externally synchronized by Vulkan anyway~~ moving the ranges out of the descriptor lock.
Fixes the correctness of a pool reset that's not created with the individual reset flag.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -115,7 +115,7 @@ type CommandBufferInnerPtr = Arc<RefCell<CommandBufferInner>>;
 
 pub struct CommandPool {
     pub(crate) shared: Arc<Shared>,
-    pub(crate) managed: Option<Vec<CommandBufferInnerPtr>>,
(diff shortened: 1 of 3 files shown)


kvark:
Actually, Dota doesn't like our new `RefCell` things. Need to investigate.

kvark:
Hmm, I think I see where my misunderstanding comes from. Binding descriptor sets from the pool being updated on a different thread seems to be OK according to the spec. So we may need to keep the locks in there :( At the very least, we can separate the lock into 2: one for the ranges (most allocations only need those) and one for the actual descriptors (used for immutable samplers and updates).

kvark:
@grovesNL alright, the new changes are good to review. Dota works, and reftests/quad too. The changes are less drastic (and risky). We can probably do more in the future, but this should be a good step anyway. TL;DR: for descriptors, we only lock the inner data of a pool if immutable samplers are provided. This should reduce the lock contention on some pools.

PR #2239, opened 2018-07-16
Mtl service locks

### LINKED ISSUE #2227: Optimize locking of `ServicePipes`
`ServicePipes`  in Metal is one struct with many semi-independent parts responsible for clears, blits, depth-stencil states, and more. We keep it in a `Mutex` shared by everyone, and I'm seeing the locks in the profiler. In particular, we lock it every time the depth-stencil needs updated.

We need to find a more efficient way to share these states:
  - investigate the performance of `Mutex` on OSX specifically. Mandatory reading - https://blog.mozilla.org/nfroyd/2017/03/29/on-mutex-performance-part-1/
  - split the struct into separate areas locked independenty

kvark:
Fixes #2227
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -3,7 +3,7 @@ use {
     BufferPtr, TexturePtr, SamplerPtr,
 };
 use {conversions as conv, native, soft, window};
-use internal::{BlitVertex, Channel, ClearKey, ClearVertex, ServicePipes};
+use internal::{BlitVertex, Channel, ClearKey, ClearVertex};
 
 use std::borrow::Borrow;
 use std::cell::RefCell;
@@ -36,6 +36,14 @@ const WORD_ALIGNMENT: u64 = WORD_SIZE as _;
 const CLEAR_IMAGE_ARRAY: bool = false;
 /// Number of frames to average when reporting the performance counters.
 const COUNTERS_REPORT_WINDOW: usize = 0;
+
+//TODO: use for inserting points of interest into System Trace
+#[allow(dead_code)]
+extern "C" {
+    fn kdebug_signpost(code: u32, arg1: *const u32, arg2: *const u32, arg3: *const u32, arg4: *const u32) -> u32;
+    fn kdebug_signpost_start(code: u32, arg1: *const u32, arg2: *const u32, arg3: *const u32, arg4: *const u32) -> u32;
+    fn kdebug_signpost_end(code: u32, arg1: *const u32, arg2: *const u32, arg3: *const u32, arg4: *const u32) -> u32;
+}
 
 pub struct QueueInner {
     raw: metal::CommandQueue,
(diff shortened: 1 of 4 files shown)


src/backend/metal/src/command.rs
```diff
@@ -37,6 +37,14 @@ const CLEAR_IMAGE_ARRAY: bool = false;
 /// Number of frames to average when reporting the performance counters.
 const COUNTERS_REPORT_WINDOW: usize = 0;
 
+//TODO: use for inserting points of interest into System Trace
```
grovesNL: This seems useful! 👀

src/backend/metal/src/command.rs
```diff
@@ -378,12 +386,13 @@ impl State {
             })
     }
 
-    fn sync_depth_stencil<'a>(
-        &'a self, pipes: &'a mut ServicePipes, device: &Mutex<metal::Device>,
-    ) -> Option<soft::RenderCommand<&'a soft::Own>> {
+    fn sync_depth_stencil<F>(&self, shared: &Shared, mut fun: F)
+    where
+        F : for<'a> FnMut(soft::RenderCommand<&'a soft::Own>),
```
grovesNL: This complicates things a bit, is there no way around it?

kvark:
>  I think the passing of fun around (in both sync_depth_stencil and with_depth_stencil) makes it harder to read. It's also slightly unfortunate that we have to split up render_commands in a few places. Perhaps we can invert these calls somehow?

I'd totally love to do this! Unfortunately, what's blocking is that we can no longer return a single (temporary) object from looking at an entry if we can dynamically decided to either read it or write it... So while technically we can return an `enum LockedMap { ReadGuard, WriteGuard }`, we'll then have to access the exact entry *again* when we want to use it, which introduce undesired overhead and code noise :(

A thing that would totally save the day is something like `map()` method on a lock guard, similar to [Ref::map](https://doc.rust-lang.org/std/cell/struct.Ref.html#method.map), which was pointed to me by @Gankro. It's sad that neither mutexes or rwlocks expose a similar function (or even plan to).

src/backend/metal/src/command.rs
kvark: an important caveat that those only show up in the "System Trace", *not* "Metal System Trace".

Gankro:
I discussed the lock::map thing with the libs team, they said it's plausible but a bit more messy than you'd expect. In particular a mapped LockGuard must be a new type (e.g. MappedLockGuard) as there are a few operations it *must not* support. Notably, temporarily unlocking with a CondVar. Consider:

```
let lock: Lock<Vec<u8>> = ...'
let val = lock.lock(),map(|v| &v[0])
cond_var(val)  // let other things use the lock for a bit
print(val)

// other thread
lock.lock().push(0); // oops realloc, invalidate the reference in the condvar
```

src/backend/metal/src/command.rs
kvark: note: this PR is getting dropped, so we need to keep an eye for those extern interfaces and bring them in separately

PR #2240, opened 2018-07-16
Faster passes and PSO bindings in Metal

kvark:
This PR has a few changes that lighten up our most heavy operations at the moment: render pass start and PSO binding. In particular, it re-uses the heap allocation properly now at the latter, ~~and the former tries to avoid waiting on a mutex (for the RP descriptor) if there is another spare available~~.

PR checklist:

### DIFF
--- a/src/backend/auxil/range_alloc.rs
+++ b/src/backend/auxil/range_alloc.rs
@@ -118,9 +118,36 @@ where
         self.free_ranges.insert(i, range);
     }
 
+    /// Returns an iterator over allocated non-empty ranges
+    pub fn allocated_ranges<'a>(&'a self) -> impl 'a + Iterator<Item = Range<T>> {
+        let first = match self.free_ranges.first() {
+            Some(Range { ref start, .. }) if *start > self.initial_range.start => Some(self.initial_range.start .. *start),
+            _ => None,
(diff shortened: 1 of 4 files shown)


kvark:
Profiling shows that the first commit doesn't actually affect any numbers. I suppose I'll remove it then, we can always reconsider.

src/backend/auxil/range_alloc.rs
```diff
@@ -118,10 +118,37 @@ where
...
+
+        let last = match self.free_ranges.last() {
+            Some(Range { end, .. }) if *end < self.initial_range.end => Some(*end .. self.initial_range.end),
+            _ => None,
+        };
+
+        let mid = self.free_ranges
+            .iter()
+            .zip(self.free_ranges.iter().skip(1))
+            .map(|(ra, rb)| ra.end .. rb.start);
```
grovesNL: I assume we have the guarantee that these are ordered (appears to be, at least)

kvark: Yeah, that's what the rest of the code assumes, at least ;)

kvark:
Thanks for reviewing!
bors r=grovesNL

PR #2241, opened 2018-07-17
Metal parking lot

kvark:
Addresses a part of #2229
~~Based on #2240~~
PR checklist:

~~Unfortunately, I see no performance advantage from switching to `parking_lot` so far. cc @Amanieu~~

After a series of unsuccessful attempts to hook up a third party library (what could possibly be easier, eh?), I got the performance numbers - we are up from 80 to almost 90, which is more than a 10% bump. Given that `parking_lot` is not a drop-in replacement for `std::sync`, I suggest we ship it by default and provide `antidote` as an optional dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ EXCLUDES:=
 FEATURES_EXTRA:=mint serialize
 FEATURES_HAL:=
 FEATURES_HAL2:=
+CHECK_SPECIAL:=-p gfx-backend-vulkan
 
 SDL2_DEST=$(HOME)/deps
 SDL2_CONFIG=$(SDL2_DEST)/usr/bin/sdl2-config
@@ -32,6 +33,7 @@ else
 	ifeq ($(UNAME_S),Darwin)
 		EXCLUDES+= --exclude gfx-backend-vulkan
 		FEATURES_HAL=metal
+		CHECK_SPECIAL=-p gfx-backend-metal --features antidote
 	endif
 endif
 
@@ -46,6 +48,7 @@ help:
 check:
 	@echo "Note: excluding \`warden\` here, since it depends on serialization"
 	cargo check --all $(EXCLUDES) --exclude gfx-warden
+	#cargo check $(CHECK_SPECIAL) # TODO: Condva API is different between antidote and parking_lot
 	cd examples && cargo check --features "gl"
 	cd examples && cargo check --features "$(FEATURES_HAL)"
 	cd examples && cargo check --features "$(FEATURES_HAL2)"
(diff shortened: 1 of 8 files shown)


Amanieu:
`feature = "partking_lot"`

You have a typo in the feature name, so parking_lot isn't actually being used.

kvark:
@Amanieu wow, thanks for pointing out this silly mistake of mine!
I tested it with proper feature name now, on top of #2240, and the fps wasn't affected still in a measurable way.

Amanieu:
Do you have instructions for a benchmark I can try? I'd like to have a look at the profiler results and possibly make tweaks to parking_lot if necessary.

grovesNL:
@kvark there's still a typo, it's `parting_lot` instead of `parking_lot`

kvark:
You must be kidding me... I need some rest.

kvark:
@Amanieu ok, now I get a ton of compile errors because all the preciouse `unwrap()` calls we fostered are no longer needed. I wish `parking_lot` had a smooth transition path, e.g. a feature that allows it to return results instead of objects.

Amanieu:
@kvark If you're just calling `unwrap` every time, you can use the [antidote](https://crates.io/crates/antidote) crate which is a wrapper around the std `Mutex`. That way you can switch between `antidote` and `parking_lot` to compare performance.

kvark:
@Amanieu looks like antidote's `Condvar` API is different from `parking_lot`, unfortunately, making it a non-drop-in replacement.
Anyhow, I profiled parking_lot for realz now, and we get more than 10% boost (and our locks are not optimal yet). So :tada: thanks for your awesome work and assistance! I want to depend on `parking_lot` by default now.

Makefile
```diff
@@ -46,6 +48,7 @@ help:
 check:
 	@echo "Note: excluding \`warden\` here, since it depends on serialization"
 	cargo check --all $(EXCLUDES) --exclude gfx-warden
+	#cargo check $(CHECK_SPECIAL) # TODO: Condva API is different between antidote and parking_lot
```
grovesNL: nit: `condvar`

kvark:
Bors r=grovesNL 
(The pr queue got too long)

kvark:
@Amanieu for the record, I think this PR is a showcase for the power of Rust and it's ecosystem. System-level fine-tuned unsafe locking primitives are wrapped into safe abstractions and hooked to the project in a matter of minutes, instantly increasing the project effectiveness and value. Our competition can't do this, so we shouldn't hesitate to take the advantage of this general ability to hook things up.

PR #2242, opened 2018-07-17
[dx11] add register remapping and disjoint CBs

fkaa:
Adds remapping of spirv bindings & sets to a flat hlsl register space.

Adds another "disjoint" buffer in the case of trying to create a buffer as a constant buffer and anything else (eg. storage buffer + uniform buffer). Also forces size of constant buffer to be aligned to 16 bytes.

Also fixes compressed texture updates, which needed to be done through `UpdateSubresource` since we cant get a writable UAV handle on compressed formats.

Fixes a few format inconsistencies for image & image view creation (eg. trying to create RTV/UAV on depth format)

Fixes #issue
PR checklist:


src/backend/dx11/src/device.rs
```diff
@@ -1040,9 +1185,16 @@ impl hal::Device<Backend> for Device {
             unimplemented!()
         };
 
+        println!("format={:?}, kind={:?}", image.format, image.kind);
```
kvark: leftover debugging? can use `info!` instead here

src/backend/dx11/src/device.rs
```diff
@@ -1151,21 +1303,29 @@ impl hal::Device<Backend> for Device {
                 }
             };
 
-            let copy_srv = self.view_image_as_shader_resource(&view).map_err(|_| device::BindError::WrongMemory)?;
+            let copy_srv = if !image.compressed {
+                Some(self.view_image_as_shader_resource(&view).map_err(|_| device::BindError::WrongMemory)?)
```
kvark: this doesn't currently work for stencil, right? let's at least assert on that

src/backend/dx11/src/device.rs
```diff
@@ -1225,12 +1385,26 @@ impl hal::Device<Backend> for Device {
...
                 .ok_or(image::ViewError::BadFormat)?,
+            range: range.clone(),
+        };
+
+        let srv_info = ViewInfo {
+            resource: image.internal.raw,
+            kind: image.kind,
+            flags: image.storage_flags,
+            view_kind,
+            format: match info.format {
```
kvark: this match (a similar one, at least) is also repeated quite a few times, let's share it

src/backend/dx11/src/device.rs
```diff
@@ -1320,50 +1494,138 @@ impl hal::Device<Backend> for Device {
...
+
+            // we decompose combined image samplers into a separate sampler and image internally
+            if binding.ty == pso::DescriptorType::CombinedImageSampler {
+                // TODO: for now we have to make combined image samplers share registers since
+                //       spirv-cross doesn't support setting the register of the sampler/texture
+                //       pair to separate values (only one `DescriptorSet` decorator)
+                let shared_reg = num_s.min(num_t);
+
+                num_s = shared_reg;
+                num_t = shared_reg;
```
kvark: wouldn't this break any following images/sampler? we just reset the register number to a technically unrelated value, and we are still going to increment this and use it afterwards?

src/backend/dx11/src/device.rs
```diff
@@ -1320,50 +1494,138 @@ impl hal::Device<Backend> for Device {
...
+
+                let sampler_reg = num_s - 1;
+                let image_reg = num_t - 1;
+
+                register_remap.push((pso::DescriptorType::Sampler, binding.binding, sampler_reg as u8, true));
```
kvark: let's put a stronger semantics (`enum` of something) here instead of a `bool` at the end

src/backend/dx11/src/internal.rs
```diff
@@ -539,34 +539,70 @@ impl Internal {
         T: IntoIterator,
         T::Item: Borrow<command::BufferImageCopy>,
     {
-        let (shader, scale_x, scale_y) = self.find_buffer_to_image_shader(dst.typed_raw_format).unwrap();
+        // NOTE: we have two separate paths for Buffer -> Image transfers. we need to special case
+        //       uploads to compressed formats through `UpdateSubresource` since we cannot get a
+        //       UAV of any compressed format.
+        if dst.block_dim.0 != 1 || dst.block_dim.1 != 1 {
```
kvark: similarly, let's use the existing `is_compressed` check

src/backend/dx11/src/internal.rs
```diff
@@ -539,34 +539,70 @@ impl Internal {
...
+                let info = copy.borrow();
 
-        unsafe {
-            context.CSSetShader(shader, ptr::null_mut(), 0);
-            context.CSSetConstantBuffers(0, 1, &self.copy_info.as_raw());
-            context.CSSetShaderResources(0, 1, [srv].as_ptr());
+                let surface_desc = dst.format.base_format().0.desc();
+                let bytes_per_texel  = surface_desc.bits as u32 / 8;
 
+                // we dont really care about non-4x4 block formats..
```
kvark: need to assert on that in code, not just comments

src/backend/dx11/src/lib.rs
```diff
@@ -1697,6 +1702,8 @@ pub struct UnboundImage {
     usage: image::Usage,
     flags: image::StorageFlags,
     bind: d3d11::D3D11_BIND_FLAG,
+    compressed: bool,
```
kvark: `format` already has this info, as I mentioned earlier, so no need to separately store it

src/backend/dx11/src/lib.rs
```diff
@@ -1840,8 +1848,12 @@ pub struct PipelineLayout {
 #[derive(Debug)]
 pub struct DescriptorSetLayout {
     bindings: Vec<PipelineBinding>,
-    offset_mapping: Vec<(u32, pso::DescriptorType)>,
-    handle_count: u32
+    register_remap: Vec<(pso::DescriptorType, u32, u8, bool)>,
+    handle_count: u32,
+    num_s: u8,
```
kvark: sounds like we need a struct with those - would also help sharing the code collecting the counters

src/backend/dx11/src/device.rs
fkaa: `shared_reg` should be using `max()` 😓

Ideally we shouldn't have to do this, but because of spirv-cross limitations we potentially end up with some wasted space, see image:
register allocation

src/backend/dx11/src/device.rs
```diff
@@ -593,10 +606,28 @@ impl hal::Device<Backend> for Device {
...
+                                    },
+                                    pso::DescriptorType::UniformBuffer | pso::DescriptorType::UniformBufferDynamic => {
+                                        c_offset
+                                    },
+                                    pso::DescriptorType::StorageTexelBuffer | pso::DescriptorType::StorageBuffer |
+                                    pso::DescriptorType::InputAttachment | pso::DescriptorType::StorageBufferDynamic |
+                                    pso::DescriptorType::StorageImage => {
+                                        u_offset
+                                    },
+                                    pso::DescriptorType::CombinedImageSampler => unreachable!()
```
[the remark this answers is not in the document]
fkaa: Since we decompose `CombinedImageSampler` into two separate descriptors we wont see this here.

PR #2245, opened 2018-07-17
More granular service locks in Metal - take 2

kvark:
Helps a bit with #2227, but not fixes it entirely yet.
PR checklist:

I'm getting burned (and wasting a bit too much time) trying to shove the zero overhead locking of depth-stencil states (which are the main blocker for performance) into the existing locking primitives... I came to conclusion that we should proceed with the simplified version of the PR, where there are still mutexes around. As the next step, we'll need to introduce a special map type that is meant to be append-only and accessed from multiple threads, which is to be used for all service pipe maps. If you have any ideas on how to approach it, please fire them up. cc @Amanieu


src/backend/metal/src/command.rs
```diff
@@ -147,6 +147,11 @@ struct RenderPipelineState {
     at_formats: SmallVec<[Option<Format>; 8]>,
 }
 
+/// The current state of a command buffer, used for two distict purposes:
+///   1. inherit resource bindings between passes
+///   2. avoid redundant state settings
+/// Note that these two usages are distinct and operate in tehnically different
```
grovesNL: nit: "technically"

src/backend/metal/src/command.rs
```diff
@@ -147,6 +147,11 @@ struct RenderPipelineState {
     at_formats: SmallVec<[Option<Format>; 8]>,
 }
 
+/// The current state of a command buffer, used for two distict purposes:
```
grovesNL: nit: "distinct"

PR #2246, opened 2018-07-18
Cut logging in release

### LINKED ISSUE #2237: Reduce debug logging overhead in release
See https://github.com/rust-lang-nursery/log/issues/112
This is a bit of an obscure source of overhead that we need to fix. It affects everything!

kvark:
Fixes #2237
PR checklist:


kvark:
@grovesNL that sounds good.
bors r=grovesNL

PR #2250, opened 2018-07-18
[mtl] Experimental concurrent map storage integration

### LINKED ISSUE #2227: Optimize locking of `ServicePipes`

kvark:
Fixes #2227
Adds dependency to `storage-map` crate for locking on Metal. If it proves to be useful, we'll likely want it to be used for other backends as well.

Dota performance changes:
  - immediate: 90 -> 92 fps
  - deferred: 92 -> 99 fps

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -3,7 +3,7 @@ use {
     BufferPtr, TexturePtr, SamplerPtr,
 };
 use {conversions as conv, native, soft, window};
-use internal::{BlitVertex, Channel, ClearKey, ClearVertex, DepthStencilStates};
+use internal::{BlitVertex, Channel, ClearKey, ClearVertex};
 use lock::Mutex;
 
 use std::borrow::Borrow;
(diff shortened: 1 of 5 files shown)


src/backend/metal/src/command.rs
```diff
@@ -1933,6 +1931,9 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         }
 
         for clear in clears {
+            let pso; // has to live at least as long as all the commands
```
grovesNL: What's the problem with leaving the `pso` declaration at the previous location?

kvark: lifetimes... not really as straightforward as I'd hoped. Basically, anything used in a command has to be declared prior to *all other* commands submitted in the same "batch".

kvark:
bors r=grovesNL

PR #2254, opened 2018-07-19
Metal command buffer fences and ignoring empty ones

### LINKED ISSUE #2252: CommandBuffer based fencing in Metal
Current fences use a combination of `Mutex` + `Condvar`, keeping a hold of a `bool` signifying the signaled state. The waiting code is a bit non-trivial and incorporates the best practices of `Condvar` usage.

I was thinking that it might be easier to instead track the last submitted command buffer:
```rust
enum FenceInner {
  Idle { signalled: bool },
  Pending(metal::CommandBuffer),
}
type Fence = RefCell<FenceInner>;
```
Now, resetting a fence just puts it into `Idle` state. Submitting with a fence makes it remember the last command buffer in flight (with a strong reference, so it's not reused by the queue in the meantime).
Waiting for 0 time is now equivalent of checking the command buffer status for completion. Checking for infinite time is equivalent to [waitUntilCompleted](https://developer.apple.com/documentation/metal/mtlcommandbuffer/1443039-waituntilcompleted?changes=_2&language=objc).

The only non-obvious part is waiting for an arbitrary amount of time. I was thinking that we may just pick a number (say, 1ms) and sleep on it then check again. I don't expect many apps to be very precise about waiting for a fence. At least, Dota doesn't use any numbers other than 0 or infinity.

The gains from this change are small but visible:
  - no more condvar magic, straightforward `antidote` use
  - less callbacks being assigned

kvark:
Fixes #2252

I'm not 100% convinced this is a good thing to fix, but had to try it anyway, so might as well file the PR :)
Please take a (critical) look.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -595,6 +595,7 @@ enum CommandSink {
         cmd_buffer: metal::CommandBuffer,
         token: Token,
         encoder_state: EncoderState,
+        num_passes: usize,
     },
     Deferred {
         passes: Vec<soft::Pass>,
@@ -703,7 +704,7 @@ impl CommandSink {
         I: Iterator<Item = soft::BlitCommand<&'a soft::Own>>,
     {
         match *self {
-            CommandSink::Immediate { ref cmd_buffer, ref mut encoder_state, .. } => {
+            CommandSink::Immediate { ref cmd_buffer, ref mut encoder_state, ref mut num_passes, .. } => {
                 let current = match mem::replace(encoder_state, EncoderState::None) {
                     EncoderState::None => None,
                     EncoderState::Render(enc) => {
@@ -716,9 +717,13 @@ impl CommandSink {
                         None
                     },
                 };
-                let encoder = current.unwrap_or_else(|| {
-                    cmd_buffer.new_blit_command_encoder().to_owned()
-                });
+                let encoder = match current {
+                    Some(enc) => enc,
+                    None => {
+                        *num_passes += 1;
+                        cmd_buffer.new_blit_command_encoder().to_owned()
+                    },
+                };
 
                 for command in commands {
                     exec_blit(&encoder, command);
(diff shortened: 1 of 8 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1938,49 +1939,43 @@ impl hal::Device<Backend> for Device {
...
-                while !*guard {
-                    fence.condvar.wait(&mut guard);
-                }
-                true
+        let inner = fence.0.borrow();
+        let cmd_buf = match *inner {
+            native::FenceInner::Idle { signaled } => return signaled,
+            native::FenceInner::Pending(ref cmd_buf) => cmd_buf,
+        };
+        if timeout_ms == !0 {
```
grovesNL: If I'm understanding correctly, I don't see how we can do these actions for a single fence in isolation, because we may be running multiple fences in parallel on the same thread (https://github.com/gfx-rs/gfx/blob/master/src/hal/src/device.rs#L528).

It seems like the `loop` would need to be around all running fences so we can start the timer from the same point. But even then with `WaitFor::Any` we want to race and avoid waiting for the longest completion or timeout, so we need to end the loop after any fence completes.

src/backend/metal/src/command.rs
```diff
@@ -1336,9 +1336,7 @@ impl RawCommandQueue<Backend> for CommandQueue {
                     trace!("\timmediate {:?} with {} passes", token, num_passes);
                     self.retained_buffers.extend(retained_buffers.drain(..));
                     self.retained_textures.extend(retained_textures.drain(..));
-                    if num_passes != 0 {
-                        cmd_buffer.commit();
-                    }
+                    cmd_buffer.commit();
```
grovesNL: Why did we have to back out the `num_passes` check? It seems reasonable unless need to attach completion handlers, but maybe we could handle that some other way

src/backend/metal/src/device.rs
kvark: Hmm, but this very call `wait_for_fence` is about a single fence, and the gluing of multiple fences lies on the shoulders of the default implementation. Are you saying that the very API is broken at the moment? I don't think the new implementation changing anything in this regard.

src/backend/metal/src/command.rs
kvark: oh, that looks like a mistake on my part

src/backend/metal/src/device.rs
grovesNL: 👍 clarified on Gitter

As mentioned, we should reconsider the `sleep` policy and timing a bit more in the future, but it's not urgent at the moment.

kvark:
Bors r=grovesNL

PR #2257, opened 2018-07-20
[mtl] add flags controlling command buffer recording logic

kvark:
PR checklist:


kvark:
This is rather harmless, merging.
bors r+

PR #2258, opened 2018-07-20
[dx11] add fences and some debugging

fkaa:
Adds a "software" Fence implementation, taken from the Metal backend. Initially I tried to implement fences using `D3D11_QUERY_EVENT`, but it turns out it would end up looking the same anyway (Queue and Device share the same ID3D11DeviceContext which has to be synchronized between threads).
 
The only difference from the Metal backend is that this isn't using a completed handler to signal the fence when the queue is done, but rather when they are scheduled (after `ExecuteCommandLists`). This should be fine since we only care about protecting our emulated objects (eg. DescriptorSet), while the driver should handle access of objects in use.

Also adds object leak checking through the `ID3D11Debug` interface.

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -7,19 +7,22 @@ use winapi::Interface;
 use winapi::shared::dxgi::{IDXGISwapChain, DXGI_SWAP_CHAIN_DESC, DXGI_SWAP_EFFECT_DISCARD};
 use winapi::shared::minwindef::{TRUE};
 use winapi::shared::{dxgiformat, dxgitype, winerror};
-use winapi::um::{d3d11, d3dcommon};
+use winapi::um::{d3d11, d3d11sdklayers, d3dcommon};
 
 use wio::com::ComPtr;
 
+use std::sync::Arc;
 use std::borrow::Borrow;
 use std::cell::RefCell;
 use std::mem;
 use std::ops::Range;
 use std::ptr;
 
+use parking_lot::{Condvar, Mutex};
+
 use {
     Backend, Buffer, BufferView, CommandPool, ComputePipeline, DescriptorPool, DescriptorSetLayout,
-    Fence, Framebuffer, GraphicsPipeline, Image, ImageView, InternalBuffer, InternalImage, Memory,
+    Fence, RawFence, Framebuffer, GraphicsPipeline, Image, ImageView, InternalBuffer, InternalImage, Memory,
     PipelineLayout, QueryPool, RenderPass, Sampler, Semaphore, ShaderModule, Surface, Swapchain,
     UnboundBuffer, UnboundImage, ViewInfo, PipelineBinding, Descriptor, MemoryHeapFlags, RegisterRemapping,
     RegisterMapping,
@@ -40,7 +43,17 @@ pub struct Device {
     pub(crate) context: ComPtr<d3d11::ID3D11DeviceContext>,
     memory_properties: hal::MemoryProperties,
     memory_heap_flags: [MemoryHeapFlags; 3],
-    pub(crate) internal: internal::Internal
+    pub(crate) internal: internal::Internal,
+}
+
+impl Drop for Device {
+    fn drop(&mut self) {
+        if let Ok(debug) = self.raw.cast::<d3d11sdklayers::ID3D11Debug>() {
+            unsafe {
+                debug.ReportLiveDeviceObjects(d3d11sdklayers::D3D11_RLDO_DETAIL);
+            }
+        }
+    }
 }
 
 unsafe impl Send for Device { }
@@ -61,7 +74,7 @@ impl Device {
                 MemoryHeapFlags::HOST_NONCOHERENT,
                 MemoryHeapFlags::HOST_COHERENT
             ],
-            internal: internal::Internal::new(&device)
+            internal: internal::Internal::new(&device),
         }
     }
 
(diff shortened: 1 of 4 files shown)


kvark:
> This should be fine since we only care about protecting our emulated objects (eg. DescriptorSet), while the driver should handle access of objects in use.

What if an application waits for a fence to be able to map something for reading/writing to the host?

src/backend/dx11/src/device.rs
```diff
@@ -7,10 +7,13 @@ use winapi::Interface;
...
 use winapi::shared::minwindef::{TRUE};
 use winapi::shared::{dxgiformat, dxgitype, winerror};
-use winapi::um::{d3d11, d3dcommon};
+use winapi::um::{d3d11, d3d11sdklayers, d3dcommon};
 
 use wio::com::ComPtr;
 
+use parking_lot::{Condvar, Mutex};
+
+use std::sync::Arc;
```
kvark: nit: could be ordered after `std::ptr`

src/backend/dx11/src/device.rs
```diff
@@ -40,7 +43,22 @@ pub struct Device {
...
     memory_properties: hal::MemoryProperties,
     memory_heap_flags: [MemoryHeapFlags; 3],
-    pub(crate) internal: internal::Internal
+    pub(crate) internal: internal::Internal,
+}
+
+impl Drop for Device {
+    fn drop(&mut self) {
+        match self.raw.cast::<d3d11sdklayers::ID3D11Debug>() {
+            Ok(debug) => {
```
kvark: nice! could be written as `Ok(debug) => unsafe {... }`

src/backend/dx11/src/device.rs
```diff
@@ -1786,27 +1804,50 @@ impl hal::Device<Backend> for Device {
...
     }
 
-    fn wait_for_fences<I>(&self, _fences: I, _wait: device::WaitFor, _timeout_ms: u32) -> bool
-    where
-        I: IntoIterator,
-        I::Item: Borrow<Fence>,
-    {
-        // TODO:
-        true
+    fn wait_for_fence(&self, fence: &Fence, timeout_ms: u32) -> bool {
```
kvark: note that Metal backend is no longer using this waiting logic

src/backend/dx11/src/lib.rs
```diff
@@ -1865,6 +1904,75 @@ pub struct DescriptorSetLayout {
     register_remap: RegisterRemapping,
 }
 
+#[derive(Debug)]
+struct CoherentBuffers {
+    // descriptor set writes containing coherent resources go into these vecs and are added to the
+    // command buffers own Vec on binding the set.
+    flush_coherent_buffers: RefCell<Vec<(*mut d3d11::ID3D11Buffer, *mut u8, Range<u64>)>>,
```
kvark: sounds like we'd better have a struct here instead of a tuple

src/backend/dx11/src/lib.rs
```diff
@@ -1865,6 +1904,75 @@ pub struct DescriptorSetLayout {
...
+    invalidate_coherent_buffers: RefCell<Vec<(*mut d3d11::ID3D11Buffer, *mut u8, Range<u64>)>>,
+}
+
+impl CoherentBuffers {
+    fn add_flush(&self, old: *mut d3d11::ID3D11Buffer, buffer: &Buffer) {
+        let new = buffer.internal.raw;
+
+        if old != new {
+            let pos = self.flush_coherent_buffers.borrow().iter().position(|&(b, _, _)| old == b);
```
kvark: let's just `borrow_mut` this once and for all

src/backend/dx11/src/lib.rs
```diff
@@ -1877,10 +1985,7 @@ pub struct DescriptorSet {
     len: usize,
     handles: *mut Descriptor,
     register_remap: RegisterRemapping,
-    // descriptor set writes containing coherent resources go into these vecs and are added to the
-    // command buffers own Vec on binding the set.
-    flush_coherent_buffers: RefCell<Vec<(*mut d3d11::ID3D11Buffer, *mut u8, Range<u64>)>>,
-    invalidate_coherent_buffers: RefCell<Vec<(*mut d3d11::ID3D11Buffer, *mut u8, Range<u64>)>>,
+    coherent_buffers: Mutex<CoherentBuffers>
```
kvark: nit: missing trailing comma

src/backend/dx11/src/lib.rs
```diff
@@ -1978,17 +2067,19 @@ impl hal::DescriptorPool<Backend> for DescriptorPool {
         let len = layout.handle_count.max(1) as _;
 
         self.allocator.allocate_range(len).map(|range| {
-			for i in range.clone() {
-				self.handles[i] = Descriptor(ptr::null_mut());
-			}
+            for i in range.clone() {
```
kvark: nit: `for h in &mut self.handles[range.clone()]`

src/backend/dx11/src/lib.rs
```diff
@@ -1978,17 +2067,19 @@ impl hal::DescriptorPool<Backend> for DescriptorPool {
...
+            }
 
             DescriptorSet {
                 offset: range.start,
                 len,
                 handles: unsafe { self.handles.as_mut_ptr().offset(range.start as _) },
                 register_remap: layout.register_remap.clone(),
-                flush_coherent_buffers: RefCell::new(Vec::new()),
-                invalidate_coherent_buffers: RefCell::new(Vec::new()),
+                coherent_buffers: Mutex::new(CoherentBuffers {
```
kvark: note: I tend to abuse `derive(Default)` for cases like this, but I don't actually recommend it

fkaa:
> What if an application waits for a fence to be able to map something for reading/writing to the host?

On calling `Map` on D3D11 resources I would think that the driver blocks until the resources in use are finished on the gpu, but since we have the rather ugly `Vec` as our mapped pointer it shouldn't matter I don't think.

fkaa:
Rebased on master

kvark:
Bors r+

PR #2262, opened 2018-07-23
Add comments to hal/src/device.rs

Michael-Lfx:
Fixes #issue
PR checklist:

### DIFF
--- a/src/hal/src/device.rs
+++ b/src/hal/src/device.rs
@@ -141,15 +141,15 @@ pub trait Device<B: Backend>: Any + Send + Sync {
     /// * `size` - Size of the allocation.
     fn allocate_memory(&self, memory_type: MemoryTypeId, size: u64) -> Result<B::Memory, OutOfMemory>;
 
-    ///
+    /// Free device memory
     fn free_memory(&self, memory: B::Memory);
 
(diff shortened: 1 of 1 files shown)


src/hal/src/device.rs
```diff
@@ -278,12 +291,17 @@ pub trait Device<B: Backend>: Any + Send + Sync {
     /// which references the framebuffer, has finished execution.
     fn destroy_framebuffer(&self, buf: B::Framebuffer);
 
+    /// Creates a new shader module object through the SPIR-V binary data
```
kvark: let's be a little more consistent: `Create` instead of `Creates`. Also - dots at the end of sentences.

src/hal/src/device.rs
```diff
@@ -385,42 +412,44 @@ pub trait Device<B: Backend>: Any + Send + Sync {
         J: IntoIterator,
         J::Item: Borrow<B::Sampler>;
 
-    ///
+    /// Destroy a descriptor set layout object
     fn destroy_descriptor_set_layout(&self, layout: B::DescriptorSetLayout);
 
-    ///
+    /// Structure specifying the parameters of a descriptor set write operation
```
kvark: Structure?

Issue #2249, opened 2018-07-18
[RFC] BAL: Graphics API framework


msiglreith:
## Introduction

We have multiple (upcoming) clients of low level gfx-rs with strong performance or validation constraints. Our current HAL interface strongly adapts the Vulkan API together with the constraints burdened upon the backends (e.g pipeline layout compatibility).

gpuweb will have a higher level API giving room for backend specific optimization without having to go through the Vulkan/HAL bottleneck. Additionally, it will require specific validation steps in contrast to other APIs. Other clients like Amethyst or WebRender could also benefit from optimization in case of higher level or small specialized interfaces.

![current](https://user-images.githubusercontent.com/7253845/42895384-12166eee-8aba-11e8-9c47-df89d3f34b7b.png)

## Concept (BAL)

Gradually lift the bottleneck imposed by HAL by adding a smaller domain specific framework for building **graphics API** on top of the lower level APIs. This framework, called **BAL**, should be seen as floating target not stabilizing anytime soon (or at all) and provide basic backend-specific and agnostic primitives for constructing higher level APIs (portability, WR, gpuweb). These primitives might be special cache structures required for only a part of the frontend APIs.

![bal](https://user-images.githubusercontent.com/7253845/42895635-c79c1606-8aba-11e8-92ed-3f60f561c520.png)

## Implementation

The transition towards the new BAL structure can happen gradually by starting to publically expose more primitives from the backends, shifting HAL slowly towards a higher level frontend API for Rust users only and allow users to hookup a basic Vulkan library (e.g Ash) to be able to work starting with a more or less fixed API.

## Graphics API implementation

Being able to quickly implement and optimize a graphics frontend API (e.g gpuweb) should be the main goal of BAL. A library author could start working with the Rust Vulkan library for implementation of the Vulkan path, automatically gaining support for the other backends due to portability initiative. Certain fast paths can be sped up writing backend specific paths on top of the primitives provided by BAL.

## Future of HAL

HAL would be mainly interesting as safer and rustic wrapper of Vulkan or moved more towards a middleground of all APIs. It should be seen as low overhead frontend but not as the only API exposed from gfx-rs. Future higher levels abstractions like gpuweb could be more appealing for certain user groups due to slimmer and simpler interface.

msiglreith:
[...]

PR #2266, opened 2018-07-24
dx12: Adopt BAL approch

msiglreith:
PoC for switching towards the BAL concept suggested in #2249 
* New `/src` structure:
```
- /bal
--- /dx12 (gfx-bal-dx12)
- /hal
--- /dx12 (gfx-hal-dx12)
--- /src (gfx-hal)
```

* `backend/dx12` moved to `hal/dx12`
* `hal/dx12` has a new dependency on `bal/dx12`

Source code from HAL will be gradually refactored into the corresponding BAL backend in order to slim down the HAL implementation and move 'unsafe' portions into BAL (ie zero cost wrappers over d3d12 structs)

Upcoming PR should be a rustfmt as the backend code doesnt seem to be formatted at all (?)


kvark:
Hmm, I'm not sure it's worth it for us to split the backends into HAL/BAL sections. It would introduce longer dependency chains for us and clients. Couldn't the backend crates just provide both? Assuming HAL implementation is mostly just wrapping around BAL primitives.

src/bal/dx12/src/copy.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+impl Image {
+    pub fn calc_subresource(&self, mip_level: u32, layer: u32, plane: u32) -> u32 {
+        mip_level + (layer * self.levels) + (plane * self.levels * self.layers)
+    }
+}
+
+#[derive(Clone)]
+pub struct Copy {
```
omni-viral: Naming conflict with `std::prelude::Copy` is undesirable.

src/bal/dx12/src/native/com.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.0
+    }
+
+    pub fn as_mut_ptr(&self) -> *mut T {
+        self.0
+    }
+}
+
+impl<T: Interface> WeakPtr<T> {
+    pub fn as_unknown(&self) -> &IUnknown {
```
omni-viral: This function must be marked `unsafe`.
Otherwise this "safe" expression `WeakPtr::from_raw(1usize as *mut u32).as_unknown()` will trigger UB.

src/bal/dx12/src/native/com.rs
```diff
@@ -0,0 +1,101 @@
...
+        debug_assert!(!self.is_null());
+        unsafe { &*(self.0 as *mut IUnknown) }
+    }
+
+    // Cast creates a new WeakPtr requiring explicit destroy call.
+    pub fn cast<U>(&self) -> D3DResult<WeakPtr<U>>
```
omni-viral: `unsafe` function

src/bal/dx12/src/native/com.rs
```diff
@@ -0,0 +1,101 @@
...
+                &U::uuidof(),
+                &mut obj.as_mut_ptr() as *mut *mut _ as *mut *mut _,
+            )
+        };
+        (obj, hr)
+    }
+
+    // Destroying one instance of the WeakPtr will invalidate all
+    // copies and clones.
+    pub fn destroy(&self) {
```
omni-viral: `unsafe` function

src/bal/dx12/src/native/com.rs
```diff
@@ -0,0 +1,101 @@
...
+        WeakPtr(self.0)
+    }
+}
+
+impl<T> Copy for WeakPtr<T> {}
+
+impl<T> Deref for WeakPtr<T> {
+    type Target = T;
+    fn deref(&self) -> &T {
+        debug_assert!(!self.is_null());
```
omni-viral: This type is `pub` so `debug_assert` doesn't seem enough.

msiglreith:
The `native` part is moved out into the d3d12-rs crate.
The transition of the current HAL backend is partially done in https://github.com/gfx-rs/gfx/pull/2374

Moving out common pieces like blitting will done in a separate PR.

PR #2267, opened 2018-07-24
[mtl] new autorelease

kvark:
Fixes the time wasted on setting up temporary autorelease pools, mostly. Although the resulting FPS doesn't appear to be affected. At least we no longer depend on core-foundation :)
cc @jrmuizel (and thanks!) for the way it looks now

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -1,5 +1,5 @@
 use {
-    AutoreleasePool, Backend, PrivateDisabilities, Shared, validate_line_width,
+    Backend, PrivateDisabilities, Shared, validate_line_width,
     BufferPtr, TexturePtr, SamplerPtr,
 };
 use {conversions as conv, native, soft, window};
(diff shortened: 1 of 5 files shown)


src/backend/metal/src/command.rs
```diff
@@ -721,7 +722,6 @@ impl Journal {
     }
 
     fn record(&self, command_buf: &metal::CommandBufferRef) {
-        let _ap = AutoreleasePool::new(); // for encoder creation
```
grovesNL: This removal seems ok but maybe we should indicate that this is handled by https://github.com/gfx-rs/gfx/pull/2267/files#diff-abc29e1de3205709a964ff5e3b0e6333R1494 instead

kvark: I don't know the best way for us to enforce the autorelease pools, tbh. Just trying to put comments doesn't protect us against a breakage during refactor, or from using a method accidentally that requires a pool but we don't have one...

kvark:
Let's keep thinking on how to enforce this nicer...
bors r=grovesNL

PR #2268, opened 2018-07-24
[mtl] fix compute resource override

kvark:
Fixes Dota on high settings.
PR checklist:


kvark:
Merging as trivial (and blocking).
bors r=common sense

PR #2274, opened 2018-07-26
Fix dynamic buffer offsets on Metal

### LINKED ISSUE #2272: Stretched character in Dota's hero selection menu
I looked at this visual glitch we get with visual settings above normal. Apparently, there is a value in one of the constant buffers, which is supposed to contain 1/size. Ours have (1/1024, 1/1024), while the actual render target size is 1280x800. Perhaps, there is a vulkan query that we don't report properly? sounds weird though, since clients should know the image sizes they supply.

kvark:
Fixes #2272

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -2924,9 +2924,9 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
 
                         if layout.content.contains(native::DescriptorContent::BUFFER) {
                             let mut buffer = data.buffers[data_offset.buffers].clone();
-                            if let Some(offset_index) = layout.dynamic_offset_index {
+                            if layout.content.contains(native::DescriptorContent::DYNAMIC_BUFFER) {
                                 if let Some((_, ref mut offset)) = buffer {
-                                    *offset += self.temp.dynamic_offsets[offset_index as usize] as u64;
+                                    *offset += self.temp.dynamic_offsets[layout.associated_data_index as usize] as u64;
                                 }
                             }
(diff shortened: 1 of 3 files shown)


msiglreith:
This fixes #2272 ?

kvark:
@msiglreith yes, thanks for noting that :)

src/backend/metal/src/command.rs
```diff
@@ -2864,62 +2864,92 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    for layout in layouts.iter() {
+                        //TODO: there is quite a bit code duplication happening below between vertex and fragment stages
+                        // I inlined everything as the most efficient code path I could see. Any abstraction from here
+                        // needs to ensure that the assembly code doesn't regress. Be my guess to give it a shot :)
+                        // The general idea is to only go through layouts once, and only fetch the data once.
+
+                        if layout.content.contains(native::DescriptorContent::SAMPLER) {
+                            let sampler = data.samplers[data_offset.samplers];
+                            if layout.stages.contains(pso::ShaderStageFlags::VERTEX) {
+                                let index = res_offset.vs.samplers;
```
grovesNL: It would probably be easier to abstract if we could access some of these values by shader stage instead of being explicitly named `vs` etc. For example, `res_offset[stage]` and `self.state.resources[stage]`. I guess the challenge is to ensure this would be compiler-friendly enough (here and other places) so it would be unrolled to roughly the same output

grovesNL:
I think it's safe to proceed with this, especially since we've confirmed it to fix the visual bug.
bors r+

PR #2276, opened 2018-07-27
[mtl] discover physical devices

kvark:
Exposes real set of physical devices.
Depends on https://github.com/gfx-rs/metal-rs/pull/63
PR checklist:


src/backend/metal/src/device.rs
```diff
@@ -192,6 +192,13 @@ impl PhysicalDevice {
 
     pub(crate) fn new(device: metal::Device) -> Self {
         let private_caps = PrivateCapabilities {
+            msl_version: if device.supports_feature_set(MTLFeatureSet::macOS_GPUFamily1_v3) {
```
grovesNL: It appears to be unrelated to the feature set if we read between the lines in the [Metal Shading Language Specification (6.4)](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf):

* ios-metal1.0
 – support the unified graphics / compute language revision 1.0 programs for iOS 8.
* ios-metal1.1
 – support the unified graphics / compute language revision 1.1 programs for iOS 9.
* ios-metal1.2
 – support the unified graphics / compute language revision 1.2 programs for iOS 10.
* ios-metal2.0
 – support the unified graphics / compute language revision 2.0 programs for iOS 11.
* ios-metal2.1
 – support the unified graphics / compute language revision 2.1 programs for iOS 12.
* osx-metal1.1
 – support the unified graphics / compute language revision 1.1 programs for macOS 10.11.
* osx-metal1.2
 – support the unified graphics / compute language revision 1.2 programs for macOS 10.12.
* osx-metal2.0
 – support the unified graphics / compute language revision 2.0 programs for macOS 10.13.
* osx-metal2.1
 – support the unified graphics / compute language revision 2.0 programs for macOS 10.14.

kvark: So basically, it's the OS version and not the feature set. hmm

kvark:
It seems farly safe for now to proceed with this incorrect shading language selection though, given that Molten is fine with it. We should definitely make this nicer afterwards!

grovesNL:
Looks like you've published metal-rs already 👍
bors r+

kvark:
Jeez, late night multitasking... I realized that the sorting was the opposite of what we need. I.e. we need the first device to have `low_power == false` and `headless == false`.
Sorry about that! Should be good now.
bors r=grovesNL

PR #2278, opened 2018-07-27
Change timeout_ms parameter in wait_for_fence(s) to nanoseconds and u64

### LINKED ISSUE #2243: Change timeout_ms parameter in wait_for_fence(s) to nanoseconds and u64
[HAL device trait](https://github.com/gfx-rs/gfx/blob/master/src/hal/src/device.rs#L510-L516) currently has the timeout parameter as meaning milliseconds and specified as u32. Vulkan (and by extension portability) requires it to be nanoseconds and u64 (see [vkWaitForFences](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkWaitForFences.html))

kocsis1david:
Fixes #2243

### Converting nanoseconds to milliseconds

This is needed by the DX12 backend. I tried to round up when converting, because it would be strange to convert 1 nanoseconds to 0 milliseconds and not wait anything. Is this correct behavior?

### Integer overflow

`u64` with nanoseconds can express much longer time period than `u32` milliseconds. I use `<u32>::max_value()` in case of overflow, but that means to wait INFINITY in win32.

### Code duplication

There is a loop in the HAL's `wait_for_fences` implementation, mostly it is the same as the metal backend's wait loop. Should I create a function for it? But I don't know where to place it.

PR checklist:

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -2333,7 +2333,7 @@ impl d::Device<B> for Device {
         let descriptor_pools = descriptor_pools.into_iter()
                                                .map(|desc| *desc.borrow())
                                                .collect::<Vec<_>>();
-        
+
         for desc in &descriptor_pools {
             match desc.ty {
                 pso::DescriptorType::Sampler => {
@@ -2778,7 +2778,7 @@ impl d::Device<B> for Device {
         });
     }
 
-    fn wait_for_fences<I>(&self, fences: I, wait: d::WaitFor, timeout_ms: u32) -> bool
+    fn wait_for_fences<I>(&self, fences: I, wait: d::WaitFor, timeout_ns: u64) -> bool
     where
         I: IntoIterator,
         I::Item: Borrow<n::Fence>,
@@ -2808,6 +2808,16 @@ impl d::Device<B> for Device {
             d::WaitFor::All => TRUE,
         };
         let hr = unsafe {
+            // This block handles overflow when converting to u32 and always rounds up
+            // The Vulkan specification allows to wait more than specified
+            let timeout_ms = {
+                if timeout_ns > (<u32>::max_value() as u64) * 1_000_000 {
+                    <u32>::max_value()
+                } else {
+                    ((timeout_ns + 999_999) / 1_000_000) as u32
+                }
+            };
+
             synchapi::WaitForMultipleObjects(fences.len() as u32, events.as_ptr(), all, timeout_ms)
         };
 
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -1942,25 +1942,30 @@ impl hal::Device<Backend> for Device {
     fn reset_fence(&self, fence: &n::Fence) {
         *fence.0.borrow_mut() = n::FenceInner::Idle { signaled: false };
     }
-    fn wait_for_fence(&self, fence: &n::Fence, mut timeout_ms: u32) -> bool {
-        debug!("wait_for_fence {:?} for {} ms", fence, timeout_ms);
+    fn wait_for_fence(&self, fence: &n::Fence, mut timeout_ns: u64) -> bool {
+        fn to_ns(duration: time::Duration) -> u64 {
+            duration.as_secs() * 1_000_000_000 + duration.subsec_nanos() as u64
+        }
+
+        debug!("wait_for_fence {:?} for {} ms", fence, timeout_ns);
         let inner = fence.0.borrow();
         let cmd_buf = match *inner {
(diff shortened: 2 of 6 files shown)


fkaa:
> I tried to round up when converting, because it would be strange to convert 1 nanoseconds to 0 milliseconds and not wait anything. Is this correct behavior?

Yup, as you mention in your comments, we're allowed (and should, in this case) to wait for longer than the specified timeout

[Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkWaitForFences.html#_parameters):
> timeout is adjusted to the closest value allowed by the implementation-dependent timeout accuracy, which **may** be substantially longer than one nanosecond, and **may** be longer than the requested period.

> `u64` with nanoseconds can express much longer time period than `u32` milliseconds. I use `<u32>::max_value()` in case of overflow, but that means to wait INFINITY in win32.

I think this is fine as well. The spec says that the function **must** return in finite time, but we're talking about an explicit request to wait for >50 days, and the previous behaviour matches this already (cc @msiglreith)

> There is a loop in the HAL's `wait_for_fences` implementation, mostly it is the same as the metal backend's wait loop. Should I create a function for it? But I don't know where to place it.

It's fine to leave it as it is

src/backend/dx12/src/device.rs
```diff
@@ -2808,6 +2808,16 @@ impl d::Device<B> for Device {
             d::WaitFor::All => TRUE,
         };
         let hr = unsafe {
+            // This block handles overflow when converting to u32 and always rounds up
+            // The Vulkan specification allows to wait more than specified
+            let timeout_ms = {
+                if timeout_ns > (<u32>::max_value() as u64) * 1_000_000 {
```
kvark: this block should be rewritten, it doesn't properly handle the overflow anyway: if `timeout_ns == <u32>::max_value() as u64 * 1_000_000`, for example, the condition will be false, but it would still overflow below.
I suggest just doing `((timeout_ns + 500_000) / 1_000_000).min(<u32>::max_value() as u64) as u32`

src/backend/metal/src/device.rs
```diff
@@ -1942,25 +1942,31 @@ impl hal::Device<Backend> for Device {
     fn reset_fence(&self, fence: &n::Fence) {
         *fence.0.borrow_mut() = n::FenceInner::Idle { signaled: false };
     }
-    fn wait_for_fence(&self, fence: &n::Fence, mut timeout_ms: u32) -> bool {
-        debug!("wait_for_fence {:?} for {} ms", fence, timeout_ms);
+    fn wait_for_fence(&self, fence: &n::Fence, mut timeout_ns: u64) -> bool {
+        use std::{time, thread};
```
kvark: why do we need to have this `use` if those types were already used here without one?

src/backend/vulkan/src/device.rs
```diff
@@ -1404,7 +1404,7 @@ impl d::Device<B> for Device {
             d::WaitFor::All => true,
         };
         let result = unsafe {
-            self.raw.0.wait_for_fences(&fences, all, timeout_ms as u64 * 1000)
```
kvark: was this code incorrect all the time? should have been `timeout_ms * 1_000_000`

src/backend/dx12/src/device.rs
kocsis1david: yeah, min is a nice idea.
but why add `500_000` instead of `999_999`?

kocsis1david: no, it's not good, because `timeout_ns + anything` could fail if timeout_ns is `!0`

kvark: on a second consideration, rounding up to highest makes more sense. Let's keep 999_999 there

src/backend/vulkan/src/device.rs
kocsis1david: it seems like that

src/backend/metal/src/device.rs
kocsis1david: it's not needed, I removed it

src/backend/dx12/src/device.rs
kocsis1david: It won't overflow when `timeout_ns == <u32>::max_value() as u64 * 1_000_000`. `(4294967295000000 + 999_999) / 1_000_000` equals `4294967295`, so it's not an overflow.

And I cannot use `min`, because it would overflow:

`((timeout_ns + 500_000) / 1_000_000).min(<u32>::max_value() as u64) as u32`

This would mean that the addition runs before the overflow check by the `min` function.

kocsis1david:
Yes, it should be one commit now.

kvark:
Thank you!
Bors r+

PR #2284, opened 2018-07-31
[dx11] add render passes and debugging

fkaa:
Re-do format conversion to be more comprehensive and less hacky

Add Bgra8 -> Buffer transfer

Add basic render passes (missing partial clear/resolves)

Add debugging using `ID3DUserDefinedAnnotation`. This functionality is behind a `#[cfg(debug_assertions)]` and is further restricted from being used on deferred contexts when RenderDoc is not used.

Separate `MemorySync` into Flush/Invalidate variants

r? @kvark 

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/dx11/shaders/copy.hlsl
+++ b/src/backend/dx11/shaders/copy.hlsl
@@ -116,6 +116,11 @@ uint Uint8x2ToUint16(uint2 data)
     return dot(min(data, 0xFF), 1 << uint2(0, 8));
 }
 
+uint4 Float4ToUint8x4(float4 data)
+{
+    return uint4(data * 255 + .5f);
+}
+
 // Buffers are always R32-aligned
 ByteAddressBuffer   BufferCopySrc : register(t0);
 RWByteAddressBuffer BufferCopyDst : register(u0);
@@ -124,6 +129,8 @@ Texture2DArray<uint4>   ImageCopySrc     : register(t0);
 RWTexture2DArray<uint>  ImageCopyDstR    : register(u0);
 RWTexture2DArray<uint2> ImageCopyDstRg   : register(u0);
 RWTexture2DArray<uint4> ImageCopyDstRgba : register(u0);
+
+Texture2DArray<float4>  ImageCopySrcBgra : register(t0);
 
 // Image<->Image copies
 [numthreads(1, 1, 1)]
@@ -323,6 +330,15 @@ void cs_copy_image2d_r8g8b8a8_buffer(uint3 dispatch_thread_id : SV_DispatchThrea
     uint3 src_idx = GetImageSrc(dispatch_thread_id);
 
     BufferCopyDst.Store(dst_idx, Uint8x4ToUint32(ImageCopySrc[src_idx]));
+}
+
+// B8G8R8A8
+[numthreads(1, 1, 1)]
+void cs_copy_image2d_b8g8r8a8_buffer(uint3 dispatch_thread_id : SV_DispatchThreadID) {
+    uint dst_idx = GetBufferDst32(dispatch_thread_id);
+    uint3 src_idx = GetImageSrc(dispatch_thread_id);
+
+    BufferCopyDst.Store(dst_idx, Uint8x4ToUint32(Float4ToUint8x4(ImageCopySrcBgra[src_idx])));
 }
 
 // R16
--- a/src/backend/dx11/src/conv.rs
+++ b/src/backend/dx11/src/conv.rs
@@ -34,120 +34,6 @@ pub fn viewable_format(format: DXGI_FORMAT) -> DXGI_FORMAT {
     }
 }
 
-pub fn typeless_format(format: DXGI_FORMAT) -> Option<(DXGI_FORMAT, DXGI_FORMAT)> {
-    match format {
-        DXGI_FORMAT_R8G8B8A8_UNORM |
-        DXGI_FORMAT_R8G8B8A8_SNORM |
-        DXGI_FORMAT_R8G8B8A8_UINT |
-        DXGI_FORMAT_R8G8B8A8_SINT |
-        DXGI_FORMAT_R8G8B8A8_UNORM_SRGB => Some((DXGI_FORMAT_R8G8B8A8_TYPELESS, DXGI_FORMAT_R8G8B8A8_UINT)),
-
-        // ?`
-        DXGI_FORMAT_B8G8R8A8_UNORM |
-        DXGI_FORMAT_B8G8R8A8_UNORM_SRGB => Some((DXGI_FORMAT_B8G8R8A8_TYPELESS, DXGI_FORMAT_B8G8R8A8_UNORM)),
-
-        DXGI_FORMAT_R8_UNORM |
-        DXGI_FORMAT_R8_SNORM |
-        DXGI_FORMAT_R8_UINT |
-        DXGI_FORMAT_R8_SINT => Some((DXGI_FORMAT_R8_TYPELESS, DXGI_FORMAT_R8_UINT)),
-
-        DXGI_FORMAT_R8G8_UNORM |
-        DXGI_FORMAT_R8G8_SNORM |
-        DXGI_FORMAT_R8G8_UINT |
-        DXGI_FORMAT_R8G8_SINT => Some((DXGI_FORMAT_R8G8_TYPELESS, DXGI_FORMAT_R8G8_UINT)),
-
-        DXGI_FORMAT_D16_UNORM |
-        DXGI_FORMAT_R16_UNORM |
-        DXGI_FORMAT_R16_SNORM |
-        DXGI_FORMAT_R16_UINT |
-        DXGI_FORMAT_R16_SINT |
-        DXGI_FORMAT_R16_FLOAT => Some((DXGI_FORMAT_R16_TYPELESS, DXGI_FORMAT_R16_UINT)),
-
-        DXGI_FORMAT_R16G16_UNORM |
-        DXGI_FORMAT_R16G16_SNORM |
-        DXGI_FORMAT_R16G16_UINT |
-        DXGI_FORMAT_R16G16_SINT |
-        DXGI_FORMAT_R16G16_FLOAT => Some((DXGI_FORMAT_R16G16_TYPELESS, DXGI_FORMAT_R16G16_UINT)),
-
-        DXGI_FORMAT_R16G16B16A16_UNORM |
-        DXGI_FORMAT_R16G16B16A16_SNORM |
-        DXGI_FORMAT_R16G16B16A16_UINT |
-        DXGI_FORMAT_R16G16B16A16_SINT |
-        DXGI_FORMAT_R16G16B16A16_FLOAT => Some((DXGI_FORMAT_R16G16B16A16_TYPELESS, DXGI_FORMAT_R16G16B16A16_UINT)),
-
-        DXGI_FORMAT_D32_FLOAT_S8X24_UINT => Some((DXGI_FORMAT_R32G8X24_TYPELESS, DXGI_FORMAT_R32_FLOAT_X8X24_TYPELESS)),
-
-        DXGI_FORMAT_D32_FLOAT |
-        DXGI_FORMAT_R32_UINT |
-        DXGI_FORMAT_R32_SINT |
-        DXGI_FORMAT_R32_FLOAT => Some((DXGI_FORMAT_R32_TYPELESS, DXGI_FORMAT_R32_UINT)),
-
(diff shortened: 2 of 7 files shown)


src/backend/dx11/shaders/copy.hlsl
```diff
@@ -116,6 +116,11 @@ uint Uint8x2ToUint16(uint2 data)
     return dot(min(data, 0xFF), 1 << uint2(0, 8));
 }
 
+uint4 Float4ToUint8x4(float4 data)
+{
+    return uint4(floor(data * 255 + .5f));
```
kvark: does it need a `floor`?

src/backend/dx11/src/conv.rs
```diff
@@ -225,6 +111,321 @@ pub fn map_format(format: Format) -> Option<DXGI_FORMAT> {
     Some(format)
 }
 
+#[derive(Debug, Clone)]
+pub struct DecomposedDxgiFormat {
+    pub typeless_format: DXGI_FORMAT,
```
kvark: whoa, this is looking very thorough!
Perhaps, we can omit `_format` suffices in the fields?

src/backend/dx11/src/conv.rs
```diff
@@ -225,6 +111,321 @@ pub fn map_format(format: Format) -> Option<DXGI_FORMAT> {
...
+    pub copy_uav_format: Option<DXGI_FORMAT>,
+    pub copy_srv_format: Option<DXGI_FORMAT>,
+}
+
+// TODO: we probably want to pass in usage flags or similar to allow for our `typeless_format`
+//       field to only contain the input format (eg. depth only rather than typeless likely
+//       improves perf since the driver doesn't need to expose internals)
+//
+// TODO: we also want aspect for determining depth/stencil
+pub fn decompose_format(format: DXGI_FORMAT) -> DecomposedDxgiFormat {
```
kvark: let's make it a constructor funftion of the target struct

src/backend/dx11/src/debug.rs
```diff
@@ -0,0 +1,101 @@
...
+mod temp {
+    use winapi::shared::minwindef::{BOOL, INT}; 
+    use winapi::um::unknwnbase::{IUnknown, IUnknownVtbl}; 
+    use winapi::um::winnt::LPCWSTR; 
+
+    RIDL!{#[uuid(0xb2daad8b, 0x03d4, 0x4dbf, 0x95, 0xeb, 0x32, 0xab, 0x4b, 0x63, 0xd0, 0xab)]
+    interface ID3DUserDefinedAnnotation(ID3DUserDefinedAnnotationVtbl):
+        IUnknown(IUnknownVtbl) {
+        fn BeginEvent(
+            Name: LPCWSTR,
```
kvark: any reason those are split into multiple lines?

src/backend/dx11/src/device.rs
```diff
@@ -1341,21 +1348,44 @@ impl hal::Device<Backend> for Device {
...
+                        view_kind: image::ViewKind::D2,
+                        format: decomposed.dsv_format.unwrap(),
+                        range: image::SubresourceRange {
+                            aspects: format::Aspects::COLOR,
+                            levels: mip..(mip + 1),
+                            layers: layer..(layer + 1)
+                        }
+                    };
+
+                    depth_stencil_views.push(self.view_image_as_depth_stencil(&view).map_err(|_| device::BindError::WrongMemory)?);
```
kvark: nit: let's not stretch this line as long

src/backend/dx11/src/device.rs
```diff
@@ -1685,6 +1715,9 @@ impl hal::Device<Backend> for Device {
                             pso::DescriptorType::StorageImage => {
                                 unsafe { *handle = Descriptor(image.uav_handle.clone().unwrap().as_raw() as *mut _); }
                             },
+							pso::DescriptorType::InputAttachment => {
```
kvark: indentation is broken here

src/backend/dx11/src/device.rs
```diff
@@ -1614,7 +1644,7 @@ impl hal::Device<Backend> for Device {
                 StorageTexelBuffer | StorageBuffer |
                 InputAttachment | StorageBufferDynamic |
                 StorageImage => {
-                    binding.handle_offset = num_s + num_t + num_u + u;
+                    binding.handle_offset = num_s + num_t + num_c + u;
```
kvark: generally, a small fix like this would go as a separate commit, to have this one strictly as a feature commit
I don't mind mixing it at this point, when things are still in flux

src/backend/dx11/src/device.rs
```diff
@@ -1764,12 +1797,15 @@ impl hal::Device<Backend> for Device {
...
     {
+        let _scope = debug::DebugScope::with_name(&self.context, &format!("FlushMappedRanges"));
 
         // go through every range we wrote to
         for range in ranges.into_iter() {
             let &(memory, ref range) = range.borrow();
             let range = memory.resolve(range);
 
+
+            let _scope = debug::DebugScope::with_name(&self.context, &format!("Range({:?})", range));
```
kvark: Unfortunate thing is that the string is constructed even if debug scopes are disabled. Sounds like the creation of a scope needs to be a macro itself

src/backend/dx11/src/device.rs
```diff
@@ -2035,11 +2085,10 @@ impl hal::Device<Backend> for Device {
                 format: config.color_format,
                 storage_flags: image::StorageFlags::empty(),
                 // NOTE: not the actual format of the backbuffer(s)
-                typed_raw_format: dxgiformat::DXGI_FORMAT_UNKNOWN,
-                dxgi_format: format,
+                decomposed_format: decomposed.clone(),
                 num_levels: 1,
                 num_mips: 1,
-                internal
+                internal: internal
```
kvark: uh?

src/backend/dx11/src/internal.rs
```diff
@@ -591,7 +606,14 @@ impl Internal {
                     let info = copy.borrow();
                     self.update_buffer_image(context, &info);
 
-                    let uav = dst.get_uav(info.image_layers.level, 0).unwrap().as_raw();
+                    debug::debug_marker(context, &format!("{:?}", info));
+
+                    // TODO: multiple layers? do we introduce a stride and do multiple dispatch
```
kvark: sounds reasonable

src/backend/dx11/src/lib.rs
```diff
@@ -268,11 +270,11 @@ impl hal::Instance for Instance {
                         heap_index: 0,
                     },
                     hal::MemoryType {
-                        properties: Properties::CPU_VISIBLE,
+                        properties: Properties::CPU_VISIBLE | Properties::CPU_CACHED,
```
kvark: we expose no uncached CPU memory any more?

src/backend/dx11/src/lib.rs
```diff
@@ -723,6 +741,86 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
 
 }
 
+#[derive(Derivative, Clone)]
+#[derivative(Debug)]
+struct AttachmentClear {
+    subpass_id: Option<pass::SubpassId>,
+    #[derivative(Debug="ignore")]
+    value: Option<command::ClearValueRaw>,
```
kvark: I think having `Debug` implemented for `ClearValueRaw` (manually) would be handy for other backends and even users

src/backend/dx11/src/lib.rs
```diff
@@ -997,19 +1100,50 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
...
+                            self.context.ClearRenderTargetView(
+                                image.get_rtv(level, layer).unwrap().as_raw(),
+                                &color.float32
+                            );
+                        }
+                    }
+                }
+            }
+
+            // TODO: clear foreach level, layer on DSVs
```
kvark: aren't you iterating the levels/layers already?

src/backend/dx11/src/lib.rs
```diff
@@ -1196,32 +1337,32 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
...
-                        host_memory: sync.host_ptr,
-                        sync_range: sync.range.clone(),
-
-                        buffer: sync.device_buffer
-                    });
+            {
+                let coherent_buffers = set.coherent_buffers.lock();
+                for sync in coherent_buffers.flush_coherent_buffers.borrow().iter() {
+                    // TODO: merge sync range if a flush already exists
+                    if self.flush_coherent_memory.iter().position(|m| m.buffer == sync.device_buffer).is_none() {
```
kvark: `iter().any(...)`

src/backend/dx11/src/lib.rs
```diff
@@ -1255,32 +1398,31 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
...
-                    self.flush_coherent_memory.push(MemorySync {
-                        working_buffer: None,
-                        working_buffer_size: 0,
-
-                        host_memory: sync.host_ptr,
-                        sync_range: sync.range.clone(),
-
-                        buffer: sync.device_buffer
-                    });
+            {
```
kvark: is there a point in this scope now?

src/backend/dx11/src/lib.rs
```diff
@@ -1454,9 +1596,23 @@ bitflags! {
...
+enum SyncRange {
+    Whole,
+    Partial(Range<u64>)
+}
+
+#[derive(Debug, Clone)]
+pub struct MemoryFlush {
+    host_memory: *mut u8,
+    sync_range: SyncRange,
+
```
kvark: nit: why empty line?

src/backend/dx11/src/lib.rs
```diff
@@ -1615,33 +1791,75 @@ impl Memory {
...
+                //      the whole range (TODO: allow for partial updates)
+                //
+                //   3. the general case, without any `UNIFORM` usage has no restrictions on
+                //      partial updates, so we upload the specified range
+                //
+                if buffer.usage.contains(Usage::UNIFORM) && buffer.usage != Usage::UNIFORM {
+                    MemoryFlush {
+                        host_memory: unsafe { ptr.offset(buffer_range.start as _) },
+                        sync_range: SyncRange::Whole,
+
```
kvark: weird newlines

src/backend/dx11/src/lib.rs
```diff
@@ -1691,9 +1909,31 @@ impl ::std::fmt::Debug for ShaderModule {
 unsafe impl Send for ShaderModule { }
 unsafe impl Sync for ShaderModule { }
 
-#[derive(Debug)]
-pub struct RenderPass;
-#[derive(Debug)]
+#[derive(Debug, Clone)]
+pub struct SubpassDesc {
+    pub(crate) color_attachments: Vec<pass::AttachmentRef>,
```
kvark: why not `pub`?

src/backend/dx11/src/device.rs
fkaa: I'd hope that the compiler could optimize it out, but haven't actually checkedi if that's the case. Do you mean a macro that simply does `#[cfg(...)]` on the function call?

src/backend/dx11/src/debug.rs
fkaa: Taken from winapi, which has some unorthodox formatting. Hopefully there'll be a new winapi release around next week

src/backend/dx11/src/lib.rs
fkaa: Agree. How would it handle [the union](https://github.com/gfx-rs/gfx/blob/master/src/hal/src/command/raw.rs#L45)? just print out all the variants?

src/backend/dx11/src/lib.rs
fkaa: shorter scope one the mutex lock

src/backend/dx11/src/device.rs
kvark: I mean a macro like `debug_scope!("tralalal {:?}", something)`

src/backend/dx11/src/lib.rs
kvark: Just print out the `u32` one

src/backend/dx11/src/lib.rs
fkaa: yup, heat must be getting to me 😓

src/backend/dx11/src/lib.rs
fkaa: Required for WR, not to mention that reads are cached already since we copy over from gpu memory to our "mapped" Vec<u8>

PR #2285, opened 2018-08-01
Improve Descriptor-related docs and make it more obvious what should be in the DescriptorSetWrite descriptors field

termhn:
I believe I've been accurate in describing these, let me know if not. I also changed DescriptorSetWrite in a way that makes more sense for people looking through docs to figure out how it's supposed to be used, and I don't think it should have broken anything.

PR checklist:

### DIFF
--- a/src/hal/src/pso/descriptor.rs
+++ b/src/hal/src/pso/descriptor.rs
@@ -1,9 +1,20 @@
 //! Descriptor sets and layouts.
-//! A descriptor is an object that describes the connection between a resource, such as
+//! 
+//! A [`Descriptor`] is an object that describes the connection between a resource, such as
 //! an `Image` or `Buffer`, and a variable in a shader. Descriptors are organized into
-//! sets, each of which contains multiple descriptors that are bound and unbound to
-//! shaders as a single unit. Each descriptor set may contain descriptors to multiple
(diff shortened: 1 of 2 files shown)


src/hal/src/pso/descriptor.rs
```diff
@@ -45,15 +45,18 @@ pub enum DescriptorType {
     UniformBuffer = 6,
     /// Read-Write, structured buffer.
     StorageBuffer = 7,
-    ///
+    /// A uniform buffer that can be indexed into when binding its associated descriptor
```
kvark: I wouldn't say "indexed into". It's a buffer that the user can has to provide an offset for on each binding, allowing them to use different parts of the buffer for different draw calls with the same descriptor set

src/hal/src/pso/descriptor.rs
```diff
@@ -159,20 +167,24 @@ pub trait DescriptorPool<B: Backend>: Send + Sync + fmt::Debug {
     fn reset(&mut self);
 }
 
-/// DOC TODO
+/// Writes information about how and where to bind one or more specific descriptors to
```
kvark: It writes the actual descriptors, not the information about how/where

kvark:
Thank you for addressing the comments! Would you mind squashing the commits before we go?

PR #2286, opened 2018-08-01
Small fix/doc improvement for colour uniform example

termhn:
Currently the colour uniform example produces 
```
ERROR gfx_backend_vulkan > [Validation]  [ VUID-VkMappedMemoryRange-size-01390 ] Object: 0x6 (Type = 8) | vkFlushMappedMemoryRanges: Size in pMemRanges[0] is 0x84, which is not a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize (0x40). The spec valid usage text states 'If size is not equal to VK_WHOLE_SIZE, size must either be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize, or offset plus size must equal the size of memory.' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkMappedMemoryRange-size-01390)
```
when creating and updating buffers because it is using calculated sizing instead of received buffer size from memory requirements for creating memory mapped ranges.

In addition, added a note that using CPU visible buffers is not scalable for most uses and especially potentially large/unchanging data like vertex buffers.

PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -781,6 +781,7 @@ impl<B: Backend> BufferState<B> {
     {
         let memory: B::Memory;
         let buffer: B::Buffer;
+        let size: u64;
 
         let stride = size_of::<T>() as u64;
         let upload_size = data_source.len() as u64 * stride;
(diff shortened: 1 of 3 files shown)


kvark:
bors delegate=ZeGentzy

kvark:
@ZeGentzy want to review it? :)

examples/colour-uniform/main.rs
```diff
@@ -818,7 +825,7 @@ impl<B: Backend> BufferState<B> {
             memory: Some(memory),
             buffer: Some(buffer),
             device: device_ptr,
-            size: upload_size,
+            size: size,
```
kvark: nit: just `size,`

ZeGentzy:
@termhn Functions on my computer and changes look good I guess. Can you explain to me why these changes are necessary for colour-uniform but not quad? I'd think they'd be necessary for both, if they are can you add them to quad too?

termhn:
@ZeGentzy I hadn't checked on quad, I'll do that now, but I suspect it needs the same change as well if it was originally how colour-uniform was laid out.

termhn:
Yep, same thing was true of both `quad` and `compute` examples, fixed there as well.

ZeGentzy:
Thanks, I suspected so, @termhn. Colour uniform was originally written to be for the most part functionally identical.

bors r+

PR #2287, opened 2018-08-02
[mtl] rewrite handling of dynamic offsets and immutable samplers

kvark:
Fixes dynamic offsets (once again).
I accidentally noticed that piece of Vulkan spec:
> Values are taken from pDynamicOffsets in an order such that all entries for set N come before set N+1; **within a set, entries are ordered by the binding numbers in the descriptor set layouts**; and within a binding array, elements are in order. dynamicOffsetCount must equal the total number of dynamic descriptors in the sets being bound.

My previous code was written within an assumption that the user-given order of things define how both immutable samplers and dynamic buffers are handled. Apparently, the latter are always handled in the sorted order, which we already do (yay!), so with this (and a bit of extra reordering of immutable samplers) we are able to get rid of the associated data index (of descriptor layouts) and the corresponding indirection. This should make things simpler and even a tiny bit faster ;)

Note that the previous code also had another bug of not tracking the count of dynamic offsets per descriptor set, so we used the same set of offsets for all the descriptors.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -177,7 +177,6 @@ unsafe impl Sync for CommandBuffer {}
 struct Temp {
     clear_vertices: Vec<ClearVertex>,
     blit_vertices: FastHashMap<(Aspects, Level), Vec<BlitVertex>>,
-    dynamic_offsets: Vec<com::DescriptorSetOffset>,
 }
 
 #[derive(Clone)]
@@ -1706,7 +1705,6 @@ impl pool::RawCommandPool<Backend> for CommandPool {
             temp: Temp {
                 clear_vertices: Vec::new(),
                 blit_vertices: FastHashMap::default(),
-                dynamic_offsets: Vec::new(),
             },
         }).collect();
 
@@ -2877,11 +2875,10 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         J: IntoIterator,
         J::Item: Borrow<com::DescriptorSetOffset>,
     {
-        self.temp.dynamic_offsets.clear();
-        self.temp.dynamic_offsets.extend(dynamic_offsets.into_iter().map(|off| *off.borrow()));
         self.state.resources_vs.pre_allocate(&pipe_layout.total.vs);
         self.state.resources_ps.pre_allocate(&pipe_layout.total.ps);
 
+        let mut dynamic_offset_iter = dynamic_offsets.into_iter();
         let mut inner = self.inner.borrow_mut();
         let mut pre = inner.sink().pre_render();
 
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/command.rs
```diff
@@ -3088,6 +3084,8 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                                 }
                                 target_offset.buffers += 1;
                             }
+                        } else if layout.content.contains(native::DescriptorContent::DYNAMIC_BUFFER) {
+                            dynamic_offset_iter.next();
```
grovesNL: What's the reason we don't increment `*offset` in this case too? i.e. in comparison to `bind_graphics_descriptor_sets`, where increase `offset` regardless of the stage flags

src/backend/metal/src/device.rs
```diff
@@ -1354,43 +1354,48 @@ impl hal::Device<Backend> for Device {
...
                 let slb = set_layout_binding.borrow();
                 let mut content = native::DescriptorContent::from(slb.ty);
                 if slb.immutable_samplers {
                     content |= native::DescriptorContent::IMMUTABLE_SAMPLER;
+                    tmp_samplers.extend(immutable_sampler_iter
+                        .by_ref()
+                        .take(slb.count)
+                        .enumerate()
+                        .map(|(array_index, sm)| TempSampler {
+                            sampler: sm.borrow().0.clone(),
```
grovesNL: Alternatively we could create the key tuple here immediately instead mapping through the temporary struct – `((binding, array_index), sampler)`

src/backend/metal/src/command.rs
kvark: There are 2 offsets in play:
  - data offset that points to our actual descriptors stored by the pool. This is bumped regardless of the stage, just a line below
  - target offset, which is the Metal binding slot. It's only increased for a stage that a descriptor binds to, and it happens the same way in graphics/compute

This branch you highlighted is only suspicious at first because of the changed order of conditions:
  - graphics goes for each descriptor, then for each sub-type, then for each stage
  - compute goes for each descriptor, then for compute stage, then for each sub-type

So I added this little branch to advance the dynamic offsets properly... maybe worth rewriting it to be totally like in graphics, will just going to be a tiny bit slower since there is only one stage to check for.

src/backend/metal/src/device.rs
kvark: Yes, that is an option. I don't mind either way :)

kvark:
I think the concerns are addressed. Follow-up is possible but not required, let's proceed.
bors r=grovesNL

src/backend/metal/src/command.rs
grovesNL: Thanks, I think I understand this part. I'll attempt to rephrase, my confusion is:
- In `bind_graphics_descriptor_sets` we do `*offset += *dynamic_offset_iter.next().unwrap().borrow() as u64;` for the current `layout`, regardless of whether or not we have vertex or fragment flags
- In `bind_compute_descriptor_sets` we do `*offset += *dynamic_offset_iter.next().unwrap().borrow() as u64;` for the current `layout` only if we have a compute flag, otherwise `dynamic_offset_iter.next();`

So I'm trying to understand why the buffer's offset is not updated in the latter compute case, because it seems inconsistent – this all seems to occur in the `layouts` loop, so I thought the logic would match.

kvark: Ah, I see your confusion better now. Thanks for explanation!
The offset only needs to be modified if we actually use it (to write into our resource state, or to issue a command). In `bind_compute_descriptor_sets` if there is no `COMPUTE` flag, we don't use that offset for anything, and thus we just skip it.

kvark:
I think your confusion is about what offset is being modified. It's not the one in descriptor pool, it's a local copy of it. So if no stages end up using it - no problem.

grovesNL:
Yeah, that was exactly it – not sure how I missed it. This makes sense now, thanks!

PR #2288, opened 2018-08-02
[mtl] Refined passes

kvark:
Prevents rebinding of all the state and resources in case of consecutive compute dispatches.
~~Also, tries to make a list of commands for graphics pipeline setup and pass it as one big thing (instead of calling `pre.issue` for each individual one). This is piggy-backing on the `issue_many` function of the base commit here.~~ Edit: removed this commit for causing a performance regression.

Changes the sink API to be more clear about how the passes are started/ended, it's now the same across compute/render and distinct from blits.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -739,7 +739,7 @@ enum PassDoor<'a> {
 /// A helper temporary object that consumes state-setting commands only
 /// applicable to a render pass currently encoded.
 enum PreRender<'a> {
-    Immediate(&'a metal::RenderCommandEncoder),
+    Immediate(&'a metal::RenderCommandEncoderRef),
     Deferred(&'a mut Vec<soft::RenderCommand<soft::Own>>),
     Void,
 }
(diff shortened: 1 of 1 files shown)


kvark:
hmm, I detect a performance regression here, investigating the cause now...

kvark:
Good news - the regression is solely caused by the extra commit here, so I dropped it. It's an interesting piece that's not obvious to me: so building one gigantic chain of commands and recording them in one go is slower than calling `pre.issue()` for each individual one, contrary to my intuition. The news is good because it allows us to improve the performance expectations and revise all the places where we are currently using the long command chains.

src/backend/metal/src/command.rs
```diff
@@ -759,12 +759,29 @@ impl<'a> PreRender<'a> {
             PreRender::Void => (),
         }
     }
+
+    fn issue_many<'b, I>(&mut self, commands: I)
```
grovesNL: Not really important for now, but it seems like we should be able to collapse these functions by being generic over the command type

src/backend/metal/src/command.rs
```diff
@@ -921,6 +985,21 @@ impl CommandSink {
         }
     }
 
+    fn quick_compute<'a, I>(&mut self, label: &str, commands: I)
+    where
+        I: Iterator<Item = soft::ComputeCommand<&'a soft::Own>>
+    {
+        {
+            let (mut pre, switch) = self.switch_compute();
+            assert!(switch);
```
grovesNL: Why is this assertion guaranteed? It seems valid to have an encoder left open but call `{fill|copy}_buffer`, although I'm probably missing something here

kvark: good point! my assumption was wrong here

kvark:
bors=grovesNL

kvark:
bors r=grovesNL

PR #2300, opened 2018-08-08
Flattened descriptor sets on Metal

### LINKED ISSUE #2295: Starting a render pass is too slow on Metal
Takes about 6% of our library's time. Presumably, the main cost comes from the need to reset all the states when we begin a pass.

What we can do is use the `set_vertex_buffers` instead of `set_vertex_buffer` and such to reset all the states at once instead of filtering them and producing separate commands for each single binding. Those "batch" commands need to be added to metal-rs first, and we need to figure out a way to use them.

kvark:
Fixes #2295
Based on #2298

This gives us another 2.5% performance on the benchmark. ~~However, it is really going to shine when we start binding the descriptors in bulk (hence WIP here).~~ Edit: actually, not. I'll get that shining PR separately.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -777,6 +777,13 @@ enum PreCompute<'a> {
 }
 
 impl<'a> PreCompute<'a> {
+    fn is_void(&self) -> bool {
+        match *self {
+            PreCompute::Void => true,
+            _ => false,
+        }
+    }
+
     fn issue<'b>(&mut self, command: soft::ComputeCommand<&'b soft::Own>) {
         match *self {
             PreCompute::Immediate(encoder) => exec_compute(encoder, command),
@@ -2940,7 +2947,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         self.state.resources_vs.pre_allocate(&pipe_layout.total.vs);
         self.state.resources_ps.pre_allocate(&pipe_layout.total.ps);
 
-        let mut _dynamic_offset_iter = dynamic_offsets.into_iter();
+        let mut dynamic_offset_iter = dynamic_offsets.into_iter();
         let mut inner = self.inner.borrow_mut();
         let mut pre = inner.sink().pre_render();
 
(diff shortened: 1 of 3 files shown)


kvark:
Here is what this is about. We have descriptors following this route:
  1. `Device::create_descriptor_set_layout`
  2. `Device::create_pipeline_layout`
  3. `DescriptorPool::allocate_set`
  4. `Device::write_descriptor_sets`
  5. `CommandBuffer::bind_graphics_descriptor_sets`

Originally, we were keeping a hold of the set layout up until the last stage. This caused issues with spilling on writes (4) and wasn't super efficient on the binding side (5). Then, semi-recently, we started processing the descriptor layouts and sorting them in the binding order. This made the writes (4) trivial and improved the binding a bit. But essentially, the split of descriptor data into shader stages was still happening at binding (5).

What this PR does is moving the stage split earlier - to the allocation and writes (2-4). This makes the size of descriptor pools a little bit bigger, since they can hold the same descriptor multiple times (one per shader stage). It also makes the writes (4) a tiny bit slower, since it's doing one to many writes now. But the benefit is that the binding (5) is straightforward: we now have just arrays of linear data that need to be mapped directly to Metal resource space, no extra work involved.

This trade-off makes sense because we expect the binding to occur more often than the earlier stages. And it does show the benefit on our benchmarks. It also opens the way for us to bind multiple descriptors at once (resource bundling), if we find it useful. There is a bit of an annoyance about handling the dynamic buffer offsets (yes, those again...) but this looks nicer with resource bundles.

src/backend/metal/src/command.rs
```diff
@@ -2944,100 +2951,65 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                            let out = &mut cache.textures[index as usize];
+                            if *out != texture {
+                                *out = texture;
+                                pre.issue(soft::RenderCommand::BindTexture { stage, index, texture });
                             }
                         }
+                        for (index, &buffer) in (res_offset.buffers .. )
+                            .zip(&data.buffers[pool_offset.buffers.start as usize .. pool_offset.buffers.end as usize])
+                        {
+                            cache.buffers[index as usize] = buffer;
```
grovesNL: Is there any risk of `cache.buffers` becoming stale when `pre.is_void()` occurs (i.e. we don't submit `BindBuffer`)? What's the reason only `BindBuffer` is optional in that case?

grovesNL: So wouldn't `if !pre.is_void() { pre.issue(soft::RenderCommand::BindSampler { stage, index, sampler }); }` be valid, for example?

kvark:
It would be valid but useless, since issue() checks for void anyway

src/backend/metal/src/command.rs
kvark: all of the `is_void` checks we have are solely for optimization, they aren't functional. It's just to spare some work done for producing the commands that are passed to `pre.issue`

PR #2303, opened 2018-08-09
Metal resource bundles

### LINKED ISSUE #2308: Bind all descriptor sets at once in Metal
Currently, we issue resource bundle binding per descriptor set, coming from an expectation that descriptor sets are fairly big. Since we have a contiguous range of descriptor sets, and the pipeline layout assigns the resource indices sequentially, we can instead just bundle all the sets at once. This should help with small descriptor sets.

kvark:
Based on #2300 
Fixes #2308
Interestingly, I don't see this being an improvement on a dual core machine, despite the fact we are issuing less Metal commands. Not sure why this is, any ideas are welcome.

PR checklist:

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -24,7 +24,7 @@ bitflags = "1.0"
 log = { version = "0.4", features = ["release_max_level_error"] }
 winit = { version = "0.16", optional = true }
 dispatch = { version = "0.1", optional = true }
-metal = "0.11"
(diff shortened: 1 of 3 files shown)


kvark:
Actually, this is not too bad. Here are the numbers for baseline/flat-descriptors/bundles
  - Intel: no difference observed
  - AMD: 58.6 / 58.9 / 59.2
  - NV: 41.4 / 41.4 / 43.1

Looks like a win to me!

PR #2311, opened 2018-08-11
[WIP] Take attachment indices in begin_render_pass

### LINKED ISSUE #2248: Fortify the clear color iterator
This is another unsound aspect of today's HAL that would be great to fix.

In `begin_render_pass` we receive an [iterator over clear values](https://github.com/gfx-rs/gfx/blob/9162cd25dfa2e58f3ee00241dbd4902cf21cd6a7/src/hal/src/command/raw.rs#L284). It's semantics currently matches Vulkan - a value per attachment, but not necessarily enough for all attachment, and some values are ignored. This is rather sad from the API standpoint, and we should fix it.

Perhaps, we could say that the iterator has to only contain the values for *cleared* attachments? Given that framebuffer/renderpass creation is rare, we could have the portability layer building a mask of cleared attachments, and then filter the given clear value slice based on that. Shouldn't introduce any visible overhead.

kocsis1david:
Fixes #2248 

I would like to get feedback for this change if it's a good direction.
`begin_render_pass` became a bit verbose. Also I couldn't sort the attachment array without `Vec`

PR checklist:

### DIFF
--- a/src/hal/src/command/graphics.rs
+++ b/src/hal/src/command/graphics.rs
@@ -12,6 +12,7 @@ use super::{
     Shot, Level, Primary,
     ClearColorRaw, ClearDepthStencilRaw, ClearValueRaw, DescriptorSetOffset,
 };
+use pass::AttachmentId;
 
 
 /// A universal clear color supporting integer formats
@@ -314,8 +315,7 @@ impl<'a, B: Backend, C: Supports<Graphics>, S: Shot> CommandBuffer<'a, B, C, S, 
         clear_values: T,
     ) -> RenderPassInlineEncoder<B, Primary>
     where
-        T: IntoIterator,
-        T::Item: Borrow<ClearValue>,
+        T: Iterator<Item=(AttachmentId, ClearValue)>
     {
         RenderPassInlineEncoder::new(self, render_pass, frame_buffer, render_area, clear_values)
     }
@@ -329,8 +329,7 @@ impl<'a, B: Backend, C: Supports<Graphics>, S: Shot> CommandBuffer<'a, B, C, S, 
         clear_values: T,
     ) -> RenderPassSecondaryEncoder<B>
     where
-        T: IntoIterator,
-        T::Item: Borrow<ClearValue>,
+        T: Iterator<Item=(AttachmentId, ClearValue)>
     {
         RenderPassSecondaryEncoder::new(self, render_pass, frame_buffer, render_area, clear_values)
     }
--- a/src/hal/src/command/render_pass.rs
+++ b/src/hal/src/command/render_pass.rs
@@ -1,10 +1,12 @@
 use std::borrow::Borrow;
 use std::ops::{Range, Deref, DerefMut};
 use std::marker::PhantomData;
+use std::vec::Vec;
 
 use {buffer, pso};
 use {Backend, DrawCount, IndexCount, InstanceCount, VertexCount, VertexOffset};
 use queue::{Supports, Graphics};
+use pass::AttachmentId;
 use super::{
     AttachmentClear, ClearValue, ClearValueRaw, CommandBuffer, RawCommandBuffer,
     Shot, Level, Primary, Secondary, Submittable, Submit, DescriptorSetOffset,
@@ -173,18 +175,13 @@ impl<'a, B: Backend, L: Level> RenderPassInlineEncoder<'a, B, L> {
     ) -> Self
     where
         C: Supports<Graphics>,
-        T: IntoIterator,
-        T::Item: Borrow<ClearValue>,
+        T: Iterator<Item=(AttachmentId, ClearValue)>
     {
-        let clear_values = clear_values
-            .into_iter()
-            .map(|cv| ClearValueRaw::from(*cv.borrow()));
-
         cmd_buffer.raw.begin_render_pass(
             render_pass,
             frame_buffer,
             render_area,
-            clear_values,
+            clear_values.map(|(id, value)| (id, ClearValueRaw::from(value))),
             SubpassContents::Inline,
         );
 
@@ -250,18 +247,13 @@ impl<'a, B: Backend> RenderPassSecondaryEncoder<'a, B> {
     ) -> Self
     where
         C: Supports<Graphics>,
-        T: IntoIterator,
-        T::Item: Borrow<ClearValue>,
+        T: Iterator<Item=(AttachmentId, ClearValue)>
     {
-        let clear_values = clear_values
-            .into_iter()
-            .map(|cv| ClearValueRaw::from(*cv.borrow()));
-
         cmd_buffer.raw.begin_render_pass(
             render_pass,
             frame_buffer,
             render_area,
-            clear_values,
+            clear_values.map(|(id, value)| (id, ClearValueRaw::from(value))),
             SubpassContents::SecondaryBuffers,
         );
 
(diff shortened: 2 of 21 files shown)


src/hal/src/command/render_pass.rs
```diff
@@ -1,12 +1,15 @@
 use std::borrow::Borrow;
 use std::ops::{Range, Deref, DerefMut};
 use std::marker::PhantomData;
+use std::iter;
+use std::iter::Iterator;
```
kvark: shouldn't be needed

src/hal/src/command/graphics.rs
```diff
@@ -315,7 +326,7 @@ impl<'a, B: Backend, C: Supports<Graphics>, S: Shot> CommandBuffer<'a, B, C, S,
     ) -> RenderPassInlineEncoder<B, Primary>
     where
         T: IntoIterator,
-        T::Item: Borrow<ClearValue>,
+        T::Item: Borrow<BeginRenderPassClear>,
```
kvark: do we really need the `Borrow` here? I would prefer `T: Iterator<Item = (AttachmentId, ClearValue)>` instead

src/hal/src/command/render_pass.rs
```diff
@@ -156,6 +159,34 @@ impl<'a, B: Backend> RenderSubpassCommon<'a, B> {
     // TODO: begin/end query
 }
 
+fn convert_begin_render_pass_clear_values<T>(clear_values: T) -> impl Iterator<Item=ClearValueRaw>
```
kvark: This is unfortunate. It would be nice to not have this piece of logic between typed gfx-hal and raw one

src/hal/src/command/graphics.rs
kocsis1david: I don't think it's needed, but every function that needs an iterator has it, a trait bound of Borrow inside a IntoIterator. It seemed to me that this is a pattern we need to follow.

src/hal/src/command/render_pass.rs
kocsis1david: yeah, maybe inside each backend? As I see backends enumerate over clear values anyways.

src/hal/src/command/graphics.rs
kvark: `Borrow` is needed for immutable stuff, e.g. if you need a list of `B::Buffer`, it's not reasonable to accept an iterator over them, since the user has them wrapped into something at all times. For the clear colors, they are relatively light and copy-able, so there is no need for borrow.

src/hal/src/command/render_pass.rs
kvark: Can it go into https://github.com/gfx-rs/gfx/tree/master/src/backend/auxil ?

kocsis1david: yes, but if it goes there, each backend calls it, is that right?

src/hal/src/command/render_pass.rs
kocsis1david: yes. will be removed

src/hal/src/command/graphics.rs
kocsis1david: agree, it will be `T: Iterator<Item = (AttachmentId, ClearValue)>`

src/hal/src/command/render_pass.rs
kvark: yes. Also, we should make it work on an existing `Vec` rather than allocating one

kocsis1david: I'm not sure how to make it work on an existing `Vec`. I have to sort the clear values which can be only done on the incoming clear values, not on the existing `Vec`s. Currently it uses `SmallVec`.

kocsis1david:
I tried to build gfx-portability, and the only thing that needs to be changed there is to call `.enumerate()` when calling `begin_render_pass`.

src/hal/src/command/graphics.rs
AIOOB: @kvark  The reason that these were `IntoIterator<Item = Borrow<>>` initially is because arrays don't have an `into_iter` method and so users would have to type `&[clear1, clear2].iter().cloned()` whenever they used this method. However, I haven't looked but if this method needs to use the value of these variables rather than a borrow then this isn't helpful. Although it might still be sensible to require an `IntoIterator` over an `Iterator`.

kocsis1david: I had to call `.iter().cloned()` when specifying the clear values.

You can call `into_iter` on an array, but it will give `&T` instead of `T`, just like calling `into_iter` on a slice.

Btw if the parameter was just a slice, then the sorting wouldn't be needed and it wourld be possible to use `.iter().position()` instead of sorting.

src/hal/src/command/render_pass.rs
kvark: the user would provide you with an iterator and a `&mut Vec<>` for the output

kocsis1david: By user you mean the caller? How would an output `Vec` help as a parameter? As I see the only difference is that caller would have to do the allocation instead of the callee, but it's still one extra allocation.

anderejd: The caller can choose to pre-allocate and re-use the same vector many times to avoid all but a single allocation.

kocsis1david: I can place the `Vec` inside the `CommandBuffer` struct, so it can be reused, but from a maintainability perspective, it doesn't sound the best to place every `Vec` inside `CommandBuffer`.

kvark: I think this complication is rather unfortunate, and it's only caused by the fact that you want to include attachment indices in there in the first place. We don't have to, and without indices there is no need to collect anything on the backend/hal side or sort.

Let's get back to the original idea and just pass the clear colors with expected ordering.

kocsis1david: What was the original idea?

The simplest way seems to be the one that Vulkan is using now, an array of clear values that has the values at the same index as the attachment indices.

If the array was only containing the clear values for attachments with a clearop, then you would have to track the attachments in the vulkan hal backend to readd the dummy values. Also portability would have a hard time filtering for those clear values that are for attachments with clearop.

kvark: The original idea is to have an iterator of clear values that are only for the attachments that need clears. The current API has a clear value per attachment, and not only some of them are unused, it may have less total values than attachments. This is unsound.
We need to either have strictly a value per attachment, or (better) a value per cleared attachment.

kocsis1david: The better solution is a bigger work in my opinion, but someone can do it in a new PR.

Having a value per attachment, is almost the same as it is now, only need to check if the iterator has the same length as the bound render pass in the command buffer. This can be also complicated for portability, because it has to know how many attachments are there when calling `begin_render_pass`.

kvark: > The better solution is a bigger work in my opinion, but someone can do it in a new PR.

Why do you think so?

> This can be also complicated for portability, because it has to know how many attachments are there when calling begin_render_pass

It's not a big stretch for the render pass to know how many attachments it has, at the portability level.

kocsis1david: Sorry, I just realized that my explanation was a flawed, I'm tired a bit.

So these are the things that needs to be done to get there if I understand right:

The render pass needs to be extended on the vulkan backend to know what attachments are there. Other backends already have this information as far as I know. And the portability layer needs to be able to query the same information about the render pass's attachments. After these are done, it's easy to implement the clear values per cleared attachments.

kvark:
@kocsis1david looks like this is abandoned?

kocsis1david:
Yes, I don't really want to work on it nowadays, so anybody can take the issue or use my code.

kvark:
Closing in favor of #2535

PR #2315, opened 2018-08-11
[dx11] add partial attachment clears & misc fixes

fkaa:
Implemented partial attachment clears with some internal shaders & depth stencil state.

Also some other small(er) fixes:

* Add Geometry & Tessellation shaders to pipeline
* Add D1/D1Array creation
* Image & buffer binding resources are now initialized to the memory contents
* Copy shaders now work in 8x8 thread groups and have bounds checking
* The memory type order has been switched (coherent comes before non-coherent)

r? @kvark 

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/dx11/shaders/clear.hlsl
+++ b/src/backend/dx11/shaders/clear.hlsl
@@ -1,0 +1,22 @@
+cbuffer ClearColorF32   : register(b0) { float4 ClearF32; };
+cbuffer ClearColorU32   : register(b0) { uint4 ClearU32; };
+cbuffer ClearColorI32   : register(b0) { int4 ClearI32; };
+cbuffer ClearColorDepth : register(b0) { float ClearDepth; };
+
+// fullscreen triangle
+float4 vs_partial_clear(uint id : SV_VertexID) : SV_Position
+{
+    return float4(
+        float(id / 2) * 4.0 - 1.0,
+        float(id % 2) * 4.0 - 1.0,
+        0.0,
+        1.0
+    );
+}
+
+// TODO: send constants through VS as flat attributes
+float4 ps_partial_clear_float() : SV_Target0 { return ClearF32; }
+uint4  ps_partial_clear_uint() : SV_Target0 { return ClearU32; }
+int4   ps_partial_clear_int() : SV_Target0 { return ClearI32; }
+float  ps_partial_clear_depth() : SV_Depth { return ClearDepth; }
+void   ps_partial_clear_stencil() { }
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -71,8 +71,8 @@ impl Device {
             memory_properties,
             memory_heap_flags: [
                 MemoryHeapFlags::DEVICE_LOCAL,
+                MemoryHeapFlags::HOST_COHERENT,
                 MemoryHeapFlags::HOST_NONCOHERENT,
-                MemoryHeapFlags::HOST_COHERENT
             ],
             internal: internal::Internal::new(&device),
         }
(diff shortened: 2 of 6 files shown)


src/backend/dx11/shaders/clear.hlsl
```diff
@@ -0,0 +1,22 @@
...
+{
+    return float4(
+        float(id / 2) * 4.0 - 1.0,
+        float(id % 2) * 4.0 - 1.0,
+        0.0,
+        1.0
+    );
+}
+
+// TODO: send constants through VS as flat attributes
```
kvark: I don't think this is going to be better. Constants are fine, and I doubt we are ALU/register limited here anyway ;)

src/backend/dx11/src/device.rs
```diff
@@ -1208,9 +1328,64 @@ impl hal::Device<Backend> for Device {
...
+                (image::ViewKind::D1Array, unsafe { ComPtr::<d3d11::ID3D11Resource>::from_raw(resource) })
+            },
             image::Kind::D2(width, height, layers, _) => {
+                let mut initial_datas = Vec::new();
+
+                for _layer in 0..layers {
+                    for level in 0..image.mip_levels {
+                        let width = image.kind.extent().at_level(level).width;
+
+                        // TODO: layer offset?
```
kvark: yeah. If you don't want to apply it, might as well put something like `assert_eq!(layer, 0)` for now

src/backend/dx11/src/device.rs
```diff
@@ -1794,22 +1982,13 @@ impl hal::Device<Backend> for Device {
...
         R: RangeArg<u64>,
     {
-        if let Some(ref host_visible) = memory.host_visible {
-            let ptr = host_visible.borrow_mut().as_mut_ptr();
-            memory.mapped_ptr.replace(Some(ptr));
-
-            Ok(unsafe { ptr.offset(*range.start().unwrap_or(&0) as isize) })
-        } else {
-            error!("Tried to map non-host visible memory");
+        assert_eq!(memory.host_visible.is_some(), true);
```
kvark: just `assert!`

src/backend/dx11/src/internal.rs
```diff
@@ -53,14 +58,44 @@ struct BlitInfo {
...
+    data: [u32; 4],
+}
+
+// the threadgroup count we use in our copy shaders
+const COPY_THREAD_GROUP_X: u32 = 8;
+const COPY_THREAD_GROUP_Y: u32 = 8;
+
+// Holds everything we need for fallback implementations of features that are not in DX. 
+//
+// TODO: maybe get rid of `Clone`? there's _a lot_ of refcounts here and it is used as a singleton
```
kvark: why is it clonable in the first place?

src/backend/dx11/src/internal.rs
```diff
@@ -691,4 +834,124 @@ impl Internal {
...
+
+                    let attachment = {
+                        let dsv_id = subpass.depth_stencil_attachment.unwrap();
+                        &cache.framebuffer.attachments[dsv_id.0]
+                    };
+
+                    unsafe {
+                        match (depth, stencil) {
+                            (Some(_), Some(stencil)) => {
+                                context.OMSetDepthStencilState(self.partial_clear_depth_stencil_state.as_raw(), stencil);
```
kvark: do you not need to reset the states somewhere after those draws?

src/backend/dx11/src/lib.rs
```diff
@@ -1057,69 +1101,137 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
...
+            let subpass_id = render_pass.subpasses.iter().position(|sp| sp.is_using(idx));
+
+            if attachment.ops.load == Clear ||
+               attachment.stencil_ops.load == Clear
+            {
+                let raw_clear_value = *clear_iter.next().unwrap().borrow();
+
+                match (attachment.ops.load, attachment.stencil_ops.load) {
+                    (Clear, Clear) => {
+                        if format.is_depth() {
```
kvark: let's put the condition inside the match, i.e. `(Clear, Clear) if format.is_depth() => ...`

src/backend/dx11/src/lib.rs
```diff
@@ -1615,9 +1740,9 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
 
 bitflags! {
     struct MemoryHeapFlags: u64 {
-        const DEVICE_LOCAL = 1 << 0;
-        const HOST_NONCOHERENT = 1 << 1;
-        const HOST_COHERENT = 1 << 2;
+        const DEVICE_LOCAL = 0x1;
+        const HOST_NONCOHERENT = 0x4 | 0x8;
```
kvark: why should those flags be complex?

src/backend/dx11/src/internal.rs
fkaa: Moved around to a bunch of types. I suppose `Arc` is a good fit here

src/backend/dx11/src/internal.rs
fkaa: Yeah, I think something more general for handling state would be good. For now just caching the stuff touched in `clear_attachments` should be fine (DS state, shaders, render targets, viewports, etc.). Even then it's almost like replicating the whole state machine..

Something like this perhaps?
```rust
struct RenderTargetsGuard {
    rtvs: [*mut ID3D11RenderTargetView; MAX_RTVS],
}

impl Drop for RenderTargetsGuard {
    fn drop() {
        OMSetRenderTargets(rtvs);
    }
}

struct DxState {
    rtvs: [*mut ID3D11RenderTargetView; MAX_RTVS],
    ...
}

impl DxState {
    fn set_rtvs(self, rtvs) {
        OMSetRenderTargets(rtvs);
        self.rtvs = rtvs;
    }

    fn set_temp_rtvs(self, rtvs) -> RenderTargetsGuard {
        // set temporary
        OMSetRenderTargets(rtvs);

        // and return the state "guard" which resets state after out of scope
        RenderTargetsGuard {
            self.rtvs
        }
    }
}
```

kvark: alternatively, you can cache all the state like Metal does and restore from the cache

kvark:
@fkaa ping

Issue #2291, opened 2018-08-03
Respect `extent` parameter in swapchain creation


kvark:
We should be able to make a swapchain that covers the surface partially. This is needed by https://github.com/szeged/webrender/pull/203

msiglreith:
Windows doesn't support this. Min/Current/Max will always be the size of the surface, therefore the extent parameter can be ignored for dx12.

PR #2323, opened 2018-08-15
Updated SwapchainConfig logic

### LINKED ISSUE #2322: Metal: vkGetPhysicalDeviceSurfaceCapabilitiesKHR returns invalid currentExtent
The following code returns -1 for both width and height when used immediately after initializing the instance, device and creating the swap chain:

```
    VkSurfaceCapabilitiesKHR scaps;
    VK_CHECK(vkGetPhysicalDeviceSurfaceCapabilitiesKHR(physicalDevice, wd.presentationSurface, &scaps));

    unsigned int width = scaps.currentExtent.width, height = scaps.currentExtent.height;
```

The call doesn't error and other fields of scaps seem reasonable.

kvark:
Fixes #2322
Addresses the Metal part of #2291

The PR changes our swapchain configuration to reflect the capabilities better and be constructed from it.

It also removes the earlier hack in Metal backend  w.r.t. applying the hiDPI factor, rewriting the logic of how `CAMetalLayer` is created and configured. Now, it's logical and physical spaces strongly correspond to the underlying view. It respects the given extents not by changing the layer, but by configuring the images (drawables) produced from it.
Most importantly, the render space matches *logical* space of the layer. This means that if we have a window of size 1440x900 with hiDPI scale of 2.0, then it will behave as if it's a 1440x900 window and report size and create drawables accordingly. The scaling/stretching is left for the window system to apply. This works for both our examples and Dota. Ultimately, we don't want to be processing more pixels just because our hiDPI is high.

PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -58,19 +58,6 @@ struct Vertex {
     a_uv: [f32; 2],
 }
 
-
-trait SurfaceTrait {
-    #[cfg(feature = "gl")]
-    fn get_window_t(&self) -> &back::glutin::GlWindow;
-}
-
-impl SurfaceTrait for <back::Backend as hal::Backend>::Surface {
-    #[cfg(feature = "gl")]
-    fn get_window_t(&self) -> &back::glutin::GlWindow {
-        self.get_window()
-    }
-}
-
 const QUAD: [Vertex; 6] = [
     Vertex { a_pos: [ -0.5, 0.33 ], a_uv: [0.0, 1.0] },
     Vertex { a_pos: [  0.5, 0.33 ], a_uv: [1.0, 1.0] },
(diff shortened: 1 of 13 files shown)


examples/colour-uniform/main.rs
```diff
@@ -617,7 +610,7 @@ fn create_backend(window_state: &mut WindowState) -> (BackendState<back::Backend
         BackendState {
             adapter: AdapterState::new(&mut adapters),
             surface,
-            window,
+            //window,
```
msiglreith: nit: remove?

kvark:
Thanks for taking a look! I believe the example doesn't work as intended for me so needs a follow-up fix anyway.

Bors r=msiglreith

kvark:
Oops, didn't realize windows CI isn't happy...

PR #2325, opened 2018-08-16
[mtl] New vertex binding logic

### LINKED ISSUE #2058: Quad example cannot run on Metal
Short info header:
- GFX version:master
- OS:MacOS
cargo run --bin quad --features metal
 Compiling hal-examples v0.1.0 (file:///Users/inmath/GitHub/gfx/examples/hal)
    Finished dev [unoptimized + debuginfo] target(s) in 9.17 secs
     Running `/Users/inmath/GitHub/gfx/target/debug/quad`
2018-05-21 23:56:50.472 quad[28500:3579820] -[MTLComputePipelineDescriptorInternal buffers]: unrecognized selector sent to instance 0x7fad8d5324e0
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
fatal runtime error: failed to initiate panic, error 5
Abort trap: 6

### LINKED ISSUE #2307: Refactor vertex buffer setting to bundle the bindings on Metal
Binding vertex buffers happens fairly often, and we are still using single-buffer binding there. Need to refactor the code to bind all vertex buffers at once.

### LINKED ISSUE #2324: Metal: vertex descriptor is built without using the location information
This is the code that creates Metal vertex descriptors:

https://github.com/gfx-rs/gfx/blob/a571ae5a15b2fd7bdd457c9d2d6189167ccfbf54/src/backend/metal/src/device.rs#L1011-L1014

The expected behavior is that the location in the vertex shader matches the location in the attributes array; instead this code fills the array from 0 to count-1, which doesn't work for vertex shaders that have 3 input attributes with locations 0, 2, 4. The array index should instead use VkVertexInputAttributeDescription::location.

kvark:
Fixes #2324
Fixes #2307
Fixes #2058 (unrelated to the PR, but harmless)

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -176,7 +176,8 @@ struct Temp {
 struct RenderPipelineState {
     raw: metal::RenderPipelineState,
     ds_desc: pso::DepthStencilDesc,
-    vbuf_map: native::VertexBufferMap,
+    vertex_buffers: native::VertexBufferVec,
+    attribute_buffer_index: ResourceIndex,
     at_formats: Vec<Option<Format>>,
 }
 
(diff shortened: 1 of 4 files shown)


kvark:
CTS draw section as well as Dota2 seem to work good with this.

kvark:
I couldn't resist to fix #2307 since I had to touch the vertex buffer binding, and this area was the last one not following the bundle path. New code is much simpler :D I detected no regressions in Dota or CTS.

Basically, we store an array of vertex buffer descriptors and their extra offsets, indexed by the relative Metal buffer index. And we are no longer trying to detect if a vertex binding changed, build a mask, and limit the bindings. We just bind the whole sub-range of vertex buffers at once.

PR #2330, opened 2018-08-20
[vulkan] error reporting using VK_EXT_debug_utils extension (requires integration with upcoming version of `ash`)

bzm3r:
This is just a "test" PR to give an idea of how things will change with the upcoming changes to `ash`, which among things, provides a wrapper for the extension `VK_EXT_debug_utils`. This PR depends upon another PR: https://github.com/gfx-rs/gfx/pull/2346

Using `VK_EXT_debug_utils` allows for [much nicer validation layer message reporting](https://www.lunarg.com/new-tutorial-for-vulkan-debug-utilities-extension/). Furthermore, `VK_EXT_debug_report` and `VK_EXT_debug_marker` will be going stale as they are going to be superseded by `VK_EXT_debug_utils`.

`VK_EXT_debug_utils` still provides a callback function which makes collecting validation layer messages to console, *regardless of which system you are on*: all one needs to do is set up something like [`env_logger`](https://github.com/sebasmagri/env_logger/) to print the logs collected using the [`log` crate.](https://github.com/rust-lang-nursery/log).

For now, I have just implemented basic error reporting functionality using `VK_EXT_debug_utils`. I have left things like setting markers, naming objects, etc. (see tutorial) for the future. 

I'll be putting in examples of how to use this functionality in [gfx-hal-tutorial](https://github.com/grovesNL/gfx-hal-tutorial). Basically, all one needs to do is create a set up function like this:

```
fn log_output_setup() {
    use env_logger::{Builder, Target};

    let mut builder = Builder::new();
    // print validation info to stdout
    builder.target(Target::Stdout);
    // want to show all logs
    builder.parse("error,warn,info,trace");
    builder.init();
}
```

...and then call the setup function as early as possible in the application, and you are ready to go.

PR checklist:
- [ ] show example usage in [gfx-hal-tutorial](https://github.com/grovesNL/gfx-hal-tutorial)

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -5,26 +5,26 @@ use std::{ffi, mem, ptr, slice};
 
 use spirv_cross::{hlsl, spirv, ErrorCode as SpirvErrorCode};
 
-use winapi::Interface;
-use winapi::um::{d3d12, d3dcommon, d3dcompiler, synchapi, winbase, winnt};
 use winapi::shared::minwindef::{FALSE, TRUE, UINT};
 use winapi::shared::{dxgi, dxgi1_2, dxgi1_4, dxgiformat, dxgitype, winerror};
+use winapi::um::{d3d12, d3dcommon, d3dcompiler, synchapi, winbase, winnt};
+use winapi::Interface;
 use wio::com::ComPtr;
 
-use hal::{self, buffer, device as d, error, format, image, mapping, memory, pass, pso, query};
 use hal::format::{Aspects, Format};
 use hal::memory::Requirements;
 use hal::pool::CommandPoolCreateFlags;
-use hal::queue::{RawCommandQueue, QueueFamilyId};
+use hal::queue::{QueueFamilyId, RawCommandQueue};
 use hal::range::RangeArg;
+use hal::{self, buffer, device as d, error, format, image, mapping, memory, pass, pso, query};
 
-use {
-    conv, descriptors_cpu, native as n, root_constants, window as w,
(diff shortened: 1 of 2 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -117,42 +122,152 @@ pub struct Instance {
...
         unimplemented!()
     }
 }
 
-extern "system" fn callback(
+extern "system" fn debug_utils_messenger_callback(
+    message_severity: vk::DebugUtilsMessageSeverityFlagsEXT,
+    message_type: vk::DebugUtilsMessageTypeFlagsEXT,
+    p_callback_data: *const vk::DebugUtilsMessengerCallbackDataEXT,
+    user_data: *mut libc::c_void,
```
grovesNL: I don't think the extra `libc` dependency should be necessary for this – other functions already use `std::os::raw::c_void`

kvark:
Amazing work!

> fn log_output_setup() {

Do we really need to opt into the detailed builder-like initialization of env_logger as opposed to just calling `env_logger::init()`? I think the excercize of configuring the logger should be left to the logger tutorial, not gfx-rs one.

bzm3r:
@kvark i prefer the builder notation because it makes it clear what `env_logger`'s settings are; otherwise, one has to mess with environment variables, etc. 

i think the builder code is succinct and self-commenting really:

```
    use env_logger::{Builder, Target};

    // create struct that will set up new logger
    let mut builder = Builder::new();
    // specify: print validation info to stdout
    builder.target(Target::Stdout);
    // specify: want to show all logs
    builder.parse("error,warn,info,trace");

    // initialize logger
    builder.init();
```

bzm3r:
@grovesNL i have updated code according to your review comments

PR #2331, opened 2018-08-20
Metal-rs and winit updates

kvark:
PR checklist:

### DIFF
--- a/src/hal/src/device.rs
+++ b/src/hal/src/device.rs
@@ -616,7 +616,7 @@ pub trait Device<B: Backend>: Any + Send + Sync {
     ///
     /// Queries are managed using query pool objects. Each query pool is a collection of a specific
     /// number of queries of a particular type.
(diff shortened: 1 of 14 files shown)


src/hal/src/device.rs
```diff
@@ -616,7 +616,7 @@ pub trait Device<B: Backend>: Any + Send + Sync {
     ///
     /// Queries are managed using query pool objects. Each query pool is a collection of a specific
     /// number of queries of a particular type.
-    fn create_query_pool(&self, ty: query::QueryType, count: u32) -> B::QueryPool;
+    fn create_query_pool(&self, ty: query::QueryType, count: query::QueryId) -> B::QueryPool;
```
grovesNL: Should fix this in the backends too, I guess

kvark:
Thanks for taking a quick look!
Backends are updated now.
bors r=grovesNL

PR #2334, opened 2018-08-20
Query API and basic support on Metal

### LINKED ISSUE #1638: [RFC] Query resolve
## Overview

We have 3 types of queries in Vulkan: Occlusion, PipelineStatistics and Timestamp.
The first two are handled via `vkCmdBeginQuery`/`vkCmdEndQuery` and timestamps via `vkCmdWriteTimestamp`. Queries are allocated in heaps and marked as 'unavailable' at the beginning.
`vkCmdBeginQuery` calls activate queries and once the results can be obtained (after `vkCmdEndQuery`) the state will switch to `available`. `vkResetQueryPool` will reset them to `unavailable`.

## Mapping to Metal
Metal doesn't expose any significant kind of timestamp or pipeline statistics support (afaik) and only supports `Counting occlusion query`.
Therefore, `pipelineStatisticsQuery` feature should be exposed as `false`, `timestampComputeAndGraphics` as false and `timestampValidBits` as 0 for all queues.
Occlusion queries are configured at renderpass boundary. Binary occlusion support is required and counting optional. Results are directly written to a buffer.

## Mapping to D3D12

D3D12 uses query heaps, similar to Vulkan's query pools. The begin/end/write commands can be mapped without significant overhead to D3D12. Reset command implementation depends on the further Vulkan Portability API.
Resolving query data in D3D12 is implemented as a command list function `ResolveQueryData`(https://msdn.microsoft.com/de-de/library/windows/desktop/dn903896(v=vs.85).aspx). It write the current query values contiguously to a buffer (upload/download/local) as 64bit integers.
Vulkan exposes two functions to resolve queries: `vkGetQueryPoolResults` (device-side) and  `vkCmdCopyQueryPoolResults` (command buffer-side). Beside additional effort to submit a command buffer with a query resolve to emulate the device function, the `VkQueryResultFlags` are difficult to handle:

- `VK_QUERY_RESULT_64_BIT` denotes that values are written as u64 values (default: u32)
- `VK_QUERY_RESULT_WITH_AVAILABILITY_BIT` adds an additional integer value indicating if the specific query is available to the buffer.
- `VK_QUERY_RESULT_WAIT_BIT` waits on all queries to be available, else resulting in a possible device loss.

Additionally, users can define which pipeline statistics can should be recorded and only marked statistics get written to the buffer. D3D12 will emit all statistics.

#### Issues
* Track the status (`available/unavailable`) of each query in the query heap to correctly emit the availability flags if requested. As these happen async it needs to be done in the command buffer. For example, compute shader to write availability values into an hidden buffer on command buffer end if we have Begin/EndQuery commands.
* Compute shader to handle the case where `VK_QUERY_RESULT_64_BIT` is not set.
* vkGetQueryPoolResults requires additional synchronization to ensure that queries have been processed by the command queues when `VK_QUERY_RESULT_WAIT_BIT` is set.

I think these are the main issues so far. Metal requires a bit further analysis but shouldn't cause major complications.

kvark:
Fixes #1638 (for the most part)
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -12,13 +12,12 @@ use std::ops::{Deref, Range};
 use std::sync::Arc;
 use std::{iter, mem, slice, time};
 
-use hal::{buffer, command as com, error, memory, pool, pso};
+use hal::{buffer, command as com, error, memory, pool, pso, query};
 use hal::{DrawCount, SwapImageIndex, VertexCount, VertexOffset, InstanceCount, IndexCount, IndexType, WorkGroupCount};
 use hal::backend::FastHashMap;
 use hal::format::{Aspects, Format, FormatDesc};
 use hal::image::{Extent, Filter, Layout, Level, SubresourceRange};
 use hal::pass::{AttachmentLoadOp, AttachmentOps};
-use hal::query::{Query, QueryControl, QueryId};
 use hal::queue::{RawCommandQueue, RawSubmission};
 use hal::range::RangeArg;
 
@@ -3477,14 +3476,14 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
 
     fn begin_query(
         &mut self,
-        query: Query<Backend>,
-        flags: QueryControl,
+        query: query::Query<Backend>,
+        flags: query::ControlFlags,
     ) {
         match query.pool {
             native::QueryPool::Occlusion(ref pool_range) => {
                 debug_assert!(pool_range.start + query.id < pool_range.end);
                 let offset = (query.id + pool_range.start) as buffer::Offset + mem::size_of::<u64>() as buffer::Offset;
-                let mode = if flags.contains(QueryControl::PRECISE) {
+                let mode = if flags.contains(query::ControlFlags::PRECISE) {
                     metal::MTLVisibilityResultMode::Counting
                 } else {
                     metal::MTLVisibilityResultMode::Boolean
@@ -3502,7 +3501,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
 
     fn end_query(
         &mut self,
-        query: Query<Backend>,
+        query: query::Query<Backend>,
     ) {
         match query.pool {
             native::QueryPool::Occlusion(ref pool_range) => {
@@ -3521,7 +3520,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
     fn reset_query_pool(
         &mut self,
         pool: &native::QueryPool,
-        queries: Range<QueryId>,
+        queries: Range<query::Id>,
     ) {
         let pool_shared = self.pool_shared.borrow_mut();
         match *pool {
@@ -3544,10 +3543,22 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         }
     }
 
+    fn copy_query_pool_results(
+        &mut self,
+        _pool: &native::QueryPool,
+        _queries: Range<query::Id>,
+        _buffer: &native::Buffer,
+        _offset: buffer::Offset,
+        _stride: buffer::Offset,
+        _flags: query::ResultFlags,
+    ) {
+        unimplemented!()
+    }
+
     fn write_timestamp(
         &mut self,
         _: pso::PipelineStage,
-        _: Query<Backend>,
+        _: query::Query<Backend>,
     ) {
         // nothing to do, timestamps are unsupported on Metal
     }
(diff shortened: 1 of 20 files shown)


kvark:
Looks like we are blocked on https://github.com/MaikKlein/ash/issues/100 for Vulkan :(
cc @MaikKlein

kvark:
Current status: 136 / 150 occlusion tests passing
Known 2 issues:
  1. sometimes getting the query results synchronously gets stuck
  2. in some cases copying results with availability doesn't produce correct results

kvark:
Vulkan is blocked on https://github.com/MaikKlein/ash/issues/105 now

kvark:
Update on Metal: all the known issues are resolved. We pass 150/150 tests without failures or hangs :tada: 
The PR is blocked on reviews and getting Vulkan backend to compile. In the worst case we'll comment out the Vulkan parts that are blocking.

src/backend/vulkan/src/command.rs
```diff
@@ -849,10 +849,10 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
     fn begin_query(
         &mut self,
         query: query::Query<Backend>,
-        control: query::QueryControl,
+        control: query::ControlFlags,
     ) {
         let mut flags = vk::QueryControlFlags::empty();
-        if control.contains(query::QueryControl::PRECISE) {
+        if control.contains(query::ControlFlags::PRECISE) {
```
grovesNL: `map_query_control_flags` instead?

src/backend/metal/src/device.rs
```diff
@@ -2291,12 +2304,102 @@ impl hal::Device<Backend> for Device {
...
+        data: &mut [u8], stride: buffer::Offset,
+        flags: query::ResultFlags,
+    ) -> Result<bool, query::Error> {
+        let is_ready = match *pool {
+            native::QueryPool::Occlusion(ref pool_range) => {
+                let visibility = &self.shared.visibility;
+                let is_ready = if flags.contains(query::ResultFlags::WAIT) {
+                    let mut guard = visibility.allocator.lock();
+                    while !visibility.are_available(pool_range.start, &queries) {
+                        println!("query: not available"); //TEMP
```
grovesNL: Is this left here intentionally?

src/backend/metal/src/command.rs
```diff
@@ -1656,13 +1682,21 @@ impl RawCommandQueue<Backend> for CommandQueue {
...
                 let free_textures = self.retained_textures
                     .drain(..)
                     .collect::<SmallVec<[_; BLOCK_BUCKET]>>();
+                let visibility = if self.active_visibility_queries.is_empty() {
+                    None
+                } else {
+                    let queries = self.active_visibility_queries
+                        .drain(..)
+                        .collect::<SmallVec<[_; BLOCK_BUCKET]>>();
+                    Some((Arc::clone(&self.shared), queries))
```
grovesNL: Why do we need to clone`self.shared`?

src/backend/metal/src/command.rs
```diff
@@ -3454,31 +3501,192 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                            let com_avail = soft::BlitCommand::CopyBuffer {
+                                src: AsNative::from(visibility.buffer.as_ref()),
+                                dst: AsNative::from(buffer.raw.as_ref()),
+                                region: com::BufferCopy {
+                                    src: visibility.availability_offset + absolute_index * size_meta,
+                                    dst: dst_offset + size_payload,
+                                    size: size_meta,
+                                },
+                            };
+                            // An extra paddig is requred if the client expects 64bits availability without a wait
```
grovesNL: `paddig` -> `padding`

src/backend/metal/src/command.rs
grovesNL: Oh, I didn't notice the callback block. Sounds good

src/backend/metal/src/command.rs
```diff
@@ -3454,31 +3501,192 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                            region: com::BufferCopy {
+                                src: absolute_index * size_data,
+                                dst: dst_offset,
+                                size: size_payload,
+                            },
+                        };
+
+                        let (com_avail, com_pad) = if flags.contains(query::ResultFlags::WITH_AVAILABILITY | query::ResultFlags::WAIT) {
+                            // Technically waiting is a no-op on a single queue. However,
+                            // the client expects the avaiability to be set regardless.
```
grovesNL: `avaiability` -> `availability`

kvark:
Landing, as per approval on gitter.
bors r=grovesNL

kvark:
Bors retry

kvark:
Bors r=grovesNL

kvark:
Oh yes you are, silly bors!

PR #2335, opened 2018-08-22
[pre-ll/GL] query current front/back buffer for reading

### LINKED ISSUE #2310: When reading from a texture: InvalidOperation executing command: CopyTextureToBuffer
Short info header:
- GFX version: pre-ll
- OS: Linux
- GPU: Intel Integrated HD 4000

I have a headless context, and I'm calling:

```
    let (screen_w, screen_h, _, _) = data.out.get_dimensions();
    encoder.copy_texture_to_buffer_raw(
        data.out.raw().get_texture(),
        None,
        gfx::texture::RawImageInfo {
            xoffset: 0,
            yoffset: 0,
            zoffset: 0,
            width: screen_w,
            height: screen_h,
            depth: 0,
            format: ColorFormat::get_format(),
            mipmap: 0,
        },
        download.raw(),
        0
    ).unwrap();
```

Where `data.out` is the screen framebuffer, and `download` is a download buffer.

I get an InvalidOperation error on this:

```
thread 'main' panicked at 'Error InvalidOperation executing command: CopyTextureToBuffer(TextureCopyRegion { texture: Surface(0), kind: D2(512, 512, Single), cube_face: None, info: ImageInfoCommon { xoffset: 0, yoffset: 0, zoffset: 0, width: 512, height: 512, depth: 0, format: Format(R8_G8_B8_A8, Unorm), mipmap: 0 } }, 2, 0, 1)'
```

This looksl ike a GL error, but i'm having no success with `apitrace` - it just doesn't produce a trace file. Potentially because this is a headless context.

Any info on this? Totally lost._.

Thanks.

tomc1998:
Fixes #2310 

This is more of a temporary fix, and just allows code to run. Not sure what the performance implications of this are - however, since downloading a GPU buffer probably doesn't happen every frame (?) and extra glGet shouldn't hurt too much (?)

EDIT:

I also applied a minor fix to the gamma example which I couldn't get to run, and was needed to test whether the fix works. I'm using i3, which does some funny things with window dimensions, so potentially that was causing the example to fail on my machine.

### DIFF
--- a/src/backend/gl/src/tex.rs
+++ b/src/backend/gl/src/tex.rs
@@ -663,8 +663,10 @@ fn bind_read_fbo(gl: &gl::Gl, texture: NewTexture, level: t::Level, fbo: FrameBu
     if texture == NewTexture::Surface(0) {
         //Warning: assuming the back buffer
         unsafe {
+            let mut buffer : gl::types::GLint = 0;
+            gl.GetIntegerv(gl::DRAW_BUFFER, &mut buffer);
             gl.BindFramebuffer(target, 0);
-            gl.ReadBuffer(gl::BACK);
+            gl.ReadBuffer(buffer as u32);
         }
         return
     }
--- a/examples/gamma/main.rs
+++ b/examples/gamma/main.rs
@@ -86,9 +86,6 @@ pub fn main() {
     };
 
     let mut screenshot = false;
-    let (w, h, _, _) = data.out.get_dimensions();
-    let mut download = factory.create_download_buffer::<SurfaceData>(w as usize * h as usize)
-        .unwrap();
 
     let mut running = true;
     while running {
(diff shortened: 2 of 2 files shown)


src/backend/gl/src/tex.rs
```diff
@@ -663,8 +663,10 @@ fn bind_read_fbo(gl: &gl::Gl, texture: NewTexture, level: t::Level, fbo: FrameBu
     if texture == NewTexture::Surface(0) {
         //Warning: assuming the back buffer
         unsafe {
+            let mut buffer : gl::types::GLint = 0;
+            gl.GetIntegerv(gl::DRAW_BUFFER, &mut buffer);
```
kvark: do you know what is the current draw buffer/FBO here? I don't think we can rely on that. E.g. what if we are currently copying data from the default FBO into a user FBO, the DRAW_BUFFER would then be `gl::COLOR_ATTACHMENT1`, which is not expected here.

tomc1998:
The problem is I don't know what the plumbing is that I need to modify to
pass this along, I just need a solution for now. How should I pass which
color buffer we're talking about, does it go in the surface struct?

kvark:
@tomc1998 the `Command` struct (from `gfx_device_gl`) could just have a public field `pub main_face: GLenum,`, which is defaulted to `gl::BACK` but users set it to `gl::FRONT` if they know it's a single-buffered swapchain. This is similar to [display_fb field](https://github.com/gfx-rs/gfx/blob/463969d3d0902eccce4ee4ade40d3a96a3d038d4/src/backend/gl/src/command.rs#L344).

kvark:
@tomc1998 ping

kvark:
Looks like this PR is abandoned... please feel free to resume the work @tomc1998 !

PR #2337, opened 2018-08-23
[WIP] [mtl] [iOS] hacking on iOS metal support

mtak-:
Here's the start of a PR to introduce support for iOS in gfx-backend-metal. With this PR gfx-backend-metal does build on iOS, but I have yet to get into testing. I'm far from a metal expert, so I would appreciate any feedback.

### Questions
1. `Instance` has two impls, one for iOS and one for macOS. Is there a preferred way to organize these differences? For example, having a platform submodule, with ios.rs, and macos.rs submodules.
2. iOS does not support `[[render_target_array_index]]`, but as far as I can tell, it does fully support array textures depending on the specific iOS version. blit.metal and clear.metal use `[[render_target_array_index]]` so I added a macros.metal file which will conditionally use the attribute. This allows the shaders to build, but I have no idea what the runtime behavior for iOS is. I'm hoping someone more knowledgable could point me in the correct direction.
3. Should whitespace changes via `rustfmt` be a separate PR, or just a standalone commit?

PR checklist:

TODO:
- [x] ~~`MTLLanguageVersion` detection based on iOS version in device.rs - macOS already needs fixing~~ fixed by #2338 
- [ ] check `broken_viewport_near_depth` in `Shared::new`'s PrivateDisabilities detection. Currently based on `MTLFeatureSet_macOS_GPUFamily1_v4` which might be incorrect for iOS
- [ ] do testing on iOS

### DIFF
--- a/src/backend/metal/shaders/macros.metal
+++ b/src/backend/metal/shaders/macros.metal
@@ -1,0 +1,5 @@
+#ifdef __METAL_MACOS__
+#   define GFX_RENDER_TARGET_ARRAY_INDEX [[render_target_array_index]]
+#else
+#   define GFX_RENDER_TARGET_ARRAY_INDEX
+#endif
--- a/src/backend/metal/src/window.rs
+++ b/src/backend/metal/src/window.rs
@@ -9,8 +9,7 @@ use hal::{self, format, image};
 use hal::{Backbuffer, SwapchainConfig};
 use hal::window::Extent2D;
 
-use cocoa::foundation::{NSRect};
-use core_graphics::geometry::{CGSize};
+use core_graphics::geometry::{CGRect, CGSize};
 use foreign_types::{ForeignType, ForeignTypeRef};
 use parking_lot::{Mutex, MutexGuard};
 use metal;
@@ -27,7 +26,7 @@ pub struct Surface {
 
 #[derive(Debug)]
 pub(crate) struct SurfaceInner {
-    pub(crate) nsview: *mut Object,
+    pub(crate) view: *mut Object,
     pub(crate) render_layer: Mutex<CAMetalLayer>,
 }
 
@@ -36,7 +35,7 @@ unsafe impl Sync for SurfaceInner {}
 
 impl Drop for SurfaceInner {
     fn drop(&mut self) {
-        unsafe { msg_send![self.nsview, release]; }
+        unsafe { msg_send![self.view, release]; }
     }
 }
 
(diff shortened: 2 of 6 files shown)


kvark:
> Instance has two impls, one for iOS and one for macOS. Is there a preferred way to organize these differences?

If it's a bit of code, we can have it in a single module (e.g. `lib.rs`), otherwise splitting is fine. Really, your call to make :)

> iOS does not support [[render_target_array_index]]

This is unfortunate. We've also had quite a few driver issues when using this, so it would make sense to have all the code using render target arrays behind an internal flag. I don't expect you to refactor/rewrite all the old code to conform to that, of course. The beginning was done with the introduction of [CLEAR_IMAGE_ARRAY](https://github.com/gfx-rs/gfx/blob/174a4ff59ca20f195ed438fd8cf89354db18f935/src/backend/metal/src/command.rs#L41), but we'll need to extend it further. For your port, just disabling the use of layered rendering specifically on iOS is fine - we'll revisit this once the code is truly conditional on macOS.

> This allows the shaders to build, but I have no idea what the runtime behavior for iOS is

If we can't use that index, then we can't use layered rendering, and all the layered operations that we currently do (blit/clear) are going to affect the first layer only. We'll fix this, not required for now.

> Should whitespace changes via rustfmt be a separate PR, or just a standalone commit?

I'd certainly prefer a separate commit.

src/backend/metal/shaders/macros.metal
```diff
@@ -0,0 +1,5 @@
+#ifdef __METAL_MACOS__
+#   define GFX_RENDER_TARGET_ARRAY_INDEX [[render_target_array_index]]
+#else
+#   define GFX_RENDER_TARGET_ARRAY_INDEX
+#endif
```
kvark: nit: missing newline

src/backend/metal/src/window.rs
```diff
@@ -9,8 +9,11 @@ use hal::{self, format, image};
 use hal::{Backbuffer, SwapchainConfig};
 use hal::window::Extent2D;
 
-use cocoa::foundation::{NSRect};
 use core_graphics::geometry::{CGSize};
+#[cfg(target_os = "macos")]
+use cocoa::foundation::NSRect as OSRect;
```
grovesNL: Could we just use `CGSize` for both or is this problematic?

mtak-: That would work.

According to [this](https://developer.apple.com/documentation/foundation/nsrect?language=objc) `NSRect` is typedef'd to `CGRect` on 64-bit platforms, and according to [this](https://support.apple.com/en-us/HT208436) metal only works on 64-bit OSX (also true for iOS).

mtak-:
@kvark Sorry for the delay. I'm in the middle of moving. Everything should be dandy now.

PR #2338, opened 2018-08-24
[mtl] Detect OS version

grovesNL:
Detect OS version to set `MTLLanguageVersion` based on the minimum version specified in [the Apple documentation](https://developer.apple.com/documentation/metal/mtllanguageversion?language=objc) (on each enum value's page)

Note `NSProcessInfo` already exists in foundation, but `operatingSystemVersion`/`NSOperatingSystemVersion` appear to be missing for now (eventually we can upstream it).

PR checklist:


kvark:
I don't understand what bors is waiting for here :(

kvark:
Trying to wake up bors
bors r-

PR #2340, opened 2018-08-24
[mtl 2D texel buffer view

kvark:
Makes our buffer views bigger, relying on https://github.com/KhronosGroup/SPIRV-Cross/issues/664
PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -172,10 +172,6 @@ const BASE_INSTANCE_SUPPORT: &[MTLFeatureSet] = &[
 const PUSH_CONSTANTS_DESC_SET: u32 = !0;
 const PUSH_CONSTANTS_DESC_BINDING: u32 = 0;
 
-//The offset and bytesPerRow parameters must be byte aligned to the size returned by the
-// minimumLinearTextureAlignmentForPixelFormat: method. The bytesPerRow parameter must also be
-// greater than or equal to the size of one pixel, in bytes, multiplied by the pixel width of one row.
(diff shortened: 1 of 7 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1957,22 +1955,27 @@ impl hal::Device<Backend> for Device {
             // Vadlidator says "Linear texture: cannot create compressed, depth, or stencil textures"
             return Err(buffer::ViewCreationError::UnsupportedFormat { format: format_maybe })
         }
-        let block_count = (end_rough - start) * 8 / format_desc.bits as u64;
+
+        //Note: we rely on SPIRV-Cross to use the proper 2D texel indexing here
+        let texel_count = (end_rough - start) * 8 / format_desc.bits as u64;
+        let col_count = cmp::min(texel_count, self.private_caps.max_texture_size);
+        let row_count = cmp::max(1, texel_count / self.private_caps.max_texture_size);
```
grovesNL: I'm a bit surprised `row_count` doesn't depend on `col_count` at all here, how does this work when either dimension has been clamped to `max_texture_size`?

grovesNL: Does the stride account for this somehow?

kvark: if `col_count < max_texture_size` then `row_count` is guaranteed to be `1` anyway. Otherwise, it is equal to `max_texture_size` and thus we don't need it in the `row_count` computation.

grovesNL: Got it, thanks! Should we assert that `row_count` is less than `self.private_caps.max_texture_size` too then?

PR #2349, opened 2018-08-27
[mtl] fix Bc1Rrgba support and the alpha swizzle

kvark:
Fixes the early RPC3/Uncharted failures
PR checklist:


kvark:
I think this is straightforward. Gotta merge to update portability and make a new build.
bors r+

kvark:
Strangely enough, Dota2 relies on us mapping Bc1Rgb -> Bc1Rgba, which doesn't sound right at all. Leaving this unaffected by the PR.
bors r+

kvark:
@Thomspoon they are somewhat non-obvious. TL;DR: Metal "packed" is the opposite of Vulkan "packed"

Issue #2358, opened 2018-08-29
make reftests fails on clear-image


jackmott:
Short info header:
- GFX version:master
- OS:linux
- GPU:nvidia 1070

```
 Scene 'transfer':
                Test 'fill-last' ...    ran: PASS
                Test 'fill-first' ...   ran: PASS
                Test 'clear-image' ...  ran: FAIL [127, 127, 127, 127]
```

kvark:
Looks like a precision error, since the expected values are 128 and we clear to 0.5. I guess we could be clearing to something like 0.5001 for confidence?

PR #2359, opened 2018-08-29
Expose Vulkan device_type from hal

jackmott:
Notes:

I ran rustfmt on files I touched, this caused many changes in some, hence the wild diffs.
make reftests failed, I believe unrelated, see the note below in the PR checklist
I have no way to build the metal backend, so I have not built it. Relying on bors for now.

### Backends

* Vulkan - This gets exposed exactly as it does by Vulkan/ash
* Metal - Metal docs say that is_low_power() is true for integrated, false for discrete. So these are the only two values Metal will return, using this to decide.
* DX11/12 - I return VirtualGpu if software rendering is true, otherwise discrete. I did not find that any more detailed info is exposed, but I am uncertain.
* OpenGL - I always return discrete, it does not appears OpenGL can expose this information.

PR checklist:
- [x] `make reftests` succeeds  (it did not, filed issue #2358 )
- [x] tested examples with the following backends:Vulkan

### DIFF
--- a/src/backend/dx11/src/dxgi.rs
+++ b/src/backend/dx11/src/dxgi.rs
@@ -1,4 +1,5 @@
 use hal::AdapterInfo;
+use hal::adapter::DeviceType;
 
 use winapi::shared::guiddef::GUID;
 use winapi::shared::{dxgi, dxgi1_2, dxgi1_3, dxgi1_4, dxgi1_5, winerror};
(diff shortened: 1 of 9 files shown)


jackmott:
I can, I suppose best way is start a fresh PR?
Alternatively I could tag the lines actually changed in a comment, there aren't many. Would that work?

kvark:
@jackmott `git rebase -i master` and then squash and edit some

jackmott:
ok I gave it a try but appear to have made of mess of things with git. I'm going to do a new PR.

kvark:
You know you can just force-push to this branch and avoid creation of new PRs?
Also, something is wrong with the "Exposing Vulkan's ..." commit, it's always put the last in Github history. Maybe the date is wrong?

jackmott:
oh yeah, time was wrong on my windows boot. it does then when i switch.  new pr is already up.

PR #2360, opened 2018-08-29
Expose Vulkan's device_type from hal

jackmott:
Notes:

make reftests failed, I believe unrelated, see the note below in the PR checklist
I have no way to build the metal backend, so I have not built it. Relying on bors for now.

### Backends

* Vulkan - This gets exposed exactly as it does by Vulkan/ash
* Metal - Metal docs say that is_low_power() is true for integrated, false for discrete. So these are the only two values Metal will return, using this to decide.
* DX11/12 - I return VirtualGpu if software rendering is true, otherwise discrete. I did not find that any more detailed info is exposed, but I am uncertain.
* OpenGL - I always return discrete, it does not appears OpenGL can expose this information.

PR checklist:
- [x] `make reftests` succeeds  (it did not, filed issue #2358 )
- [x] tested examples with the following backends:Vulkan

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -9,19 +9,19 @@ extern crate bitflags;
 extern crate log;
 extern crate gfx_gl as gl;
 extern crate gfx_hal as hal;
-extern crate smallvec;
-extern crate spirv_cross;
 #[cfg(feature = "glutin")]
 pub extern crate glutin;
+extern crate smallvec;
+extern crate spirv_cross;
 
 use std::cell::Cell;
 use std::fmt;
-use std::sync::Arc;
 use std::ops::Deref;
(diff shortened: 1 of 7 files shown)


src/backend/gl/src/lib.rs
```diff
@@ -213,10 +214,10 @@ impl PhysicalDevice {
         hal::Adapter {
             info: hal::AdapterInfo {
                 name,
-                vendor: 0, // TODO
-                device: 0, // TODO
-                device_type: hal::adapter::DeviceType::DiscreteGpu,  // TODO Is there a way to detect this?
-                software_rendering: false, // not always true ..
+                vendor: 0,                                          // TODO
```
kvark: wooot, rustfmt aligns the comments? nice!

PR #2361, opened 2018-08-30
[mtl] carefully restore resources and push constants on pipeline change

kvark:
Fixes the only Metal validation issue occurring in Mario Kart run via Doplhin.
PR checklist:
- [x] tested examples with the following backends: quad

This is tricky but super important to get right for us. First part, a minor one, is that the ID of the push constants depend on the pipeline layout (which was a recent change), and thus we didn't re-bind them properly on a pipeline change. This PR does it.

The second part is most complex. Both push constants and descriptors are bound with the pipeline layout, but the vertex buffers are bound according to a particular graphics pipeline. And they tend to collide with the others. Vertex buffers may occupy any slots starting from `attribute_buffer_index` member of the layout (which is also copied to the pipeline on creation). Thus, when a new pipeline is getting assigned, we can't trust any vertex buffers at this index and above. The new logic attempts to re-bind the push constants and vertex buffers affected.

In an ideal world, vertex buffers would be a part of the pipeline layout, and we wound't have to resort to any state restoration like this. Life would be easier, and we'd all ride unicorns.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -10,7 +10,7 @@ use std::borrow::Borrow;
 use std::cell::RefCell;
 use std::ops::{Deref, Range};
 use std::sync::Arc;
-use std::{iter, mem, slice, time};
+use std::{cmp, iter, mem, slice, time};
 
 use hal::{buffer, command as com, error, memory, pool, pso, query};
 use hal::{DrawCount, SwapImageIndex, VertexCount, VertexOffset, InstanceCount, IndexCount, IndexType, WorkGroupCount};
@@ -174,11 +174,13 @@ struct Temp {
     blit_vertices: FastHashMap<(Aspects, Level), Vec<BlitVertex>>,
 }
 
+type VertexBufferMaybeVec = Vec<Option<(pso::VertexBufferDesc, pso::ElemOffset)>>;
+
 #[derive(Clone)]
 struct RenderPipelineState {
     raw: metal::RenderPipelineState,
     ds_desc: pso::DepthStencilDesc,
-    vertex_buffers: native::VertexBufferVec,
+    vertex_buffers: VertexBufferMaybeVec,
     attribute_buffer_index: ResourceIndex,
     at_formats: Vec<Option<Format>>,
 }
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1079,14 +1079,6 @@ impl hal::Device<Backend> for Device {
         let pipeline_layout = &pipeline_desc.layout;
         let pass_descriptor = &pipeline_desc.subpass;
 
-        let attribute_buffer_index = pipeline_layout.attribute_buffer_index();
-        if attribute_buffer_index + pipeline_desc.vertex_buffers.len() as ResourceIndex > self.private_caps.max_buffers_per_stage {
```
grovesNL: Will something else catch this case now?

src/backend/metal/src/command.rs
```diff
@@ -3000,6 +3004,33 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+            // re-bind push constants
+            if let Some(index) = pipeline.vs_pc_buffer_index {
+                if Some(index) != self.state.resources_vs.push_constants_buffer_id
+                    || index >= old_attribute_buffer_index
+                {
+                    pre.issue(self.state.push_vs_constants(index));
+                }
+            }
+            if let Some(index) = pipeline.ps_pc_buffer_index {
+                if Some(index) != self.state.resources_ps.push_constants_buffer_id {
```
grovesNL: Do we need to bother updating push constants when `push_constants_buffer_id` is `None`? Why not `if let Some(id) = push_constants_buffer_id` inside?

src/backend/metal/src/command.rs
```diff
@@ -3000,6 +3004,33 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             if let Some(command) = self.state.set_vertex_buffers() {
                 pre.issue(command);
             }
+            // Note: all VS resources past the `old_attribute_buffer_index` have to be re-bound
```
grovesNL: What prevents this from happening with CS buffers too?

kvark: the "attribute_index" is for vertex buffers only, so it's only relevant for VS stage

src/backend/metal/src/command.rs
kvark: The one in resources is the old bound buffer id. If it wasn't bound and the coming pipeline needs it, we still need to bind it, I think. `self.state.push_ps_constants(index) ` will set it to `Some`

src/backend/metal/src/device.rs
kvark: there is another check in the loop below that catches this more robustly

kvark:
I believe the concerns are addressed. Let's proceed with patches if not :)
bors r=grovesNL

PR #2374, opened 2018-09-01
Porting dx12 backend to use d3d12

msiglreith:
Starting to sit dx12 on top of the new d3d12 crate:
* `native` module renamed to `resource`, `d3d12` is included as `native` (might just call it d3d12 at the end once we got fully rid of winapi::um::d3d12)
* explicit `destroy` calls in comparison to internal COM reference counting

TODO: 
* Larger portions like graphics pipeline creation still need to be done
* Reiterate a few of the d3d12 functions (e.g newtypes for descriptor registers+set)

Tested with our examples


kvark:
Do you want to finish the TODO before merging?

kvark:
Also, what would be the best way to review this. Should we take a look at particular modules (with an assumption that everything else is just renaming/moving)?

msiglreith:
> Do you want to finish the TODO before merging?

nop, d3d12-rs isn't ready yet.

>  Should we take a look at particular modules (with an assumption that everything else is just renaming/moving)?

I think so, yes. There isn't any logic change beside the introduction of the destroy functions.

msiglreith:
ping

msiglreith:
thanks!
bors r+

kvark:
R=me, not you :P

msiglreith:
bors r=kvark
🙈

PR #2377, opened 2018-09-04
simplify quad resize logic

### LINKED ISSUE #2326: Simplify the resize logic of Quad example
The resize logic is super generic currently. It takes into account the fact that swapchain may need a different format, which forces the pipeline states to be recreated, then render passes, framebuffers, etc - imposes a ton of headache. I don't think it's worth the complexity of the code. This is our first and simplest example, we can still have other examples showing a more complicated resize functionality. For the Quad, it should be sufficient to simply assert that the format PSO was created for is still supported.

jackmott:
Fixes #2326

First pass at simplifying logic, would like review here before tightening it up more.
Some logic is repeated, should it be pulled out into a function or left as is so that the logic is linear down the page?

Anything I am missing?

Have not run rustfmt yet, will make that a separate commit.

PR checklist:
- [x] tested examples with the following backends:vulkan, gl

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -548,24 +548,19 @@ fn main() {
             }
         });
 
+        // Window was resized so we must recreate swapchain and framebuffers
         if recreate_swapchain {
             device.wait_idle().unwrap();
 
             let (caps, formats, _present_modes) = surface.compatibility(&mut adapter.physical_device);
-                println!("formats: {:?}", formats);
-                let format = formats.
-                    map_or(f::Format::Rgba8Srgb, |formats| {
-                        formats
-                            .iter()
-                            .find(|format| format.base_format().1 == ChannelType::Srgb)
-                            .map(|format| *format)
-                            .unwrap_or(formats[0])
-            });
-            
+            // Verify that previous format still exists so we may resuse it.
+            assert!(formats.iter().any(|fs| fs.contains(&format)));         
+
             let swap_config = SwapchainConfig::from_caps(&caps, format);
             println!("{:?}", swap_config);            
             let extent = swap_config.extent.to_extent();
             
+            // Clean up the old framebuffers, images and swapchain
             for framebuffer in framebuffers {
                 device.destroy_framebuffer(framebuffer);
             }
@@ -574,8 +569,7 @@ fn main() {
             }
             device.destroy_swapchain(swap_chain);
 
-            let (new_swap_chain, new_backbuffer) = device.create_swapchain(&mut surface, swap_config, None);
-            
+            let (new_swap_chain, new_backbuffer) = device.create_swapchain(&mut surface, swap_config, None);            
             backbuffer = new_backbuffer;
             swap_chain = new_swap_chain;
 
@@ -610,9 +604,7 @@ fn main() {
             };
 
             framebuffers = new_framebuffers;
-            frame_images = new_frame_images;
-            
-           
+            frame_images = new_frame_images;                       
             viewport.rect.w = extent.width as _;
             viewport.rect.h = extent.height as _;
             recreate_swapchain = false;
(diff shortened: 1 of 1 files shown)


examples/quad/main.rs
```diff
@@ -727,13 +509,195 @@ fn swapchain_stuff(
...
+                    }
+                    _ => (),
+                }
+            }
+        });
+
+        if recreate_swapchain {
+            device.wait_idle().unwrap();
+
+            let (caps, formats, _present_modes) = surface.compatibility(&mut adapter.physical_device);
```
kvark: Since we aren't re-creating the pipeline state and render passes here, let's just `assert_eq!` on the old format being available.

jackmott: like this? 
```rust
  let (caps, formats, _present_modes) = surface.compatibility(&mut adapter.physical_device);
  assert_eq!(formats.iter().any(|fs| fs.iter().any(|f| *f == format)),true);
```

kvark: yes, but just `assert!(x)` instead of `assert_eq!(x, true)` :)

jackmott: ha, of course.

examples/quad/main.rs
```diff
@@ -47,13 +49,30 @@ struct Vertex {
...
 
 const QUAD: [Vertex; 6] = [
-    Vertex { a_Pos: [ -0.5, 0.33 ], a_Uv: [0.0, 1.0] },
-    Vertex { a_Pos: [  0.5, 0.33 ], a_Uv: [1.0, 1.0] },
-    Vertex { a_Pos: [  0.5,-0.33 ], a_Uv: [1.0, 0.0] },
-
-    Vertex { a_Pos: [ -0.5, 0.33 ], a_Uv: [0.0, 1.0] },
-    Vertex { a_Pos: [  0.5,-0.33 ], a_Uv: [1.0, 0.0] },
-    Vertex { a_Pos: [ -0.5,-0.33 ], a_Uv: [0.0, 0.0] },
+    Vertex {
```
ZeGentzy: The old formatting was intentional. Can we add `#[cfg_attr(rustfmt, rustfmt_skip)]`?

examples/quad/main.rs
```diff
@@ -21,21 +21,23 @@ extern crate glsl_to_spirv;
...
+    buffer, command, format as f, image as i, memory as m, pass, pool, pso, window::Extent2D,
+};
+use hal::{Backbuffer, DescriptorPool, FrameSync, Primitive, SwapchainConfig};
+use hal::{Device, Instance, PhysicalDevice, Surface, Swapchain};
 
 use std::fs;
 use std::io::{Cursor, Read};
 
-const DIMS: Extent2D = Extent2D { width: 1024, height: 768 };
+const DIMS: Extent2D = Extent2D {
```
ZeGentzy: I liked the old formatting here.

examples/quad/main.rs
jackmott: agree, rustfmt skip was made for vertex array literals =)

kvark:
thanks everyone!
bors r+

kvark:
ha, interestingly @ZeGentzy review doesn't show up as one big red cross for me

kvark:
bors delegate=ZeGentzy

ZeGentzy:
Neither for me, funky.
Anyways.
bors r+

ZeGentzy:
Opps, @jackmott, you need to remove the question mark on the end of the skips. So ` #[cfg_attr(rustfmt, rustfmt_skip)]? ` -> ` #[cfg_attr(rustfmt, rustfmt_skip)]?`

jackmott:
Thanks, sorry about that.

ZeGentzy:
No worries.
bors r+

Issue #2059, opened 2018-05-21
Support composite specialization constants


grovesNL:
We support [basic data types at the moment](https://github.com/gfx-rs/gfx/blob/master/src/hal/src/pso/mod.rs#L223) for specialization constants at the moment. However we should also determine how to support composites such as structs, vectors, matrices, etc. (which are supported by Vulkan/SPIR-V as far as I know).

kvark:
As a sub-issue here, we need to support named specialization constants on Metal for vkQuake.

PR #2381, opened 2018-09-06
Raw specialization data

kvark:
Fixes a large portion of CTS "dEQP-VK.pipeline.spec_constant" group. The compute sub-part has 83 successes and 38 failures. The graphics part seems to suffer from a descriptor allocation issue, which is unrelated.
Related to https://github.com/gfx-rs/gfx/issues/2059#issuecomment-416439423

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -216,30 +216,24 @@ impl Device {
     fn specialize_ast(
         &self,
         ast: &mut spirv::Ast<glsl::Target>,
-        specializations: &[pso::Specialization],
+        specialization: pso::Specialization,
     ) -> Result<(), d::ShaderError> {
         let spec_constants = ast
             .get_specialization_constants()
             .map_err(gen_unexpected_error)?;
 
         for spec_constant in spec_constants {
-            if let Some(constant) = specializations
+            if let Some(constant) = specialization.constants
                 .iter()
                 .find(|c| c.id == spec_constant.constant_id)
             {
                 // Override specialization constant values
-                unsafe {
-                    let value = match constant.value {
-                        pso::Constant::Bool(v) => v as u64,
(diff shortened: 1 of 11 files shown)


kvark:
Hmm, actually now I wonder if a constant should just be `&[u8]` with an ID. This way we don't need to take care of a separate `data` array on HAL side, and it's not going to be any more complex on Vulkan backend side.

src/backend/gl/src/device.rs
```diff
@@ -216,30 +216,25 @@ impl Device {
...
-                        pso::Constant::U32(v) => v as u64,
-                        pso::Constant::U64(v) => v,
-                        pso::Constant::I32(v) => *(&v as *const _ as *const u64),
-                        pso::Constant::I64(v) => *(&v as *const _ as *const u64),
-                        pso::Constant::F32(v) => *(&v as *const _ as *const u64),
-                        pso::Constant::F64(v) => *(&v as *const _ as *const u64),
-                    };
-                    ast.set_scalar_constant(spec_constant.id, value).map_err(gen_unexpected_error)?;
-                }
+                // Override specialization constant values
```
grovesNL: This comment has been duplicated now

grovesNL:
I guess we could still have an enum with `From` implemented or similar, if we wanted to provide some common type conversions.

kvark:
Filed #2382 to follow up with a nicer API. Not seeing a clear way forward there, and the current state at least works.
bors r=grovesNL

kvark:
CTS update: we pretty much nail down the specialization category of tests, but with a few caveats:
  - arrays are not supported for specialization constants (likely a SPIRV-Cross issue)
  - vertex shaders are generated with `void` return type (likely a SPIRV-Cross issue)
  - a workaround for #2383 is needed

PR #2384, opened 2018-09-06
[mtl] unflatten descriptor pools

### LINKED ISSUE #2383: Metal descriptor pools don't take shader stages into account
It appears that with #2300 we no longer know ahead of time how many descriptors are needed. The pool creation isn't receiving any information about shader stages, so we can't duplicate each entry properly per stage... As a result, some of the CTS sections (in specialization constants, see #2381) fail because they can't allocate sufficient number of descriptors.

There are two ways to address this:
  1. as a hack, just multiply the count by 2. This would work for descriptors that are either used in graphics or compute, but not in both.
  2. revert #2300 logic in some way, or at least re-think the layout of descriptor pools

kvark:
Partially reverts #2300
Fixes #2383

PR checklist:

Tested on Dota2 and some CTS categories.

### DIFF
--- a/src/backend/metal/src/native.rs
+++ b/src/backend/metal/src/native.rs
@@ -161,6 +161,15 @@ impl<T> ResourceData<T> {
     }
 }
 
+impl<T: Copy + Ord> ResourceData<Range<T>> {
+    pub fn expand(&mut self, point: ResourceData<T>) {
+        //TODO: modify `start` as well?
+        self.buffers.end = self.buffers.end.max(point.buffers);
+        self.textures.end = self.textures.end.max(point.textures);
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/native.rs
```diff
@@ -159,6 +159,12 @@ impl<T> ResourceData<T> {
             samplers: fun(&self.samplers),
         }
     }
+
+    pub fn consume<V, F: FnMut(&mut T, V)>(&mut self, other: ResourceData<V>, mut fun: F) {
```
grovesNL: This is always passed the same function, could we just put the logic here instead?

kvark:
Thanks for the review! `consume` is fixed now
bors r=grovesNL

PR #2387, opened 2018-09-07
Metal multi-pass support

kvark:
Fixes vkQuake:
gfx-quake

PR checklist:

~~TODO: need to fix Dolphin~~

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -4,7 +4,7 @@ use {
     validate_line_width,
 };
 use {conversions as conv, command, native as n};
-use internal::FastStorageMap;
+use internal::{Channel, FastStorageMap};
 use native;
 use range_alloc::RangeAllocator;
 
@@ -15,7 +15,7 @@ use std::collections::hash_map::Entry;
 use std::ops::Range;
 use std::path::Path;
 use std::sync::Arc;
-use std::{cmp, mem, ptr, slice, thread, time};
+use std::{cmp, iter, mem, ptr, slice, thread, time};
 
 use hal::{self, error, image, pass, format, mapping, memory, buffer, pso, query};
 use hal::device::{BindError, OutOfMemory, FramebufferError, ShaderError};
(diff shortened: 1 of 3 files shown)


kvark:
Everything works now :tada: , and this PR should be ready to land.
A small perf regression is expected because of disabled RP caching, but this is the only path forward now.

src/backend/metal/src/device.rs
```diff
@@ -1128,7 +1189,7 @@ impl hal::Device<Backend> for Device {
             None => {
                 // TODO: This is a workaround for what appears to be a Metal validation bug
                 // A pixel format is required even though no attachments are provided
-                if pass_descriptor.main_pass.attachments.is_empty() {
+                if subpass.colors.is_empty() {
```
grovesNL: Should we check `depth_stencil` here too?

src/backend/metal/src/device.rs
```diff
@@ -1366,25 +1414,15 @@ impl hal::Device<Backend> for Device {
...
-        descriptor.set_render_target_array_length(extent.depth as NSUInteger);
-
-        let mut inner = n::FramebufferInner {
-            extent,
-            aspects: format::Aspects::empty(),
-            colors: SmallVec::new(),
-            depth_stencil: None,
-        };
-
+        /*
```
grovesNL: Should this block just be removed entirely, or are we planning to revisit it soon?

src/backend/metal/src/device.rs
```diff
@@ -860,10 +859,69 @@ impl hal::Device<Backend> for Device {
...
+                                self.private_caps.map_format(
+                                    attachments[id].format.expect("No depth-stencil format provided")
+                                ).expect("Unable to map depth-stencil format!")
+                            }),
+                    },
+                }
+            })
+            .collect();
+
+        // sprinkle load operations
```
grovesNL: This makes sense, but might be useful to expand to the comments here a bit just to explain the intent. Also where do we distinguish between `loadOp` and `stencilLoadOp` (and store) for mixed depth stencil attachments?

kvark:
Thanks for the review! All notes are addressed now.
Bors r=grovesNL

PR #2390, opened 2018-09-08
[mtl] obey push constant ranges

kvark:
Fixes a validation error happening in vkQuake because we always bind the whole push constant array. This PR makes us only bind the relevant slice (which should also reduce the data passed to GPU), and avoid doing so if we know in advance that more data has to come in.

Tested on vkQuake and the relevant CTS section.

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -312,11 +312,11 @@ impl State {
                     index: 0,
                     samplers: &resources.samplers[..],
                 };
-                let com_push_constants = resources.push_constants_buffer_id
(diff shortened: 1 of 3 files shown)


kvark:
Actually, when going further there is still a validation error related to push constant sizes. I'm going to look at this problem more tomorrow, we can leave PR open for now.

kvark:
Looks like something (SPIRV/SPIRV-Cross/MSL) rounds up the expected push constant buffer size, so now we do the same on our side to provide enough data. vkQuake runs without validation errors :tada:

kvark:
Thank you! TODO is added now.
bors r=grovesNL

PR #2347, opened 2018-08-26
[NOMERGE] Merge window creation, attempt 2

ZeGentzy:
I've been sitting on this for a while, so here it is. Attempt 1: https://github.com/gfx-rs/gfx/pull/2136 

Signed-off-by: Hal Gentz <zegentzy@protonmail.com>

Fixes #1619
PR checklist:

- [x] Implement on X11

Depends on: https://github.com/tomaka/glutin/pull/1058 https://github.com/tomaka/winit/pull/627  https://github.com/tomaka/glutin/pull/1029

PR #2393, opened 2018-09-08
Backports from Merge Window Creation 2

ZeGentzy:
An assortment of changes done while working on this: https://github.com/gfx-rs/gfx/pull/2347

Backporting things like `GlContainer`, even if they do nothing, will decrease the # of accidental merge conflicts between master and #2347 , which will make my life easier.

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -88,6 +88,8 @@ pub enum Command {
     ClearBufferColorI(DrawBuffer, [i32; 4]),
     /// Clear depth-stencil drawbuffer of bound framebuffer.
     ClearBufferDepthStencil(Option<pso::DepthValue>, Option<pso::StencilValue>),
+    /// Clear the currently bound texture with the given color.
+    ClearTexture([f32; 4]),
 
     /// Set list of color attachments for drawing.
     /// The buffer slice contains a list of `GLenum`.
@@ -209,7 +211,7 @@ pub struct RawCommandBuffer {
     pub(crate) id: u64,
     individual_reset: bool,
 
-    fbo: n::FrameBuffer,
+    fbo: Option<n::FrameBuffer>,
     /// The framebuffer to use for rendering to the main targets (0 by default).
     ///
     /// Use this to set the framebuffer that will be used for the screen display targets created
@@ -231,7 +233,7 @@ pub struct RawCommandBuffer {
 
 impl RawCommandBuffer {
     pub(crate) fn new(
-        fbo: n::FrameBuffer,
+        fbo: Option<n::FrameBuffer>,
         limits: Limits,
         memory: Arc<Mutex<BufferMemory>>,
     ) -> Self {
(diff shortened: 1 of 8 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -445,6 +444,10 @@ impl CommandQueue {
 
                 self.share.context.ClearBufferfi(target, 0, depth, stencil as _);
             }
+            com::Command::ClearTexture(color) => unsafe {
+                self.share.context.ClearColor(color[0], color[1], color[2], color[3]);
```
kvark: this isn't clearing the currently bound texture, it's clearing the framebuffer

ZeGentzy: Ah, I see. I got no clue why it still worked, I guess something else might have bound the texture to the framebuffer, I guess. 

Replaced with `unimplemented!()`

src/backend/gl/src/command.rs
```diff
@@ -629,23 +631,33 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
-        self.push_cmd(Command::BindTargetView(gl::DRAW_FRAMEBUFFER, gl::COLOR_ATTACHMENT0, view));
-        self.push_cmd(Command::SetDrawColorBuffers(1));
-
-        match image.channel {
-            ChannelType::Unorm | ChannelType::Inorm | ChannelType::Ufloat |
-            ChannelType::Float | ChannelType::Srgb | ChannelType::Uscaled |
-            ChannelType::Iscaled => self.push_cmd(Command::ClearBufferColorF(0, unsafe { color.float32 })),
-            ChannelType::Uint => self.push_cmd(Command::ClearBufferColorU(0, unsafe { color.uint32 })),
-            ChannelType::Int => self.push_cmd(Command::ClearBufferColorI(0, unsafe { color.int32 })),
+        if let Some(fbo) = self.fbo {
```
msiglreith: let's `match` instead of `if .. else`

src/backend/gl/src/info.rs
```diff
@@ -124,10 +124,16 @@ fn get_string(gl: &gl::Gl, name: gl::types::GLenum) -> &'static str {
     }
 }
 
-fn get_usize(gl: &gl::Gl, name: gl::types::GLenum) -> usize {
+fn get_usize(gl: &GlContainer, name: gl::types::GLenum) -> Result<usize, Error> {
     let mut value = 0 as gl::types::GLint;
     unsafe { gl.GetIntegerv(name, &mut value) };
-    value as usize
+
+    let err = Error::from_error_code(unsafe { gl.GetError() });
```
msiglreith: do these calls fail at all?

src/backend/gl/src/info.rs
```diff
@@ -295,13 +301,13 @@ impl Info {
...
 /// capabilities.
-pub fn query_all(gl: &gl::Gl) -> (Info, Features, LegacyFeatures, Limits, PrivateCaps) {
+pub(crate) fn query_all(gl: &GlContainer) -> (Info, Features, LegacyFeatures, Limits, PrivateCaps) {
     use self::Requirement::*;
     let info = Info::get(gl);
 
     let mut limits = Limits {
-        max_texture_size: get_usize(gl, gl::MAX_TEXTURE_SIZE),
-        max_texel_elements: get_usize(gl, gl::MAX_TEXTURE_BUFFER_SIZE),
+        max_texture_size: get_usize(gl, gl::MAX_TEXTURE_SIZE).unwrap_or(0),
```
msiglreith: should use the actual limits here

src/backend/gl/src/lib.rs
```diff
@@ -39,6 +39,25 @@ mod window;
...
+    fn make_current(&self) {
+        // Unimplemented
+    }
+}
+
+impl Deref for GlContainer {
+    type Target = gl::Gl;
+    fn deref(&self) -> &gl::Gl {
+        #[cfg(feature = "glutin")]
+        self.make_current();
```
msiglreith: Calling `make_current` on each context access is worrisome.
Couldn't we just switch the active contexts before and after presenting under the constraint that everything must happen in the 'main' thread?

src/backend/gl/src/lib.rs
```diff
@@ -197,6 +264,18 @@ impl PhysicalDevice {
         }
         let name = info.platform_name.renderer.into();
 
+        // Because some drivers don't actually support this stuff, but report
```
msiglreith: why not do these checks inside `query_all`?

src/backend/gl/src/info.rs
ZeGentzy: Yes, I think it was failing with `MAX_TEXTURE_BUFFER_SIZE`, but I honestly can't remember.

src/backend/gl/src/info.rs
ZeGentzy: What would be the actual limit for a failed `MAX_TEXTURE_BUFFER_SIZE`?

src/backend/gl/src/lib.rs
ZeGentzy: Ok, I've moved it into `query_all`.

src/backend/gl/src/lib.rs
ZeGentzy: Most of the time calling `make_current` won't do anything, if the context is already current. Currently we call `is_current` to check, but if that proves too costly, we could make a thread_local uint containing the id (assigned by us) of the currently active context.

Switching contexts might prove useful in the future for other things, like emulating FBOs in OpenGL 2.1 (I'm not really sure how this would work out yet, so that's a future me problem). In such case, switching from and back to the `main` context might not end up being the best thing (or it might, who knows).

Until switching contexts proves too costly, I'd like to avoid adding additional constraints.

src/backend/gl/src/info.rs
msiglreith: buffer textures are 3.something, so 0 would be fine in this case as we don't support it at all for lower version. texture size should be supported for all and just in case for weird stuff we should use the minimum limit guarenteed by GL.

ZeGentzy: Will do. :+1:

Issue #2364, opened 2018-08-31
metal: Can't generate mipmaps for cubemaps using vkCmdBlitImage because it's not implemented


zeux:
We use vkCmdBlitImage to do progressive blitting from mip 0 to mip 1 to mip 2 etc., including cubemaps. Getting this panic:

```
thread '<unnamed>' panicked at 'not yet implemented', /Users/travis/.cargo/git/checkouts/gfx-e86e7f3ebdbc4218/4a1b551/src/backend/metal/src/internal.rs:382:30
stack backtrace:
  0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
  1: std::sys_common::backtrace::print
  2: std::panicking::default_hook::{{closure}}
  3: std::panicking::default_hook
  4: std::panicking::rust_panic_with_hook
  5: std::panicking::begin_panic
  6: gfx_backend_metal::internal::ImageBlitPipes::create
  7: gfx_backend_metal::internal::ImageBlitPipes::get::{{closure}}
  8: <storage_map::StorageMap<L, std::collections::hash::map::HashMap<K, V, S>>>::get_or_create_with
  9: gfx_backend_metal::internal::ImageBlitPipes::get
 10: <gfx_backend_metal::command::CommandBuffer as gfx_hal::command::raw::RawCommandBuffer<gfx_backend_metal::Backend>>::blit_image
 11: portability_gfx::impls::gfxCmdBlitImage
 12: _ZN3RBX8Graphics19DeviceContextVulkan15generateMipmapsEPNS0_7TextureE
 13: _ZN3RBX8Graphics6EnvMap10fullUpdateEPNS0_13DeviceContextE
 14: _ZN3RBX8Graphics12SceneManager11renderBeginEPNS0_13DeviceContextEjjRKNS0_12RenderCameraE
 15: _ZN3RBX8Graphics12SceneManager11renderSceneEPNS0_13DeviceContextEPNS0_11FramebufferERKNS0_12RenderCameraE
 16: _ZN3RBX8Graphics10RenderView17renderPerformImplEdPNS0_11FramebufferENS_10RenderTypeE
 17: _ZN15RenderScheduler18renderIntoViewBaseEv
 18: _ZN11QOgreWidget5eventEP6QEvent
 19: _ZN19QApplicationPrivate13notify_helperEP7QObjectP6QEvent
 20: _ZN12QApplication6notifyEP7QObjectP6QEvent
[...]

PR #2398, opened 2018-09-11
[mtl] build internal shaders with debug info

kvark:
Helps investigating #2364
PR checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,18 +19,18 @@ matrix:
     - env: MACOSX_DEPLOYMENT_TARGET=10.9
       os: osx
       rust: stable
-      osx_image: xcode9
+      osx_image: xcode10
       compiler: clang
     - env: MACOSX_DEPLOYMENT_TARGET=10.9
       os: osx
       rust: nightly
-      osx_image: xcode9
+      osx_image: xcode10
       compiler: clang
 
     # iPhoneOS 64bit
     - env: TARGET=aarch64-apple-ios
       os: osx
-      osx_image: xcode9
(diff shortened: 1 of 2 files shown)


grovesNL:
This doesn't seem to work on Travis:
```
--- stderr
metal: error: unknown argument: '-MO'
thread 'main' panicked at 'shader compilation failed', src/backend/metal/build.rs:72:13
```
But we rely on Travis for building portability... so not sure how to proceed.

Michael-Lfx:
Is the PR abandoned? cc @kvark

kvark:
Please feel free to complete it.

kvark:
@Michael-Lfx the only thing missing here is adding a check at run-time for XCode 10 before adding those debug flags. Similar but different flags exist on XCode 9 as well, so using those appropriately would be good too.

kvark:
I really procrastinate when it comes to build systems, sorry. Don't know what I'd be doing without @grovesNL help, but this PR is still kept in the air.

kvark:
@grovesNL any idea on how to finish this? :)
just trying to clean up our open PRs

GabrielMajeri:
@kvark According to [this page](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#xcode), GH Actions already has Xcode 11 by default, so we can drop the `.travis.yml` changes.

As for the build flags, I'm guessing they can stay the same?

kvark:
Right, that sounds reasonable. Note that travis YML is no more.
These changes now need to go through our github action scripts.

Issue #2400, opened 2018-09-12
Metal swapchain lost on creation


kvark:
[Filament](https://github.com/google/filament) samples fail to start because we return OUT_OF_DATE error on one of the first acquires. They aren't currently handling this, but even though they should, it's not good that we return the error right off the bat. We need to look into what's going on and try to make this smoother.

kvark:
Looks like it only happens with "Wait" acquire mode. It wasn't tested as much as the "Oldest", so we should probably just switch to the latter by default.

PR #2401, opened 2018-09-12
[mtl] Filament compatibility fixes

kvark:
gfx-filament-pbr-small

There are still occasional asserts on frame acquisition, so leaving #2400 open for now.

PR checklist:

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -48,7 +48,6 @@ use core_graphics::base::CGFloat;
 use core_graphics::geometry::CGRect;
 use foreign_types::ForeignTypeRef;
 use objc::runtime::{Class, Object};
-#[cfg(target_os = "ios")]
 use objc::runtime::{BOOL, YES};
 use parking_lot::{Condvar, Mutex};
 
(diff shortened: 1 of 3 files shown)


kvark:
Problem - RPCS3 regresses with this, it doesn't like that we don't report OUT_OF_DATE on `Oldest` acquiring.

src/backend/metal/src/lib.rs
```diff
@@ -194,18 +193,31 @@ impl Instance {
                 panic!("window does not have a valid contentView");
             }
 
+            let existing: *mut Object = msg_send![view, layer];
             let class = Class::get("CAMetalLayer").unwrap();
```
grovesNL: Unrelated but probably should use [`class!`](https://github.com/SSheldon/rust-objc/blob/master/src/lib.rs#L13) here instead

src/backend/metal/src/window.rs
```diff
@@ -152,16 +164,23 @@ impl Drop for Swapchain {
...
             .inner
             .lock();
-        assert!(!frame.available);
-        frame.available = true;
-        frame.drawable
-            .take()
-            .expect("Drawable has not been acquired!")
+        assert!(!frame.available && frame.linked);
+
+        match frame.drawable.take() {
```
grovesNL: Could we use `ok_or` here?

kvark: we don't want to set the `available` state unconditionally, so I don't think `ok_or` would help here

src/backend/metal/src/lib.rs
kvark: good point! hopefully, this is only a one-time operation

kvark:
thanks for the review!
bors r=grovesNL

PR #2402, opened 2018-09-14
[mtl] keep borrowed clear value longer

kvark:
Intended to fix https://github.com/szeged/webrender/issues/216
Note: this is a little scary of an issue. Doesn't appear that we have any more of them, but we need to keep an eye on those cases.

PR checklist:


kvark:
This is important and fairly trivial. Assuming it compiles:)
bors r+

PR #2404, opened 2018-09-17
dx12: support individual command buffer reset

### LINKED ISSUE #2140: dx12: Indivial command buffer reset
Command allocators in D3D12 need to be reset to avoid running out of memory over time. Resetting will also not free the allocated resources. When using individual command buffer reset we will run out of memory over time as we cannot reset the allocator. Possible solution would be one allocator per command list in that case. Additionally handling the release resources flag by destructing and rebuilding the individual allocators.

caitp:
dx12::RawCommandPool now holds a collection of ID3D12CommandAllocator,
introducing a new one each time a new command buffer is allocated.

dx12::RawCommandPool::reset() now performs a reset() on each allocator in
the collection, mimicking the prior single-allocator behaviour.

dx12::CommandBuffer::reset() now handles the `release_resources` flag by
reseting its allocator.

Fixes #2140 

PR checklist:
- [ ] `make reftests` succeeds*
- [x] tested examples with the following backends: **dx12**
- [ ] `rustfmt` run on changed code*

*make reftests fails here:

```
Warding DX12:
        Scene 'basic':
thread 'main' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`', src\warden\src\gpu.rs:1152:9
```

I haven't looked too closely into this, but it also fails with the patch reverted. Other than this, there are 15 passing reftests.

---

*rustfmt --check requests many changes (not limited to things inserted in this patch), but I believe the style of the code added in this patch is consistent with the rest of the file.

---

TODO: add a test (I'm not sure how easy it will be to simulate an EOM reliably to verify that it doesn't happen with the patch applied, though).

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -1044,7 +1044,10 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         }
     }
 
-    fn reset(&mut self, _release_resources: bool) {
+    fn reset(&mut self, release_resources: bool) {
+        if release_resources {
+            unsafe { self.allocator.Reset(); }
+        }
         self.reset();
     }
 
--- a/src/backend/dx12/src/pool.rs
+++ b/src/backend/dx12/src/pool.rs
@@ -7,8 +7,13 @@ use hal::{command, pool};
 use native::command_list::CmdListType;
 use {native, Backend, Shared};
 
+pub enum CommandPoolAllocator {
+    Shared(native::CommandAllocator),
+    Individual(Vec<native::CommandAllocator>),
+}
+
 pub struct RawCommandPool {
-    pub(crate) allocators: Vec<native::CommandAllocator>,
+    pub(crate) allocator: CommandPoolAllocator,
     pub(crate) device: native::Device,
     pub(crate) list_type: CmdListType,
     pub(crate) shared: Arc<Shared>,
@@ -16,12 +21,20 @@ pub struct RawCommandPool {
 
 impl RawCommandPool {
     fn create_command_list(&mut self) -> (native::GraphicsCommandList, native::CommandAllocator) {
-        let (command_allocator, hr) = self.device.create_command_allocator(self.list_type);
+        let command_allocator = match self.allocator {
+            CommandPoolAllocator::Shared(ref allocator) => allocator.clone(),
+            CommandPoolAllocator::Individual(ref mut allocators) => {
+                let (command_allocator, hr) = self.device.create_command_allocator(self.list_type);
 
-        // TODO: error handling
-        if !SUCCEEDED(hr) {
-            error!("error on command allocator creation: {:x}", hr);
-        }
+                // TODO: error handling
+                if !SUCCEEDED(hr) {
+                    error!("error on command allocator creation: {:x}", hr);
+                }
+
+                allocators.push(command_allocator);
+                command_allocator
+            }
+        };
 
         // allocate command lists
         let (command_list, hr) = self.device.create_graphics_command_list(
@@ -39,14 +52,19 @@ impl RawCommandPool {
         // But only one command list can be recording for each allocator
         let _hr = command_list.close();
 
-        self.allocators.push(command_allocator);
-
         (command_list, command_allocator)
     }
 
     pub(crate) fn destroy(self) {
-        for allocator in self.allocators {
-            unsafe { allocator.destroy(); }
+        match self.allocator {
+            CommandPoolAllocator::Shared(ref allocator) => unsafe { allocator.destroy(); },
+            CommandPoolAllocator::Individual(ref allocators) => {
+                for allocator in allocators.iter() {
+                    unsafe {
+                        allocator.destroy();
+                    }
+                }
+            }
         }
     }
 }
@@ -56,11 +74,16 @@ unsafe impl Sync for RawCommandPool {}
 
 impl pool::RawCommandPool<Backend> for RawCommandPool {
     fn reset(&mut self) {
-        self.allocators.iter_mut().for_each(|allocator| {
-            unsafe {
-                allocator.Reset();
+        match self.allocator {
+            CommandPoolAllocator::Shared(ref allocator) => unsafe { allocator.Reset(); },
+            CommandPoolAllocator::Individual(ref mut allocators) => {
+                for allocator in allocators.iter_mut() {
+                    unsafe {
+                        allocator.Reset();
+                    }
+                }
             }
-        })
+        }
     }
 
     fn allocate(&mut self, num: usize, level: command::RawLevel) -> Vec<CommandBuffer> {
--- a/src/backend/dx12/src/pool.rs
+++ b/src/backend/dx12/src/pool.rs
@@ -8,20 +8,25 @@ use native::command_list::CmdListType;
 use {native, Backend, Shared};
 
 pub struct RawCommandPool {
-    pub(crate) raw: native::CommandAllocator,
+    pub(crate) allocators: Vec<native::CommandAllocator>,
     pub(crate) device: native::Device,
     pub(crate) list_type: CmdListType,
     pub(crate) shared: Arc<Shared>,
 }
 
 impl RawCommandPool {
-    fn create_command_list(&mut self) -> native::GraphicsCommandList {
+    fn create_command_list(&mut self) -> (native::GraphicsCommandList, native::CommandAllocator) {
+        let (command_allocator, hr) = self.device.create_command_allocator(self.list_type);
+
         // TODO: error handling
+        if !SUCCEEDED(hr) {
+            error!("error on command allocator creation: {:x}", hr);
+        }
 
         // allocate command lists
         let (command_list, hr) = self.device.create_graphics_command_list(
             self.list_type,
-            self.raw,
+            command_allocator,
             native::PipelineState::null(),
             0,
         );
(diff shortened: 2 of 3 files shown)


kvark:
Thank you for the PR!
Let's wait a bit for @msiglreith to have a look at it.

In the meantime, individual command buffer reset is only needed for command pools created with [RESET_INDIVIDUAL](https://github.com/gfx-rs/gfx/blob/1d8b50e55d3fb2c3923e39c19fc735ead7cc0318/src/hal/src/pool.rs#L23) flag. We aren't currently checking for it. Would we get a simpler code path (in terms of run-time costs mostly, since implementation wise it will only make more code) for those command pools that don't need individual resets?

caitp:
> In the meantime, individual command buffer reset is only needed for command pools created with RESET_INDIVIDUAL flag. We aren't currently checking for it. Would we get a simpler code path (in terms of run-time costs mostly, since implementation wise it will only make more code) for those command pools that don't need individual resets?

so, a few different ways come to mind, what do you prefer? they all seem to make it more complicated than always behaving as though RESET_INDIVIDUAL were specified, though.

1. store the create flags in RawCommandPool, and always use the first-created allocator for all command lists
1. allocator is an enum variant of either a raw allocator, or a vector of allocators
1. create a trait that handles different strategies for the allocators "allocator_for_new_command_list()" and "reset()", I guess --- and store a Box  of the trait object in place of a variant or extra bit flags.
1. the entire RawCommandPool struct is generic, with implementations for `RawCommandPool<ComPtr<d3d12::ID3D12CommandAllocator>>` and `RawCommandPool<Vec<ComPtr<d3d12::ID3D12CommandAllocator>>>`

kvark:
I like the (2). But let's make this work only if there is a promise of run-time advantage (less work to do) for those pools not supporting individual cmdbuf resets.

src/backend/dx12/src/command.rs
```diff
@@ -980,7 +980,10 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
         unsafe { self.raw.Close(); }
     }
 
-    fn reset(&mut self, _release_resources: bool) {
+    fn reset(&mut self, release_resources: bool) {
+        if release_resources {
+            unsafe { self.allocator.Reset(); }
```
msiglreith: - `release_resource` should destroy the allocator if possible as it's the only way in dx12 to regain the memory allocated.
- else we reset the allocator, which should be always valid as in the case of indiviual reset we kind have to assign an own pool to each commandbuffer

src/backend/dx12/src/pool.rs
```diff
@@ -11,22 +11,42 @@ use command::{CommandBuffer};
...
 }
 
 impl RawCommandPool {
-    fn create_command_list(&mut self) -> ComPtr<d3d12::ID3D12GraphicsCommandList> {
+    fn create_command_list(&mut self) -> (ComPtr<d3d12::ID3D12GraphicsCommandList>, ComPtr<d3d12::ID3D12CommandAllocator>) {
+        let command_allocator = {
+            let mut command_allocator: *mut d3d12::ID3D12CommandAllocator = ptr::null_mut();
+
+            let hr = unsafe {
+                self.device.clone().CreateCommandAllocator(
```
msiglreith: (I wish the d3d12 migration PR would have been merged already)

src/backend/dx12/src/pool.rs
```diff
@@ -54,19 +76,24 @@ unsafe impl Sync for RawCommandPool { }
 
 impl pool::RawCommandPool<Backend> for RawCommandPool {
     fn reset(&mut self) {
```
msiglreith: (we really need to provide an equivalent for release resources for pools!)

src/backend/dx12/src/pool.rs
```diff
@@ -54,19 +76,24 @@ unsafe impl Sync for RawCommandPool { }
...
-        unsafe { self.inner.Reset(); }
+        self.allocators.iter_mut().for_each(|allocator| {
```
msiglreith: a normal for loop should be fine here

src/backend/dx12/src/pool.rs
msiglreith: https://github.com/gfx-rs/gfx/issues/2407

msiglreith:
Regarding optimization for indiviual resets let's follow the proposal [here](https://github.com/gfx-rs/gfx/issues/2140#issuecomment-396446497) as we also have to take into account multiple command buffers recorded at the same time.

src/backend/dx12/src/command.rs
caitp: > release_resource should destroy the allocator if possible as it's the only way in dx12 to regain the memory allocated.

What conditions would allow for that safely? `allocator.clone().as_unknown().Release() == 1`?

caitp:
> Regarding optimization for indiviual resets let's follow the proposal here as we also have to take into account multiple command buffers recorded at the same time.

I don't fully understand what the proposal is wanting to do. In this bug, it sounds like we're wanting to only keep multiple allocators if requested --- but it sounds like you want to always support multiple allocators, but only allocate them if needed (if an allocator is being recorded while a new one is being allocated)?

msiglreith:
We basically have to take care of two things in D3D12 for emulating the behavior of Vulkan (beside the large memory preallocations..):

* Releasing the memory of command allocators to avoid OoM (requires allocator release + recreate)
* Supporting to record multiple command buffers at the same time per pool (Vulkan concept)

With respect to the 2nd point we don't really have control over how many allocators we will need. Therefore, I would just store them in a `Vec` and create new allocators on demand (#allocators >= #command buffers currently recording)

For the 1st aspect we also need one pool per command buffer in case of indiviual command buffer reset. If we only have one allocator we wouldn't be able to tell when we can release the underlying allocator. In this case #allocators **must** be equal to the number of command buffers allocated from the HAL/Vulkan pool.

> What conditions would allow for that safely? allocator.clone().as_unknown().Release() == 1?

As mentioned above for the individual reset case (as it is here) the buffer more or less owns the allocator, so it's always safe to release the pool and recreate one. Note that cloning will increment the internal reference counter so we actually end up with the same internal ref count after the `Release` call as before the operation.

msiglreith:
These are actually more or less optimization, so the PR is also fine without this further changes.
I'm just noticing that it's still quite a bit of work to grasp all the details between Vulkan's and D3D12's approach and implement it.

caitp:
This still doesn't release any allocators until the CommandPool and all of its CommandBuffers have been dropped (assuming all the addref/release logic is sound)

caitp:
> As mentioned above for the individual reset case (as it is here) the buffer more or less owns the allocator, so it's always safe to release the pool and recreate one. Note that cloning will increment the internal reference counter so we actually end up with the same internal ref count after the Release call as before the operation.

Right, I wrote that thinking this was RawCommandPool doing this.

Still, RawCommandPool and CommandBuffer both hold a reference to the allocator, so dropping it in the CommandBuffer doesn’t really destroy it. Dropping it twice would leave an invalid ComPtr in RawCommandPool, which would break RawCommandPool::reset();

So, like you say, CommandBuffer could reference back to the pool, and ask the Pool to remove that ComPtr from its list, effectively dropping it (but I’m not sure how speedy all this would be, if that matters).

msiglreith:
If it doesn't work with ComPtrs let's just fallback to pointers. `ComPtr` will be replaced anyway once https://github.com/gfx-rs/gfx/pull/2374 lands. (`WeakPtr` is basically just a pointer which needs to be manually tracked)
I'm not sure what's the best way to handle it, but at some pointer in the future we need to do else the pool will just eat away all out memory :<

caitp:
So, I understand:

1) CommandBuffer::reset() can happen on any different threads
2) CommandBuffer needs to be able to have the Vec<CommandPoolAllocator> drop an allocator

So, we need to rethink this a bit.

1) hold a reference to RawCommandPool (a RefCell or unsafe pointer?), and synchronize access to the Pool's allocators via a mutex
2) use a mpsc::Sender to notify the RawCommandPool's owner thread that it needs to drop an allocator from the vector.

msiglreith:
ehh - let's look what's the result of https://github.com/KhronosGroup/Vulkan-Docs/issues/802 before we start to fiddle with mutexes or lock-free queues :D 

As mentioned we can also postpone it for now, we are also kind of blocked a bit.

kvark:
Is this PR getting stale? I thought there are no blockers atm. Ping @caitp

msiglreith:
Looks fine to me, needs rebasing tho

caitp:
is there a good reason we have explicit `destroy_command_pool` methods, instead of just `drop()` --- and the `destroy()` methods don't borrow mutably?

I ask because it seems there's no guarantee that the command pool will actually be dropped after `destroy_command_pool` is called, and it's not possible to clear the allocators vector because destroy_command_pool() doesn't borrow the pool mutably, so it will have an array of dead allocators which would probably crash if destroy_command_pool is called twice, --- or if anything continues to reference one of the allocators.

it seems safer to just implement Drop for everything, and allow users to control the lifetime of these objects with Rc/Arc if they wish to.

msiglreith:
`WeakPtr` is `Copy`, so we can't implement drop for it,  We could ofc implement it for `CommandPool` but I'm not sure what would be the advantage as we have explicit `Device::destroy_command_pool` calls anyway. Feel free to move towards `destroy(&mut self)` or even `self`.

caitp:
> WeakPtr is Copy, so we can't implement drop for it, We could ofc implement it for CommandPool but I'm not sure what would be the advantage as we have explicit Device::destroy_command_pool calls anyway. Feel free to move towards destroy(&mut self) or even self.

yes, I specifically was referring to the gfx-rs wrappers around the native bindings, because the explicit `destroy` calls technically leave open the possibility of accessing invalid memory, but if we used Drop instead, that should be impossible

caitp:
anyways, rebasing has been done --- we still don't destroy the allocator if `release_resources` is true, because a dead WeakPtr would still be in the RawCommandPool, and might be dereferenced again via destroy_command_pool().

kvark:
bors r=msiglreith

PR #2409, opened 2018-09-18
Don't #[allow(type_alias_bounds)].

eddyb:
https://github.com/rust-lang/rust/pull/54090 will fix false positives (not relevant here) and make it an error in the upcoming Rust 2018 edition, which means that if you want to migrate to Rust 2018, it shouldn't be ignored.

See also https://github.com/rust-lang/rust/issues/49441#issuecomment-421136550 for the decision and some background.


kvark:
Thank you!
bors r+

PR #2416, opened 2018-09-20
[mtl] return an error on Swapchain::present if drawable not found

mtak-:
PR checklist:
- [x] tested examples with the following backends: metal (ios/mac)


kvark:
I need to double check how vkQuake and others like this change.

kvark:
I needed to patch `gfxQueuePresent` to actually pass through the error, and it does happen on RPCS3, but with the patch everything seems to work good, so let's go ahead :+1: 
bors r+

mtak-:
Nice. I should work on getting set up with vkQuake...

PR #2419, opened 2018-09-22
Just some simple error impls.

derekdreery:
To make error handling easier :)

PR checklist:

### DIFF
--- a/src/hal/src/pso/mod.rs
+++ b/src/hal/src/pso/mod.rs
@@ -1,6 +1,6 @@
 //! Raw Pipeline State Objects
 //!
-//! This module contains items used to create and manage Pipelines. 
+//! This module contains items used to create and manage Pipelines.
 
 use {device, pass};
 use std::error::Error;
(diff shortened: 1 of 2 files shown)


derekdreery:
I could update the examples to use

```rust
fn main() -> Result<(), failure::Error>
```

if you like.

src/hal/src/pso/mod.rs
```diff
@@ -145,6 +145,19 @@ impl From<Stage> for ShaderStageFlags {
     }
 }
 
+impl fmt::Display for Stage {
```
kvark: does it make sense to have this implementation? Arguably, it brings nothing extra on top of `Debug`, which is derived

derekdreery: We don't need this, I basically added it because it made the text read better (no random capitals) and because it did no harm. I'm happy to remove it and use the Debug format if you prefer?

derekdreery:
I did this in my local version so I could use `failure` in the `quad` example. Would it be worth including that?

For example, if you `cargo run` in any subdirectory of `quad` it won't find the shader files, because the cwd will be wrong. With failure/context it tells you exactly what it couldn't find, otherwise you just get Os error 2 (not found).

kvark:
> I could update the examples to use

That sounds intriguing, tbh. If you see that this would reduce boilerplate, please go for it ;)

src/hal/src/pso/mod.rs
kvark: The text is only seen on error, it's not something you read every day anyway. In general, I like less code if possible, but we can leave that implementation since you already wrote it.

derekdreery:
It doesn't reduce boilerplate (I think it makes the code *slightly* longer). What it does do is to add contextual information to errors, so e.g. I might get an I/O error when accessing a file, but that doesn't contain info about which file I'm trying to access. By using `failure::Error` and `failure::ResultExt::context` I can add a ring to the error chain with information about which file is the problem, a bit like building your own stack-trace.

derekdreery:
You can almost do all of this without the `failure` crate, but it puts some extra constraints on the `Fail` type (it's version of `std::error::Error`, for example `'static` so you can downcast.

kvark:
Ok, let's not do this for `quad` then :)

PR #2420, opened 2018-09-22
Allow access to extra types.

derekdreery:
This is necessary for me to keep a handle on the backend surface (so
that I can get information about it, for example when its size changes).

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -50,7 +50,7 @@ mod info;
 mod native;
 mod pool;
 mod result;
-mod window;
+pub mod window;
 
 const LAYERS: &'static [&'static str] = &[#[cfg(debug_assertions)]
 "VK_LAYER_LUNARG_standard_validation"];
@@ -292,7 +292,7 @@ impl hal::Instance for Instance {
                         ash::vk::PhysicalDeviceType::DiscreteGpu => DeviceType::DiscreteGpu,
                         ash::vk::PhysicalDeviceType::VirtualGpu => DeviceType::VirtualGpu,
                         ash::vk::PhysicalDeviceType::Cpu => DeviceType::Cpu,
-                    },                    
+                    },
                 };
                 let physical_device = PhysicalDevice {
                     instance: self.raw.clone(),
--- a/src/backend/vulkan/src/window.rs
+++ b/src/backend/vulkan/src/window.rs
@@ -25,7 +25,7 @@ pub struct Surface {
     pub(crate) samples: NumSamples,
 }
 
-pub struct RawSurface {
+pub(crate) struct RawSurface {
     pub(crate) handle: vk::SurfaceKHR,
     functor: ext::Surface,
     pub(crate) instance: Arc<RawInstance>,


src/backend/vulkan/src/lib.rs
```diff
@@ -50,7 +50,7 @@ mod info;
 mod native;
 mod pool;
 mod result;
-mod window;
+pub mod window;
```
kvark: if `RawSurface` is the only type that needs to be visible, let's re-export it instead of making `window` module public

derekdreery: RawSurface is the only one that *doesn't* need to be visible

derekdreery:
I think that the quad example actually uses the concrete `gfx_backend_vulkan::Surface`, because of type inference it doesn't need to be able to explicitly name the type. But if I want to store it in a struct I do have to name the type. I could use trait objects, but I don't think this feels right. I think other backends expose the `Surface` type (at least the dummy backend and gl).

grovesNL:
Could you store `<back::Backend as Backend>::Surface`?

kvark:
@derekdreery `Backend` is the only concrete type that each of the backends needs to provide, in theory. If something can be obtained as `<ConcreteBackend as hal::Backend>::XXX`, then this *is* the recommended way to access the type. With this consideration, I think the PR can be closed.

derekdreery:
Ok let's turn this PR into 2 things

 1. A documentation note on `Backend` that you can access the types using `<.. as ..>::..` notation
 2. Removing `pub` from other types so that the backends are consistent.

does that sound good?

kvark:
@derekdreery sounds excellent! :+1:

derekdreery:
I experimented with this, and I think it only works because the other types are in private modules. Maybe all types could be in private modules, but it would kinda make a mess.

I think the situation will change in rust 2018, so if you plan to update the codebase we could revist then? It may be that all the types on `Backend` will need to be public, I don't understand the new rules yet.

kvark:
I don't understand. What "only works"? There is no need for the concrete types to be public in the backends.

derekdreery:
I mean that if you try to make, say, `PhysicalDevice` private in vulkan/lib.rs, you get

```
error[E0446]: private type `PhysicalDevice` in public interface
   --> src/backend/vulkan/src/lib.rs:819:5
    |
343 | struct PhysicalDevice {
    | - `PhysicalDevice` declared as private
...
819 |     type PhysicalDevice = PhysicalDevice;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak private type
```

kvark:
oh wow, that doesn't seem consistent (as in, Rust way of handling this...)

derekdreery:
Yeah I believe it may be changing for 2018, but I'm not 100%.

kvark:
ping

kvark:
FWIW, we are on Rust 2018 now. @derekdreery any plans on finishing this?

omni-viral:
Rust disallows private types to be exported in any way, including making it associated type for a public type.
This is consistent as public types in private modules are still public, just without visible symbol. You can mske them accessible via export, via type alias, via associated type. You can even use them in public API without letting user name it.
Nothing has changed in 2018 in this regard afaik.

Instead of making all types private I propose do it otherwise and make them public to allow user hack around in backend specific way.

kvark:
> Instead of making all types private I propose do it otherwise and make them public to allow user hack around in backend specific way.

But they can already do that (via the associated types of `Backend`), and exposing more types in public just increases the API surface for no apparent benefit. I.e. any refactoring of those names and location would become a breaking change out of a sudden.

derekdreery:
> > Instead of making all types private I propose do it otherwise and make them public to allow user hack around in backend specific way.
> 
> But they can already do that (via the associated types of `Backend`), and exposing more types in public just increases the API surface for no apparent benefit. I.e. any refactoring of those names and location would become a breaking change out of a sudden.

I agree with this statement, but accessing types through associated types of traits isn't something you do all that often in rust - so I think it should be well documented, with examples. When I get chance I'll change this PR into a doc-change PR.

kvark:
@derekdreery let's open a new PR then. Closing this one as agreed.

PR #2422, opened 2018-09-24
[pre-ll] Add accessors to allow getting a raw GLuint from a texture

magnonellie:
I needed this for submitting frames to my VR compositor. I'm very unsure of the quality of this PR however :stuck_out_tongue:

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -127,6 +127,9 @@ impl ResourceView {
             owned: true,
         }
     }
+    pub fn gl_texture(&self) -> Texture {
+        self.object
+    }
 }
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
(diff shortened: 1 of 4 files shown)


src/backend/gl/src/lib.rs
```diff
@@ -127,6 +127,9 @@ impl ResourceView {
             owned: true,
         }
     }
+    pub fn raw_texture(&self) -> &Texture {
```
kvark: since GL knows this is integer, we might just return it by copy. Also, for clarity we can rename this to `gl_texture`

src/core/src/handle.rs
```diff
@@ -129,6 +129,11 @@ pub struct ShaderResourceView<R: Resources, T>(
     PhantomData<T>
 );
 
+impl<R: Resources, T> ShaderResourceView<R, T> {
+    ///
+    pub fn raw_resource(&self) -> &R::ShaderResourceView { &(self.0).0 }
```
kvark: actually, this isn't a resource, this is a view :)

PR #2428, opened 2018-10-01
Add serde defaults for `SamplerInfo` fields

torkleyy:
I'm not sure if there should be a default filter (currently there is not) or if the default wrap mode should be removed. All the other defaults are equal to what's been used by `SamplerInfo::new`.

### DIFF
--- a/src/core/src/texture.rs
+++ b/src/core/src/texture.rs
@@ -491,12 +491,16 @@ pub struct SamplerInfo {
     /// This bias is added to every computed mipmap level (N + lod_bias). For
     /// example, if it would select mipmap level 2 and lod_bias is 1, it will
     /// use mipmap level 3.
+    #[cfg_attr(feature = "serialize", serde(default = "SamplerInfo::def_lod_bias"))]
     pub lod_bias: Lod,
     /// This range is used to clamp LOD level used for sampling.
+    #[cfg_attr(feature = "serialize", serde(default = "SamplerInfo::def_lod_range"))]
     pub lod_range: (Lod, Lod),
     /// Comparison mode, used primary for a shadow map.
+    #[cfg_attr(feature = "serialize", serde(default))]
     pub comparison: Option<state::Comparison>,
     /// Border color is used when one of the wrap modes is set to border.
+    #[cfg_attr(feature = "serialize", serde(default = "SamplerInfo::def_border"))]
     pub border: PackedColor,
 }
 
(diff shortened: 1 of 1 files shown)


torkleyy:
I'm going to call those default methods from `SamplerInfo::new` so the defaults aren't duplicated.

kvark:
`SamplerInfo` has the liberty to be more opinionated, since it's just a helper. Serde implementation needs to be more strict, it can only use the defaults that *really* make sense as defaults. `WrapMode` isn't one of them.

torkleyy:
Thanks, that's what I thought.

torkleyy:
Done, this is ready for review now.

torkleyy:
Fixed. After this is merged, could you please release a new version?

src/core/src/texture.rs
```diff
@@ -491,12 +491,16 @@ pub struct SamplerInfo {
     /// This bias is added to every computed mipmap level (N + lod_bias). For
     /// example, if it would select mipmap level 2 and lod_bias is 1, it will
     /// use mipmap level 3.
+    #[cfg_attr(feature = "serialize", serde(default = "SamplerInfo::def_lod_bias"))]
     pub lod_bias: Lod,
     /// This range is used to clamp LOD level used for sampling.
+    #[cfg_attr(feature = "serialize", serde(default = "SamplerInfo::def_lod_range"))]
     pub lod_range: (Lod, Lod),
     /// Comparison mode, used primary for a shadow map.
+    #[cfg_attr(feature = "serialize", serde(default = "None"))]
```
kvark: this could be `serde(default)` without any mention of `None`, since it's a natural default

kvark:
Actually, CI complains that `= "None"` is illegal.
I'll be happy to publish an update, assuming this is all you need for serialization, but we'll need to back-port this PR to v0.17 branch before that (and bump the patch version).

src/core/src/texture.rs
torkleyy: Oh right, I've gotten so used to this annoying `= path` pattern I didn't notice ^^

torkleyy:
Oh, so there've been breaking changes and you don't want to release 0.18 yet?

torkleyy:
Done: https://github.com/gfx-rs/gfx/pull/2429

kvark:
@torkleyy there were things I wanted to change in 0.18 rather brutally (i.e. really breaking changes), and I've not got any requests to release 0.18, so I'm procrastinating doing those changes :)

torkleyy:
Oops sorry I didn't see you did bors r+ already. I removed the version bump commit so it got cancelled.

PR #2429, opened 2018-10-01
Backport #2428 to v0.17

torkleyy:


src/render/Cargo.toml
```diff
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx"
-version = "0.17.1"
+version = "0.17.2"
```
kvark: we shouldn't touch this version, let's instead touch the `src/core` version

torkleyy: core version is already bumped

torkleyy: Done.

PR #2430, opened 2018-10-02
[mtl] simple blit with scaling between the swapchains

### LINKED ISSUE #2414: quad example flickers during resize on macos/ios using metal backend
Short info header:
- GFX version: master (14789bb04f192a19587cf40004addf8f0c62d50f)
- OS: iOS/macos
- GPU: Radeon Pro 455 2048 MB, and A11 Bionic (iOS)

This is what @kvark had to say about it
> macos swapchain handing is pretty hacky, but it would be nice to make it work as if it's not hacky:)
> for the quad example, first need to check if we report the OutOfDate swapchain so that the quad example keeps re-creating it
> and then verify that the re-creation doesn't causes flicker

kvark:
Fixes #2414 

Also hang on resizing in general in some applications... like Dota2. Apparently, it's now resizing by re-creating the surface completely (previously, it only re-created the swapchain). In this case, we fail to track if the old swapchain is still presented, and the hack becomes inactive.

This new version of the hack is smarter: it's always active, and if the user provided the old swapchain, we'll even do a nice blit from the old frame to the new one, to address the flicker.
cc @mtak- 

PR checklist:
- [x] tested examples with the following backends: Vulkan, Metal

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -573,6 +573,9 @@ fn main() {
             println!("{:?}", swap_config);
             let extent = swap_config.extent.to_extent();
 
+            let (new_swap_chain, new_backbuffer) =
+                device.create_swapchain(&mut surface, swap_config, Some(swap_chain));
+
             // Clean up the old framebuffers, images and swapchain
             for framebuffer in framebuffers {
                 device.destroy_framebuffer(framebuffer);
@@ -580,10 +583,7 @@ fn main() {
             for (_, rtv) in frame_images {
                 device.destroy_image_view(rtv);
             }
-            device.destroy_swapchain(swap_chain);
 
-            let (new_swap_chain, new_backbuffer) =
-                device.create_swapchain(&mut surface, swap_config, None);
             backbuffer = new_backbuffer;
             swap_chain = new_swap_chain;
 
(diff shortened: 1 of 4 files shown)


examples/quad/main.rs
```diff
@@ -573,17 +573,17 @@ fn main() {
...
+                device.create_swapchain(&mut surface, swap_config, Some(swap_chain));
+
             // Clean up the old framebuffers, images and swapchain
             for framebuffer in framebuffers {
                 device.destroy_framebuffer(framebuffer);
             }
             for (_, rtv) in frame_images {
                 device.destroy_image_view(rtv);
             }
-            device.destroy_swapchain(swap_chain);
```
grovesNL: Don't we still want to destroy it at some point?

mtak-:
This fixes the flicker for me 🎉 . I hoped you wouldn't get to it before me!

I was looking MoltenVK handling of `MTLDrawable`'s, and it grabs them a little more lazily. Does that seem right? I haven't seen similar blit hacks in MoltenVK, so I'm gonna keep poking at it more it in my spare time.

Thank you!

kvark:
It's moved in here, so the implementation destroys it.
Bors r=grovesNL 

As for MVK, they have quite a bit of lazyness, and generally operate in a deferred environment where commands are recorded at submission time... yikes!
For resize, their solution is: just always expose only 2 swapchain images, which makes 3rd one a spare for the resize migration:)

Issue #2431, opened 2018-10-05
Quad example crashes with dx12


zakorgy:
Short info header:
- GFX version: 2e4a9081453883e566471bdaddee457045502b44
- OS: Windows 10
- GPU: Intel(R) HD Graphics 630
Here is the trace:
```
AdapterInfo { name: "Intel(R) HD Graphics 630", vendor: 32902, device: 22802, device_type: DiscreteGpu }
AdapterInfo { name: "Microsoft Basic Render Driver", vendor: 5140, device: 140, device_type: VirtualGpu }
Memory types: [MemoryType { properties: DEVICE_LOCAL, heap_index: 0 }, MemoryType { properties: DEVICE_LOCAL | COHERENT | CPU_VISIBLE | CPU_CACHED, heap_index: 0 }, MemoryType { properties: DEVICE_LOCAL | COHERENT | CPU_VISIBLE | CPU_CACHED, heap_index: 0 }]
formats: Some([Bgra8Srgb, Bgra8Unorm, Rgba8Srgb, Rgba8Unorm, A2b10g10r10Unorm, Rgba16Float])
SwapchainConfig { present_mode: Fifo, format: Bgra8Srgb, extent: Extent2D { width: 1024, height: 768 }, image_count: 2, image_layers: 1, image_usage: COLOR_ATTACHMENT }
resized to LogicalSize { width: 1024.0, height: 768.0 }
SwapchainConfig { present_mode: Fifo, format: Bgra8Srgb, extent: Extent2D { width: 1024, height: 768 }, image_count: 2, image_layers: 1, image_usage: COLOR_ATTACHMENT }
ERROR 2018-10-05T11:11:49Z: gfx_backend_dx12::device: error on swapchain creation 0x80070005
thread 'main' panicked at 'assertion failed: !self.is_null()', C:\Users\zakorgy\.cargo\git\checkouts\d3d12-rs-9104402ad61c26c8\124f9fc\src\com.rs:37:9
stack backtrace:
   0: std::sys::windows::backtrace::set_frames
             at /rustc/8c4ad4e9e42f4bcbdbb784692b936c2462061abc\src/libstd\sys\windows\backtrace\mod.rs:104
   1: std::sys::windows::backtrace::set_frames
             at /rustc/8c4ad4e9e42f4bcbdbb784692b936c2462061abc\src/libstd\sys\windows\backtrace\mod.rs:104
   2: std::sys_common::backtrace::_print
[...]

PR #2433, opened 2018-10-05
dx12: Destroy old swapchain on recreation

msiglreith:
Fixes https://github.com/gfx-rs/gfx/issues/2431
PR checklist:
- [x] tested quad with the following backends: dx12

Currently crashes on shutdown, not totally sure. Might be that the window handle is already invalid but the swapchain is still alive? Need to investigate further..

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -676,6 +676,8 @@ fn main() {
     }
 
     // cleanup!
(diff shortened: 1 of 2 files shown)


msiglreith:
Remaining issue was that some resources associated with the swapchain were still in use.
bors r=kvark

PR #2440, opened 2018-10-15
[v0.17] Fix for panic when creating an empty vertex buffer dx11

### LINKED ISSUE #2370: Error creating empty vertex buffer with DX11 backend
- GFX version: 0.17.1
- OS: Windows 10
- GPU: Geforce GTX 660

See this example program: https://github.com/djcsdy/gfx-dx11-empty-vertex-buffer/blob/master/src/main.rs

`factory.create_vertex_buffer::<Vertex>(&[Vertex { pos: [0.0, 0.0] }]);` is ok, but `factory.create_vertex_buffer::<Vertex>(&[]);` panics with `CreationError::Other`.

I believe this is because [`ID3D11Device::CreateBuffer`](https://docs.microsoft.com/en-us/windows/desktop/api/d3d11/nf-d3d11-id3d11device-createbuffer) fails when attempting to create an empty buffer, but that behaviour is not documented by Microsoft and I haven’t double checked.

This is preventing me from using [conrod](https://github.com/PistonDevelopers/conrod) in conjunction with gfx-rs and DX11, because one of the first things conrod does on startup is to create an empty vertex buffer.

FWIW, the OpenGL backend has a [specific workaround](https://github.com/gfx-rs/gfx/pull/1187/commits/fc29cde45e75c467e3b9c59e0a1eaa4b547b9f7a) for the same problem.

Napokue:
Fixes #2370 for v0.17
PR checklist:
- [x] tested examples with the following backends: dx11

### DIFF
--- a/src/backend/dx11/Cargo.toml
+++ b/src/backend/dx11/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_device_dx11"
-version = "0.7.0"
+version = "0.7.1"
 description = "DirectX-11 backend for gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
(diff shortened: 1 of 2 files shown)


kvark:
Please bump that patch version of the dx11 backend, and we are good to go :)

PR #2445, opened 2018-10-17
lifetime elision

Michael-Lfx:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/metal/src/internal.rs
+++ b/src/backend/metal/src/internal.rs
@@ -200,11 +200,11 @@ impl DepthStencilStates {
     }
 
     // TODO: avoid locking for writes every time
-    pub fn get<'a>(
-        &'a self,
+    pub fn get(
+        &self,
         desc: pso::DepthStencilDesc,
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/internal.rs
```diff
@@ -200,11 +200,11 @@ impl DepthStencilStates {
     }
 
     // TODO: avoid locking for writes every time
-    pub fn get<'a>(
-        &'a self,
+    pub fn get(
+        &self,
         desc: pso::DepthStencilDesc,
         device: &Mutex<metal::Device>,
```
kvark: doesn't the self lifetime now affect this parameter as well?

Michael-Lfx: According to [Clippy.needless_lifetimes](https://rust-lang-nursery.github.io/rust-clippy/master/index.html#needless_lifetimes) we can simplify 'a lifetime marker.

kvark:
great!
bors r+

PR #2461, opened 2018-11-07
Added supports_transfer function.

bbrown683:
Fixes #issue
PR checklist:


src/hal/src/queue/capability.rs
```diff
@@ -50,7 +50,9 @@ impl Capability for Compute {
 impl Capability for Transfer {
     fn supported_by(qt: QueueType) -> bool {
         match qt {
-            _ => true
+            QueueType::General |
+            QueueType::Transfer => true,
```
omni-viral: Compute and Graphics queues supports transfer too.

bbrown683: Somehow slipped my mind.

bbrown683: Fixed

PR #2475, opened 2018-11-18
Fixes to get gl backend to work with opengl 3.0

ZeGentzy:

Fixes #issue
PR checklist:
- [x] tested examples with the following backends: Linux, quad

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -410,6 +410,58 @@ impl Device {
     }
 }
 
+pub(crate) unsafe fn set_sampler_info<SetParamFloat, SetParamFloatVec, SetParamInt>(
+    share: &Starc<Share>,
+    info: &i::SamplerInfo,
+    mut set_param_float: SetParamFloat,
+    mut set_param_float_vec: SetParamFloatVec,
+    mut set_param_int: SetParamInt,
+)
+where
+    SetParamFloat: FnMut(GLenum, GLfloat),
+    SetParamFloatVec: FnMut(GLenum, &[GLfloat]),
+    SetParamInt: FnMut(GLenum, GLint),
+{
+    let (min, mag) = conv::filter_to_gl(info.mag_filter, info.min_filter, info.mip_filter);
+    match info.anisotropic {
+        i::Anisotropic::On(fac) if fac > 1 => {
+            if share.private_caps.sampler_anisotropy_ext {
+                set_param_float(gl::TEXTURE_MAX_ANISOTROPY_EXT, fac as GLfloat);
+            } else if share.features.contains(c::Features::SAMPLER_ANISOTROPY) {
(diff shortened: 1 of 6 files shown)


src/backend/gl/src/device.rs
```diff
@@ -1071,7 +1090,28 @@ impl d::Device<B> for Device {
...
+                        let mut h = h;
+                        for i in 0..num_levels {
+                            gl.TexImage2D(
+                                gl::TEXTURE_2D,
+                                i as _,
+                                int_format as _,
+                                w as _,
+                                h as _,
+                                0,
+                                gl::RGBA,
```
msiglreith: hardcoded values

src/backend/gl/src/queue.rs
```diff
@@ -564,6 +564,20 @@ impl CommandQueue {
                 let gl = &self.share.context;
                 gl.BindSampler(index, sampler);
             }
+            com::Command::SetTextureSamplerSettings(index, texture, ref sinfo) => unsafe {
+                let gl = &self.share.context;
+                gl.ActiveTexture(gl::TEXTURE0 + index);
+                gl.BindTexture(gl::TEXTURE_2D, texture);
```
msiglreith: we should at least warn/panic here is the texture is not 2D

ZeGentzy: Currently the OGL backend does not keep track of dimensionality, as `create_image` currently only supports making 2D images, therefore, all textures will be 2D. All the other code in the backend also assumes textures will only be 2D. Solving this problem is beyond the scope of this PR.

src/backend/gl/src/info.rs
```diff
@@ -304,6 +305,7 @@ impl Info {
 pub(crate) fn query_all(gl: &GlContainer) -> (Info, Features, LegacyFeatures, Limits, PrivateCaps) {
     use self::Requirement::*;
     let info = Info::get(gl);
+    println!("{:?}", info);
```
kvark: debug leftover?

src/backend/gl/src/device.rs
```diff
@@ -410,6 +410,58 @@ impl Device {
...
+pub(crate) unsafe fn set_sampler_info<SetParamFloat, SetParamFloatVec, SetParamInt>(
+    share: &Starc<Share>,
+    info: &i::SamplerInfo,
+    mut set_param_float: SetParamFloat,
+    mut set_param_float_vec: SetParamFloatVec,
+    mut set_param_int: SetParamInt,
+)
+where
+    SetParamFloat: FnMut(GLenum, GLfloat),
+    SetParamFloatVec: FnMut(GLenum, &GLfloat),
```
kvark: this should be accepting `&[GLfloat]`

src/backend/gl/src/info.rs
ZeGentzy: Yup, removed.

PR #2484, opened 2018-11-27
Add a default size parameter to SwapchainConfig::from_caps

### LINKED ISSUE #2421: Quad example thinks window is 1x1 pixel when using Vulkan backend under Wayland
Short info header:
- GFX version: Commit `a449dc5`
- OS: Fedora 28 (using Wayland)
- GPU: Intel UHD 620

When using the Vulkan backend, the drawn area is only 1x1 pixel in size when running the quad example. This issue does not occur when using the GL backend. The window is the 'correct' size, I can tell because the titlebar is the right width and the resize grab handles appear in the right places. However the window is almost totally transparent apart from a single pixel in the upper-left corner.

Whenever I resize the window the following message is printed in the console output:

    SwapchainConfig { present_mode: Fifo, format: Bgra8Srgb, extent: Extent2D { width: 1, height: 1 }, image_count: 4, image_layers: 1, image_usage: COLOR_ATTACHMENT }

keringar:
Fixes #2421 by adding a default parameter to SwapchainConfig::from_caps. In Vulkan on Wayland, the surface capabilities don't give us a current extent because `vkGetPhysicalDeviceSurfaceCapabilitiesKHR` sets current_extent to `0xFFFFFFFF`. Which means that the surface size is determined by the swapchain, not the surface capabilities.

PR checklist:
    - Vulkan (on Wayland and X11)
    - OpenGL (on Wayland and X11)
    - A very noisy diff so I left it as a separate commit

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -1,5 +1,10 @@
 #![cfg_attr(
-    not(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl")),
+    not(any(
+        feature = "vulkan",
+        feature = "dx12",
+        feature = "metal",
+        feature = "gl"
+    )),
     allow(dead_code, unused_extern_crates, unused_imports)
 )]
 
@@ -63,7 +68,12 @@ const COLOR_RANGE: i::SubresourceRange = i::SubresourceRange {
     layers: 0..1,
 };
 
-#[cfg(any(feature = "vulkan", feature = "dx12", feature = "metal", feature = "gl"))]
+#[cfg(any(
+    feature = "vulkan",
+    feature = "dx12",
+    feature = "metal",
+    feature = "gl"
+))]
 fn main() {
     env_logger::init();
 
@@ -73,8 +83,7 @@ fn main() {
         .with_dimensions(winit::dpi::LogicalSize::new(
             DIMS.width as _,
             DIMS.height as _,
-        ))
-        .with_title("quad".to_string());
+        )).with_title("quad".to_string());
     // instantiate backend
     #[cfg(not(feature = "gl"))]
     let (_window, _instance, mut adapters, mut surface) = {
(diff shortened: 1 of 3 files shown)


examples/quad/main.rs
```diff
@@ -359,9 +360,18 @@ fn main() {
             .unwrap_or(formats[0])
     });
 
-    let swap_config = SwapchainConfig::from_caps(&caps, format);
+    let (min_width, max_width) = (caps.extents.start.width, caps.extents.end.width - 1);
+    let (min_height, max_height) = (caps.extents.start.height, caps.extents.end.height - 1);
+
+    let mut swap_config = SwapchainConfig::from_caps(&caps, format);
```
kvark: maybe we need to add a parameter to `from_caps` specifying the default sizes, so that the branch can be moved into the implementation?

keringar: I added a `default_extent` param to `SwapchainConfig::from_caps`. If the passed in parameter doesn't fit within the range of the surface sizes, than it's clamped to fit automatically. I also added a debug assertion if it was clamped automatically and the size was changed to fit.

src/hal/src/window.rs
```diff
@@ -282,12 +283,31 @@ impl SwapchainConfig {
...
+            None => {
+                let (min_width, max_width) = (caps.extents.start.width, caps.extents.end.width - 1);
+                let (min_height, max_height) = (caps.extents.start.height, caps.extents.end.height - 1);
+
+                // clamp the default_extent to within the allowed surface sizes
+                let width = min(max_width, max(default_extent.width, min_width));
+                let height = min(max_height, max(default_extent.height, min_height));
+
+                // assert that the default_extent fits within the allowed sizes
+                debug_assert_eq!(default_extent.height, height);
```
kvark: I don't think these assertions are needed: the parameter isn't supposed to be strict - merely a hint on what configuration to build

keringar: I've removed the assertions

PR #2485, opened 2018-11-27
Add support for attribute binding rate on the GL backend

siriux:
This allows instancing on the GL backend.


kvark:
@siriux thank you for the PR! I'd be fine with merging this, however we'll need proper checks for instancing to be supported in place before using it.
@ZeGentzy want to have a look?

siriux:
Yes, you are right, but for that I need to modify the LegacyFeatures struct that is already "full". Is it ok if I upgrade it to u32? Or are you thiking about a different way of checking.

kvark:
@siriux yeah, I think legacy features can be extended no problem

ZeGentzy:
A quick skim makes me think it looks good (just needs a legacy features check), haven't actually tested however. If @siriux has an small example which uses instancing, that would be nice.

PR #2488, opened 2018-11-30
[vk] extend dual-source blending disable on Intel/windows

### LINKED ISSUE #1930: [vk] crash in dual-source blending on Windows + Intel

zakorgy:
This is a follow up on #1931 and fixes #1930.
cc @kvark 

PR checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/2488)

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -552,10 +552,12 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
 
     fn features(&self) -> Features {
         // see https://github.com/gfx-rs/gfx/issues/1930
-        let is_windows_intel_kaby = cfg!(windows)
+        let is_windows_intel_dual_src_bug = cfg!(windows)
             && self.properties.vendor_id == info::intel::VENDOR
-            && self.properties.device_id & info::intel::DEVICE_KABY_LAKE_MASK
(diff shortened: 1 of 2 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -557,6 +557,11 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
             && self.properties.device_id & info::intel::DEVICE_KABY_LAKE_MASK
                 == info::intel::DEVICE_KABY_LAKE_MASK;
 
+        let is_windows_intel_sky = cfg!(windows)
```
kvark: let's have a single boolean for this workaround, e.g. `is_windows_intel_dual_src_bug`

Issue #2479, opened 2018-11-23
How to use feature auto-capture


Michael-Lfx:
I find it that `auto-capture` is using `MTLCaptureManager` which supports Capture GPU Frame of Xcode. How to use it when running Dota2 due to no launching from Xcode?

kvark:
@Michael-Lfx you can only use it *when* running from XCode. Also, it doesn't make sense for Dota - frame boundaries trigger capture automatically. It's useful for workloads that are either compute or don't present frame, mostly Vulkan CTS. Here are instructions on how to set up XCode to run anything external, I have all the things set up this way (Dota, Vulkan CTS, Filament, etc):

![xcode-capture](https://user-images.githubusercontent.com/107301/48951515-d0692d80-ef0c-11e8-8b57-16a01c0b2274.png)
ode.

Michael-Lfx:
I think it's better to make it as a wiki page for new latecomers.

kvark:
I think it'd be good to have this sort of documentation to be a part of the repo, e.g. by existing in the backend Readme, or even as comments in Cargo.toml.

Michael-Lfx:
Another proposal: leaving a fine-configured project/workspace in examples to save time.

Michael-Lfx:
Unfortunately I failed to make Capture GPU Frame work because no matter quad or colour-uniform they crashed ASAP when Build & Run is pressed. 
![2281543485351_ pic_hd](https://user-images.githubusercontent.com/9588926/49214024-aa59f680-f400-11e8-8369-e4b3186a5414.jpg)
<img width="530" alt="image" src="https://user-images.githubusercontent.com/9588926/49214061-c2ca1100-f400-11e8-9f5a-0868d19acc77.png">
<img width="1045" alt="image" src="https://user-images.githubusercontent.com/9588926/49214783-52bc8a80-f402-11e8-99e8-f5eef1e18462.png">
<img width="896" alt="image" src="https://user-images.githubusercontent.com/9588926/49214799-5c45f280-f402-11e8-9300-24c8093bb1d5.png">
<img width="896" alt="image" src="https://user-images.githubusercontent.com/9588926/49214833-6bc53b80-f402-11e8-846a-8cb40ff669ea.png">

With or without "Main Thread Checker" quad crashed anyway. My codebase is `6fa48b1209ee19b98abb10c0feecff9acbf44b29` which log is Merge #2484.

[...]

PR #2489, opened 2018-11-30
Add concat env for shader file reading, make it more convenient for …

Michael-Lfx:
Xcode Capture GPU Frame as #2479 discussed

Fixes #issue
PR checklist:
  - [x] Metal

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -1292,7 +1292,7 @@ impl<B: Backend> PipelineState<B> {
 
         let pipeline = {
             let vs_module = {
-                let glsl = fs::read_to_string("colour-uniform/data/quad.vert").unwrap();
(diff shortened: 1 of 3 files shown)


kvark:
I mean, surely the examples aren't meant to be "movable" but then new applications tend to copy code from them, so it's good to show best practices in general, not something unique to examples.

Michael-Lfx:
Adding absolute path for shader file results in no requirement for setting the working directory of Xcode Run action, which is easily to make mistakes to beginners. Or replace `fs::read_to_string()` with `include_bytes!()` then debugging shaders directly with Reload Shader of Xcode?

kvark:
Setting up a working dir is a one time operation that you'd do for any other non-trivial project anyway, so why try to avoid it?

Michael-Lfx:
Everyone will not have to repeat that operation. And I think it's better to provide a Xcode project within `examples`.

kvark:
if we do provide an XCode project for the examples, can't we set the working dir in it to be relative to something (e.g. project dir)?

Michael-Lfx:
Sure.

PR #2493, opened 2018-12-02
Ignore when using concrete ShaderModule

Michael-Lfx:
Both vs_module and fs_module are concrete ShaderModules.

Fixes #issue
PR checklist:
   - [x] Metal


kvark:
The reason for that was that Rustc refused to infer the types properly. If it's overcome it now, great!
bors r+

PR #2495, opened 2018-12-05
PRESERVE shortcut for attachment ops

kvark:


kvark:
You can approve it for realz :)
Bors r=grovesNL

PR #2499, opened 2018-12-06
Add queue family resource ownership transfer to memory barriers for Vulkan

termhn:
This is untested so far because I am not currently at a computer with Vulkan to test on... seems simple enough though.

cc @omni-viral 

PR checklist:


termhn:
@msiglreith It appears that `pipeline_barrier` doesn't do anything on `metal` backend currently anyways, and I'm not familiar enough with `dx12` to know if there's a similar concept of ownership passing between queue families to take care of there, but at the cursory glance I took of the code it didn't seem like it... of course you would definitely know better than I :P

msiglreith:
Relevant docs for resource ownership transfer: https://docs.microsoft.com/en-us/windows/desktop/direct3d12/user-mode-heap-synchronization#multi-queue-resource-access
But this would require move in depth analysis and it corresponds with the current resource state logic.
I guess ignoring it for now is fine, maybe just add a TODO

omni-viral:
You could explicitly panic if not equal queue index are specified in dx12

termhn:
Done both. Not sure if it's actually appropriate to panic in the case that they are not equal for dx12, though. Thoughts, @msiglreith @kvark ?

omni-viral:
`unimplemented!()` is appropriate for unimplemented cases

termhn:
Used unimplemented and fixed it so it builds and works, also updated examples

msiglreith:
> Used unimplemented and fixed it so it builds and works, also updated examples
thanks!
warden fails to build currently

msiglreith:
followup issue for dx12 https://github.com/gfx-rs/gfx/issues/2503

termhn:
Done :)

PR #2504, opened 2018-12-10
Removed "max_level_*" features from log dependencies

olivia-fl:
Fixes discussion from #2237.

PR checklist:
- [x] tested examples with the following backends: gl, vulkan


kvark:
Thank you for the PR! Could you clarify one thing: is there a case where macro wrappers are really needed as opposed to just having the target application driving the log level?

Benjamin-L:
No reason in particular, other than that it's a rather unexpected performance pitfall. I feel like the way I implemented the wrapper is fairly ugly either way, so possibly just remove the `max_level_*` features and document somewhere that this is a possible performance problem.

kvark:
Let's do it in this PR, please

PR #2505, opened 2018-12-10
subbuffer-barrier

omni-viral:
Fixes #issue
PR checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/2505)

### DIFF
--- a/examples/compute/main.rs
+++ b/examples/compute/main.rs
@@ -147,7 +147,8 @@ fn main() {
             Some(memory::Barrier::Buffer {
                 states: buffer::Access::TRANSFER_WRITE .. buffer::Access::SHADER_READ | buffer::Access::SHADER_WRITE,
                 families: None,
-                target: &device_buffer
+                target: &device_buffer,
+                range: None .. None,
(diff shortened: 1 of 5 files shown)


msiglreith:
not sure about this one, unable to support it at least on dx12.

omni-viral:
@msiglreith if I understand something in barriers you always can specify more than you need.

msiglreith:
@omni-viral I believe it legal to use a subrange of a buffer as storage buffer for example and other subrange as index buffer without interference. This would most likely cause issues with dx12 as inferes flushes and invalidations via resource state transitions, which apply to the whole resource Vulkan is a bit closer to the hardware which makes it difficult to map such more fine granular use cases to other backends ):
In practice, it might work okish or we would need to fallback to whole pipeline flushes in this case, not totally sure.

kvark:
I agree this may be problematic on D3D12, but it's already in Vulkan so with this API change we'll get less info lost in portability-gfx translation, so it's good to merge.

PR #2513, opened 2018-12-14
[WIP] Auto-allocated resources

### LINKED ISSUE #2511: Provide "committed resource" semantics in HAL
Primary motivation is getting advantage of [vk-mem](https://crates.io/crates/vk-mem) by @gwihlidal in our Vulkan backend. The new API entry points for buffers and images could return BoundXxx right away.

Implementing in D3D11/D3D12 and Metal is also going to be straightforward.

kvark:
Fixes #2511
PR checklist:


kvark:
Ugh, this is right, but it would require more apis...

Issue #2500, opened 2018-12-06
Switch the API away from Borrow bounds


kvark:
It is a very common pattern in gfx-hal API right now to have `I: IntoIterator` and `I::Item: Borrow<Something>`. Unfortunately, this falls apart if the user actually wants to have something like this:
```rust
struct Foo<B: Backend>(B::Something);
impl<B: Backend> Borrow<B::Something> for Foo<B> {...}
```
Compiler isn't happy, because `T` impls `Borrow<T>` and there is no way to convince the compiler that this implementation doesn't conflict with our new one...

I'm going to change everything to use `AsRef` therefore. Any thoughts? Objections?

omni-viral:
There is a way to convince the compiler

```
struct Foo<B, S=B::Something>(S);
impl<B> Borrow<B::Something> for Foo<B> { ... }
```

kvark:
Apparently, `AsRef` isn't a solution either, since it's not implemented for `T` or `&T`. Next alternative: `Iterator<&'a T>`

omni-viral:
`Iterator<Item = &'a T>` isn't better than `Iterator<Item = impl Borrow<T>>`

kvark:
@omni-viral it *is* better because `Borrow<T>` just isn't possible for structs that just depend on `<B: Backend>`, including our own `CommandBuffer`. The trade-off here is that for `Borrow` we'd require all the users to change their data types/structures, while for `&'a T` it's only a single `iter()` call (or similar) at the use site.

kvark:
Interestingly, `Iterator<Item = &'a T>` isn't a solution either :( If the user has an iterator over actual items, I don't see how Rust would allow us to make an iterator over their references instead (sounds like a case for streaming iterators, which we don't have). See [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d0f9d491bb9483041d1d6fd4710ad6ac).

TL;DR: no solution is good atm?
  - `Borrow` is bad because we can't `impl<B: Backend> Borrow<B::AssociatedType> for Something<B>`, where `Something<B: Backend>(B::AssociatedType)` or similar
    - a workaround exist by adding an extra generic type, but it's a bit awkward
  - `AsRef` is bad because it's not implemented for `T` or `&T`
  - `&'a T` is bad because we can't get an iterator over references if we have one over owned items...

@Gankro - maybe you can make a recommendation for us?

Gankro:
[...]

PR #2514, opened 2018-12-14
Remove CommandBuffer: Clone bound

### LINKED ISSUE #2247: Remove RawCommandBuffer: Clone bound
This is one of the worst sounding API aspects of HAL at the moment.
It's currently defined as:
```rust
pub trait RawCommandBuffer<B: Backend>: Clone + Any + Send + Sync
```
This was likely needed for secondary command buffers and submissions.
Steps to fix:
  1. remove the `Clone` bound
  2. fix build errors
  3. update portability
  4. profit!

kvark:
Fixes #2247

PR checklist:

The `CommandBuffer` type and borrow implementation are extremely confusing to me :(


src/hal/src/command/mod.rs
```diff
@@ -65,100 +67,74 @@ impl Level for Primary { }
...
+    pub fn as_raw_mut(&mut self) -> &mut B::CommandBuffer {
+        &mut self.raw
+    }*/
+
     /// Downgrade a command buffer to a lesser capability type.
     ///
     /// This is safe as a downgraded version can't be `submit`'ed
     /// since `submit` requires `self` by move.
-    pub fn downgrade<D>(&mut self) -> &mut CommandBuffer<'a, B, D, S>
+    pub fn downgrade<D>(&mut self) -> &mut CommandBuffer<B, D, S>
```
omni-viral: While you're at it. This method should be marked unsafe.

src/hal/src/pool.rs
```diff
@@ -107,15 +87,12 @@ impl<B: Backend, C> CommandPool<B, C> {
     pub fn acquire_command_buffer<S: Shot>(
         &mut self, allow_pending_resubmit: bool
     ) -> CommandBuffer<B, C, S> {
-        self.reserve(1);
-
-        let buffer = &mut self.buffers[self.next_buffer];
+        let mut buffer = self.raw.allocate_one(RawLevel::Secondary);
```
omni-viral: Primary

src/hal/src/pool.rs
```diff
@@ -107,15 +87,12 @@ impl<B: Backend, C> CommandPool<B, C> {
...
         let mut flags = S::FLAGS;
         if allow_pending_resubmit {
             flags |= CommandBufferFlags::SIMULTANEOUS_USE;
         }
         buffer.begin(flags, CommandBufferInheritanceInfo::default());
```
omni-viral: Now user should always begin recording manually.
Because after reset user would have to call `begin` anyway.

src/hal/src/command/render_pass.rs
```diff
@@ -280,80 +286,68 @@ impl<'a, B: Backend> RenderPassSecondaryEncoder<'a, B> {
...
 
         RenderPassSecondaryEncoder(
-            Some(cmd_buffer.raw),
+            &mut cmd_buffer.raw,
         )
     }
 
     /// Executes the given commands as a secondary command buffer.
-    pub fn execute_commands<I>(&mut self, submits: I)
+    pub fn execute_commands<I, P: Shot>(&mut self, submits: I)
```
termhn: change argument name here as well

src/hal/src/command/mod.rs
kvark: Why should this method be unsafe? It explicitly says why it's safe in the doc comment.

src/hal/src/pool.rs
kvark: oops :)

kvark:
I'm hitting some nasty compile errors with that extra default generic parameter on `CommandBuffer` :(

src/hal/src/command/mod.rs
omni-viral: `std::mem::replace`

src/hal/src/queue/mod.rs
```diff
@@ -105,14 +105,15 @@ impl<B: Backend, C> CommandQueue<B, C> {
 
     /// Submits the submission command buffers to the queue for execution.
     /// `fence` will be signalled after submission and _must_ be unsignalled.
-    pub fn submit<D>(&mut self,
-        submission: Submission<B, D>,
+    pub fn submit<IC>(&mut self,
```
omni-viral: Now user can submit any command buffer into any queue.

src/hal/src/command/mod.rs
```diff
@@ -65,100 +67,103 @@ impl Level for Primary { }
...
 
-impl<'a, B: Backend, C, S: Shot> CommandBuffer<'a, B, C, S, Primary> {
+impl<B: Backend, C, S: Shot> CommandBuffer<B, C, S, Primary> {
     /// Identical to the `RawCommandBuffer` method of the same name.
-    pub fn execute_commands<I, K>(&mut self, submits: I)
+    pub fn execute_commands<I, K, P: Shot>(&mut self, cmd_buffers: I)
     where
-        I: IntoIterator,
-        I::Item: Submittable<'a, B, K, Secondary>,
+        I: IntoIterator<Item = CommandBuffer<B, K, P, Secondary>>,
```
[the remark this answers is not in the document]
kvark: what do you mean?

src/hal/src/queue/mod.rs
kvark: yes, compromises

src/hal/src/command/mod.rs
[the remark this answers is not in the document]
omni-viral: This function takes non-copiable command buffers by value. Who will destroy them? Not to mention I want to reuse them.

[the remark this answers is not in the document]
kvark: right, we shouldn't be consuming them here

kvark:
ok, only a few backends left to fix...

kvark:
@omni-viral PTAL!

src/hal/src/command/render_pass.rs
```diff
@@ -109,70 +120,70 @@ impl<'a, B: Backend> RenderSubpassCommon<'a, B> {
...
     ///
     pub fn write_timestamp(&mut self, stage: pso::PipelineStage, query: query::Query<B>) {
-        self.0.write_timestamp(stage, query)
+        self.cmb.borrow_mut().write_timestamp(stage, query)
     }
 }
 
 /// An object that records commands into a command buffer inline, that is,
 /// without secondary command buffers.
-pub struct RenderPassInlineEncoder<'a, B: Backend, L: Level>(pub(crate) Option<RenderSubpassCommon<'a, B>>, PhantomData<L>)
```
omni-viral: `RenderPassInlineEncoder` cannot be on secondary level.

src/hal/src/queue/submission.rs
```diff
@@ -2,119 +2,20 @@
...
-pub struct RawSubmission<'a, B: Backend + 'a, IC>
+use std::borrow::Borrow;
+
+
+/// Submission information for a command queue.
+pub struct Submission<'a, B: Backend + 'a, T, IC>
 where
-    IC: IntoIterator,
-    IC::Item: Borrow<B::CommandBuffer>,
+    T: 'a + Borrow<B::CommandBuffer>,
```
omni-viral: Don't put bounds on types.

src/hal/src/command/render_pass.rs
kvark: nicely spotted!

kvark:
I think all the pieces come together, at last:
  - we don't have the obscure `Borrow` API for submissions specifically which is inconsistent with the rest of borrows in the API
  - we have a `RawSubmittable` + `Submittable` traits
  - we still enforce the `Capability` to an extent, or at least don't silently ignore it

There is still a question on what to do with the `Borrow` things in the code, but now it's somewhat an orthogonal issue here - #2500.

@omni-viral PTAL again!

src/hal/src/command/mod.rs
```diff
@@ -65,100 +64,131 @@ impl Level for Primary { }
...
 impl Level for Secondary { }
 
-/// Thread-safe finished command buffer for submission.
-pub struct Submit<B: Backend, C, S, L>(pub(crate) B::CommandBuffer, pub(crate) PhantomData<(C, S, L)>);
-impl<B: Backend, C, S, L> Submit<B, C, S, L> {
-    fn new(buffer: B::CommandBuffer) -> Self {
-        Submit(buffer, PhantomData)
-    }
+/// A property of a command buffer to be submitted to a queue.
+pub trait RawSubmittable<B: Backend> {
```
omni-viral: I think this should be implemented for `B::CommandBuffer` at least.
Otherwise user woudln't be able to submit raw `&B::CommandBuffer` into raw `B::CommandQueue` which seems odd :)

Maybe `Borrow<B::CommandBuffer>` would make more sense than `RawSubmittable`

src/hal/src/command/mod.rs
```diff
@@ -65,100 +64,131 @@ impl Level for Primary { }
...
+    /// Borrow a raw command buffer from this submittable.
+    fn as_command_buffer(&self) -> &B::CommandBuffer;
 }
-unsafe impl<B: Backend, C, S, L> Send for Submit<B, C, S, L> {}
 
+/// A property of a command buffer to be submitted to a queue with specific capability.
+pub trait Submittable<B: Backend, K: Capability, L: Level>: RawSubmittable<B> {}
```
omni-viral: In all other places `C` is used for `Capability`. Except implementation of this trait, where two capability type parameters are defined and using `K` perfectly justified.

omni-viral: I believe we should avoid unused bounds.

src/hal/src/command/mod.rs
```diff
@@ -65,100 +64,131 @@ impl Level for Primary { }
...
     {
         unsafe { ::std::mem::transmute(self) }
     }
 }
 
-impl<'a, B: Backend, C, S: Shot> CommandBuffer<'a, B, C, S, Primary> {
+impl<B: Backend, C, S: Shot> CommandBuffer<B, C, S, Primary> {
     /// Identical to the `RawCommandBuffer` method of the same name.
-    pub fn execute_commands<I, K>(&mut self, submits: I)
+    pub fn execute_commands<'a, I, T, K, P: Shot>(&mut self, cmd_buffers: I)
```
omni-viral: Is `P` used?

src/hal/src/command/mod.rs
kvark: I think these bounds are fine on a trait (as opposed to a struct). Consider them a part of documentation ;)

src/hal/src/command/mod.rs
kvark: nope. I wonder why Rust doesn't complain about this

src/hal/src/command/mod.rs
omni-viral: It is the same problem as with struct - impl block would require those bounds even though they aren't used anywhere.

src/hal/src/command/mod.rs
omni-viral: `#![deny(unused)` will make Rust complain.

src/hal/src/command/mod.rs
kvark: There is a slight distinction though. A struct defines data, a trait defines logic. The constraints are more relevant to logic. Also, you typically want to implement multiple traits for a struct, hence it quickly becomes annoying. In our case, you are likely to only have at most one type to implement the trait on, so it's not nearly as annoying.

src/hal/src/command/mod.rs
kvark: Nope. Rust warns by default, and I don't see any warnings either. This must be a bug in `rustc` then.
Anyhow, I added the deny now, it makes sense to have it in gfx-hal crate at least.

omni-viral: Indeed. `unused` doesn't cover unused type parameters in functions.

src/hal/src/queue/mod.rs
```diff
@@ -42,6 +42,16 @@ pub enum QueueType {
     Transfer,
 }
 
+/// Submission information for a command queue.
+pub struct Submission<'a, B: Backend + 'a, I> {
+    /// Command buffers to submit.
+    pub command_buffers: I,
+    /// Semaphores to wait being signalled before submission.
+    pub wait_semaphores: &'a [(&'a B::Semaphore, pso::PipelineStage)],
```
omni-viral: Why this is not an iterator type? This seems inconsistent with rest of API.

`WI: IntoIterator<Item = (&'a W, pso::PipelineStage)>, W: Borrow<B::Semaphore>` on methods should do :)

kvark: holy cow, that would make `submit` to carry 7 generic parameters :/ How about doing this as a follow up? I don't think my PR regresses this anyhow.

omni-viral: Lucky number :)

omni-viral: Old `Submission` uses `IntoIterator` for semaphores. So this PR regress it.
But it was done suboptimal `Borrow<(&B::Semaphore, PipelineStage)>` is not very convenient.

omni-viral:
If there are no objections.
bors r+

omni-viral:
:(

kvark:
Bors r=omni-viral

PR #2515, opened 2018-12-18
[0.17/gl] don't use temporary manager for resource creation

kvark:
Fixes a heavy memory leak per frame when using host-mapped resources.
What was happening is the fence being created on a temporary manger, which never cleans up the resources properly (just clears instead, since it assumes it's never the last one).
Ideally, we'd want to design the internal handle manager in such a way that it's not possible at the API level, but this isn't a priority right now.

### DIFF
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name = "gfx_device_gl"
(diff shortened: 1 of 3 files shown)


kvark:
Alright, proceeding with this. Please feel free to review while it's in flight.
bors r+

PR #2518, opened 2018-12-18
Remove Unbound resource variants

### LINKED ISSUE #2517: Remove Unbound resources
Remove `UnboundBuffer` and `UnboundImage` as they are leftover from the original API and don't fit well into the sparse resource concept (resources don't need to be bound).
The proposal would be to adjust the HAL api as following (similar for images):

### Original
```rust
fn create_buffer(
        &self,
        size: u64,
        usage: buffer::Usage,
) -> Result<B::UnboundBuffer, buffer::CreationError>;

fn bind_buffer_memory(
        &self,
        memory: &B::Memory,
        offset: u64,
        buf: UnboundBuffer,
) -> Result<B::Buffer, BindError>;
```

### Proposed
```rust
fn create_buffer(
        &self,
        size: u64,
        usage: buffer::Usage,
) -> Result<B::Buffer, buffer::CreationError>;

fn bind_buffer_memory(
        &self,
        memory: &B::Memory,
        offset: u64,
        buf: &mut B::Buffer,
) -> Result<(), BindError>;
```

msiglreith:
Remove unbound resources from backend and adjust creation and binding api
Fixes #2517 

#### Backends
- [x] Vulkan
- [x] empty
- [x] GL
- [x] Metal
- [x] dx11
- [x] dx12

#### PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -812,7 +812,7 @@ impl<B: Backend> BufferState<B> {
         T: Copy,
     {
         let memory: B::Memory;
-        let buffer: B::Buffer;
+        let mut buffer: B::Buffer;
         let size: u64;
 
         let stride = size_of::<T>() as u64;
@@ -821,8 +821,8 @@ impl<B: Backend> BufferState<B> {
         {
             let device = &device_ptr.borrow().device;
 
-            let unbound = device.create_buffer(upload_size, usage).unwrap();
-            let mem_req = device.get_buffer_requirements(&unbound);
+            buffer = device.create_buffer(upload_size, usage).unwrap();
+            let mem_req = device.get_buffer_requirements(&buffer);
 
             // A note about performance: Using CPU_VISIBLE memory is convenient because it can be
             // directly memory mapped and easily updated by the CPU, but it is very slow and so should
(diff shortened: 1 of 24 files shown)


Michael-Lfx:
I am confused about these Unbound*  at the beginning. I think it's good to keep consistent with Vulkan.

kvark:
Note that this will technically result in more overhead for direct clients of gfx-hal: on each access of buffer/texture in, say, Metal, we'd now need to check if it's bound or not.

omni-viral:
@kvark instead of checking resource is bound metal backend can assume it is bound.

kvark:
The resource doesn't exist in metal before we bind it.

omni-viral:
@kvark I can't see why this is an issue with my suggestion

kvark:
Well, assumption in this case is just ignored checking, which is quite dangerous: unsafety within the driver/hw is not the same as unsafety in our rust code.

The old api fit better because you couldn't screw if up, or the compiler shouted at you. This means no checks, enums, or unsafety on our side at least.

kvark:
Interestingly, when updating metal I had to add an accessor to `native::Buffer`, which I was meant to do anyway for a while, given that in the code we often used `raw` buffer without bothering to check the ranges... And with this change I discovered the missing erroneous spots: index buffer binding and indirect draw. So this change actually fixes a few Metal bugs ;)

kvark:
I took the liberty of rebasing the code, since it conflicted with the changes I landed earlier.
:rocket:  bors r+

PR #2519, opened 2018-12-18
export `gfx_backend_dx11::device::Device` as a public type.

hinaria:
in the dx11 backend, this type is reachable (it's returned by `open_with<_, gfx_hal::Graphics>`) but not nameable, which means we can't store it in a struct.

`Device` *is* publicly exported in the vulkan backend though, so this pr just unifies the dx11 api to be the same as the the vulkan one.

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -89,6 +89,8 @@ mod internal;
 mod range_alloc;
 mod shader;
 
+pub use device::Device;
+
 #[derive(Clone, Derivative)]
 #[derivative(Debug)]
 pub(crate) struct ViewInfo {


kvark:
The thing is, concrete types of the backends are implementation details. Once you have `Instance` and something that implements `Backend`, you should only operate on the associated types of the `Backend`. Thus, the type you are looking for is `<gfx_backend_vulkan::Backend as gfx_hal::Backend>::Device`. I'd rather hide the concrete type of the vulkan backend for consistency than expose it in other backends.

hinaria:
oh okay, that works. i can submit a pr that removes that, but that's a big type name to have in your struct declarations.

what's the benefit of this since every backend currently implements one `gfx_hal::Backend`, and as i understand it, `open_with` in its current form will always return only one concrete implementation of said backend?

kvark:
That concrete implementation can be named arbitrarily and be in a submodule, etc
Typically an application would have structures generic over B: Backend, so the long notation of the type is never used.

PR #2523, opened 2018-12-21
Actually stop deriving Clone for backend's command buffers

kvark:
Follow-up to #2514 that actually tries to lift the `Clone` bounds in the backends. This is going to be a long CI fight, I feel, but it's good that there is a ton of places where it was used.

PR checklist:


kvark:
This is straightforward
bors r+

PR #2526, opened 2018-12-24
Add few unsafe markers

### LINKED ISSUE #2453: Mark most functions as unsafe
Because we follow Vulkan closely for the HAL API, we don't have many safety guarantees for most of the API usage. A few users have asked to have the explicit unsafe annotation to make it more obvious that no guarantees are provided here.

omni-viral:
Fixes #2453
PR checklist:
  metal

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -3,6 +3,7 @@ name = "hal-examples"
 version = "0.1.0"
 publish = false
 workspace = ".."
+edition = "2018"
 
 [features]
 default = []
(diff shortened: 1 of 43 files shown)


examples/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "hal-examples"
 version = "0.1.0"
 publish = false
 workspace = ".."
+edition = "2018"
```
kvark: let's leave the edition update to other PRs, the scope is already high, unless you require it for this to work

omni-viral: The code needs nll.

kvark:
Can you remove the second commit please?

kvark:
The issue https://github.com/rust-lang/rust/issues/57111 is resolved now, please remove the last commit.

omni-viral:
If you insist

kvark:
Thank you!
Bors r+

PR #2534, opened 2018-12-28
Fix warden memory

kvark:
just something I screwed up when porting the `Unbound` changes earlier


kvark:
merging as trivial
bors r+

PR #2537, opened 2018-12-29
Updated Ash dependency from 0.24 to 0.26

### LINKED ISSUE #2510: Update Ash dependency from 0.24 to 0.26
We are lagging quite behind. The new versions bring more API exposure, also nicer bitflags.

keringar:
Fixes #2510. A lot of the changes were just renaming various constants that were moved around with no major changes. I went through the diff and checked to make sure everything was updated correctly, but it's possible there could have been a typo somewhere in the 500 or so changes. 

I tested the changes with warden and the examples and nothing appears to be wrong, but I can't be sure since it looks like warden is broken with 7 failing tests on the Vulkan backend. But this appears to be a separate issue from updating Ash since warden is also broken on at least the 0.1 release and maybe earlier.

PR checklist:
    - vulkan

### DIFF
--- a/src/backend/vulkan/src/conv.rs
+++ b/src/backend/vulkan/src/conv.rs
@@ -1,15 +1,14 @@
 use ash::vk;
 
-use hal::{buffer, command, format, image, pass, pso, query};
-use hal::{IndexType, Primitive, PresentMode};
 use hal::range::RangeArg;
+use hal::{buffer, command, format, image, pass, pso, query};
+use hal::{IndexType, PresentMode, Primitive};
 
 use native as n;
 
-use std::mem;
 use std::borrow::Borrow;
+use std::mem;
 use std::ptr;
-
 
 pub fn map_format(format: format::Format) -> vk::Format {
     // Safe due to equivalence of HAL format values and Vulkan format values
@@ -17,9 +16,7 @@ pub fn map_format(format: format::Format) -> vk::Format {
 }
 
 pub fn map_vk_format(format: vk::Format) -> Option<format::Format> {
-    if (format as usize) < format::NUM_FORMATS &&
-        format != vk::Format::Undefined
-    {
+    if (format.as_raw() as usize) < format::NUM_FORMATS && format != vk::Format::UNDEFINED {
         // Safe due to equivalence of HAL format values and Vulkan format values
         Some(unsafe { mem::transmute(format) })
     } else {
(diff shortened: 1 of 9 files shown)


kvark:
The failures are fixed by #2534 , fwiw

src/backend/vulkan/src/conv.rs
```diff
@@ -208,86 +208,86 @@ pub fn map_mip_filter(filter: image::Filter) -> vk::SamplerMipmapMode {
...
-        Primitive::TriangleList           => vk::PrimitiveTopology::TriangleList,
-        Primitive::TriangleListAdjacency  => vk::PrimitiveTopology::TriangleListWithAdjacency,
-        Primitive::TriangleStrip          => vk::PrimitiveTopology::TriangleStrip,
-        Primitive::TriangleStripAdjacency => vk::PrimitiveTopology::TriangleStripWithAdjacency,
-        Primitive::PatchList(_)           => vk::PrimitiveTopology::PatchList,
+        Primitive::PointList              => vk::PrimitiveTopology::POINT_LIST,
+        Primitive::LineList               => vk::PrimitiveTopology::LINE_LIST,
+        Primitive::LineListAdjacency      => vk::PrimitiveTopology::LINE_LIST_WITH_ADJACENCY,
+        Primitive::LineStrip              => vk::PrimitiveTopology::LINE_STRIP,
+        Primitive::LineStripAdjacency     => vk::PrimitiveTopology::LINE_LIST_WITH_ADJACENCY,
```
kvark: this should be `LINE_STRIP_WITH_ADJACENCY`

src/backend/vulkan/src/lib.rs
```diff
@@ -395,21 +401,18 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
                 self.instance
                     .0
                     .create_device(self.handle, &info, None)
-                    .map_err(|err| match err {
-                        ash::DeviceError::LoadError(err) => panic!("{:?}", err),
-                        ash::DeviceError::VkError(err) => Into::<result::Error>::into(err),
-                    })
+                    .map_err(|err| Into::<result::Error>::into(err))
```
kvark: nit: could just be `map_err(Into::<result::Error>::into)`

src/backend/vulkan/src/conv.rs
keringar: Fixed

src/backend/vulkan/src/lib.rs
keringar: Fixed

keringar:
I squashed and ran rustfmt. The failing Windows builds were also fixed. While I don't have access to an Android device to test, I believe the android stuff in `window.rs` has also been correctly updated.

kvark:
Thanks!
Bors r+

PR #2538, opened 2018-12-29
[vulkan] Remove `mem::transmute` usages in `conv.rs` 

bzm3r:
Now that conversion methods abound in `ash` and `hal`, `mem::transmute` may be a thing of the past? 

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/conv.rs
+++ b/src/backend/vulkan/src/conv.rs
@@ -5,27 +5,24 @@ use hal::{buffer, command, format, image, pass, pso, query};
 use hal::{IndexType, PresentMode, Primitive};
 
 use native as n;
-
 use std::borrow::Borrow;
 use std::mem;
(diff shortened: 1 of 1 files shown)


src/hal/Cargo.toml
```diff
@@ -23,6 +23,8 @@ mint = { version = "0.5", optional = true }
 failure = "0.1"
 serde = { version = "1", features = ["serde_derive"], optional = true }
 fxhash = "0.2.1"
+num-traits = "0.2"
```
kvark: I don't think it's worth it including the two extra dependencies for `FromPrimitive`, which can be replaced by a simple `transmute`

bzm3r: @kvark ok, sounds good! those should be removed now,

kvark:
Thanks!
Bors r+

PR #2539, opened 2018-12-30
Support device specific feature requests

### LINKED ISSUE #1928: Device feature requests
We need a mechanism to request particular features of the supported ones for the device.
see https://github.com/szeged/webrender/issues/136#issuecomment-380745757

keringar:
Fixes #1928 by adding an optional parameter containing device features to enable. Currently only actually does anything on the Vulkan back-end since both Metal and DirectX use feature sets rather than fine grained features. If a device does not support all the requested features, than `DeviceCreationError::MissingFeature` is returned. I wasn't too sure what to do with `open_with`, but since it's just a convenience method, I thought it was alright to just not request any additional features. But I'm not sure if we want to expose it or not. Otherwise people can just use `PhysicalDevice::open` instead of `Adapter::open_with`

PR checklist:
    - Vulkan on Linux
    - Vulkan, DX12, DX11, OpenGL on Windows

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -463,8 +463,13 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     unsafe fn open(
         &self,
         families: &[(&QueueFamily, &[hal::QueuePriority])],
+        requested_features: hal::Features,
     ) -> Result<hal::Gpu<Backend>, error::DeviceCreationError> {
         let (device, cxt) = {
+            if !self.features().contains(requested_features) {
+                return Err(error::DeviceCreationError::MissingFeature);
+            }
+
             let feature_level = get_feature_level(self.adapter.as_raw());
             let mut returned_level = d3dcommon::D3D_FEATURE_LEVEL_9_1;
 
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -353,6 +353,7 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     unsafe fn open(
         &self,
         families: &[(&QueueFamily, &[hal::QueuePriority])],
+        requested_features: Features,
     ) -> Result<hal::Gpu<Backend>, DeviceCreationError> {
         let family_infos = families
             .iter()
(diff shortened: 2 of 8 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -463,8 +463,10 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     unsafe fn open(
         &self,
         families: &[(&QueueFamily, &[hal::QueuePriority])],
+        _requested_features: Option<hal::Features>
```
kvark: nit: here and in all similar places, let's have a closing comma at the end

src/backend/vulkan/src/lib.rs
```diff
@@ -353,6 +353,7 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
     unsafe fn open(
         &self,
         families: &[(&QueueFamily, &[hal::QueuePriority])],
+        requested_features: Option<Features>,
```
kvark: let's assert that requested features are a subset of available features

src/hal/src/adapter.rs
```diff
@@ -51,11 +51,13 @@ pub struct MemoryProperties {
 
 /// Represents a physical device (such as a GPU) capable of supporting the given backend.
 pub trait PhysicalDevice<B: Backend>: Any + Send + Sync {
-    /// Create a new logical device.
+    /// Create a new logical device with the requested features. If `requested_features` is
+    /// `None` than only the core features are supported.
```
kvark: wouldn't `None` be equivalent of passing the empty bit mask? in this case, let's not have `Option` at all

src/backend/vulkan/src/lib.rs
keringar: I added an assertion for the Vulkan backend, should the same be done for the other backends as well?

src/hal/src/adapter.rs
keringar: I've changed it to take a Features bitflag directly now.

src/backend/vulkan/src/lib.rs
keringar: On second thought actually, wouldn't it be better to just return an error rather than panicking? We already have the `MissingFeature` error anyway.

kvark:
Sure, that would work too

keringar:
Ok, I've made all the requested changes and added feature support checks to all the back-ends. If the feature isn't supported than MissingFeature is returned.

PR #2542, opened 2018-12-31
Fix typo in physical device Features struct

keringar:
As suggested in PR 2539, this fixes the Features struct to have the correct names. Going by https://www.khronos.org/registry/vulkan/specs/1.1-khr-extensions/html/vkspec.html#features-features the `SHADER_RESIDENCY_IMAGE_2D` and `SHADER_RECIDENSY_IMAGE_3D` should be changed to     `SPARSE_RESIDENCY_IMAGE_2D` and `SPARSE_RESIDENCY_IMAGE_3D`.

PR checklist:


kvark:
Thank you!
bors r+

PR #2548, opened 2019-01-02
[mtl] iOS 9.3 support

mtak-:
- bump metal version
- PrivateCapabilities checking to let basic stuff run on iOS 9
- support MTLLanguageVersion 2.1

Before this PR, gfx-backend-metal was not running on iOS < 12. After this atleast iOS 9.3 is supported (perhaps earlier too).

PR checklist:
- [x] tested examples with the following backends: macos 10.14 metal,
- [x] also tested:  ios 9.3 iphone6, ios 10.3 iphone5s


kvark:
bors try

src/backend/metal/src/command.rs
```diff
@@ -129,6 +129,7 @@ pub struct CommandPool {
     shared: Arc<Shared>,
     allocated: Vec<CommandBufferInnerPtr>,
     pool_shared: PoolSharedPtr,
+    layered_rendering: bool,
```
kvark: The approach of just adding the flags here isn't scalable. Let's make a new type, a bitfield perhaps, something like `EncodingCapabilities`.

mtak-: How about moving `PrivateCapabilities` into `Shared`?

kvark:
That would be fine by me as well

Michael-Lfx:
@mtak- Nice. How about iOS 8? We want to test on iOS 8 device but we haven't those devices.

mtak-:
@Michael-Lfx We don't have any iOS 8 devices either unfortunately. We could try downgrading, but I'm haven't looked into whether that's doable without installing firmware from some third party site.

Michael-Lfx:
> Before this PR, gfx-backend-metal was not running on iOS < 12. 

It's weird. We can runn on iOS 10 iPad mini 4 and iOS 11 iPhone 7p but failed on iOS 12 iPhone 8.

PR #2551, opened 2019-01-03
fix for iphone5s drawable pool size being locked at 3

mtak-:
Before this, if using a default `SwapchaingConfig`, gfx would think the swapchain drawables were changed every 3rd frame under certain conditions on iphone5s release builds (very specifically...). This is because the Drawable pool size is fixed at 3 in release builds for that device, and `SwapchainConfig` defaults to a pool size of 2.

The [API](https://developer.apple.com/documentation/quartzcore/cametallayer/2938720-maximumdrawablecount?language=objc) for setting the drawable pool size is fairly new, but appears to work on _some_ devices, and possibly on older macos versions? So there's only version checking on iOS.

Also, I moved `PrivateCapabilities` into `Shared` as discussed in #2548.

PR checklist:
- [x] tested examples with the following backends: metal macos 10.14
- [x] tested on: iphone 6 (9.3), and iphone5s

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -42,10 +42,13 @@ use std::sync::Arc;
 
 use hal::queue::QueueFamilyId;
 
+use cocoa::foundation::NSInteger;
 use core_graphics::base::CGFloat;
 use core_graphics::geometry::CGRect;
 use foreign_types::ForeignTypeRef;
-use objc::runtime::{BOOL, YES};
+use metal::MTLFeatureSet;
+use metal::MTLLanguageVersion;
+use objc::runtime::{BOOL, Object, YES};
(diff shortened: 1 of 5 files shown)


src/backend/metal/src/lib.rs
```diff
@@ -42,10 +42,13 @@ use std::sync::Arc;
 
 use hal::queue::QueueFamilyId;
 
+use cocoa::foundation::NSInteger;
 use core_graphics::base::CGFloat;
 use core_graphics::geometry::CGRect;
 use foreign_types::ForeignTypeRef;
-use objc::runtime::{BOOL, YES};
+use metal::MTLFeatureSet;
+use metal::MTLLanguageVersion;
```
kvark: nit: let's combine these entries into a single `use`

PR #2553, opened 2019-01-04
[vk] fix push constants offset

### LINKED ISSUE #2552: `push_graphics_constants`/`push_compute_constants` offset parameter is inconsistent across different backends
Here's something that I just ran into.. The Metal backend expects the offset passed to `push_graphics_constants`/`push_compute_constants` [be in bytes](https://github.com/gfx-rs/gfx/blob/master/src/backend/metal/src/command.rs#L506-L507), but the Vulkan backend [expects it to be a multiple of 4 bytes (the size of u32).](https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkan/src/command.rs#L958) This causes user code to break across the different backends.

I'm not sure what the DirectX 12 backend expects in this regard. Could someone familiar with that backend check it out? I'd do it myself, but I'm having problems getting my code to run on that particular backend..

aleksijuvani:
Fixes #2552
PR checklist:

cc @kvark

### DIFF
--- a/src/backend/vulkan/src/command.rs
+++ b/src/backend/vulkan/src/command.rs
@@ -937,7 +937,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                 self.raw,
                 layout.raw,
                 vk::ShaderStageFlags::COMPUTE,
-                offset * 4,
+                offset,
                 memory::cast_slice(constants),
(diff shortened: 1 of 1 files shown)


kvark:
Bors r+
Cc @msiglreith

Issue #1900, opened 2018-04-01
Feature: webgl/stdweb backend


Diggsey:
Using https://docs.rs/webgl_stdweb/0.1.0/webgl_stdweb/struct.GLContext.html

I've tried a couple of times to integrate this, but it's very non-trivial: obviously I'd like to reuse as much as possible from the GL backend, but it was never designed to have a separation between the code which translates GFX concepts into GL concepts, and the code which actually interacts with the OpenGL API.

chpio:
is there a easy way to compile *spirv_cross* to wasm/include it in the build? also it would be really nice to run *spriv_cross* as a build step.

Diggsey:
I made some progress on this: I managed to create true C-like bindings to webgl by borrowing `opengl.js` from emscripten, running it through a custom-built preprocessor to handle the emscripten-specific macros and then incorporating that into a new stdweb-based `gl-generator` backend.

The net result is that I can call the normal OpenGL API on the `wasm32-unknown-unknown` target. Since then I've been working on extricating the dependency on `winit` from the `gfx` opengl backend.

kvark:
@Diggsey u r da the best, man! Please keep up cracking this nut, we are excited to see this coming :)

Boscop:
I'm also very interested in this, what's the current status of this? :)

grovesNL:
@Diggsey is there a branch somewhere that contains your progress on this?

cretz:
As an alternative to stdweb, consider the [recently announced](https://rustwasm.github.io/2018/09/26/announcing-web-sys.html) web-sys crate, e.g. [canvas.get_context("webgl")](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.HtmlCanvasElement.html#method.get_context).unwrap().dyn_into::<web_sys::[WebGlRenderingContext](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.WebGlRenderingContext.html)>()

aep:
whats missing here, anything we can help with?

luckielordie:
Is anyone working on this?

kvark:
Looks like we are back to square one here, with @Diggsey disappearing. At least we can consider `web-sys` without risking of losing any work ;)

Vurich:
I'm working on a Processing-like framework in Rust and this would be amazing to allow me to have an online sketchpad similar to a combination of https://valentin.dasdeck.com/processing/ and the Rust playground.

[...]

PR #2554, opened 2019-01-04
Add wasm32-unknown-unknown/WebGL support

grovesNL:
Start to make a more maintainable version of my old branch from https://github.com/gfx-rs/gfx/issues/1900#issuecomment-449983693. Heavily WIP but maybe somebody wants to start giving some feedback as I work through the remainder of issues.

This PR replaces gl_generator with glow (https://github.com/grovesNL/glow), a experimental crate meant to unify WebGL/OpenGL types and function signatures. Currently unsupported functions will just panic, but I think it makes sense to move some of the version logic and extension fallbacks in there too (removing them from the gl backend). The WebGL types are now just keys (and `Copy`) which fixes some of the issues I mentioned previously.

This is for the wasm32-unknown-unknown target and uses web-sys.

TODO:
- [X] ~~(major) spirv_cross wrapper on wasm32-unknown-unknown and C++ library through emscripten. For now the quad shaders are hardcoded.~~ See https://github.com/grovesNL/spirv_cross/pull/92. (I'll probably leave the PR open until we work out how the integration between wasm bundles should work)
- [X] ~~Consolidate render loop somehow (see WIP in glow example https://github.com/grovesNL/glow/blob/master/examples/hello/src/main.rs#L118). I need some more ideas on how to work around the `'static` bounds for wasm32 here. The basic idea is that wasm32 has to use callbacks and we use spinloops on the native target, so we would ideally like to unify these (at least for use in examples but also simple applications). For wasm32 with the quad example I just render once and quit instead.~~ Currently winit is working through the implementation of "Event Loop 2.0" and there's been [some effort to write a stdweb backend for this](https://github.com/tomaka/winit/pull/797), which we could later help port to wasm32-unknown-unknown. I think for now we can just keep the wasm render loop path separate in gfx examples, and things will work naturally once winit wasm32-unknown-unknown support is ready.
- [X] ~~Remove all remaining `#[cfg(not(target_arch = "wasm32"))]` wherever possible to make this more maintainable.~~ There are a few places remaining that can be removed when we move some logic into glow, like version parsing and extension checking. Some other parts like the runloop can't be addressed without winit support for wasm32-unknown-unknown or other workarounds in quad.
- [X] ~~Decide where to put things like `index.html` for examples (I've excluded it for now) and guide about how to create wasm bundle with wasm-bindgen.~~
- [X] ~~Consider how logging should work – I don't see how to redirect stdout to `console.log` so it's a bit manually at the moment. It would be really cool if things like `info!` just work automatically with wasm32-unknown-unknown.~~ We should be able to use https://github.com/iamcodemaker/console_log for the wasm backend without any major challenges.
- [X] ~~Publish glow~~ Published 0.2.0
- [x] ~~Publish updated spirv_cross~~ Published 0.14.0
- [X] ~~Add wasm-bindgen to CI~~ This has been added

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/2554)


grovesNL:
I spoke to the wasm working group:
* it's not planned for stdout redirection to be stabilized in std anytime soon
* macros `println!` will continue to no-op (probably for a long time, though possibly will have some functionality when a new target is available)
* for `log` it looks like we can use https://github.com/iamcodemaker/console_log

grovesNL:
For spirv_cross, I've been able to compile the C++ portion of spirv_cross to a WebAssembly+JavaScript bundle through Emscripten.

The example below loads a SPIR-V, passes it to SPIRV-Cross, and compiles to GLSL.

screen shot 2019-01-13 at 12 40 14 pm

Next we need to redirect the Rust FFI calls. When building for native targets these calls will still go direct to C++, but when building for wasm32-unknown-unknown they should go through the JS bindings provided in the Emscripten bundle (probably just exposed as globals on the `window` object in JS). This step is a bit unfortunate, because I'd ideally like to keep calls within wasm only, but I don't see a good way around this for now. If anyone knows a better way please let me know 😃

kvark:
>they should go through the JS bindings provided in the Emscripten bundle

I find it weird that you are targeting `unknown` while still using Emscripten. I thought the whole point was to avoid it?

grovesNL:
@kvark 
> I find it weird that you are targeting unknown while still using Emscripten. I thought the whole point was to avoid it?

The Emscripten build I'm talking about is only the C++ portion of SPIRV-Cross, so we can run SPIRV-Cross in the browser (the part in my screenshot). The idea is to have something like a `window.SPIRVCross` (the C++ portion, compiled by Emscripten) that will later be referenced by spirv_cross (the Rust portion, compiled as wasm32-unknown-unknown). I'm not sure of a way around this besides rewriting the C++ portion in Rust.

@minecrawler 
> I don't think it's an issue, because isn't that the norm? It's an inconvenience at most, and might have to be documented or tooling has to be created for it.

Right, but I'm talking about the transitive dependency on the C++ library here: application (Rust, compiled with wasm32-unknown-unknown) -> gfx-backend-gl (Rust) -> spirv_cross (Rust) -> SPIRV-Cross (C++, compiled with Emscripten to JS and wasm)

The application's index.html has to include both the wasm produced by rustc but also the Emscripten bundle so the `window.SPIRVCross` is already in place by the time the application's wasm runs.

The alternative is that spirv_cross includes the necessary JS and wasm inline somewhere and causes it to be evaluated by the JS engine somehow during initialization of the application, or we could use a callback/return a Future somewhere in spirv_cross. The problem is that I don't know the best way to do this 😄 I can't find much precedent for this either but I expect lots of libraries wrapping C or C++ libraries will have to deal with this. I think it will be even more complicated when using bundlers like Parcel (I just load it manually in the HTML file for now).

grovesNL:
To clarify, I'm able to compile SPIRV-Cross to wasm with Emscripten without a problem. The problem is I'm not sure of the best method to link the Rust wasm (wasm32-unknown-unknown) and C++ wasm (Emscripten) parts together besides the ideas I mentioned above.

It would be great if we could statically link the wasm modules together, but it's not clear how this would be possible without compiling both parts with Emscripten in the first place (at least with the Emscripten output generated from SPIRV-Cross), which we'd like to avoid.

carlosdp:
The stack overflow link in my last comment walks through a way to link the
wasm directly using the binaryen tools

grovesNL:
I spent a few hours today looking into how dynamic linking should work in WebAssembly generally, and Rust with WebAssembly. It seems like both of these are still under consideration, so it looks like it's too early to make any progress with more direct linking.

FWIW there is a clang target for wasm32-unknown-unknown (just like rustc has), so one idea could be to generate LLVM IR or LLVM bitcode for SPIRV-Cross, combine this directly with IR/bitcode generated by rustc, and eventually generate a single wasm bundle. However there needs to be improvements in tooling on both the C++ and Rust sides for this, so I don't think I'm going to keep looking into this option for now. https://github.com/yurydelendik/wasmception is a great reference for playing with this target.

grovesNL:
Another small update: I continued experimenting with the idea of having Rust wasm32-unknown-unknown make calls to JavaScript (through wasm bindgen), where the JavaScript methods are just the Emscripten generated C++ bindings.

So each call into the C++ SPIRV-Cross library roughly looks like Rust (wasm, wasm32-unknown-unknown) -> JavaScript -> C++ (wasm Emscripten). I manually wrote some bindings for the GLSL compiler and it works really well so far:

screen shot 2019-02-02 at 1 04 48 am

The screenshot is running a simple Rust binary compiled to wasm32-unknown-unknown which has spirv_cross as a dependency. It includes a SPIR-V file, creates a GLSL compiler using the regular spirv_cross API and compiles SPIR-V to GLSL. I think it's pretty impressive that we can do this without changing the API of spirv_cross at all.

As I mentioned earlier, there are a few downsides to this approach:
- The Emscripten C++ files (one JS file and one wasm file) need to be loaded by the application (in their own HTML file), because I don't see we could easily embed it as part of gfx at the moment. This is conceptually similar to linking to installed libraries. There might be a better way to do this in the future, so that the JS and wasm are included inline somehow (basically JS snippets, which are planned for wasm-bindgen).
- Calling through JS adds overhead, so it would be nice to be able to call directly between both these wasm modules instead.
- The Emscripten bundle is a bit large, so it there could be some improvements by using clang wasm32-unknown-unknown directly when it's more stable. After this step, it would be great to have both Rust and C++ code in the same wasm module.

carlosdp:
I have your branch running locally now and would like to help. I can try to tackle the render loop consolidation if that helps. Have you started trying to make the spirv_cross shim yet?

grovesNL:
Sounds great! The render loop consolidation would be useful. I'm mostly interested in an idiomatic callback-based way to create and run a renderloop with a cleanup at the end, and allows moving resources etc. into the cleanup. I played around with the idea wrapping all resources into struct fields that are `Option` but maybe there are some other ideas.

I'm also wondering whether we could add wasm32/requestAnimationFrame support to winit somehow, if they're ready to support wasm32-unknown-unknown. Then most projects would be able to just switch to winit's callback-driven API instead.

I've started to work on the spirv_cross shim (the screenshot above is running an example), currently I'm just filling out the last few bindings we should need. Calls are made from Rust/wasm32-unknown-unknown -> JS -> C++/Emscripten wasm as mentioned above. So far the spirv_cross work has been mostly adding some helpers to handle pointer offsets and string conversion correctly wherever the previous assumptions made for C FFI aren't valid anymore (i.e. reading a string from the Emscripten heap vs. reading a string from a pointer).

grovesNL:
spirv_cross as wasm32-unknown-unknown (using the ideas mentioned above) is mostly ready now: see grovesNL/spirv_cross#92. It would be great if somebody could review and consider any possible issues with this approach.

Here's a spirv_cross example with native and wasm32-unknown-unknown running side-by-side:

image

grovesNL:
Some more updates:

- Rebased this branch onto master again
- Added [console error panic hook](https://github.com/rustwasm/console_error_panic_hook) (to get stack traces when panics occur) and [console log](https://crates.io/crates/console_log) (to write `info!` etc. to the console) to the quad example
- Confirmed the wasm32-unknown-unknown version of spirv_cross works with this branch, by loading spirv and converting it to ESSL at runtime (in the browser)

Screen Shot 2019-03-18 at 11 38 57 PM

grovesNL:
Also I just looked up Emscripten's emulation for `glMapBufferRange` to compare as a reference, it's located at https://github.com/emscripten-core/emscripten/blob/incoming/src/library_webgl.js#L3320

Comparing my emulation vs. Emscripten's emulation, they appear to use the same approach:
- during `glMapBufferRange`, `malloc` some memory internally, return a pointer to it
- during `glUnmapBufferRange`, upload it

The emulation in this PR is less complete than Emscripten's version for now, but we could expand it afterwards.

kyren:
I've been testing out the PR, and I noticed that I think based on browser defaults and blacklisting, the webgl2 context is sometimes created with multisampling enabled by default.  This interferes with glBlitFramebuffer, which cannot blit to the multisampled screen.

Changing [this code](https://github.com/grovesNL/gfx/blob/9573a23019d16c8f9aa1281eef9deb9e78b25ad7/src/backend/gl/src/lib.rs#L255) that creates the webgl2 context to something like:

```
        #[cfg(target_arch = "wasm32")]
        let context = {
            use wasm_bindgen::JsCast;
            use js_sys::Map;
            let options = js_sys::JSON::parse(r#"{ "antialias": false }"#).unwrap();

            let canvas = web_sys::window()
                .unwrap()
                .document()
                .unwrap()
                .get_element_by_id(webgl_context_id.unwrap())
                .unwrap()
                .dyn_into::<web_sys::HtmlCanvasElement>()
                .unwrap();
            let webgl2_context = canvas
                .get_context_with_context_options("webgl2", &options.into())
                .unwrap()
                .unwrap()
                .dyn_into::<web_sys::WebGl2RenderingContext>()
                .unwrap();
            glow::web::Context::from_webgl2_context(webgl2_context)
        };
```

Seems to solve the issue.  Also, should `stencil`, `depth`, and `alpha` also be set to `false` in options as well?  I think so based on my understanding of the gl backend, but I might be wrong here?

grovesNL:
Sounds god, we could probably disable `antialias` and `alpha` for now at least. I think `stencil` and `depth` on the default framebuffer should be ok.

FWIW for testing locally I just set it directly in index.html. `getContext("webgl2", {antialias: false})` on the canvas element should cause later `getContext` calls to get the same instance

msiglreith:
Would it be possible to import `glow` as `gl`? I would prefer keeping the current style regarding `gl::TEXTURE_2D` etc.

grovesNL:
> Would it be possible to import `glow` as `gl`?

Sure, I don't have a preference here if everyone is fine with renaming it at the import level. It should reduce the diff slightly too.

kvark:
I don't have a strong feeling about this. The diff will only be spared on the constant usage (e.g. `gl::TEXTURE_2D`), while entry points are still different. @msiglreith do you feel strongly? or was it just a suggestion to consider?

PR #2555, opened 2019-01-05
MTLCompareFunction requires iOS 9.0

Michael-Lfx:
Fixes #issue
PR checklist:
   - [x]metal

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -1566,7 +1566,7 @@ impl hal::Device<Backend> for Device {
         unsafe { descriptor.set_lod_bias(info.lod_bias.into()) };
         descriptor.set_lod_min_clamp(info.lod_range.start.into());
         descriptor.set_lod_max_clamp(info.lod_range.end.into());
-        
+
         let caps = &self.private_caps;
         // TODO: Clarify minimum macOS version with Apple (43707452)
         if (caps.os_is_mac && caps.has_version_at_least(10, 13)) ||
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1575,7 +1575,10 @@ impl hal::Device<Backend> for Device {
         }
 
         if let Some(fun) = info.comparison {
-            descriptor.set_compare_function(conv::map_compare_function(fun));
+            if (caps.os_is_mac && caps.has_version_at_least(10, 11)) ||
+                (!caps.os_is_mac && caps.has_version_at_least(9, 0)) {
+                descriptor.set_compare_function(conv::map_compare_function(fun));
```
kvark: We should return an error otherwise. If the user requests a comparison function, we can't just ignore it.

Michael-Lfx: But there is not suitable error of `AllocationError` for the case.

kvark:
We can reuse an existing code and error! the cause, or add a new code aka generic Unsupported.

src/backend/metal/src/device.rs
kvark: If we can't return an error, we should panic. It's not like the user is going to get anything meaningful anyway if they ask for a comparison sampler but don't get it.

Michael-Lfx: Should we add a bool field in `PrivateCapabilities` to indicate that whether `set_compare_function` works on the current platform?

kvark:
Yes, sure!
Note that this issue has been discussed in MoltenVK repo with the conclusion that we still support comparison samplers if they are immutable (provided at the descriptor set layout creation), so your application can still use them. (Link needed)

kvark:
@Michael-Lfx 
> Should we add a bool field in PrivateCapabilities to indicate that whether set_compare_function works on the current platform?

Yes, we need that.
Important to note that we'd still allow the samplers to be created, but disallow them being used in immutable sampler bindings (in `create_xxx_descriptor_sets`).
And as I mentioned, we need to panic right away if this isn't followed by the user.

kvark:
Closed by #2885

PR #2558, opened 2019-01-07
Fix crash on iPhone5s iOS 12.1

Michael-Lfx:
Crash log: -[MTLDebugRenderCommandEncoder setDepthClipMode:]:2621: failed assertion `-[MTLDebugRenderCommandEncoder setDepthClipMode:] is only supported on MTLFeatureSet_iOS_GPUFamily2_v1 and later'

Fixes #issue
PR checklist:
   - [x] Metal iPhone5s iOS 12.1


mtak-:
@Michael-Lfx thanks for fixing that! I didn't see depth clip mode mentioned in the feature sets. The apple docs say it was introduced in iOS 11. I guess it was undocumented before that.

PR #2562, opened 2019-01-09
Improve secondary and subpass command buffers

madadam:
This PR:
- implements `Submittable` for `SubpassCommandBuffer`, thus allowing it to be recorded into primary command buffers. 
- Modifies `CommandPool::free` to allow freeing also secondary and subpass command buffers.

PR checklist:
- [x] tested examples with the following backends: quad (vulkan, gl)

### DIFF
--- a/src/hal/src/command/raw.rs
+++ b/src/hal/src/command/raw.rs
@@ -526,3 +526,9 @@ pub trait RawCommandBuffer<B: Backend>: Any + Send + Sync {
         T: 'a + Borrow<B::CommandBuffer>,
         I: IntoIterator<Item = &'a T>;
 }
+
+/// A trait for types that can be converted into raw command buffer.
+pub trait IntoRawCommandBuffer<B: Backend, C> {
+    /// Converts into raw command buffer.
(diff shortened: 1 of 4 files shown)


src/hal/src/command/raw.rs
```diff
@@ -526,3 +526,9 @@ pub trait RawCommandBuffer<B: Backend>: Any + Send + Sync {
         T: 'a + Borrow<B::CommandBuffer>,
         I: IntoIterator<Item = &'a T>;
 }
+
+/// A trait for types that can be converted into raw command buffer.
+pub trait IntoRawCommandBuffer<B: Backend, C> {
```
kvark: Did you consider trying the standard `Into` instead of rolling out a new trait?

kvark: why do you need the `Capability` bound on the trait anyway?

kvark: Good point. The unsafety story is rather poor, still, but this simplification doesn't worth it. Let's proceed with your trait and `C` then.

PR #2566, opened 2019-01-11
[vk] Extend pipeline cache functionalities

zakorgy:
Depends on https://github.com/gfx-rs/gfx/pull/2565

Added the `get_pipeline_cache_data` function to query pipeline cache data from Device. Added the optional `data` member to `create_pipeline_cache` where we can supply the bytes of a loaded cache.

PR checklist:
  - Vulkan (Windows, Linux)

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/2566)

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -904,7 +904,10 @@ impl hal::Device<Backend> for Device {
         })
     }
 
-    unsafe fn create_pipeline_cache(&self, _data: Option<&[u8]>) -> Result<(), device::OutOfMemory> {
+    unsafe fn create_pipeline_cache(
+        &self,
+        _data: Option<&[u8]>,
+    ) -> Result<(), device::OutOfMemory> {
         Ok(())
     }
 
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -904,8 +904,13 @@ impl hal::Device<Backend> for Device {
         })
     }
 
-    fn create_pipeline_cache(&self) -> Result<(), device::OutOfMemory> {
+    unsafe fn create_pipeline_cache(&self, _data: Option<&[u8]>) -> Result<(), device::OutOfMemory> {
         Ok(())
+    }
+
+    unsafe fn get_pipeline_cache_data(&self, cache: &()) -> Result<Vec<u8>, device::OutOfMemory> {
+        //empty
+        Ok(Vec::new())
     }
 
     unsafe fn destroy_pipeline_cache(&self, _: ()) {
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -285,7 +285,7 @@ impl d::Device<B> for Device {
         &self,
         data: Option<&[u8]>,
     ) -> Result<n::PipelineCache, d::OutOfMemory> {
-        let (data_len, data) = if let Some(d) =  data {
+        let (data_len, data) = if let Some(d) = data {
             (d.len(), d.as_ptr())
         } else {
             (0_usize, ptr::null())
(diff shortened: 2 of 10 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -904,10 +904,15 @@ impl hal::Device<Backend> for Device {
         })
     }
 
-    fn create_pipeline_cache(&self) -> Result<(), device::OutOfMemory> {
+    fn create_pipeline_cache(&self, _data: Option<Vec<u8>>, device: &PhysicalDevice) -> Result<(), device::OutOfMemory> {
```
kvark: this should be `Option<&[u8]>`

kvark: why do we need the physical device here? It's not in https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkCreatePipelineCache.html

src/backend/vulkan/src/device.rs
```diff
@@ -281,13 +281,70 @@ impl d::Device<B> for Device {
         }
     }
 
-    fn create_pipeline_cache(&self) -> Result<n::PipelineCache, d::OutOfMemory> {
+    fn create_pipeline_cache(
+        &self,
+        data: Option<Vec<u8>>,
+        device: &PhysicalDevice,
+    ) -> Result<n::PipelineCache, d::OutOfMemory> {
+        fn bad_cache(d: &[u8], properties: &vk::PhysicalDeviceProperties) -> bool {
```
kvark: nit: let's reverse the logic and call it `is_valid` or something

src/backend/vulkan/src/device.rs
```diff
@@ -281,13 +281,70 @@ impl d::Device<B> for Device {
...
+            if vendor_id != properties.vendor_id as u32 {
+                warn!(
+                    "Vendor ID mismatch. Device: {:?}, cache: {:?}.",
+                    properties.vendor_id,
+                    vendor_id,
+                );
+                return true;
+            }
+
+            let device_id = unsafe { std::mem::transmute::<[u8; 4], u32>([d[12], d[13], d[14], d[15]]) };
```
kvark: looks like we should just cast the first 4 dwords right away, using `std::slice::from_raw_parts`

src/backend/dx11/src/device.rs
kvark: these methods should be unsafe, just like all the other ones here

zakorgy: I couldn't find other way to access the `PhysicalDeviceProperties`  (for the cache validation) from the Device, so I placed the `&PhysicalDevice` here.
I also meant to propagate the validation to the create users, but they can't access backend specific resources e.g. `vk::PipelineCacheHeaderVersion` or `properties.pipeline_cache_uuid` from `gfx-hal`.
My other idea is to move the cache validation into the `PhysicalDevice` and the crate users can call it (e.g. `adpater.physical_device.is_valid_cache(&cache_bytes)`) before `create_pipeline_cache`.

kvark: We don't have to, strictly speaking, validate the caches on `create_pipeline_cache`, at least not at the API level. The backend can still keep VkPhysicalDevice around when creating a logical device and use it for invalidation.

src/backend/vulkan/src/lib.rs
```diff
@@ -798,6 +798,50 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
             max_sampler_anisotropy: limits.max_sampler_anisotropy,
         }
     }
+
+    fn is_valid_cache(&self, cache: &[u8]) -> bool {
+        let cache_info: &[u32] = unsafe { slice::from_raw_parts(cache as *const _ as *const _, 4) };
```
kvark: need to assert/check here that the length is at least 16

src/backend/vulkan/src/lib.rs
```diff
@@ -798,6 +798,50 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+
+        if self.properties.pipeline_cache_uuid != cache[16..16 + vk::UUID_SIZE] {
+            warn!(
+                "Pipeline cache UUID mismatch. Device: {:?}, cache: {:?}.",
+                self.properties.pipeline_cache_uuid,
+                &cache[16..16 + vk::UUID_SIZE],
+            );
+            return false;
+        }
+        return true;
```
kvark: nit: no need for `return` here

PR #2451, opened 2018-10-21
Add windows support for travis CI

msiglreith:
Let's see if it works

PR #2567, opened 2019-01-14
Add note about `examples` directory

grovesNL:
Add note about switching `cwd` to `examples` to further clarify the comment in the code block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,16 +32,18 @@ gfx-rs is a low-level, cross-platform graphics abstraction library in Rust. It c
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


kvark:
bors retry

grovesNL:
bors r=kvark

grovesNL:
@kvark is it because AppVeyor needs to report to bors unconditionally or something (even if AppVeyor skips this because it doesn't see a diff)?

kvark:
Time to drop appveyor? :)

grovesNL:
If there are no concerns about #2451 then I think we could try it out and see how it goes. @msiglreith what do you think?

PR #2569, opened 2019-01-14
[MTL] Set viewport and scissor for clear

dati91:
PR checklist:

Without setting the viewport/scissor, metal will use a default 0x0 rect and clear nothing.
Also there is an Y flip for the clear. (Note: the blit code also uses the same Y flip)


kvark:
I don't understand why setting the viewport/scissor is needed here? I believe the default render pass has those as non-zero.

PR #2575, opened 2019-01-16
update doc for Adapter and Device

Michael-Lfx:
Fixes #issue
PR checklist:

### DIFF
--- a/src/hal/src/device.rs
+++ b/src/hal/src/device.rs
@@ -6,7 +6,7 @@
 //! and managing graphics resources such as buffers, images and memory.
 //!
 //! The `Adapter` and `Device` types are very similar to the Vulkan concept of
-//! "physical devices" vs. "logical devices"; an `Adapter` is single GPU
+//! "logical devices" vs. "physical devices"; an `Adapter` is single GPU
 //! (or CPU) that implements a backend, a `Device` is a
 //! handle to that physical device that has the requested capabilities
 //! and is used to actually do things.


src/hal/src/device.rs
```diff
@@ -6,7 +6,7 @@
 //! and managing graphics resources such as buffers, images and memory.
 //!
 //! The `Adapter` and `Device` types are very similar to the Vulkan concept of
-//! "physical devices" vs. "logical devices"; an `Adapter` is single GPU
+//! "logical devices" vs. "physical devices"; an `Adapter` is single GPU
```
kvark: I don't think this change is correct. The line above says "Adapter" followed by "Device, so here it should be "physical device" followed by "logical device". The old comment is correct, therefore.

Michael-Lfx: Here’s to say the “Adapter” equals to “physical device” and “Device” is “logical device”?

kvark: That is correct, yes

Michael-Lfx: Thank you.

grovesNL: @kvark what do you think about aligning the naming with Vulkan here?

kvark:
We do have a PhysicalDevice, it's just included in the Adapter with more info, so I don't think getting closer to Vk would be giving us much.

grovesNL:
Yeah I was just wondering because I think a couple others were confused about this too (can't search for "adapter" in Vulkan docs etc.)

PR #2577, opened 2019-01-17
Add note about `examples` directory

grovesNL:
Add note about switching `cwd` to examples to further clarify the comment in the code block.

#2567 accidentally used the wrong fork – not sure if that was causing issues with bors. Also fixed a typo in dx12, which should run AppVeyor just in case.


grovesNL:
Oh, AppVeyor is actually having a major outage right now

PR #2590, opened 2019-01-20
[dx12] Immediately bind VertexBuffers when called instead of caching

### LINKED ISSUE #2280: `bind_vertex_buffer` inside render pass on dx12
Short info header:
- GFX version: commit e8254cd
- OS: Windows 10
- GPU: Nvidia GTX 980 ti

Project:
https://gitlab.com/termhn/game/blob/new-renderer-indirect/voxygen/src/renderer.rs#L364

When using vulkan backend, performs as expected with each draw call having bound a new vertex buffer, producing a normal looking image:
image

When using dx12 backend, only the final bound vertex buffer is bound for all of the draw calls, producing:
image

Brekcel:
Fixes #2280
PR checklist:
- [*] `rustfmt` run on changed code

Previously it would cache the vertex buffer's to bind until a draw was called as it required info from the pipeline about how to bind them (specifically the stride). It would also, as an optimization, remove that information after it bound something so that nothing else could be bound until a new pipeline was bound. This PR moves the binding to actually happen when bind_vertex_buffers is called instead of caching them. This gives a more expected behavior and removes the incorrect optimization, however now a pipeline MUST be bound BEFORE bind_vertex_buffers is called.

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -297,11 +297,8 @@ pub struct CommandBuffer {
     occlusion_query: Option<OcclusionQuery>,
     pipeline_stats_query: Option<UINT>,
 
-    // Cached vertex buffer views to bind.
-    // `Stride` values are not known at `bind_vertex_buffers` time because they are only stored
-    // inside the pipeline state.
+    // VertexBindings for the currently bound pipeline
     vertex_bindings_remap: [Option<r::VertexBinding>; MAX_VERTEX_BUFFERS],
-    vertex_buffer_views: [d3d12::D3D12_VERTEX_BUFFER_VIEW; MAX_VERTEX_BUFFERS],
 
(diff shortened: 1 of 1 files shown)


kvark:
If the vertex buffer binding depends on the pipeline, we can't *just* bind them right the `bind_vertex_buffers` call, because this one has no associated pipeline layout (unlike, say, descriptor binding, where the user does explicitly specify the layout). So there could be any pipelines switching after the vertex buffers are bound, and we need to adjust for that, on every pipeline switch.

What Metal backend does is binding at both `bind_vertex_buffers` call and `bind_graphics_pipeline`, but not strictly: only binding what could be bound (i.e. compatible), and just asserting on draw that all the binding succeeded.

msiglreith:
Thanks for taking care of this issue. I still need some time to review and test how this interacts with the remapping and missing stride information.

Issue #2579, opened 2019-01-17
Improve DX12 initialization time


kvark:
See https://github.com/gfx-rs/gfx/issues/2406 for some numbers.
See https://github.com/gpuweb/gpuweb/pull/67#issuecomment-422079901 for the number of things we could be doing lazily.

Also a small benchmark by @bzm3r :

| state | Vulkan | D3D12 |
| ----- | -------  | -------  |
|start  | 0.0012000000000000001ms | 0.0006000000000000001ms |
|init...  | 0.803ms | 3.8476ms |
|creating instance...  | 0.9356000000000001ms | 3.9911ms |
|done | 168.6831ms | 48.5743ms |
|adapter picked | 169.2583ms | 184.7913ms |
|device opened | 178.2687ms | 316.2624ms |
|compiling... | 179.2792ms | 318.06170000000003ms |
|done: 1168 bytes | 256.3115ms | 410.0108ms |
|fill_payload... | 264.73240000000004ms | 445.28880000000004ms |
|execute_compute... | 288.8599ms | 470.3587ms |
|check_result... | 291.2221ms | 471.33090000000004ms |
|clean_up... | 291.54620000000006ms | 471.65510000000006ms |
|done... | 357.7824ms | 493.82220000000007ms |

msiglreith:
These could be done lazily:
* Collect all feature information
* Query all format features
* Create command signatures for indirect execution

Veykril:
I'd be interested in trying to work on this but I am not quite sure how to make the mentioned things lazy like for example format_properties (I assume this is meant with format features) of the physical device. Especially cause this is behind an `Arc` afterwards so lazily initializing it on first access would mean to put it behind an `Arc<Mutex>` or something similar if i am not mistaken right?
So if someone could mentor me slightly on this I would really appreciate it.

kvark:
@Veykril that would be great!
Yes, locking the format properties in `Arc<Mutex>` would be fine, it's not meant to be accessed too often, so no contention is expected. Please hop on our [gitter](https://gitter.im/gfx-rs/gfx) for more details about how this can work.

PR #2594, opened 2019-01-23
Make Format Properties Lazy for dx12 backend

Veykril:
Progresses #2579

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -1946,7 +1946,7 @@ impl d::Device<B> for Device {
         let buffer = buffer.expect_bound();
         let buffer_features = {
             let idx = format.map(|fmt| fmt as usize).unwrap_or(0);
-            self.format_properties[idx].buffer_features
+            self.format_properties.get(idx).buffer_features
         };
         let (format, format_desc) = match format.and_then(conv::map_format) {
             Some(fmt) => (fmt, format.unwrap().surface_desc()),
@@ -2044,7 +2044,7 @@ impl d::Device<B> for Device {
         let block_dim = format_desc.dim;
         let extent = kind.extent();
 
-        let format_properties = &self.format_properties[format as usize];
+        let format_properties = self.format_properties.get(format as usize);
         let (layout, features) = match tiling {
             image::Tiling::Optimal => (
                 d3d12::D3D12_TEXTURE_LAYOUT_UNKNOWN,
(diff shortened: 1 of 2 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -2044,7 +2041,7 @@ impl d::Device<B> for Device {
         let block_dim = format_desc.dim;
         let extent = kind.extent();
 
-        let format_properties = &self.format_properties[format as usize];
+        let format_properties = &self.format_properties.get(format as usize);
```
kvark: nit: the reference is no longer needed here

src/backend/dx12/src/device.rs
```diff
@@ -2215,7 +2212,7 @@ impl d::Device<B> for Device {
         //TODO: the clear_Xv is incomplete. We should support clearing images created without XXX_ATTACHMENT usage.
         // for this, we need to check the format and force the `RENDER_TARGET` flag behind the user's back
         // if the format supports being rendered into, allowing us to create clear_Xv
-        let format_properties = &self.format_properties[image_unbound.format as usize];
+        let format_properties = &self.format_properties.get(image_unbound.format as usize);
```
kvark: and here

src/backend/dx12/src/lib.rs
```diff
@@ -1172,3 +1090,116 @@ fn validate_line_width(width: f32) {
...
+    fn drop(&mut self) {
+        unsafe {
+            self.1.destroy();
+        }
+    }
+}
+
+impl FormatProperties {
+    fn new(device: native::Device) -> Self {
+        let mut arr: [Mutex<Option<f::Properties>>; f::NUM_FORMATS] =
```
kvark: let's just make it a `Box<[]>`, given that we can construct a `Vec` and convert it into that with https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.into_boxed_slice

src/backend/dx12/src/lib.rs
```diff
@@ -1172,3 +1090,116 @@ fn validate_line_width(width: f32) {
...
+        for entry in arr.iter_mut().skip(1) {
+            unsafe { ptr::write(entry, Mutex::new(None)) };
+        }
+        FormatProperties(arr, device)
+    }
+
+    fn get(&self, idx: usize) -> f::Properties {
+        let mut guard = self.0[idx].lock().unwrap();
+        if let Some(props) = *guard {
+            props
```
kvark: let's just `return` here to avoid indenting the rest of the code

src/hal/src/device.rs
```diff
@@ -265,10 +265,16 @@ pub trait Device<B: Backend>: Any + Send + Sync {
     unsafe fn destroy_pipeline_layout(&self, layout: B::PipelineLayout);
 
     /// Create a pipeline cache object.
-    unsafe fn create_pipeline_cache(&self, data: Option<&[u8]>) -> Result<B::PipelineCache, OutOfMemory>;
+    unsafe fn create_pipeline_cache(
```
kvark: note: if you are applying rustfmt changes, please make it a separate commit independent of the functional changes

Veykril: Reverted them, didnt notice this when I ran it

kvark:
bors retry

PR #2595, opened 2019-01-23
add device_type inference to gl backend

### LINKED ISSUE #2593: Use gl.VERSION to infer device_type?
Short info header:
- GFX version:master
- OS:all
- GPU:all

Could we use the string provided by gl.VERSION to infer the device_type instead of always returning `DiscreteGpu`.  We should be able to infer correctly a high % of the time by just inspect for the presence of nvidia, intel, or amd in the version string.  If we collected some of the common cases here we could get a heuristic possibly.

Relevant code would be here:
https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/lib.rs#L274

Failing this, should we add an unknown type for cases where we have no idea?

jackmott:
A first pass at inferring the device type.  If we see "intel" in the device string we can be confident that it is an integratedgpu.  If we see "nvidia", "ati", or "amd" it is either going to be discrete or have performance much better than the intel integrated gpu. Otherwise we set device_type to Other since we really have no idea.

Fixes #2593 
PR checklist:
- [x] tested examples with the following backends:gl

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -256,7 +256,7 @@ impl PhysicalDevice {
             debug!("- {}", *extension);
         }
         let name = info.platform_name.renderer.into();
-        let vendor: std::string::String = info.platform_name.vendor.into();
+        let renderer: std::string::String = info.platform_name.renderer.into();
 
         // create the shared context
         let share = Share {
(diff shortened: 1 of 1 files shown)


jackmott:
Maybe this whole idea is no good. Intel has discrete cards coming:
https://www.tomshardware.com/news/intel-discrete-gpu-2020-raja,37289.html

Should we always say Other?  Should we add an Unknown to the Device_Type enum? This would break from the Vulkan API though.

kvark:
Note that AMD also has APUs, which are integrated... I wonder we can do a more precise matching of the adapter name to figure this out. As for the current PR, I don't feel strongly either way. Anyone got a strong opinion here?

grovesNL:
Not really, I'd slightly prefer a stronger heuristic than vendor.

jackmott:
I'm going to do some research on vendor/version/renderer strings and see if we can do better. I've got quite a few computers here to survey.

jackmott:
I've got a simple command line program that will print renderer vender and version:
https://github.com/jackmott/rustglstats

I've started compiling results for computers I have access to:
https://docs.google.com/spreadsheets/d/1va19XlGv9VuUEF99ekkWhDoTroYOp-QRAIU457lAnM4/edit?usp=sharing

if any of you want to add to it or pass this around please do

kvark:
Could also use `glxinfo` that is included by default in Linux. There is also a database in https://opengl.gpuinfo.org/

jackmott:
I figured something like that would exist already

jackmott:
if anyone has an "IGP" or "nForce" nvidia card to run glxinfo on I'd like to see the output.  the opengl database doesn't have any examples.

src/backend/gl/src/lib.rs
```diff
@@ -272,17 +272,32 @@ impl PhysicalDevice {
...
-            hal::adapter::DeviceType::IntegratedGpu
-        } else if vender_lower.contains("nvidia")
-            || vender_lower.contains("ati")
-            || vender_lower.contains("amd")
+        // opengl has no way to discern device_type, so we can try to infer it from the renderer string
+        
+        let renderer_lower = renderer.to_lowercase();
+        let strings_that_imply_integrated = [
+            " xpress", // space here is on purpose so we don't match express
+            "radeon hd 4200",
```
kvark: any reason to not match just `"radeon hd "?

jackmott: yes, most radeon hd gpus are not integrated

kvark:
oh, ok
let's proceed!
bors r+

PR #2596, opened 2019-01-24
Add missing acquire image/present errors

omni-viral:
Fixes #issue
PR checklist:
- [x] tested examples with the following backends: metal, gl

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -15,7 +15,7 @@ use hal::{
     queue::{RawCommandQueue, Submission},
     range::RangeArg,
     DrawCount, IndexCount, IndexType, InstanceCount, SwapImageIndex, VertexCount, VertexOffset,
-    WorkGroupCount, window::PresentError,
+    WorkGroupCount, window::{PresentError, Suboptimal},
 };
 
 use block::ConcreteBlock;
@@ -2078,7 +2078,7 @@ impl RawCommandQueue<Backend> for CommandQueue {
         &mut self,
         swapchains: Is,
         wait_semaphores: Iw,
-    ) -> Result<(), PresentError>
+    ) -> Result<Option<Suboptimal>, PresentError>
     where
         W: 'a + Borrow<window::Swapchain>,
         Is: IntoIterator<Item = (&'a W, SwapImageIndex)>,
@@ -2131,7 +2131,7 @@ impl RawCommandQueue<Backend> for CommandQueue {
             }
         }
 
-        Ok(())
+        Ok(None)
     }
 
     fn wait_idle(&self) -> Result<(), error::HostExecutionError> {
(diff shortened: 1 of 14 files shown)


kvark:
LGTM, let's just make CI happy

omni-viral:
👍

kvark:
@omni-viral please rebase

omni-viral:
I should return to this soon

src/backend/metal/src/command.rs
```diff
@@ -2072,7 +2072,7 @@ impl RawCommandQueue<Backend> for CommandQueue {
 
             for (swapchain, index) in swapchains {
                 debug!("presenting frame {}", index);
-                let drawable = swapchain.borrow().take_drawable(index)?;
+                let drawable = swapchain.borrow().take_drawable(index).map_err(|()| PresentError::OutOfDate)?; // What `Err(())` represents?
```
termhn: This seems related to #2659

src/backend/vulkan/src/lib.rs
```diff
@@ -959,8 +960,12 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
         };
 
         match self.swapchain_fn.queue_present_khr(*self.raw, &info) {
-            vk::Result::SUCCESS => Ok(()),
-            vk::Result::SUBOPTIMAL_KHR | vk::Result::ERROR_OUT_OF_DATE_KHR => Err(()),
+            vk::Result::SUBOPTIMAL_KHR | vk::Result::SUCCESS => Ok(()),
```
termhn: Should we return Suboptimal as part of the Ok(()) in case user wants to handle it?

src/hal/src/window.rs
```diff
@@ -414,7 +444,7 @@ pub trait Swapchain<B: Backend>: Any + Send + Sync {
         timeout_ns: u64,
         semaphore: Option<&B::Semaphore>,
         fence: Option<&B::Fence>,
-    ) -> Result<SwapImageIndex, AcquireError>;
+    ) -> Result<(SwapImageIndex, Option<Suboptimal>), AcquireError>;
```
kvark: would you consider instead returning a struct like
```rust
struct AcquireFrame {
  index: SwapImageIndex,
  suboptimal: bool,
}
```

omni-viral: And for `present`?

kvark: Oh, this is a bit unfortunate. Would it be simpler if we just add `is_optimal` method to the `Swapchain`? It's not an error here, so we aren't forcing anyone to check for it anyway.

omni-viral: It is not an error, but it **should** be checked, and making it part of return type kinda encourages user to do the check.

kvark:
thank you!
bors r+

PR #2599, opened 2019-01-25
[gl] Device type inference enhanced

jackmott:
While researching how to support vendor id's I came across: http://vulkan.gpuinfo.org/
Which lists hardware and whether it is discrete or integrated.  By cross referencing with the opengl database I was able to expand on the logic for detecting integrated gpus.

Also it will set the vendor id based on the vendor string.

Supporting device id is not practical unless you want to pull in a huge database of strings and ids!

The last commit removes unnecessary unsafe blocks which cleans up the warning spam a lot.

PR checklist:


jackmott:
we could potentially use this same logic for dx11 and dx12, could we put it somewhere that all 3 backends could reference it?

jackmott:
DX12/11 only discerns virtual vs discrete, not integrated vs discrete.

PR #2600, opened 2019-01-25
[mtl] secondary command buffers

### LINKED ISSUE #2564: Secondary command buffer support on Metal
Supporting SCBs is required to run Angle on VkPI. cc @null77

kvark:
Fixes #2564 
PR checklist:

This is basic support for SCBs, which allows us to pass those few tests in CTS that don't use events.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -765,6 +765,49 @@ impl Journal {
             }
         }
     }
+
+    fn extend(&mut self, other: &Self, inherit_pass: bool) {
+        if inherit_pass {
+            assert_eq!(other.passes.len(), 1);
+            match *self.passes.last_mut().unwrap() {
+                (soft::Pass::Render(_), ref mut range) => {
+                    range.end += other.render_commands.len();
+                }
+                (soft::Pass::Compute, _) |
+                (soft::Pass::Blit, _) => panic!("Only render passes can inherit"),
+            }
+        } else {
+            for (ref pass, ref range) in &other.passes {
+                let offset = match *pass {
+                    soft::Pass::Render(_) => self.render_commands.len(),
+                    soft::Pass::Compute => self.compute_commands.len(),
+                    soft::Pass::Blit => self.blit_commands.len(),
+                };
+                self.passes.push((
+                    pass.clone(),
+                    range.start + offset .. range.end + offset,
+                ));
+            }
+        }
+
+        // Note: journals contain 3 levels of stuff:
+        // resources, commands, and passes
+        // Each upper level points to the lower one with index
+        // sub-ranges. In order to merge two journals, we need
+        // to fix those indices of the one that goes on top.
+        // This is referred here as "rebasing". 
+        for mut com in other.render_commands.iter().cloned() {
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/command.rs
```diff
@@ -4241,11 +4316,47 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+
+            match *self.inner
+                .borrow_mut()
+                .sink()
+            {
+                CommandSink::Immediate { ref mut cmd_buffer, ref mut encoder_state, ref mut num_passes, .. } => {
+                    if is_inheriting {
+                        let encoder = match encoder_state {
+                            EncoderState::Render(ref encoder) => encoder,
+                            _ => panic!("Expecgted Render encoder!"),
```
grovesNL: `Expected`

src/backend/metal/src/command.rs
```diff
@@ -4241,11 +4316,47 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                    (journal, is_inheriting)
+                }
+                _ => panic!("Unexpected secondary sink!"),
+            };
+
+            match *self.inner
+                .borrow_mut()
+                .sink()
+            {
+                CommandSink::Immediate { ref mut cmd_buffer, ref mut encoder_state, ref mut num_passes, .. } => {
```
grovesNL: If I understand correctly, it looks like `begin` only records SCBs into `Deferred` sinks? If this is the case then when would this branch be used?

src/backend/metal/src/command.rs
```diff
@@ -2210,11 +2258,38 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
             }
             _ => CommandSink::Deferred {
                 is_encoding: false,
+                is_inheriting: info.subpass.is_some(),
                 journal: inner.backup_journal.take().unwrap_or_default(),
             },
         };
         inner.sink = Some(sink);
-        self.state.reset_resources();
```
grovesNL: Does dropping `self.state.reset_resources()` here impact the non-SCB case? Was it not necessary before?

src/backend/metal/src/command.rs
```diff
@@ -765,6 +765,43 @@ impl Journal {
...
+                    soft::Pass::Blit => self.blit_commands.len(),
+                };
+                self.passes.push((
+                    pass.clone(),
+                    range.start + offset .. range.end + offset,
+                ));
+            }
+        }
+
+        for mut com in other.render_commands.iter().cloned() {
```
grovesNL: Could you explain the rebasing part a bit?

src/backend/metal/src/command.rs
kvark: The SCB itself is always a `CommandSink::Deferred`, but here we are matching on the command buffer that executes SCB externally, and it could be anything.

src/backend/metal/src/command.rs
kvark: I dropped it because it's already done in `self.reset()` :)

omni-viral:
:tada:

PR #2603, opened 2019-01-25
Fix default `CompositeAlpha` mode

### LINKED ISSUE #2583: "Inherit" is not always a safe alpha compositing mode in Vulkan
Short info header:
- GFX version: db3393f3073ada8ca1a9e1a581e5528b9d68ce76 (hal)
- OS: Linux (Arch Linux)
- GPU: Nvidia (GTX 970)
- Driver: [`nvidia` package](https://www.archlinux.org/packages/extra/x86_64/nvidia/), version 415.25-8

Running the "quad" or "colour-uniform" example on my machine gives a validation error that looks like this:

```
VALIDATION [VUID-VkSwapchainCreateInfoKHR-compositeAlpha-01280 (0)] : vkCreateSwapChainKHR() called with a non-supported pCreateInfo->compositeAlpha (i.e. VK_COMPOSITE_ALPHA_INHERIT_BIT_KHR).  Supported values are:
    VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR
. The Vulkan spec states: compositeAlpha must be one of the bits present in the supportedCompositeAlpha member of the VkSurfaceCapabilitiesKHR structure returned by vkGetPhysicalDeviceSurfaceCapabilitiesKHR for the surface (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkSwapchainCreateInfoKHR-compositeAlpha-01280)
object info: (type: DEVICE, hndl: 140390740535376)
```

(In fact, it gives that error twice because a resize event causes the swapchain to be recreated immediately.)

It's a simple problem, and we don't even have to talk about Vulkan to understand it -- `CompositeAlpha::Inherit` isn't in the list of composite alphas returned by `surface.compatibility()` on my machine, but the `composite_alpha` field set by both `SwapchainConfig::new` and `SwapchainConfig::from_caps` is always `CompositeAlpha::Inherit`. I can work around the issue by just setting the field myself, but I don't think that's very obvious or convenient.

This will require an API change to fix. `SwapchainConfig::from_caps` will have to be able to see the list of valid composite alphas, which it can't right now because the vecs are all factored out of the `SurfaceCapabilities` struct that gets passed in to it (for low-cost cloning I guess?). `SwapchainConfig::new` should probably get a way to set `composite_alpha`, too.

aleksijuvani:
Fixes #2583 
PR checklist:
- [x] tested examples with the following backends: vk, mtl, dx12, gl

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -1438,7 +1438,7 @@ struct SwapchainState<B: Backend> {
 
 impl<B: Backend> SwapchainState<B> {
     unsafe fn new(backend: &mut BackendState<B>, device: Rc<RefCell<DeviceState<B>>>) -> Self {
-        let (caps, formats, _present_modes, _composite_alphas) = backend
+        let (caps, formats, _present_modes) = backend
             .surface
             .compatibility(&device.borrow().physical_device);
         println!("formats: {:?}", formats);
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -364,7 +364,7 @@ fn main() {
             .expect("Can't wait for fence");
     }
 
-    let (caps, formats, _present_modes, _composite_alphas) =
+    let (caps, formats, _present_modes) =
         surface.compatibility(&mut adapter.physical_device);
     println!("formats: {:?}", formats);
     let format = formats.map_or(f::Format::Rgba8Srgb, |formats| {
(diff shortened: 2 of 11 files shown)


aleksijuvani:
I had the same thought. From a cursory glance, it looks like Vulkan returns an array instead of a bitset from `vkGetPhysicalDeviceSurfacePresentModesKHR`, so we might not be able to (or rather, it might not make sense to do so). I can investigate this further tomorrow.

PR #2605, opened 2019-01-28
[dx12] fix push constants offset

aleksijuvani:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -685,23 +685,6 @@ impl CommandBuffer {
         );
     }
 
-    fn push_constants(
-        user_data: &mut UserData,
-        layout: &r::PipelineLayout,
-        offset: u32,
-        constants: &[u32],
-    ) {
-        let num = constants.len() as u32;
-        for root_constant in &layout.root_constants {
-            assert!(root_constant.range.start <= root_constant.range.end);
-            if root_constant.range.start >= offset && root_constant.range.start < offset + num {
-                let start = (root_constant.range.start - offset) as _;
-                let end = num.min(root_constant.range.end - offset) as _;
-                user_data.set_constants(offset as _, &constants[start..end]);
-            }
-        }
-    }
-
     fn flush_user_data<F, G>(
         pipeline: &mut PipelineCache,
         mut constants_update: F,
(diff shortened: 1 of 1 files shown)


src/backend/dx12/src/command.rs
```diff
@@ -691,13 +691,28 @@ impl CommandBuffer {
...
+                debug_assert!(intersection.start <= intersection.end);
+                debug_assert!(intersection.start >= pc_start);
+                debug_assert!(intersection.start >= root_constant.range.start);
+
+                let pc_start = (intersection.start - pc_start) as usize;
+                let pc_end = pc_start + intersection.len();
+
+                let rc_start = (intersection.start - root_constant.range.start) as usize;
+
+                user_data.set_constants(rc_start, &constants[pc_start..pc_end]);
```
msiglreith: Should be `intersection.start` instead of `rc_start` I think

aleksijuvani:
The Metal backend was taking the offset in bytes, but the Vulkan backend was taking it in multiples of 4 bytes. This was addressed in #2553 and they now both take the offset in bytes. This PR changes the behaviour of the DirectX 12 backend to match the new behaviour of the Metal and Vulkan backends.

I also noticed other problems with this function (hence the excessive diff), such as the `root_constant.range.start >= offset` check, which always evaluated to false and caused the push constants to be never set when the offset was greater than 0 and the root constant range started from 0.

src/backend/dx12/src/command.rs
aleksijuvani: I think you're right. I'll take a look at this tomorrow.

aleksijuvani: I feel like I'm missing something. If we don't need to know the root constant range here at all, there's no reason this entire function can't be just the following? Could it really be that simple?

```rust
fn push_constants(
    user_data: &mut UserData,
    layout: &r::PipelineLayout,
    offset: u32,
    constants: &[u32],
) {
    assert!(offset % 4 == 0);
    user_data.set_constants(offset as usize / 4, &constants);
}
```

msiglreith: Actually, I think that's valid as we delayed the actual pushing of the root constants.

aleksijuvani: I went ahead and changed it.

kvark:
bors retry

PR #2607, opened 2019-01-29
[mtl] fix DPI handling in surface extents

aleksijuvani:
On macOS with a Retina display, the surface `compatibility()` function was returning a non-DPI scaled value for `current_extent` (e.g. 640x360 when we were expecting 1280x720), which is not consistent with the behaviour on Windows, where a DPI-scaled value is returned.

My understanding is that the cause of this is that `NSView` `bounds` returns a value in "points" and you're meant to use `convertRectToBacking` to convert the value to pixels. Someone familiar with this backend, please review.

PR checklist:


kvark:
Ultimately, I think we'll end up not having `NSView` at all. For now, the fix appears to bring consistency. Thank you!
bors r+

PR #2611, opened 2019-02-03
Use enum for vertex input rate

termhn:
Vulkan doesn't support input rates > 1 (instance rate). This PR changes the `VertexBufferDesc` rate to be an enum between `Vertex` and `Instance` to match Vulkan spec. Previously it was allowed to use a number > 1 and it would be properly passed through to the metal/dx12/dx11 backends, but there is no emulation for the Vulkan backend so it would create inconsistent behavior. 

PR checklist:
- [x] tested examples with the following backends: vulkan, dx12, gl

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -21,6 +21,7 @@ use hal::device::{
 };
 use hal::memory::Properties;
 use hal::pool::CommandPoolCreateFlags;
+use hal::pso::VertexInputRate;
 use hal::queue::{QueueFamilyId, Queues};
 use hal::range::RangeArg;
 use hal::{self, buffer, error, format, image, mapping, memory, pass, pso, query, window};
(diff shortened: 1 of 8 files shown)


msiglreith:
There is an extension for user defined divisor numbers. Inconsistency is not really an issue as larger values (without enabling the potential extension) indicates invalid usage.
Nonetheless, I think the change is good. We could add an field to `Instance` to cover different divisors or add a new enum variant.

termhn:
Added optional field to Instance to cover different divisor. It should work for other backends, but not for Vulkan yet. I believe we'll need to first add support for `VK_EXT_vertex_attribute_divisor` to `ash` and then detect and use it when needed...

src/backend/metal/src/device.rs
```diff
@@ -1289,11 +1290,14 @@ impl hal::Device<Backend> for Device {
...
-                } else {
-                    mtl_buffer_desc.set_step_function(MTLVertexStepFunction::PerInstance);
-                    mtl_buffer_desc.set_step_rate(vb.rate as u64);
+                match vb.rate {
+                    VertexInputRate::Vertex => {
+                        mtl_buffer_desc.set_step_function(MTLVertexStepFunction::PerVertex);
+                    }
+                    VertexInputRate::Instance(_rate) => {
+                        mtl_buffer_desc.set_step_function(MTLVertexStepFunction::PerInstance);
+                        mtl_buffer_desc.set_step_rate(vb.rate.as_uint() as u64);
```
kvark: would be cleaner to use the `_rate` here

msiglreith:
@kvark 
How about something like
```rust
enum VertexInputRate {
    Vertex,
    Instance,
    InstanceExt { divisor: usize },
}
```

kvark:
Better, but I I don't think we should have Instance (without args) there at all. Let's specify the divisor and error if it's not supported.

msiglreith:
Well that's fair enough. I'm good with both solutions.

kvark:
@termhn ping! would you mind removing the `Option` from instance rate?

termhn:
@kvark Done and squashed :)

kvark:
thanks!
bors r+

PR #2615, opened 2019-02-06
[WIP] Enable surfaces in gl backend

### LINKED ISSUE #2376: Depth buffer creation is unimplemented in OpenGL backend
Attempts to create a depth buffer with
```
device.create_image(
    image::Kind::D2(extent.width as Size, extent.height as Size, 1, 1),
    1,
    Format::D32FloatS8Uint,
    image::Tiling::Optimal,
    image::Usage::DEPTH_STENCIL_ATTACHMENT,
    image::StorageFlags::empty(),
);
```
fail with an unimplemented panic. I'm not sure how difficult it is to fix this, but depth buffers seem like a reasonably basic and useful feature.

Thomspoon:
Fixes #2376
PR checklist:

The issue is that the backend doesn't support creating a `n::ImageView::Surface`, which can be found via:

```rust
device.create_image(
    image::Kind::D2(extent.width as Size, extent.height as Size, 1, 1),
    1,
    Format::D32FloatS8Uint,
    image::Tiling::Optimal,
    image::Usage::DEPTH_STENCIL_ATTACHMENT,
    image::StorageFlags::empty(),
);
```
The former will devolve to a `n::ImageView::Surface` because the count is <= 1, and the image::Usage flags do not include SAMPLED or STORAGE.

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -765,6 +765,49 @@ impl Journal {
             }
         }
     }
+
+    fn extend(&mut self, other: &Self, inherit_pass: bool) {
+        if inherit_pass {
+            assert_eq!(other.passes.len(), 1);
+            match *self.passes.last_mut().unwrap() {
+                (soft::Pass::Render(_), ref mut range) => {
+                    range.end += other.render_commands.len();
+                }
+                (soft::Pass::Compute, _) |
+                (soft::Pass::Blit, _) => panic!("Only render passes can inherit"),
+            }
+        } else {
+            for (ref pass, ref range) in &other.passes {
+                let offset = match *pass {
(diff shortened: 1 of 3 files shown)


src/backend/gl/src/command.rs
```diff
@@ -984,6 +986,14 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
                             self.push_cmd(Command::BindTexture(*binding, *texture))
                         }
                     }
+                    n::DescSetBindings::Surface(binding, surface) => {
```
kvark: this shouldn't be needed. A render buffer in GL can only be used for rendering into. You can't create a descriptor set with it, because descriptors are resources, and any operations to them require there to be a full-featured texture instead of a render buffer.

PR #2616, opened 2019-02-06
Metal surface fix, extended features and limits for portability

kvark:
Needed for https://github.com/gfx-rs/portability/pull/174
PR checklist:

### DIFF
--- a/src/backend/metal/src/window.rs
+++ b/src/backend/metal/src/window.rs
@@ -12,7 +12,7 @@ use hal::{self, format, image};
 use hal::{Backbuffer, SwapchainConfig, CompositeAlpha};
 
 use core_graphics::base::CGFloat;
-use core_graphics::geometry::{CGPoint, CGRect, CGSize};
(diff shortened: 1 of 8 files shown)


kvark:
Justification for the Metal surface changes is the following: using the `NSView` only to figure out the pixel size is wrong, because the layer itself may have different content scaling factor. This code makes it always ask the layer on MacOS.

src/backend/metal/src/window.rs
```diff
@@ -105,38 +105,31 @@ impl SurfaceInner {
     }
 
     fn dimensions(&self) -> Extent2D {
-        let size = match self.view {
+        let size: CGSize = match self.view {
+            #[cfg(not(target_os = "macos"))]
             Some(view) => unsafe {
```
grovesNL: nit: `Some(view) if cfg!(not(target_os = "macos")) => unsafe {` seems slightly clearer of the intent

kvark:
Thanks for the review!
Bors r=grovesNL

PR #2623, opened 2019-02-10
Add enumerate_adapters_by_gpu_preference

Connicpu:
Adds a method for enumerating the GPUs ordered by preference on Windows 10 v1803+ with the Direct3D 12 backend. Simply returns None on lower versions of the platform.

PR checklist:
- [x] tested examples with the following backends: dx12 (quad, colour-uniform, compute)

### DIFF
--- a/src/backend/dx12/Cargo.toml
+++ b/src/backend/dx12/Cargo.toml
@@ -26,5 +26,5 @@ d3d12 = "0.1"
 log = { version = "0.4" }
 smallvec = "0.6"
 spirv_cross = "0.12.1"
-winapi = { version = "0.3", features = ["basetsd","d3d12","d3d12sdklayers","d3d12shader","d3dcommon","d3dcompiler","dxgi1_2","dxgi1_3","dxgi1_4","dxgiformat","dxgitype","handleapi","minwindef","synchapi","unknwnbase","winbase","windef","winerror","winnt","winuser"] }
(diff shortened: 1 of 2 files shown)


kvark:
Thank you for the contribution!
Would there be any downside from always using the preferred order on a platform that supports it? I don't recall any ordering requirements from Vulkan, and it would make sense to me.

Connicpu:
There shouldn't be any downside! Passing `Unspecified` is the same as using the old iteration method, the only issue would be to make sure people don't rely on adapter 0 always being the high-performance one if they pass that without checking whether it was supported, but always using adapter 0 is pretty common anyways so maybe that point is moot haha

kvark:
Ok, good. Let's just make this behavior default then (when supported, of course).

Connicpu:
Is this about what you were thinking? Or did you mean to not have the separate platform method at all?

kvark:
Which reftests are failing?

Also, can we refactor the code to simplify it by avoiding unnecessary closure functions and helpers? I.e. it could be done this way:
  1. check if the new ability is there
  2. if so, gather adapters into a Vec<>
  3. else, gather adapters into a Vec by the old way
  4. proceed with the code based on the vec of adapters

Connicpu:
When it starts Warding DX12, it immediately panics from a failing assertion at `src\warden\src\gpu.rs:253:25` (`assert_ne!(upload_req.type_mask & (1 << upload_type.0), 0);`)

I did the helper so that I wouldn't have to duplicate that massive function to have both the Trait method and a platform-specific method on Instance, but I can take that out if that's what you'd prefer ^^

kvark:
I don't think we need a platform-specific method at all there, we should just do everything within the regular adapter enumeration function.

Connicpu:
Alright, will do!

kvark:
Perfecto, thank you!
Just one last thing - could you please squash those commits together, since they aren't additive (i.e. the follow-ups are replacing parts that the first commit adds)?

Connicpu:
Absolutely!

PR #2628, opened 2019-02-16
Expose `RawCommandBuffer::reset()` in `CommandBuffer`

aleksijuvani:
Fixes #issue
PR checklist:
- [x] tested examples with the following backends: dx12 vk

### DIFF
--- a/src/hal/src/command/mod.rs
+++ b/src/hal/src/command/mod.rs
@@ -155,9 +155,20 @@ impl<B: Backend, C, S: Shot, L: Level> CommandBuffer<B, C, S, L> {
 
     /// Finish recording commands to the command buffers.
     ///
-    /// The command pool must be reset to able to re-record commands.
+    /// The command buffer must be reset to able to re-record commands.
     pub unsafe fn finish(&mut self) {
         self.raw.finish();
+    }
+
+    /// Empties the command buffer, optionally releasing all resources from the
+    /// commands that have been submitted. The command buffer is moved back to
+    /// the "initial" state.
+    ///
+    /// The command buffer must not be in the "pending" state. Additionally, the
+    /// command pool must have been created with the RESET_INDIVIDUAL flag to be
+    /// able to reset individual buffers.
+    pub unsafe fn reset(&mut self, release_resources: bool) {
+        self.raw.reset(release_resources);
     }
 
     /*


src/hal/src/command/mod.rs
```diff
@@ -155,11 +155,20 @@ impl<B: Backend, C, S: Shot, L: Level> CommandBuffer<B, C, S, L> {
...
     pub unsafe fn finish(&mut self) {
         self.raw.finish();
     }
 
+    /// Empties the command buffer, optionally releasing all resources from the
+    /// commands that have been submitted.
+    ///
+    /// The command pool must have been created with the RESET_INDIVIDUAL flag
+    /// to be able to reset individual buffers.
+    pub unsafe fn reset(&mut self, release_resources: bool) {
```
kvark: I don't think `reset` makes much sense/use for single-shot buffers. Let's move it under the impls for `Multishot` only?

aleksijuvani: Sure, that sounds reasonable. I've pushed some changes. Is this better?

omni-viral: Single shot buffers are not "one time submit and delete". Rather they should be reset before used again.

src/hal/src/command/mod.rs
```diff
@@ -144,6 +144,17 @@ impl<B: Backend, C> CommandBuffer<B, C, MultiShot, Secondary> {
     }
 }
 
+impl<B: Backend, C, L: Level> CommandBuffer<B, C, MultiShot, L> {
+    /// Empties a multi-shot command buffer, optionally releasing resources from
+    /// commands that have been submitted.
+    ///
+    /// The command pool must have been created with the RESET_INDIVIDUAL flag
+    /// to be able to reset individual buffers.
```
omni-viral: Add also that it must not be in pending state.

kvark:
Ah, true. Thanks for correction, @omni-viral!
Looks like the original variant was good.

aleksijuvani:
I've changed this back into its original form. I've also added a comment that the command buffer must not be in the "pending" state when it's reset.

I was also under the impression that `OneShot` buffers can only be submitted once and must then be discarded. We should probably address this in the docs for `OneShot` and `MultiShot` so that this is less likely to be misunderstood.

kvark:
Your impression was right: reset() is effectively a discard.

Thanks!
Bors r+

kvark:
Bors retry

PR #2632, opened 2019-02-20
Metal 0.1.1 with fixes back-ported from master

kvark:
Gets us the secondary command buffers, proper surface layer handling, and a few important iOS fixes.
PR checklist:


kvark:
alright, let's proceed. What's the worst that could happen, right? :)
bors r+

Issue #2608, opened 2019-01-30
Add featured interface for creating Image and ImageView from CoreVideo created Texture


Michael-Lfx:
As written in title I have to support instance of `CVMetalTextureRef *` created by `CVMetalTextureCacheCreateTextureFromImage` because camera and hardware decoder return `CVImageBufferRef`. The CV function is faster than the path that `CVImageBufferRef` -> staging buffer -> image due to no memory copy operations. 
Is it ok to add some featured interfaces within metal backend to address it? cc @kvark

kvark:
Absolutely!

Michael-Lfx:
```c
CV_EXPORT CVReturn CVMetalTextureCacheCreate(
					CFAllocatorRef CV_NULLABLE allocator,
					CFDictionaryRef CV_NULLABLE cacheAttributes,
					id <MTLDevice> CV_NONNULL metalDevice,
					CFDictionaryRef CV_NULLABLE textureAttributes,
					CV_RETURNS_RETAINED_PARAMETER CVMetalTextureCacheRef CV_NULLABLE * CV_NONNULL cacheOut ) API_AVAILABLE(macosx(10.11), ios(8.0), tvos(9.0)) __WATCHOS_PROHIBITED;
```

As the above shows the `CVMetalTextureCacheCreate()` requires `id <MTLDevice>` but both `pub struct PhysicalDevice` and `pub struct Device` make its device field public inside crate. **Should I make it public to the outside caller?** cc @kvark 

Here's the definition.

```rust
struct Shared {
    device: Mutex<metal::Device>,
    queue: Mutex<command::QueueInner>,
    service_pipes: internal::ServicePipes,
    disabilities: PrivateDisabilities,
    private_caps: PrivateCapabilities,
    visibility: VisibilityShared,
}

pub struct PhysicalDevice {
    pub(crate) shared: Arc<Shared>,
    memory_types: Vec<hal::MemoryType>,
}

pub struct Device {
    pub(crate) shared: Arc<Shared>,
    memory_types: Vec<hal::MemoryType>,
    pub online_recording: OnlineRecording,
}
```

kvark:
Exposing `MTLDevice` isn't solving the problem: you'd still need to get `B::Image` on the output. I think we just need a `pub fn create_image_from_cv_buffer(..) -> Image` function added to `impl Device` in the metal backend.

Michael-Lfx:
[...]

PR #2635, opened 2019-02-23
[mtl] add interface for creating ImageView from CoreVideo

Michael-Lfx:
[mtl] add interface for creating ImageView from CVMetalTextureCacheCreateTextureFromImage created CVMetalTextureRef object

Fixes https://github.com/gfx-rs/gfx/issues/2608  cc @kvark 
PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -703,6 +703,48 @@ impl Device {
 
         arg
     }
+
+    pub unsafe fn create_image_view_with_metal_texture(
+        &self,
+        texture: metal::Texture,
+        kind: image::ViewKind,
+        format: format::Format,
+        swizzle: format::Swizzle,
+        range: image::SubresourceRange,
+    ) -> Result<n::ImageView, image::ViewError> {
+        let mtl_format = match self
+            .shared
+            .private_caps
+            .map_format_with_swizzle(format, swizzle)
+            {
+                Some(f) => f,
+                None => {
+                    error!("failed to swizzle format {:?} with {:?}", format, swizzle);
+                    return Err(image::ViewError::BadFormat(format));
+                }
+            };
+        let like = native::ImageLike::Texture(texture);
+        let raw = like.as_texture();
+        let mtl_type = conv::map_texture_type(kind);
+
+         let view = raw.new_texture_view_from_slice(
+           mtl_format,
+           mtl_type,
+           NSRange {
+               location: range.levels.start as _,
+               length: (range.levels.end - range.levels.start) as _,
+           },
+           NSRange {
+               location: range.layers.start as _,
+               length: (range.layers.end - range.layers.start) as _,
+           },
+        );
+
+         Ok(n::ImageView {
+            raw: view,
+            mtl_format,
+        })
+    }
 }
 
 impl hal::Device<Backend> for Device {


src/backend/metal/src/device.rs
```diff
@@ -703,6 +703,49 @@ impl Device {
...
+        kind: image::ViewKind,
+        format: format::Format,
+        swizzle: format::Swizzle,
+        range: image::SubresourceRange,
+    ) -> Result<n::ImageView, image::ViewError> {
+        let mtl_format = match self
+            .shared
+            .private_caps
+            .map_format_with_swizzle(format, swizzle)
+            {
```
kvark: indentation is off here

src/backend/metal/src/device.rs
```diff
@@ -703,6 +703,49 @@ impl Device {
...
+
+    #[cfg(feature = "core_video")]
+    pub unsafe fn create_image_view_with_metal_texture(
+        &self,
+        texture: metal::Texture,
+        kind: image::ViewKind,
+        format: format::Format,
+        swizzle: format::Swizzle,
+        range: image::SubresourceRange,
+    ) -> Result<n::ImageView, image::ViewError> {
```
kvark: is this mostly the same code as the device has for regular view creation? can one use another internally?

src/backend/metal/src/device.rs
```diff
@@ -703,6 +703,49 @@ impl Device {
 
         arg
     }
+
+    #[cfg(feature = "core_video")]
```
kvark: since it doesn't bring any new dependencies, we can expose this unconditionally (with no features)

src/backend/metal/src/device.rs
Michael-Lfx: It's formatted by rustfmt...

src/backend/metal/src/device.rs
Michael-Lfx: The texture returned by `CVMetalTextureGetTexture()` isn't a framebuffer-only texture and I think it's alright to call '.to_owned()'. I should give it a try.

src/backend/metal/src/device.rs
Michael-Lfx: Ok.

src/backend/metal/src/device.rs
kvark: oh really? that's an unfortunate way to format :(
perhaps, I need to play with our config more

kvark:
Is this ready for the review?

Michael-Lfx:
Not yet...

Michael-Lfx:
Should we change it to create `Image` from `CVMetalTextureCacheCreateTextureFromImage()` then use the same routine to create `ImageView`?

kvark:
@Michael-Lfx is there a use of this image if it's created from `CVMetalTextureCacheCreateTextureFromImage` other than for creating a view?

Michael-Lfx:
We will use it as input texture and color output attachment.

kvark:
So in both cases you just need a view, right? Sounds like you don't need the `Image` object constructed then and can provide the API just to create a view.

Michael-Lfx:
Yes. I think an `ImageView` is good enough for both `ShaderProgram` and `Framebuffer` module.

kvark:
@Michael-Lfx are you planning on finishing this work?

kvark:
We'll likely need that down the road, but this PR has not been finish, so it needs to be closed.

PR #2640, opened 2019-02-23
Improve frame synchronization in quad example

termhn:
Fixes TODOs in quad example and demonstrates a more proper method of frame synchronization

PR checklist:

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -298,10 +298,8 @@ fn main() {
         ]);
     }
 
-    let mut frame_semaphore = device.create_semaphore().expect("Can't create semaphore");
-    let mut frame_fence = device.create_fence(false).expect("Can't create fence"); // TODO: remove
-
     // copy buffer to texture
+    let mut copy_fence = device.create_fence(false).expect("Could not create fence");
     unsafe {
         let mut cmd_buffer = command_pool.acquire_command_buffer::<command::OneShot>();
         cmd_buffer.begin();
@@ -357,11 +355,15 @@ fn main() {
 
         cmd_buffer.finish();
 
-        queue_group.queues[0].submit_nosemaphores(Some(&cmd_buffer), Some(&mut frame_fence));
+        queue_group.queues[0].submit_nosemaphores(Some(&cmd_buffer), Some(&mut copy_fence));
 
         device
-            .wait_for_fence(&frame_fence, !0)
+            .wait_for_fence(&copy_fence, !0)
             .expect("Can't wait for fence");
+    }
+
+    unsafe {
+        device.destroy_fence(copy_fence);
     }
 
     let (caps, formats, _present_modes) = surface.compatibility(&mut adapter.physical_device);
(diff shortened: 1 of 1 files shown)


examples/quad/main.rs
```diff
@@ -442,6 +444,47 @@ fn main() {
         Backbuffer::Framebuffer(fbo) => (Vec::new(), vec![fbo]),
     };
 
+    let frames_in_flight = frame_images.len();
```
msiglreith: It would be nicer to not strongly couple it with the swapchain size and make the decision on which element of the ringbuffers to use independent of the acquired frame id.

examples/quad/main.rs
```diff
@@ -698,18 +762,21 @@ fn main() {
...
             let submission = Submission {
-                command_buffers: Some(&cmd_buffer),
-                wait_semaphores: Some((&frame_semaphore, PipelineStage::BOTTOM_OF_PIPE)),
-                signal_semaphores: &[],
+                command_buffers: Some(&*cmd_buffer),
+                wait_semaphores: Some((
+                    &image_acquire_semaphores[frame],
+                    PipelineStage::COLOR_ATTACHMENT_OUTPUT,
+                )),
+                signal_semaphores: Some(&submission_complete_semaphores[frame]),
```
kvark: I don't think this signal is really needed, given that we are presenting on the same queue

msiglreith: It's needed. Presentation engine is abstracted by vulkan but requires synchronization with the presenting queue.

examples/quad/main.rs
```diff
@@ -442,6 +444,47 @@ fn main() {
...
 
+    let frames_in_flight = frame_images.len();
+
+    let mut free_acquire_semaphore = device
+        .create_semaphore()
+        .expect("Could not create semaphore");
+    let mut image_acquire_semaphores = Vec::with_capacity(frames_in_flight);
+    let mut submission_complete_semaphores = Vec::with_capacity(frames_in_flight);
+    let mut submission_complete_fences = Vec::with_capacity(frames_in_flight);
+    let mut cmd_pools = Vec::with_capacity(frames_in_flight);
```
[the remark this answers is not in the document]
msiglreith: There seem to be some voices who want a more non-'toyish' (?) example.
This might be the right direction ..

examples/quad/main.rs
msiglreith: As there are concerns regarding increased complexity. I think it would only require minor changes, but I'm fine with keeping it as it as well!

examples/quad/main.rs
kvark: uh oh, thanks for correction!

msiglreith: For reference beside the specs: https://github.com/KhronosGroup/Vulkan-Docs/wiki/Synchronization-Examples#combined-graphicspresent-queue

kvark:
shipit :)
bors r+

PR #2644, opened 2019-02-25
[hal] more succinctly expressed by calling .find(p) instead.

Michael-Lfx:
Fixes #issue
PR checklist:

### DIFF
--- a/src/hal/src/adapter.rs
+++ b/src/hal/src/adapter.rs
@@ -187,12 +187,11 @@ impl<B: Backend> Adapter<B> {
         let requested_family = self
             .queue_families
             .iter()
-            .filter(|family| {
+            .find(|family| {
                 C::supported_by(family.queue_type())
-                    && selector(&family)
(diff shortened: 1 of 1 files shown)


src/hal/src/adapter.rs
```diff
@@ -187,12 +187,11 @@ impl<B: Backend> Adapter<B> {
         let requested_family = self
             .queue_families
             .iter()
-            .filter(|family| {
+            .find(|family| {
                 C::supported_by(family.queue_type())
                     && selector(&family)
```
kvark: nit: I don't think we need `&` before `familty` here now

Michael-Lfx:
done.  @kvark

kvark:
@Michael-Lfx is it? I don't see it

Michael-Lfx:
> @Michael-Lfx is it? I don't see it

Sorry. I forget to force push it.

kvark:
Bors r+

PR #2657, opened 2019-02-27
Implement Drop for Device

chtotut:
Device `context` field (`ID3D11DeviceContext`) have to call `.Release()` in `Drop` impl.


kvark:
bors retry

PR #2663, opened 2019-03-02
Implementing missing gl backend functions

Thomspoon:
WIP for fixing some of the shortcomings of the GL backend. 
Currently the fixes would allow Lokathor's gfx-hal tutorials to be ran on OpenGL as well.

Fixes: 
- [x] Push constant translation to uniforms
- [x] Depth/Stencil

PR checklist:
- [x] tested examples with the following backends: vulkan, gl

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -70,8 +70,8 @@ pub enum Command {
     BindIndexBuffer(gl::types::GLuint),
     //BindVertexBuffers(BufferSlice),
     BindUniform {
-        uniform: n::UniformDesc, 
-        buffer: BufferSlice
+        uniform: n::UniformDesc,
+        buffer: BufferSlice,
     },
     SetViewports {
         first_viewport: u32,
@@ -473,7 +473,7 @@ impl RawCommandBuffer {
 
                             return Some(cmd);
                         }
-                    } 
+                    }
 
                     // Clear depth-stencil target
                     let depth = if view_format.is_depth() {
@@ -1297,7 +1297,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         }
 
         let uniform = if offset == 0 {
-            // If offset is zero, we can just return the first item 
+            // If offset is zero, we can just return the first item
             // in our uniform list
             uniforms.get(0).unwrap()
         } else {
@@ -1318,12 +1318,10 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             }
 
             needle.unwrap()
-        }.clone();
+        }
+        .clone();
 
-        self.push_cmd(Command::BindUniform {
-            uniform,
-            buffer,
-        });
+        self.push_cmd(Command::BindUniform { uniform, buffer });
     }
 
     unsafe fn push_compute_constants(
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -69,6 +69,10 @@ pub enum Command {
     },
     BindIndexBuffer(gl::types::GLuint),
     //BindVertexBuffers(BufferSlice),
+    BindUniform {
+        uniform: n::UniformDesc, 
+        buffer: BufferSlice
+    },
     SetViewports {
         first_viewport: u32,
         viewport_ptr: BufferSlice,
@@ -171,6 +175,8 @@ struct Cache {
     vertex_buffer_descs: Vec<Option<pso::VertexBufferDesc>>,
     // Active attributes.
     attributes: Vec<n::AttributeDesc>,
+    // Active uniforms
+    uniforms: Vec<n::UniformDesc>,
 }
 
 impl Cache {
@@ -188,6 +194,7 @@ impl Cache {
             vertex_buffers: Vec::new(),
             vertex_buffer_descs: Vec::new(),
             attributes: Vec::new(),
+            uniforms: Vec::new(),
         }
     }
 }
@@ -466,23 +473,23 @@ impl RawCommandBuffer {
 
                             return Some(cmd);
                         }
+                    } 
+
+                    // Clear depth-stencil target
+                    let depth = if view_format.is_depth() {
+                        clear.value.map(|cv| unsafe { cv.depth_stencil.depth })
                     } else {
-                        // Clear depth-stencil target
-                        let depth = if view_format.is_depth() {
-                            clear.value.map(|cv| unsafe { cv.depth_stencil.depth })
-                        } else {
-                            None
-                        };
+                        None
+                    };
 
-                        let stencil = if view_format.is_stencil() {
-                            clear.stencil_value
-                        } else {
-                            None
-                        };
+                    let stencil = if view_format.is_stencil() {
+                        clear.stencil_value
+                    } else {
+                        None
+                    };
 
-                        if depth.is_some() || stencil.is_some() {
-                            return Some(Command::ClearBufferDepthStencil(depth, stencil));
-                        }
+                    if depth.is_some() || stencil.is_some() {
+                        return Some(Command::ClearBufferDepthStencil(depth, stencil));
                     }
 
                     None
@@ -911,6 +918,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             ref blend_targets,
             ref attributes,
             ref vertex_buffers,
+            ref uniforms,
         } = *pipeline;
 
         if self.cache.primitive != Some(primitive) {
@@ -932,6 +940,8 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         self.cache.attributes = attributes.clone();
 
         self.cache.vertex_buffer_descs = vertex_buffers.clone();
+
+        self.cache.uniforms = uniforms.clone();
 
         self.update_blend_targets(blend_targets);
     }
@@ -1276,10 +1286,44 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         &mut self,
         _layout: &n::PipelineLayout,
         _stages: pso::ShaderStageFlags,
-        _offset: u32,
-        _constants: &[u32],
+        offset: u32,
+        constants: &[u32],
     ) {
-        unimplemented!()
+        let buffer = self.add(constants);
+
+        let uniforms = &self.cache.uniforms;
+        if uniforms.len() == 0 {
+            panic!("No uniforms bound when trying to bind push constant!");
+        }
+
+        let uniform = if offset == 0 {
+            // If offset is zero, we can just return the first item 
+            // in our uniform list
+            uniforms.get(0).unwrap()
+        } else {
+            // If offset is non-zero, we need to iterate through our uniform list
+            // and find what uniform is expected
+            let mut needle = None;
+            let mut acc: u32 = 0;
+            for uniform in uniforms {
+                acc += uniform.size as u32;
+                if acc == offset {
+                    needle = Some(uniform);
+                    break;
+                }
+            }
+
+            if needle.is_none() {
+                panic!("No uniform found at offset: {}", offset);
+            }
+
+            needle.unwrap()
+        }.clone();
+
+        self.push_cmd(Command::BindUniform {
+            uniform,
+            buffer,
+        });
     }
 
     unsafe fn push_compute_constants(
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -73,6 +73,13 @@ pub enum Command {
         uniform: n::UniformDesc,
         buffer: BufferSlice,
     },
+    BindRasterizer {
+        rasterizer: pso::Rasterizer,
+        is_embedded: bool,
+    },
+    BindDepth {
+        depth: pso::DepthTest,
+    },
     SetViewports {
         first_viewport: u32,
         viewport_ptr: BufferSlice,
@@ -919,6 +926,8 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             ref attributes,
             ref vertex_buffers,
             ref uniforms,
+            rasterizer,
+            depth,
         } = *pipeline;
 
         if self.cache.primitive != Some(primitive) {
@@ -944,6 +953,14 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         self.cache.uniforms = uniforms.clone();
 
         self.update_blend_targets(blend_targets);
+
+        self.push_cmd(Command::BindRasterizer { 
+            rasterizer, 
+            is_embedded: false 
+        });
+        self.push_cmd(Command::BindDepth { 
+            depth,
+        });
     }
 
     unsafe fn bind_graphics_descriptor_sets<I, J>(
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -75,7 +75,6 @@ pub enum Command {
     },
     BindRasterizer {
         rasterizer: pso::Rasterizer,
-        is_embedded: bool,
     },
     BindDepth {
         depth: pso::DepthTest,
@@ -956,7 +955,6 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
 
         self.push_cmd(Command::BindRasterizer { 
             rasterizer, 
-            is_embedded: false 
         });
         self.push_cmd(Command::BindDepth { 
             depth,
@@ -1318,23 +1316,10 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             // in our uniform list
             uniforms.get(0).unwrap()
         } else {
-            // If offset is non-zero, we need to iterate through our uniform list
-            // and find what uniform is expected
-            let mut needle = None;
-            let mut acc: u32 = 0;
-            for uniform in uniforms {
-                acc += uniform.size as u32;
-                if acc == offset {
-                    needle = Some(uniform);
-                    break;
-                }
+            match uniforms.binary_search_by(|uniform| uniform.offset.cmp(&offset as _)) {
+                Ok(index) => uniforms.get(index).unwrap(),
+                Err(_) => panic!("No uniform found at offset: {}", offset),
             }
-
-            if needle.is_none() {
-                panic!("No uniform found at offset: {}", offset);
-            }
-
-            needle.unwrap()
         }
         .clone();
 
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -1292,8 +1292,8 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         let buffer = self.add(constants);
 
         let uniforms = &self.cache.uniforms;
-        if uniforms.len() == 0 {
-            panic!("No uniforms bound when trying to bind push constant!");
+        if uniforms.is_empty() {
+            unimplemented!()
         }
 
         let uniform = if offset == 0 {
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -254,6 +254,27 @@ impl Device {
         Ok(())
     }
 
+    fn set_push_const_layout(
+        &self,
+        ast: &mut spirv::Ast<glsl::Target>,
+    ) -> Result<(), d::ShaderError>  {
+        let resources = ast.get_shader_resources()
+            .map_err(gen_unexpected_error)?;
+        
+        match self.share.info.shading_language.tuple() {
+            // Explicit uniform locations are only valid on OpenGL 4.3+
+            (major, minor) if major >= 4 && minor >= 3 => {
+                for (index, pushconst) in resources.push_constant_buffers.iter().enumerate() {
+                    let _ = ast.set_decoration(pushconst.id, spirv::Decoration::Location, index as _)
+                        .map_err(gen_unexpected_error);
+                }
+            },
+            _ => () 
+        };
+
+        Ok(())
+    }
+
     fn translate_spirv(
         &self,
         ast: &mut spirv::Ast<glsl::Target>,
@@ -436,9 +457,10 @@ impl Device {
                     desc_remap_data,
                     name_binding_map,
                 );
+                self.set_push_const_layout(&mut ast).unwrap();
 
                 let glsl = self.translate_spirv(&mut ast).unwrap();
-                info!("Generated:\n{:?}", glsl);
+                debug!("SPIRV-Cross generated shader:\n{}", glsl);
                 let shader = match self
                     .create_shader_module_from_source(glsl.as_bytes(), stage)
                     .unwrap()
@@ -579,14 +601,20 @@ impl d::Device<B> for Device {
         let subpasses = subpasses
             .into_iter()
             .map(|subpass| {
+                let subpass = subpass.borrow();
                 let color_attachments = subpass
-                    .borrow()
                     .colors
                     .iter()
                     .map(|&(index, _)| index)
                     .collect();
+                
+                let depth_stencil = if subpass.depth_stencil.is_some() {
+                    Some(subpass.depth_stencil.unwrap().0)
+                } else {
+                    None
+                };
 
-                n::SubpassDesc { color_attachments }
+                n::SubpassDesc { color_attachments, depth_stencil }
             })
             .collect();
 
@@ -788,6 +816,36 @@ impl d::Device<B> for Device {
             vertex_buffers[vb.binding as usize] = Some(*vb);
         }
 
+        let mut uniforms = Vec::new();
+        {
+            let gl = &self.share.context;
+            let mut count = 0;
+            let mut uniform_max_size = 0;
+
+            gl.GetProgramiv(program, gl::ACTIVE_UNIFORMS, &mut count);
+            gl.GetProgramiv(program, gl::ACTIVE_UNIFORM_MAX_LENGTH, &mut uniform_max_size);
+            for uniform in 0..count {
+                let mut length = 0;
+                let mut size = 0;
+                let mut utype = 0;
+                let mut name = Vec::with_capacity(uniform_max_size as usize);
+                name.set_len(uniform_max_size as usize - 1); 
+                gl.GetActiveUniform(program, uniform as _, uniform_max_size - 1 as i32, &mut length, &mut size, &mut utype, name.as_mut_ptr() as *mut _);
+                
+                let location = gl.GetUniformLocation(program, name.as_ptr() as _);
+                
+                // Sampler2D won't show up in UniformLocation and the only other uniforms
+                // should be push constants
+                if location >= 0 {
+                    uniforms.push(n::UniformDesc {
+                        location: location as _,
+                        size,
+                        utype,
+                    });
+                }                
+            }
+        }
+
         Ok(n::GraphicsPipeline {
             program,
             primitive: conv::primitive_to_gl_primitive(desc.input_assembler.primitive),
@@ -810,6 +868,7 @@ impl d::Device<B> for Device {
                     }
                 })
                 .collect(),
+            uniforms,
         })
     }
 
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -603,11 +603,7 @@ impl d::Device<B> for Device {
                 let subpass = subpass.borrow();
                 let color_attachments = subpass.colors.iter().map(|&(index, _)| index).collect();
 
-                let depth_stencil = if subpass.depth_stencil.is_some() {
-                    Some(subpass.depth_stencil.unwrap().0)
-                } else {
-                    None
-                };
+                let depth_stencil = subpass.depth_stencil.map(|ds| ds.0);
 
                 n::SubpassDesc {
                     color_attachments,
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -101,7 +101,7 @@ fn create_fbo_internal(share: &Starc<Share>) -> Option<gl::types::GLuint> {
 /// GL device.
 #[derive(Debug)]
 pub struct Device {
-    share: Starc<Share>,
+    pub (crate) share: Starc<Share>,
 }
 
 impl Drop for Device {
@@ -846,6 +846,9 @@ impl d::Device<B> for Device {
             }
         }
 
+        state::bind_rasterizer(gl, &desc.rasterizer, false);
+        state::bind_depth(gl, &desc.depth_stencil.depth);
+
         Ok(n::GraphicsPipeline {
             program,
             primitive: conv::primitive_to_gl_primitive(desc.input_assembler.primitive),
@@ -1212,6 +1215,7 @@ impl d::Device<B> for Device {
         let (int_format, iformat, itype) = match format {
             Format::Rgba8Unorm => (gl::RGBA8, gl::RGBA, gl::UNSIGNED_BYTE),
             Format::Rgba8Srgb => (gl::SRGB8_ALPHA8, gl::RGBA, gl::UNSIGNED_BYTE),
+            Format::D32Sfloat => (gl::DEPTH32F_STENCIL8, gl::DEPTH_STENCIL, gl::FLOAT_32_UNSIGNED_INT_24_8_REV),
             _ => unimplemented!(),
         };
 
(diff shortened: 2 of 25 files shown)


src/backend/gl/src/command.rs
```diff
@@ -67,8 +67,14 @@ pub enum Command {
...
         instances: Range<hal::InstanceCount>,
     },
+    DrawIndirect {
+        primitive: gl::types::GLenum,
+        offset: buffer::Offset,
+        draw_count: hal::DrawCount,
+    },
     BindIndexBuffer(gl::types::GLuint),
     //BindVertexBuffers(BufferSlice),
+    BindUniform(gl::types::GLuint, BufferSlice),
```
msiglreith: We didn't do it so far but we should try to reflect the underyling gl function call name in these cmd names as it easily becomes quite confusing once the gl implementation becomes more sophisticated. Also difficult to reason about the required version.

src/backend/gl/src/command.rs
```diff
@@ -1274,10 +1294,14 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
         _layout: &n::PipelineLayout,
         _stages: pso::ShaderStageFlags,
-        _offset: u32,
-        _constants: &[u32],
+        offset: u32,
+        constants: &[u32],
     ) {
-        unimplemented!()
+        let constants = self.add(constants);
+        self.push_cmd(Command::BindUniform(
```
msiglreith: Is this valid? How does spirv-cross translate push constants to GLSL? I would assume a uniform struct, so I'm wondering a bit if filling it with a raw u32 array is allowed

src/backend/gl/src/queue.rs
```diff
@@ -386,6 +386,22 @@ impl CommandQueue {
...
             }
+            com::Command::DrawIndirect {
+                primitive,
+                offset,
+                draw_count
+            } => {
+                let gl = &self.share.context;
+
+                unsafe {
+                    gl.DrawArraysInstanced(
```
msiglreith: That's not an indirect draw call. Also the `draw_count` parameter seems to be misused and an instance count of 0 appears a bit suspicious to me.

src/backend/gl/src/command.rs
Thomspoon: spirv-cross translates push constants to a uniform struct, so my idea is that with 32-bit words push constants in this fashion would be similar to a memcpy. Is there a more elegant way to do this @msiglreith?

src/backend/gl/src/queue.rs
Thomspoon: @msiglreith Looking at this documentation https://www.khronos.org/opengl/wiki/Vertex_Rendering#Indirect_rendering it sounded like DrawIndirect would call glDrawArraysInstancedBaseInstance underneath, which I thought was the same as glDrawArraysInstanced with 0 as the instance.

termhn: Indirect drawing gets its information from a GPU buffer though; specifically, the one bound to the  `GL_DRAW_INDIRECT_BUFFER` binding, which `DrawArraysInstanced` does not. `draw_count` is only a parameter for `glMultiDrawArraysIndirect` and it is telling it how many successive indirect draw call information structures to look for after `offset` into the `GL_DRAW_INDIRECT_BUFFER`. The `count` parameter in `DrawArraysInstanced` that you are passing it into is telling the call how may *indices* to render in the currently bound vertex buffers, not how many draw calls to make.

src/backend/gl/src/command.rs
msiglreith: Could you print an example of the spirv-cross output?
I would assume we would need to do some sort of reflection here at least to correctly fill all the fields

Thomspoon:
```rust
#version 450

struct PushConsts
{
    mat4 view_proj;
};

uniform PushConsts push;

layout(location = 2) in vec4 model_col1;
layout(location = 3) in vec4 model_col2;
layout(location = 4) in vec4 model_col3;
layout(location = 5) in vec4 model_col4;
layout(location = 0) in vec3 position;
layout(location = 1) out vec2 frag_uv;
layout(location = 1) in vec2 vert_uv;

void main()
{
    mat4 model = mat4(vec4(model_col1), vec4(model_col2), vec4(model_col3), vec4(model_col4));
    gl_Position = (push.view_proj * model) * vec4(position, 1.0);
    frag_uv = vert_uv;
}
```

msiglreith: IIRC gl assignes different locations for fields of uniforms so each field needs to be manually assigned -> reflection required

src/backend/gl/src/queue.rs
Thomspoon: I removed this code because I'm not actually doing any indirect draws. I was also doing it terribly wrong :P

src/backend/gl/src/command.rs
Thomspoon: Hey @msiglreith check out my new push constant to uniform translations.

src/backend/gl/src/queue.rs
```diff
@@ -488,16 +491,15 @@ impl CommandQueue {
...
+                    (Some(depth), Some(stencil)) => (depth, stencil),
+                    (Some(depth), None) => (depth, 0),
+                    (None, Some(stencil)) => (0.0, stencil),
                     _ => unreachable!(),
                 };
-
                 self.share
                     .context
-                    .ClearBufferfi(target, 0, depth, stencil as _);
+                    .ClearBufferfi(gl::DEPTH_STENCIL, 0, depth, stencil as _);
```
msiglreith: why is this hardcoded?

src/backend/gl/src/queue.rs
```diff
@@ -835,7 +895,26 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
         S: 'a + Borrow<native::Semaphore>,
         Iw: IntoIterator<Item = &'a S>,
     {
+        let gl = &self.share.context;
+
         for swapchain in swapchains {
+            let extent = swapchain.0.borrow().extent;
+
+            gl.BlitNamedFramebuffer(
+                self.state.fbo,              // Default framebuffer
```
msiglreith: this shouldn't be the current fbo but rather a swapchain owned fbo, which has the backbuffer image attached.

src/hal/src/window.rs
```diff
@@ -386,15 +386,6 @@ impl SwapchainConfig {
     // TODO: depth-only, stencil-only, swapchain size, present modes, etc.
 }
 
-/// Swapchain backbuffer type
-#[derive(Debug)]
-pub enum Backbuffer<B: Backend> {
```
msiglreith: Nice! 👍

src/backend/gl/src/queue.rs
Thomspoon: Khronos documentation:
> clearBufferfi may be used to clear the depth and stencil buffers. buffer must be DEPTH_STENCIL and drawBuffer must be zero. depth and stencil are the depth and stencil values, respectively.

src/backend/gl/src/queue.rs
Thomspoon: When we do a renderpass we bind a framebuffer, and we want to draw to that framebuffer, right?

src/backend/gl/src/device.rs
```diff
@@ -788,6 +816,39 @@ impl d::Device<B> for Device {
...
+                    uniforms.push(n::UniformDesc {
+                        location: location as _,
+                        size,
+                        utype,
+                    });
+                }                
+            }
+        }
+
+        state::bind_rasterizer(gl, &desc.rasterizer, false);
```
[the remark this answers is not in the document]
Thomspoon: These exist to generate the glEnable's that are required. Should we move them to bind_graphics_pipeline instead? It does make sense not to have them in the create.

src/backend/gl/src/native.rs
```diff
@@ -265,12 +266,14 @@ pub struct RenderPass {
...
     pub(crate) color_attachments: Vec<usize>,
+    pub(crate) depth_stencil: Option<usize>,
 }
 
 impl SubpassDesc {
     /// Check if an attachment is used by this sub-pass.
     pub(crate) fn is_using(&self, at_id: pass::AttachmentId) -> bool {
-        self.color_attachments.iter().any(|id| *id == at_id)
+        self.color_attachments.iter().any(|id| *id == at_id) ||
+        (self.depth_stencil.is_some() && self.depth_stencil.unwrap() == at_id)
```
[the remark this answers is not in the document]
Thomspoon: What would that match look like?

src/backend/gl/src/queue.rs
ZeGentzy: @Thomspoon 
>  **glClearBufferfi and glClearNamedFramebufferfi are used to clear the depth and stencil buffers simultaneously.** buffer must be GL_DEPTH_STENCIL and drawbuffer must be zero. depth and stencil are the values to clear the depth and stencil buffers to, respectively. Clamping and type conversion of depth for fixed-point depth buffers are performed in the same fashion as for glClearDepth. Masking of stencil for stencil buffers is performed in the same fashion as for glClearStencil. These commands are equivalent to clearing the depth and stencil buffers separately, but may be faster when a buffer of internal format GL_DEPTH_STENCIL is being cleared. The same per-fragment and masking operations defined for glClear are applied. 

-- https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glClearBuffer.xhtml

When `depth` or `stencil` is `None`,  `glClearStencil` or `glClearDepth` should be used, respectively.

src/backend/gl/src/queue.rs
```diff
@@ -835,7 +987,26 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
         S: 'a + Borrow<native::Semaphore>,
         Iw: IntoIterator<Item = &'a S>,
     {
+        let gl = &self.share.context;
+
         for swapchain in swapchains {
+            let extent = swapchain.0.borrow().extent;
+
+            gl.BlitNamedFramebuffer(
```
ZeGentzy: This is only introduced in OGL 4.5. As a fallback, with OGL pre-4.5, bind the read and write frame buffers separately and use `glBlitFramebuffer`.

Thomspoon: Ah darn. I liked this function more :)

src/backend/gl/src/native.rs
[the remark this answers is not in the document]
ZeGentzy:
```
match self.depth_stencil {
Some(at_id) => true,
_ => self.color_attachments.iter().any(|id| *id == at_id),
}
```
?

src/backend/gl/src/device.rs
```diff
@@ -788,6 +810,48 @@ impl d::Device<B> for Device {
...
+                program,
+                gl::ACTIVE_UNIFORM_MAX_LENGTH,
+                &mut uniform_max_size,
+            );
+            for uniform in 0..count {
+                let mut length = 0;
+                let mut size = 0;
+                let mut utype = 0;
+                let mut name = Vec::with_capacity(uniform_max_size as usize);
+                name.set_len(uniform_max_size as usize - 1);
```
ZeGentzy: Why a minus one?
> GL_ACTIVE_UNIFORM_MAX_LENGTH 
> params returns the length of the longest active uniform variable name for program, including the null termination character (i.e., the size of the character buffer required to store the longest uniform variable name). If no active uniform variables exist, 0 is returned.

ZeGentzy: `set_len` is unsafe, and we are not meeting one of it's in variants:
> The elements at old_len..new_len must be initialized.

 - https://doc.rust-lang.org/std/vec/struct.Vec.html

`glGetActiveUniform` might not init the whole range. This could cause UB in `Vec`'s `Drop` impl.

src/backend/gl/src/device.rs
```diff
@@ -788,6 +810,48 @@ impl d::Device<B> for Device {
...
+                    &mut utype,
+                    name.as_mut_ptr() as *mut _,
+                );
+
+                let location = gl.GetUniformLocation(program, name.as_ptr() as _);
+
+                // Sampler2D won't show up in UniformLocation and the only other uniforms
+                // should be push constants
+                if location >= 0 {
+                    uniforms.push(n::UniformDesc {
```
ZeGentzy: Why is this code even a thing? Surely we can attain this info via SPIRV reflection?

src/backend/gl/src/device.rs
```diff
@@ -254,6 +256,27 @@ impl Device {
...
     }
 
+    fn set_push_const_layout(
+        &self,
+        ast: &mut spirv::Ast<glsl::Target>,
+    ) -> Result<(), d::ShaderError> {
+        let resources = ast.get_shader_resources().map_err(gen_unexpected_error)?;
+
+        match self.share.info.shading_language.tuple() {
+            // Explicit uniform locations are only valid on OpenGL 4.3+
```
ZeGentzy: What are we doing pre-4.3?

src/backend/gl/src/device.rs
```diff
@@ -254,6 +256,27 @@ impl Device {
...
+            (major, minor) if major >= 4 && minor >= 3 => {
+                for (index, pushconst) in resources.push_constant_buffers.iter().enumerate() {
+                    let _ = ast
```
ZeGentzy: Afaik, push_constants need to be remaped, like everything else.

https://github.com/gfx-rs/gfx/blob/58245e48a19758c5ac05f301c9cfb14cf3a138af/src/backend/gl/src/device.rs#L336

I can't imagine how this code might possibly work. This is the only place in the code which mentions push constants.

To my knowledge, multiple things would need to happen to support push constants:
 - Use spirv reflection to change PSBs to just another UBO.
 - Remap it like everything else.
 - Copy what we get from `vkCmdPushConstants` into the UBO.
 - Maybe a couple more steps that I've missed.
 - Profit!

src/backend/gl/src/command.rs
```diff
@@ -1274,10 +1301,42 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
+            // in our uniform list
+            uniforms.get(0).unwrap()
+        } else {
+            // If offset is non-zero, we need to iterate through our uniform list
+            // and find what uniform is expected
+            let mut needle = None;
+            let mut acc: u32 = 0;
+            for uniform in uniforms {
+                acc += uniform.size as u32;
+                if acc == offset {
```
ZeGentzy: If we are only going to check for the uniforms' offsets, not their size, then we should only store the uniform offsets, not sizes. Then we can just do a `binary_search`.

src/backend/gl/src/queue.rs
```diff
@@ -491,15 +491,24 @@ impl CommandQueue {
...
+                match (depth, stencil) {
+                    (Some(depth), Some(stencil)) => {
+                        self.share
+                            .context
+                            .ClearBufferfi(gl::DEPTH_STENCIL, 0, depth, stencil as _);
+                    },
+                    (Some(depth), None) => {
+                        self.share
+                            .context
+                            .ClearBufferfv(gl::DEPTH, 0, &depth);
```
ZeGentzy: No. You got to use `glClearDepth`.

src/backend/gl/src/queue.rs
```diff
@@ -491,15 +491,24 @@ impl CommandQueue {
...
+                    },
+                    (None, Some(stencil)) => {
+                        self.share
+                            .context
+                            .ClearBufferiv(gl::STENCIL, 0, &(stencil as i32));
```
ZeGentzy: No. You got to use ` glClearStencil`.

src/backend/gl/src/native.rs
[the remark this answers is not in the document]
msiglreith: I had some like this in mind:
```rust
let uses_ds = match self.depth_stencil {
       Some(ds) => ds == at_id,
       None => false,
};
let uses_color = self.color_attachments.iter().any(|id| *id == at_id);
uses_color || uses_ds
```

src/backend/gl/src/queue.rs
msiglreith: right, it isn't correct but should use the other glClearBuffer variants

src/backend/gl/src/queue.rs
msiglreith: yes, but a framebuffer is only a container object. At the end we want to display the content of the backbuffer images. So we want to blit the color images of the backbuffer to the display framebuffer. The currently bound fbo might not include the current backbuffer image we want to present.

Thomspoon: The swapchain doesn't own framebuffers in any other backend, I'm confused why it would own one in the gl backend? The framebuffer is just something to write to, right? So we have multiple framebuffers for our backbuffer images so we can write to them while another one is displaying. In this case when we do a renderpass we bind a framebuffer, which changes the state fbo, and then we draw to that fbo, and then blit from that fbo to default... I'm not really sure what the change would be here

src/backend/gl/src/device.rs
Thomspoon: To not include the null terminator mostly.

Thomspoon: Iterating through all of the active uniforms, one of them will initialize the rest of the vec, so I don't think this is an issue.

src/backend/gl/src/device.rs
Thomspoon: Nothing because it's not supported... We could probably remove this since it's not actively being used.

src/backend/gl/src/device.rs
Thomspoon: Push constants don't need to be remapped, spirv-cross translates push constants to a single uniform called PushConst, and we use OpenGL reflection to get the information from it.

src/backend/gl/src/command.rs
Thomspoon: It's difficult when one command needs to do many things. It would get pretty ugly if we had one command for one operation... *shrug* let me know if you want anything split out

src/backend/gl/src/queue.rs
Thomspoon: glClearStencil and glClearDepth aren't correct because we want to clear them on the default framebuffer. I'm using glClearBufferfi to clear both, glClearBufferfv to clear depth and glClearBufferiv to clear stencil as recommended here https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glClearBuffer.xhtml

src/backend/gl/src/queue.rs
msiglreith: > The swapchain doesn't own framebuffers in any other backend, I'm confused why it would own one in the gl backend?

To be sure OpenGL Framebuffers and Vulkan Framebuffers are not exactly the same concept. At higher level, they are equal in terms of being a collection of attachments which will be used in draw commands, but differ due to the concept of renderpasses introduced by vulkan, which lead to a slight separation of responsibilites: a framebuffer alone only stores attachment while a subpasses hold the additional information about which attachments will be used during the draw commands.

> The framebuffer is just something to write to, right?

Just but we don't write to the framebuffer but rather access the attachments associated with it.

> So we have multiple framebuffers for our backbuffer images so we can write to them while another one is displaying

The backbuffer framebuffers would be purely for blitting and we don't bind them for any draw operations.

> In this case when we do a renderpass we bind a framebuffer, which changes the state fbo, and then we draw to that fbo, and then blit from that fbo to default... I'm not really sure what the change would be here

An example which should better show the intended logic:
```rust
cmd_buffer.begin_renderpass(pass1);  // bind GL-FBO of first subpass (e.g contains 2 color targets (one of these is present image #2) and one depth stencil)
..
cmd_buffer.next_subpass(); // bind GL-FBO of second subpass (e.g containts 1 color target (non-present image) and depth-stencil
...
cmd_buffer.end_renderpass();
...

queue.submit(cmd_buffer);
queue.present(); // currently bound GL-FBO of second subpass, we want to present image #1 of prev frame -> we bind a GL-FBO which only contains present image #1 and blit it to our display FBO
```

Thomspoon: That would complicate the backend and make it slow imo. If we have a framebuffer per subpass we'd have to blit each framebuffer into the default framebuffer. Rather than just keeping a single framebuffer for an entire renderpass.

src/backend/gl/src/device.rs
ZeGentzy: All UBOs must be remapped. OGL bindings are per shader program. Vulkan bindings are per stage.

Consider this, what if two different shader stages receive two different PSBs? The two PSBs must be assigned two different bindings. These two bindings must also not conflict with the bindings of any other UBO. The solution to this is to remap them like all the other UBOs.

src/backend/gl/src/device.rs
ZeGentzy: `glGetUniformLocation`'s string must be null terminated:
> name:
> Points to a null terminated string containing the name of the uniform variable whose location is to be queried.

Therefore, `name` must be null terminated. `set_len` should be given `uniform_max_size`, and the last byte should be manually set to '\0'.

`glGetActiveUniform` should still receive `uniform_max_size - 1`

ZeGentzy: Untrue. If all the active uniforms have a length less than `uniform_max_size - 1` then none of them will initialize the whole range.

Thomspoon: That's what this is querying:
```rust
            gl.GetProgramiv(
                program,
                gl::ACTIVE_UNIFORM_MAX_LENGTH,
                &mut uniform_max_size,
            );
```

src/backend/gl/src/device.rs
Thomspoon: I don't really understand. What we have here works, but I might only be using one stage? What stages are you referring to?

src/backend/gl/src/device.rs
ZeGentzy: As it stands, `glGetActiveUniform` could write to all `0..uniform_max_size - 1` bytes, and then `glGetUniformLocation` would read past the buffer.

ZeGentzy: Hmmm, forgot the "ACTIVE" parts, although `glGetProgramiv` isn't to be trusted, as drivers are buggy: https://stackoverflow.com/questions/12555165/incorrect-value-from-glgetprogramivprogram-gl-active-uniform-max-length-outpa

Let me modify what I said just slightly: If any of the active uniforms have a length less than uniform_max_size - 1 then one of them will not initialize the whole range.

I'm afraid you aren't reusing the vector between iterations, instead making a new one each time.

src/backend/gl/src/device.rs
ZeGentzy: Sort've rushing to type this, forgive errors.

Multiple push constant ranges can be used in a single pipeline:

> pushConstantRangeCount is the number of push constant ranges included in the pipeline layout.
>
> pPushConstantRanges is a pointer to an array of VkPushConstantRange structures defining a set of push constant ranges for use in a single pipeline layout. In addition to descriptor set layouts, a pipeline layout also describes how many push constants can be accessed by each stage of the pipeline.

-- Vulkan specs, 13.2.2

There may only be one push constant block per stage:
> There must be no more than one push constant block statically used per shader entry point.

-- Vulkan specs, 14.5.1

Now consider this, application A marks bytes 0..8 as accessible from the vertex shader, and bytes 8..16 as accessible from the fragment shader.

Let's say their vertex shader has this:
```
layout(push_constant) uniform pushConstants {
    layout(offset = 0) float test1;
    layout(offset = 4) float test2;
} u_pushConstants;
```

And the fragment this:
```
layout(push_constant) uniform pushConstants {
    layout(offset = 8) float test3;
    layout(offset = 12) float test4;
} u_pushConstants;
```

Now, afaik (haven't tested), spriv-cross would translate them to something like this:
```
struct PushConsts
{
    float test{3,1};
    float test{4,2};
};

uniform PushConsts push;
```

In OGL, UBO binding locations are per shader program, not per shader stage. In OGL, both PCB would need to be bound to a binding which:
  A) Does not conflict with the other PCB
  B) Does not conflict with any UBOs.

For A:
If our fictional application A writes the floats F to the first PCB and F' the second PCB, we could not grant them the same binding, as they could contain different information, and we can't be binding different sets of info to the same binding. *

For B:
If our fictional application A also had four UBOs, which have reserved the binding location 1, 2, 3, and 4 (post-remapping), we would have to assign our first PCB binding 5, and the second binding 6. We couldn't assign them bindings 1 and 3, because those are used.

Now, if we treat PCBs as UBOs, and share the existing infrastructure (primarily our `DescRemapData`), we could simplify the two requirements to just one: The binding must not conflict with any other UBOs. 

Indeed, we would just grab one binding of `btype == n::BindingTypes::UniformBuffers` per PCB, and call it a day.

*Actually, instead of having separate UBOs for each PCB, we could just have one big UBO, and make the floats `test{3,4}` as offsets into that one big UBO. That would avoid having to copy the same data into multiple different UBOs if the PCBs have overlapping ranges (which is allowed). Afaik, PCBs don't have to be continuous, so this might end up being necessary. Nevertheless, the binding for this big UBO must still not conflict with the binding of any of the other UBOs.

ZeGentzy: **It should be noted that we also remap UBOs. (I think I've said this already.)

Consider if fictional application A had two UBOs in the vertex shader (UB1 and UB2, respectively), bound to bindings 1 and 2. It also has two UBOs in the fragment shader (UB1' and UB2', respectively), also bound to bindings 1 and 2. We would in this case remap UBOs [UB1, UB2, UB1', UB2'] to bindings [1, 2, 3, 4]. That's why I state post-remapping.

src/backend/gl/src/device.rs
```diff
@@ -822,16 +822,18 @@ impl d::Device<B> for Device {
...
                 let mut length = 0;
                 let mut size = 0;
                 let mut utype = 0;
-                let mut name = Vec::with_capacity(uniform_max_size as usize);
-                name.set_len(uniform_max_size as usize - 1);
                 gl.GetActiveUniform(
                     program,
                     uniform as _,
-                    uniform_max_size - 1 as i32,
+                    uniform_max_size as i32,
```
ZeGentzy: This one stays as minus one.

src/backend/gl/src/device.rs
```diff
@@ -891,28 +960,81 @@ impl d::Device<B> for Device {
...
         gl.BindFramebuffer(target, name);
 
-        let att_points = [
-            gl::COLOR_ATTACHMENT0,
-            gl::COLOR_ATTACHMENT1,
-            gl::COLOR_ATTACHMENT2,
-            gl::COLOR_ATTACHMENT3,
-        ];
+        let mut max_color_attachments = 0;
+        gl.GetIntegerv(gl::MAX_COLOR_ATTACHMENTS, &mut max_color_attachments);
```
msiglreith: minor: query these in the physical device limits and use those values

src/backend/gl/src/device.rs
```diff
@@ -891,28 +960,81 @@ impl d::Device<B> for Device {
...
+        for attachment in &pass.attachments {
+            if color_attachment_index > max_color_attachments as _ {
+                panic!(
+                    "Invalid number of color attachments: {} color_attachment of {}",
+                    color_attachment_index, max_color_attachments
+                );
+            }
+
+            let color_attachment = match color_attachment_index {
+                0 => gl::COLOR_ATTACHMENT0,
```
msiglreith: shorter: gl::COLOR_ATTACHMENT0 + index

src/backend/gl/src/device.rs
```diff
@@ -891,28 +960,81 @@ impl d::Device<B> for Device {
...
+        }
 
         let mut attachments_len = 0;
-        //TODO: exclude depth/stencil attachments from here
-        for (&att_point, view) in att_points.iter().zip(attachments.into_iter()) {
+        for (&render_attachment, view) in render_attachments.iter().zip(attachments.into_iter()) {
             attachments_len += 1;
             if self.share.private_caps.framebuffer_texture {
-                Self::bind_target(gl, target, att_point, view.borrow());
+                Self::bind_target(gl, target, render_attachment, view.borrow());
```
msiglreith: for more complex renderpasses this won't be possible, but we can keep it as is so far

src/backend/gl/src/command.rs
```diff
@@ -69,6 +69,17 @@ pub enum Command {
     },
     BindIndexBuffer(gl::types::GLuint),
     //BindVertexBuffers(BufferSlice),
+    BindUniform {
+        uniform: n::UniformDesc,
+        buffer: BufferSlice,
+    },
+    BindRasterizer {
+        rasterizer: pso::Rasterizer,
+        is_embedded: bool,
```
msiglreith: what's the meaning of this one?

src/backend/gl/src/queue.rs
```diff
@@ -835,7 +996,27 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
         S: 'a + Borrow<native::Semaphore>,
         Iw: IntoIterator<Item = &'a S>,
     {
+        let gl = &self.share.context;
+
         for swapchain in swapchains {
+            let extent = swapchain.0.borrow().extent;
+
+            gl.BindFramebuffer(gl::READ_FRAMEBUFFER, self.state.fbo);
```
msiglreith: we won't come around blitting from the actual present image to the framebuffer in the future, but also fine for now. Should work for a large number of cases.

src/backend/gl/src/command.rs
Thomspoon: The command or the flag? I'm not sure what the flag is for.. I copied it from commented code..

src/backend/gl/src/device.rs
Thomspoon: I wasn't sure if all of them were contiguous or not... it makes sense that they would be on every platform.

src/backend/gl/src/device.rs
Thomspoon: Assuming that's an old bug, and would have been fixed in the seven years since it was found *crosses fingers*, I hope we don't have to rely on that. This code is a little squirrely since we're relying on glGetActiveUniform to read up until a null terminator but giving it a string of size `uniform_max_size`. I've tested this code with many different variants of uniforms and it seems to be alright for now.

src/backend/gl/src/device.rs
Thomspoon: Due to my lack of knowledge of SPIRV reflection, I just looked at the spec and found that you cannot query Sampler2D uniforms the same as normal uniforms.

kvark:
If I understand correctly, the only remaining bit for the PR is fixing the CI?

Thomspoon:
@kvark That and figure out what to do with @ZeGentzy's UBO remap. Is that a simple change that I can add? I'm just not sure how it helps.

kvark:
@Thomspoon after reading this now, here is how I understand the push constant situation:
  - we are putting all the push constants in a single UBO
  - whether it's a UBO per pipeline, per pipeline layout, or something else is not important (let's start with per pipeline).
  - when the user pushes constants, we just ask the driver to update the (sub) contents of this UBO
  - this UBO may need to be bound with a different offset per pipeline stage
  - this UBO needs to participate in UBO remapping (whatever this is, I don't recall) like all the other UBOs

@ZeGentzy did I get close to your idea?
Anyhow, it doesn't seem obvious at all, and more importantly - this doesn't have to be done in this PR.

ZeGentzy:
Nearly:
> this UBO may need to be bound with a different offset per pipeline stage

The members of the UBO in the GLSL should have their offsets set individually. The whole UBO should be bound without an offset.

src/backend/gl/src/device.rs
ZeGentzy: Are you saying this can't be done with SPIRV reflection? If it can, than I'd rather it is.

kvark:
@ZeGentzy 

> The members of the UBO in the GLSL should have their offsets set individually. The whole UBO should be bound without an offset.

That sounds like it would require a SPIRV-Cross change?

> Afaik, it should be per pipeline.

Is there a strong reason for this to be the case? I can only think of an expectation that push constants don't change *too much* within a pipeline. But then, again, the whole point of push constants is to change them often, so...

ZeGentzy:
> That sounds like it would require a SPIRV-Cross change?

Correct.

> Is there a strong reason for this to be the case? I can only think of an expectation that push constants don't change _too much_ within a pipeline. But then, again, the whole point of push constants is to change them often, so...

Couldn't different pipelines have different push constants?

kvark:
> Couldn't different pipelines have different push constants?

Push constants are a state of a command buffer being recorded, so technically it doesn't really matter for the implementation if it's going to be changing one uniform buffer or the other, as long as we stay within the same pipeline layout.

grovesNL:
Agreed that we can follow up with that afterwards. Are we ready to proceed with this PR for now?

PR #2674, opened 2019-03-04
[dx12] coerse image views to arrays

kvark:
Fixes the ability to create 2D views out of 2DArray images.
PR checklist:

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -553,6 +553,7 @@ impl Device {
 
         match info.view_kind {
             image::ViewKind::D1 => {
+                assert_eq!(info.range.layers, 0 .. 1);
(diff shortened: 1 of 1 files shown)


msiglreith:
Assigning a array view to a non-array resource on shader side is fine?

kvark:
Apparently! It's not like we have a choice anyway :) And yes, the debug layer doesn't complain.

msiglreith:
Could you add an additional condition for non-0 base array level to use arrays?
Else looks fine for me

kvark:
> Could you add an additional condition for non-0 base array level to use arrays?

Could you clarify this please?
What this is doing is switching to XxArray view kind whenever the *source* image is an array. So it doesn't care if the view is 0-base or not.

msiglreith:
My bad, I mixed something up
bors r+

PR #2675, opened 2019-03-04
implement fmt::Debug for RawDevice

### LINKED ISSUE #2660: impl fmt::Debug for RawDevice is unimpemented
Short info header:
- GFX version: 0.1.0
- OS: window (all)
- GPU: vulkan

It's ok if it prints nothing, but unimplemented is really annoying as setting log level to trace results in panic. I have to explicitly disable trace for some modules.

hgallagher1993:
Fixes #2660  
PR checklist:
- [X] tested examples with the following backends: Vulkan


hgallagher1993:
Woops didn't see that #2660 was done

PR #2680, opened 2019-03-06
Improve vertex buffer binding documentation

### LINKED ISSUE #2662: Improve the docs around VertexBufferDesc, AttributeDesc, bind_vertex_buffers
They are a bit misleading as they could make you think that vertex buffer binding points are the same as descriptor binding points, and don't describe what the 'Offset' is doing in the bind_vertex_buffers iterator. This issue is mostly a reminder for myself to go fix it...

termhn:
Fixes #2662

PR checklist:

### DIFF
--- a/src/hal/src/command/raw.rs
+++ b/src/hal/src/command/raw.rs
@@ -209,7 +209,18 @@ pub trait RawCommandBuffer<B: Backend>: Any + Send + Sync {
     /// will operate on.
     ///
     /// Each buffer passed corresponds to the vertex input binding with the same index,
-    /// starting from an offset index `first_binding`.
+    /// starting from an offset index `first_binding`. For example an iterator with
+    /// two items and `first_binding` of 1 would fill vertex buffer binding numbers
+    /// 1 and 2.
+    ///
+    /// This binding number refers only to binding points for vertex buffers and is
+    /// completely separate from the binding numbers of `Descriptor`s in `DescriptorSet`s.
+    /// It needs to match with the `VertexBufferDesc` and `AttributeDesc`s to which the
+    /// data from each bound vertex buffer should flow.
+    ///
+    /// The `buffers` iterator should yield the `Buffer` to bind, as well as an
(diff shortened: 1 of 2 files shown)


src/hal/src/command/raw.rs
```diff
@@ -209,7 +209,18 @@ pub trait RawCommandBuffer<B: Backend>: Any + Send + Sync {
     /// will operate on.
     ///
     /// Each buffer passed corresponds to the vertex input binding with the same index,
-    /// starting from an offset index `first_binding`.
+    /// starting from an offset index `first_binding`. For example an iterator with
+    /// two items and `first_binding` of 1 would fill vertex buffer binding numbers
+    /// 1 and 2.
+    ///
+    /// This binding number is local to the currently bound pipeline and is separate
```
kvark: I don't think this is correct to say. It reads as if the binding gets reset upon changing the pipeline. The binding is just separate from descriptor sets, and pipeline knows how to fetch the data from those bindings.

src/hal/src/pso/input_assembler.rs
```diff
@@ -45,11 +45,14 @@ pub struct Element<F> {
...
 
-/// Vertex buffer descriptor
+/// Vertex buffer description. Notably, completely separate from resource `Descriptor`s
+/// used in `DescriptorSet`s.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct VertexBufferDesc {
-    /// Binding number of this vertex buffer descriptor.
+    /// Binding number of this vertex buffer. This binding point is
+    /// unique to a specific pipeline, and is completely separate from
```
kvark: same wording here

src/hal/src/command/raw.rs
termhn: Yeah that wording was a bit bad. Fixed.

kvark:
Bors r+

PR #2681, opened 2019-03-06
Only discard rasterization if no fragment shader and the Depth/Stencil tests are both disabled (Vulkan backend)

ForestCSharp:
…ell as there being no pixel shader)

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -543,10 +543,9 @@ impl d::Device<B> for Device {
                     } else {
                         vk::FALSE
                     },
-                    rasterizer_discard_enable: if desc.shaders.fragment.is_none() {
-                        vk::TRUE
-                    } else {
-                        vk::FALSE
+                    rasterizer_discard_enable: match (&desc.shaders.fragment, &desc.depth_stencil.depth, &desc.depth_stencil.stencil) {
+                        (None, pso::DepthTest::Off, pso::StencilTest::Off) => vk::TRUE,
+                                                                         _ => vk::FALSE,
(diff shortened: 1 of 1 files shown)


kvark:
We need to double check with Vulkan spec on what is the expected behavior when the fragment shader is not set. IIRC, it's D3D that allows no fragment shader, but Vulkan expects it anyway, even if empty.

ForestCSharp:
From the Spec:  "...and fragment shaders, if present, operating on fragments generated by Rasterization." seems to imply that fragment shaders are optional. Also, the VkGraphicsPipelineCreateInfo valid usage states that only a stage with VK_SHADER_STAGE_VERTEX_BIT is required.

kvark:
Right. We need more info though: fragment shadersbeing optional is not enough, we need to find out what the spec does in this case.

ForestCSharp:
Some more info from the Spec (9.2.1. Valid Combinations of Stages for Graphics Pipelines): 
"If a fragment shader is omitted, fragment color outputs have undefined values, and the fragment depth value is unmodified. This can be useful for depth-only rendering."

It then goes to provide an example of depth-only rendering, which requires 
1. Vertex Shader
2. VkPipelineVertexInputStateCreateInfo
3. VkPipelineInputAssemblyStateCreateInfo
4. VkPipelineViewportStateCreateInfo
5. VkPipelineRasterizationStateCreateInfo
6. VkPipelineMultisampleStateCreateInfo
7. VkPipelineDepthStencilStateCreateInfo

kvark:
Thanks!
Bors r+

PR #2683, opened 2019-03-07
[mtl] expose custom sampler configuration

### LINKED ISSUE #2682: [mtl] provide platform-specific API to create unnormalized sampler coordinates
Metal supports unnormalized coordinates, unlike Vulkan:
https://developer.apple.com/documentation/metal/mtlsamplerdescriptor/1516289-normalizedcoordinates

We need to use them to emulate `GL_TEXTURE_RECT` used by Gecko on Apple
cc @zakorgy

kvark:
Fixes  #2682
PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -703,6 +703,74 @@ impl Device {
 
         arg
     }
+
+    pub fn make_sampler_descriptor(
+        &self,
+        info: image::SamplerInfo,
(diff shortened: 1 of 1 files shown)


msiglreith:
Vulkan also supports unnormalized coordinates as sampler parameter, no need for a backend specific functionality.

kvark:
Oh, good, I missed that!
The PR is still fine to land though

kvark:
we can roll back after we get the good API, or even just turn those entry points into private
bors r+

PR #2684, opened 2019-03-08
Add documentation to `gfx_app`

tangmi:
(Note: this is for `pre-ll`)

I understand that `gfx_app` is meant to be a framework for the `gfx` examples, but I've found it to be very useful in prototype or toy renderers. My original intention was to generalize it a little such that more kinds of projects could use it, but I figured I'd start by documenting the existing code (and potentially creating a new project that just does the cross platform window/backend initialization that `gfx_app` does).

Please let me know if this PR is useful and what improvements I can make on it, if so!

Cheers!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,11 +57,7 @@
 //!     }
 //! }
 //! 
-//! fn main() {
-//!     use gfx_app::Application;
-//! 
-//!     App::launch_simple("Window title");
-//! }
+//! // Then just call `App::launch_simple("Window title");` from `main`.
 //! ```
 //! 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -12,6 +12,56 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Cross platform GFX example application framework.
+//! 
+//! Supports OpenGL 3.2, OpenGL ES 2.0, WebGL 2, DirectX 11, Vulkan, and Metal. Handles windowing via `winit`.
```
kvark: There is no Vulkan and Metal support in the old gfx-rs. The existing work has not been completed.

tangmi:
Thanks for taking the time to look at this! I've updated the crate documentation to exclude Vulkan and Metal from the list of supported backends.

PR #2686, opened 2019-03-08
[pre-ll] Add format::R8_G8_B8 for opengl

alexheretic:
This change adds `format::R8_G8_B8` support for opengl. This is useful to me as I have some large rgb pngs and this avoid me having to convert them to rgba before uploading to the gpu. It may help with vram usage in this case too, although maybe some drivers already optimise this.

PR checklist:

### DIFF
--- a/src/backend/dx11/src/data.rs
+++ b/src/backend/dx11/src/data.rs
@@ -39,7 +39,7 @@ pub fn map_format(format: Format, is_target: bool) -> Option<DXGI_FORMAT> {
     use core::format::SurfaceType::*;
     use core::format::ChannelType::*;
     Some(match format.0 {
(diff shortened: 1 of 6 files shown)


kvark:
@alexheretic I don't believe that the actual drivers/GPU support this format, and that's why it's not present in DX11. So you aren't really saving anything by trying to use it. You'd be sacrificing portability (i.e. wgpu-rs doesn't expose it either) and just pushing the trouble down the stack, forcing the poor GL driver to accomodate and pretend it's 3 bytes per texel (while in fact it's still 4).

I strongly recommend just using RGBA textures for this case to start with. If you have large textures and concerned about space, texture compression is the right thing to do (and yes, sorry, it's not quite available in pre-ll right now).

alexheretic:
Ah ok that explains why there are no vram savings. The cpu savings are probably just in debug mode then. Thanks for the explanation!

PR #2687, opened 2019-03-08
[mtl] expose line fill, fix MSAA views

kvark:
Fixes vkQuake2 running on gfx-portability.
Changes are relatively harmless.
PR checklist:

### DIFF
--- a/src/backend/metal/src/conversions.rs
+++ b/src/backend/metal/src/conversions.rs
@@ -1176,6 +1176,22 @@ pub fn map_winding(face: pso::FrontFace) -> MTLWinding {
     }
 }
 
+pub fn map_polygon_mode(mode: pso::PolygonMode) -> MTLTriangleFillMode {
+    match mode {
+        pso::PolygonMode::Point => {
+            warn!("Unable to fill with points");
+            MTLTriangleFillMode::Lines
+        }
+        pso::PolygonMode::Line(width) => {
+            if width != 1.0 {
+                warn!("Unsupported line width: {:?}", width);
+            }
+            MTLTriangleFillMode::Lines
+        }
+        pso::PolygonMode::Fill => MTLTriangleFillMode::Fill,
+    }
+}
+
 pub fn map_cull_face(face: pso::Face) -> Option<MTLCullMode> {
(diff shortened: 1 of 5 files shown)


kvark:
@kondrak hi! Just a small note that MoltenVK is not the only thing capable of providing Vulkan on Apple platforms. With this fix in, https://github.com/gfx-rs/portability runs vkQuake2 fine as well.

kondrak:
Neat! I think I might give it a shot and see what it's capable of. Great job! :+1:

kondrak:
So looking at the patch it seems that vkQuake2 is the only title you ran into that uses lines and points? Curious!

src/backend/metal/src/device.rs
```diff
@@ -2399,7 +2401,14 @@ impl hal::Device<Backend> for Device {
             levels: 0..raw.mipmap_level_count() as image::Level,
             layers: 0..image.kind.num_layers(),
         };
-        let mtl_type = conv::map_texture_type(kind);
+        let mtl_type = if image.mtl_type == MTLTextureType::D2Multisample {
```
grovesNL: There's also `MTLTextureType2DMultisampleArray` in macOS 10.14+, I just opened #2689 to track it

src/backend/metal/src/conversions.rs
```diff
@@ -1176,6 +1176,22 @@ pub fn map_winding(face: pso::FrontFace) -> MTLWinding {
     }
 }
 
+pub fn map_polygon_mode(mode: pso::PolygonMode) -> MTLTriangleFillMode {
+    match mode {
+        pso::PolygonMode::Point => {
+            warn!("Unable to fill with points");
+            MTLTriangleFillMode::Lines
```
grovesNL: Is the last line removed if we submit an odd number of points?

kvark:
Thanks @grovesNL ! Last line is not removed, since this is the fill mode, not to be confused with input geometry.

Bors r=grovesNL

grovesNL:
Oops :) right, the amount of vertices doesn’t matter because this only applies to polygons.

PR #2690, opened 2019-03-10
[v0.18] Update winit/glutin -> 0.19/0.20

alexheretic:
Bumped for release
* gfx_window_dxgi -> `0.19`
* gfx_window_glutin -> `0.30`

Also updated gfx_app `env_logger` and some dev-dependencies because why not?

PR checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,14 +24,14 @@ name = "gfx_app"
 
 [dependencies]
 log = "0.4"
-env_logger = "0.5"
-glutin = "0.19"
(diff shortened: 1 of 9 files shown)


alexheretic:
> I assume you need those updates to be published?

Yes please, at least `gfx_window_glutin` which is what I'm using.

kvark:
https://crates.io/crates/gfx_window_glutin/0.30.0
https://crates.io/crates/gfx_window_dxgi/0.19.0

PR #2694, opened 2019-03-12
Fix acquire semaphore initialization in quad example

Veykril:
Fixes a small error in the quad example that would cause problems for when `frames_in_flight` < `frame_images.len()`. Aligns the code with what the comments say.


kvark:
Thank you!
bors r+

PR #2696, opened 2019-03-14
[dx12] clear attachments and SRV component swizzles

kvark:
Fixes WebRender issue with clear rectangles on DX12. cc @zakorgy 
Note: the swizzles are only implemented for SRV. I don't know how to get them for other views.
PR checklist:


src/backend/dx12/src/device.rs
```diff
@@ -538,6 +541,9 @@ impl Device {
     ) -> Result<(), image::ViewError> {
         #![allow(non_snake_case)]
 
+        if info.component_mapping != IDENTITY_MAPPING {
+            warn!("Unsupported component_mapping for RTV: {:?}", info.component_mapping);
```
msiglreith: I don't think we should warn for these as it's not like this is allowed in the first case for vulkan in any use case.

PR #2697, opened 2019-03-14
[0.1/dx12] array attachment clear fix and SRV component mapping

kvark:
Sibling of #2696
PR checklist:

### DIFF
--- a/src/backend/dx12/src/conv.rs
+++ b/src/backend/dx12/src/conv.rs
@@ -5,11 +5,11 @@ use std::mem;
 
 use winapi::shared::basetsd::UINT8;
 use winapi::shared::dxgiformat::*;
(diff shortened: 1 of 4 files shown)


kvark:
Err, something is horribly broken here (5K loc changes)

kvark:
bors r=msiglreith

src/backend/dx12/src/conv.rs
```diff
@@ -104,6 +104,25 @@ pub fn map_format(format: Format) -> Option<DXGI_FORMAT> {
     Some(format)
 }
 
+pub fn map_swizzle(swizzle: Swizzle) -> UINT {
+    use hal::format::Component::*;
+
+    [swizzle.0, swizzle.1, swizzle.2, swizzle.3]
```
msiglreith: that looks really clean!

PR #2201, opened 2018-06-30
[WIP] Add more physical device limits

AIOOB:
Unfortunately, I'm away for a week or so but I thought I may as well stick this up, anyone is welcome to grab what I've done or commit to this branch in the meantime, I'll be on Gitter.

I think I may have got a few limits wrong already so some hints from people who know the backends better would be appreciated :smiley:

PR checklist:
- [ ] `make` succeeds (on *nix)
- [ ] `make reftests` succeeds
- [ ] tested examples with the following backends:

PR #2698, opened 2019-03-14
More detailed limits

### LINKED ISSUE #2041: dx12: Expose texture limits
Expose the (at least minimum) supported texture dimensions for each texture type.
Requires additional fields in limits.

Blocking `texture3d` sample for portability.

kvark:
Closes #2201 
Fixes #2041
Allows WebRender to avoid hard-coding the texture limits. cc @zakorgy 

PR checklist:

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -180,8 +180,8 @@ static QUEUE_FAMILIES: [QueueFamily; 4] = [
 
 pub struct PhysicalDevice {
     adapter: native::WeakPtr<dxgi1_2::IDXGIAdapter2>,
-    features: hal::Features,
-    limits: hal::Limits,
+    features: Features,
+    limits: Limits,
     format_properties: Arc<FormatProperties>,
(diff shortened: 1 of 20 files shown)


src/backend/dx12/src/lib.rs
```diff
@@ -1036,16 +1036,20 @@ impl hal::Instance for Instance {
                     Features::INSTANCE_RATE |
                     Features::SAMPLER_MIP_LOD_BIAS,
                 limits: Limits { // TODO
-                    max_texture_size: 0,
+                    max_image_1d_size: d3d12::D3D12_RESOURCE_DIMENSION_TEXTURE1D as _,
```
msiglreith: these enums doesn't makes sense in this context.
these sgould be used instead similar to the dx11 implementation: https://retep998.github.io/doc/winapi/d3d12/constant.D3D12_REQ_TEXTURE1D_U_DIMENSION.html

kvark:
@msiglreith great catch! PTAL

PR #2699, opened 2019-03-14
Metal presentation signposts

kvark:
Allows us to get a better idea of the swapchain performance when looking at a system trace. Disabled by default.
PR checklist:


kvark:
alright, no reviews, but seemingly harmless
bors r+

PR #2700, opened 2019-03-15
[mtl] retain NSView when used with existing layer

kvark:
Fixes the crash on exit in Dolphin
PR checklist:

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -204,12 +204,13 @@ impl Instance {
             panic!("window does not have a valid contentView");
         }
 
-        // temporary, hopefully!
+        // Deprecated! Clients should use `create_surface_from_layer` instead.
         let is_actually_layer: BOOL = msg_send![view, isKindOfClass: class];
         if is_actually_layer == YES {
             return self.create_from_layer(view);
         }
 
+        msg_send![view, retain];
         let existing: CAMetalLayer = msg_send![view, layer];
         let use_current = if existing.is_null() {
             false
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/lib.rs
```diff
@@ -210,6 +210,9 @@ impl Instance {
             return self.create_from_layer(view);
         }
 
+        // TODO: figure out why this is needed. Dolphin goes the `use_currect` path
+        // and crashes on exit when trying to release `NSView` otherwise.
+        msg_send![view, retain];
```
grovesNL: This seems right to me (even for the `existing` path), so maybe the `TODO` isn't needed?

Unless I'm misunderstanding, both `create_from_layer` and `create_from_uiview` appear to always `retain` already because we'll always `release` during `SurfaceInner`'s `drop` if `self.view` is `Some` (which it appears to be here).

In other words:

- With this change, we unconditionally `retain` here, and `release` during `drop` (i.e. add `1` to the ref count here and subtract `1` during `drop`). This seems right.
- Without this change, we don't `retain` here for `existing`, yet still `release` during `drop` (i.e. add `0` to the ref count here and subtract `1` during `drop`). This doesn't seem right, because we might cause `view` to be released too early if something else was holding a reference.

kvark:
That's the confusing part: we store it as cocoa::id, which is just a typedef for a pointer. Is it released on drop? I expect not.

kvark:
Bors r=grovesNL 

Let's continue discussion anyway, it's just clear that this is fixing a problem, for now.

grovesNL:
We do have `msg_send![object, release];` in `SurfaceInner`'s `drop`, so we'd need `retain`/`release` to be balanced at least.

I think `id` still uses the same referenced object for these memory management calls (i.e. `retain` and `release` aren't acting on the pointer itself, instead the actual view is used).

kvark:
Ah, right! Ok, we should be good then, great :)

Issue #2702, opened 2019-03-15
Eliminate excessive memcpy instructions


kvark:
Rust is lazy when constructing LLVM IR, relies a lot on LLVM optimization passes, which don't always deliver. One of the biggest issues is `memcpy` of larger chunks across the stack. We now have a helper tool to analyze those: https://github.com/jrmuizel/memcpy-find

kvark:
Here comes https://github.com/kvark/copyless

PR #2705, opened 2019-03-17
[mtl] use copyless vec helper

kvark:
Metal piece of #2702

This is just an optimization. Main idea is to bring structure construction closer to the placement spot. I.e. instead of doing 1) construct 2) try extending the vector 3) write into it, we now do 2-1-3, which allows the compiler to construct the data in place (without any `box` magic).

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -20,6 +20,7 @@ use hal::{
 
 use block::ConcreteBlock;
 use cocoa::foundation::{NSRange, NSUInteger};
+use copyless::VecHelper;
 use foreign_types::ForeignType;
 use metal::{
     self,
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/command.rs
```diff
@@ -785,8 +786,10 @@ impl Journal {
                     soft::Pass::Compute => self.compute_commands.len(),
                     soft::Pass::Blit => self.blit_commands.len(),
                 };
-                self.passes
-                    .push((pass.clone(), range.start + offset..range.end + offset));
+                self.passes.alloc().init((
```
grovesNL: Is it possible to only have `push_no_copy` to avoid repeating `alloc().init()` and make it clear what's happening here?

kvark: nope :) this is the tricky bit - the whole point of that change is to separate value creation from the memory allocation. So the creation *has to* happen after `alloc()` is called, and wrapping it into a helper would ruin the idea.

kvark:
> Do we know whether this has a positive performance impact?

Nope. We did see visible impact on WebRender when doing a similar change though.
I'm planning on coming up with some crazy tests in `copyless` that verify when it works, and allows us to catch the moment of compiler maturing enough ;)

Let's go ahead, it's not a major change.
bors r=grovesNL

kvark:
Note: I screwed up one bit here, causing a panic. Quickly hot-fixed in the follow-up 3afab46c73a5cf711cd398deda8acb9a009a5b2a

PR #2706, opened 2019-03-17
[hal-0.1] Make Backend impl Default

termhn:
This allows types that are generic over Backend to be stored in a `specs::World`; backported to 0.1 so that it can be used with `rendy` until hal 0.2 release

PR checklist:


kvark:
Would it be possible for the user to implement `Default` manually as a workaround?
Also, I'm suspecting that changing `enum` to `struct` on public type is a breaking change...

termhn:
I've realized that I was being dumb and this isn't required. Perhaps deriving default should still be done, but it's not a blocker in this case

PR #2708, opened 2019-03-22
Remove FrameSync, add support for Metal acquiring fences

### LINKED ISSUE #2711: Remove FrameSync
Apparently, there are legit cases for applications wanting to do both semaphore and fence synchronization on acquiring a swapchain frame:
  - semaphore used to chain up following submissions
  - fence used for waiting later on, as back pressure

kvark:
Fixes #2711
It makes us compatible with the updated Diligent Engine.
Also improves Metal limits and logging of fences.

PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -49,7 +49,7 @@ use std::{fs, iter};
 
 use hal::{
     buffer, command, format as f, image as i, memory as m, pass, pool, pso, window::Extent2D,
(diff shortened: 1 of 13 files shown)


kvark:
post-landing review is still needed
bors r+

kvark:
It's a fence versus the FrameSync as a semaphore. It's more strict for pointing to a specific iteration of a frame, i.e. we allow the user to check that fence long after the frame was acquired, even over multiple acquire cycles. We don't allow that for semaphores atm, since it's less practicle, but still possible.

Thanks for all the post-landing reviews!
Hope you landed smooth yourself as well ;)

PR #2715, opened 2019-03-26
[dx12] Remove extra flags on certain MemoryTypes

Brekcel:
Fixes #issue
PR checklist:

I might be misunderstanding how closely gfx-hal and the backends follow the Vulkan Spec but, according to the [Vulkan Spec](https://vulkan.lunarg.com/doc/view/1.0.30.0/linux/vkspec.chunked/ch10s02.html), MemoryTypes can not be `COHERENT` or `CPU_CACHED` without also being `CPU_VISIBLE`. Before this change, removing just the `CPU_VISIBLE` flag would cause some MemoryTypes to have a flag of `COHERENT` or `COHERENT | CPU_CACHED`. This change removes the extra flags as they shouldn't be on there with MemoryTypes that are not `CPU_VISIBLE`.


kvark:
bors retry

kvark:
bors retry

PR #2716, opened 2019-03-26
[mtl] handle incompatible pipelines better

kvark:
Fixes the case when an incompatible pipeline is bound to a pass, but then used in the next pass.
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -190,13 +190,23 @@ struct SubpassInfo {
 /// The current state of a command buffer, used for two distinct purposes:
 ///   1. inherit resource bindings between passes
 ///   2. avoid redundant state settings
+///
+/// ## Spaces
 /// Note that these two usages are distinct and operate in technically different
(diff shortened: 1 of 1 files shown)


kvark:
bors r+
post-landing review is still needed

src/backend/metal/src/command.rs
```diff
@@ -3367,72 +3372,74 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                pre.issue(soft::RenderCommand::SetDepthStencilState(state));
+            }
+        } else {
+            // This may be tricky: we expect either another pipeline to be bound
+            // (this overwriting these), or a new render pass started (thus using these).
+            self.state.rasterizer_state = pipeline.rasterizer_state.clone();
+            self.state.primitive_type = pipeline.primitive_type;
         }
 
-        if set_stencil_references {
```
grovesNL: Why did we have to move this up? Or is it unrelated?

kvark:
Yes, that's correct.
Moved that up because we can't make it optional.

PR #2728, opened 2019-04-04
Fix `gfx_hal::memory::Properties` to match `VkMemoryPropertyFlagBits`

demurgos:
This commit updates the `COHERENT`, `CPU_VISIBLE` and `LAZILY_ALLOCATED` flags to match the order and values defined by Vulkan.

See https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkMemoryPropertyFlagBits.html

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -634,6 +634,12 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
                     type_flags |= Properties::DEVICE_LOCAL;
                 }
                 if mem
+                  .property_flags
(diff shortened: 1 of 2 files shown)


kvark:
That looks good, thank you!
Do you mind also fixing the Vulkan backend to simplify the mapping of this flags type?

demurgos:
I looked into it but I am not familiar enough with the codebase to update the Vulkan backend. I was debugging my own code when I noticed the discrepancy between the values of `gfx-hal` and the Khronos API, but I am just getting started with this lib.

kvark:
Ok, thanks!
Bors r+

Issue #2546, opened 2019-01-02
Replace glsl_to_spirv in examples


LTheGreats:
Short info header:
- GFX version:
- OS:
- GPU:

Michael-Lfx:
glsl-to-spirv is also used in vulkan backend.

LTheGreats:
Really this isn't just about the examples, but all of the gfx-rs code that depends on the deprecated library. (Sorry about the many name changes.)

grovesNL:
I think it should be removed from the Vulkan backend already, see https://github.com/gfx-rs/gfx/pull/2533

kvark:
To clarify: GLSL is in now way a part of gfx-hal API, so the issue is rather harmless. It just affects the way people start using the project, since they replicate the examples, and we shouldn't promote a deprecated library there. So, a documentation issue for the most part.

aleksijuvani:
I don't think shaderc-rs is a very good alternative. It's kind of a pain to build on Windows, requiring a C++ compiler (e.g. usually Visual Studio, which is an ordeal to install), Python and CMake in the path. glsl_to_spirv used glslang instead and included a pre-built binary for Windows.

I'm sure it's not a problem for experienced developers, but it won't make for a good experience for anyone looking to learn Rust or gfx-hal and doesn't have these already installed.

kvark:
Good point! Maybe we can just undeprecate in then?

Michael-Lfx:
> I don't think shaderc-rs is a very good alternative. It's kind of a pain to build on Windows, requiring a C++ compiler (e.g. usually Visual Studio, which is an ordeal to install), Python and CMake in the path. glsl_to_spirv used glslang instead and included a pre-built binary for Windows.
> 
> I'm sure it's not a problem for experienced developers, but it won't make for a good experience for anyone looking to learn Rust or gfx-hal and doesn't have these already installed.

IIRC shaderc is based on glslang, spriv-cross and other KHR projects. And yes it requires a complex environment to build which is difficult for beginners. Here we just need glslang to convert glsl to Vulkan compatible spirv binary. I think it's good to make it simple.  Because [TimNN/glslang-sys](https://github.com/TimNN/glslang-sys) stays in May 10, 2015, I have a proposal that gfx should hold a glslang-rs crate to follow the upstream. @kvark

kvark:
> I have a proposal that gfx should hold a glslang-rs crate to follow the upstream

[...]

PR #2731, opened 2019-04-06
Replace glsl-to-spirv with shaderc

djcsdy:
Replaces glsl-to-spirv with shaderc.

glsl-to-spirv is deprecated and the author states: ["use shaderc-rs instead"](https://crates.io/crates/glsl-to-spirv).

PR checklist:

Note: `make reftests` is failing for dx12 on master, so it’s failing on this branch too. It passes for vulkan.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ before_install:
   # see https://docs.travis-ci.com/user/trusty-ci-environment/
   - if [[ $TRAVIS_OS_NAME == "linux" ]]; then export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start && make travis-sdl2 && export CXX=g++-5; fi
   - if [[ $TRAVIS_OS_NAME == "osx" ]]; then brew update && brew install sdl2; fi
-  - if [[ $TRAVIS_OS_NAME == "windows" ]]; then choco install make; fi
+  - if [[ $TRAVIS_OS_NAME == "windows" ]]; then choco install make ninja; fi
   - rustup self update
   - rustup target add $TARGET; true
 
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -31,9 +31,9 @@ env_logger = "0.5"
 image = "0.19"
 log = "0.4"
 winit = "0.18"
-glsl-to-spirv = "0.1.4"
 gfx-hal = { path = "../src/hal", version = "0.1" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.1" }
+shaderc = "0.3.16"
 
 [dependencies.gfx-backend-gl]
 path = "../src/backend/gl"
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -32,8 +32,8 @@ extern crate gfx_backend_vulkan as back;
 extern crate log;
 extern crate env_logger;
 extern crate gfx_hal as hal;
-extern crate glsl_to_spirv;
 extern crate image;
+extern crate shaderc;
 extern crate winit;
 
 struct Dimensions<T> {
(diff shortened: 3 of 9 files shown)


djcsdy:
Urgh. The CI build is failing on Windows due to the 260 character path limit 🤢.

djcsdy:
This PR now also installs ninja on Travis Windows before the build.

ninja is an optional build-time dependency of shaderc. It is recommended on Windows because without it the shaderc build process generates files with very long pathnames, which can cause build failures.

msiglreith:
Hi and thanks for the PR!
How is the experience for a user who wants to start using gfx-rs by trying out an example or building code on top of existing ones? Are there further build systems required to get started, which might especially a blocker on windows systems?

djcsdy:
This adds a dependency on Python on all platforms, and on Windows sometimes adds a dependency on ninja.

I am using Windows myself and I didn’t encounter significant problems. The standard Windows distribution of Python 3 from https://www.python.org/downloads/windows/ works fine. Of course Python has to be on the PATH. The Python installer can do this for you – it is optional but IIRC turned on by default.

Ninja is required on Windows if the path to the gfx source tree is longer than about 30 characters. This is because, without Ninja, the shaderc build process uses a different project generator that generates excessively long path names, which can cause a build failure on Windows. I didn’t experience this problem on my machine because I had the project checked out at a shorter path, but as you can see it blew up on Travis and it was not immediately obvious what the solution was. I think it would be a good idea for the shaderc project to just add a hard dependency on ninja at build time on Windows, since at least then it could produce a helpful error message.

I’m not aware of any other additional dependencies compared to what this project already depended on – shaderc depends on cmake and a C compiler, but those were already required.

djcsdy:
I proposed a PR to shaderc which would make Ninja mandatory when building for Windows with MSVC: https://github.com/google/shaderc-rs/pull/47

djcsdy:
It turns out that there is also work in progress to [bundle pre-built binaries with shaderc-rs](https://github.com/google/shaderc-rs/issues/44). IIUC this would also make the experience a lot more straightforward for people building on Windows, since it would remove the need for Python and ninja.

termhn:
IMO we should *not* adopt shaderc for the examples but we should in warden. We can make a note that shaderc is a better choice for a real project, but since glsl-to-spirv works for the quite simple use case we need it for, and honestly doesn't really break until you get a fair bit deeper, it "just works", and it doesn't double or more the build times.... I think it is worth it to leave in.

aleksijuvani:
There's some earlier discussion about this subject in #2546.

I agree that shaderc build times are not great. I'm also not a fan of the additional dependencies to Python and ninja. Was CMake already a dependency for gfx-hal? Where is this dependency?

We might be able to avoid these dependencies by wrapping glslang ourselves and invoking it directly instead of going through shaderc.

djcsdy:
I thought CMake was a dependency of gfx already, but I guess I was mistaken. I installed CMake for some Rust project, but I guess it wasn’t this one.

aleksijuvani:
Looking into it, glsl-to-spirv does have CMake as a dependency, but since it includes pre-built binaries on Windows, it's only used on the other platforms.

knappador:
Static lib options and crate behavior have been made more normal and default behaviors are expanded in https://github.com/google/shaderc-rs/pull/46

The expanded static options just make it easier to pick a path that builds really fast with few dependencies.  The old `--no-defaults` option was funky (but still supported, just deprecated).

I don't think this makes life easier on gfx since it's pretty heavily focused on expanding options for Linux CI.  @djcsdy if you want support on getting the build script working on Windows or have anything you want but didn't get, just say so.  

Appveyor is currently advertising a build failure, but I don't have access to the failure info to go fix it.  Time for other-platform people to :sparkles: :sparkles:

djcsdy:
@knappador The build does work on Windows. Check my last commit.

djcsdy:
@knappador Scratch that sorry, I misunderstood. I see you are referring to google/shaderc-rs#46, which is currently failing on Windows.

knappador:
@djcsdy   I'm fixing a line that got regressed as soon as I can shove it into history  _realtime github chat_

knappador:
@djcsdy  okay tests passed on the CI servers.  No idea what platforms or coverage they have.  I don't think you need ninja to do the static lib build paths.  I pretty much just cargo-culted forward what was there for the `--no-defaults` legacy path in build.rs for non-Linux builds

knappador:
[Shaderc's crate split has landed](https://github.com/google/shaderc-rs/pull/46) with some additional options to get pre-built binaries linked.  I'm available to guide integration attempts and land additional PR's against shaderc-rs if more build variations are desired.

imkow:
functionality side, shaderc looks fine, but I think it's not ready for cross platform support. And it's depends on too many 3rd party stuff.

djcsdy:
I don’t see any path forward for me here.

If someone else wants to pick this up, please do. But consider me thoroughly discouraged.

kvark:
@djcsdy sorry about the way this PR progressed (and essentially ended up unmerged)!
Please don't hesitate to visit [our chat](https://gitter.im/gfx-rs/gfx) to talk or just provide feedback.

PR #2733, opened 2019-04-09
[pre-ll] Add BC1 (DXT1), BC3 (DXT5) S3TC texture formats

### LINKED ISSUE #2688: [pre-ll] texture compression support
Short info header:
- GFX version: 0.18
- OS: Linux/Windows
- GPU: OpenGL 3.3 compatible ones

Hey I'd like try texture compression in Robo Instructus. I have some big background & sprite sheet pngs, these can take up a fair bit of VRAM and I'd like to offer texture compression to my users with less of that.

I'm willing to implement this, but it's probably best to discuss how best to go about this. Ie what kinds of compression have best support etc. I'm only interested in opengl, so I'd probably just leave the other APIs untouched, or with any new enums unimplemented.

alexheretic:
Adds compressed BC1 & BC3 image formats for gfx_device_gl/opengl. Image data is expected to be already compressed.

New formats:
* `SurfaceType::BC1_R8_G8_B8` (DXT1)
* `SurfaceType::BC3_R8_G8_B8_A8` (DXT5)

Each supports rgb & srgb.

Resolves #2688

## Motive
I've been investigating texture compression to reduce VRAM usage in my game Robo Instructus. For this end I wanted more efficient 8-bit channel rgb & rgba storage with good opengl driver & hardware support.

S3TC is old and the best supported compression available (see https://opengl.gpuinfo.org/listcompressedformats.php). Patent issues are now history (see https://en.wikipedia.org/wiki/S3_Texture_Compression). For testing I used [GPUOpen-Tools/Compressonator](https://github.com/GPUOpen-Tools/Compressonator) for encoding, which is quite fast and seems to have decent quality.

* `BC1_R8_G8_B8` offers 8:1 compression over `R8_G8_B8_A8`
* `BC3_R8_G8_B8_A8` offers 4:1 compression over `R8_G8_B8_A8`

I didn't include BC2/DXT3, or BC7 as I don't plan to use it. However, support for this format can be trivially added on top of this change.

## No ETC2?
I did investigate ETC2 compression too which is newer yet features wide support. However, I found actual _hardware_ support for ETC2 on the desktop is much more limited that it may appear. It was a bit of a downer to see no VRAM usage reduction on my polaris RX480 when using ETC2 compressed textures. ETC2 formats also didn't seem to work without the `GL_ARB_texture_storage`/immutable storage code paths, whereas I've tested the BC1 & BC3 formats working either way. That can probably be addressed, but I didn't.

For the above reasons I decided against using ETC2 in the game & adding it here. Note though; similarly to BC2, this change will make it simpler to support ETC2 formats in future if desired.

### OpenGL Extensions
~This code requires S3TC extensions added to `gfx_gl`. The changes there are ready to go if this PR is wanted. The gfx_gl dependency line needs to be changed:~
~https://github.com/gfx-rs/gfx_gl/pull/38~
Extensions now available in gfx_gl `0.6`.

### API support
Robo Instructus is an OpenGL 3.3 game so I'm not personally interested in other API support for these formats at the moment. Also considering pre-ll code is in support mode, and unsupported formats already seem common I thought it wasn't a big deal not supporting other APIs.

### Naming
I went with Vulkan naming style ie prefix with BC1. Maybe you guys have a better idea of what you'd like to name the formats. I'm not hugely fussed here.

### Format documentation
~It would be nice to have documentation for the new formats. Mentioning there various other pseudonyms & API support will probably help people wondering what BC1 does & means. However, the `impl_formats!` macro prevents easy documentation.~ I've added this. Other formats can now be fairly easily documented too.

### DIFF
--- a/src/backend/gl/src/tex.rs
+++ b/src/backend/gl/src/tex.rs
@@ -64,9 +64,18 @@ fn format_to_glpixel(format: NewFormat) -> GLenum {
     match format.0 {
         S::R8 | S::R16 | S::R32=> r,
         S::R4_G4 | S::R8_G8 | S::R16_G16 | S::R32_G32 => rg,
-        S::R16_G16_B16 | S::R32_G32_B32 | S::R5_G6_B5 | S::R11_G11_B10 => rgb,
-        S::R8_G8_B8_A8 | S::R16_G16_B16_A16 | S::R32_G32_B32_A32 |
-        S::R4_G4_B4_A4 | S::R5_G5_B5_A1 | S::R10_G10_B10_A2 => rgba,
+        S::R16_G16_B16
+        | S::R32_G32_B32
+        | S::R5_G6_B5
+        | S::R11_G11_B10
+        | S::BC1_R8_G8_B8 => rgb,
+        S::R8_G8_B8_A8
+        | S::R16_G16_B16_A16
+        | S::R32_G32_B32_A32
+        | S::R4_G4_B4_A4
+        | S::R5_G5_B5_A1
+        | S::R10_G10_B10_A2
+        | S::BC3_R8_G8_B8_A8 => rgba,
         S::D24_S8 => gl::DEPTH_STENCIL,
         S::D16 | S::D24 | S::D32 => gl::DEPTH_COMPONENT,
         S::B8_G8_R8_A8 => bgra,
@@ -89,7 +98,12 @@ fn format_to_gltype(format: NewFormat) -> Result<GLenum, ()> {
         S::R4_G4_B4_A4 => gl::UNSIGNED_SHORT_4_4_4_4,
         S::R5_G5_B5_A1 => gl::UNSIGNED_SHORT_5_5_5_1,
         S::R5_G6_B5 => gl::UNSIGNED_SHORT_5_6_5,
-        S::B8_G8_R8_A8 | S::R8 | S::R8_G8 | S::R8_G8_B8_A8 => fm8,
+        S::B8_G8_R8_A8
+        | S::R8
+        | S::R8_G8
+        | S::R8_G8_B8_A8
+        | S::BC1_R8_G8_B8
+        | S::BC3_R8_G8_B8_A8 => fm8,
         S::R10_G10_B10_A2 => gl::UNSIGNED_INT_10_10_10_2,
         S::R11_G11_B10 => return Err(()),
         S::R16 | S::R16_G16 | S::R16_G16_B16 | S::R16_G16_B16_A16 => fm16,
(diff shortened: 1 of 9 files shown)


alexheretic:
Added some docs

alexheretic:
> gfx_gl needs to be changed as you mentioned

Yep with https://github.com/gfx-rs/gfx_gl/pull/38 merged & released I can update to refer to that gfx_gl version.

alexheretic:
Ok the extensions are in gfx_gl and documentation for the new formats is there. I don't see only supporting OpenGL as an issue, as other backends could be added later anyway.

I've been using these formats for all the textures in Robo Instructus for the last two beta updates and everything seems fine with my testers.

Let me know if you have better ideas for the format names, otherwise this is ready to go I think.

src/backend/gl/src/tex.rs
```diff
@@ -1085,3 +1195,38 @@ pub fn generate_mipmap(gl: &gl::Gl, name: Texture, target: gl::types::GLenum) {
     gl.BindTexture(target, name);
     gl.GenerateMipmap(target);
 }}
+
+trait CompressionAware {
+    fn is_compressed(&self) -> bool;
+}
+
+impl CompressionAware for core::format::SurfaceType {
```
kvark: this logic should probably be in `gfx_core` crate since it's not GL specific

alexheretic: Yeah that probably will make sense when the formats are supported in more backends. Right now this is really just an implementation detail of the gl backend, not part of the public API. I'm not sure if it even _will_ be useful in other backends, so it makes sense to leave it private for now (it's not exactly complicated either).

alexheretic:
@kvark can this be released in v0.18? I guess gfx-core will need a breaking bump, but I'm not sure if gfx "soaks" that bump or has to step with it.

kvark:
@alexheretic unfortunately, no. The format stuff is re-exported from `gfx` crate, and there is a few more surface types now in the enum...

Saying that, it would only be breaking for code that has a comprehensive match statement across the surface types, which I don't know any (for pre-ll), so in practice this would be a breaking change, especially considering that 0.18 was released relatively recent and hasn't gotten that many users.

Therefore, my weak preference would be on releasing this as a patch to 0.18 rather than the new version. @msiglreith @grovesNL @alexheretic opinions?

alexheretic:
Yes I'd have thought the only users matching on all surface types would be the gfx backends themselves. That along with the fairly recent v0.18 version I'd also favour sticking with 0.18 & yanking on complaint. 

But either way will work in the end.

alexheretic:
I've raised #2755 as a minimally breaking release proposal.

PR #2735, opened 2019-04-10
[vk] fix sporadic failures in `create_render_pass`

aleksijuvani:
We were taking pointers to `attachment_refs`, but pushing to a `Vec` causes reallocation when the capacity is exceeded (`attachment_refs` has an initial capacity of 0), which invalidates existing pointers. Hence, if we have more than a single subpass, the first subpasses would contain invalid pointers.

Solves the issue by allocating `attachment_refs` in one go before iterating over the subpasses. I also noticed that we are able to use `Box<[_]>` here instead of `Vec<_>` because we don't need to be able to grow these containers, but this is not crucial in terms of function.

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -140,55 +140,49 @@ impl d::Device<B> for Device {
             })
             .sum();
 
-        let mut attachment_refs = Vec::new();
-
-        let subpasses = subpasses
+        let attachment_refs = subpasses
             .into_iter()
             .map(|subpass| {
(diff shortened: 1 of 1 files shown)


aleksijuvani:
I don't think it's possible to infer the array size for `ArrayVec` or `SmallVec`, so we would have to specify some arbitrary number of subpasses and attachments that we're expecting that would get allocated on the stack. It's hard to guess how many parameters will get passed in so I would avoid thinking about this for now and allocate the memory dynamically until Rust gains support for VLAs.

PR #2741, opened 2019-04-12
[mtl] Use private_caps data for reported texture sizes

Frizi:
I also found out that the queried feature set was off by one version.

PR checklist:


kvark:
bors retry

PR #2742, opened 2019-04-14
[pre-ll] Update glutin -> 0.21

alexheretic:
Update to the latest quite breaking glutin api changes.

PR checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,9 @@ matrix:
     # Windows 64bit
     - os: windows
       rust: stable
-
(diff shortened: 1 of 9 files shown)


kvark:
Would you want to back-port this to 0.18 branch (or 0.17 even?) for a minor release?

alexheretic:
> Would you want to back-port this to 0.18 branch (or 0.17 even?) for a minor release?

I'm using git at the moment as I need #2733, I can't use a release without that so I'll leave the 0.18 cherry picks till later or for someone else.

PR #2745, opened 2019-04-16
[pre-ll] initialize device

### LINKED ISSUE #2727: [pre-ll] Initialize Device and Factory from existing context
Short info header:
- GFX version: `gfx 0.18.0`

I'm wondering how receptive we'd be to adding functions to create a Device and Factory from existing backend-specific contexts, similar to how the hal DX11 backend has a `new` function that takes an existing `ID3D11DeviceContext`:

https://github.com/gfx-rs/gfx/blob/fcc3806aba14e4c0f54adb51dd34057e192800d3/src/backend/dx11/src/device.rs#L68-L72

I imagine for all backends, it will be adding a backend-specific method to the Device type. For DX11 I imagine it involves just passing the immediate context in and for GL there might not need to be any work done (just need to be careful to call the creation on a gl thread?).

The use case for this is to allow non-`gfx`/`winit` entrypoints to call my `gfx` code as a library.

Cheers!

tangmi:
Fixes #2727
PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -191,6 +191,28 @@ pub fn create(driver_type: d3dcommon::D3D_DRIVER_TYPE, desc: &dxgi::DXGI_SWAP_CH
     let mut swap_chain = ptr::null_mut();
     let create_flags = 0; //D3D11_CREATE_DEVICE_DEBUG;
     let mut device = ptr::null_mut();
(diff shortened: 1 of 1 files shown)


tangmi:
This was tested in a separate C++ DX11 app with one set of initialization flags. I'm unaware if there's configurations that this can break on, but this code worked for my purposes.

PR #2747, opened 2019-04-18
[pre-ll] Support HLSL semantic indexes

### LINKED ISSUE #2748: [pre-ll] Support HLSL semantic indexes
Short info header:
- GFX version: pre-ll 0.18.0
- OS: Windows, DX11

I'm wondering how to correctly support HLSL semantics with indexes like these:

```hlsl
struct SPIRV_Cross_Input
{
    float2 in_var_POSITION : TEXCOORD0;
    float2 in_var_UV : TEXCOORD1;
};

// Vertex main
SPIRV_Cross_Output main(SPIRV_Cross_Input stage_input)
{
    // ..
}
```

I've made an attempt here: https://github.com/gfx-rs/gfx/pull/2747, but I can tell this solution is pretty sketchy.

My best guess is that `Structure<Format>::query` needs to take a `shade::AttributeVar` instead of a `&str` and the `gfx_defines!` macros should generate a `query` implementation that is aware of indexed semantics?

Thanks for your time!

tangmi:
Fixes #2748
PR checklist:

This changeset adds support for HLSL semantic indices.

Users should now be able to specify a semantic with an index (such as `"TEXCOORD1"`) in their pipeline. If one does not include the index, an index of `0` is assumed (as HLSL assumes).

Introduces a struct that can be converted to and from a packed String.

Here are some sample mappings:
- `"TEXCOORD"` -> `VertexSemantic { name: "TEXCOORD", index: 0 }` -> `"TEXCOORD0"`
- `"TEXCOORD0"` <-> `VertexSemantic { name: "TEXCOORD", index: 0 }`
- `"TEXCOORD10"` <-> `VertexSemantic { name: "TEXCOORD", index: 10 }`

On link_input, we will try linking against the semantic name and index together (e.g. `"TEXCOORD0"`) and also just the semantic name (e.g. `"TEXCOORD"`) if the semantic index is 0, in the case that the user did not explicitly provide the semantic index, esp. for GLSL compatibility).

### DIFF
--- a/src/backend/dx11/src/factory.rs
+++ b/src/backend/dx11/src/factory.rs
@@ -587,9 +587,10 @@ impl core::Factory<R> for Factory {
                     attrib.name, elem.offset);
                 return Err(core::pso::CreationError);
             }
+            let vertex_semantic: VertexSemantic = attrib.name.as_str().into();
             layouts.push(d3d11::D3D11_INPUT_ELEMENT_DESC {
                 SemanticName: &charbuf[charpos],
-                SemanticIndex: 0,
+                SemanticIndex: vertex_semantic.index,
                 Format: match map_format(elem.format, false) {
                     Some(fm) => fm,
                     None => {
@@ -597,7 +598,7 @@ impl core::Factory<R> for Factory {
                         return Err(core::pso::CreationError);
                     }
                 },
-                InputSlot: attrib.slot as _,
+                InputSlot: attrib.slot as _, // NOTE: gfx_backend_dx11 has a vertex buffer binding per attribute.
                 AlignedByteOffset: elem.offset as _,
                 InputSlotClass: if bdesc.rate == 0 {
                     d3d11::D3D11_INPUT_PER_VERTEX_DATA
@@ -606,7 +607,7 @@ impl core::Factory<R> for Factory {
                 },
                 InstanceDataStepRate: bdesc.rate as _,
             });
-            for (out, inp) in charbuf[charpos..].iter_mut().zip(attrib.name.as_bytes().iter()) {
+            for (out, inp) in charbuf[charpos..].iter_mut().zip(vertex_semantic.name.as_bytes().iter()) {
                 *out = *inp as i8;
             }
             charpos += attrib.name.as_bytes().len() + 1;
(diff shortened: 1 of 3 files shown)


src/backend/dx11/src/factory.rs
```diff
@@ -589,7 +589,7 @@ impl core::Factory<R> for Factory {
             }
             layouts.push(d3d11::D3D11_INPUT_ELEMENT_DESC {
                 SemanticName: &charbuf[charpos],
-                SemanticIndex: 0,
+                SemanticIndex: u32::from(attrib.slot),
```
kvark: why isn't this just `attrib.slot as u32` or something like this?

msiglreith:
Just to be sure, this change is currently backwards incompatible and would require adjusting all the HLSL shaders, right?

src/backend/dx11/src/factory.rs
tangmi: Clippy has yelled at me in the past about this: https://rust-lang.github.io/rust-clippy/master/#cast_lossless. Though I'm happy to change it to `as u32` if that's preferred.

tangmi:
I've done some work to this branch, as my previous attempt made it require a vertex attribute have a semantic index equal to the attribute slot assigned to it.

Ideally this change should be backwards compatible, but in the latest revision I *believe* I have fixed a bug in the DX backend regarding multiple vertex buffers in a single pipeline (though I don't have a test for this yet).

Here's a bit of a brain dump:

- `AttributeVar`'s `slot` doesn't seem to do anything on directx (I am unaware of being able to explicitly set registers on individual vertex attributes is possible?)
- `attrib.slot` was previously the value set for the `InputSlot`, but that's a field meant for the buffer binding index when there are multiple vertex buffers (https://docs.microsoft.com/en-us/windows/desktop/api/d3d11/ns-d3d11-d3d11_input_element_desc)
  - I'm unsure how this worked in the past
- My current solution involves adding a "semantic_index" field on `AttributeVar` (that is set to 0 on all other platforms)
  - I tried to reuse `attrib.slot` for the semantic index, but it looks like that value is used as the index for some arrays in the DX impl (and setting it to existing values on multiple attributes causes indices in those arrays to be overridden)
- I don't like that my solution for `link_input` is "try the semantic, then try the semantic with the index, then fail"

src/backend/dx11/src/factory.rs
```diff
@@ -630,7 +630,33 @@ impl core::Factory<R> for Factory {
                 &mut vertex_layout)
         };
         if !winerror::SUCCEEDED(hr) {
-            error!("Failed to create input layout from ?, error {:x}"/*, layouts*/, hr);
+            struct DebugInputLayout<'a> {
```
tangmi: I can remove this if it's unsightly.

I'm working on https://github.com/retep998/winapi-rs/pull/760, which would additionally make this unnecessary.

kvark:
@tangmi thank you for the investigation! It would take me a bit of time to remember the internals of pre-ll DX11 backend. Overall, don't put too much trust it in - it was likely rolled out with the examples being the only test case. Saying that, the instancing example has multiple vertex buffers, which according to your findings shouldn't be working correctly, but it does.

tangmi:
Unfortunately I think the instancing example just happens to work. If I reorder the vertex attributes in the shader, the example stops displaying output.

Here are some observations:
- It looks like DX's input element's input slot is clamped to the range of bound vertex buffers
- Gfx reads the AttributeVar::slot from the order of the HLSL reflection data

So if the following vertex buffers are bound in the instance example
```
vertex buffer 0 => position data
vertex buffer 1 => translate and color instance data
```

The HLSL has a vertex defined as
```
struct VsInput {
	float2 pos: a_Position; // <-- will be assigned AttributeVar::slot = 0
	float2 trans: a_Translate; // <-- will be assigned AttributeVar::slot = 1
	uint color: a_Color; // <-- will be assigned AttributeVar::slot = 2
};
```

Then:
- `a_Position`'s InputSlot will get set to 0 (vertex data)
- `a_Translate`'s InputSlot will get set to 1 (instance data)
- `a_Color`'s InputSlot will get set to 2, but clamped to 1 (instance data)

And the example works

Reordering the vertex attributes in the shader to something like:
```
struct VsInput {
	float2 trans: a_Translate; // <-- will be assigned AttributeVar::slot = 0
	float2 pos: a_Position; // <-- will be assigned AttributeVar::slot = 1
	uint color: a_Color; // <-- will be assigned AttributeVar::slot = 2
};
```

Then:
- `a_Translate`'s InputSlot will get set to 0 (vertex data) <-- incorrectly set
- `a_Position`'s InputSlot will get set to 1 (instance data) <-- incorrectly set
- `a_Color`'s InputSlot will get set to 2, but clamped to 1 (instance data)

tangmi:
@kvark I think this is ready to review. All the examples run without modification (excluding `render_target`, which is missing HLSL shaders).

My reservations with these changes are:
- Adding a DX-only `semantic_index` field to `AttributeVar` is hacky
- `link_input` could be implemented better (i.e. `T::query` could generate code to look for semantic indices instead of `link_input` trying both, but idk how well this would work for code that manually implements `query`)

kvark:
Thank you for evolving the PR into shippable state! Could you please rebase instead of introducing the merge commit?
Also, does @msiglreith want to have a look before we proceed?

msiglreith:
So, the semantic name can basically be chosen randomly, right? In case of SPIRV-Cross outputted shaders, user need to be aware of the ordering of the vertex input. They need to match the declaration on the host side. Does my understanding match with the code?

tangmi:
The semantic name still needs to be set correctly, unfortunately. Since HLSL can't bind attributes by index, SPIRV-Cross maps the binding location to index of a consistent, but arbitrary, semantic: `TEXCOORD`.

As an aside, I believe any semantic could've been chosen, but I think people were mostly already using `TEXCOORD` for packing custom data in DX9/10. In DX11+, I think any semantic that doesn't conflict with the `SV_` ones would work.

tangmi:
@kvark This should be rebased and ready for re-review! Please let me know if there's anything else I should do.

src/core/src/shade.rs
```diff
@@ -360,6 +360,8 @@ impl From<Stage> for Usage {
 pub struct AttributeVar {
     /// Name of this attribute.
     pub name: String,
+    /// Hack: Semantic index used by DirectX. Should be set to 0 in all other backends.
+    pub semantic_index: u32,
```
kvark: this is a bit unfortunate - would require a new minor release for the whole project. Is there a way for us to only limit the changes to dx11 backend?

kvark: ping @tangmi

tangmi: Hi! Sorry for the late reply!

The only thing I can think of is making `AttributeVar` parameterized over a "name" type that dx11 could specialize as `(String, u32)` and other backends would use `String`. This would only ensure that the size/layout of the data doesn't change for the other backends. I'm not sure if this is sufficient to avoid a minor release for whole project?

`DataLink::link_input` for `VertexBufferCommon` needs the semantic index, but it only gets `AttributeVar`. `Factory::create_pipeline_state_raw` also needs the semantic index, and `AttributeVar` seems like the only thing that makes sense to hold it.

kvark: Sorry, I'm still trying to figure out what the best solution here is. So here is another question: would it be feasible for your case to require the user to *only* use semantic index == 0 in their shaders? It seems to me that it would be compatible with the current code. What would be the downside of this path?

tangmi: I'm not sure I understand the question fully, sorry!

Here's an alternate implementation that might be better: https://github.com/tangmi/gfx/pull/5 It doesn't change `AttributeVar` at the cost of very slightly higher cost marshaling in/out of that struct for the DX11 backend and slightly more complex vertex attribute handling on the DX11 backend.

tangmi:
@kvark I've rebased the new changes into this branch and updated the PR comment at the top. There *should* be no actual changes from tangmi#5, but please sanity check me if you think it's a good idea!

PR #2751, opened 2019-04-20
[mtl] add multisampling basics

aleksijuvani:
Sets multisampling state for `MTLRenderPipelineDescriptor` and maps the `Image::Usage::INPUT_ATTACHMENT` flag to `MtlTextureUsage::ShaderRead`. This is enough to implement multisampling if you resolve the samples yourself using a fragment shader on a fullscreen quad in a separate subpass. Doesn't implement resolve attachments, but it looks like that could be added relatively easily in the future.

PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -1462,6 +1462,14 @@ impl hal::Device<Backend> for Device {
             .depth_stencil_states
             .prepare(&pipeline_desc.depth_stencil, &*device);
 
+        if let Some(multisampling) = &pipeline_desc.multisampling {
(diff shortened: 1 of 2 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1462,6 +1462,14 @@ impl hal::Device<Backend> for Device {
             .depth_stencil_states
             .prepare(&pipeline_desc.depth_stencil, &*device);
 
+        if let Some(multisampling) = &pipeline_desc.multisampling {
```
kvark: taking a reference isn't needed here. Instead, should be `ref multisampling`

PR #2760, opened 2019-04-25
[pre-ll] DX11 add ID3D11InfoQueue in debug builds

tangmi:
Fixes #issue

PR checklist:
- [x] tested examples with the following backends: dx11

This adds debug output from the ID3D11InfoQueue (e.g. it'll send a message whenever a method is called incorrectly and sometimes call out which arguments are wrong). This code could also be adapted for `gfx-hal`'s dx11 backend to aid in the development experience there.


kvark:
Thank you, this looks neat!
I wonder if those translations can be auto-generated though, they are pretty mechanical...
Could you rebase (instead of introducing a merge commit) please?

tangmi:
@kvark I've rebased this and should be ready for re-review!

tangmi:
Thanks!

Re: to this comment, @kvark 

>  I wonder if those translations can be auto-generated though, they are pretty mechanical...

I toyed around with writing a macro to help a little, but didn't go for it:

- The macro I wrote added complexity
- These consts likely won't change between DirectX versions (although some might be added?)
- An ideal solution would be for `winapi` to expose some list containing each of the values for each enum. (it currently flattens an e.g. declared `D3D_MESSAGE_ID` enum to many const declarations).

PR #2765, opened 2019-04-30
Strongly typed specialization constants API

### LINKED ISSUE #2382: Refine specialization constants API
#2381 makes them just ranges into raw `[u8]` data. This isn't very convenient.

One way to improve it could be to carry the data inside the specialization constant instead of storing it separately:
```rust
pub struct Specialization<V: Borrow<[u8]>> {
    /// Constant identifier in shader source.
    pub id: u32,
    /// Value to override specialization constant.
    pub value: V,
}
```
I made such an attempt in [this branch](https://github.com/kvark/gfx/commit/ca1e4bfa5a6e91174d9a041a83e79356a7ceccf7) but it appears to be adding a level of indirection that isn't very convenient for either portability or Vulkan backend.

Another direction of improvement would be adding a trait that allows conversion of primitive types (like `f32` or `u32`) to a specialization constant, because otherwise using them directly is very inconvenient.

omni-viral:
Attempt to bring type safety into specialization constants API.

No backend/examples change is required except types renaming.

Fixes #2382

Fixes #issue
PR checklist:

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -541,10 +541,7 @@ fn main() {
                 pso::EntryPoint {
                     entry: ENTRY_NAME,
                     module: &vs_module,
-                    specialization: pso::Specialization {
-                        constants: &[pso::SpecializationConstant { id: 0, range: 0..4 }],
-                        data: unsafe { std::mem::transmute::<&f32, &[u8; 4]>(&0.8f32) },
-                    },
+                    specialization: hal::spec_const_list![0.8f32],
                 },
                 pso::EntryPoint {
                     entry: ENTRY_NAME,
--- a/src/hal/src/pso/mod.rs
+++ b/src/hal/src/pso/mod.rs
@@ -2,9 +2,8 @@
 //!
 //! This module contains items used to create and manage Pipelines.
 
-use std::fmt;
-use std::ops::Range;
-use crate::{device, pass};
+use crate::{device, pass, Backend};
+use std::{borrow::Cow, fmt, ops::Range};
 
 mod compute;
 mod descriptor;
@@ -12,13 +11,7 @@ mod graphics;
 mod input_assembler;
 mod output_merger;
 
-pub use self::compute::*;
-pub use self::descriptor::*;
-pub use self::graphics::*;
-pub use self::input_assembler::*;
-pub use self::output_merger::*;
-
-use crate::Backend;
+pub use self::{compute::*, descriptor::*, graphics::*, input_assembler::*, output_merger::*};
 
 /// Error types happening upon PSO creation on the device side.
 #[derive(Clone, Debug, PartialEq, Fail)]
@@ -151,7 +144,7 @@ impl fmt::Display for Stage {
 }
 
 /// Shader entry point.
-#[derive(Debug, Copy)]
+#[derive(Debug)]
 pub struct EntryPoint<'a, B: Backend> {
     /// Entry point name.
     pub entry: &'a str,
@@ -166,7 +159,7 @@ impl<'a, B: Backend> Clone for EntryPoint<'a, B> {
         EntryPoint {
             entry: self.entry,
             module: self.module,
-            specialization: self.specialization,
+            specialization: self.specialization.clone(),
         }
     }
 }
@@ -221,29 +214,85 @@ pub struct SpecializationConstant {
 }
 
 /// Specialization information structure.
-#[derive(Debug, Copy)]
+#[derive(Debug, Clone)]
 pub struct Specialization<'a> {
     /// Constant array.
-    pub constants: &'a [SpecializationConstant],
+    pub constants: Cow<'a, [SpecializationConstant]>,
     /// Raw data.
-    pub data: &'a [u8],
+    pub data: Cow<'a, [u8]>,
 }
 
-impl<'a> Default for Specialization<'a> {
+impl Specialization<'_> {
+    /// Empty specialization instance.
+    pub const EMPTY: Self = Specialization {
+        constants: Cow::Borrowed(&[]),
+        data: Cow::Borrowed(&[]),
+    };
+}
+
+impl Default for Specialization<'_> {
     fn default() -> Self {
+        Specialization::EMPTY
+    }
+}
+
+#[doc(hidden)]
+#[derive(Default)]
+pub struct SpecializationStorage {
+    constants: Vec<SpecializationConstant>,
+    data: Vec<u8>,
+}
+
+/// List of specialization constants.
+#[doc(hidden)]
+pub trait SpecConstList: Sized {
+    fn fold(self, storage: &mut SpecializationStorage);
+}
+
+impl<T> From<T> for Specialization<'_>
+where
+    T: SpecConstList,
+{
+    fn from(list: T) -> Self {
+        let mut storage = SpecializationStorage::default();
+        list.fold(&mut storage);
         Specialization {
-            constants: &[],
-            data: &[],
+            data: Cow::Owned(storage.data),
+            constants: Cow::Owned(storage.constants),
         }
     }
 }
 
-impl<'a> Clone for Specialization<'a> {
-    fn clone(&self) -> Self {
-        Specialization {
-            constants: self.constants,
-            data: self.data,
-        }
+#[doc(hidden)]
+pub struct SpecConstListNil;
+
+#[doc(hidden)]
+pub struct SpecConstListCons<H, T> {
+    pub head: (u32, H),
+    pub tail: T,
+}
+
+impl SpecConstList for SpecConstListNil {
+    fn fold(self, _storage: &mut SpecializationStorage) {}
+}
+
+impl<H, T> SpecConstList for SpecConstListCons<H, T>
+where
+    T: SpecConstList,
+{
+    fn fold(self, storage: &mut SpecializationStorage) {
+        let size = std::mem::size_of::<H>();
+        assert!(storage.data.len() + size <= u16::max_value() as usize);
+        let offset = storage.data.len() as u16;
+        storage.data.extend_from_slice(unsafe {
+            // Inspecting bytes is always safe.
+            std::slice::from_raw_parts(&self.head.1 as *const H as *const u8, size)
+        });
+        storage.constants.push(SpecializationConstant {
+            id: self.head.0,
+            range: offset..offset + size as u16,
+        });
+        self.tail.fold(storage)
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/hal/src/pso/mod.rs
```diff
@@ -151,26 +151,52 @@ impl fmt::Display for Stage {
...
+}
+
+impl<'a, B, S> EntryPoint<'a, B, S>
+where
+    B: Backend,
+    S: Specialization<'a>,
+{
+    /// Convert `specialization` of the `EntryPoint` into raw type
+    /// acceptable by pipeline creation functions.
+    pub fn into_raw(&self, storage: &'a mut SpecializationStorage) -> EntryPoint<'a, B> {
```
kvark: this method should follow rust conventions. `into_` is supposed to be moving out of `self`

src/hal/src/pso/mod.rs
```diff
@@ -151,26 +151,52 @@ impl fmt::Display for Stage {
...
+        EntryPoint {
+            entry: self.entry,
+            module: self.module,
+            specialization: self.specialization.into_raw(storage),
+        }
+    }
 }
 
-impl<'a, B: Backend> Clone for EntryPoint<'a, B> {
+impl<'a, B, S> Clone for EntryPoint<'a, B, S>
```
kvark: Given that `Backend` already implements `Clone`, we should just put `#[derive(Clone)]` on the struct

src/hal/src/pso/mod.rs
```diff
@@ -151,26 +151,52 @@ impl fmt::Display for Stage {
...
         EntryPoint {
             entry: self.entry,
             module: self.module,
-            specialization: self.specialization,
+            specialization: self.specialization.clone(),
         }
     }
 }
 
+impl<'a, B, S> Copy for EntryPoint<'a, B, S>
```
kvark: why do you need `Copy`?

src/hal/src/pso/mod.rs
```diff
@@ -213,40 +239,77 @@ pub enum BasePipeline<'a, P: 'a> {
...
+/// Temporary storage for specialization constants.
+pub struct SpecializationStorage {
+    data: Vec<u8>,
+    constants: Vec<RawSpecializationConstant>,
+}
+
+/// Convert typed specialization into untyped one.
+pub trait Specialization<'a> {
+    /// Build `Specialization` value.
+    fn into_raw(&self, storage: &'a mut SpecializationStorage) -> RawSpecialization<'a>;
```
kvark: what role does `SpecializationStorage` play here?

src/hal/src/pso/mod.rs
```diff
@@ -213,40 +239,77 @@ pub enum BasePipeline<'a, P: 'a> {
...
+impl<'a> Specialization<'a> for frunk_core::hlist::HNil {
+    fn into_raw(&self, storage: &'a mut SpecializationStorage) -> RawSpecialization<'a> {
+        RawSpecialization {
+            data: &storage.data,
+            constants: &storage.constants,
         }
     }
 }
 
+impl<'a, H, T> Specialization<'a> for frunk_core::hlist::HCons<(u32, H), T>
```
kvark: can we make `frunk_core` to be an optional dependency?

src/hal/src/pso/mod.rs
```diff
@@ -213,40 +239,77 @@ pub enum BasePipeline<'a, P: 'a> {
...
+where
+    T: Specialization<'a>,
+{
+    fn into_raw(&self, storage: &'a mut SpecializationStorage) -> RawSpecialization<'a> {
```
kvark: It's not clear to me how this is going to be used. Could you provide a doc example here?

src/hal/src/pso/mod.rs
omni-viral: It does not work that way :(
It requires `B::ShaderModule: Clone` for some reason.

src/hal/src/pso/mod.rs
omni-viral: I can `Copy` that's enough reason for impl

src/hal/src/pso/mod.rs
omni-viral: It owns what `RawSpecialization`, borrows.

src/hal/src/pso/mod.rs
omni-viral: Yes.

src/hal/src/pso/mod.rs
omni-viral: User will use this or `into_raw` on `EntryPoint` to convert typed into untyped. Untyped is required to create pipeline.

src/hal/src/pso/mod.rs
omni-viral: Okey

src/hal/src/pso/mod.rs
kvark: I respectfully disagree. If the payload is big-ish, having the programmer to call `clone` can at least make the costs more explicit. Not all copies are good.

src/hal/src/pso/mod.rs
kvark: I can't reproduce this in playground. This code compiles:
```rust
trait Foo: Clone {
    type Bar;
}

#[derive(Clone)]
struct Tar<'a, F: Foo>(&'a F::Bar);
```

omni-viral: But does it implement `Clone`?

src/hal/src/pso/mod.rs
omni-viral: I agree. Yet if `S` is big-ish it shall not implement `Copy` and then `EntryPoint` won't too.
And if `S` is small `EntryPoint` is small too.

examples/quad/main.rs
```diff
@@ -541,24 +541,21 @@ fn main() {
                 pso::EntryPoint {
                     entry: ENTRY_NAME,
                     module: &vs_module,
-                    specialization: pso::Specialization {
-                        constants: &[pso::SpecializationConstant { id: 0, range: 0..4 }],
-                        data: unsafe { std::mem::transmute::<&f32, &[u8; 4]>(&0.8f32) },
-                    },
+                    specialization: hal::spec_const_list![0.8f32],
```
kvark: wow, fancy absolute address for a macro! I like it :)

src/hal/src/lib.rs
```diff
@@ -15,6 +15,10 @@ extern crate mint;
 #[macro_use]
 extern crate serde;
 
+#[cfg(feature = "hlist-spec-const")]
+#[doc(hidden)]
+pub use frunk_core;
```
kvark: does it need to be pub?

src/hal/src/pso/mod.rs
```diff
@@ -151,26 +151,53 @@ impl fmt::Display for Stage {
...
 
-impl<'a, B: Backend> Clone for EntryPoint<'a, B> {
+impl<'a, B, S> EntryPoint<'a, B, S>
+where
+    B: Backend,
+    S: Specialization,
+{
+    /// Convert `specialization` of the `EntryPoint` into raw type
+    /// acceptable by pipeline creation functions.
+    pub fn as_raw(&'a self) -> EntryPoint<'a, B> {
```
kvark: The convention of `as_something()` in Rust is to borrow and return a reference. In this case, let's have `to_raw()`

src/hal/src/pso/mod.rs
```diff
@@ -221,32 +248,112 @@ pub struct SpecializationConstant {
...
     /// Constant array.
     pub constants: &'a [SpecializationConstant],
     /// Raw data.
     pub data: &'a [u8],
 }
 
-impl<'a> Default for Specialization<'a> {
+impl RawSpecialization<'_> {
+    /// Create empty specialization instance.
+    pub fn new() -> Self {
```
kvark: do we still need this?

src/hal/src/pso/mod.rs
```diff
@@ -221,32 +248,112 @@ pub struct SpecializationConstant {
...
+pub trait Specialization {
+    /// Build `Specialization` value.
+    fn as_raw(&self) -> RawSpecialization<'_>;
+}
+
+/// Temporary storage for specialization constants.
+#[derive(Clone, Debug, Default)]
+pub struct SpecializationStorage {
+    /// Actual data for specialization constants.
+    pub data: Vec<u8>,
```
kvark: do we need to expose everything here to the user? I thought this is built with helper methods anyway

src/hal/src/pso/mod.rs
```diff
@@ -221,32 +248,112 @@ pub struct SpecializationConstant {
...
+        RawSpecialization {
+            data: &self.data,
+            constants: &self.constants,
         }
     }
 }
 
+/// Empty `Specialization` implementation.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Unspecialized;
```
kvark: can this just be something like `RawSpecialization::EMPTY`?

src/hal/src/pso/mod.rs
```diff
@@ -221,32 +248,112 @@ pub struct SpecializationConstant {
...
+    }
+}
+
+#[cfg(feature = "hlist-spec-const")]
+impl SpecializationList for frunk_core::hlist::HNil {
+    fn fold(&self, _storage: &mut SpecializationStorage) {}
+}
+
+#[cfg(feature = "hlist-spec-const")]
+impl<H, T> SpecializationList for frunk_core::hlist::HCons<(u32, H), T>
```
kvark: why don't we `use frunk_core::hlist` to begin with?

src/hal/src/lib.rs
omni-viral: No. And I'm getting rid of the dependency anyway ;)

src/hal/src/pso/mod.rs
omni-viral: Okey. But this changes to the way it was originally.

src/hal/src/pso/mod.rs
omni-viral: I don't really like types without `new` without reason like impossibility to implement.

src/hal/src/pso/mod.rs
omni-viral: What if user would want to make his own `Specialization` trait implementation?
This type get merged with original `Specialization` anyway though.

src/hal/src/pso/mod.rs
omni-viral: I forgot associated constants exists. How could I?

src/hal/src/pso/mod.rs
omni-viral: Why do we need fancy `frunk_core::hlist` to begin with?

PR #2771, opened 2019-05-15
Expand docs and add some Default impl's for image

icefoxen:
Fixes https://github.com/amethyst/rendy/issues/117
PR checklist:

### DIFF
--- a/src/hal/src/image.rs
+++ b/src/hal/src/image.rs
@@ -19,6 +19,8 @@ pub type Layer = u16;
 pub type Level = u8;
 /// Maximum accessible mipmap level of an image.
 pub const MAX_LEVEL: Level = 15;
+/// A texel coordinate in an image.
+pub type TexelCoordinate = i32;
 
 /// Describes the size of an image, which may be up to three dimensional.
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


src/hal/src/image.rs
```diff
@@ -56,15 +56,20 @@ impl Extent {
...
+/// copy operations.  All offsets are in texels, and
+/// specifying offsets other than 0 for dimensions
+/// that do not exist is undefined behavior -- for
+/// example, specifying a `z` offset of `1` in a
+/// two-dimensional image.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Offset {
-    ///
+    /// X offset, in texels
```
kvark: I think we can have a `type TexelCoordinate = i32` to reflect this

src/hal/src/image.rs
```diff
@@ -208,6 +213,13 @@ pub enum Filter {
     Linear,
 }
 
+impl Default for Filter {
```
kvark: why do we want a default filter here?

src/hal/src/image.rs
```diff
@@ -469,7 +490,10 @@ impl Into<[f32; 4]> for PackedColor {
     }
 }
 
-/// Specifies how to sample from an image.
+/// Specifies how to sample from an image.  This is all the parameters
```
kvark: This is -> These are

src/hal/src/image.rs
icefoxen: Just to impl `Default` for `SamplerInfo`, which is something I wanted recently to make basic hello-world style setup just a little simpler.

src/hal/src/image.rs
icefoxen: :+1:

icefoxen:
I recently wanted the defaults for writing something that was basically a hello-world program, where I was saying "okay I don't care what values these are yet, I just want them to be something vaguely sensible".  Up to you guys whether or not you actually want them.

src/hal/src/image.rs
icefoxen: :+1:

kvark:
Problem is, in gfx-hal the goal of reducing user's boilerplate is quite low priority. Cleanness and soundness of the semantics matter most. I would suggest leaving the opinionated defaults to other wrapping higher level libraries (ggez?).

icefoxen:
> Problem is, in gfx-hal the goal of reducing user's boilerplate is quite low priority.

That's why I'm doing it for you.  :-P  But, if you want to leave those sorts of things out on purpose, just confirm it and I'll update the PR's to remove the `Default` impl's.

kvark:
Yes, please. I confirm we don't want gfx-hal to be opinionated about what the user wants. Defaults are accepted when they are for disabling an optional function block, i.e. default for a depth test is OFF, default for any option<> is fine to have None, etc. This is not the case for `SamplerState`.

icefoxen:
Makes sense.  Done!

kvark:
Thank you! Please squash a bit (adding + removing something shouldn't be in history within the same PR)

icefoxen:
Done.

PR #2776, opened 2019-05-19
Changes to support wgpu gl backend

kyren:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -176,14 +176,14 @@ impl Device {
             },
             n::ImageView::Texture((textype, texture), level) => unsafe {
                 gl.BindTexture(textype, texture);
-                gl.FramebufferTexture2D(point, attachment, gl::TEXTURE_2D, texture, level as _);
+                gl.FramebufferTexture2D(point, attachment, textype, texture, level as _);
             },
             n::ImageView::TextureLayer((textype, texture), level, layer) => unsafe {
                 gl.BindTexture(textype, texture);
                 gl.FramebufferTexture3D(
                     point,
                     attachment,
-                    gl::TEXTURE_2D,
+                    textype,
                     texture,
                     level as _,
                     layer as _,
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1595,8 +1595,23 @@ impl d::Device<B> for Device {
         }
     }
 
-    unsafe fn get_fence_status(&self, _: &n::Fence) -> Result<bool, d::DeviceLost> {
-        Ok(true)
+    unsafe fn get_fence_status(&self, fence: &n::Fence) -> Result<bool, d::DeviceLost> {
+        let gl = &self.share.context;
+
+        let mut len = 0;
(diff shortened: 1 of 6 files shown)


kyren:
It's very possible I've done some of this in the wrong way, especially the gl texture type change.  I'm happy to re-work anything that is done in the wrong way.

kyren:
> I don't think the change regarding fences is correct. Take a look at the documentation of glFenceSync for fences in OpenGL and glGetSynciv for querying the status of a fence in particular.

Yeah, you're absolutely right.  It would have been obvious that they should have been implemented had I looked at any of the other gl fence methods as well.  I hope the implementation is correct now.

src/backend/gl/src/device.rs
```diff
@@ -1548,8 +1595,23 @@ impl d::Device<B> for Device {
...
+    unsafe fn get_fence_status(&self, fence: &n::Fence) -> Result<bool, d::DeviceLost> {
+        let gl = &self.share.context;
+
+        let mut len = 0;
+        let mut values = [0];
+        gl.GetSynciv(fence.0.get(), gl::SYNC_STATUS, values.len() as i32, &mut len, values.as_mut_ptr());
+
+        if len == 0 {
+            if let Err(err) = self.share.check() {
+                error!("Error getting fence status: {:?}", err);
```
kvark: nit: could just be `error!("Error getting fence status: {:?}", self.share.check());`

src/backend/gl/src/native.rs
```diff
@@ -13,7 +13,7 @@ pub type Shader = gl::types::GLuint;
 pub type Program = gl::types::GLuint;
 pub type FrameBuffer = gl::types::GLuint;
 pub type Surface = gl::types::GLuint;
-pub type Texture = gl::types::GLuint;
+pub type Texture = (gl::types::GLenum, gl::types::GLuint);
```
kvark: There is something wrong about this. The first piece of that tuple - a type of the texture - should be associated with the texture itself. I.e. be a member of the `ImageLike::Texture` variant. It would be cleaner than pretending that the type is a part of the identifier.

kyren: So just to be clear, you're suggesting that the gl type be part of the gl backend's `ImageView::Texture` variant?  If so, that makes perfect sense to me, and shouldn't be very difficult.

kyren: It was slightly more involved than I thought, but it still wasn't bad.

src/backend/gl/src/device.rs
```diff
@@ -812,8 +812,7 @@ impl d::Device<B> for Device {
                 );
                 assert!(uniform_max_size > 0);
 
-                let mut name = Vec::with_capacity(uniform_max_size as usize);
-                name.set_len(uniform_max_size as usize);
+                let mut name = vec!['\0'; uniform_max_size as usize];
                 name[uniform_max_size as usize - 1usize] = '\0';
```
kvark: this line is redundant now

kyren: Sorry, good catch!

PR #2782, opened 2019-05-23
android only supports a subset of the standard validation layers

mtak-:
Fixes https://github.com/gfx-rs/gfx/issues/1805 I think.

The android ndk only ships with five validation layers and does not support `VK_LAYER_LUNARG_standard_validation`. This PR only enables the validation layers android supports.

```
VK_LAYER_LUNARG_core_validation
VK_LAYER_LUNARG_object_tracker
VK_LAYER_LUNARG_parameter_validation
VK_LAYER_GOOGLE_threading
VK_LAYER_GOOGLE_unique_objects
```

Additionally, the vulkan validation layers repo is only setup to build those 5 validation layers on android [source](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/master/build-android/jni/Android.mk).

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -59,18 +59,18 @@ mod window;
 lazy_static! {
     static ref LAYERS: Vec<&'static CStr> = if cfg!(all(target_os = "android", debug_assertions)) {
         vec![
-            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_core_validation\0").expect("Wrong extension string"),
-            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_object_tracker\0").expect("Wrong extension string"),
-            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_parameter_validation\0").expect("Wrong extension string"),
-            CStr::from_bytes_with_nul(b"VK_LAYER_GOOGLE_threading\0").expect("Wrong extension string"),
-            CStr::from_bytes_with_nul(b"VK_LAYER_GOOGLE_unique_objects\0").expect("Wrong extension string"),
+            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_core_validation\0").unwrap(),
+            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_object_tracker\0").unwrap(),
+            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_parameter_validation\0").unwrap(),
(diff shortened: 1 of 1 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -57,7 +57,19 @@ mod window;
...
+    static ref LAYERS: Vec<&'static CStr> = if cfg!(all(target_os = "android", debug_assertions)) {
+        vec![
+            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_core_validation\0").expect("Wrong extension string"),
+            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_object_tracker\0").expect("Wrong extension string"),
+            CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_parameter_validation\0").expect("Wrong extension string"),
+            CStr::from_bytes_with_nul(b"VK_LAYER_GOOGLE_threading\0").expect("Wrong extension string"),
+            CStr::from_bytes_with_nul(b"VK_LAYER_GOOGLE_unique_objects\0").expect("Wrong extension string"),
+        ]
+    } else if cfg!(debug_assertions) {
+        vec![CStr::from_bytes_with_nul(b"VK_LAYER_LUNARG_standard_validation\0").expect("Wrong extension string")]
```
kvark: let's just `unwrap` or `from_bytes_with_nul_unchecked`. There is not much use from those `expect` messages

mtak-: Replaced with unwrap

kvark:
@mtak- can you confirm if Android now successfully runs the examples?

mtak-:
I have no idea how @torkleyy was running the quad example. `cargo apk` isn't working for me, and it's a nontrivial amount of effort to setup an android rust project manually.

I don't expect the examples to work even after getting a project setup due to the dependence on `glsl_to_spirv` which attempts to run an executable (something only rooted android phones can do AFAIK).

However, I can confirm that validation is working fine in my own project on android. Before this change, I was seeing similar errors to #1805.

Now I have a validation error :), but it appears to not be a big deal, as the game still renders.

```
2019-05-24 13:10:25.585 7587-7620/com.our.package D/RustAndroidGlueStdouterr: DS(ERROR): object: 0x0 type: 0 location: 2953 msgCode: 6: Layout for color attachment is VK_IMAGE_LAYOUT_PRESENT_SRC_KHR but can only be COLOR_ATTACHMENT_OPTIMAL or GENERAL.
```

mtak-:
device-2019-05-24-140919

PR #2787, opened 2019-05-31
[hal] Add deny(missing_debug_implementations) to enforce debug printers

### LINKED ISSUE #2560: Add a Debug impl to everything
Every single type within `gfx-hal`, and all backends, should have a `Debug` impl available in the `0.2` release. This is not a controversial idea.

However, I would add that while we're setting it up we might want to make an effort to have them be _optional_ Debug impls, gated by a feature flag (probably one that is _on_ by default). Why? When there's a lot of types it ends up taking build time just to generate the Debug code. Even if the user doesn't call the Debug impl, it still gets generated and then thrown away. It's maybe not the _biggest_ deal, but it's worth consideration. This isn't totally crazy, `winapi` already does it.

Right now most types don't have Debug, and most types _might_ have content that isn't friendly to simply deriving Debug. That's fine, but we want them to have Debug anyway so that end users can derive a Debug even if they're holding a `gfx-hal` type in their struct. It's just the polite thing to do. What's needed is a "dummy" impl that just prints the type's name and a note that other info has been cut. This is a trivial macro_rules to write:

```rust
macro_rules! generate_dummy_Debug_for {
  ($t:ty) => {
    impl core::fmt::Debug for $t {
      /// This only prints the type name, with no field data
      fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
        write!(f, concat!(stringify!($t), " {{ fields omitted }} "))
      }
    }
  };
}
```

BourgondAries:
Fixes #2560
PR checklist:

### DIFF
--- a/src/hal/src/backend.rs
+++ b/src/hal/src/backend.rs
@@ -11,6 +11,7 @@ use fxhash::FxHasher;
 /// Bare-metal queue group.
 ///
 /// Denotes all queues created from one queue family.
+#[derive(Debug)]
 pub struct RawQueueGroup<B: Backend> {
     pub family: B::QueueFamily,
     pub queues: Vec<B::CommandQueue>,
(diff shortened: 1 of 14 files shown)


BourgondAries:
Note: one of the debug implementations is for a `union` and just prints out the name of the union. I'm not sure if it's desirable to transmute it and print its contents.

Issue #2749, opened 2019-04-18
Make OpenGL backend abstract


skyne98:
Hello there! 

At the moment `gfx-backend-gl` seems to be very closely related to `glutin`, which makes providing different OpenGL backends for it very hard (if not even impossible). Coming from https://github.com/Smithay/smithay/issues/129 there is an idea to abstract away the need for `glutin` behind the trait making it possible for other projects to utilize gfx-hal while being in their own specific contexts (as described by https://github.com/Smithay/smithay/issues/129#issuecomment-484464241 and https://github.com/Smithay/smithay/issues/129#issuecomment-459541840).

At the moment this will allow Smithay to implement such a backend creating an entrypoint into the world of gfx-hal rendering (e.g. rendy, amethyst, ggez, gfx-webrender) while the other, more complex backends, are in the works (e.g. vulkan). I personally think that Smithay can become a great foundation for people willing to make not only fast and pretty desktop experiences, but also games and software that doesn't rely on compositors.

Thank you for your time!

msiglreith:
Hi 👋 

that sounds like an interesting project! It's a bit difficult for me to grasp the scope of `smithay` atm. Therefore, I would quickly give an overview about the window system interface which would be the connection between `gfx` and `smithay` I believe. This mostly follows https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/window/glutin.rs

Ideally, we would have an headless context acting as `Instance` which allows us to query basic information about the system (memory, capabilities, ..), additionally a `Surface` acting as software abstraction of the OS/Hardware/Display/Compositor(?) providing information about these. The surface can be then used to instantiate a `Swapchain` which will do the actual presentation handling (providing images to draw to, compositing, displaying stuff on the monitor, ..).

From the API side the best case would be that only the surface creation will be platform dependent and all the other parts completely opaque and platform agnostic. Meaning, some kind of `create_surface_from_smithay` function and the rest is done internally. But we are still far from this point even with glutin under the hood.

Given this as current basis, I would propose to start with a similar implementation as https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/window/glutin.rs and using a feature switch for selecting the implementation. This would probably a good point to better understand the interface and requirements/possibilities from both sides!

vberger:
Hi,

To give some more context, smithay is already integrated with glium as a drawing API. This was pretty easy because glium is generic over its OpenGL provider.

[...]

PR #2793, opened 2019-06-01
gl: Add wgl WSI backend

msiglreith:
Add a wgl based window interface backend for the gl backend. This allows to get closer to the Vulkan API and a step towards multithreaded OpenGL.

It's based around context sharing where we seperate two different context types: PresentContext and DeviceContext. Each swapchain will have a present context which only does blitting towards the default framebuffer of the underlying surface and presentation. The device context is backed by no real surface/window and does the usual command submission etc. All contexts are shared. Present and device contexts only share the swapchain images.

Future work would include adding glx, egl and web support. Wayland + EGL turned out to be really tricky, need to look a layer deeper I guess (below EGL?). This PR should be an initial test for this approach and therefore doesn't full employ all possible features of the hal/wgl API.

Long term goal would be to unify all backends and make `Instance` the single entrypoint for all APIs..

Related https://github.com/gfx-rs/gfx/issues/2749

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ EXCLUDES:=
 FEATURES_EXTRA:=mint serialize
 FEATURES_HAL:=
 FEATURES_HAL2:=
+FEATURES_HAL3:=
 
 SDL2_DEST=$(HOME)/deps
 SDL2_CONFIG=$(SDL2_DEST)/usr/bin/sdl2-config
@@ -25,6 +26,7 @@ ifeq ($(OS),Windows_NT)
 	else
 		FEATURES_HAL2=dx12
 	endif
+	FEATURES_HAL3=wgl
 else
 	UNAME_S:=$(shell uname -s)
 	EXCLUDES+= --exclude gfx-backend-dx12
@@ -55,6 +57,7 @@ check:
 	cd examples && cargo check $(CHECK_TARGET_FLAG) --features "gl"
 	cd examples && cargo check $(CHECK_TARGET_FLAG) --features "$(FEATURES_HAL)"
 	cd examples && cargo check $(CHECK_TARGET_FLAG) --features "$(FEATURES_HAL2)"
+	cd examples && cargo check $(CHECK_TARGET_FLAG) --features "$(FEATURES_HAL3)"
 	cd src/warden && cargo check $(CHECK_TARGET_FLAG) --no-default-features
 	cd src/warden && cargo check $(CHECK_TARGET_FLAG) --features "env_logger gl gl-headless $(FEATURES_HAL) $(FEATURES_HAL2)"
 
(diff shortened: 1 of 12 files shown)


src/backend/gl/src/device.rs
```diff
@@ -1713,7 +1714,146 @@ impl d::Device<B> for Device {
...
+        let mut images = Vec::new();
+
+        for _ in 0..config.image_count {
+            unsafe {
+                let mut fbo = 0;
+                gl.GenFramebuffers(1, &mut fbo);
+                gl.BindFramebuffer(gl::FRAMEBUFFER, fbo);
+                fbos.push(fbo);
+
+                let image = if config.image_layers > 1
```
kvark: can we use existing routines to create a texture? I think there is an opportunity to share the code here. We must already have the code path that checks for the use and goes through glTexStorage versus glRenderbufferStorage based on it.

src/backend/gl/src/device.rs
```diff
@@ -1713,7 +1714,146 @@ impl d::Device<B> for Device {
...
+                                    config.image_layers as _,
+                                    int_format,
+                                    w as _,
+                                    h as _,
+                                );
+                            } else {
+                                gl.TexParameteri(
+                                    gl::TEXTURE_2D,
+                                    gl::TEXTURE_MAX_LEVEL,
+                                    (config.image_layers - 1) as _,
```
kvark: why are we treating `image_layers` as the mipmap count and allocating accordingly?

src/backend/gl/src/lib.rs
```diff
@@ -125,6 +133,12 @@ impl Error {
 /// Internal struct of shared data between the physical and logical device.
 struct Share {
     context: GlContainer,
+    /// Context associated with an instance.
+    ///
+    /// Parenting context for all device contexts shared with it.
+    /// Used for querying basic information and spawning shared contexts.
+    instance_ctxt: DeviceContext,
```
kvark: first time I'm seeing "context" shortened to "ctxt"

src/backend/gl/src/device.rs
msiglreith: Appears to be a bug in the current implementation, fixed it. Nice catch!

msiglreith:
Thanks for the review! Will fix some open points and do a cleanup round to finish it up.

msiglreith:
Updated the PR, still has a lot of open points but I believe it's in a good place to get started with. Not sure how the other WSIs will look like so far.

kvark:
Can we land this in some form?

msiglreith:
yes, rebased

kvark:
How are we testing it on CI?

Makefile
```diff
@@ -3,6 +3,7 @@ EXCLUDES:=
 FEATURES_EXTRA:=mint serialize
 FEATURES_HAL:=
 FEATURES_HAL2:=
+FEATURES_HAL3:=
```
kvark: we should probably make it a list, which is later iterated upon

msiglreith:
bors r=kvark

PR #2795, opened 2019-06-03
[pre-ll] Bump versions to match crates

tangmi:
Last month, a handful of pre-ll crates had a patch version bumped, but the crate versions were not updated in Github. Additionally, the prior version was yanked from crates (e.g. gfx was bumped to 0.18.1, which 0.18.0 yanked).

Both of these things together causes issues with [cargo patches](https://doc.rust-lang.org/cargo/reference/manifest.html#the-patch-section), as the patch has the original version, which cargo tries to resolve against the new version. Since the original is yanked, the patch is not valid and then is not used if it's not explicitly added to the `Cargo.lock` file (which is pretty painful to do manually).

This PR bumps these versions in the `pre-ll` branch so that the repo better reflects what is published and those branched off this repo can be a valid `[patch.crates-io]` patch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ gfx_core = { path = "src/core", version = "0.9" }
 gfx = { path = "src/render", version = "0.18" }
 gfx_macros = { path = "src/macros", version = "0.2" }
 gfx_device_gl = { path = "src/backend/gl", version = "0.16" }
-gfx_window_glutin = { path = "src/window/glutin", version = "0.30" }
+gfx_window_glutin = { path = "src/window/glutin", version = "0.31" }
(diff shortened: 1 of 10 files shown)


kvark:
I don't think you are expected to have "cargo patch" working with "pre-ll" branch. Instead, you should be checking out the branch relevant to the version being used, which is likely [v0.18](https://github.com/gfx-rs/gfx/tree/v0.18). If you agree, I don't think it makes much sense to bother changing the versions until we decide to publish a new breaking release (0.19), in which case the versions will need to be changed again regardless...

tangmi:
!!!

I totally didn't realize those branches existed! You're completely right, that makes more sense, I'll rebase my patch off of that branch.

Thanks!

PR #2798, opened 2019-06-04
[hal] rename present_nosemaphores

Michael-Lfx:
Fixes #issue
PR checklist:
  - [x] metal

### DIFF
--- a/src/hal/src/window.rs
+++ b/src/hal/src/window.rs
@@ -470,7 +470,7 @@ pub trait Swapchain<B: Backend>: fmt::Debug + Any + Send + Sync {
     }
 
     /// Present one acquired image without any semaphore synchronization.
-    unsafe fn present_nosemaphores<'a, C>(
+    unsafe fn present_without_semaphores<'a, C>(
         &'a self,
         present_queue: &mut CommandQueue<B, C>,
         image_index: SwapImageIndex,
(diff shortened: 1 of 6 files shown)


src/hal/src/window.rs
```diff
@@ -470,7 +470,7 @@ pub trait Swapchain<B: Backend>: fmt::Debug + Any + Send + Sync {
     }
 
     /// Present one acquired image without any semaphore synchronization.
-    unsafe fn present_nosemaphores<'a, C>(
+    unsafe fn present_no_semaphores<'a, C>(
```
kvark: we should go further and call it `present_without_semaphores`.
Also `submit_nosemaphores` needs to be updated as well

Michael-Lfx:
Okay. Let me rename it.

src/hal/src/window.rs
Michael-Lfx: Done. r? cc @kvark

grovesNL: @Michael-Lfx `quad/main.rs` still uses `submit_nosemaphores`

Michael-Lfx: Fixed 10 minutes ago.

grovesNL: Sorry it looks like there is one more (at least) `submit_nosemaphores` in `compute/main.rs`

Michael-Lfx: https://github.com/gfx-rs/gfx/commit/71fe6b5c0f94b3c530c93532205efb678ee58be5 should work

PR #2802, opened 2019-06-05
[mtl] impl create_surface() the way vulkan backend did

Michael-Lfx:
Fixes #issue
PR checklist:
  - [x] metal

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -184,77 +184,21 @@ impl Instance {
         Instance
     }
 
-    unsafe fn create_from_layer(&self, layer: CAMetalLayer) -> window::SurfaceInner {
-        let class = class!(CAMetalLayer);
-        let proper_kind: BOOL = msg_send![layer, isKindOfClass: class];
-        assert_eq!(proper_kind, YES);
-        msg_send![layer, retain];
-        window::SurfaceInner::new(None, layer)
-    }
-
-    pub fn create_surface_from_layer(&self, layer: CAMetalLayer, enable_signposts: bool) -> Surface {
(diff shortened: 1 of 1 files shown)


aleksijuvani:
This removes the public functions `create_surface_from_uiview` and `create_surface_from_nsview` which makes window creation significantly more difficult when the winit feature is disabled. Only the `create_surface_from_layer` function remains. Was this intentional?

Michael-Lfx:
Yes. Both iOS and macOS supports `CAMetalLayer` and has almost the same operation. If you are not using `winit` I think you are using `MTLView` or rewriting `+ layerClass` to change the underlaying layer to `CAMetalLayer`. In such case `create_surface_from_layer()` can be called without any problem, right?

aleksijuvani:
If I wanted to use `create_surface_from_layer` with an `NSView`, I think I would have to duplicate all the logic that already exists in `create_from_nsview`. At least I think that's the case? I see that it checks for an existing `CAMetalLayer` on the `NSView`, but creates it if it doesn't exist. When does the layer not exist exactly?

To be more specific about what I'm doing, I'm trying to create a surface from an `NSView` handle I get from winit (and winit only provides an `NSView`), but I'm using a version of winit (eventloop-2.0 specifically) than the one used by gfx-hal, so I cannot use the winit feature in gfx-hal. Additionally, disabling the winit feature allows me to be more agnostic about the windowing library I'm using, so I can easily switch it out in the future.

kvark:
I missed the fact it removes the entry points, sorry!
At the end of the day, everyone should just be providing a layer to us to use. But this time has not yet come, and we'll need to expose the entry points for longer.

PR #2803, opened 2019-06-05
[hal] elide lifetime parameter

Michael-Lfx:
Fixes #issue
PR checklist:
  - [x] metal

### DIFF
--- a/src/hal/src/window.rs
+++ b/src/hal/src/window.rs
@@ -470,13 +470,13 @@ pub trait Swapchain<B: Backend>: fmt::Debug + Any + Send + Sync {
     }
 
     /// Present one acquired image without any semaphore synchronization.
-    unsafe fn present_without_semaphores<'a, C>(
-        &'a self,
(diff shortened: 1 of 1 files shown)


kvark:
Can we do the same for submit_nosemaphores?

Michael-Lfx:
No. Tried and failed when I remove the `'a` param. Compiler requires an explicit lifetime for `submit_nosemaphores`.

kvark:
Are you seeing a compiler/lint warning about the lifetime you remove here? We might otherwise just leave it for consistency: it's not harming.

Michael-Lfx:
Removing lifetime of present_without_semaphores  causes no compile error or warning. It just work with Rust 1.35.0.

kvark:
Ok. If there is no warnings with the current code, I'd slightly prefer to keep the lifetime consistent between present and submit functions, so for it to say as it is.

Michael-Lfx:
Ok.

PR #2806, opened 2019-06-06
Add Events API and implement for Vulkan backend

### LINKED ISSUE #2804: Support Events
`Event` object is probably the largest concept we are missing from Vulkan, and we still need it:
  - as an associated type of `Backend` trait
  - as a few `Device` and `CommandBuffer` functions dealing with these objects

The first (and most important) step would be exposing these APIs, having them implemented on Vulkan only for now and leaving `unimplemented!()` elsewhere.

cormac-obrien:
Also adds placeholder implementations for all other backends.

Fixes #2804 
PR checklist:
- [x] tested examples with the following backends: dx11, dx12, vulkan, gl

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -2144,8 +2144,7 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
     unsafe fn wait_events<'a, I, J>(
         &mut self,
         _: I,
-        _: pso::PipelineStage,
-        _: pso::PipelineStage,
+        _: Range<pso::PipelineStage>,
         _: J
     ) where
         I: IntoIterator,
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -2133,6 +2133,29 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
         unimplemented!()
     }
 
+    unsafe fn set_event(&mut self, _: &(), _: pso::PipelineStage) {
+        unimplemented!()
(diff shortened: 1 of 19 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -2133,6 +2133,29 @@ impl hal::command::RawCommandBuffer<Backend> for CommandBuffer {
...
+    }
+
+    unsafe fn reset_event(&mut self, _: &(), _: pso::PipelineStage) {
+        unimplemented!()
+    }
+
+    unsafe fn wait_events<'a, I, J>(
+        &mut self,
+        _: I,
+        _: pso::PipelineStage,
```
kvark: we use ranges in other places for src .. dst, let's keep the convention here, for now

src/backend/vulkan/src/command.rs
```diff
@@ -790,6 +790,123 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+        let events = events
+            .into_iter()
+            .map(|e| e.borrow().0)
+            .collect::<Vec<_>>();
+
+        let mut global_bars: SmallVec<[vk::MemoryBarrier; 4]> = SmallVec::new();
+        let mut buffer_bars: SmallVec<[vk::BufferMemoryBarrier; 4]> = SmallVec::new();
+        let mut image_bars: SmallVec<[vk::ImageMemoryBarrier; 4]> = SmallVec::new();
+
+        for barrier in barriers {
```
kvark: we already have the same code for `pipeline_barrier`, can we re-use it?

PR #2814, opened 2019-06-09
Specify limits for portability validation layers

Thomspoon:
Fixes a few invalid validation layer messages when running with `VK_LAYER_LUNARG_standard_validation`

PR checklist:


src/hal/src/lib.rs
```diff
@@ -252,7 +252,8 @@ pub struct Limits {
     pub max_sampler_allocation_count: usize,
     ///
     pub max_bound_descriptor_sets: pso::DescriptorSetIndex,
-
+    ///
+    pub max_framebuffer_layers: usize,
```
kvark: That likely requires changes in all the other backends

PR #2815, opened 2019-06-09
Allow multisampling sample mask to be set to null

tristan:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VkSampleMask states:
> If pSampleMask is NULL, it is treated as if the mask has all bits enabled, i.e. no coverage is removed from fragments.

I believe it makes sense to follow this behaviour and make the field optional in gfx-hal.

Example of usage before this change:
```
pipeline_desc.multisampling = Some(pso::Multisampling {
    rasterization_samples: rasterization_samples,
    sample_shading: None,
    sample_mask: 0xFFFFFFFFFFFFFFFF,
    alpha_coverage: false,
    alpha_to_one: false
});
```
After:
```
pipeline_desc.multisampling = Some(pso::Multisampling {
    rasterization_samples: rasterization_samples,
    sample_shading: None,
    sample_mask: None,
    alpha_coverage: false,
    alpha_to_one: false
});
```

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -600,10 +600,13 @@ impl d::Device<B> for Device {
 
                 let multisampling_state = match desc.multisampling {
                     Some(ref ms) => {
-                        let sample_mask = [
-                            (ms.sample_mask & 0xFFFFFFFF) as u32,
-                            ((ms.sample_mask >> 32) & 0xFFFFFFFF) as u32,
-                        ];
+                        let sample_mask = match ms.sample_mask {
+                            Some(mask) => Some([
+                                (mask & 0xFFFFFFFF) as u32,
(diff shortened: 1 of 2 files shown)


kvark:
I don't think it has much value, tbh. Users can just set it to `!0` which is still shorter and easier on the type system than having an option.

tristan:
Ok, I'm not experienced enough with rust or graphics in general to argue otherwise. I'll just say, that as a beginner trying to wrap my head around gfx-hal using the vulkan-tutorials and gfx examples as a base it took me a bit of time to find what the right defaults to use for this field. But I understand that this is probably something obvious to anyone who's more familiar with graphics programming, and was probably a good lesson for me going through the pain of figuring it out. 

And `!0` is a neat tip that i didn't know about yet, thanks for that!

kvark:
Thanks for understanding, @tristan !

PR #2817, opened 2019-06-11
[hal-0.2] Back-port Debug implementations

kvark:
Sibling of #2787 
cc @BourgondAries 
PR checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## hal-0.2.1 (10-06-2019)
+  - `Debug` implementations
+
 ## hal-0.2.0 (10-05-2019)
   - pipeline cache support
   - rich presentation errors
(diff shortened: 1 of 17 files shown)


Michael-Lfx:
Will it increase the binary size when adding `derive Debug` to these struct?

m4b:
> Will it increase the binary size when adding `derive Debug` to these struct?

yes

Michael-Lfx:
Can we use to conditional compilation to remove them to get a small binary package? It's good to keep our app small both on AppStore and Google Play.

m4b:
> Can we use to conditional compilation to remove them to get a small binary package? It's good to keep our app small both on AppStore and Google Play.

Debug print functions (which will account for the increased binary size) will only be included in binary if they're used (if you link with `--gc-sections`) so I wouldn't worry about it.

PR #2822, opened 2019-06-12
Rework of gl memory / buffer handling

### LINKED ISSUE #2812: GL panics with "No buffer has been bound yet, can't map memory!"
Spot in the code https://github.com/gfx-rs/gfx/blob/416d9ff65cd4559dcda5e640d7baf79e606be4e8/src/backend/gl/src/device.rs#L1153
Reported in https://github.com/gfx-rs/wgpu-rs/pull/10#issuecomment-499659509
cc @kyren @grovesNL

kyren:
Allocates native buffers to back memory as soon as allocate_memory is called,
hal buffers become logical sub-ranges of native buffers.

This is still WIP, there is a lot of room for improvement here.  Separate INDEX
memory could be reserved only for webgl, there could be separate upload /
download memory, and the end of buffer sub-ranges could be checked with asserts
in debug mode.

Fixes #2812
PR checklist:

### DIFF
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -23,7 +23,7 @@ bitflags = "1"
 log = { version = "0.4" }
 gfx-hal = { path = "../../hal", version = "0.2" }
 smallvec = "0.6"
-glow = { version = "0.2.0" }
+glow = { git = "https://github.com/grovesNL/glow", rev = "054aabc1bfbad472e4a08bc55912552d8211daee" }
 spirv_cross = { version = "0.14.0", features = ["glsl"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -23,7 +23,7 @@ bitflags = "1"
 log = { version = "0.4" }
 gfx-hal = { path = "../../hal", version = "0.2" }
 smallvec = "0.6"
-glow = { git = "https://github.com/grovesNL/glow", rev = "054aabc1bfbad472e4a08bc55912552d8211daee" }
+glow = { git = "https://github.com/grovesNL/glow", rev = "7131fc75c1afb460161dd84b9d44fadef24f5a90" }
 spirv_cross = { version = "0.14.0", features = ["glsl"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -23,7 +23,7 @@ bitflags = "1"
 log = { version = "0.4" }
 gfx-hal = { path = "../../hal", version = "0.2" }
 smallvec = "0.6"
-glow = { git = "https://github.com/grovesNL/glow", rev = "7131fc75c1afb460161dd84b9d44fadef24f5a90" }
+glow = { git = "https://github.com/grovesNL/glow", rev = "6c74ffbea64e8fbaa1ec9e94e7f5f6791663a70e" }
 spirv_cross = { version = "0.14.0", features = ["glsl"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -148,7 +148,7 @@ struct Cache {
     // Active primitive topology, set by the current pipeline.
     primitive: Option<u32>,
     // Active index type and buffer range, set by the current index buffer.
-    index_type_range: Option<(hal::IndexType, Range<u64>)>,
+    index_type_range: Option<(hal::IndexType, Range<buffer::Offset>)>,
     // Stencil reference values (front, back).
     stencil_ref: Option<(pso::StencilValue, pso::StencilValue)>,
     // Blend color.
@@ -165,7 +165,7 @@ struct Cache {
     // Blend per attachment.
     blend_targets: Option<Vec<Option<pso::ColorBlendDesc>>>,
     // Maps bound vertex buffer offset (index) to handle / buffer range
-    vertex_buffers: Vec<Option<(n::RawBuffer, Range<u64>)>>,
+    vertex_buffers: Vec<Option<(n::RawBuffer, Range<buffer::Offset>)>>,
     // Active vertex buffer descriptions.
     vertex_buffer_descs: Vec<Option<pso::VertexBufferDesc>>,
     // Active attributes.
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -147,8 +147,8 @@ pub struct RenderPassCache {
 struct Cache {
     // Active primitive topology, set by the current pipeline.
     primitive: Option<u32>,
-    // Active index type, set by the current index buffer.
-    index_type: Option<hal::IndexType>,
+    // Active index type and buffer range, set by the current index buffer.
+    index_type_range: Option<(hal::IndexType, Range<u64>)>,
     // Stencil reference values (front, back).
     stencil_ref: Option<(pso::StencilValue, pso::StencilValue)>,
     // Blend color.
@@ -164,8 +164,8 @@ struct Cache {
     program: Option<n::Program>,
     // Blend per attachment.
     blend_targets: Option<Vec<Option<pso::ColorBlendDesc>>>,
-    // Maps bound vertex buffer offset (index) to handle.
-    vertex_buffers: Vec<Option<n::RawBuffer>>,
+    // Maps bound vertex buffer offset (index) to handle / buffer range
+    vertex_buffers: Vec<Option<(n::RawBuffer, Range<u64>)>>,
     // Active vertex buffer descriptions.
     vertex_buffer_descs: Vec<Option<pso::VertexBufferDesc>>,
     // Active attributes.
@@ -178,7 +178,7 @@ impl Cache {
     pub fn new() -> Cache {
         Cache {
             primitive: None,
-            index_type: None,
+            index_type_range: None,
             stencil_ref: None,
             blend_color: None,
             framebuffer: None,
@@ -382,7 +382,10 @@ impl RawCommandBuffer {
                 error!("No vertex buffer bound at {}", binding);
             }
 
-            let handle = vertex_buffers[binding];
+            let (handle, range) = vertex_buffers[binding].as_ref().unwrap();
+
+            let mut attribute = attribute.clone();
+            attribute.offset += range.start as u32;
 
             match vertex_buffer_descs.get(binding) {
                 Some(&Some(desc)) => {
@@ -391,8 +394,8 @@ impl RawCommandBuffer {
                         &mut self.memory,
                         &mut self.buf,
                         Command::BindAttribute(
-                            attribute.clone(),
-                            handle.unwrap(),
+                            attribute,
+                            *handle,
                             desc.stride as _,
                             desc.rate.as_uint() as u32,
                         ),
@@ -749,13 +752,10 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     }
 
     unsafe fn bind_index_buffer(&mut self, ibv: buffer::IndexBufferView<Backend>) {
-        // TODO: how can we incorporate the buffer offset?
-        if ibv.offset > 0 {
-            warn!("Non-zero index buffer offset currently not handled.");
-        }
+        let (raw_buffer, range) = ibv.buffer.as_bound();
 
-        self.cache.index_type = Some(ibv.index_type);
-        self.push_cmd(Command::BindIndexBuffer(ibv.buffer.raw));
+        self.cache.index_type_range = Some((ibv.index_type, range.start + ibv.offset..range.end));
+        self.push_cmd(Command::BindIndexBuffer(raw_buffer));
     }
 
     unsafe fn bind_vertex_buffers<I, T>(&mut self, first_binding: pso::BufferIndex, buffers: I)
@@ -768,10 +768,9 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
             if self.cache.vertex_buffers.len() <= index {
                 self.cache.vertex_buffers.resize(index + 1, None);
             }
-            self.cache.vertex_buffers[index] = Some(buffer.borrow().raw);
-            if offset != 0 {
-                error!("Vertex buffer offset {} is not supported", offset);
-            }
+
+            let (raw_buffer, range) = buffer.borrow().as_bound();
+            self.cache.vertex_buffers[index] = Some((raw_buffer, range.start + offset..range.end));
         }
     }
 
@@ -1088,7 +1087,9 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     }
 
     unsafe fn dispatch_indirect(&mut self, buffer: &n::Buffer, offset: buffer::Offset) {
-        self.push_cmd(Command::DispatchIndirect(buffer.raw, offset));
+        let (raw_buffer, range) = buffer.borrow().as_bound();
+        assert_eq!(range.start, 0, "buffer offset unsupported in indirect draw");
+        self.push_cmd(Command::DispatchIndirect(raw_buffer, offset));
     }
 
     unsafe fn copy_buffer<T>(&mut self, src: &n::Buffer, dst: &n::Buffer, regions: T)
@@ -1098,9 +1099,13 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     {
         let old_size = self.buf.size;
 
+        let (src_raw, src_range) = src.as_bound();
+        let (dst_raw, dst_range) = dst.as_bound();
         for region in regions {
-            let r = region.borrow().clone();
-            let cmd = Command::CopyBufferToBuffer(src.raw, dst.raw, r);
+            let mut r = region.borrow().clone();
+            r.src += src_range.start;
+            r.dst += dst_range.start;
+            let cmd = Command::CopyBufferToBuffer(src_raw, dst_raw, r);
             self.push_cmd(cmd);
         }
 
@@ -1148,11 +1153,13 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     {
         let old_size = self.buf.size;
 
+        let (src_raw, src_range) = src.as_bound();
         for region in regions {
-            let r = region.borrow().clone();
+            let mut r = region.borrow().clone();
+            r.buffer_offset += src_range.start;
             let cmd = match dst.kind {
-                n::ImageKind::Surface(s) => Command::CopyBufferToSurface(src.raw, s, r),
-                n::ImageKind::Texture(t, tt) => Command::CopyBufferToTexture(src.raw, t, tt, r),
+                n::ImageKind::Surface(s) => Command::CopyBufferToSurface(src_raw, s, r),
+                n::ImageKind::Texture(t, tt) => Command::CopyBufferToTexture(src_raw, t, tt, r),
             };
             self.push_cmd(cmd);
         }
@@ -1173,12 +1180,14 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
         T::Item: Borrow<command::BufferImageCopy>,
     {
         let old_size = self.buf.size;
+        let (dst_raw, dst_range) = dst.as_bound();
 
         for region in regions {
-            let r = region.borrow().clone();
+            let mut r = region.borrow().clone();
+            r.buffer_offset += dst_range.start;
             let cmd = match src.kind {
-                n::ImageKind::Surface(s) => Command::CopySurfaceToBuffer(s, dst.raw, r),
-                n::ImageKind::Texture(t, tt) => Command::CopyTextureToBuffer(t, tt, dst.raw, r),
+                n::ImageKind::Surface(s) => Command::CopySurfaceToBuffer(s, dst_raw, r),
+                n::ImageKind::Texture(t, tt) => Command::CopyTextureToBuffer(t, tt, dst_raw, r),
             };
             self.push_cmd(cmd);
         }
@@ -1218,22 +1227,27 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     ) {
         self.bind_attributes();
 
-        let (start, index_type) = match self.cache.index_type {
-            Some(hal::IndexType::U16) => (indices.start * 2, glow::UNSIGNED_SHORT),
-            Some(hal::IndexType::U32) => (indices.start * 4, glow::UNSIGNED_INT),
+        let (index_type, buffer_range) = match &self.cache.index_type_range {
+            Some((index_type, buffer_range)) => (index_type, buffer_range),
             None => {
                 warn!("No index type bound. An index buffer needs to be bound before calling `draw_indexed`.");
                 self.cache.error_state = true;
                 return;
             }
         };
+
+        let (start, index_type) = match index_type {
+            hal::IndexType::U16 => (indices.start as buffer::Offset * 2 + buffer_range.start, glow::UNSIGNED_SHORT),
+            hal::IndexType::U32 => (indices.start as buffer::Offset * 4 + buffer_range.start, glow::UNSIGNED_INT),
+        };
+
         match self.cache.primitive {
             Some(primitive) => {
                 self.push_cmd(Command::DrawIndexed {
                     primitive,
                     index_type,
                     index_count: indices.end - indices.start,
-                    index_buffer_offset: start as _,
+                    index_buffer_offset: start,
                     base_vertex,
                     instances,
                 });
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -1088,8 +1088,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
 
     unsafe fn dispatch_indirect(&mut self, buffer: &n::Buffer, offset: buffer::Offset) {
         let (raw_buffer, range) = buffer.borrow().as_bound();
-        assert_eq!(range.start, 0, "buffer offset unsupported in indirect draw");
-        self.push_cmd(Command::DispatchIndirect(raw_buffer, offset));
+        self.push_cmd(Command::DispatchIndirect(raw_buffer, range.start + offset));
     }
 
     unsafe fn copy_buffer<T>(&mut self, src: &n::Buffer, dst: &n::Buffer, regions: T)
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -5,8 +5,7 @@ use std::sync::{Arc, Mutex, RwLock};
 use std::slice;
 
 use glow::Context;
-use crate::{GlContainer, GlContext, IMAGE_MEM_TYPE_MASK, INDEX_MEM_TYPE_MASK, OTHER_MEM_TYPE_MASK,
-            DEVICE_LOCAL_MASK, COHERENT_MASK};
+use crate::{GlContainer, GlContext};
 
 use crate::hal::backend::FastHashMap;
 use crate::hal::format::{Format, Swizzle};
@@ -22,7 +21,7 @@ use spirv_cross::{glsl, spirv, ErrorCode as SpirvErrorCode};
 use crate::info::LegacyFeatures;
 use crate::pool::{BufferMemory, OwnedBuffer, RawCommandPool};
 use crate::{conv, native as n, state};
-use crate::{Backend as B, Share, Starc, Surface, Swapchain};
+use crate::{Backend as B, Share, MemoryRole, Starc, Surface, Swapchain};
 
 /// Emit error during shader module creation. Used if we don't expect an error
 /// but might panic due to an exception in SPIRV-Cross.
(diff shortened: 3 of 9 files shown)


kyren:
This is still highly WIP, I just went ahead and created the PR to get feedback.

Notably, this *doesn't fix wgpu-rs with rendy* yet, rendy errors with `NoSuitableMemory(0, CPU_VISIBLE)`.  I need to look into exactly why that is, and it might either need further gfx changes or potentially rendy changes as well.

kyren:
Okay, so the cube example in wgpu-rs is still broken, and it has something to do with the fact that you can't call `glTexSubImage2D` with a buffer bound to  `GL_PIXEL_UNPACK_BUFFER` that is also currently memory mapped.

In the cube example, this is coming from uploading texture data to a new buffer and calling `copy_buffer_to_texture` on that buffer.  I believe this ends up being allocated on a rendy-memory "dynamic" heap, which actually does *not* unmap its buffers when `DynamicBlock::unmap` is called.  HOWEVER, even if you patch rendy's `DynamicBlock::unmap` to call `device.unmap_memory`, the cube example will *still* not work.

I don't know how opengl mapping of a buffer *range* really works, I guess you can't unmap a buffer *range*, you can only unmap all of the mapped buffer ranges at once?  Does opengl support mapping multiple different ranges of a single buffer simultaneously?

Curiously, if you patch rendy memory to call gfx `unmap_memory` on `DynamicBlock::unmap` as described above, the cube example errors with the exact same error!  AND, even more curiously, if you unmap the buffer AGAIN as a hack in gl/src/queue.rs handling of `Command::CopyBufferToTexture`, the call to `glTexSubImage2D` will *not* error.  I don't *think* I'm missing a call to `glMapBufferRange` anywhere, so my current theory is that `glUnmapBuffer` unmaps the *most recent* call to `glMapBufferRange`, but not *all* of them.

In any case, I think the best way forward might be not to use the rendy dynamic allocator with the opengl backend and instead substitute the dedicated one?  I'm not really super sure!

src/backend/gl/src/command.rs
```diff
@@ -147,8 +147,8 @@ pub struct RenderPassCache {
 struct Cache {
     // Active primitive topology, set by the current pipeline.
     primitive: Option<u32>,
-    // Active index type, set by the current index buffer.
-    index_type: Option<hal::IndexType>,
+    // Active index type and buffer range, set by the current index buffer.
+    index_type_range: Option<(hal::IndexType, Range<u64>)>,
```
kvark: we could use `hal::buffer::Offset` typedef here

src/backend/gl/src/command.rs
```diff
@@ -1088,7 +1087,9 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
     }
 
     unsafe fn dispatch_indirect(&mut self, buffer: &n::Buffer, offset: buffer::Offset) {
-        self.push_cmd(Command::DispatchIndirect(buffer.raw, offset));
+        let (raw_buffer, range) = buffer.borrow().as_bound();
+        assert_eq!(range.start, 0, "buffer offset unsupported in indirect draw");
```
kvark: we actually support indirect buffer offset, it's a part of `DispatchIndirect` variant.
What we do *not* support is `draw_indexed_indirect` when index buffer offset is non-zero

src/backend/gl/src/device.rs
```diff
@@ -1011,47 +1053,15 @@ impl d::Device<B> for Device {
...
-
-        if self.share.private_caps.buffer_storage {
-            //TODO: glow::DYNAMIC_STORAGE_BIT | glow::MAP_PERSISTENT_BIT
-            let flags = memory.map_flags();
-            //TODO: use *Named calls to avoid binding
-            gl.bind_buffer(target, Some(buffer.raw));
-            gl.buffer_storage(target, buffer.requirements.size as _, None, flags);
-            gl.bind_buffer(target, None);
-        } else {
-            let flags = if cpu_can_read && cpu_can_write {
```
kvark: Removing these flags appears to be a regression. It's fine to catch up to this after the fact, since your change is much more important, but need to keep it on the radar.

src/backend/gl/src/lib.rs
```diff
@@ -484,35 +490,34 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+        let memory_types = vec![
+            // Faked DEVICE_LOCAL memory for images, no gl buffer is actually allocated for it.
+            hal::MemoryType {
                 properties: Properties::DEVICE_LOCAL,
                 heap_index: 0,
-            }]
-        };
+            },
+            // Memory type for uses other than images and INDEX
+            hal::MemoryType {
```
kvark: would be good to also expose a device-local memory type for buffers

src/backend/gl/src/lib.rs
```diff
@@ -484,35 +490,34 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+                heap_index: 1,
+            },
+            // For security reasons, WebGL does not allow "element array buffers" to be used as any
+            // other kind of buffer.  We need to provide a unique type of memory specifically for
+            // buffers with INDEX usage.
+            hal::MemoryType {
+                properties: Properties::CPU_VISIBLE
+                    | Properties::COHERENT
+                    | Properties::CPU_CACHED,
+                heap_index: 2,
```
kvark: this doesn't need to be a separate heap index. We should have 0 for device local types and 1 for cpu visible types.

src/backend/gl/src/queue.rs
```diff
@@ -612,7 +612,7 @@ impl CommandQueue {
                     r.image_extent.height as _,
                     glow::RGBA,
                     glow::UNSIGNED_BYTE,
-                    0,
+                    r.buffer_offset as i32,
```
kvark: we should use the offset in `CopyTextureToBuffer` as well

src/backend/gl/src/command.rs
kyren: good catch, done!

src/backend/gl/src/lib.rs
kyren: Yeah that makes perfect sense, done.

src/backend/gl/src/queue.rs
kyren: Whoops!  I did in fact forget about `CopyTextureToBuffer`, and it's now waiting on https://github.com/grovesNL/glow/pull/13

src/backend/gl/src/command.rs
kyren: Right, okay that makes sense, I was confused between `DispatchIndirect` and `draw_indirect` and `draw_indexed_indirect`.  OpenGL doesn't support `draw_indexed_indirect` at all right now so that's not an issue.

src/backend/gl/src/device.rs
kyren: The way the gl backend was before, allocated memory had `n::Memory` properties hard-coded to `memory::Properties::CPU_VISIBLE | memory::Properties::CPU_CACHED`, which means that it was always really allocated with `glow::MAP_READ_BIT | glow::MAP_WRITE_BIT` or `glow::DYNAMIC_DRAW`, the rest of it was dead code.

Now, it's still hard-coded, but since the allocation takes place in the same place that the `n::Memory` flags are set, there's nothing to check to to see what flags to use when allocating.

I agree it would probably be better if you could control the buffer flags more than this, but it should have the same behavior as it did before.

src/backend/gl/src/lib.rs
kyren: That sounded right to me too, but I didn't know enough to know how to do it correctly.  If we have a device-local buffer type, what should the gl backend do when copying from a cpu-visible buffer (backed by a real native buffer) to a device-local buffer?  I imagine the correct answer is *nothing*, but what happens then when the user tries to copy from a fake device-local buffer to e.g. an image buffer?  This might be another case where if I knew more about how HAL works it would make more sense to me, but as of right now I need it explained a bit slower :D

kyren:
I think getting this to work on webgl will require mapping *without* `COHERENT` and relying on explicit flushing.  Currently working on it.

kyren:
I'm *very* unsure about whether the non-coherent gl buffer mapping is correct...

src/backend/gl/src/queue.rs
kyren: Done, but it requires depending on a git rev of glow again.

kyren:
Okay, I have tested this PR via wgpu-rs and can confirm the cube demo works again!

I have also tested this PR with several other required patches and can confirm this allows wgpu-rs to work on webgl again as it did before!  Only simple translucent quad is tested, and this requires several additional in-flight patches to `wgpu-rs`, `wgpu-native`, and `gfx-*` (start [here](https://github.com/kyren/wgpu-rs/tree/mothra) if you're reading along and are interested).

src/backend/gl/src/lib.rs
kyren: Well, now that I think about it I suppose just having an actual, real buffer without any memory mapping would be the right thing to do, even if it needlessly duplicates memory on platforms with emulated memory mapping?

kyren: Okay I think I pushed something that works, it at least fixes the wgpu-rs shadow example

kyren:
Now the wgpu-rs shadow example works again (to the same extent it used to work, with some graphical errors).

src/backend/gl/Cargo.toml
```diff
@@ -23,7 +23,7 @@ bitflags = "1"
 log = { version = "0.4" }
 gfx-hal = { path = "../../hal", version = "0.2" }
 smallvec = "0.6"
-glow = { version = "0.2.0" }
+glow = { git = "https://github.com/grovesNL/glow", rev = "054aabc1bfbad472e4a08bc55912552d8211daee" }
```
kvark: @grovesNL what's the plan here? Are we waiting for more fixes to `glow` before a patch can be released?

src/backend/gl/src/device.rs
```diff
@@ -515,7 +514,6 @@ impl d::Device<B> for Device {
 
             Ok(n::Memory {
                 properties: memory::Properties::CPU_VISIBLE
-                    | memory::Properties::COHERENT
```
kvark: I'm sorry to bring it up, but it looks like we should expose both coherent and non-coherent memory types for CPU visible buffers (where persistent mapping is available), given that GL has the corresponding `MAP_FLUSH_EXPLICIT_BIT` bit that we can control.

It is totally fine to delay for later. My only concern would be that some of gfx-hal users may expect at least one coherent memory type exposed, since this is a guarantee of Vulkan.

src/backend/gl/src/lib.rs
kvark: Copying from one buffer to another would be going through `glCopyBufferSubData`.

>what happens then when the user tries to copy from a fake device-local buffer to e.g. an image buffer

Device-local buffer is just a GL buffer (or a sub-range of it, if we have a buffer associated with `Memory`) without MAP_XXX flags. Copying from it, to it, using as index/vertex, all those things are done the same way as usual.

src/backend/gl/src/device.rs
```diff
@@ -1096,38 +1121,56 @@ impl d::Device<B> for Device {
...
+            let buffer = mem.buffer.expect("cannot flush DEVICE_LOCAL memory");
+            gl.bind_buffer(mem.target, Some(buffer));
+
+            let offset = *range.start().unwrap_or(&0);
+            let size = *range.end().unwrap_or(&mem.size) - offset;
+
+            if self.share.private_caps.emulate_map {
+                let ptr = mem.emulate_map_allocation.get().unwrap();
+                let slice = slice::from_raw_parts_mut(ptr.offset(offset as isize), size as usize);;
+                gl.buffer_data_u8_slice(mem.target, slice, glow::DYNAMIC_DRAW);
```
kvark: should probably use `buffer_sub_data` instead of rewriting the whole buffer

src/backend/gl/src/device.rs
```diff
@@ -1096,38 +1121,56 @@ impl d::Device<B> for Device {
...
-        gl.bind_buffer(target, None);
+        gl.bind_buffer(memory.target, None);
 
         if let Err(err) = self.share.check() {
             panic!("Error unmapping memory: {:?} for memory {:?}", err, memory);
         }
     }
 
-    unsafe fn flush_mapped_memory_ranges<'a, I, R>(&self, _: I) -> Result<(), d::OutOfMemory>
+    unsafe fn flush_mapped_memory_ranges<'a, I, R>(&self, ranges: I) -> Result<(), d::OutOfMemory>
```
kvark: we are also missing `invalidate_mapped_memory_ranges` implementation, which is required to be sensible for non-coherent memory

src/backend/gl/src/lib.rs
kyren: Yeah, that's where I ended up.

src/backend/gl/src/device.rs
kyren: Wow, that's awful I'm very sorry!  There *is* no buffer_sub_data in glow, I will add that.

src/backend/gl/src/device.rs
kyren: I can add that as well.

kyren:
I've added an implementation of `invalidate_mapped_memory_ranges`, added both incoherent and coherent memory types, and changed to use `glBufferSubData` for flushing buffer in [my personal branch of gfx](https://github.com/kyren/gfx/tree/mothra).  Waiting on `glow` to merge my PRs and I will update this PR with those changes.

kyren:
I have probably added a lot of portability issues with this PR, I'm open to suggestions about what needs to be fixed right now, but I don't have a great way of testing everything I might do.

I know that with the changes I'm going to add as soon as the glow PRs are merged, the gl backend will be using `glInvalidateBufferSubData` which I guess is core only since 4.3.  A possible improvement here is to either ONLY provide incoherent memory if that function is available, or potentially provide incoherent memory for platforms with emulated buffer mapping and coherent memory only for platforms with native buffer mapping.

I'm open to suggestions here.  I don't mind doing more work on this, but I do feel like the PR has gotten a bit more complex than I was really intending from the outset (and I feel a smidge unqualified to be doing some of the things I'm doing).

src/backend/gl/src/device.rs
kyren: done (in personal branch), waiting on glow.

src/backend/gl/src/device.rs
kyren: done in personal branch, waiting on glow

src/backend/gl/src/device.rs
kyren: Added `glInvalidateBufferSubData` to glow and used that (in personal branch), when using emulated memory maps I'm currently doing... nothing?  I guess that's the right thing to do?

src/backend/gl/Cargo.toml
grovesNL: Nope, 0.2.0 was published a week ago but there have been a few PRs since then. We can publish whenever this branch is ready to replace the git dependency

src/backend/gl/src/device.rs
kvark: Well, flush() updates the data from mapping to GPU, while invalidate() should update the data *from* GPU to the mapping. It sounds like this still needs to be done with the emulated maps?

src/backend/gl/src/device.rs
```diff
@@ -468,16 +469,74 @@ pub(crate) unsafe fn set_sampler_info<SetParamFloat, SetParamFloatVec, SetParamI
...
         size: u64,
     ) -> Result<n::Memory, d::AllocationError> {
-        // TODO
-        Ok(n::Memory {
-            properties: memory::Properties::CPU_VISIBLE | memory::Properties::CPU_CACHED,
-            first_bound_buffer: Cell::new(None),
-            size,
-            emulate_map_allocation: RefCell::new(std::ptr::null_mut()),
-        })
+        if (1 << mem_type.0) & IMAGE_MEM_TYPE_MASK != 0 {
```
kvark: should this just be a comparison `1 << mem_type.0  == IMAGE_MEM_TYPE_MASK`?

src/backend/gl/src/device.rs
```diff
@@ -468,16 +469,74 @@ pub(crate) unsafe fn set_sampler_info<SetParamFloat, SetParamFloatVec, SetParamI
...
+            Ok(n::Memory {
+                properties: memory::Properties::DEVICE_LOCAL,
+                buffer: None,
+                target: 0,
+                size,
+                emulate_map_allocation: Cell::new(None),
+                map_flags: 0,
+            })
+        } else {
+            assert!(self.share.private_caps.buffer_role_change);
```
kvark: wait, so this code can't allocate buffer memory at all if `buffer_role_change` is false?

src/backend/gl/src/device.rs
```diff
@@ -468,16 +469,74 @@ pub(crate) unsafe fn set_sampler_info<SetParamFloat, SetParamFloatVec, SetParamI
...
+
+            let target = if (1 << mem_type.0) & INDEX_MEM_TYPE_MASK != 0 {
```
kvark: similarly, could we compare directly instead?

src/backend/gl/src/device.rs
```diff
@@ -468,16 +469,74 @@ pub(crate) unsafe fn set_sampler_info<SetParamFloat, SetParamFloatVec, SetParamI
...
+                glow::ELEMENT_ARRAY_BUFFER
+            } else {
+                glow::ARRAY_BUFFER
+            };
+
+            let is_device_local = (1 << mem_type.0) & DEVICE_LOCAL_MASK != 0;
```
kvark: nit: since this flag is needed in both branches, let's move this assignment up to the beginning of the function?

src/backend/gl/src/device.rs
```diff
@@ -468,16 +469,74 @@ pub(crate) unsafe fn set_sampler_info<SetParamFloat, SetParamFloatVec, SetParamI
...
+            }
+
+            let mut map_flags = glow::MAP_READ_BIT | glow::MAP_WRITE_BIT | glow::MAP_PERSISTENT_BIT;
+            if (1 << mem_type.0) & COHERENT_MASK == 0 {
+                map_flags |= glow::MAP_FLUSH_EXPLICIT_BIT;
+            }
+
+            Ok(n::Memory {
+                properties: memory::Properties::CPU_VISIBLE
+                    | memory::Properties::CPU_CACHED,
```
kvark: in the future, we'd want to expose "read" memory type independent of the "write", but it doesn't need to happen here

src/backend/gl/src/device.rs
```diff
@@ -978,31 +1037,36 @@ impl d::Device<B> for Device {
...
+        let (size, usage) = match *buffer {
+            n::Buffer::Unbound { size, usage } => (size, usage),
+            n::Buffer::Bound { .. } => panic!("Unexpected Buffer::Bound"),
+        };
+
+        let (type_mask, alignment) = if usage.contains(buffer::Usage::INDEX) {
+            // Buffers are only allowed to be INDEX usage XOR another type of usage, they are not
+            // allowed to have both INDEX and non-INDEX usage.
+            if !(buffer::Usage::INDEX | buffer::Usage::TRANSFER_SRC | buffer::Usage::TRANSFER_DST).contains(usage) {
+                (0, 1)
```
kvark: let's `error!` log here for the users to find out what happened earlier :)

src/backend/gl/src/device.rs
```diff
@@ -1096,51 +1127,91 @@ impl d::Device<B> for Device {
...
 
         if self.share.private_caps.emulate_map {
-            let raw = memory.emulate_map_allocation.replace(std::ptr::null_mut());
-            let mapped = slice::from_raw_parts_mut(raw, memory.size as usize);
+            let ptr = memory.emulate_map_allocation.replace(None).unwrap();
+            let allocation = slice::from_raw_parts_mut(ptr, memory.size as usize);
             // TODO: Access
-            gl.buffer_data_u8_slice(target, mapped, glow::DYNAMIC_DRAW);
-            let _ = *Box::from_raw(raw);
+            gl.buffer_sub_data_u8_slice(memory.target, 0, &allocation);
```
kvark: we shouldn't be doing this on `unmap`. In Vulkan and gfx-hal, mapping as an operation doesn't actually involve any synchronization. I.e. the only thing that synchronizes is:
  - submit() for coherent memory (implicitly synchronizes all active coherent mappings)
  - explicit flush() and invalidate() for non-coherent memory

src/backend/gl/src/device.rs
kyren: For this case, yes, but that assumes that there is only one type of image memory.  For the other classes there are multiple types of memory that could work, so the intention was to see whether the given type was in the bit mask at all.

src/backend/gl/src/device.rs
kyren: There's more than one type of index memory, so the intention here is to check whether the given memory type is any of the index memory types.  Is there a clearer way to write this?

src/backend/gl/src/device.rs
kyren: Yeah, I know it's bad, I figured you'd ask about this.  We could either offer a unique type of memory for each role and work differently whether or not `buffer_role_change` was allowed, OR we could force mmap emulation if `buffer_role_change` is not there.  I'm fine with doing either one of those now, this change was just getting to be larger than I expected it to be so I wasn't sure what the best way forward right now was.

src/backend/gl/src/device.rs
kyren: Yeah, agreed!

src/backend/gl/src/device.rs
kyren: Done, and I went ahead and gave all of the properties clearer names at the top of the function because I felt that was easier to follow.

src/backend/gl/src/device.rs
kyren: Done!

src/backend/gl/src/device.rs
kyren: That makes sense, thanks for being patient with me and helping me learn all of the different memory semantics for modern gfx apis!

src/backend/gl/src/device.rs
kyren: Okay yeah that makes sense.  This will also require a new glow method, I'll merge into this PR once that's merged into glow.

kyren:
> Thank you for driving this forward! I think we've narrowed down most of rough spots, here is just a few left (noted below)

You're very very welcome, thanks for being patient with me and teaching me a lot about modern gfx APIs in the process :stuck_out_tongue:

src/backend/gl/src/device.rs
kvark: > OR we could force mmap emulation if buffer_role_change is not there.

Is this really an option though? If the user wants a buffer to support both being a vertex buffer and a uniform buffer, no amount of mmap emulation would give you that.

So it looks to me that we just need to expose this "no role change" policy in the exposed memory types and live with that, for now. Later, if we find that applications are too restricted with this, we could think of some workarounds involving copying data around to temporary buffers, but I don't want us to go there just now.

src/backend/gl/src/lib.rs
```diff
@@ -481,37 +495,68 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
...
+                    | Properties::CPU_CACHED,
+                heap_index: 1,
+            },
+        ];
+
+        // If we are not emulating buffer mapping, we can provide coherent buffer mapping by
+        // omitting the GL_MAP_FLUSH_EXPLICIT_BIT flag.
+        if !self.0.private_caps.emulate_map {
+            memory_types.push(
+                // Coherent CPU_VISIBLE memory for INDEX buffers
```
kvark: let's move these to the beginning of `memory_types`. In Vulkan there is an interesting requirement that a memory type with more flags should be exposed first (i.e one with flags A | B should come before A).

src/backend/gl/src/device.rs
kyren: So, just to clarify, you're saying that the gl backend should have memory types based on no role changes at all, or that this should be dynamic depending on `buffer_role_change`?  Doing the first one is... relatively easy, but the second one is a more involved change.  It's easier for me to do the first one and has a higher probability of me getting it right, but if you think the that the gl backend should definitely support multi purpose buffers if available, I can do that.

src/backend/gl/src/lib.rs
kyren: That makes the index masks dynamic based on capabilities, darn.

kyren:
> Alright, I suppose there is only getting `glow` published and maybe reordering the memory types that's left to do here? Or do you think we should land this as is now and follow-up?

I guess it would be better to fix the memory types thing and the brokenness of `!buffer_role_change` before merging...

src/backend/gl/src/device.rs
kvark: > should be dynamic depending on buffer_role_change

We definitely need the exposed memory types being different, based on the `buffer_role_change` capability.

> is the PIXEL_PACK_BUFFER target special somehow on non-webgl and allowed

We only *have* to use this binding when doing texture-buffer copies. Technically, this is subset of TRANSFER_DST capability (while PIXEL_UNPACK_BUFFER is a subset with TRANSFER_SRC). So with this in mind there is pretty much 1:1 correspondence between a buffer binding point ("role") and `buffer::Usage`.

When the implementation *just* needs to bind a buffer, regardless of the particular bind point, it should pick the bind point that the buffer supports. For example, if the user created a buffer with only VERTEX usage, we'll use the GL_ARRAY_BUFFER binding point for it, etc.

kyren:
Okay, I have a new system for handling memory types that is dynamic based on capabilities rather than the old simplistic system.  I've also fixed what I think are several bugs that I caught in review, so there's kind of a lot of new changes.

This should fix handling the `!buffer_role_change` case, improve the backend on non-webgl, and order the memory types according to vulkan requirements.

src/backend/gl/src/device.rs
```diff
@@ -464,73 +463,81 @@ impl d::Device<B> for Device {
...
-            let mut map_flags = glow::MAP_READ_BIT | glow::MAP_WRITE_BIT | glow::MAP_PERSISTENT_BIT;
-            if !is_coherent_memory {
-                map_flags |= glow::MAP_FLUSH_EXPLICIT_BIT;
+                let mut map_flags = glow::MAP_WRITE_BIT
+                    | glow::MAP_PERSISTENT_BIT
+                    | glow::MAP_FLUSH_EXPLICIT_BIT;
+                if is_readable_memory {
+                    map_flags |= glow::MAP_READ_BIT;
+                }
+                if !is_coherent_memory {
```
kvark: the condition appears wrong

src/backend/gl/src/device.rs
```diff
@@ -464,73 +463,81 @@ impl d::Device<B> for Device {
...
-                emulate_map_allocation: Cell::new(None),
-                map_flags,
-            })
+            MemoryRole::Image => {
+                assert!(is_device_local_memory);
+                Ok(n::Memory {
+                    properties: memory::Properties::DEVICE_LOCAL,
+                    buffer: None,
+                    size,
+                    target: 0,
```
kvark: let's move this field behind the `buffer` option?

src/backend/gl/src/device.rs
```diff
@@ -1044,24 +1051,36 @@ impl d::Device<B> for Device {
...
-            // Buffers are only allowed to be INDEX usage XOR another type of usage, they are not
-            // allowed to have both INDEX and non-INDEX usage.
-            if !(buffer::Usage::INDEX | buffer::Usage::TRANSFER_SRC | buffer::Usage::TRANSFER_DST).contains(usage) {
-                error!("gl backend does not allow a buffer to be used for both INDEX and non-INDEX usage");
-                (0, 1)
-            } else {
-                // Alignment of 4 covers indexes of type u16 and u32
-                (INDEX_MEM_TYPE_MASK, 4)
-            }
+        let alignment = if usage.contains(buffer::Usage::INDEX) {
```
kvark: why don't we just return 4 unconditionally here? It's a reasonable alignment to adhere to.

src/backend/gl/src/lib.rs
```diff
@@ -315,6 +316,162 @@ impl PhysicalDevice {
...
+        // allowed opengl roles.
+        let mut buffer_role_memory_types = Vec::new();
+
+        // Mimicking vulkan, memory types with more flags should come before those with fewer flags
+        if private_caps.map {
+            buffer_role_memory_types.push(hal::MemoryType {
+                properties: memory::Properties::CPU_VISIBLE | memory::Properties::CPU_CACHED | memory::Properties::COHERENT,
+                heap_index: CPU_VISIBLE_HEAP,
+            });
+            buffer_role_memory_types.push(hal::MemoryType {
```
kvark: are we missing the CPU-visible non-cached one?

src/backend/gl/src/lib.rs
```diff
@@ -315,6 +316,162 @@ impl PhysicalDevice {
...
+                            target: glow::PIXEL_PACK_BUFFER,
+                        }
+                    ));
+                } else {
+                    // If we have `buffer_role_change` capability and are not on WebGL, a buffer can
+                    // be used for any role.
+                    memory_types.push((
+                        buffer_role_memory_type,
+                        MemoryRole::Buffer {
+                            allowed_usage: buffer::Usage::TRANSFER_SRC
```
kvark: could we use `buffer::Usage::all()` instead?

src/backend/gl/src/device.rs
kyren: Oh!  It definitely is wrong, it used to control the explicit flush bit before I figured out that it should control the coherent flag instead, and I forgot to reverse it.

src/backend/gl/src/device.rs
kyren: Yeah, I almost did that before actually...

src/backend/gl/src/device.rs
kyren: Sure thing, that seems reasonable.

src/backend/gl/src/lib.rs
kyren: You know what, I actually did all this still with the assumption that there would only be R/W memory types this patch and not write-only ones (and leave that for a later patch).  I guess though, with the way everything else is set up, the ONLY THING left to do is just add an entry here and it will work?  I guess that's really all that's left to do so it's silly not to do it!

src/backend/gl/src/lib.rs
kyren: Oh sure thing I just forgot it existed :P

src/backend/gl/src/device.rs
kyren: And it was hiding another bug, which is now fixed 😥

kyren:
Okay all of your new comments are addressed.  Based on the discussion in gitter (I think I have this right?) I added a new memory type that is *cpu-visible*, *non-cached*, and *coherent*, so there is now also a *non-cached* gl memory type available (non-cached implies coherent, right?).

kvark:
Awesome!
Bors r+

PR #2823, opened 2019-06-12
Fix command buffer leak in the color-uniform example

### LINKED ISSUE #2818: color-uniform example drops in framerate over time
Short info header:
- GFX version: hal-0.2 (branch)
- OS: Windows 10 x64
- GPU: AMD RX580 4GB
- Backends: Vulkan

How to reproduce:
```
cd gfx/examples
cargo run --bin colour-uniform --features vulkan
```
(i have made this branch with frame time print out if you are feeling lazy https://github.com/rishflab/gfx/tree/issue-2818)

The framerate drops gradually over time when running the color-uniform example. It starts at around 150fps and after about 3 minutes it is at around 50fps. 

This only happens with the vulkan backend. The dx12 backend maintains a steady 150fps.

My guess it is due to the way the synchronisation is setup in the example. I have used the synchronisation setup from this example in my own project and I am having the same issue. I profiled it the radeon profiler and what I found was that there was a period of time with 0% gpu utilisation which was gradually increasing.

The quad example uses a different synchronisation setup and does not suffer from this issue.

Tested on 2014 Macbook pro:
I tried the example on a 2014 macbook pro (Mac OSX, metal, nvidia gt750m) and it works fine.

kvark:
Fixes #2818 
PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -528,7 +528,7 @@ impl<B: Backend> RendererState<B> {
                 .framebuffer
                 .get_frame_data(Some(frame as usize), Some(sem_index));
 
-            let (framebuffer_fence, framebuffer, command_pool) = fid.unwrap();
+            let (framebuffer_fence, framebuffer, command_pool, command_buffers) = fid.unwrap();
             let (image_acquired, image_present) = sid.unwrap();
 
             unsafe {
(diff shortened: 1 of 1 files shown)


kvark:
Bors r=omni-viral

examples/colour-uniform/main.rs
```diff
@@ -1639,7 +1649,13 @@ impl<B: Backend> Drop for FramebufferState<B> {
                 device.destroy_fence(fence);
             }
 
-            for command_pool in self.command_pools.take().unwrap() {
+            for (mut command_pool, comamnd_buffer_list) in self.command_pools
```
msiglreith: minor: this change isn't required as command pool destruction should implicitly free all associated command buffers if we follow vulkan semantics

PR #2824, opened 2019-06-12
Update Glutin to 0.21.0

### LINKED ISSUE #2757: Update glutin to 0.21
glutin was updated to 0.20 with https://github.com/gfx-rs/gfx/pull/2756 to be able to use the latest winit version.

glutin 0.21 is the current release at the moment, there were quite a few changes.

See:
https://github.com/rust-windowing/glutin/blob/master/CHANGELOG.md
https://gentz.rocks/posts/glutin-v0-21-0-migration-guide/

arashikou:
Fixes #2757

Note that this represents a significant change to the public API of the OpenGL backend. In addition to the various type changes needed due to changes in Glutin 0.21:

1) In Glutin 0.21, `make_current()` consumes the `Context` struct. So it is not possible for the types in `gfx-backend-gl::window::glutin` to make themselves the current context anymore. After this PR, the calling code must make the context current before instantiating any gfx windowing types.

1) `gfx-backend-gl::window::glutin::Headless` must now carry a Starc because only _non_-current contexts implement `Send` and `Sync` in Glutin 0.21, but only _current_ contexts can be used to implement `hal::Instance`. To help with constructing a `Headless`, I have added `Headless::from_context()`.

1) `get_window()` has been renamed `get_context()` to reflect that, in Glutin 0.21, `glutin::WindowedContext` dereferences to `glutin::Context` rather than `glutin::Window` and additional steps must be taken to access the `Window`.

PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -105,13 +105,13 @@ const COLOR_RANGE: i::SubresourceRange = i::SubresourceRange {
 
 trait SurfaceTrait {
     #[cfg(feature = "gl")]
-    fn get_window_t(&self) -> &back::glutin::WindowedContext;
+    fn get_window_t(&self) -> &back::glutin::WindowedContext<back::glutin::PossiblyCurrent>;
 }
 
 impl SurfaceTrait for <back::Backend as hal::Backend>::Surface {
     #[cfg(feature = "gl")]
-    fn get_window_t(&self) -> &back::glutin::WindowedContext {
-        self.get_window()
+    fn get_window_t(&self) -> &back::glutin::WindowedContext<back::glutin::PossiblyCurrent> {
+        self.get_context()
     }
 }
 
@@ -387,7 +387,7 @@ impl<B: Backend> RendererState<B> {
                             winit::WindowEvent::Resized(dims) => {
                                 #[cfg(feature = "gl")]
                                 backend.surface.get_window_t().resize(dims.to_physical(
-                                    backend.surface.get_window_t().get_hidpi_factor(),
+                                    backend.surface.get_window_t().window().get_hidpi_factor(),
                                 ));
                                 recreate_swapchain = true;
                             }
@@ -685,12 +685,11 @@ fn create_backend(window_state: &mut WindowState) -> (BackendState<back::Backend
         let builder =
             back::config_context(back::glutin::ContextBuilder::new(), ColorFormat::SELF, None)
                 .with_vsync(true);
-        back::glutin::WindowedContext::new_windowed(
+        let context = builder.build_windowed(
             window_state.wb.take().unwrap(),
-            builder,
             &window_state.events_loop,
-        )
-        .unwrap()
+        ).unwrap();
+        unsafe { context.make_current() }.expect("Unable to make context current")
     };
 
     let surface = back::Surface::from_window(window);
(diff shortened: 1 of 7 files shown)


arashikou:
One design decision I'd particularly welcome input on is that none of the types in `gfx-backend-gl::window::glutin` support being used with `glutin::Context`s that are `glutin::NotCurrent`. Non-current contexts are basically useless for anything until they're made current. The only value I could find in storing `NotCurrent` Contexts was making the containing type sendable to another thread, but the OpenGL backend already uses `Starc` for that. But if I am missing some reason support for these would be important, please let me know.

On a similar note, I am concerned that the solution I've come up with here unduly hampers apps' ability to use multiple `Context`s. Because of the way making a context current/not-current in 0.21 consumes the struct and potentially creates a new one and/or changes its type, it's awkward to store them in other structs unless you have mutable access to move things in/out of the struct. This is further complicated because the existing code needs to make the `Context` current inside functions where `self` is not mutable.

I started building a solution by making these types store an `Option` inside a `Cell` inside a `Starc`, but it feels dirty to me, and it's not clear to me if this is even a supported way to use gfx-backend-gl. Could someone who is better-versed in gfx tell me if this is even a valid use case to worry about?

kvark:
@arashikou sorry for letting this sit for a bit. It's a complex piece to review (for me), and just occurred at the same time as our Mozilla all hands. We'll get this sorted out ASAP.

kvark:
> Could someone who is better-versed in gfx tell me if this is even a valid use case to worry about?

Well, we definitely need to support GL operations on a device with `&self` borrowing. And that requires GL context to be active. So there needs to be *some* way to make it active, likely by utilizing some form of internal mutability (like `RefCell`).

examples/colour-uniform/main.rs
```diff
@@ -105,13 +105,13 @@ const COLOR_RANGE: i::SubresourceRange = i::SubresourceRange {
 
 trait SurfaceTrait {
     #[cfg(feature = "gl")]
-    fn get_window_t(&self) -> &back::glutin::WindowedContext;
+    fn get_window_t(&self) -> &back::glutin::WindowedContext<back::glutin::PossiblyCurrent>;
```
kvark: It feels like someone has toyed with type classes a bit too much... (not you, I'm talking about glutin authors)

arashikou: Yeah, seconded. 😅

arashikou:
@kvark No worries! It was a complex piece for me to understand well enough to write, so I know it's not easy to review. Is there anything I could do to make reviewing easier next time? I'm pretty new to OSS and open to advice on how to effectively contribute.

Thank you for taking a look at this. Just to be sure I've been clear: There is no way to enforce that the Context is _active_. At best, we can require that it be _possibly_ active. So right now, the code as I've written it assumes the caller made the context active before passing it to gfx and that it will never become inactive. i.e. There will never be any other contexts but this one. That much is the minimum necessary to get the tests passing.

But the tests never have multiple contexts active at once. Having multiple contexts means continually deactivating and reactivating them as you switch between them. If we want to allow for the possibility of the context becoming inactive and gfx needing to reactivate it, then we will have to do something with internal mutability.

kvark:
bors got stuck?
bors r+

arashikou:
Bors has been pending on this PR since I submitted it, TBH.

kvark:
bors retry

kvark:
Bors retry

PR #2828, opened 2019-06-13
MacOS vulkan portability support

### LINKED ISSUE #2796: Vulkan compile fails on OS X
I just pulled the latest Vulkan SDK for OS X (1.1.106.0) from here:
https://vulkan.lunarg.com/sdk/home

and attempted to compile it and got the following error:

```
$ cargo run --bin quad --features=vulkan
   Compiling gfx-backend-vulkan v0.2.0 (/Users/andrewl/rust/gfx/src/backend/vulkan)
error[E0463]: can't find crate for `x11`
  --> src/backend/vulkan/src/lib.rs:22:1
   |
22 | extern crate x11;
   | ^^^^^^^^^^^^^^^^^ can't find crate
```

Thanks.

kvark:
Fixes #2796 
PR checklist:
- [x] tested examples with the following backends: Vulkan, Metal

Users need to ensure that Vulkan loader is available and ICD paths are set up. I launched the example with the following command line:
```bash
LD_LIBRARY_PATH=<vulkan_sdk>/macOS/lib/ VK_ICD_FILENAMES=<gfx-portability>/libportability-icd/portability-macos-debug.json cargo run --bin quad --features vulkan
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 # Change Log
 
+### backend-vulkan-0.2.2 (14-06-2019)
+  - allow building on macOS for Vulkan Portability
+
 ### backend-metal-0.2.1 (14-06-2019)
   - fixed memory leaks in render pass descriptors and function strings
 
 ### hal-0.2.1 (10-06-2019)
   - `Debug` implementations
+
+### backend-vulkan-0.2.1 (23-05-2019)
+  - fix `VK_EXT_debug_utils` check at startup
 
 ## hal-0.2.0 (10-05-2019)
   - pipeline cache support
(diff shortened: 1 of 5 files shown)


kvark:
Note: I don't think we should rush into adding iOS support here, since the code path is aimed only to get the SDK validation output. For iOS, the only thing that would change is exposed device geometry (similar to running on a different macOS device), which can be simulated with [devsim](https://vulkan.lunarg.com/doc/view/1.0.61.0/windows/device_simulation.html).

kvark:
Also, moving forward, this code would be using the *real* VkPI WSI extension (instead of the MVK variant). We just need to wait for Ash to catch up (cc @MaikKlein ). See [VK_ext_metal_surface](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VK_EXT_metal_surface.html).

kvark:
uh, we need to have Makefile aware of the iOS/macOS and only enable Vulkan on the latter...

kvark:
bors r=msiglreith

PR #2829, opened 2019-06-14
[mtl] fix leaked render pass descriptor data

kvark:
Fixes #2647 
PR checklist:
- [x] `make` succeeds (on *nix)
- [x] `make reftests` succeeds
- [ ] tested examples with the following backends:
- [ ] `rustfmt` run on changed code

PR #2830, opened 2019-06-14
[mtl] wrap render pass configuration into an AR pool

kvark:
Sibling of #2829
PR checklist:


kvark:
reviewed for 0.2
bors r+

PR #2831, opened 2019-06-14
[0.2/mtl] metal-rs update

kvark:
Fixes the string leaking.
PR checklist:
- [ ] tested examples with the following backends: Metal

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -28,7 +28,7 @@ copyless = "0.1"
 log = { version = "0.4" }
 winit = { version = "0.19", optional = true }
 dispatch = { version = "0.1", optional = true }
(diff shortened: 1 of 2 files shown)


kvark:
On colour-uniform example, the only remaining leak I see is CFString coming from `winit`, presumably.

aleksijuvani:
Can we get this on the `master` branch as well?

kvark:
Help would be great! I'll only be able to do this later tonight, it's on my list.

aleksijuvani:
There's no rush, just wanted to make sure this hadn't slipped through the cracks. Won't have time for this today, but I can take care of it tomorrow if you haven't gotten to it by then!

kvark:
Please do! This and #2828 are the two things that need porting to `master`.

PR #2833, opened 2019-06-15
Fix incorrect alignment assumptions in create_shader_module

Ralith:
The Vulkan spec [mandates](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkShaderModuleCreateInfo.html):

> pCode must be a valid pointer to an array of codeSize / 4 uint32_t values

The address of some random `&[u8]` does not satisfy this for the obvious alignment reasons. Additionally, most other backends used unsafe code to transmute `&[u8]` to `&[u32]`, which is similarly UB. This is a breaking change.

It'd be nice to have somewhere to factor the `read_spv` helper out to.

PR checklist:
  - vulkan

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -42,7 +42,7 @@ struct Dimensions<T> {
 }
 
 use std::cell::RefCell;
-use std::io::{Cursor, Read};
+use std::io::Cursor;
 use std::mem::size_of;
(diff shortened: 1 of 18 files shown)


Ralith:
Moved `read_spv` to `hal::pso::read_spirv` and added a top-level reexport; let me know if that's not preferred. Metal fix attempted.

PR #2838, opened 2019-06-16
[vk] create `CAMetalLayer` for `NSView` if it doesn't exist

aleksijuvani:
Fixes `VK_ERROR_INITIALIZATION_FAILED: vkCreate_PLATFORM_SurfaceMVK(): On-screen rendering requires a layer of type CAMetalLayer` with MoltenVK.

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/window.rs
+++ b/src/backend/vulkan/src/window.rs
@@ -230,8 +230,38 @@ impl Instance {
     #[cfg(target_os = "macos")]
     pub fn create_surface_from_nsview(&self, view: *mut c_void) -> Surface {
         use ash::extensions::mvk;
-        use core_graphics::geometry::CGRect;
-        use objc::runtime::Object;
+        use core_graphics::{
+            geometry::CGRect,
+            base::CGFloat,
+        };
+        use objc::runtime::{Object, YES, BOOL};
+
+        // TODO: this logic is duplicated from gfx-backend-metal, refactor?
+        unsafe {
+            let view = view as *mut Object;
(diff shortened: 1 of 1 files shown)


src/backend/vulkan/src/window.rs
```diff
@@ -244,10 +243,10 @@ impl Instance {
         let surface = {
             let mac_os_loader = mvk::MacOSSurface::new(entry, &self.raw.0);
             let info = vk::MacOSSurfaceCreateInfoMVK {
-                s_type: vk::StructureType::WIN32_SURFACE_CREATE_INFO_KHR,
+                s_type: vk::StructureType::MACOS_SURFACE_CREATE_INFO_M,
                 p_next: ptr::null(),
                 flags: vk::MacOSSurfaceCreateFlagsMVK::empty(),
-                p_view: view,
+                p_view: layer as *mut c_void,
```
kvark: I don't think this is correct. Here is what the extension [documentation](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkMacOSSurfaceCreateInfoMVK.html) says:
> pView is a reference to a NSView object which will display this surface. This NSView must be backed by a CALayer instance of type CAMetalLayer.

aleksijuvani: Good catch! This is fixed now. MoltenVK didn't seem to mind.

kvark: it doesn't bother to specify its own extension properly :/

PR #2840, opened 2019-06-17
[mtl] argument buffer support

kvark:
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -31,11 +31,13 @@ use parking_lot::Mutex;
 #[cfg(feature = "dispatch")]
 use dispatch;
 
-use std::borrow::Borrow;
-use std::cell::RefCell;
-use std::ops::{Deref, Range};
-use std::sync::Arc;
-use std::{cmp, iter, mem, slice, time};
+use std::{
+    cmp, iter, mem, ptr, slice, time,
+    borrow::Borrow,
+    cell::RefCell,
+    ops::{Deref, Range},
+    sync::Arc,
+};
 
 
 const WORD_SIZE: usize = 4;
@@ -1523,6 +1525,9 @@ where
         Cmd::BindPipeline(ref pipeline_state) => {
             encoder.set_render_pipeline_state(pipeline_state.borrow());
         }
+        Cmd::UseResource { resource, usage } => {
+            encoder.use_resource(resource.as_native(), usage);
+        }
         Cmd::Draw {
             primitive_type,
             ref vertices,
@@ -1815,6 +1820,9 @@ where
         }
         Cmd::BindPipeline(ref pipeline) => {
             encoder.set_compute_pipeline_state(pipeline.borrow());
+        }
+        Cmd::UseResource { resource, usage } => {
+            encoder.use_resource(resource.as_native(), usage);
         }
         Cmd::Dispatch { wg_size, wg_count } => {
             encoder.dispatch_thread_groups(wg_count, wg_size);
@@ -3556,6 +3564,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                     ref raw,
                     offset,
                     stage_flags,
+                    ref resources,
                     ..
                 } => {
                     //Note: this is incompatible with the binding scheme below
@@ -3582,6 +3591,14 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                             buffer: AsNative::from(raw.as_ref()),
                             offset,
                         });
+                    }
+                    if stage_flags.intersects(pso::ShaderStageFlags::VERTEX | pso::ShaderStageFlags::FRAGMENT) {
+                        for used_resource in resources.values() {
+                            pre.issue(soft::RenderCommand::UseResource {
+                                resource: ptr::NonNull::new(used_resource.ptr.get()).unwrap(),
+                                usage: used_resource.usage,
+                            });
+                        }
                     }
                 }
             }
@@ -3698,6 +3715,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                     ref raw,
                     offset,
                     stage_flags,
+                    ref resources,
                     ..
                 } => {
                     if stage_flags.contains(pso::ShaderStageFlags::COMPUTE) {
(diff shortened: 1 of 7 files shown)


kvark:
I believe the logic is correct, but something is broken on SPIRV-Cross side with regards to how the argument buffer is generated:
```
struct spvDescriptorSetBuffer0
{
    sampler u_sampler [[id(1)]];
    texture2d<float> u_texture; /// MISSING ID!
};
```

kvark:
Now it's trying to set the overrides for all the contained items in order to get the `id(xxx)` generated correctly. The PR also handles array descriptors, in theory, with an exception of the combined image-sampler (for which the array semantics is troublesome).

The PR isn't *yet* setting the proper texture type on a descriptor.

The blocker now is that Metal doesn't seem to like us using `MTLDataType::Struct` when describing the buffer-type arguments...
> colour-uniform[2929:78936] -[_MTLDevice newArgumentEncoderWithArguments:structType:], line 2262: error 'Invalid argument type at index 0'

The only piece of documentation [I found](https://developer.apple.com/documentation/metal/buffers/basic_argument_buffers?language=objc) uses "float" type, but it's not appropriate to us, since there can be anything in an uniform buffer. I tried changing the type to Float4 for a test, and it does run, but the colour-uniform example shows blank (pixel shader produces NaN, presumably when accessing that darn argument buffer).

kvark:
Another thing missing here is detecting the descriptors that are incompatible with the argument buffers (like writable texture storage) and binding them separately...

kvark:
Running with Float4 type again, I'm getting this validation error:
> indexRangeIsValueType:273: failed assertion `Trying to set a Buffer at index 0 but the argument buffer has a Constant at this index'

kvark:
I resolved the data type problem. It needs to be `Pointer`. Another crucial bit I was missing is keeping track of used resources and resetting them at the start of any pass... this is somewhat sub-optimal, to say the least. We probably don't need to re-use them on *every* pass, especially not *all* the resources. It's yet to figure out how to reduce this to a meaningful subset.

kvark:
Here is a bigger problem, perhaps. We need to know the texture type (1D/2D/3D/etc) at the time we create the argument encoder, so in `create_descriptor_set_layout`. Vulkan only tells us that there is a texture to expect, it doesn't tell us what kind of a texture there is.

A brute force solution would be to delay the argument encoder creation to a point where we create the descriptor set itself... and of course we'd need to cache those encoders, assuming in most cases that only a single one is needed per particular descriptor set layout.

kvark:
I think I know the answer to a question about `used_resources`.  We should basically just keep a list per descriptor set binding point in the command buffer state. Once (during recording) we decide that a particular descriptor set is compatible with the current layout, we can issue the `useResource` calls for this descriptor.

kvark:
Implemented better tracking of used resources. The only unresolved issue is the texture type atm.

kvark:
I've refined this to a point where I can run vkQuake3. After a few moments of playing, stuff starts to break visibly:
Screen Shot 2019-06-19 at 22 19 35

I presume this is about us calling `useResource` incorrectly, but I haven't identified exactly where the bug is.

kvark:
With the recent fix, we can *mostly* run vkQuake. There is still a bit of visual issues, but nothing nearly as awesome as that last screen shot... I'm going to miss it.
Anyhow, argument buffer support is turned off by default now. I think we are at the point where we need to review and land this. @grovesNL would you be able to take a look?

kvark:
fwiw, rev 7c8a78c6ffefb997dac1560083f6d0b04795a8d5 is where the fun stuff is (fixed in latest):
Screen Shot 2019-06-20 at 17 42 55

kvark:
Interestingly, vkQuake3 plays totally fine without showing any vertex buffer issues, unlike vkQuake1. It is possible that the app itself has a bug (I've already fixed some in them before, one recently even) that is not caught by the validation and was obscured in our non-argument code path.

src/backend/metal/src/command.rs
```diff
@@ -1523,6 +1561,9 @@ where
         Cmd::BindPipeline(ref pipeline_state) => {
             encoder.set_render_pipeline_state(pipeline_state.borrow());
         }
+        Cmd::UseResource { resource, usage } => {
+            encoder.use_resource(resource.as_native(), usage);
```
grovesNL: We'll need to update this soon because this is deprecated now https://developer.apple.com/documentation/metal/mtlrendercommandencoder/2866168-useresource

src/backend/metal/src/command.rs
```diff
@@ -352,6 +367,16 @@ impl State {
                     .chain(iter::once(com_samplers))
                     .chain(com_push_constants)
             });
+        let com_used_resources = self.descriptor_sets
```
grovesNL: Why do we always have to do this? Should we only track used resources when arguments buffers are enabled? Or maybe this is part of the fix for #2848?

src/backend/metal/src/command.rs
```diff
@@ -194,6 +198,12 @@ struct SubpassInfo {
     formats: native::SubpassFormats,
 }
 
+#[derive(Debug, Default)]
+struct DescriptorSetInfo {
+    graphics_resources: Vec<(ResourcePtr, metal::MTLResourceUsage)>,
```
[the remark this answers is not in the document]
kvark: I think in this case a regular vector makes more sense: we don't have a particular expectation on that list other than "many or zero", and it gets reused within a command buffer, so this is what `Vec` works best for.

src/backend/metal/src/command.rs
kvark: Yeah, I saw that. Not exactly sure when the method we do call becomes unavailable though.

src/backend/metal/src/command.rs
kvark: If there are no argument buffers, these lists are empty. At the end of the day, argument buffer should be our preferred path to go, so that's what we optimize for.

kvark:
Thank you for the review @grovesNL !
It doesn't look like there are big concerns. I have a whole lot of stuff piled on top, and this functionality is disabled by default, so we are safe to continue.
bors r=grovesNL

kvark:
Interesting, bors appears to be stuck?..
Last status is green, but bors didn't get the memo - https://travis-ci.org/gfx-rs/gfx/builds/549781361?utm_source=github_status&utm_medium=notification

PR #2851, opened 2019-06-21
Only set persistent | coherent map flags when we have glBufferStorage

kyren:
Also disables coherent memory types when glBufferStorage is not available

Fixes #issue
PR checklist:


kyren:
This makes OpenGL <4.4 platforms *with* mapping somewhat less "capable" than those without mapping at all, as emulated mapping *does* emulate persistent mapping.

PR #2852, opened 2019-06-21
Quite a few gl backend fence fixes

kyren:
~~gl fences can now be in three states, raw (meaning they are backed by a raw
fence which is signaled or unsignaled), fake-signaled, and fake-unsignaled.
When creating a new fence, fences will start in the fake-unsignaled or
fake-signaled fence depending on how they are created.  waiting on a
fake-unsignaled fence will turn it into an unsignaled real raw fence and then
that and any further wait will wait on the real fence.~~

~~Before, fences created in the 'signaled' state would be created with a
raw *unsignaled* fence which was wrong, and reset fences would be set to None,
causing incorrect panics on future fence operations.  This change should fix
both these issues.~~

I wasn't thinking clearly about how fences were supposed to work, the gl fence implementation now mimics the metal backend, which is vastly more correct.

Additionally, on wasm32, a replacement `Device::wait_fences` is added which does
not use `thread::sleep` or rely on `std::time` (which are unavailable in wasm32).

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1610,7 +1610,7 @@ impl d::Device<B> for Device {
         let performance = web_sys::window().unwrap().performance().unwrap();
         let start = performance.now();;
         let get_elapsed = || {
-            ((performance.now() - start) * 1000.0) as u64
+            ((performance.now() - start) * 1_000_000.0) as u64
         };
 
         match wait {
@@ -1629,10 +1629,12 @@ impl d::Device<B> for Device {
                 Ok(true)
             }
             d::WaitFor::Any => {
+                const FENCE_WAIT_NS: u64 = 100_000;
+
                 let fences: Vec<_> = fences.into_iter().collect();
                 loop {
                     for fence in &fences {
-                        if self.wait_for_fence(fence.borrow(), 1000)? {
+                        if self.wait_for_fence(fence.borrow(), FENCE_WAIT_NS)? {
                             return Ok(true);
                         }
                     }
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1584,7 +1584,11 @@ impl d::Device<B> for Device {
                         }
                         Ok(false)
                     }
-                    _ => Ok(true),
+                    glow::CONDITION_SATISFIED | glow::ALREADY_SIGNALED => {
+                        fence.0.set(n::FenceInner::Idle { signaled: true });
+                        Ok(true)
+                    },
+                    _ => unreachable!(),
                 }
             }
             n::FenceInner::Pending(None) => {
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1543,31 +1543,13 @@ impl d::Device<B> for Device {
         Ok(n::Semaphore)
     }
 
-    fn create_fence(&self, signalled: bool) -> Result<n::Fence, d::OutOfMemory> {
-        let sync = if signalled && self.share.private_caps.sync {
-            let gl = &self.share.context;
-            Some(unsafe { gl.fence_sync(glow::SYNC_GPU_COMMANDS_COMPLETE, 0).unwrap() })
-        } else {
-            None
-        };
-        Ok(n::Fence::new(sync))
+    fn create_fence(&self, signaled: bool) -> Result<n::Fence, d::OutOfMemory> {
+        let cell = Cell::new(n::FenceInner::Idle { signaled });
+        Ok(n::Fence(cell))
     }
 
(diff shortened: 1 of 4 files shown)


kyren:
Sorry for American spelling of 'signaled', if you'd rather standardize on the British spelling I can change it.

kyren:
Also, I'm not 100% sure of the previous fence logic, it seemed really wrong so there's always the possibility that I'm misunderstanding the OpenGL api... again 😛

src/backend/gl/src/device.rs
```diff
@@ -1744,9 +1795,21 @@ pub(crate) fn wait_fence(fence: &n::Fence, share: &Starc<Share>, timeout_ns: u64
     let gl = &share.context;
     unsafe {
         if share.private_caps.sync {
+            let sync = match fence.0.get() {
+                n::FenceState::Raw(sync) => sync,
+                n::FenceState::Unsignaled => {
+                    let sync = gl.fence_sync(glow::SYNC_GPU_COMMANDS_COMPLETE, 0).unwrap();
```
kvark: This seems wrong. I read it as "if user wants to wait for a fence that has Unsignaled state, we just wait for GPU to complete the operation". This is wrong because what is desired here is to wait for the signal state of the fence, not for GPU to complete commands. Fences can be scheduled to signal after submission or image acquisition. So these are the places were I'd expect `gl.fence_sync()` call to be inserted, and the fence to be initialized with the `Raw` variant.

If we follow this logic, in `wait_fence` we expect the fence to be either `Raw` or `Signaled`, since otherwise with `Unsignaled` there is no ability for it to ever turn to signaled. Interestingly, this matches Metal backend logic: https://github.com/gfx-rs/gfx/blob/a0dfe9ca10c28437875081ddf45cbe6a561cc02e/src/backend/metal/src/device.rs#L2509

As a side note, you can follow the Metal backend and declare the fence as:
```rust
pub enum FenceInner {
    Idle { signaled: bool },
    Pending(<GlContext as glow::Context>::Fence),
}
```

kyren: Aha, yeah I wasn't thinking about this correctly, and I like the way the Metal backend does this so I'll follow that pattern.

So, okay what state should the fence be after being reset?  Should it be a new unsignaled raw fence?  Why does the metal API have an `Idle { signaled: false }` state?

src/backend/gl/src/device.rs
```diff
@@ -1568,26 +1550,100 @@ impl d::Device<B> for Device {
         fence: &n::Fence,
         timeout_ns: u64,
     ) -> Result<bool, d::OomOrDeviceLost> {
-        if !self.share.private_caps.sync {
-            return Ok(true);
```
kyren: So one thing this patch does right now is *remove* this line.  I am not sure why this is there, but the behavior beforehand was *not* actually to call `glFlush` on waiting for a fence with no sync capability due to this early return.  If you want, I can keep this behavior and just eliminate `FenceInner::Pending(None)` as a state entirely (without sync capability, fences could just be created `FenceInner::Idle { signaled: true }`.

src/backend/gl/src/device.rs
```diff
@@ -1568,26 +1550,100 @@ impl d::Device<B> for Device {
...
+                    timeout_ns as i32,
+                ) {
+                    glow::TIMEOUT_EXPIRED => Ok(false),
+                    glow::WAIT_FAILED => {
+                        if let Err(err) = self.share.check() {
+                            error!("Error when waiting on fence: {:?}", err);
+                        }
+                        Ok(false)
+                    }
+                    _ => Ok(true),
```
kyren: Would it be faster / more correct here to transition the fence here to `FenceStatus::Idle { true }` to avoid future `glClientWaitSync` calls?  The metal backend did not do this so I also didn't do it.

src/backend/gl/src/device.rs
kvark: I think this is too late for this check anyway. If "sync" capability is not available, we'd not be able to create a `Pending` sync in the first place. So removing the check from here is fine.

src/backend/gl/src/device.rs
kvark: we should replace `_` by a real arm, and leave `_` for unexpected stuff. And yes, we should change it to `Idle { signaled: true }` both here and in Metal

src/backend/gl/src/device.rs
kyren: With this patch, if "sync" is not available the gl backend will create a `FenceInner::Pending(None)` in submit, so it is possible to create a `FenceInner::Pending(None)` and hit the code below this (which is not from this patch).  Beforehand, I *think* it was not possible to hit the code below.

src/backend/gl/src/device.rs
kyren: That's from previosly, but yeah I agree and I will change it.

PR #2853, opened 2019-06-22
[0.2] Backport renderpass resolve attachments fix

rukai:
Backport of https://github.com/gfx-rs/gfx/pull/2813/files

I am able to get MSAA working on wgpu only with these changes.

Includes a new minor release, is there a process I should be aware of here?
The new release is needed because it is impractical to get wgpu to use these changes without a new release because of rendy's dependency on gfx-hal.
Simply using a specific git commit fails because rendy and wgpu both need to refer to the same trait from gfx-hal.

Fixes #issue
PR checklist:
- ~~[ ] `rustfmt` run on changed code~~

### DIFF
--- a/src/backend/vulkan/Cargo.toml
+++ b/src/backend/vulkan/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-backend-vulkan"
-version = "0.2.2"
+version = "0.2.3"
 description = "Vulkan API backend for gfx-rs"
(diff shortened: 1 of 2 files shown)


rukai:
Looks like a `cargo publish` needs to be manually run?

kvark:
@msiglreith could you publish it please?

kvark:
https://crates.io/crates/gfx-backend-vulkan/0.2.3

What we are missing still is a CHANGELOG update (just not to forget...)

PR #2854, opened 2019-06-22
More carefully set blend settings in the gl backend

kyren:
This should enable the gl backend to support multiple draw buffers with all the
same blend settings even if glColorMaski / glBlendEquationi / glBlendFunci are
not available.

(Edit: also it should fix several problems, blend settings do not appear to work at all without this even with a single output target if glColorMaski / glBlendEquationi / glBlendFunci are not available)

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -325,7 +325,7 @@ impl RawCommandBuffer {
         slice
     }
 
-    fn update_blend_targets(&mut self, blend_targets: &Vec<pso::ColorBlendDesc>) {
+    fn update_blend_targets(&mut self, blend_targets: &[pso::ColorBlendDesc]) {
         let max_blend_slots = blend_targets.len();
         if max_blend_slots == 0 {
             return;
@@ -335,13 +335,7 @@ impl RawCommandBuffer {
             self.cache.blend_targets.resize(max_blend_slots, None);
         }
 
-        let mut all_targets_same = true;
-        for blend_target in &blend_targets[1..] {
-            if blend_target != &blend_targets[0] {
-                all_targets_same = false;
-                break;
-            }
-        }
+        let all_targets_same = blend_targets[1..].iter().all(|target| target == &blend_targets[0]);
 
         if all_targets_same {
             let mut update_blend = false;
@@ -356,14 +350,10 @@ impl RawCommandBuffer {
             }
         } else {
             for (slot, blend_target) in blend_targets.iter().enumerate() {
-                let cached_target = self.cache.blend_targets.get_mut(slot).unwrap();
+                let cached_target = &mut self.cache.blend_targets[slot];
                 let update_blend = match cached_target {
-                    Some(cache) => {
-                        cache != blend_target
-                    }
-                    None => {
-                        true
-                    }
+                    Some(cache) => cache != blend_target,
+                    None => true,
                 };
 
                 if update_blend {
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -108,7 +108,7 @@ pub enum Command {
     SetPatchSize(i32),
     BindProgram(<GlContext as glow::Context>::Program),
     BindBlendSlot(ColorSlot, pso::ColorBlendDesc),
-    BindAllBlendSlots(pso::ColorBlendDesc),
+    SetAllBlendSlots(pso::ColorBlendDesc),
     BindAttribute(n::AttributeDesc, n::RawBuffer, i32, u32),
     //UnbindAttribute(n::AttributeDesc),
     CopyBufferToBuffer(n::RawBuffer, n::RawBuffer, command::BufferCopy),
@@ -346,11 +346,13 @@ impl RawCommandBuffer {
                 }
             }
             if update_blend {
-                self.push_cmd(Command::BindAllBlendSlots(blend_targets[0]));
+                self.push_cmd(Command::SetAllBlendSlots(blend_targets[0]));
             }
         } else {
-            for (slot, blend_target) in blend_targets.iter().enumerate() {
-                let cached_target = &mut self.cache.blend_targets[slot];
+            for (slot, (blend_target, cached_target)) in blend_targets
+                .iter()
+                .zip(&mut self.cache.blend_targets)
+                .enumerate() {
                 let update_blend = match cached_target {
                     Some(cache) => cache != blend_target,
                     None => true,
@@ -358,7 +360,12 @@ impl RawCommandBuffer {
 
                 if update_blend {
                     *cached_target = Some(*blend_target);
-                    self.push_cmd(Command::BindBlendSlot(slot as _, *blend_target));
+                    push_cmd_internal(
+                        &self.id,
+                        &mut self.memory,
+                        &mut self.buf,
+                        Command::BindBlendSlot(slot as _, *blend_target)
+                    );
                 }
             }
         }
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -108,6 +108,7 @@ pub enum Command {
     SetPatchSize(i32),
     BindProgram(<GlContext as glow::Context>::Program),
     BindBlendSlot(ColorSlot, pso::ColorBlendDesc),
+    BindAllBlendSlots(pso::ColorBlendDesc),
     BindAttribute(n::AttributeDesc, n::RawBuffer, i32, u32),
     //UnbindAttribute(n::AttributeDesc),
     CopyBufferToBuffer(n::RawBuffer, n::RawBuffer, command::BufferCopy),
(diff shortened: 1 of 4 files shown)


kyren:
I *think* this PR is basically in the right direction, but I'm not sure of some of the details.

src/backend/gl/src/command.rs
```diff
@@ -108,6 +108,7 @@ pub enum Command {
     SetPatchSize(i32),
     BindProgram(<GlContext as glow::Context>::Program),
     BindBlendSlot(ColorSlot, pso::ColorBlendDesc),
+    BindAllBlendSlots(pso::ColorBlendDesc),
```
kyren: Is this the right name, its not really binding...

src/backend/gl/src/command.rs
```diff
@@ -326,44 +327,44 @@ impl RawCommandBuffer {
...
+                all_targets_same = false;
+                break;
+            }
+        }
+
+        if all_targets_same {
+            for cached_target in &mut self.cache.blend_targets {
+                *cached_target = Some(blend_targets[0]);
+            }
+            self.push_cmd(Command::BindAllBlendSlots(blend_targets[0]));
```
kyren: I guess this is wrong because it shouldn't set the blend state again if all of the cached targets also match... really the cache state should be in one of two states (all set to X settings, per-targets set to X settings), but that would require potentially dropping and re-creating a Vec or getting pretty messy.  I think I can at least avoid this command if all of the blend slots match already.

src/backend/gl/src/command.rs
```diff
@@ -326,44 +327,50 @@ impl RawCommandBuffer {
...
-                            if cache != blend_target {
-                                update_blend = true;
-                            }
-                        }
-                        &None => {
-                            update_blend = true;
-                        }
-                    }
+            for cached_target in &mut self.cache.blend_targets {
+                if cached_target.as_ref() != Some(&blend_targets[0]) {
```
kyren: *theoretically*, this could cache the case where there's more render buffers now than there were before, but before settings were changed across all render buffers so we wouldn't need to set the settings on the new render buffers to match it.  It does not do that currently.

src/backend/gl/src/info.rs
```diff
@@ -469,6 +471,11 @@ pub(crate) fn query_all(gl: &GlContainer) -> (Info, Features, LegacyFeatures, Li
         legacy |= LegacyFeatures::INSTANCED_ATTRIBUTE_BINDING;
     }
 
+    let per_draw_buffer_blending = info.is_supported(&[Core(4, 0), Es(3, 2)]) && !info.is_webgl();
+    if per_draw_buffer_blending {
+        features |= Features::INDEPENDENT_BLENDING;
```
kyren: this wasn't set before, so that makes me think maybe there's some reason it should't be set now?

kyren:
Hmm, hold on, the cache behavior for the all targets cace still isn't right...

src/backend/gl/src/command.rs
```diff
@@ -326,44 +327,50 @@ impl RawCommandBuffer {
 
     fn update_blend_targets(&mut self, blend_targets: &Vec<pso::ColorBlendDesc>) {
```
kvark: should be `&[pso::ColorBlendDesc]`

src/backend/gl/src/command.rs
```diff
@@ -326,44 +327,50 @@ impl RawCommandBuffer {
...
-                    self.cache.blend_targets = Some(vec![None; max_blend_slots]);
-                }
-            };
+        if self.cache.blend_targets.len() < max_blend_slots {
+            self.cache.blend_targets.resize(max_blend_slots, None);
         }
 
-        for (slot, blend_target) in blend_targets.iter().enumerate() {
+        let mut all_targets_same = true;
+        for blend_target in &blend_targets[1..] {
```
kvark: should use iterators here, i.e. `let all_targets_same = blend_targets[1..].iter().all(|target| target == blend_targets[0])`

src/backend/gl/src/command.rs
```diff
@@ -326,44 +327,50 @@ impl RawCommandBuffer {
...
+                    *cached_target = Some(blend_targets[0]);
+                    update_blend = true;
                 }
+            }
+            if update_blend {
+                self.push_cmd(Command::BindAllBlendSlots(blend_targets[0]));
+            }
+        } else {
+            for (slot, blend_target) in blend_targets.iter().enumerate() {
+                let cached_target = self.cache.blend_targets.get_mut(slot).unwrap();
```
kvark: nit: let's iterate both zipped, i.e. `for (cached, blend_target) = self.cache.blend_targets.iter_mut().zip(blend_targets) {..}`

src/backend/gl/src/info.rs
```diff
@@ -469,6 +471,11 @@ pub(crate) fn query_all(gl: &GlContainer) -> (Info, Features, LegacyFeatures, Li
         legacy |= LegacyFeatures::INSTANCED_ATTRIBUTE_BINDING;
     }
 
+    let per_draw_buffer_blending = info.is_supported(&[Core(4, 0), Es(3, 2)]) && !info.is_webgl();
```
kvark: should add "EXT_draw_buffers2" extension here

src/backend/gl/src/command.rs
kyren: yeah, definitely, I can change that with this PR

src/backend/gl/src/command.rs
kyren: Yeah I definitely should do that, wasn't thinking

src/backend/gl/src/info.rs
kyren: Good to know, will do that!

src/backend/gl/src/command.rs
kyren: That would be better, except:
1) you still need the slot later so you also have to enumerate (which is easy enough, but it might be a bit big for a single for statement)
2) there's a NLL trick going on where `self.cache.blend_targets` is borrowed for a sub-range of the inner part of the loop.  If I `iter_mut` over it then I borrow part of `self` across `self.push_cmd`.  Ideally we wouldn't, but it wasn't immediately obvious how to do it.

kvark: I don't think (2) is the case here. For (1) it's fine to split into multiple lines if it becomes long,i.e.
```rust
for (slot, (cached, blend_target)) in self.cache.blend_targets
  .iter_mut()
  .zip(blend_targets.iter())
  .enumerate()
{
///
}
```

kyren: 2 is right, that line you just typed will have a borrowck error because mutable iterator holds a borrow of `blend_targets` across the body of the for loop, which includes the `push_cmd` method call:

```
error[E0499]: cannot borrow `*self` as mutable more than once at a time
   --> src/backend/gl/src/command.rs:368:21
    |
352 |               for (slot, (cached_target, blend_target)) in self.cache.blend_targets
    |                                                            ------------------------
    |                                                            |
    |  __________________________________________________________first mutable borrow occurs here
    | |
353 | |                 .iter_mut()
354 | |                 .zip(blend_targets.iter())
355 | |                 .enumerate()
    | |____________________________- first borrow later used here
...
368 |                       self.push_cmd(Command::BindBlendSlot(slot as _, *blend_target));
    |                       ^^^^ second mutable borrow occurs here

```

But yeah, you're right the long for loop is fine, really, other than the borrowck error.

kyren: One thing, I have no idea why I wrote:

`let cached_target = self.cache.blend_targets.get_mut(slot).unwrap();`

I think it was an artifact of how it was written previously, but in any case I replaced that with the more obvious version.

src/backend/gl/src/command.rs
kvark: good point. Let's rename to `SetXxx`

src/backend/gl/src/command.rs
```diff
@@ -324,46 +325,42 @@ impl RawCommandBuffer {
...
+                let cached_target = &mut self.cache.blend_targets[slot];
+                let update_blend = match cached_target {
+                    Some(cache) => cache != blend_target,
+                    None => true,
+                };
 
                 if update_blend {
-                    cached_targets[slot] = Some(*blend_target);
+                    *cached_target = Some(*blend_target);
+                    self.push_cmd(Command::BindBlendSlot(slot as _, *blend_target));
```
kvark: could you consider using `push_cmd_internal` instead of `self.push_cmd` to avoid `self` borrowing conflict with `cache.blend_targets`?

src/backend/gl/src/info.rs
kvark: this looks correct to me

src/backend/gl/src/command.rs
kyren: oh, good point

src/backend/gl/src/command.rs
```diff
@@ -108,6 +108,7 @@ pub enum Command {
     SetPatchSize(i32),
     BindProgram(<GlContext as glow::Context>::Program),
     BindBlendSlot(ColorSlot, pso::ColorBlendDesc),
```
kyren: Did you want both BindAllBlendSlots *and* BindBlendSlot changed to SetXXX?

kvark: yes, thank you!

kyren:
Okay, I'm done now, sorry about all the bors chaos 😅

PR #2855, opened 2019-06-22
Fall back to f32 depth range on gl backend

kyren:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/src/queue.rs
+++ b/src/backend/gl/src/queue.rs
@@ -427,8 +427,8 @@ impl CommandQueue {
                         if self.share.private_caps.depth_range_f64_precision {
                             gl.depth_range_f64(depth_range[0], depth_range[1]);
                         } else {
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -427,8 +427,8 @@ impl CommandQueue {
                         if self.share.private_caps.depth_range_f64_precision {
                             gl.depth_range_f64(depth_range[0], depth_range[1]);
                         } else {
-                            warn!("Depth ranges with f64 precision are not supported");
-                            // TODO: fallback to f32?
+                            debug!("Depth ranges with f64 precision are not supported, falling back to f32");
```
kvark: I wouldn't even `debug!` here, tbh

kyren: honestly I only did it because it *was* a warning 😛

PR #2856, opened 2019-06-22
Support copying buffer to 2d texture array in gl backend, copy with proper texture format

kyren:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -117,7 +117,7 @@ pub enum Command {
         texture_target: n::TextureTarget,
         texture_format: n::TextureFormat,
         pixel_type: n::DataType,
-        command: command::BufferImageCopy,
+        data: command::BufferImageCopy,
     },
     CopyBufferToSurface(n::RawBuffer, n::Surface, command::BufferImageCopy),
     CopyTextureToBuffer {
@@ -126,7 +126,7 @@ pub enum Command {
         texture_format: n::TextureFormat,
         pixel_type: n::DataType,
         dst_buffer: n::RawBuffer,
-        command: command::BufferImageCopy,
+        data: command::BufferImageCopy,
     },
     CopySurfaceToBuffer(n::Surface, n::RawBuffer, command::BufferImageCopy),
     CopyImageToTexture(n::ImageKind, n::Texture, n::TextureTarget, command::ImageCopy),
@@ -1183,7 +1183,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
                         texture_target: target,
                         texture_format: format,
                         pixel_type,
-                        command: r,
+                        data: r,
                     }
                 },
             };
@@ -1220,7 +1220,7 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
                         texture_format: format,
                         pixel_type: pixel_type,
                         dst_buffer: dst_raw,
-                        command: r,
+                        data: r,
                     }
                 },
             };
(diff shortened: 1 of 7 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -598,22 +598,46 @@ impl CommandQueue {
...
+                    glow::TEXTURE_2D => {
+                        gl.bind_texture(glow::TEXTURE_2D, Some(texture));
+                        gl.tex_sub_image_2d_pixel_buffer_offset(
+                            glow::TEXTURE_2D,
+                            r.image_layers.level as _,
+                            r.image_offset.x,
+                            r.image_offset.y,
+                            r.image_extent.width as _,
+                            r.image_extent.height as _,
+                            glow::RGBA,
```
kvark: we should fix the format here... RGBA8 should not be the only format supporting buffer transfers :)

kyren:
> Would you want to fix the formats as well here? I'm fine with landing as is.

Yeah... if it's easy to do it really needs to be fixed, let me see...

kyren:
Well, it wasn't trivial exactly, but it isn't too bad of a change.  Sorry about the PR scope creep again 😅

Also includes a drive-by rename of `TextureType` to `TextureTarget`, as that is actually the correct OpenGL term.

src/backend/gl/src/queue.rs
```diff
@@ -595,32 +595,56 @@ impl CommandQueue {
...
+                            glow::TEXTURE_2D_ARRAY,
+                            r.image_layers.level as _,
+                            r.image_offset.x,
+                            r.image_offset.y,
+                            r.image_layers.layers.start as i32,
+                            r.image_extent.width as _,
+                            r.image_extent.height as _,
+                            r.image_layers.layers.end as i32 - r.image_layers.layers.start as i32,
+                            format,
+                            glow::UNSIGNED_BYTE,
```
kvark: this should also depend on the format, i.e. 16 bit signed formats would use `glow::SHORT`, etc

kyren: OH, you're right it would...

kyren:
I couldn't in good conscience add yet another field to the `CopyBufferToTexture` / `CopyTextureToBuffer` tuple enum variants, so I also changed that...

Arguably, maybe storing the target, format, and pixel data type in the `native::ImageKind::Texture` is too much and you could store something smaller and then compute the appropriate target / format / pixel data type?

src/backend/gl/src/command.rs
```diff
@@ -111,17 +111,31 @@ pub enum Command {
     BindAttribute(n::AttributeDesc, n::RawBuffer, i32, u32),
     //UnbindAttribute(n::AttributeDesc),
     CopyBufferToBuffer(n::RawBuffer, n::RawBuffer, command::BufferCopy),
-    CopyBufferToTexture(n::RawBuffer, n::Texture, n::TextureType, command::BufferImageCopy),
+    CopyBufferToTexture {
```
kvark: great change!

src/backend/gl/src/command.rs
```diff
@@ -111,17 +111,31 @@ pub enum Command {
...
+        src_buffer: n::RawBuffer,
+        dst_texture: n::Texture,
+        texture_target: n::TextureTarget,
+        texture_format: n::TextureFormat,
+        pixel_type: n::DataType,
+        command: command::BufferImageCopy,
```
kvark: should probably be renamed to "data", since it's not a command

kvark:
@kyren fwiw, storing stuff associated with `Texture` is totally fine, since they aren't meant to be `Copy` anyway.

kyren:
Whoops, sorry I didn't see that you had already approved the PR to be merged!  I interrupted bors 😞

PR #2857, opened 2019-06-23
[mtl] better pipeline layout

### LINKED ISSUE #2848: Problem with compatible layouts on Metal
I discovered a scenario that is technically valid in gfx-hal and Vulkan, but would cause big troubles for our Metal backend:
  1. user sets pipeline A of layout X
  2. user sets vertex buffer V. The actual Metal buffer index is chosen to be right after X bindings.
  3. user binds descriptor sets for pipeline layout Y != X:
    - if Y is a superset of X, then the bindings could overwrite our vertex buffer V
    - if Y is incompatible, we expect there to `bind_graphics_pipeline` to follow with a compatible pipeline, but we still need to realize that the vertex buffers are thrashed...

We need to find a solution that doesn't involve checks at the draw()/dispatch() frequency level.

kvark:
Fixes #2848
Basic approach is described in the comments. We put push constants first, descriptors later, and vertex buffers at the end.

Based on #2840
Actually, it makes it impossible to use argument buffers together with push constants. We need a way to set argument buffer overrides in spirv-cross to fix that.

PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -922,6 +922,55 @@ impl hal::Device<Backend> for Device {
         let mut res_overrides = BTreeMap::new();
         let mut infos = Vec::new();
 
+        // First, place the push constants
+        let mut pc_buffers = [None; 3];
+        let mut pc_limits = [0u32; 3];
+        for pcr in push_constant_ranges {
+            let (flags, range) = pcr.borrow();
+            for (limit, &(stage_bit, _, _)) in pc_limits.iter_mut().zip(&stage_infos) {
+                if flags.contains(stage_bit) {
+                    *limit = range.end.max(*limit);
+                }
+            }
+        }
+
+        const LIMIT_MASK: u32 = 3;
+        // round up the limits alignment to 4, so that it matches MTL compiler logic
+        //TODO: figure out what and how exactly does the alignment. Clearly, it's not
+        // straightforward, given that value of 2 stays non-aligned.
+        for limit in &mut pc_limits {
+            if *limit > LIMIT_MASK {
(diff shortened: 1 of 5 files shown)


kvark:
Interestingly, since WebGPU doesn't have push constants yet, we may try enabling argument buffers there even after this change lands and see if it makes a difference.

src/backend/metal/src/device.rs
```diff
@@ -1351,8 +1349,8 @@ impl hal::Device<Backend> for Device {
                     vertex_buffers.alloc().init((original.clone(), base_offset));
                     vertex_buffers.len() - 1
                 });
-            let mtl_buffer_index = attribute_buffer_index as usize + relative_index;
-            if mtl_buffer_index >= self.shared.private_caps.max_buffers_per_stage as usize {
+            let mtl_buffer_index = self.shared.private_caps.max_buffers_per_stage - 1 - (relative_index as ResourceIndex);
+            if mtl_buffer_index < pipeline_layout.total.vs.buffers {
```
grovesNL: Could you explain this condition? I expected to test `mtl_buffer_index < 0` here but maybe I'm misunderstanding

kvark: We are allocating the vertex buffers from top to bottom. It starts at 31, then goes down, but we can't get lower than `pipeline_layout.total.vs.buffers` since this is where the pipeline layout bindings (for resources and push constants) end.

kvark:
Thank you for the review!
bors r=grovesNL

src/backend/metal/src/device.rs
grovesNL: :+1: This makes sense, thanks! I thought we were also including the vertex buffers in that count but I see the distinction now (for the pipeline layout)

PR #2864, opened 2019-06-24
[mtl] Events

### LINKED ISSUE #2809: Implement Event on Apple platforms
This is a follow up to #2804, which would allow us to switch `gfx-portability` to use the new event API. It's also relevant to #2130, but is about a simpler scheme where we wouldn't use any recent Metal features, see https://github.com/gfx-rs/gfx/issues/2130#issuecomment-457355182

kvark:
Fixes #2809
Passes the relevant Vulkan CTS test! Although, might still have logic flows due to complexity of the synchronization here.

Based on #2840 and #2857

PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -122,6 +122,50 @@ impl QueueInner {
 }
 
 #[derive(Debug)]
+pub struct BlockedSubmission {
+    wait_events: Vec<Arc<AtomicBool>>,
+    command_buffers: Vec<metal::CommandBuffer>,
+    affect_events: Vec<Arc<AtomicBool>>,
+}
+
+/// Class responsible for keeping the state of submissions between the
+/// requested user submission that is blocked by a host event, and
+/// setting the event itself on the host.
+#[derive(Debug, Default)]
+pub struct QueueBlocker {
+    submissions: Vec<BlockedSubmission>,
+}
+
+impl QueueBlocker {
+    fn submit_impl(&mut self, cmd_buffer: &metal::CommandBufferRef) {
+        match self.submissions.last_mut() {
+            Some(blocked) => blocked.command_buffers.push(cmd_buffer.to_owned()),
+            None => cmd_buffer.commit(),
+        }
+    }
+
+    pub(crate) fn set_host_event(&mut self, event: &Arc<AtomicBool>) {
+        // clean up the relevant blocks
+        for blocked in self.submissions.iter_mut() {
+            blocked.wait_events.retain(|ev| !Arc::ptr_eq(ev, event));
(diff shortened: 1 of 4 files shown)


kvark:
Rebased and ready for review (@grovesNL ?)

src/backend/metal/src/command.rs
```diff
@@ -4228,26 +4311,50 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
...
+                .map_or(false, |ev| ev.1);
+            if is_local {
+                need_barrier = true;
+            } else {
+                inner.host_events.push(Arc::clone(event));
+            }
+        }
+
+        if need_barrier {
+            self.pipeline_barrier(stages, memory::Dependencies::empty(), barriers);
```
grovesNL: I think events could allow the pipeline barrier operation to be delayed until the data is actually used, so it could be non-optimal to force the pipeline barrier at this point.

Although none of this really matters as long as we continue to no-op `pipeline_barrier`

src/backend/metal/src/command.rs
```diff
@@ -144,18 +143,19 @@ impl QueueBlocker {
...
-        for blocked in self.submissions.iter_mut() {
-            blocked.wait_events.retain(|ev| !Arc::ptr_eq(ev, event));
-            if blocked.affect_events.iter().any(|ev| Arc::ptr_eq(ev, event)) {
-                // Execution of this command buffer affects the event,
-                // we can't look further ahead.
-                break;
-            }
-        }
+        let done = {
+            let blocked = match self.submissions.first_mut() {
```
grovesNL: If I understand correctly, why do we only check the first submission here?

For example, is it possible for `self.submissions[0]` to be blocked but `self.submissions[1]` to be unblocked – in which case we don't try to triage `self.submissions[1]` despite it being ready?

kvark: We can't submit submissions[1] anyway for as long as submissions[0] is empty.
This method is not comprehensive, it's expected to be called "occasionally" and advance the state of things. So taking one submission seems fine.

grovesNL: Oh ok, so these are ordered too. Sounds good :+1:

src/backend/metal/src/native.rs
```diff
@@ -1007,6 +1007,10 @@ pub struct Fence(pub(crate) RefCell<FenceInner>);
 unsafe impl Send for Fence {}
 unsafe impl Sync for Fence {}
 
+//TODO: review the atomic ordering
```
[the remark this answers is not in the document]
kvark: on a second look, the atomics here are misplaced, I think. It should really be just an externally synchronized bool... Which one can express as a `SeqCst` atomic, but that doesn't appear to be the most straightforward representation.

kvark:
bors r=grovesNL

PR #2867, opened 2019-06-25
Some long-needed rustfmt changes

kvark:
> blank_lines_upper_bound = 2

This allows us to put more than a single blank line between important sections. I was really confused when rustfmt started erasing my empty lines separating `use` sections from the other items, for example. No more!

> spaces_around_ranges = true

Ranges are often used with non-trivial values on both sides, so spaces look much better.

> imports_layout = "HorizontalVertical"

A compromise...


kvark:
Bors r=grovesNL

PR #2869, opened 2019-06-26
Derive Debug for All Backend Instances

arashikou:
Fixing https://github.com/gfx-rs/wgpu/issues/76 is blocked on the `Instance` types for all the backends not implementing `Debug`. This patch adds `#[derive(Debug)]` to all the instances.

PR checklist:
- [x] tested examples with the following backends: dx11, dx12, vulkan

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -99,6 +99,7 @@ pub(crate) struct ViewInfo {
     range: image::SubresourceRange,
 }
 
+#[derive(Debug)]
 pub struct Instance {
     pub(crate) factory: ComPtr<IDXGIFactory>,
     pub(crate) dxgi_version: dxgi::DxgiVersion,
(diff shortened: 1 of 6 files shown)


arashikou:
My Linux box is down until tomorrow, so I will get to the two steps requiring a *nix box then.

Also, I tried to verify the examples with the gl backend, but I got a crash in colour-uniform. I can't fathom how it could be related to this change, but I will verify that it's unrelated.

Personal todos:
- [ ] Run make-based builds
- [ ] Verify that OpenGL crash is unrelated

kvark:
Thank you!
Note that wgpu would not be able to use it unless we either:
  - backport the change to hal-0.2 branch and release patches
  - switch wgpu-native to depend on gfx-rs from git. It will only need to be switched back to crates version when wgpu itself needs to publish

bors r+

arashikou:
Yeah. Or wait for gfx-hal to release 0.3. Any preference on which way to go? My leaning is to backport and then pick it up with the next 0.2.x release, just because I have no idea what updating wgpu to use the current tip of gfx master is going to require. I'm willing to do the legwork to backport it.

kvark:
I forgot that wgpu depends on Rendy as well, so we can't simply switch it to use gfx from git...

Issue #2866, opened 2019-06-24
Exclude certain resources from Metal argument buffers


kvark:
Apparently, Tier 1 support for argument buffers doesn't not allow writable textures to be placed there: https://developer.apple.com/documentation/metal/buffers/about_argument_buffers?language=objc

We need to bind them separately.

kvark:
This also applies to dynamic-offset buffers.

kvark:
marking as blocked on shader translation changes

PR #2870, opened 2019-06-26
[mtl] override argument buffer sets

kvark:
Makes argument buffers compatible with push constants again. Also reduce the binding of argument buffers to only the affected stages.
PR checklist:

I think this brings us to the end of the second push for argument buffers. They seem pretty solid and operational as of today, vkQuake is a good test case. There are serious limitations though that ask for big changes:
  1. Storage images and dynamic-offset buffers can't be a part of the argument buffer - #2866 . This requires shader translation changes.
  2. We need to know the dimensionality and msaa-ness of textures at the descriptor set layout creation time. This would mean another derivation from Vulkan proper.

Once we have these implemented, we'll be able to benchmark Dota and Dolphin again. But until then, argument buffer support will be kept as an experimental feature disabled by default.

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -1046,24 +1046,35 @@ impl hal::Device<Backend> for Device {
                         }
                     }
                 }
-                n::DescriptorSetLayout::ArgumentBuffer { ref bindings, .. } => {
-                    //Note: we specifically ignore the stage flags and unconditionally
-                    // add resource overrides and bump the counter, because there is no
-                    // way to override the argument buffer binding itself in SPIRV-Cross.
-                    //TODO: ^^
-                    for &mut (_stage_bit, stage, ref mut counters) in stage_infos.iter_mut() {
-                        res_overrides.extend(bindings
-                            .iter()
-                            .map(|(&binding, arg)| {
-                                let key = msl::ResourceBindingLocation {
-                                    stage,
-                                    desc_set: set_index as _,
-                                    binding,
-                                };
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/device.rs
```diff
@@ -1046,24 +1046,35 @@ impl hal::Device<Backend> for Device {
...
+                    for &mut (stage_bit, stage, ref mut counters) in stage_infos.iter_mut() {
+                        let has_stage = stage_flags.contains(stage_bit);
+                        res_overrides.insert(
+                            msl::ResourceBindingLocation {
+                                stage,
+                                desc_set: set_index as _,
+                                binding: msl::ARGUMENT_BUFFER_BINDING,
+                            },
+                            msl::ResourceBinding {
+                                buffer_id: if has_stage { counters.buffers } else { !0 },
```
grovesNL: Why is `counters.buffers` guaranteed to be correct here?

kvark: We are at the stage of going through resources and bumping the counters. Before, we were first going though resources and then through the push constants. With #2840, we were basically bumping the buffer counters 1:1 with enumerating descriptor sets, which matched spirv-cross default mapping of argument buffers. As of #2857, we are scanning the push constants first, which broke the matching.

This PR uses the new overrides, so that we can force the mapping again.

kvark:
bors r=grovesNL

kvark:
bors retry

kvark:
bors r=grovesNL

PR #2881, opened 2019-07-03
Rust formatting pass

kvark:


kvark:
@msiglreith thank you for understanding! There doesn't appear to be a way to make everyone happy, and I'm glad we aren't fighting about this :)

PR #2882, opened 2019-07-03
Presentation Surface API

### LINKED ISSUE #2863: Alternative swapchain model
## Situation
We have a lot of swapchain-related issues:
  - it doesn't map to Metal:
    - the current solution is hacky and occasionally fires back, with things like full-screen, or orientation change on iOS
  - it doesn't map to most GL WSI
    - GL doesn't let you us `VkImage` of the framebuffer, although may be possible in some configurations (i.e. Wayland, or `IOSurface`-backed WSI)
  - it doesn't map nice to DX12
    - users aren't allowed to do anything with any other image than currently acquired

it's really the most rough API piece of all the things we have... and the problem comes from how Vulkan exposes the swapchain. Good thing is - designers of Vulkan anticipated the WSI being the pain spot and excluded it from the core specification.

## Idea

What if we expose a very simple (and limited) API for the swapchain: you get the next frame as as a structure that can be borrowed as `Framebuffer`, you draw to it (only using one-time-submit command buffers), and you pass it back by value on `present()`. That's it. No enumeration of images ahead of time, no semaphores, less possible errors on the way and definitely less pain for the backends to implement it.

My bet is that this simple model would work for *most* of the client apps. I'd like to hear from our users if it wouldn't work. The only known use that doesn't it this scenario is gfx-portability... so my hope is to structure that simple API in such a way that the current complex logic of Metal backend (mapping Vulkan to Metal) is basically moved and adopted there, in gfx-portability. It could in fact even be simpler since at this level we have the leverage of providing more hidden state info from the backends.

Note: having render-only swapchain images is certainly more limited than regular Vulkan WSI, where you could ask for supported usages and potentially copy to/from the swapchain images, even bind them as storage, etc. However, there is evidence that presentation works most efficiently if it's *only* rendered to: Metal has a special flag "isFramebufferOnly" indicating possible optimizations, and DXGI is known to operate in a different mode if you need readbacks and such.

kvark:
Fixes #2863
PR checklist:

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -49,8 +49,13 @@ use hal::{
     window,
 };
 
-use std::io::Cursor;
-use std::mem::ManuallyDrop;
+use std::{
+    borrow::Borrow,
+    io::Cursor,
+    iter,
+    mem::{self, ManuallyDrop},
+    ptr,
+};
 
 #[cfg_attr(rustfmt, rustfmt_skip)]
 const DIMS: window::Extent2D = window::Extent2D { width: 1024, height: 768 };
@@ -99,11 +104,17 @@ fn main() {
     let event_loop = winit::event_loop::EventLoop::new();
 
     #[cfg(not(target_arch = "wasm32"))]
+    let dpi = event_loop.primary_monitor().hidpi_factor();
+
+    #[cfg(not(target_arch = "wasm32"))]
     let wb = winit::window::WindowBuilder::new()
         .with_min_inner_size(winit::dpi::LogicalSize::new(1.0, 1.0))
-        .with_inner_size(winit::dpi::LogicalSize::new(
-            DIMS.width as _,
-            DIMS.height as _,
+        .with_inner_size(winit::dpi::LogicalSize::from_physical(
+            winit::dpi::PhysicalSize::new(
+                DIMS.width as _,
+                DIMS.height as _,
+            ),
+            dpi,
         ))
         .with_title("quad".to_string());
     // instantiate backend
@@ -177,14 +188,13 @@ fn main() {
                     println!("resized to {:?}", dims);
                     #[cfg(feature = "gl")]
                     {
-                        let context = renderer.surface.get_context();
+                        let context = renderer.surface.context();
                         context.resize(dims.to_physical(window.hidpi_factor()));
                     }
-                    let dimensions = window::Extent2D {
-                        width: dims.width as u32,
-                        height: dims.height as u32,
+                    renderer.dimensions = window::Extent2D {
+                        width: (dims.width * dpi) as u32,
+                        height: (dims.height * dpi) as u32,
                     };
-                    renderer.dimensions = dimensions;
                     renderer.recreate_swapchain();
                 }
                 _ => {}
@@ -204,10 +214,7 @@ struct Renderer<B: hal::Backend> {
     surface: B::Surface,
     adapter: hal::adapter::Adapter<B>,
     format: hal::format::Format,
-    swap_chain: Option<B::Swapchain>,
     dimensions: window::Extent2D,
-    framebuffers: Vec<B::Framebuffer>,
-    frame_images: Vec<(B::Image, B::ImageView)>,
     viewport: hal::pso::Viewport,
     render_pass: ManuallyDrop<B::RenderPass>,
     pipeline: ManuallyDrop<B::GraphicsPipeline>,
@@ -215,8 +222,6 @@ struct Renderer<B: hal::Backend> {
     desc_set: B::DescriptorSet,
     set_layout: ManuallyDrop<B::DescriptorSetLayout>,
     submission_complete_semaphores: Vec<B::Semaphore>,
-    image_acquire_semaphores: Vec<B::Semaphore>,
-    free_acquire_semaphore: Option<B::Semaphore>,
     submission_complete_fences: Vec<B::Fence>,
     cmd_pools: Vec<B::CommandPool>,
     cmd_buffers: Vec<B::CommandBuffer>,
@@ -313,7 +318,7 @@ where
         // Buffer allocations
         println!("Memory types: {:?}", memory_types);
 
-        let buffer_stride = std::mem::size_of::<Vertex>() as u64;
+        let buffer_stride = mem::size_of::<Vertex>() as u64;
         let buffer_len = QUAD.len() as u64 * buffer_stride;
 
         assert_ne!(buffer_len, 0);
@@ -444,7 +449,7 @@ where
                     set: &desc_set,
                     binding: 0,
                     array_offset: 0,
-                    descriptors: Some(pso::Descriptor::Image(&*image_srv, i::Layout::Undefined)),
+                    descriptors: Some(pso::Descriptor::Image(&*image_srv, i::Layout::ShaderReadOnlyOptimal)),
                 },
                 pso::DescriptorSetWrite {
                     set: &desc_set,
@@ -536,12 +541,11 @@ where
 
         let swap_config = window::SwapchainConfig::from_caps(&caps, format, DIMS);
         println!("{:?}", swap_config);
-        let extent = swap_config.extent.to_extent();
-
-        let (swap_chain, backbuffer) =
-            unsafe { device.create_swapchain(&mut surface, swap_config, None) }
-                .expect("Can't create swapchain");
-        let swap_chain = Some(swap_chain);
+        let extent = swap_config.extent;
+        unsafe {
+            surface.configure_swapchain(&device, swap_config)
+                .expect("Can't configure swapchain");
+        };
 
         let render_pass = {
             let attachment = pass::Attachment {
@@ -577,46 +581,9 @@ where
             )
         };
 
-        let (frame_images, framebuffers) = {
-            let pairs = backbuffer
-                .into_iter()
-                .map(|image| unsafe {
-                    let rtv = device
-                        .create_image_view(
-                            &image,
-                            i::ViewKind::D2,
-                            format,
-                            Swizzle::NO,
-                            COLOR_RANGE.clone(),
-                        )
-                        .unwrap();
-                    (image, rtv)
-                })
-                .collect::<Vec<_>>();
-            let fbos = pairs
-                .iter()
-                .map(|&(_, ref rtv)| unsafe {
-                    device
-                        .create_framebuffer(&render_pass, Some(rtv), extent)
-                        .unwrap()
-                })
-                .collect::<Vec<_>>();
-            (pairs, fbos)
-        };
-
         // Define maximum number of frames we want to be able to be "in flight" (being computed
         // simultaneously) at once
         let frames_in_flight = 3;
-
-        // Number of image acquisition semaphores is based on the number of swapchain images, not frames in flight,
-        // plus one extra which we can guarantee is unused at any given time by swapping it out with the ones
-        // in the rest of the queue.
-        let mut image_acquire_semaphores = Vec::with_capacity(frame_images.len());
-        let free_acquire_semaphore = Option::Some(
-            device
-                .create_semaphore()
-                .expect("Could not create semaphore"),
-        );
 
         // The number of the rest of the resources is based on the frames in flight.
         let mut submission_complete_semaphores = Vec::with_capacity(frames_in_flight);
@@ -647,14 +614,6 @@ where
             }
         }
 
-        for _ in 0 .. frame_images.len() {
-            image_acquire_semaphores.push(
-                device
-                    .create_semaphore()
-                    .expect("Could not create semaphore"),
-            );
-        }
-
         for i in 0 .. frames_in_flight {
             submission_complete_semaphores.push(
                 device
@@ -672,7 +631,7 @@ where
         let pipeline_layout = ManuallyDrop::new(
             unsafe {
                 device.create_pipeline_layout(
-                    std::iter::once(&*set_layout),
+                    iter::once(&*set_layout),
                     &[(pso::ShaderStageFlags::VERTEX, 0 .. 8)],
                 )
             }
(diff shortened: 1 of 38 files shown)


kvark:
This works to the point of rendering the modified `quad` example on macOS. The API is not quite good yet, we need some more iteration on it. Requesting feedback!

src/backend/metal/src/window.rs
```diff
@@ -355,13 +375,97 @@ impl hal::Surface<Backend> for Surface {
...
+        let display_sync = config.present_mode != PresentMode::Immediate;
+        let is_mac = caps.os_is_mac;
+        let can_set_next_drawable_timeout = if is_mac {
+            caps.has_version_at_least(10, 13)
+        } else {
+            caps.has_version_at_least(11, 0)
+        };
+        let can_set_display_sync = is_mac && caps.has_version_at_least(10, 13);
+        let drawable_size = {
+            let bounds: CGRect = msg_send![render_layer, bounds];
```
mtak-: Can this be changed to `self.inner.dimensions()` to take scaling into account?

mtak-:
The trickiest part of testing this out on our codebase was having to (maybe) create the framebuffers after `acquire_image` which is somewhat the point of this. I wound up just using a `HashMap` like the quad example. I don't have much feedback on the API except that it was easy to port over to it. The metal implementation does fix our resizing problems on iOS/macos - everything is butter smooth.

src/backend/metal/src/window.rs
kvark: I tried that and got incorrect scaling on the quad example, didn't investigate further

aleksijuvani:
I'm still seeing `IOSurface` leaks with this implementation from `nextDrawable` when the window is in fullscreen mode. Are we freeing the drawables correctly?

src/backend/metal/src/window.rs
```diff
@@ -355,13 +375,97 @@ impl hal::Surface<Backend> for Surface {
...
+        // for the drawable & texture
+        let render_layer_borrow = self.inner.render_layer.lock();
+        let (drawable, raw) = autoreleasepool(|| {
+            let drawable: &metal::DrawableRef = msg_send![*render_layer_borrow, nextDrawable];
+            assert!(!drawable.as_ptr().is_null());
+            let texture: &metal::TextureRef = msg_send![drawable, texture];
+            (drawable.to_owned(), texture.to_owned())
+        });
+
+        let id = raw.as_ptr() as usize as hal::SwapchainImageId; //HACK
```
aleksijuvani: Would we be able to use [`drawableID`](https://developer.apple.com/documentation/metal/mtldrawable/2806860-drawableid?language=objc) here?

src/backend/metal/src/window.rs
aleksijuvani: This doesn't look right to me. `bounds` returns the drawable size in logical pixels. `drawableSize` ought to be used to get the drawable size in physical pixels. Even if we change this, it appears that this is trying to set the drawable size to the same value that it already is. I'm not sure what was intended here.

Shouldn't we be storing the new extents of the swapchain image in `SurfaceSwapchainConfig` and be setting the drawable size to that instead? Currently, there doesn't seem to be any way to specify the extents of the swapchain.

src/backend/metal/src/window.rs
kvark: According to the docs, this would just be monotonically increasing. Is there use for this ID then? The original idea behind the ID was that the users can keep associated framebuffers... I'd really like to get rid of this ID semantics though, it's not very sound. The main blocker is a potential use of the swapchain image in a framebuffer that has other images :/

examples/quad/main.rs
```diff
@@ -677,78 +636,39 @@ fn main() {
...
-            };
-
-            framebuffers = new_framebuffers;
-            frame_images = new_frame_images;
-            viewport.rect.w = extent.width as _;
-            viewport.rect.h = extent.height as _;
+            viewport.rect.w = resize_dims.width as _;
+            viewport.rect.h = resize_dims.height as _;
             recreate_swapchain = false;
+            framebuffers.clear();
```
aleksijuvani: We ought to call `destroy_framebuffer` here on the framebuffers before dropping them.

src/backend/metal/src/window.rs
aleksijuvani: I interpreted the documentation as the ID increasing every time a new drawable is created, but if an existing drawable is recycled, no new drawable would be created, and thus the ID would be the same. I notice that it doesn't explicitly state this, so I would have to test what it returns. Regardless, I realize now that `drawableID` is only available on iOS 10.3+ and tvOS 10.3+. It is not available on macOS.

src/backend/metal/src/window.rs
mtak-: Quad is using logical pixels for everything. Which is why setting the drawable size to `bounds` (logical size) works for quad.

src/backend/metal/src/window.rs
mtak-: It seems borderline impossible to implement unique id without running into ABA. The metal backend would still have to track MTLTextures somehow.

kvark:
Thanks @aleksijuvani and @mtak- for review and early testing of this!

The next step in the API, as discussed here and on gitter, would be to remove the IDs completely and instead ask the users to create the framebuffers every frame and delete when appropriate. Metal and DX12 backends would make sure to release the swapchain image-associated resources on present().

kvark:
Here is the second iteration of the API. We aren't storing `MTLTexture` objects for the views of the surface swapchain any more, so I don't expect any leaks. Please take a look!

aleksijuvani:
Looks good! I'm not seeing any more leaks. The swapchain image extents will still need to be addressed. The current implementation causes incorrect scaling in my application.

mtak-:
Scaling is also incorrect in my application. Is creating a `Framebuffer` during a render OK on vulkan?

I am looking forward to this merge!

kvark:
@mtak- 
> Is creating a Framebuffer during a render OK on vulkan?

That is still a big unknown in the air. We have somewhat compelling evidence that it's fine: neither DX11/12 or Metal have concepts for framebuffers, so these must be purely logical in the driver and lightweight.

I'll fix the scaling and try to eliminate `kind()` method as suggested by @omni-viral . Also will move out the things into a separate trait.

kvark:
Alright, I addressed all the concerns in these comments now. Let's all have one more look (3rd iteration?) before we proceed with the other backends

mtak-:
Scaling issues are fixed. Looks good to me!

aleksijuvani:
Looks good to me. Scaling seems correct in my application as well. Is there any reason we're no longer allowing the user to configure the size of the swapchain images? I see that we're now always setting it to the bounds of the `NSView`.

kvark:
@aleksijuvani this is now also addressed. Looking into Vulkan and friends...

src/backend/gl/src/queue.rs
```diff
@@ -1077,6 +1077,18 @@ impl hal::queue::RawCommandQueue<Backend> for CommandQueue {
...
     }
 
+    #[cfg(any(feature = "glutin", feature = "wgl", target_arch = "wasm32"))]
+    unsafe fn present_surface(
+        &mut self,
+        surface: &mut crate::Surface,
+        _image: crate::SurfaceImage,
+    ) -> Result<Option<hal::window::Suboptimal>, hal::window::PresentError> {
+        #[cfg(all(feature = "glutin", not(target_arch = "wasm32")))]
+        surface.context.swap_buffers().unwrap();
```
msiglreith: I don't think this work that easily. How would the backend know when it should render into the default framebuffer?

src/backend/vulkan/src/window.rs
```diff
@@ -435,6 +469,125 @@ impl hal::Surface<Backend> for Surface {
...
+                Some(ssc.swapchain)
+            }
+            None => None,
+        };
+
+        let (swapchain, images) = device.create_swapchain(self, config, old)?;
+        self.swapchain = Some(SurfaceSwapchain {
+            swapchain,
+            device: Arc::clone(&device.raw),
+            fence: device.create_fence(false).unwrap(),
```
msiglreith: need to destruct this one somewhere

src/backend/gl/src/queue.rs
kvark: you are right, GL backend needs more stuff...

src/backend/vulkan/src/window.rs
```diff
@@ -435,6 +469,125 @@ impl hal::Surface<Backend> for Surface {
...
+        let old = match self.swapchain.take() {
+            Some(ssc) => {
+                ssc.device.0.destroy_fence(ssc.fence.0, None);
+                self.stale_views.push((ssc.device, ssc.semaphore, ssc.views));
+                Some(ssc.swapchain)
+            }
+            None => None,
+        };
+
+        let (swapchain, images) = device.create_swapchain(self, config, old)?;
```
aleksijuvani: There's a validation error on program exit about the swapchain not being destroyed prior to the surface being destroyed. The swapchain ought to be destroyed when the surface is destroyed.

src/backend/vulkan/src/window.rs
```diff
@@ -164,15 +165,13 @@ impl Instance {
             unsafe { w_loader.create_wayland_surface(&info, None) }.expect("WaylandSurface failed")
         };
 
-        self.create_surface_from_vk_surface_khr(surface, width, height, 1)
+        self.create_surface_from_vk_surface_khr(surface)
     }
 
     #[cfg(target_os = "android")]
     pub fn create_surface_android(
```
mtak-: The calls to `create_surface_from_wayland` and `create_surface_android` need to be updated to remove the width/height parameters.

kvark:
As I was working through DX11 implementation (see last commit), I realized that DX11/12, Metal, and Vulkan all want different things:
  - Metal wants all associated framebuffers to be removed around `present()` time
  - DX11/12 wants all the associated framebuffers to be removed before `configure_swapchain()` time. This is a limitation of `ResizeBuffers` call that requires no references to the swapchain buffers.
  - Vulkan wants the associated framebuffers to live for as long as GPU uses them...

If the user is responsible for creation/deletion of these framebuffers, the Metal and DX backends would have a hard time erasing the internal mutable state of these objects while the user is still owning them. This is possible but really clunky and not pretty.

Given that and the fact that in Metal and DX framebuffers are "logical" objects that don't have a low-level counterpart (and thus easy/fast to create), it seems like the best way to proceed would be to make the Vulkan backend cache and track the swapchain image views and associated framebuffers. From the user perspective, they'd have to remove all the associated resource before the next call to either acquire() or configure_swapchain().
@omni-viral @msiglreith does that sound feasible to you?

kvark:
Went through the pains of rebasing, and also implemented the core Vulkan logic of handling the API. It doesn't work yet...

kvark:
Current status:
  - Metal works
  - GL Glutin+Dummy work, others not ported yet
  - Vulkan is ported, doesn't work correctly yet
  - D3D11 works (took me way more time to nail down than I'd like...)
  - D3D12 isn't ported yet

@aleksijuvani want to have another look at the API in the meantime?

omni-viral:
@kvark do I understand correctly that user would have to destroy image views to swapchain images (and any object that transitively references it) before acquiring next presentable image?

kvark:
@omni-viral you can't explicitly destroy image views because you don't own them. The acquired image has to be moved back by calling present(). But you do have to destroy all objects that reference that view:
  - framebuffers
  - command buffers using it must be reset, or they are one-time-submit and actually submitted

kvark:
Status update: got DX12 working, although I had to insert a `wait_idle` in there because DX12 doesn't like resources of a swapchain being released while they are in flight used by GPU.

src/backend/metal/src/window.rs
```diff
@@ -77,10 +80,65 @@ impl SurfaceInner {
...
+
+        let device_raw = shared.device.lock().as_ptr();
+        unsafe {
+            // On iOS, unless the user supplies a view with a CAMetalLayer, we
+            // create one as a sublayer. However, when the view changes size,
+            // its sublayers are not automatically resized, and we must resize
+            // it here. The drawable size and the layer size don't correlate
+            #[cfg(target_os = "ios")]
+            {
+                if let Some(view) = surface.inner.view {
```
aleksijuvani:
```suggestion
                if let Some(view) = self.view {
```

aleksijuvani:
I've tested the latest changes with my application, and both DirectX 12 and Metal seem to be working. Vulkan crashes after presenting a single frame and complains about deleting an invalid framebuffer, but I think that's to be expected for now?

kvark:
Thank you for testing!
Yes, I'm still figuring out what to do with Vulkan. This PR ended up being a huge time/effort investment that I didn't anticipate.

kvark:
Alright, Vulkan works now. I'm nailing down the last remaining pieces.
@msiglreith please have another look

src/backend/dx12/src/window.rs
```diff
@@ -70,13 +78,27 @@ impl w::Surface<Backend> for Surface {
...
         let capabilities = w::SurfaceCapabilities {
             image_count: 2 ..= 16, // we currently use a flip effect which supports 2..=16 buffers
-            current_extent: Some(extent),
-            extents: extent ..= extent,
+            current_extent,
+            extents: w::Extent2D {
+                width: 16,
+                height: 16,
+            } ..= w::Extent2D {
+                width: 4096,
```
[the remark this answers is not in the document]
kvark: pretty much. Is this wrong? does this have to match the current extent only?

kvark:
@grovesNL there is a bit of repetition in the GL window modules now, but I'm not capable of cleaning it up nicely, since I don't want to dive into the zoo of GL WSI semantics in this PR...

kvark:
Bors retry

kvark:
Bors retry

kvark:
bors r=msiglreith

Issue #2883, opened 2019-07-04
Support for unnormalized coordinates in samplers


kvark:
Vulkan [sampler descriptor](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkSamplerCreateInfo.html) has this field: `unnormalizedCoordinates`.

The task here is to add it to our API and do a basic investigation on where it can't be supported.

kvark:
Now that #2885 has landed, we have the API in place. We need to find a way to support it on DX12, at least, and also look into DX11 and GL if possible.

kvark:
I'm thinking, the possible solution could be:
  1. provide a root constant (32 bit) with each bit corresponding to a sampler. Bit = 1 if the coordinates are unnormalized.
  2. in the generated HLSL, whenever there is a sampling operation, we can insert code that checks the relevant bit. If the bit is set, the injected code would call `GetDimensions` on the texture, and use it to produce the normalized coordinates prior to sampling.

PR #2885, opened 2019-07-04
Hardcoded immutable samplers in Metal

### LINKED ISSUE #2837: Hard-code immutable samplers into MSL
Some iOS devices don't support comparison samplers being bound dynamically at all. Instead one can hard-code them in the generated MSL, which SPIRV-Cross can already do, see https://github.com/KhronosGroup/MoltenVK/pull/644

kvark:
Also adds unnormalized sampling to HAL.
Blocked on spirv_cross 0.15.0 release.

Fixes #2837
Partially fulfills #2883 
PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -169,6 +169,7 @@ impl VisibilityShared {
 pub struct Device {
     pub(crate) shared: Arc<Shared>,
     memory_types: Vec<hal::MemoryType>,
+    features: hal::Features,
     pub online_recording: OnlineRecording,
 }
 unsafe impl Send for Device {}
@@ -313,6 +314,7 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
         let device = Device {
             shared: self.shared.clone(),
             memory_types: self.memory_types.clone(),
+            features: requested_features,
             online_recording: OnlineRecording::default(),
         };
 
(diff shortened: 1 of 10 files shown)


kvark:
fwiw, I confirmed on VK CTS that the samplers get hard-coded and work as expected.

src/backend/metal/src/device.rs
```diff
@@ -795,18 +809,67 @@ impl Device {
...
+            min_filter: match info.min_filter {
+                image::Filter::Nearest => msl::SamplerFilter::Nearest,
+                image::Filter::Linear => msl::SamplerFilter::Linear,
+            },
+            mag_filter: match info.mag_filter {
+                image::Filter::Nearest => msl::SamplerFilter::Nearest,
+                image::Filter::Linear => msl::SamplerFilter::Linear,
+            },
+            mip_filter: match info.min_filter {
+                image::Filter::Nearest if info.lod_range.end < image::Lod::from(0.5) =>
```
grovesNL: Why do we need the special LOD logic here?

src/backend/vulkan/src/device.rs
```diff
@@ -1096,7 +1096,7 @@ impl d::Device<B> for Device {
                     vk::BorderColor::FLOAT_TRANSPARENT_BLACK
                 }
             },
-            unnormalized_coordinates: vk::FALSE,
+            unnormalized_coordinates: if sampler_info.normalized { vk::TRUE } else { vk::FALSE },
```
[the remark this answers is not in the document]
kvark: well spotted!

src/backend/metal/src/device.rs
kvark: same logic as we do in `make_sampler_descriptor`

kvark:
Thanks for the review @grovesNL !
comments are addressed
bors r=grovesNL

kvark:
bors retry

PR #2887, opened 2019-07-05
Reset `vertex_attrib_divisor` to `0` when needed on GL backend

hecrj:
This fixes `vertex_attrib_divisor` not being reset to `0` properly after using instanced drawing.

PR checklist:

### DIFF
--- a/src/backend/gl/src/queue.rs
+++ b/src/backend/gl/src/queue.rs
@@ -576,16 +576,14 @@ impl CommandQueue {
                     }
                 }
 
-                if rate != 0 {
-                    if self
-                        .share
-                        .legacy_features
-                        .contains(LegacyFeatures::INSTANCED_ATTRIBUTE_BINDING)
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -576,16 +576,14 @@ impl CommandQueue {
...
-                    } else {
-                        error!("Binding attribute with instanced input rate is not supported");
-                    }
+                if self
+                    .share
+                    .legacy_features
+                    .contains(LegacyFeatures::INSTANCED_ATTRIBUTE_BINDING)
+                {
+                    gl.vertex_attrib_divisor(location, rate);
+                } else {
```
msiglreith: This branch should either be removed or requires an additional check for rates > 0

hecrj: Good catch, it should be fixed now.

Issue #2874, opened 2019-06-27
Upgrade to winit 0.20.0


rukai:
I opened this earlier https://github.com/gfx-rs/wgpu/issues/238
But it quickly became clear that gfx-hal is the first step here.

I've started working on this.

kvark:
This has been requested on the forum: https://users.rust-lang.org/t/winit-0-20-the-state-of-windowing-in-rust-and-a-request-for-help/29485/31?u=kvark
cc @norru

PR #2888, opened 2019-07-06
Upgrade to winit 0.20.0

rukai:
Fixes https://github.com/gfx-rs/gfx/issues/2874
PR checklist:
- [x] tested examples with the following:
*   Backends: vulkan and opengl
*   OS: Arch Linux
*   GPU: GTX 960
*   Driver: Nvidia proprietary driver
- [] `rustfmt` run on changed code - its panicking and changing things outside of my commit :/ I'll look into this later

Concerns:
*   I had to comment out some opengl specific resize handling code, I dont seem to be able to call the needed functions on a `B::Surface` and I dont understand the rules behind how the `back::surface` becomes a `B::Surface` But it seems to still handle resize without this code?
*   The other backends I haven't tested can be assumed to not work.
*    I have copied the code from the unstabilized ManuallyDrop::take implementation in order to cleanup the resources.
*   For now the colour-uniform example is just commented out. The plan is to implement it once we are happy with the state of winit 0.20 and want to progress further. I don't want to implement it yet as this PR could sit around for a while and I don't want to deal with merge conflicts.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -35,7 +35,7 @@ gfx-hal = { path = "../src/hal", version = "0.3" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.3" }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-winit = "0.19"
+winit = "0.20.0-alpha2"
 glsl-to-spirv = "0.1.4"
 
 [dependencies.gfx-backend-gl]
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -140,7 +140,6 @@ struct RendererState<B: Backend> {
     swapchain: Option<SwapchainState<B>>,
     device: Rc<RefCell<DeviceState<B>>>,
     backend: BackendState<B>,
-    window: WindowState,
     vertex_buffer: BufferState<B>,
     render_pass: RenderPassState<B>,
     uniform: Uniform<B>,
@@ -148,6 +147,11 @@ struct RendererState<B: Backend> {
     framebuffer: FramebufferState<B>,
     viewport: pso::Viewport,
     image: ImageState<B>,
+    recreate_swapchain: bool,
+    color: pso::ColorValue,
+    bg_color: pso::ColorValue,
+    cur_color: Color,
+    cur_value: u32,
 }
 
 #[derive(Debug)]
@@ -159,7 +163,7 @@ enum Color {
 }
 
 impl<B: Backend> RendererState<B> {
-    unsafe fn new(mut backend: BackendState<B>, window: WindowState) -> Self {
+    unsafe fn new(mut backend: BackendState<B>) -> Self {
         let device = Rc::new(RefCell::new(DeviceState::new(
             backend.adapter.adapter.take().unwrap(),
             &backend.surface,
@@ -297,7 +301,6 @@ impl<B: Backend> RendererState<B> {
         let viewport = RendererState::create_viewport(swapchain.as_ref().unwrap());
 
         RendererState {
-            window,
             backend,
             device,
             image,
@@ -310,6 +313,11 @@ impl<B: Backend> RendererState<B> {
             swapchain,
             framebuffer,
             viewport,
+            recreate_swapchain: false,
+            color: [1.0, 1.0, 1.0, 1.0],
+            bg_color: [0.8, 0.8, 0.8, 1.0],
+            cur_color: Color::Red,
+            cur_value: 0,
         }
     }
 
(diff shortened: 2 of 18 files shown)


examples/quad/main.rs
```diff
@@ -137,663 +138,751 @@ fn main() {
...
-        device.create_command_pool_typed(&queue_group, pool::CommandPoolCreateFlags::empty())
-    }
-    .expect("Can't create command pool");
+    // It is important that the closure move captures the Renderer,
+    // otherwise it will not be dropped when the event loop exits.
+    event_loop.run(move |event, _, control_flow| {
+        *control_flow = winit::event_loop::ControlFlow::Wait;
 
-    // Setup renderpass and pipeline
-    let set_layout = unsafe {
```
kvark: why are we even touching all this code? Creating the resources doesn't depend on winit, only the swapchain is.

examples/quad/main.rs
```diff
@@ -137,663 +138,751 @@ fn main() {
...
-        device.create_image_view(
-            &image_logo,
-            i::ViewKind::D2,
-            ColorFormat::SELF,
-            Swizzle::NO,
-            COLOR_RANGE.clone(),
-        )
-    }
-    .unwrap();
+struct Renderer<B: hal::Backend> {
```
kvark: do we have to make it structured here? I think having a very simple example that just does all the things in one function is a good point to start with, and it's less code than trying to structurize everything. We have colour-uniform to cover that use case

rukai: Because winit now calls us I think storing resources in a struct is now worth it:
*   Much easier to understand as the winit events loop code is cleanly separated from the renderer code
*   Putting the cleanup code in the Renderer::drop is less scary, in terms of never cleaning up or cleaning up more than once.
*   All resources that are not used in the event loop are never dropped. When resource creation is done in a `Renderer::new`, then all resources created during the function either go into the Renderer struct or are dropped at the end of `Renderer::new`. Without that external function there are some gfx structs that are not used directly in the render loop so they do not get dropped, I dont know if this is an issue or not.

I have however created a separate branch here without the Renderer struct: https://github.com/rukai/gfx/commit/7247ed8e46cbbbe57a373f02b80f129ff1d69276
Let me know if you would still prefer me to replace this branch with that one.

kvark: Thank you! Would it be possible to write this without `ManuallyDrop`?
If the structured renderer is the way to go for all of the examples, and there is enough winit-related boilerplate, we may move it into "main_loop.rs" or something, shared by all the window-related examples.

rukai: Manually drop is required because the move closure only gives us an &mut T

kvark:
Next step is to update `colour-uniform`?

aclysma:
+1, this worked for me on MacOS with a very minor tweak (https://github.com/aclysma/gfx/commit/54262a6249a0ac2313d0e305c089e62ffdea7db0)

rukai:
Thanks! I'll include this in the pr!

rukai:
Pushed some changes:
*   Updates winit dep from `0.20.0-alpha1` to `0.20.0-alpha2`
*   WIP port of `colour-uniform` example
*   Mac OS fixes

rukai:
The colour-uniform example is working now, the changes were far less invasive then I had initially expected.

I would advise waiting until I cleanup + fix window before reviewing again.

rukai:
I'm happy with this, its ready for further review.

I have tested:
Windows: dx11, dx12, gl, vulkan
Linux: vulkan, gl, wasm

ios build fails because Glutin doesn't yet support ios in 0.20.0: https://github.com/rust-windowing/glutin/issues/1170

I tried running `rustfmt` again and its still panicking, I tried both nightly and stable. :/ Any magic invocation I'm supposed to give it?

kvark:
Thank you @rukai , amazing work!
Rustfmt panics for everyone, but we haven't investigated (narrowed down) the cause yet.
I'll try to review it tomorrow (please don't hesitate to ping!).

examples/colour-uniform/main.rs
```diff
@@ -140,14 +140,23 @@ struct RendererState<B: Backend> {
...
-    window: WindowState,
     vertex_buffer: BufferState<B>,
     render_pass: RenderPassState<B>,
     uniform: Uniform<B>,
     pipeline: PipelineState<B>,
     framebuffer: FramebufferState<B>,
     viewport: pso::Viewport,
     image: ImageState<B>,
+    recreate_swapchain: bool,
+    r: f32,
```
kvark: can we use [ColorValue](https://docs.rs/gfx-hal/0.2.1/gfx_hal/pso/type.ColorValue.html) for these?

rukai:
ios build fails because Glutin doesn't yet support ios in 0.20.0: https://github.com/rust-windowing/glutin/issues/1170

examples/colour-uniform/main.rs
```diff
@@ -1712,11 +1625,74 @@ impl<B: Backend> Drop for FramebufferState<B> {
...
+                    } => {
+                        if let Some(virtual_keycode) = virtual_keycode {
+                            renderer_state.input(virtual_keycode);
+                        }
+                    }
+                    _ => (),
+                }
+            }
+            winit::event::Event::EventsCleared => {
+                renderer_state.draw();
```
ZeGentzy: You should be calling winit's `request_redraw` here.

`renderer_state.draw();` should be moved to `RedrawRequested`, so that it also redraws when the OS requests a redraw.

examples/quad/main.rs
```diff
@@ -137,663 +138,749 @@ fn main() {
...
+                            height: dims.height as u32,
+                        };
+                        renderer.dimensions = dimensions;
+                        renderer.recreate_swapchain();
+                    }
+                    _ => { }
+                }
+            }
+            winit::event::Event::EventsCleared => {
+                renderer.render();
```
ZeGentzy: You should be calling winit's `request_redraw` here.

`renderer_state.draw();` should be moved to `RedrawRequested`, so that it also redraws when the OS requests a redraw.

examples/colour-uniform/main.rs
rukai: Ah interesting, I missed RedrawRequested.
Why do we call request_redraw, shouldn't winit know when I need to redraw anyway?

ZeGentzy: Well, winit will only issue a `RedrawRequested` when the OS tells it so. So, only if there has been a window resize, or parts of the window have been (unexposed then) exposed. However, simply receiving new input does not mean anything should redraw, which is why you should manually call `request_redraw`.

rukai: So I'm thinking the quad example should not call `request_redraw` as it is a static image.
However the colour-uniform should call it at EventsCleared as it responds to user input.

ZeGentzy: Yeah, that is about right.

examples/quad/main.rs
rukai: https://github.com/rust-windowing/winit/issues/1051 This issue is clashing really badly with RedrawRequested

rukai:
Resolved merge conflicts

examples/Cargo.toml
```diff
@@ -35,7 +35,7 @@ gfx-hal = { path = "../src/hal", version = "0.2" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.2" }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-winit = "0.19"
+winit = "0.20.0-alpha2"
```
kvark: do I understand correctly that we are waiting for "0.20.0" to come out before merging this?

rukai: Yes.

Alternatively it might be nice to get this merged after the release of gfx Hal 0.3

kvark:
Is there a timeline for 0.20?

rukai:
Not that I know of, I suspect it is at least a few months away.

ZeGentzy:
The timeline is whenever all the blocking issues are resolved, which might take anywhere from a couple days to a couple centuries :p

https://github.com/rust-windowing/winit/labels/Blocking%20a%20Release

kvark:
Thanks for the info! In this case, it sounds like we should release hal-0.3 now and then pick up `winit` update at some later point.

rukai:
Thoughts on merging this now?
Positives: I dont have to maintain this branch anymore + winit 0.20 will receive more testing.
Negatives: Wont be able to release gfx hal 0.4 until winit 0.20 is released.

kvark:
We have ios users who use the master branch... is winit working on ios?

kvark:
@mtak- @aleksijuvani are you using `winit` on iOS? We'll hold the PR until this is available, in this case, or proceed otherwise.

aleksijuvani:
Yes, and it works!

mtak-:
Yeah. I've been using winit `0.20.0-alpha2` for a while now. It works!

kvark:
Alright, thanks for amazing work and feedback!
bors r+

PR #2889, opened 2019-07-06
Various GL fixes

omni-viral:
Fixes number of small issues I encountered

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -842,17 +842,17 @@ impl d::Device<B> for Device {
                 let glow::ActiveUniform { size, utype, name } =
                     gl.get_active_uniform(program, uniform).unwrap();
 
-                let location = gl.get_uniform_location(program, &name).unwrap();
+                if let Some(location) = gl.get_uniform_location(program, &name) {
+                    // Sampler2D won't show up in UniformLocation and the only other uniforms
+                    // should be push constants
+                    uniforms.push(n::UniformDesc {
+                        location: location as _,
+                        offset,
+                        utype,
+                    });
 
-                // Sampler2D won't show up in UniformLocation and the only other uniforms
-                // should be push constants
-                uniforms.push(n::UniformDesc {
-                    location: location as _,
-                    offset,
-                    utype,
-                });
-
-                offset += size as u32;
+                    offset += size as u32;
+                }
             }
         }
 
@@ -986,6 +986,10 @@ impl d::Device<B> for Device {
                     color_attachment_index += 1;
                 }
                 Some(Format::Rgba8Srgb) => {
+                    render_attachments.push(color_attachment);
+                    color_attachment_index += 1;
+                }
+                Some(Format::Bgra8Srgb) => {
                     render_attachments.push(color_attachment);
                     color_attachment_index += 1;
                 }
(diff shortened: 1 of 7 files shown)


src/backend/gl/src/info.rs
```diff
@@ -140,6 +140,15 @@ fn get_usize(gl: &GlContainer, name: u32) -> Result<usize, Error> {
         Ok(value as usize)
     }
 }
+fn get_u64(gl: &GlContainer, name: u32) -> Result<u64, Error> {
```
kvark: why is this needed if we have `get_usize` implemented the same way?

src/backend/gl/src/info.rs
```diff
@@ -347,9 +356,9 @@ pub(crate) fn query_all(gl: &GlContainer) -> (Info, Features, LegacyFeatures, Li
         max_viewports: 1,
         optimal_buffer_copy_offset_alignment: 1,
         optimal_buffer_copy_pitch_alignment: 1,
-        min_texel_buffer_offset_alignment: 1,   // TODO
-        min_uniform_buffer_offset_alignment: 1, // TODO
-        min_storage_buffer_offset_alignment: 1, // TODO
+        min_texel_buffer_offset_alignment: 1,
+        min_uniform_buffer_offset_alignment: get_u64(gl, glow::UNIFORM_BUFFER_OFFSET_ALIGNMENT).unwrap_or(1024),
```
kvark: great!

src/backend/gl/src/lib.rs
```diff
@@ -655,7 +655,17 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
         _usage: image::Usage,
         _view_caps: image::ViewCapabilities,
     ) -> Option<image::FormatProperties> {
-        None //TODO
+        Some(image::FormatProperties {
+            max_extent: image::Extent {
+                width: !0,
```
kvark: let's at least leave a TODO to return the actual limits

src/backend/gl/src/window/web.rs
```diff
@@ -153,7 +153,9 @@ impl Device {
 
         let (int_format, iformat, itype) = match config.format {
             f::Format::Rgba8Unorm => (glow::RGBA8, glow::RGBA, glow::UNSIGNED_BYTE),
+            f::Format::Bgra8Unorm => (glow::RGBA8, glow::BGRA, glow::UNSIGNED_BYTE),
             f::Format::Rgba8Srgb => (glow::SRGB8_ALPHA8, glow::RGBA, glow::UNSIGNED_BYTE),
+            f::Format::Bgra8Srgb => (glow::SRGB8_ALPHA8, glow::BGRA, glow::UNSIGNED_BYTE),
```
kvark: do we have to repeat the code from `create_swapchain` here?

omni-viral: `create_swapchain` under `#[cfg(target_arch = "wasm32")]` just calls this function.

src/backend/gl/src/device.rs
```diff
@@ -842,17 +842,17 @@ impl d::Device<B> for Device {
...
-                // Sampler2D won't show up in UniformLocation and the only other uniforms
-                // should be push constants
-                uniforms.push(n::UniformDesc {
-                    location: location as _,
-                    offset,
-                    utype,
-                });
-
-                offset += size as u32;
+                if let Some(location) = gl.get_uniform_location(program, &name) {
```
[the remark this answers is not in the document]
omni-viral: I just made the code to do the same it was doing before porting to glow.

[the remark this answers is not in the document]
kvark: ok, fair enough

kvark:
Ok, I think we can go ahead with this. The rough parts are what we already have, and the PR doesn't regress them.
bors r+

PR #2892, opened 2019-07-07
Implement surface-to-surface copy in GL backend

hecrj:
This is probably incomplete (and wrong). Now, it just always copies the color attachment. I imagine we should check the subresource layers in `data`. I would appreciate some guidance, is that easily implementable from here?

PR checklist:

### DIFF
--- a/src/backend/gl/src/queue.rs
+++ b/src/backend/gl/src/queue.rs
@@ -698,8 +698,60 @@ impl CommandQueue {
             com::Command::CopyImageToTexture(..) => {
                 unimplemented!() //TODO: use FBO
             }
-            com::Command::CopyImageToSurface(..) => {
-                unimplemented!() //TODO: use FBO
+            com::Command::CopyImageToSurface {
+                src_image,
+                dst_surface,
+                dst_format,
+                ref data
+            } => {
+                let gl = &self.share.context;
+
+                if data.src_subresource.aspects != hal::format::Aspects::COLOR
+                    || data.dst_subresource.aspects != hal::format::Aspects::COLOR {
+                        unimplemented!()
+                }
+
+                match src_image {
+                    native::ImageKind::Texture { .. } => {
+                        unimplemented!()
+                    }
+                    native::ImageKind::Surface {
+                        surface: src_surface,
+                        format: src_format
+                    } => {
+                        if src_format != dst_format {
+                            unimplemented!()
+                        }
+
+                        unsafe {
+                            let src_fbo = gl.create_framebuffer().unwrap();
+                            gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(src_fbo));
+                            gl.framebuffer_renderbuffer(glow::READ_FRAMEBUFFER, glow::COLOR_ATTACHMENT0, glow::RENDERBUFFER, Some(src_surface));
+
+                            let dst_fbo = gl.create_framebuffer().unwrap();
+                            gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, Some(dst_fbo));
+                            gl.framebuffer_renderbuffer(glow::DRAW_FRAMEBUFFER, glow::COLOR_ATTACHMENT0, glow::RENDERBUFFER, Some(dst_surface));
+
+                            gl.blit_framebuffer(
+                                data.src_offset.x,
+                                data.src_offset.y,
+                                data.src_offset.x + data.extent.width as i32,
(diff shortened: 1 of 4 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -698,8 +698,44 @@ impl CommandQueue {
...
+                        unsafe {
+                            let src_fbo = gl.create_framebuffer().unwrap();
+                            gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(src_fbo));
+                            gl.framebuffer_renderbuffer(glow::READ_FRAMEBUFFER, glow::COLOR_ATTACHMENT0, glow::RENDERBUFFER, Some(src));
+
+                            let dst_fbo = gl.create_framebuffer().unwrap();
+                            gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, Some(dst_fbo));
+                            gl.framebuffer_renderbuffer(glow::DRAW_FRAMEBUFFER, glow::COLOR_ATTACHMENT0, glow::RENDERBUFFER, Some(dst));
+
+                            gl.blit_framebuffer(
```
kvark: I was writing about 5 different comments on how to improve this, but at the end of the day I realized that blitting an FBO is just wrong here. Blitting is a graphics operation, it effectively samples a surface and writes into another surface. Those samples are writes do the format unpacking/packing, which is *not* what a copy needs. A copy doesn't respect any format conversions: you can copy from RGBA8Uint to R32Float, and it would still work in Vulkan and gfx-hal. But it wouldn't work with a blit operation, since it would only be able to write done one R component.

Unfortunately, GLES or GL Core don't have `glCopyPixels` or `glDrawPixels`, so I'm not entirely sure what to do here in general case... need to give it more thought.

We can still have this code path you wrote, but only when the formats of source and destination match *exactly*.

src/backend/gl/src/queue.rs
```diff
@@ -698,8 +698,44 @@ impl CommandQueue {
...
+                                data.src_offset.x,
+                                data.src_offset.y,
+                                data.src_offset.x + data.extent.width as i32,
+                                data.src_offset.y + data.extent.height as i32,
+                                data.dst_offset.x,
+                                data.dst_offset.y,
+                                data.dst_offset.x + data.extent.width as i32,
+                                data.dst_offset.y + data.extent.height as i32,
+                                glow::COLOR_BUFFER_BIT,
+                                glow::LINEAR,
```
kvark: should be `NEAREST`

src/backend/gl/src/queue.rs
```diff
@@ -698,8 +698,44 @@ impl CommandQueue {
...
+                            );
+
+                            gl.bind_framebuffer(glow::READ_FRAMEBUFFER, None);
```
kvark: can just do `glow::FRAMEBUFFER` to cover both

src/backend/gl/src/queue.rs
hecrj: I see. Thank you for the great explanation!

I had the idea of trying to circumvent this by implementing texture-to-texture copy in `CopyImageToTexture`. I tried changing `create_image` to create textures instead of render buffers when either `TRANSFER_SRC` or `TRANSFER_DST` usage flags are set, and then using the already implemented `CopyTextureToBuffer` and `CopyBufferToTexture` commands to perform the copy operation. However, [`CopyTextureToBuffer` doesn't seem to work on WebGL](https://github.com/grovesNL/glow/blob/master/src/web.rs#L254). In any case, this solution would probably be very slow, right?

It's quite unfortunate that there is no easy way to implement this. Do you know if there is at least an easy way to check the surface format so we can keep this code as you advised? I haven't been able to find a way.

kvark: > However, CopyTextureToBuffer doesn't seem to work on WebGL.

Oh wow, this is sad. @grovesNL - heads up in case you have ideas :)

> In any case, this solution would probably be very slow, right?

It's undesirable. Theoretically, GL renderbuffers could be more efficient than regular textures, but I don't recall any benchmarks confirming this :)

> Do you know if there is at least an easy way to check the surface format

Let's just keep it in the `Surface` type that we have. We know it at creation time, at least.

hecrj:
I have implemented the discussed changes. Let me know if there is anything that still needs to be addressed!

kvark:
bors retru

kvark:
bors retry

kvark:
Bors retry

PR #2894, opened 2019-07-08
Fix attachments confusion

omni-viral:
This changes attachments references confusion that led to `draw_buffers` to contain shifted indices.
Also this opens route to support multiple subpasses in one render pass by having fbo per subpass.

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -102,7 +102,7 @@ pub enum Command {
     /// The buffer slice contains a list of `GLenum`.
     DrawBuffers(BufferSlice),
 
-    BindFrameBuffer(FrameBufferTarget, Option<n::FrameBuffer>),
+    BindFrameBuffer(FrameBufferTarget, Option<n::RawFrameBuffer>),
     BindTargetView(FrameBufferTarget, AttachmentPoint, n::ImageView),
     SetDrawColorBuffers(usize),
     SetPatchSize(i32),
@@ -150,16 +150,16 @@ pub type DrawBuffer = u32;
 
 #[derive(Clone, Debug)]
 struct AttachmentClear {
-    subpass_id: Option<pass::SubpassId>,
-    value: Option<command::ClearValueRaw>,
-    stencil_value: Option<pso::StencilValue>,
+    subpass_id: pass::SubpassId,
+    index: u32,
+    value: command::ClearValueRaw,
 }
 
 #[derive(Debug)]
 pub struct RenderPassCache {
     render_pass: n::RenderPass,
-    framebuffer: Option<n::FrameBuffer>,
-    attachment_clears: Vec<AttachmentClear>,
+    framebuffer: n::FrameBuffer,
+    attachment_clears: Vec<Option<AttachmentClear>>,
 }
 
 // Cache current states of the command buffer
@@ -174,7 +174,7 @@ struct Cache {
     // Blend color.
     blend_color: Option<pso::ColorValue>,
     ///
-    framebuffer: Option<(FrameBufferTarget, n::FrameBuffer)>,
+    framebuffer: Option<(FrameBufferTarget, n::RawFrameBuffer)>,
     ///
     // Indicates that invalid commands have been recorded.
     error_state: bool,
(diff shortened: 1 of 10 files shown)


src/backend/gl/src/command.rs
```diff
@@ -474,54 +474,57 @@ impl RawCommandBuffer {
...
                         None
                     };
 
                     if depth.is_some() || stencil.is_some() {
                         return Some(Command::ClearBufferDepthStencil(depth, stencil));
                     }
+                }
+                None
+            })
+            .collect::<Vec<_>>();
```
kvark: do we have to collect it here?

src/backend/gl/src/command.rs
```diff
@@ -628,41 +631,33 @@ impl command::RawCommandBuffer<Backend> for RawCommandBuffer {
...
-                        Some(cv.unwrap().borrow().depth_stencil.stencil)
-                    } else {
-                        None
-                    },
-                }
+
+                let (subpass, index) = render_pass.subpasses.iter().enumerate().filter_map(|(i, sp)| {
+                    let index = sp.attachment_using(id)?;
+                    Some((i, index))
+                }).next()?;
```
kvark: this looks weird to me: so we are constructing an iterator from `subpasses` and only checking the first element of it?

src/backend/gl/src/device.rs
```diff
@@ -624,12 +624,15 @@ impl d::Device<B> for Device {
         ID: IntoIterator,
         ID::Item: Borrow<pass::SubpassDependency>,
     {
+        let max_attachments = self.share.context.get_parameter_i32(glow::MAX_COLOR_ATTACHMENTS) as usize;
```
kvark: we should have this in the limits, so this line should be re-using the limits instead of re-querying it

src/backend/gl/src/command.rs
omni-viral: Exactly.

src/backend/gl/src/command.rs
omni-viral: We need access by index

src/backend/gl/src/command.rs
kvark: I suppose it would look cleaner with `iter().position()` instead, but it hardly matters :)

src/backend/gl/src/command.rs
kvark: where are we accessing `clear_cmds` by index?

src/backend/gl/src/command.rs
omni-viral: It was cleaner with `.position()`, but I need not only position now.

src/backend/gl/src/command.rs
omni-viral: We don't, I confused it with another piece of code

omni-viral: Apparently we need to let go of `self` borrow, so collecting is unavoidable.

kvark: The `self` is only borrowed for `self.cur_subpass`? In this case, we can copy it into a variable and not hold `self`

src/backend/gl/src/command.rs
```diff
@@ -433,37 +433,37 @@ impl RawCommandBuffer {
...
+            vec![glow::BACK_LEFT]
+        } else {
+            subpass
+                .color_attachments
                 .iter()
-                .zip(state.attachment_clears.iter())
-                .filter_map(|(attachment, clear)| {
+                .enumerate()
+                .map(|(index, _)| glow::COLOR_ATTACHMENT0 + index as u32)
+                .collect::<Vec<_>>()
```
kvark: technically, this doesn't need to collect either

src/backend/gl/src/command.rs
omni-viral: No. `self` is borrowed for `state` which owns collections we iterate over.

src/backend/gl/src/command.rs
omni-viral: How? We need a slice.

kvark: does `self.add` really need a slice as opposed to an iterator?

PR #2895, opened 2019-07-09
Fix WebGL always presenting latest bound FBO

hecrj:
When working on #2892, I noticed that when using WebGL with the GL backend, the latest bound FBO was being blitted to the default framebuffer, instead of the correct swapchain FBO.

The changes in this PR fix this issue while:
  - Unifying swapchain creation and presentation
  - Removing related unused code
  - Removing a related unnecessary unsafe block

PR checklist:

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1838,17 +1838,6 @@ impl d::Device<B> for Device {
         unimplemented!()
     }
 
-    #[cfg(target_arch = "wasm32")]
-    unsafe fn create_swapchain(
-        &self,
-        surface: &mut Surface,
-        config: c::SwapchainConfig,
-        _old_swapchain: Option<Swapchain>,
-    ) -> Result<(Swapchain, Vec<n::Image>), c::window::CreationError> {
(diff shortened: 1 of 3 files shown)


src/backend/gl/src/device.rs
```diff
@@ -1857,16 +1846,6 @@ impl d::Device<B> for Device {
...
 
-        let (int_format, iformat, itype) = match config.format {
-            Format::Rgba8Unorm => (glow::RGBA8, glow::RGBA, glow::UNSIGNED_BYTE),
-            Format::Bgra8Unorm => (glow::RGBA8, glow::BGRA, glow::UNSIGNED_BYTE),
-            Format::Rgba8Srgb => (glow::SRGB8_ALPHA8, glow::RGBA, glow::UNSIGNED_BYTE),
-            Format::Bgra8Srgb => (glow::SRGB8_ALPHA8, glow::BGRA, glow::UNSIGNED_BYTE),
-            _ => unimplemented!(),
-        };
-
-        let channel = config.format.base_format().1;
```
hecrj: This code wasn't actually being used...

kvark:
bors r=msiglreith

PR #2897, opened 2019-07-10
backport iOS metal sim support to hal0.2 branch

mtak-:
- [x] tested examples with the following backends: metal ios/mac


kvark:
https://crates.io/crates/gfx-backend-metal/0.2.3

Issue #2571, opened 2019-01-15
Occasional incorrect use of Range


Lokathor:
The [Range](https://doc.rust-lang.org/core/ops/struct.Range.html) struct is _exclusive_ while [RangeInclusive](https://doc.rust-lang.org/core/ops/struct.RangeInclusive.html) is naturally the _inclusive_ version. There are some parts of the system that use `Range` to provide inclusive values. This is a semantic error that can lead to very confused programmers down the line.

Specifically, with [SurfaceCapabilities](https://docs.rs/gfx-hal/0.1.0/gfx_hal/window/struct.SurfaceCapabilities.html) on win32 Vulkan and DX12:
* `extents`: Range<Extent2D>
* `image_count`: Range<SwapImageIndex>

(Possibly other places in the library, I'm very new here obviously.)

Now this isn't quite as simple as switching from `Range` to `RangeInclusive`, because the `RangeInclusive` carries an extra `Option<bool>` field for whatever obscure iteration based reason. So it throws extra overhead into the pile if we use `RangeInclusive`. It's probably better to just use tuples (really) and document that they're inclusive at both ends.

rileylyman:
Is it a concern that changing `Range` to `(u32, u32)`, for example, could break a lot of user code? Would it be better to change how other parts of the library's code uses these fields? For example, instead of `surface_caps.image_count = start..end` we write `surface_caps.image_count = start..(end + 1)`.

Lokathor:
0.2 will probably have a whole bunch of minor breaks in general. It's probably better to just do the break, because the `Range` type doesn't super fit in the first place. Using `Range` means that plain description things that we'd want to just be `Copy` don't end up being `Copy`, which is annoying.

rileylyman:
Ok, I can try changing them to tuples.

kvark:
My vote would be to use `RangeInclusive` where appropriate. Ranges have strong semantics (that is better than a comment on a tuple) and good ergonomics. The only downside is that they aren't `Copy`, but this is most of the time easy to work with (please show me cases where it's not the case).

Lokathor:
It's more annoying than anything else. You have to put `clone` all over the place, which freaks people out because they think you're doing some expensive thing when you're not.

memoryruins:
[...]

PR #2901, opened 2019-07-12
[mtl] fix incorrect `image_count` caps

aleksijuvani:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/metal/src/window.rs
+++ b/src/backend/metal/src/window.rs
@@ -328,14 +328,16 @@ impl hal::Surface<Backend> for Surface {
         let device_caps = &device.shared.private_caps;
 
         let can_set_maximum_drawables_count =
-            device_caps.os_is_mac || device_caps.has_version_at_least(11, 2);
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/window.rs
```diff
@@ -328,14 +328,16 @@ impl hal::Surface<Backend> for Surface {
...
-            2 .. 4
+            // You can set this value to 2 or 3 only; other values are ignored
+            // and an exception is thrown
+            2 .. 3
         } else {
-            // 3 is the default in `CAMetalLayer` documentation
-            // iOS 10.3 was tested to use 3 on iphone5s
-            3 .. 4
+            // The default value is 3
+            3 .. 3
```
kvark: the `Range` in Rust has a non-inclusive semantics on the right, so this would be a wrong range

aleksijuvani: I see. I was confused because the Vulkan backend was providing inclusive values with `Range` for `image_count`. So we should instead address the other backends so that the usage of `Range` is consistent?

kvark: Yes. I fact, we have #2571 mentioning this.

PR #2904, opened 2019-07-12
Decouple window from gl context

omni-viral:
Also remove `Surface::kind` and fix few rough points.

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -998,7 +998,20 @@ impl d::Device<B> for Device {
                 }
 
             }
-            let _status = gl.check_framebuffer_status(target); //TODO: check status
+            
+            let status = gl.check_framebuffer_status(target);
+            match status {
+                glow::FRAMEBUFFER_COMPLETE => {},
+                glow::FRAMEBUFFER_INCOMPLETE_ATTACHMENT => panic!("One of framebuffer attachmet points are incomplete"),
+                glow::FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT => panic!("Framebuffer does not have any image attached"),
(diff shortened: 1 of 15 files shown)


src/backend/gl/src/device.rs
```diff
@@ -998,7 +998,20 @@ impl d::Device<B> for Device {
...
-            let _status = gl.check_framebuffer_status(target); //TODO: check status
+            
+            let status = gl.check_framebuffer_status(target);
+            match status {
+                glow::FRAMEBUFFER_COMPLETE => {},
+                glow::FRAMEBUFFER_INCOMPLETE_ATTACHMENT => panic!("One of framebuffer attachmet points are incomplete"),
+                glow::FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT => panic!("Framebuffer does not have any image attached"),
+                glow::FRAMEBUFFER_INCOMPLETE_DRAW_BUFFER => panic!("FRAMEBUFFER_INCOMPLETE_DRAW_BUFFER"),
+                glow::FRAMEBUFFER_INCOMPLETE_READ_BUFFER => panic!("FRAMEBUFFER_INCOMPLETE_READ_BUFFER"),
+                glow::FRAMEBUFFER_UNSUPPORTED => panic!("FRAMEBUFFER_UNSUPPORTED"),
```
kvark: why can't we just `panic!("{:?}", status)` instead of repeating ourselves?

kvark:
There is a lot in common with https://github.com/gfx-rs/gfx/pull/2882/commits/93932abf832570c239778d44f1886465ced7eff8
Looks good to me!.
bors r+

src/backend/gl/src/device.rs
omni-viral: Because you will get an integer

kvark:
bors retry

PR #2907, opened 2019-07-18
Use all device features provided by vulkan

4lDO2:
PR checklist:


kvark:
bors retry

PR #2911, opened 2019-07-23
No Winit by default

kvark:
Unblocks https://github.com/gfx-rs/wgpu/issues/64
cc @seivan @cosmicchipsocket
PR checklist:
- [x] tested examples with the following backends: Vulkan/X11, GL/glutin/X11

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -38,16 +38,9 @@ gfx-backend-empty = { path = "../src/backend/empty", version = "0.2" }
 winit = "0.19"
 glsl-to-spirv = "0.1.4"
 
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies.gfx-backend-gl]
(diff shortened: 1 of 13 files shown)


cosmicchipsocket:
Is a patch coming for wgpu as well? wgpu seems to depend on wgpu-native with feature "window-winit"

https://github.com/gfx-rs/wgpu-rs/blob/d12c7329cc39ad74ac6c3d46b766be0610b66d73/Cargo.toml#L27

kvark:
@cosmicchipsocket these need to be published first in order for the wgpu crates to pick up the changes. I think we'll try to fit this into 0.3 gfx-hal release, ideally, since it's not an urgent issue.
After wgpu-native is updated, we can start talking about `wgpu-rs` patching.

seivan:
Nice, thank you for this!

PR #2914, opened 2019-07-25
[WIP] Experimental RangeBounds use

### LINKED ISSUE #2910: Better Range APIs
Current API has things like `Range<Option<Index>>` to simulate ranges that can be non-bounded from either side. This is not looking good: `None .. None`. We could introduce our own enums for begin/end of those ranges, or better - we could make our methods generic over [std RangeBounds](https://doc.rust-lang.org/stable/std/ops/trait.RangeBounds.html), which is stable.

kvark:
Fixes #2910
PR checklist:

Missing backends: dx11, dx12, metal

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -44,6 +44,7 @@ struct Dimensions<T> {
 use std::cell::RefCell;
 use std::io::Cursor;
 use std::mem::size_of;
+use std::ops::{RangeBounds, RangeFull};
 use std::rc::Rc;
 use std::{fs, iter};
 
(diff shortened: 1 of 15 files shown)


kvark:
I tried to prototype quickly and see if this works. The ergonomic suffers because, again, we can't set the default range type for cases where no buffers are involved. Wins are quite arguable. All in all, it doesn't seem to be worth it exploring.

Thoughts?

kvark:
thanks @Gankro for the link to https://github.com/rust-lang/rust/issues/27336 explaining why we can't have nice things

msiglreith:
hmmm.. that looks quite unergonomic. We can even support different ranges types when it comes to arrays I guess? I'm still wondering if exposing a `!0` constant would have been easier..

kvark:
>  I'm still wondering if exposing a !0 constant would have been easier..

I think it would be slightly better than what we have today.
Given 2 thumbs down so far, I think this can be closed.

PR #2922, opened 2019-07-29
[dx12] allow view creation to fail

### LINKED ISSUE #2921: [dx12] don't create RTVs when not possible
Current DX12 code for view creation checks for the image usage and attempts to create all of DSV/RTV/SRV/UAV based on the original image usage. When the user creates a view over multiple levels, neither of DSV/RTV are possible, and we currently panic. We should be handling that gracefully and just accept the fact there is no associated DSV/RTV if it's not possible.

This affects wgpu-rs mipmap example.

kvark:
Fixes #2921
PR checklist:

### DIFF
--- a/src/backend/dx12/Cargo.toml
+++ b/src/backend/dx12/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-backend-dx12"
-version = "0.2.1"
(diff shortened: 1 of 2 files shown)


kvark:
Published in https://crates.io/crates/gfx-backend-dx12

PR #2923, opened 2019-07-29
[dx12] allow view creation to fail

kvark:
Sibling of #2922


kvark:
The sibling got reviewed already, merging.
bors r=msiglreith

Issue #2875, opened 2019-06-27
Support multisample resolve targets


kvark:
#2813 added support for resolve targets in the Vulkan backend
We still need to add `ATTACHMENT_UNUSED` in the API - #2861
Also related to #2859 rename.

This issue is about catching up with other backends by implementing resolve targets in DX/GL/Metal.

msiglreith:
Resolve should be mostly done for dx12, I tested it via [portability](https://github.com/SaschaWillems/Vulkan#02---multi-sampling) IIRC.
ATTACHMENT_UNUSED is still an open point ofc

PR #2924, opened 2019-07-30
[hal-0.2/mtl] attachment resolves

kvark:
Addresses Metal side of #2875 
PR checklist:
- [ ] tested examples with the following backends

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -3221,7 +3221,7 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                     descriptor.set_render_target_array_length(framebuffer.extent.depth as _);
                 }
 
-                for (i, &(at_id, op_flags)) in subpass.colors.iter().enumerate() {
+                for (i, &(at_id, op_flags, resolve_id)) in subpass.colors.iter().enumerate() {
                     let rat = &render_pass.attachments[at_id];
                     let texture = &framebuffer.attachments[at_id];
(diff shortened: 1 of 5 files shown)


src/backend/metal/src/command.rs
```diff
@@ -3237,7 +3237,15 @@ impl com::RawCommandBuffer<Backend> for CommandBuffer {
                             desc.set_clear_color(channel.interpret(raw));
                         }
                     }
-                    if op_flags.contains(native::SubpassOps::STORE) {
+                    if let Some(id) = resolve_id {
+                        let resolve = &framebuffer.attachments[id];
+                        desc.set_resolve_texture(Some(resolve));
+                        // these should only be encoded into `resolve` MTLTexture
+                        //desc.set_resolve_level();
```
grovesNL: Are these still `TODO`, or could you clarify what the comment means?

kvark: These aren't TODOs, it's something that we could be doing, but we don't need to, since our image views are already technically child MTLTexture objects that point to the right slices/levels. Do you want these comments removed?

grovesNL: Oh ok, sounds good :+1: I thought that's what it meant but wasn't sure based on the comment. Maybe we could say explicitly that the image views already handle it.

kvark: fixed now :)

kvark:
bors r=grovesNL

PR #2929, opened 2019-07-31
hal: dedicated buffer/image memory

### LINKED ISSUE #2511: Provide "committed resource" semantics in HAL

kvark:
Fixes #2511 
Closes #2513

The idea is that backends can implement this more efficiently than the default implementation.

PR checklist:

### DIFF
--- a/src/backend/vulkan/Cargo.toml
+++ b/src/backend/vulkan/Cargo.toml
@@ -29,6 +29,7 @@ ash = "0.29.0"
 gfx-hal = { path = "../../hal", version = "0.2" }
 smallvec = "0.6"
 winit = { version = "0.19", optional = true }
+#vk-mem = { version = "0.1", optional = true }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3", features = ["libloaderapi", "windef", "winuser"] }
--- a/src/hal/src/device.rs
+++ b/src/hal/src/device.rs
@@ -411,6 +411,24 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
         buf: &mut B::Buffer,
     ) -> Result<(), BindError>;
 
+    /// Create a dedicated allocation for a buffer.
+    ///
+    /// Returns a memory object that can't be used for binding any resources.
+    unsafe fn allocate_buffer_memory(
+        &self,
+        memory_type: MemoryTypeId,
+        buffer: &mut B::Buffer,
+    ) -> Result<B::Memory, BindError> {
+        let req = self.get_buffer_requirements(buffer);
+        let memory = self.allocate_memory(memory_type, req.size)
+            .map_err(|err| match err {
+                AllocationError::OutOfMemory(oom) => BindError::OutOfMemory(oom),
+                AllocationError::TooManyObjects => BindError::OutOfBounds,
+            })?;
+        self.bind_buffer_memory(&memory, 0, buffer)
+            .map(|()| memory)
+    }
+
     /// Destroy a buffer.
     ///
     /// The buffer shouldn't be destroyed before any submitted command buffer,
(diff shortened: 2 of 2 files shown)


src/backend/vulkan/Cargo.toml
```diff
@@ -29,6 +29,7 @@ ash = "0.29.0"
 gfx-hal = { path = "../../hal", version = "0.2" }
 smallvec = "0.6"
 winit = { version = "0.19", optional = true }
+#vk-mem = { version = "0.1", optional = true }
```
msiglreith: should be removed

kvark:
It's there for future work on the topic ;)

src/hal/src/device.rs
```diff
@@ -411,6 +411,24 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
         buf: &mut B::Buffer,
     ) -> Result<(), BindError>;
 
+    /// Create a dedicated allocation for a buffer.
+    ///
+    /// Returns a memory object that can't be used for binding any resources.
+    unsafe fn allocate_buffer_memory(
```
omni-viral: To actually do that more efficiently, the function must return a view into memory with offset and size.

kvark: The user is expected to still call `get_buffer_requirements` before doing this, so the size is known. And the offset is known to be zero. Is this not enough?

omni-viral: Oh, so this will work only for dedicated allocations?

omni-viral: I thought we want to integrate vma into vulkan backend this way

kvark: I think what we can do with VMA is:
  1. make it a compile time optional dependency
  2. turn `Memory` into an enum, with one variant for normal allocations and another - for VMA-owned ones. The code dealing with memory would need to match the enum, but I don't think it's a problem.
  3. the user is then expected to treat the `Memory` as dedicated allocation

Does that sound reasonable?

omni-viral: Then you will map ranges from two of those `VmaMemory` instances that actually comes from single `RealMemory` and violate vulkan usage.

kvark: So the memory type exposed by the Vulkan backend would be:
```rust
enum Memory {
  Physical(vk::Memory),
  SubAllocated(vma::Allocation),
}
```
Mapping a `Memory::SubAllocated` will go though VMA and just give out a pointer without really doing any more Vulkan mapping, if I understand correctly, much like with Rendy-memory.

omni-viral: If VMA has no limitations here. Rendy has to map all non-dedicated mappable memory persistently to allow this, which is not optimal. I wonder how VMA does that.

kvark: > which is not optimal

Is it, really? My understanding is that it's both optimal and recommended for CPU-visible memory.

omni-viral: It's not a problem when RAM is abundant, like on typical modern PC. But in other cases it may be not the best approach.

kvark: Looks like it doesn't map memory by default on creation, unless there is a specific flag provided. Relevant docs - https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/memory_mapping.html
It confirms that VMA acts roughly like rendy-memory, and the approach this PR is taking should work.

kvark:
I suppose I can integrate VMA in this PR as a validation of the API.

PR #2930, opened 2019-08-01
[mtl] fix missing depth formats on iOS

aleksijuvani:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/metal/src/conversions.rs
+++ b/src/backend/metal/src/conversions.rs
@@ -25,8 +25,10 @@ impl PrivateCapabilities {
             f::Rg8Srgb if self.format_min_srgb_channels <= 2 => RG8Unorm_sRGB,
             f::Rgba8Srgb if self.format_min_srgb_channels <= 4 => RGBA8Unorm_sRGB,
             f::Bgra8Srgb if self.format_min_srgb_channels <= 4 => BGRA8Unorm_sRGB,
+            f::D16Unorm if self.format_depth16unorm => Depth16Unorm,
             f::D24UnormS8Uint if self.format_depth24_stencil8 => Depth24Unorm_Stencil8,
-            f::D32SfloatS8Uint if self.format_depth32_stencil8_filter => Depth32Float_Stencil8,
+            f::D32Sfloat => Depth32Float,
+            f::D32SfloatS8Uint => Depth32Float_Stencil8,
             f::R8Unorm => R8Unorm,
             f::R8Snorm => R8Snorm,
             f::R8Uint => R8Uint,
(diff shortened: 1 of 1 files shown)


src/backend/metal/src/conversions.rs
```diff
@@ -25,8 +25,10 @@ impl PrivateCapabilities {
             f::Rg8Srgb if self.format_min_srgb_channels <= 2 => RG8Unorm_sRGB,
             f::Rgba8Srgb if self.format_min_srgb_channels <= 4 => RGBA8Unorm_sRGB,
             f::Bgra8Srgb if self.format_min_srgb_channels <= 4 => BGRA8Unorm_sRGB,
+            f::D16Unorm if self.format_depth16unorm => Depth16Unorm,
             f::D24UnormS8Uint if self.format_depth24_stencil8 => Depth24Unorm_Stencil8,
-            f::D32SfloatS8Uint if self.format_depth32_stencil8_filter => Depth32Float_Stencil8,
+            f::D32Sfloat => Depth32Float,
+            f::D32SfloatS8Uint => Depth32Float_Stencil8,
```
kvark: why are we no longer checking `format_depth32_stencil8_filter` here?

aleksijuvani: This format is available if either `format_depth32_stencil8_filter` or `format_depth32_stencil8_none` is set. If we're on macOS, then the former is set, but on iOS the latter is set. So, it's always available.

kvark: I think `format_depth32_stencil8_none` means that it's not supported. These flags are somewhat confusing to me and beg for a refactor.

aleksijuvani: If we look at the [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) for Pixel Format Capabilities and compare these variables against that, it looks to me that the intention here was to indicate with `_none` that the format does not have the "Filter" capability, but is supported, while `_filter` indicates that it also has the "Filter" capability. The "Filter" capability seems to map to the `SAMPLED_LINEAR` flag in gfx-hal.

However, it's weird to me that `map_format` returns an `Option` in the first place. Surely it should map the format regardless of if it's supported? Perhaps this behaviour is relied upon elsewhere in the codebase. It might be something to keep in mind for a refactor.

kvark: We don't need the _none then and instead should imply this based on the other 2 flags.

> it's weird to me that map_format returns an Option in the first place

I can see where you are coming from. Our logic is based on the following idea: if a type is not supported by the platform, there should be no benefit of producing an unsupported mapping. We could potentially change this internal semantics, but there are still going to be types that don't have an MTL counterpart, so I don't see how you could get rid of `Option`?

aleksijuvani: > I can see where you are coming from. Our logic is based on the following idea: if a type is not supported by the platform, there should be no benefit of producing an unsupported mapping. We could potentially change this internal semantics, but there are still going to be types that don't have an MTL counterpart, so I don't see how you could get rid of `Option`?

I see! It did not occur to me that there are also formats that don't have a corresponding format in Metal. That completely makes sense.

PR #2931, opened 2019-08-01
Remove usages of range that dont describe linear values

rukai:
Fixes https://github.com/gfx-rs/gfx/issues/2571
PR checklist:
   - linux opengl + vulkan

~~- [ ] `rustfmt` run on changed code~~

I haven't fully checked that this covers every case, I just went through the examples and changed every case that pops up there. Can thoroughly replace everything in this PR or a followup.

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -792,7 +792,8 @@ impl<B: Backend> RenderPassState<B> {
                     pass::AttachmentStoreOp::Store,
                 ),
                 stencil_ops: pass::AttachmentOps::DONT_CARE,
-                layouts: i::Layout::Undefined .. i::Layout::Present,
+                initial_layout: i::Layout::Undefined,
+                final_layout: i::Layout::Present,
             };
 
             let subpass = pass::SubpassDesc {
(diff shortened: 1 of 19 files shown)


examples/colour-uniform/main.rs
```diff
@@ -804,11 +805,12 @@ impl<B: Backend> RenderPassState<B> {
             };
 
             let dependency = pass::SubpassDependency {
-                passes: pass::SubpassRef::External .. pass::SubpassRef::Pass(0),
-                stages: PipelineStage::COLOR_ATTACHMENT_OUTPUT
-                    .. PipelineStage::COLOR_ATTACHMENT_OUTPUT,
-                accesses: i::Access::empty()
-                    .. (i::Access::COLOR_ATTACHMENT_READ | i::Access::COLOR_ATTACHMENT_WRITE),
+                src_subpass: pass::SubpassRef::External,
```
kvark: My biggest concern here is duplication of code. Would you be open to a new struct that encloses the duplicated types? Something like

```
struct DependencySide {
  subpass: SubRef,
  stage: PipelineStage,
  access: image::Access,
}
struct SubpassDependency {
   src: DependencySide,
   dst: DependencySide,
}
```

examples/colour-uniform/main.rs
```diff
@@ -1228,15 +1230,18 @@ impl<B: Backend> ImageState<B> {
             cmd_buffer.begin();
 
             let image_barrier = m::Barrier::Image {
-                states: (i::Access::empty(), i::Layout::Undefined)
-                    .. (i::Access::TRANSFER_WRITE, i::Layout::TransferDstOptimal),
+                src_access: i::Access::empty(),
```
kvark: we already have a nice `type State = (Access, Layout)` for images, let's use it here

examples/colour-uniform/main.rs
rukai: Duplication of user code? Or duplication of gfx-hal code?

What are your thoughts on keeping the public API with `src_`/`dest_` fields and then using the DependencySide internally.

kvark: duplication on the API side is what I mean

rukai:
I've lost interest in this pr and can't see myself working on it again. Sorry to close it after review had started :/

Still interested in other gfx/wgpu work in the future but I'm fairly busy irl at the moment.

PR #2932, opened 2019-08-01
[mtl] update parking_lot dependency to 0.9.0

aleksijuvani:
Depends on kvark/storage-map#2 and needs a new version published of storage-map
PR checklist:


kvark:
thank you!
bors r+

PR #2934, opened 2019-08-01
[0.2/dx12] refactored root signature creation, improved shader visibility

kvark:
Just improves the root signature a bit...
PR checklist:
- [ ] `make` succeeds (on *nix)
- [ ] `make reftests` succeeds
- [ ] tested examples with the following backends:
- [ ] `rustfmt` run on changed code

PR #2935, opened 2019-08-01
[dx12] refactored root signature creation, improved shader visibility

### LINKED ISSUE #2643: dx12: Stricter shader visibility flags
e.g https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12/src/device.rs#L1333

dx12 supports resource shader visibility for one particular shader or for all. We currently just set it always to ALL, which may have some performance impacts.

kvark:
Sibling of #2934 for `master`
Fixes #2643


kvark:
bors r=msiglreith

PR #2937, opened 2019-08-02
[mtl] fix `CAMetalLayer` creation and resize on iOS

aleksijuvani:
Fixes #issue
PR checklist:

r? @mtak-

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -282,28 +282,22 @@ impl Instance {
             // If the main layer is not a CAMetalLayer, we create a CAMetalLayer sublayer and use it instead.
             // Unlike on macOS, we cannot replace the main view as UIView does not allow it (when NSView does).
             let new_layer: CAMetalLayer = msg_send![class, new];
-            let bounds: CGRect = msg_send![view, bounds];
-            msg_send![new_layer, setBounds: bounds];
 
-            let frame: CGRect = msg_send![view, frame];
-            msg_send![new_layer, setFrame: frame];
+            let bounds: CGRect = msg_send![main_layer, bounds];
+            msg_send![new_layer, setFrame: bounds];
 
             msg_send![main_layer, addSublayer: new_layer];
             new_layer
         };
 
         let window: cocoa::base::id = msg_send![view, window];
-        if window.is_null() {
-            panic!("surface is not attached to a window");
-        }
+        if !window.is_null() {
+            let screen: cocoa::base::id = msg_send![window, screen];
+            assert!(!screen.is_null(), "window is not attached to a screen");
 
-        let screen: cocoa::base::id = msg_send![window, screen];
-        if screen.is_null() {
-            panic!("window is not attached to a screen");
+            let scale_factor: CGFloat = msg_send![screen, nativeScale];
+            msg_send![view, setContentScaleFactor: scale_factor];
         }
-
-        let scale_factor: CGFloat = msg_send![screen, nativeScale];
-        msg_send![view, setContentScaleFactor: scale_factor];
 
         msg_send![view, retain];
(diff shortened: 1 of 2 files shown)


src/backend/metal/src/window.rs
```diff
@@ -405,6 +405,14 @@ impl Device {
         let cmd_queue = self.shared.queue.lock();
 
         unsafe {
+            // On iOS, unless the user supplies a view with a CAMetalLayer, we
+            // create one as a sublayer. However, when the view changes size,
+            // its sublayers are not automatically resized, and we must resize
+            // it here. The drawable size and the layer size don't correlate
+            let view = surface.inner.view.unwrap().as_ptr();
```
kvark: should this be ios-only?

aleksijuvani: It didn't seem to have any ill effect on macOS, so it probably doesn't matter. We could make it iOS only. On a separate note, it occurred to me that `view` may be `None` if the surface is created with `create_surface_from_layer` instead of `create_surface_from_uiview`. I'll address these now.

src/backend/metal/src/lib.rs
```diff
@@ -282,8 +282,6 @@ impl Instance {
             // If the main layer is not a CAMetalLayer, we create a CAMetalLayer sublayer and use it instead.
             // Unlike on macOS, we cannot replace the main view as UIView does not allow it (when NSView does).
             let new_layer: CAMetalLayer = msg_send![class, new];
-            let bounds: CGRect = msg_send![view, bounds];
-            msg_send![new_layer, setBounds: bounds];
 
             let frame: CGRect = msg_send![view, frame];
             msg_send![new_layer, setFrame: frame];
```
mtak-: I believe this is more correct as `[new_layer setFrame: [main_layer bounds]]`. Frames are always in the parents coordinate space, and `bounds` are in view/layer's own coordinate space.

src/backend/metal/src/window.rs
```diff
@@ -405,6 +405,17 @@ impl Device {
...
 
         unsafe {
+            // On iOS, unless the user supplies a view with a CAMetalLayer, we
+            // create one as a sublayer. However, when the view changes size,
+            // its sublayers are not automatically resized, and we must resize
+            // it here. The drawable size and the layer size don't correlate
+            #[cfg(target_os = "ios")] {
+                if let Some(view) = surface.inner.view {
+                    let frame: CGRect = msg_send![view.as_ptr(), frame];
+                    msg_send![render_layer, setFrame: frame];
```
mtak-: Similar situation here, but it only matters in corner cases.

mtak-:
Thanks!

kvark:
bors r-
@mtak- wow, thank you for jumping on the review!
@aleksijuvani do you want to address these here? or as a follow-up?

src/backend/metal/src/lib.rs
mtak-: Maybe `[CALayer setAutoresizingMask: kCALayerWidthSizable | kCALayerHeightSizable]` is an alternate solution to resizing in build swapchain? I haven't tested it, but the docs look good.

aleksijuvani:
Thank you for the review, @mtak-! I can address these here. It's getting late so it might take me until tomorrow to finish doing so.

src/backend/metal/src/lib.rs
mtak-: Ahh `autoresizingMask` is macos only.

kvark: :+1: moved to https://github.com/gfx-rs/gfx/issues/2938

aleksijuvani: I've now addressed this here for iOS. We should handle #2938 for macOS in a separate PR.

Issue #2933, opened 2019-08-01
Read-only storage bindings


kvark:
Vulkan has this strange semantics: at the binding model level, there is no "read-only" option for storage resources. But at the shader level, one can specify a storage resource can be decorated with `NonReadable` or `NonWritable`. When such "NonWritable" is encountered by our DX12 backend (say, produced by "readonly" GLSL SSBO), the translated shader uses non-writable variant, e.g. `ByteAddressBuffer` as opposed to `RWByteAddressBuffer`. There is a crucial difference here because DX12 binding model considers the former to be SRV and the latter to be UAV... but we have no means to detect this at the pipeline creation time.

So there is an obvious conflict with DX12 binding model semantics. I could think of a number of ways to address this:
  1. Change our API to include this semantics in the descriptor set layout, potentially refactor the descriptor types completely. This is likely acceptable for all the clients *except* for Vulkan Portability, which we don't want to break.
  2. Convince the shader translator (be it SPIRV-Cross or the future Javelin) to declare these storage resources as RW all the time. This is probably the easiest path, but it may have unintended consequences, since DX12 driver would have to consider it being mutable now.
  3. In the Root Signature, for each storage type descriptor, generate both SRV and UAV entries. When the pipeline is created, we could extract this semantics from the shader and override the binding to either the SRV or UAV accordingly. This has a downside of wasting the descriptor space for storage types.

Also, a combination of these methods is possible: have the flag in the API that specifies read-only/write-only/both. Have the Root Signature creation code respect that setting, and for the "both" case still make the override based on the shader. This solution would allow both get the maximum of the API and keep the compatibility with Vulkan. We'd also consider upstreaming the API changes to VkPI in this case.

kvark:
Good news is - we don't really need to introspect the SPIR-V and set up the overrides based on the "readonly" flag: Root Signature spaces for UAV and SRV are separate, so we can just bind both, and let it figure out which one is needed.

PR #2940, opened 2019-08-02
[dx12] read-only storage descriptors

kvark:
Addresses the DX12 side of #2933 
PR checklist:

@msiglreith please take a careful look at it. I still need to run it over CTS to see if anything regressed, but the code should be ready for the review now.

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -1361,18 +1361,7 @@ impl d::Device<B> for Device {
                 let mut sum = 0;
                 for binding in desc_set.borrow().bindings.iter() {
                     let content = r::DescriptorContent::from(binding.ty);
-                    if content.contains(r::DescriptorContent::CBV) {
-                        sum += 1;
-                    }
-                    if content.contains(r::DescriptorContent::SRV) {
-                        sum += 1;
-                    }
-                    if content.contains(r::DescriptorContent::UAV) {
-                        sum += 1;
-                    }
-                    if content.contains(r::DescriptorContent::SAMPLER) {
-                        sum += 1;
-                    }
+                    sum += content.bits().count_ones() as usize;
                 }
                 sum
             })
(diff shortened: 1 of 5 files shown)


kvark:
I wasn't able to test this properly in CTS, facing the problem in #2941 
However, I figured `gfx_ocean` would be fairly good test. It's extremely useful - it allowed me to discover a few flaws, and now everything is working properly, and I'm fairly confident in the code. Well, maybe not *too* confident to backport it to hal-0.2 branch just yet :)

@msiglreith please take a look

proof:
gfx-ocean

src/backend/dx12/src/device.rs
```diff
@@ -1358,16 +1359,21 @@ impl d::Device<B> for Device {
...
-
-                for binding in bindings {
-                    sum += if binding.ty == pso::DescriptorType::CombinedImageSampler {
-                        2
-                    } else {
-                        1
-                    };
+                for binding in desc_set.borrow().bindings.iter() {
+                    let content = r::DescriptorContent::from(binding.ty);
+                    if content.contains(r::DescriptorContent::CBV) {
```
msiglreith: might get away with `sum += content.bits().count_ones()`

src/backend/dx12/src/device.rs
```diff
@@ -2533,11 +2559,13 @@ impl d::Device<B> for Device {
         let mut src_views = Vec::new();
         let mut num_samplers = Vec::new();
         let mut num_views = Vec::new();
+        info!("write_descriptor_sets");
```
msiglreith: this should be removed

src/backend/dx12/src/device.rs
```diff
@@ -2580,74 +2645,102 @@ impl d::Device<B> for Device {
                                 CounterOffsetInBytes: 0,
                                 Flags: d3d12::D3D12_BUFFER_UAV_FLAG_RAW,
                             };
+                            if heap.is_full() {
```
msiglreith: why is this one here?

src/backend/dx12/src/resource.rs
```diff
@@ -389,6 +392,48 @@ pub struct Memory {
...
+bitflags! {
+    /// A set of D3D12 descriptor types that need to be associated
+    /// with a single gfx-hal `DescriptorType`.
+    #[derive(Default)]
+    pub struct DescriptorContent: u8 {
+        const CBV = 0x1;
+        const SRV = 0x2;
+        const UAV = 0x4;
+        const SAMPLER = 0x8;
+        const VIEW = 0x7;
```
msiglreith: `const VIEW = Self::CBV.bits | Self::SRV.bits | Self::UAV.bits;`

src/backend/dx12/src/device.rs
kvark: I think it's good to be here. We make sure the pool index is pointing to a pool that can allocate for the next iteration.

src/backend/dx12/src/resource.rs
kvark: unfortunately that doesn't work: it expects `u8` but the expression is of type `Self`

kvark:
Thank you for the review! Let's :rocket: 
bors r=msiglreith

kvark:
bors retry

src/backend/dx12/src/resource.rs
msiglreith: should be working: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9abc4c17f80253a0e4e2b809c81bbfac

kvark:
Ah I see, you are orring bits fields.

PR #2944, opened 2019-08-08
Dynamic PSO states refactor

### LINKED ISSUE #2905: More dynamic states
See https://github.com/gfx-rs/wgpu/issues/248#issuecomment-512537277
  - line width
  - stencil compare mask
  - stencil write mask
  - stencil ref

kvark:
Fixes #2905
PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -1399,10 +1399,10 @@ impl<B: Backend> PipelineState<B> {
                     &pipeline_layout,
                     subpass,
                 );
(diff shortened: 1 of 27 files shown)


kvark:
Side note on enum -> Option conversions. At some point in the past, I suggested to use enums for cleaner semantics at the type level. Now it's more clear that type level clarity is less important here as just being straight and simple. Option gives you that, without sacrificing the type safety.

kvark:
Ok, I don't think there is a lot of controversy about the PR, other than the "Sided" name :)
Bors r=grovesNL

PR #2948, opened 2019-08-10
Remove typed command wrappers

### LINKED ISSUE #2862: Consider removing typed command buffers, pools and queues
Anything on top of "raw" command types is not used by any backends, it's only for the user, and it doesn't help most important users either (rendy and wgpu-rs can't use those). So we should probably just remove those layers in favor of something like https://crates.io/crates/rendy-command
cc @omni-viral

kvark:
Here comes another midnight PR. If you ever looked for a way to make users stick to the last published version (hal-0.3 in our case), look no more. This PR removes a layer from HAL and reorders the exported types so that everyone is guaranteed to have fun fixing their programs. Seriously though, the old way of type re-exports was very inconsistent and messy, both on the HAL and using sides.

To clarify, I think having `Capability`-based wrappers is very useful, it's a great thing to have. Problems are: 1) it's not needed by the backends, and 2) it's not *universally* useful, i.e. things like Rendy, WebRender, and WebGPU can't really take full advantage of it. Therefore, our recommendation is to consider [rendy-command](https://crates.io/crates/rendy-command) as a replacement. cc @omni-viral 

With the removal of typed layers, our names became more solid, and the semantics of clear values is much simpler now.

Fixes #2862
PR checklist:
- [x] tested examples with the following backends: all

@grovesNL @msiglreith this PR should probably not hang around for too long, since it's hard to rebase. Only HAL parts needs reviewing, everything else is not as important.


omni-viral:
Just look at LOC stats!

examples/quad/main.rs
```diff
@@ -294,10 +307,13 @@ impl<B> Renderer<B> where B: hal::Backend {
         let buffer_len = QUAD.len() as u64 * buffer_stride;
 
         assert_ne!(buffer_len, 0);
-        let mut vertex_buffer =
-            ManuallyDrop::new(unsafe { device.create_buffer(buffer_len, buffer::Usage::VERTEX) }.unwrap());
+        let mut vertex_buffer =ManuallyDrop::new(unsafe {
```
grovesNL: Did cargo fmt skip this?

kvark:
@grovesNL I pushed a commit with rustfmt that added about 100 LOC...
Thanks for the reviews!
bors r=msiglreith,grovesNL

PR #2955, opened 2019-08-15
gl: Check for debug extension support

grovesNL:
This fixes the quad example on native GL.

We were unwrapping an invalid enum error when getting the `VENDOR` string, but the invalid enum was actually caused by trying to unconditionally use a constant from the KHR_debug extension.

PR checklist:
- [X] tested examples with the following backends: gl


kvark:
I removed the glutin bumps from here (which is done by #2961)
bors r+

PR #2960, opened 2019-08-16
Benchmarking support in Warden

### LINKED ISSUE #2717: Benchmarking suite
It's not very reliable to optimize a backend only when the real world usage is involved and tested upon. We may get it faster for one case and slower for the others. It's always difficult to reason about the effect a small change can bear on the performance across different applications. I believe we can solve this by introducing a few simple benchmarks here, here is some ideas about how to do it.

We can have another binary target of Warden that shares the scene definition with reftests but has a different job description. A benchmark job should be allowed to repeat something multiple times. We'd then have a number of tasks for benchmarking different areas of the API:
  - vertex buffer switches: try re-binding the same point out of many, always different point, many points at once
  - pipeline swicthes: just alternate between a few completely different pipelines a lot, also would be interesting to see how alternating between *similar* pipelines (i.e. same shader but with specialization) affects the perf
  - render pass switches: create a large texture array and clear each slice by doing a separate render pass into it with a clear op, for example
  - descriptor set switches: in a pipeline layout that occupies 4 descriptor sets, try changing the top one a lot, then changing the second from the top, then the third, then the bottom one, and see how that affects perf

kvark:
Fixes #2717
PR checklist:
- [x] `make reftests` succeeds (d'oh!)


kvark:
The main incentive for this work was benchmarking blits versus shader draw calls on Vulkan (since it's not supported elsewhere). Here are the results with "GeForce GTX 1050 Ti with Max-Q Design" on Linux:
  - _copy_: 1422.8 mcs mean, 1083 variance
  - _blit_: 1156.2 mcs mean, 860 variance
  - _shader_: 1170.6 mcs mean, 458 variance

Testing had 10 runs (by just calling `make benches`). Source texture is 1024x1024 of type RGBA8. Destination texture for shader and blits is 2048x2048 of type RG8, and same resolution with type RGBA8 for the copy. Each run does 4 repetitions of a blit (using the same method methods), totaling to about 16 megatexels of RGBA8 data.

The procedure isn't ideal, but results appear to be valid. Known imperfections:
  - all images were returned to the `ColorAttachmentOptimal` state between 4 repetitions in a single run. Switching that "stable" state to `TransferDstOptimal` ends up penalizing the _shader_ path heavily, but the _copy_ and _blit_ performance stays roughly the same.
  - blits and copies were done in 4 chunks each, operating on 1024x1024 texel rects.

All in all, there doesn't appear to be strong benefit for blitting on this GPU, other than for convenience.

kvark:
bors r=nobody

kvark:
On Intel GPU, I'm seeing the numbers in 20 mcs - 30 mcs ballpark, which likely means no actual work was done. Gotta figure out what makes it skip the work.
Another annoyance is that the timing queries are now used in the reftests, and they aren't implemented in many backends. This is easy to work around though.

krogovin:
It would be a really good idea to run this with the following tweaks:

1. Many, many blits per frame
2. To make sure a driver is not "clever" (and skips the work), blit the 1x1 top of the pyramid to the framebuffer.
3. Use much bigger texture, something like 4096x4096
4. use many different textures (with different data), not just one per frame. However do NOT reload the texture every frame
5. add a use case to first render to texture and generate mips for that. In addition, to make sure the driver has to do work on the 3D side, render to each texture a few pixels to mess it up,
6. last but not least: run on mobile devices that use tiled based rendering: PowerVR, Mali and Qualcomm being the bread and butter.
7. On GL, compare against glGenerateMipmaps

The (1) is to allow that doing a simple naive timing of the thing from start to finish with say, 1000 frames and 1000 blits per frame will give more reliable values by the volume. (2) can be important because I have seen drivers skip work if the result of the work is unused. (3) becomes more and more important for tiled based renderers. (4) and (5) are me begin paranoid because I know drivers cheat. (6) is critical since WebGPU's main use case is going to be mobile I suspect.

PR #2967, opened 2019-08-20
Replace `winit` with `raw-window-handle` in backend crates

### LINKED ISSUE #2956: Please add support for the new `raw-window-handle` crate
Recently the gamedev-wg has been trying to make it easier for windowing libs and graphical libs to agree on a protocol for communication. This has resulted in the [raw-window-handle](https://crates.io/crates/raw-window-handle) crate, where the window offers up its OS window handle thingy to the graphical lib upon request and then the graphical lib is able to do its startup.

* This would affect each backend.
* It's a strict superset of the `winit` feature, so this can be released as a `0.3.1`
* Instead of having the `winit` feature allow for the `pub fn create_surface(&self, window: &winit::window::Window) -> Surface` method, you simply _unconditionally_ have a `pub fn create_surface(&self, has_handle: &impl raw_window_handle::HasRawWindowHandle) -> Surface` method. Then you call for the raw window handle, [match on it](https://docs.rs/raw-window-handle/0.1.2/raw_window_handle/enum.RawWindowHandle.html), and then re-dispatch to the correct actual method just like before.
* After this the `winit` feature can eventually be removed in 0.4 **EDIT:** _only_ if the `raw-window-handle` crate ends up supporting the web platform by then (which is up in the air at this time).

Tracking issue where this crate was discussed and crated if you have any questions: https://github.com/rust-gamedev/wg/issues/26

jchapman127:
Closes #2956
PR checklist:
- [ ] tested examples with the following backends: dx11, dx12, vulkan, wgl

Currently Android is not supported in `raw-window-handle` and therefore unimplemented in `gfx-backend-vulkan` - waiting on https://github.com/rust-windowing/raw-window-handle/issues/12 to be resolved.


src/backend/dx11/src/lib.rs
```diff
@@ -142,10 +140,15 @@ impl Instance {
...
-    pub fn create_surface(&self, window: &winit::window::Window) -> Surface {
-        use winit::platform::windows::WindowExtWindows;
-        self.create_surface_from_hwnd(window.hwnd() as *mut _)
+    pub fn create_surface(
+        &self,
+        has_handle: &impl raw_window_handle::HasRawWindowHandle,
+    ) -> Surface {
+        let hwnd = match has_handle.raw_window_handle() {
+            raw_window_handle::RawWindowHandle::Windows(handle) => handle.hwnd,
+            _ => unreachable!(),
```
kvark: I think cases like this should be `panic!`, because they will easily be reachable once the enum gets new variants

src/backend/vulkan/src/window.rs
```diff
@@ -310,54 +310,61 @@ impl Instance {
...
         ))]
         {
-            use winit::platform::unix::WindowExtUnix;
-
-            if self.extensions.contains(&khr::WaylandSurface::name()) {
-                if let Some(display) = window.wayland_display() {
-                    let display: *mut c_void = display as *mut _;
-                    let surface: *mut c_void = window.wayland_surface().unwrap() as *mut _;
-                    return self.create_surface_from_wayland(display, surface);
+            return match has_handle.raw_window_handle() {
```
kvark: could we refactor this into a single giant `match has_handle.raw_window_handle()`, where some of the arms are gated by `cfg` and some have conditions like `if self.extensions.contains(...)`? I think it would look much better here

src/backend/dx11/src/lib.rs
jchapman127: Yes of course, I hadn't considered that. Do you have any suggestions for the errors message(s)?

src/backend/vulkan/src/window.rs
jchapman127: Agreed, will do.

src/backend/gl/src/window/wgl.rs
```diff
@@ -186,12 +183,16 @@ impl Instance {
         }
     }
 
-    #[cfg(feature = "winit")]
-    pub fn create_surface(&self, window: &winit::window::Window) -> Surface {
-        use winit::platform::windows::WindowExtWindows;
-
-        let hwnd = window.hwnd();
-        self.create_surface_from_hwnd(hwnd as *mut _)
+    #[cfg(windows)]
```
jchapman127: Not quite sure about this, or if the entire wgl feature should be cfg(target_os = windows), or whether the unreachable should instead be something like panic!("WGL is not supported for this target")?

kvark: latter sounds good to me. @msiglreith ?

src/backend/dx11/src/lib.rs
jchapman127: Yes of course, I hadn't considered that. Do you have any suggestions for the error message(s)?

src/backend/vulkan/src/window.rs
jchapman127: Agreed, will do.

src/backend/dx11/src/lib.rs
kvark: It actually sounds like we need an error code returned for that mismatch?..

jchapman127:
Wrt. how to land this change, I see now that it would be a breaking change as you mention. If you wanted to go with option 1, would that be achieved by adding `create_surface(&impl HasRawWindowHandle)` *alongside* `create_surface(&winit::Window)`, both `cfg`'d out as appropriate and mutually exclusive?

kvark:
> Wrt. how to land this change, I see now that it would be a breaking change as you mention. If you wanted to go with option 1, would that be achieved by adding `create_surface(&impl HasRawWindowHandle)` _alongside_ `create_surface(&winit::Window)`, both `cfg`'d out as appropriate and mutually exclusive?

I think we can make them exclusive, but only if `winit` takes precedence (in case of both features enabled), so that the existing code works.

Although, a simpler (and better) path would just be to just expose it under a different name. We'll break the API in the next version anyway.

jchapman127:
> Although, a simpler (and better) path would just be to just expose it under a different name. We'll break the API in the next version anyway.

Ok, in that case I'll go ahead and do that, maybe with the name `create_surface_from_raw` for now? (Obviously feel free to bikeshed.)

kvark:
> > Although, a simpler (and better) path would just be to just expose it under a different name. We'll break the API in the next version anyway.
> 
> Ok, in that case I'll go ahead and do that, maybe with the name `create_surface_from_raw` for now? (Obviously feel free to bikeshed.)

Sounds great, thank you!

(to clarify, this would need to be done in a separate PR targeting hal-0.3 branch and bumping the patch versions)

src/backend/dx11/src/lib.rs
jchapman127: > It actually sounds like we need an error code returned for that mismatch?..

Could you explain that a bit further, please?

kvark: Just adding something like `struct InitError` in `window.rs` module, so that all `create_surface_from_raw` calls would return `Result<Surface, window::InitError>` instead of panicking.

jchapman127: I thought that was what you were implying, just wanted to double check :)

jchapman127:
I've now opened #2968 with these changes as additive against the `hal-0.3` branch. Will address your remaining feedback there.

kvark:
@jchapman127 would you mind landing this PR though? We still want this very change in `master`.

jchapman127:
> @jchapman127 would you mind landing this PR though? We still want this very change in `master`.

Does that include changing `create_surface` to return `Result<InitError>`?

kvark:
> > @jchapman127 would you mind landing this PR though? We still want this very change in `master`.
> 
> Does that include changing `create_surface` to return `Result<InitError>`?

That would be great, yeah! Especially, since you've done this for hal-0.3 already, so it shouldn't be difficult to port back here.

PR #2968, opened 2019-08-20
[hal-0.3] Add support for `raw-window-handle`

### LINKED ISSUE #2956: Please add support for the new `raw-window-handle` crate

jchapman127:
Closes #2956.
PR checklist:

Non-breaking change version of #2967.

Currently Android is not supported in raw-window-handle and therefore unimplemented in gfx-backend-vulkan - waiting on https://github.com/rust-windowing/raw-window-handle/issues/12 to be resolved.

### DIFF
--- a/src/backend/dx11/Cargo.toml
+++ b/src/backend/dx11/Cargo.toml
@@ -29,7 +29,7 @@ parking_lot = "0.7"
 winapi = { version = "0.3", features = ["basetsd","d3d11", "d3d11sdklayers", "d3dcommon","d3dcompiler","dxgi1_2","dxgi1_3","dxgi1_4", "dxgi1_5", "dxgiformat","dxgitype","handleapi","minwindef","synchapi","unknwnbase","winbase","windef","winerror","winnt","winuser"] }
 winit = { version = "0.19", optional = true }
 wio = "0.2"
-raw-window-handle = { version = "0.1.2", optional = true }
+raw-window-handle = "0.1.2"
 
 # This forces docs.rs to build the crate on windows, otherwise the build fails
 # and we get no docs at all.
--- a/src/backend/dx11/Cargo.toml
+++ b/src/backend/dx11/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-backend-dx11"
-version = "0.3.0"
+version = "0.3.1"
 description = "DirectX-11 API backend for gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
@@ -18,7 +18,7 @@ default = []
 name = "gfx_backend_dx11"
 
 [dependencies]
-gfx-hal = { path = "../../hal", version = "0.3" }
+gfx-hal = { path = "../../hal", version = "0.3.1" }
 range-alloc = { path = "../../auxil/range-alloc", version = "0.1" }
 bitflags = "1"
 derivative = "1"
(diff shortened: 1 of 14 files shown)


jchapman127:
Wasn't 100% on your normal procedure for both of those, but will address
them when back at my desk.

kvark:
Wonderful, thank you!
bors r+

btw, is there a reason for "0.1.2" version specifically?

jchapman127:
No, thank you for a very pleasant first-time-contributor experience!

> btw, is there a reason for "0.1.2" version specifically?

Not really, it's probably because I used `cargo add` 🤷‍♂ That should allow cargo to select any `0.1.x` (where `x > 2`) version though, right?

kvark:
Could you please squash the macos fix with the commit that adds support to macos? We'd want every commit to be buildable.

kvark:
> Not really, it's probably because I used cargo add man_shrugging That should allow cargo to select any 0.1.x (where x > 2) version though, right?

Right. The approach I'm trying to follow is having the higher version specified only if it's required by something, instead of just always pointing to the highest. For example, if there is a new type we depend on, a new enum variant, or a `Debug` implementation for something. Or maybe there was a critical fix that we want to keep.

Briefly looking at [changelog](https://github.com/rust-windowing/raw-window-handle/blob/master/CHANGELOG.md) I'm getting an impression that we don't require anything past "0.1".

jchapman127:
>
>
> Could you please squash the macos fix with the commit that adds support to macos? We'd want every commit to be buildable.

Yes of course, sorry. I've done that now.

>
>
> > Not really, it's probably because I used cargo add man_shrugging That should allow cargo to select any 0.1.x (where x > 2) version though, right?
> 
> Right. The approach I'm trying to follow is having the higher version specified only if it's required by something, instead of just always pointing to the highest. For example, if there is a new type we depend on, a new enum variant, or a `Debug` implementation for something. Or maybe there was a critical fix that we want to keep.
> 
> Briefly looking at [changelog](https://github.com/rust-windowing/raw-window-handle/blob/master/CHANGELOG.md) I'm getting an impression that we don't require anything past "0.1".

Yes that makes complete sense, I have changed the version to `0.1`.

kvark:
Sweet, thank you!
bors r+

kvark:
@jchapman127 could you address the CI issues?

jchapman127:
> @jchapman127 could you address the CI issues?

Well, I checked out the branch on Linux and ran `make` which revealed a compile error, but it looks like the Linux CI builds still failed. Do you know what the failures mean?

kvark:
Let's see if it succeeds now with 417593cf060499254a7ac7785c451bd4a657fb11
(Also note the metal version needs a bigger bump, I'll do it later)
bors r+

PR #2969, opened 2019-08-22
Cleanup dependencies

### LINKED ISSUE #2436: Unify error types
Currently, there are multiple error types, some using the failure crate, and some implementing the Error trait. I propose that all types implement Error and to remove the failure crate. or make every error type the failure crate type. This may be a rookie question, but if it is already mostly using the Error trait, why keep the failure dependency and not use the standard library way?

grovesNL:
This cleans up our dependencies slightly through version matching or dependency removal.

Fixes #2436

- Removes `failure`
  - `failure`'s README describes that it's in a bit of flux as improvements are being made to Rust's stdlib upstream, so we might want to avoid relying on it too greatly in its current state
  - we aren't really using `failure` consistently throughout, and many of the messages are just taken directly from doc comments anyway. While we could make usage more consistent, it's not clear that this effort will be very useful if it's incompatible with upstream changes in Rust itself.
- Matches spirv_cross version across all backends
- Removes `derivative`
  - most existing usages seem like they could be treated as opaque types because they don't provide much debug information, so we could just write custom debug implementations wherever it's noticeably useful

PR checklist:

### DIFF
--- a/src/backend/dx11/Cargo.toml
+++ b/src/backend/dx11/Cargo.toml
@@ -24,7 +24,7 @@ bitflags = "1"
 derivative = "1"
 log = { version = "0.4" }
 smallvec = "0.6"
-spirv_cross = { version = "0.14.0", features = ["hlsl"] }
+spirv_cross = { version = "0.15", features = ["hlsl"] }
 parking_lot = "0.7"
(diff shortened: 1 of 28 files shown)


grovesNL:
bors r=kvark

grovesNL:
bors r=kvark

grovesNL:
bors r=kvark

grovesNL:
Looks like the GitHub outage caused some issues staging the latest set of commits
bors r=kvark

grovesNL:
bors r=kvark

PR #2970, opened 2019-08-22
Unbloat the build

kvark:
Removes dependency on failure and refactors our errors a bit.
This change reduces the gfx-hal build time from 45 sec to just 8 sec on my machine :tada: 

PR checklist:
- [x] tested examples with the following backends: linux


kvark:
bors r=grovesNL

kvark:
bors r=grovesNL

kvark:
bors r=grovesNL

matklad:
do you know which part of the speed up is attributed to the removal of failure, and which to the other refactorings? I wonder if “removing failure may make your build five times faster” is an accurate thing to say :)

kvark:
Iirc, it was failure precisely that was making us slow.

PR #2972, opened 2019-08-22
Metal: update metal-rs and wrap RP switch into an ARP

kvark:
Fixes the leak of render command encoders
PR checklist:


kvark:
bors r=aleksijuvani

seivan:
This fixes the leaks with discussed, thanks!

PR #2986, opened 2019-08-30
[0.3-mtl] make resize fill configurable

### LINKED ISSUE #2951: Metal swapchain resize triggers a validation error
Metal swapchain resize attempts to blit the contents of the last frame into a temporary frame that is has to issue in order to properly free the drawables... The blit (that we implement ourselves) requires texture usage to be shader-readable, which may not be the case, thus triggering a validation error.

This is annoying as it's effectively blocking a GPU capture of apps that run with validation enabled from XCode.

I suggest we don't do a blit if we know that the swapchain doesn't support being shader-readable.

kvark:
Fixes #2951 
PR checklist:

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-backend-metal"
-version = "0.3.1"
+version = "0.3.2"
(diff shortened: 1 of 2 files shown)


kvark:
Also sets up the properties of a swapchain to be inherited.

PR #2994, opened 2019-09-05
[vk] fix `wait_semaphore` not being used in `present_surface`

aleksijuvani:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -1213,27 +1213,31 @@ impl queue::CommandQueue<Backend> for CommandQueue {
         wait_semaphore: Option<&native::Semaphore>,
     ) -> Result<Option<Suboptimal>, PresentError> {
         let ssc = surface.swapchain.as_ref().unwrap();
-        let submit_info = vk::SubmitInfo {
-            s_type: vk::StructureType::SUBMIT_INFO,
-            p_next: ptr::null(),
-            wait_semaphore_count: 0,
-            p_wait_semaphores: wait_semaphore.map_or(ptr::null(), |s| &s.0 as *const _),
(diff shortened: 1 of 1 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -1216,7 +1216,7 @@ impl queue::CommandQueue<Backend> for CommandQueue {
         let submit_info = vk::SubmitInfo {
```
aleksijuvani: Why is this queue submission necessary? Could we pass `wait_semaphore` to `vk::PresentInfoKHR` directly?

kvark: If there is a wait semaphore, we should skip the submission. If there isn't, we need *a* semaphore to pass to the present(), so this is what the submission is used for.
Would you mind trying this change out?

aleksijuvani: I see. In the case there is no wait semaphore, we must wait for the `COLOR_ATTACHMENT_OUTPUT` pipeline stage to complete, and we cannot use pipeline barriers to do so, because pipeline barriers exist in the context of a command buffer—and this is why we require a semaphore, because we must use a submission to wait for the pipeline stage?

kvark: The idea is that an image of the new swapchain model is only going to be used in a render pass. User is still responsible to set up the initial/final layouts so that the image is presentable after it's rendered to. I.e. the quad example has `i::Layout::Undefined .. i::Layout::Present`.

We need a semaphore to synchronize rendering with presentation because Vulkan spec requires that. It allows the driver to treat presentation mechanism as an internal "queue", so it needs to be synchronized with the rendering queue.

Issue #1508, opened 2017-09-20
Expose debugging features in an implementation-agnostic manner?


AlphaModder:
The backends targeted by `gfx` all expose a number of debugging features to allow developers to more easily find and solve problems in their code. With the transition to a more low-level API, it may be appropriate to expose at least some of these features to developers in an implementation-agnostic manner where possible.

This topic has been brought up in a few issues before (#596, #1288, #1426), but these are old and did not have a full comparison between APIs, so in this one I have compiled a table of the available debugging features on each backend:

| debug feature | dx11 | dx12 | vulkan | metal | gl |
| ------------- | ---- | ---- | ------ | ----- | -- |
| Catching normally unchecked API misuse | **Yes**, via the debug layer enabled through [`D3D11_CREATE_DEVICE_FLAG::D3D11_CREATE_DEVICE_DEBUG`][1a]. Debug messages are accessible through [`ID3D11InfoQueue`][2a]. | **Yes**, via the debug layer enabled through a call to [`ID3D12Debug::EnableDebugLayer`][3a]. Debug messages are accessible through [`ID3D12InfoQueue`][4a]. | **Yes**, via the [validation layer mechanism][5a]. Debug messages are accessible to callbacks passed to `vkCreateDebugReportCallbackEXT`, exposed by the [`VK_EXT_debug_report`][6a] extension. | Set `METAL_DEVICE_WRAPPER_TYPE=1` environmental variable, see [This article][7a] on "Metal API Validation". | **Yes**, via the [`GL_KHR_debug`][8a] or [`GL_ARB_debug_output`][9a] extensions (identical interface). Enabled by setting the `GL_CONTEXT_FLAG_DEBUG_BIT` during context creation or calling `glEnable(GL_DEBUG_OUTPUT)`. Like Vulkan, messages are accessible through a callback mechanism. |
[...]

PR #3000, opened 2019-09-06
[mtl][vk] add object debug labels

aleksijuvani:
Resolves a part of #1508
PR checklist:
- [x] tested examples with the following backends: mtl vk

### DIFF
--- a/src/backend/metal/src/native.rs
+++ b/src/backend/metal/src/native.rs
@@ -96,6 +96,7 @@ pub struct Subpass {
 pub struct RenderPass {
     pub(crate) attachments: Vec<Attachment>,
     pub(crate) subpasses: Vec<Subpass>,
+    pub(crate) name: String,
 }
 
 #[derive(Debug)]
@@ -294,6 +295,7 @@ pub enum ImageLike {
         descriptor: metal::TextureDescriptor,
         mip_sizes: Vec<buffer::Offset>,
         host_visible: bool,
+        name: String,
     },
     /// This is a linearly tiled HOST-visible image, which is represented by a buffer.
     Buffer(Buffer),
@@ -413,6 +415,7 @@ pub enum Buffer {
     Unbound {
         size: u64,
         usage: buffer::Usage,
+        name: String,
     },
     Bound {
         raw: metal::Buffer,
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -2,6 +2,7 @@ use arrayvec::ArrayVec;
 use ash::extensions::khr;
 use ash::version::DeviceV1_0;
 use ash::vk;
+use ash::vk::Handle;
 use smallvec::SmallVec;
 
 use hal::{
@@ -22,7 +23,7 @@ use std::sync::Arc;
 use std::{mem, ptr};
 
 use crate::pool::RawCommandPool;
-use crate::{conv, native as n, window as w};
+use crate::{conv, native as n, window as w, command as cmd};
 use crate::{Backend as B, Device};
 
 #[derive(Debug, Default)]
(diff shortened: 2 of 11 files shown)


src/backend/metal/src/native.rs
```diff
@@ -294,6 +295,7 @@ pub enum ImageLike {
         descriptor: metal::TextureDescriptor,
         mip_sizes: Vec<buffer::Offset>,
         host_visible: bool,
+        name: RefCell<String>,
```
kvark: How can we use `RefCell` here if it's not meant to be used by multiple threads? I think we have to use `Mutex` instead.

src/backend/vulkan/src/device.rs
```diff
@@ -2234,6 +2235,82 @@ impl d::Device<B> for Device {
...
+        self.set_object_name(vk::ObjectType::RENDER_PASS, render_pass.raw.as_raw(), name)
+    }
+}
+
+impl Device {
+    unsafe fn set_object_name(&self, object_type: vk::ObjectType, object_handle: u64, name: &str)
+    -> Result<(), d::OutOfMemory> {
+        let instance = &self.raw.2;
+        if let Some(ref debug_utils_ext) = instance.1 {
+            let name = CString::new(name).unwrap();
```
kvark: let's leave a TODO here to avoid heap allocating this string

src/hal/src/device.rs
```diff
@@ -758,4 +758,42 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
     ///
     /// Host access to all queues needs to be **externally** sycnhronized!
     fn wait_idle(&self) -> Result<(), OutOfMemory>;
+
+    /// Associate a name with an image, for easier debugging in external tools or with validation
+    /// layers that can print a friendly name when referring to objects in error messages
+    unsafe fn set_image_name(&self, image: &B::Image, name: &str) -> Result<(), OutOfMemory>;
```
kvark: I think we should just ignore the errors here. They aren't useful, they just make a lot of noise.

aleksijuvani:
Should be noted that this depends on gfx-rs/metal-rs#98, so we will need to bump the dependency. It should also be noted that this doesn't cover all the types that it's possible to set debug labels for in Vulkan—I figured we can fill them in as we run into them, because we will also have to think about if they are applicable to other backends on a per-case basis.

src/hal/src/device.rs
```diff
@@ -758,4 +758,42 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    /// Associate a name with a buffer, for easier debugging in external tools or with validation
+    /// layers that can print a friendly name when referring to objects in error messages
+    unsafe fn set_buffer_name(&self, buffer: &B::Buffer, name: &str) -> Result<(), OutOfMemory>;
```
kvark: it looks like Vulkan [requires external synchronization](https://vulkan.lunarg.com/doc/view/1.0.30.0/linux/vkspec.chunked/ch32s01.html#vkDebugMarkerSetObjectNameEXT) for all the objects that we need to set names for:
> Host access to pNameInfo.object must be externally synchronized 

That means we should have `&mut B::Buffer` (and same for other methods). On the good side, the implementations don't need to have `Mutex` in them.

src/backend/vulkan/src/device.rs
aleksijuvani: Vulkan requires a null-terminated string here, but Rust strings are not null-terminated so I'm not sure how to avoid this allocation

kvark: The way to do this would be to have some kind of temporary storage associated with `Device`, which we'd use to copy the string in, append a null terminator, and provide to Vulkan then.

src/hal/src/device.rs
aleksijuvani: I'd be inclined to follow the Vulkan specification and keep the errors here, just to be consistent with the rest of the API, even if most of the time the user would not care.

aleksijuvani:
metal-rs 0.17 added `MTLLanguageVersion::V2_2` so I've opened grovesNL/spirv_cross#113 to add `msl::Version::V2_2`—needs a new version of spirv_cross published

aleksijuvani:
bors try

src/hal/src/device.rs
kvark: We follow Vulkan for the lack of a better idea most of the time. Sometimes though, we have a better idea, and we disagree. See https://github.com/gfx-rs/gfx/wiki/Deviations-from-Vulkan as relevant.

In this particular case, labels are most likely in the host memory. Running out of host memory is something that can easily happen in 99% of the gfx-rs code, given our use of the standard library and containers, and we aren't handling that.

So for now, let's not worry about these error codes. We'll cross that bridge when it will need to be crossed.

aleksijuvani: I've removed the error codes.

PR #3006, opened 2019-09-09
[mtl] add missing limits

aleksijuvani:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -74,6 +74,7 @@ use std::{cmp, iter, mem, ptr, thread, time};
 const PUSH_CONSTANTS_DESC_SET: u32 = !0;
 const PUSH_CONSTANTS_DESC_BINDING: u32 = 0;
 const STRIDE_GRANULARITY: pso::ElemStride = 4; //TODO: work around?
+const SHADER_STAGE_COUNT: usize = 3;
 
 /// Emit error during shader module parsing.
 fn gen_parse_error(err: SpirvErrorCode) -> ShaderError {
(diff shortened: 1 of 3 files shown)


src/backend/metal/src/device.rs
```diff
@@ -449,8 +449,13 @@ impl hal::PhysicalDevice<Backend> for PhysicalDevice {
             max_push_constants_size: 0x1000,
             max_sampler_allocation_count: !0,
             max_bound_descriptor_sets: MAX_BOUND_DESCRIPTOR_SETS as _,
-            max_descriptor_set_uniform_buffers: 0x100, // also arbitrary, but probably shouldn't be more than max_bound_descriptor_sets
-            max_fragment_input_components: 32, // TODO: Add to private caps since 32 is only for mac
+            max_descriptor_set_samplers: pc.max_samplers_per_stage as usize * 4,
```
kvark: Where does "* 4" come from?

aleksijuvani: This is the same logic as in MoltenVK, which seemed to make sense to me—if I'm not mistaken, there are four programmable shader stages, and if each supports a maximum of `max_samplers_per_stage`, the total ought to be `max_samplers_per_stage * 4`. Actually, looking at the [Metal Tessellation Pipeline](https://developer.apple.com/library/archive/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/Tessellation/Tessellation.html), maybe this ought to be three instead?

kvark: Ok, I see. This makes sense.
Let's have a constant somewhere, i.e. `SHADER_STAGE_COUNT` to use for multiplication here. Its value should be `3`.

aleksijuvani: Done!

PR #3008, opened 2019-09-11
Deduplicate spec const code shared between backends

aleksijuvani:
Fixes #issue
PR checklist:
- [x] tested examples with the following backends: mtl dx11 dx12 gl

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 
 members = [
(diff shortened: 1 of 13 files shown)


aleksijuvani:
I've moved this functionality to a separate crate. How do you feel about the name? I like gfx-backend-common better than gfx-backend-helper, because it implies that it's shared functionality instead of just a collection of utilities.

Issue #3009, opened 2019-09-11
D3D12 debug layer errors with barriers created in create_render_pass


zakorgy:
Short info header:
- GFX version: gfx.backend-dx12:0.3.2
- OS: Windows 10
- GPU: Intel HD Graphics 630

I have the following validation layer error from the debug layer:
```
ID3D12CommandList::ResourceBarrier: Before state (0x0: D3D12_RESOURCE_STATE_[COMMON|PRESENT]) of resource (0x00000228D8A91820:'Unnamed ID3D12Resource Object') (subresource: 0) specified by transition barrier does not match with the state (0x4: D3D12_RESOURCE_STATE_RENDER_TARGET) specified in the previous call to ResourceBarrier [ RESOURCE_MANIPULATION ERROR #527: RESOURCE_BARRIER_BEFORE_AFTER_MISMATCH]
```
which is related to a barrier created during [create_render_pass](https://github.com/gfx-rs/gfx/blob/92df79ae6765ada9ab139eb83c03c3f9049e1417/src/backend/dx12/src/device.rs#L1213). The states [here](https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12/src/device.rs#L1257) and [here](https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12/src/device.rs#L1392) are always going to be `D3D12_RESOURCE_STATE_[COMMON|PRESENT]`, since [map_image_resource_state](https://github.com/gfx-rs/gfx/blob/master/src/backend/dx12/src/conv.rs#L492) only takes `Layout` in account if it's `Layout::Present` and the `Access` provided here at this point is always `Access::empty`.
I don't know if this is an expected behavior, because if I have the resource in the expected state (in my case `D3D12_RESOURCE_STATE_RENDER_TARGET`) for the first subpass and no transition needed (I want to use it as `D3D12_RESOURCE_STATE_RENDER_TARGET` inside the render pass) the barrier is still generated whit the wrong Before/After state.
I've put together a quick workaround on my fork (which works for my use case) using the `Access` values from the provided `dependencies` to get the correct Before/After states, and this fixed my validation layer errors: https://github.com/zakorgy/gfx/commit/9ac0070a923ea313e5d72bac30532a8da9ea350b
Here are the render passes I use:  https://gist.github.com/zakorgy/2c14cdf7b8cff2d0c9189649578e25a2
If it's needed I can check which one is exactly involved.

zakorgy:
[...]

PR #3010, opened 2019-09-12
[0.3/dx12] use explicit external pass barrier access

kvark:
Improves (but doesn't fix yet) #3009 
PR checklist:

The basic thing it does is just using the access flags provided in the explicit external barriers, if any, instead of `empty()`.
@zakorgy - please check it out and tell if it's helping or not

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-dx12-0.3.3 (12-09-2019)
+  - improve external render pass barriers
+
 ### backend-metal-0.3.3 (05-09-2019)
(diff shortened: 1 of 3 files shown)


zakorgy:
@kvark unfortutnatelly we still have errors, and there are new ones: https://gist.github.com/zakorgy/89fc1d62e9cf61c17d329c9b0d5d06d8
Maybe D3D12 doesn't like the fact that the src stage is `D3D12_RESOURCE_STATE_DEPTH_WRITE |  D3D12_RESOURCE_STATE_RENDER_TARGET` with this solution in my case?

kvark:
uh oh. I should have waited for your test before publishing.

PR #3011, opened 2019-09-12
[dx12] use explicit external pass barrier access

kvark:
Sibling of #3010 
~~Also includes a tiny fix for `num_levels` being used. @msiglreith - could you peek at that please?~~


kvark:
Thanks!
Maybe it's good that this hasn't landed - I should move out the `num_levels` change into a separate commit, and also fix the barriers better...

kvark:
bors r=msiglreith

kvark:
was fighting sublime to keep Unix-style line endings...
bors r=msiglreith

PR #3013, opened 2019-09-13
[0.3/dx12] split explicit external pass barrier access for depth attachments

### LINKED ISSUE #3009: D3D12 debug layer errors with barriers created in create_render_pass

zakorgy:
Follow up for #3010

Fixes #3009
PR checklist:

Handle the case for depth access provided in the explicit external barriers separately.
Also set the `end` in case of `Range { start: Sr::External, end: Sr::Pass(_) } ` since there is no guarantee that there is always a `Range { start: Sr::Pass(_), end: Sr::External }` for transitioning back the resources, at least Vulkan doesn't need it.
cc @kvark

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -1140,6 +1140,7 @@ impl d::Device<B> for Device {
             .map(|_| 0u16)
             .collect::<SmallVec<[_; 1]>>();
         let mut external_access = image::Access::empty() .. image::Access::empty();
+        let mut external_access_depth = image::Access::empty() .. image::Access::empty();
         for dep in &dependencies {
             use hal::pass::SubpassRef as Sr;
             let dep = dep.borrow();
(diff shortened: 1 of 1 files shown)


msiglreith:
Would you mind expanding on why handling depth stencil attachments differently is required?

kvark:
> Would you mind expanding on why handling depth stencil attachments differently is required?

The access flags specified in an external barrier apply to the whole sub-pass, so it's a union of flags needed for depth and color and resolve etc. When we try to map them to D3D12 states, the conversion assumes the access flags to be required specifically, so it's confused and it's producing the wrong initial state.

kvark:
See #3014 . Please let me know if this works for you!

zakorgy:
Closing in favor of https://github.com/gfx-rs/gfx/pull/3014

PR #3014, opened 2019-09-13
[0.3/dx12] rework resource state conversion, better external barriers

kvark:
Follow-up to #3010 that tries to address #3009 better. Replaces #3013
Edit: this combines parts 2 and 3 of the big dx12 barrier refactor :)

PR checklist:

Contains 2 ideas:
  1. we try to extract more information out of the image layouts, which is more robust
  2. we are very selective in the external access dependencies, only taking ones that are related to where the attachment is used first or last.

### DIFF
--- a/src/backend/dx12/src/conv.rs
+++ b/src/backend/dx12/src/conv.rs
@@ -486,54 +486,47 @@ pub fn map_image_resource_state(
     layout: image::Layout,
 ) -> D3D12_RESOURCE_STATES {
     use self::image::Access;
-    // `D3D12_RESOURCE_STATE_PRESENT` is the same as COMMON (general state)
-    if layout == image::Layout::Present {
-        return D3D12_RESOURCE_STATE_PRESENT;
-    }
 
-    // Mutable states
-    if access.contains(Access::SHADER_WRITE) {
-        return D3D12_RESOURCE_STATE_UNORDERED_ACCESS;
-    }
-    if access.contains(Access::DEPTH_STENCIL_ATTACHMENT_WRITE) {
-        return D3D12_RESOURCE_STATE_DEPTH_WRITE;
-    }
-    if access.contains(Access::COLOR_ATTACHMENT_READ)
-        || access.contains(Access::COLOR_ATTACHMENT_WRITE)
-    {
-        return D3D12_RESOURCE_STATE_RENDER_TARGET;
-    }
+    match layout {
+        // the same as COMMON (general state)
+        image::Layout::Present => D3D12_RESOURCE_STATE_PRESENT,
+        image::Layout::ColorAttachmentOptimal => D3D12_RESOURCE_STATE_RENDER_TARGET,
+        image::Layout::DepthStencilAttachmentOptimal => D3D12_RESOURCE_STATE_DEPTH_WRITE,
+        // `TRANSFER_WRITE` requires special handling as it requires RESOLVE_DEST | COPY_DEST
+        // but only 1 write-only allowed. We do the required translation before the commands.
+        // We currently assume that `COPY_DEST` is more common state than out of renderpass resolves.
+        // Resolve operations need to insert a barrier before and after the command to transition
+        // from and into `COPY_DEST` to have a consistent state for srcAccess.
+        image::Layout::TransferDstOptimal => D3D12_RESOURCE_STATE_COPY_DEST,
+        image::Layout::TransferSrcOptimal => D3D12_RESOURCE_STATE_COPY_SOURCE,
+        // At this point, we assume that the mutable access flags have already been covered by the
+        // layout checks, and we ignore them.
+        image::Layout::General if access.contains(Access::SHADER_WRITE) => D3D12_RESOURCE_STATE_UNORDERED_ACCESS,
+        image::Layout::ShaderReadOnlyOptimal | // could be moved outside
+        image::Layout::DepthStencilReadOnlyOptimal |
+        image::Layout::General |
+        image::Layout::Undefined |
+        image::Layout::Preinitialized => {
+            // Read-only states
+            let mut state = D3D12_RESOURCE_STATE_COMMON;
 
-    // `TRANSFER_WRITE` requires special handling as it requires RESOLVE_DEST | COPY_DEST
-    // but only 1 write-only allowed. We do the required translation before the commands.
-    // We currently assume that `COPY_DEST` is more common state than out of renderpass resolves.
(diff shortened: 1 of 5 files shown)


kvark:
Tested with https://github.com/msiglreith/gfx_ocean/pull/15 and didn't spot any extra validation errors.

src/backend/dx12/src/conv.rs
```diff
@@ -486,54 +486,47 @@ pub fn map_image_resource_state(
...
+        image::Layout::ShaderReadOnlyOptimal | // could be moved outside
+        image::Layout::DepthStencilReadOnlyOptimal |
+        image::Layout::General |
+        image::Layout::Undefined |
+        image::Layout::Preinitialized => {
+            // Read-only states
+            let mut state = D3D12_RESOURCE_STATE_COMMON;
+
+            if access.contains(Access::TRANSFER_READ) {
+                state |= D3D12_RESOURCE_STATE_COPY_SOURCE;
```
msiglreith: as resource states are not flags we should just return them

src/backend/dx12/src/conv.rs
```diff
@@ -486,54 +486,47 @@ pub fn map_image_resource_state(
...
-    if access.contains(Access::TRANSFER_WRITE) {
-        return D3D12_RESOURCE_STATE_COPY_DEST;
-    }
 
-    // Read-only states
-    let mut state = D3D12_RESOURCE_STATE_COMMON;
+    match layout {
+        // the same as COMMON (general state)
+        image::Layout::Present => D3D12_RESOURCE_STATE_PRESENT,
+        image::Layout::ColorAttachmentOptimal => D3D12_RESOURCE_STATE_RENDER_TARGET,
```
msiglreith: good idea with image layouts due to more strict requirements

src/backend/dx12/src/device.rs
```diff
@@ -1131,68 +1140,77 @@ impl d::Device<B> for Device {
...
                 let old = mem::replace(&mut att_infos[id].sub_states[sid], state);
                 debug_assert_eq!(SubState::Undefined, old);
             }
-            for &(id, _layout) in sub.depth_stencil {
-                let state = SubState::New(att_infos[id].target_state);
+            for &(id, layout) in sub.depth_stencil {
+                let state = SubState::New(match layout {
+                    image::Layout::DepthStencilAttachmentOptimal => d3d12::D3D12_RESOURCE_STATE_DEPTH_WRITE,
+                    image::Layout::DepthStencilReadOnlyOptimal => d3d12::D3D12_RESOURCE_STATE_DEPTH_READ,
+                    _ => {
```
msiglreith: General layout should be perfectly legal

src/backend/dx12/src/conv.rs
kvark: no, I think you are mistaken here:
> The resource state usage bits that are used to describe a resource state are divided into read-only and read/write states. The reference topic for the D3D12_RESOURCE_STATES indicates the read/write access level for each bit in the enumeration.
> At most, only one read/write bit can be set for any resource. If a write bit is set, then no read-only bit may be set for that resource. If no write bit is set, then any number of read bits may be set.

kvark:
Thank you for the review!
I addressed the handling of `General` layout now and responded to the other concern.

src/backend/dx12/src/conv.rs
msiglreith: oh right, mixed it up thanks.

kvark:
Bors r=msiglreith

kvark:
let's wait for @zakorgy - I don't want to make this mistake again :)
bors r-

zakorgy:
@kvark this solution solves the render pass related barrier issue-s with dx12, however generates the following errors with `clear_image`:
```
D3D12 ERROR: ID3D12CommandList::ClearRenderTargetView: Resource state (0x400: D3D12_RESOURCE_STATE_COPY_DEST) of resource (0x0000027BE684DC10:'Unnamed ID3D12Resource Object') (subresource: 1) is invalid for use as a render target.  Expected State Bits (all): 0x4: D3D12_RESOURCE_STATE_RENDER_TARGET, Actual State: 0x400: D3D12_RESOURCE_STATE_COPY_DEST, Missing State: 0x4: D3D12_RESOURCE_STATE_RENDER_TARGET. [ EXECUTION ERROR #538: INVALID_SUBRESOURCE_STATE]	
D3D12 ERROR: ID3D12CommandList::ClearDepthStencilView: Resource state (0x400: D3D12_RESOURCE_STATE_COPY_DEST) of resource (0x0000027BE6719450:'Unnamed ID3D12Resource Object') (subresource: 0) is invalid for use as a depth buffer.  Expected State Bits (all): 0x10: D3D12_RESOURCE_STATE_DEPTH_WRITE, Actual State: 0x400: D3D12_RESOURCE_STATE_COPY_DEST, Missing State: 0x10: D3D12_RESOURCE_STATE_DEPTH_WRITE. [ EXECUTION ERROR #538: INVALID_SUBRESOURCE_STATE]	
```
Here is the code how we call clear_image: https://gist.github.com/zakorgy/dec93877d4983b4f2a5d2ed130a5650b, we have `TransferDstOptimal` layout for the clear.

For me this looks like `VkCmdClearImage` requires `TransferDstOptimal` (or `General` according to the spec) layout for the clear, on d3d12 `TransferDstOptimal` maps to `D3D12_RESOURCE_STATE_COPY_DEST` and `ClearRenderTargetView` requires `D3D12_RESOURCE_STATE_RENDER_TARGET`, that's why we get the error.
With `vkCmdClearDepthStencilImage` and `ClearDepthStencilView` the issue is similar.

kvark:
@zakorgy in the last commit, I made it so we are handling `clear_image` correctly by emitting the barriers (back and forth). Please check it out and report back (sorry for a lot of trial and error here - the problem is hard).

@msiglreith please have a look at the last commit

zakorgy:
@kvark I've checked the latest change and the `clear_image` issue is absent now.
There were new errors with DX12, but those turned out to be an issue on our side with the barriers in our `read_pixels` function, fixed them in https://github.com/szeged/webrender/pull/308/commits/4b3d4177065988b42739a3119d414f5a4831a4a5, so everything works fine now with both DX12 and Vulkan.

kvark:
Awesome, thank you for verifying!
Bors r=msiglreith

kvark:
bors retry

kvark:
Published in https://crates.io/crates/gfx-backend-dx12/0.3.4

PR #3025, opened 2019-09-30
[mtl] update `CAMetalLayer` contents scale automatically on macOS

aleksijuvani:
When a `CAMetalLayer` is created by the Metal backend, the contents scale factor is set to match the window scale factor. However, we weren't updating this scale factor if the window scale factor changed later on (such as when moved to a screen with a different scale factor, or when entering exclusive fullscreen mode)!

AppKit updates the contents scale factor automatically for layers that it manages, but for layers created and managed by us, we must implement [`NSViewLayerContentScaleDelegate`](https://developer.apple.com/documentation/appkit/nsviewlayercontentscaledelegate?language=objc) and explicitly opt-in to this behaviour.

PR checklist:

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -40,6 +40,7 @@ smallvec = "0.6"
 spirv_cross = { version = "0.16", features = ["msl"] }
 parking_lot = "0.9"
 storage-map = "0.2"
+lazy_static = "1"
 
 # This forces docs.rs to build the crate on mac, otherwise the build fails
 # and we get no docs at all.
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -72,10 +72,14 @@ use dispatch;
 use foreign_types::ForeignTypeRef;
 use metal::MTLFeatureSet;
 use metal::MTLLanguageVersion;
-use objc::runtime::{Object, BOOL, YES};
+use objc::{
+    declare::ClassDecl,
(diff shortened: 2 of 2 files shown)


aleksijuvani:
When a window's backing scale factor changes, AppKit iterates through every layer and calls [`layer:shouldInheritContentsScale:fromWindow:`](https://developer.apple.com/documentation/appkit/nsviewlayercontentscaledelegate/3005294-layer?language=objc) on the layer's [`delegate`](https://developer.apple.com/documentation/quartzcore/calayer/1410984-delegate?language=objc). We set an instance of this class as the delegate for the layer [here](https://github.com/gfx-rs/gfx/pull/3025/files#diff-6e80d01aaeb75f01d28d8338b0a64797R394).

If [`layer:shouldInheritContentsScale:fromWindow:`](https://developer.apple.com/documentation/appkit/nsviewlayercontentscaledelegate/3005294-layer?language=objc) returns `YES`, AppKit will set the `contentsScale` of the layer to the window's new scale factor. The documentation for [`NSViewLayerContentScaleDelegate`](https://developer.apple.com/documentation/appkit/nsviewlayercontentscaledelegate?language=objc) and [`layer:shouldInheritContentsScale:fromWindow:`](https://developer.apple.com/documentation/appkit/nsviewlayercontentscaledelegate/3005294-layer?language=objc) should further explain the nuances of when this method is called.

kvark:
Bors delegate+

aleksijuvani:
looks like a spurious network error on travis' part
bors r=kvark

PR #3029, opened 2019-10-02
[pre-ll] Remove derivative

alexheretic:
* Remove crate derivative.
* Fix some warnings: `&Error` -> `&dyn Error`.

After reading about master branch speeding up compile times by removing _derivative_ I wondered if it would help the pre-ll code too. Tbf I can see derivative is quite helpful for the `PhantomData` types where regular `#[derive(...)]` falls down. But it isn't _that_ hard to write out the impls.

It impacts clean builds more than I would expect _(linux, rust 1.38, ryzen 2700x)_.

### Before
```sh
$ hyperfine --max-runs 3 'cargo clean; cargo build --example cube' 
Benchmark #1: cargo clean; cargo build --example cube
  Time (mean ± σ):     34.765 s ±  0.126 s    [User: 248.966 s, System: 12.062 s]
  Range (min … max):   34.666 s … 34.906 s    3 runs
```
### After
```sh
$ hyperfine --max-runs 3 'cargo clean; cargo build --example cube'
Benchmark #1: cargo clean; cargo build --example cube
  Time (mean ± σ):     28.158 s ±  0.343 s    [User: 250.888 s, System: 12.067 s]
  Range (min … max):   27.819 s … 28.504 s    3 runs
```

PR checklist:

### DIFF
--- a/src/core/Cargo.toml
+++ b/src/core/Cargo.toml
@@ -30,7 +30,6 @@ path = "src/lib.rs"
 [dependencies]
 bitflags = "1"
 mint = { version = "0.5", optional = true }
-derivative = "1"
 draw_state = "0.8"
 log = "0.4"
 serde = { version = "1", features = ["serde_derive"], optional = true }
--- a/src/core/src/factory.rs
+++ b/src/core/src/factory.rs
@@ -57,7 +57,7 @@ impl Error for ResourceViewError {
         }
     }
 
-    fn cause(&self) -> Option<&Error> {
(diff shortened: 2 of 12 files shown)


alexheretic:
> Thank you! If I understand correctly, this is a non-breaking change.

Yep should function identically.

@kvark Do you think it's worth making a new release with this improvement, or wait for more stuff? I'm happy to raise the v0.18 pr if so.

kvark:
@alexheretic I don't mind either way :)

PR #3032, opened 2019-10-03
dx12: Refactor descriptor sets and pipeline layouts

msiglreith:
- Improve documentations on the descriptor set internals
- Prepare dynamic uniform/storage buffers
- Cache pipeline layout related calculations

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -272,7 +272,7 @@ impl Device {
     ) -> Result<(), d::ShaderError> {
         // Move the descriptor sets away to yield for the root constants at "space0".
         let space_offset = match layout {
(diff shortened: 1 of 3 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -1446,38 +1446,52 @@ impl d::Device<B> for Device {
     {
         // Pipeline layouts are implemented as RootSignature for D3D12.
         //
+        // Push Constants are implemented as root constants.
```
kvark: It would be great to have this architectural documentation more visible. For example, [in Metal](https://github.com/gfx-rs/gfx/blob/master/src/backend/metal/src/lib.rs#L1-L52) I tried to put everything in the doc comments of the crate, so that it's visible in the auto-generated documentation.

PR #3033, opened 2019-10-03
dx12: Implement dynamic uniform buffers

msiglreith:
Part of #1949

Implement dynamic uniform buffers via root descriptors (2 slots in the root signature per descriptor).
This approach is suitable for uniform buffers as these are translated to typed constant buffers on the shader side.

Storage buffers need to be implemented another way as we would oversubscribe the root signature and we need to deal with read-only descriptors in a way, which is not feasible with the root descriptor approach.

Tested with an modified quad example, executing two draw calls with a different offset.

TODO: d3d12-rs changes pending


msiglreith:
> Could you clarify why a dynamic CBV needs 2 root signature slots?

- Root constants are stored plain in the root signature (multiple 32 bit values)
- Root descriptors store the GPU address of the backing buffer -> 64bit == 2 slots
- Descriptor tables store an offset relative to the currently bound descriptor heaps. As their size won't surpass 4GB, an offset of 32 bit is enough -> 1 slot

msiglreith:
bors r=kvark

PR #3034, opened 2019-10-05
Impl backend for instance

### LINKED ISSUE #2963: Implement Backend for Instance
It seems like we could have one less type exposed by the API if we just make `Instance` to be that object that implements the `hal::Backend` trait.

yanchith:
Fixes #2963 

This is just a draft PR for now, but I'd like to get some feedback on whether this is the way to go. Also I currently don't have access to windows, so if there's any way to build the windows backends without it, let me know 😂 . I will be able to get to a windows machine tomorrow, most likely.

PR checklist:


yanchith:
Closing in favor of #3035

PR #3035, opened 2019-10-06
Add type Instance to trait Backend

### LINKED ISSUE #2963: Implement Backend for Instance

yanchith:
Fixes #2963 

Much simpler than #3034, but still had to add the cfgs to the gl backend.

PR checklist:
- [ ] `make reftests` succeeds (panics on macOS, but same for `master` branch, didn't have enough time on the windows machine to test)
  - macOS: metal ✅ , vulkan (moltenvk) ✅ , gl 🔴 (panics)
  - windows: vulkan ✅ , dx12 ✅ , dx11 ✅ , gl ✅ 

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -3166,6 +3166,7 @@ pub struct QueryPool;
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub enum Backend {}
 impl hal::Backend for Backend {
+    type Instance = Instance;
     type PhysicalDevice = PhysicalDevice;
     type Device = device::Device;
 
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -1144,6 +1144,7 @@ impl hal::Instance for Instance {
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub enum Backend {}
 impl hal::Backend for Backend {
+    type Instance = Instance;
     type PhysicalDevice = PhysicalDevice;
     type Device = Device;
 
--- a/src/backend/empty/src/lib.rs
+++ b/src/backend/empty/src/lib.rs
@@ -28,6 +28,7 @@ use std::ops::Range;
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 pub enum Backend {}
 impl hal::Backend for Backend {
+    type Instance = Instance;
     type PhysicalDevice = PhysicalDevice;
     type Device = Device;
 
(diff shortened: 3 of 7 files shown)


src/backend/gl/src/command.rs
```diff
@@ -552,6 +552,11 @@ impl CommandBuffer {
     }
 }
 
+#[cfg(any(
```
kvark: why do we need to add this attribute?

yanchith:
To my understanding the chain starts with:

```rust
#[cfg(any(
    all(not(target_arch = "wasm32"), feature = "glutin"),
    target_arch = "wasm32",
    feature = "wgl"
))]
impl hal::Backend for Backend {
    #[cfg(any(
        all(not(target_arch = "wasm32"), feature = "glutin"),
        feature = "wgl"
    ))]
    type Instance = Instance;

    #[cfg(all(target_arch = "wasm32", not(feature = "wgl")))]
    type Instance = Surface;
    
    // ... snip ...
}
```

The `hal::Backend` trait previously didn't provide an `Instance` associated type, but now that it does we have to select one in the impl. There are 3 candidates to choose from:

- `Instance` for the `glutin` feature
- `Instance` for the `wgl` feature
- `Surface` from `window::web` if `target_arch = "wasm32"`

This makes it difficult to produce a valid `impl hal::Backend for Backend` universally. The consumers using `Backend` require it to be constrained by `hal::Backend`, and since there is no cfg-less impl for that, I had to cfg all consumers of it.

I don't like it either 🙁. Perhaps if we could make one of the options for producing the `Instance` a default one? I am not sure, which one of those 3 it should be though. I am open to ideas.

kvark:
Today, gfx-backend-gl has a [single](https://github.com/gfx-rs/gfx/blob/97521403a8f92e197685ed7095117e59c5d04c36/src/backend/gl/src/lib.rs#L127) `impl hal::Backend`. We should just keep it, there is no need to have any `cfg()` on top.

We'll just have a few choices for the `Instance` inside, like what you wrote:
```rust
    #[cfg(any(
        all(not(target_arch = "wasm32"), feature = "glutin"),
        feature = "wgl"
    ))]
    type Instance = Instance;

    #[cfg(all(target_arch = "wasm32", not(feature = "wgl")))]
    type Instance = Surface;
```

There should be no other `cfg()` added.

yanchith:
Hmmm, I did as you asked, but I'm pretty sure the build will fail unless one of `glutin`/`wgl` features is passed, or the `wasm32-unknown-unknown`.

Do we want the user to explicitly pass the features/targets for the gl backend build? Hmm, that might actually make sense, especially given that wasm is not the usual build target and those features add more dependencies... Unfortunately the CI doesn't know it needs to add those build flags though: https://travis-ci.org/gfx-rs/gfx/jobs/594688627

kvark:
@yanchith we can always have some `struct Instance` that we define when neither of the other configurations are enabled (and we `impl hal::Instance for Instance`).

yanchith:
Damn, I should have figured that out myself 😄 On it!

PR #3037, opened 2019-10-09
[hal-0.3] Upgrade to raw-window-handle v0.3.0

antonok-edm:
PR checklist:

In the process, I've also added support for building a Vulkan surface using a raw XCB handle.


antonok-edm:
@kvark how does the versioning work for this repo? I just added it to `hal-0.3` because it's the only branch with the `raw-window-handle` dependency.

kvark:
@antonok-edm the "hal-0.3" branch contains fixes to "hal-0.3" crate and dependencies (e.g. backends). The"master" branch contains latest and greatest stuff. A change to bring raw-window-handle to "master" is #2967, and it has unfortunately stalled. Big thanks if you can get it rolling!

omni-viral:
Now it's time to upgrade to `raw-window-handle 0.3`

antonok-edm:
Ah, there's no way to change which branch a PR merges into... I can create a new one

omni-viral:
There is. Use `Edit` button on the top-right

antonok-edm:
Oh... that's news to me :sweat_smile:

PR #3042, opened 2019-10-13
Add `BASE_VERTEX_INSTANCE_DRAWING` feature flag

aleksijuvani:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -170,6 +170,7 @@ fn get_features(
         | hal::Features::FORMAT_BC
         | hal::Features::INSTANCE_RATE
         | hal::Features::SAMPLER_MIP_LOD_BIAS
+        | hal::Features::BASE_VERTEX_INSTANCE_DRAWING
 }
 
 fn get_format_properties(
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -1068,7 +1068,8 @@ impl hal::Instance for Instance {
                     Features::FORMAT_BC |
                     Features::INSTANCE_RATE |
                     Features::SAMPLER_MIP_LOD_BIAS |
-                    Features::SAMPLER_ANISOTROPY,
+                    Features::SAMPLER_ANISOTROPY |
+                    Features::BASE_VERTEX_INSTANCE_DRAWING,
                 limits: Limits { // TODO
                     max_image_1d_size: d3d12::D3D12_REQ_TEXTURE1D_U_DIMENSION as _,
                     max_image_2d_size: d3d12::D3D12_REQ_TEXTURE2D_U_OR_V_DIMENSION as _,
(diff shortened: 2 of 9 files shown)


aleksijuvani:
I intended for this to be a gfx-hal specific flag, not necessarily a flag exposed by Vulkan Portability. It looks like we already have a flag here that is only used by gfx-hal—`INSTANCE_RATE`. Can we add another? As a sidenote, we should fix the mask here.

As for Vulkan Portability, this can be implemented by re-binding the vertex buffer with a different offset, correct—we probably don't need to expose any new flags there. In any case, I think gfx-portability would need the ability to query support for this feature from gfx-hal?

kvark:
> It looks like we already have a flag here that is only used by gfx-hal—INSTANCE_RATE

Yes. The difference is that this feature is *on top* of what Vulkan provides, so gfx-portability doesn't need this flag until it wants to expose an extension that is based on it.

Unlike `INSTANCE_RATE`, the base vertex thing *is* a part of Vulkan... That is to say, we have 2 choices:
  1. implement the workaround in gfx-hal and possibly expose this as a "hint" kind of flag. Note: we don't currently have any hints, so that would be the first.
  2. have it as a *restriction* in gfx-hal and implement the workaround in gfx-portability. That seems a bit appealing, but I don't think it's practical. We already have the logic in the backends that would re-bind vertex buffers, say, upon changing the pipeline. So handling the base vertex similarly isn't radically changing the backends logic, it seems fairly fitting in there.

Saying that, it looks to me that just unconditionally supporting this feature in gfx-hal is the way to go. Whether or not we should have "hints" bitflags can be discussed separately with more things in mind.

aleksijuvani:
While I haven't benchmarked this, I'm worried that re-binding vertex buffers per draw call would have performance implications. This is why I'm more inclined to make this a restriction in gfx-hal and make it explicit to the user, rather than have it re-bind the vertex buffers automatically. Changing the pipeline is already an expensive operation, so I'm not worried about it in that situation.

kvark:
@aleksijuvani I understand your point. Doing any state changes at draw() frequency is something we want to avoid. Before we make it a restriction in gfx-hal, we need to have a rough idea of how the gfx-portability implementation would look like.

aleksijuvani:
I'll try to put together a pull request for gfx-portability by the end of this week to see what this would look like.

aleksijuvani:
Sorry for taking such a long time getting back to you! I don't think I'll have the time to finish work on this pull request. I'm no longer doing much work with Rust so it's been becoming increasingly harder to justify the time investment. For what it's worth, I think this is more or less good to merge—we just need to implement the fallback in gfx-portability if this is not natively supported (which is just on some iOS devices and the iOS Simulator, really). I don't think that should be holding back getting this merged.

kvark:
@aleksijuvani I'll pick that up. I think we'll have a new bitflag field for "hints", which this will go into.

kvark:
Replaced by #3167

Issue #3015, opened 2019-09-13
Vulkan debug layer errors in quad and colour-uniforms examples


Herschel:
`cargo run --bin=quad --features=vulkan` and `cargo run --bin=colour-uniforms --features=vulkan` on the master branch produce spew from the Vulkan validation layer:

```
VALIDATION [VUID-vkDestroyFramebuffer-framebuffer-00892 (0)] : Cannot call vkDestroyFramebuffer on VkFramebuffer 0x14e1d60000000033[] that is currently in use by a command buffer. The Vulkan spec states: All submitted commands that refer to framebuffer must have completed execution (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkDestroyFramebuffer-framebuffer-00892)
```

OS: Windows 10 Home 64-bit
GPU: Geforce GTX 1060
LunarG Vulkan SDK 1.1.121.0

kvark:
Found an interesting link for our info - https://github.com/KhronosGroup/Vulkan-Docs/issues/117#issuecomment-206540937

kvark:
It we have to change the way the fence operates on the frame, we could fence the artificial submission we do at `present_surface` and wait for it before re-using the swapchain image and destroying the framebuffer.

kvark:
Upon further investigation, the `quad` validation error appears to be a bug in the validation layer itself - https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/12 , so it's harmless.
`colour-uniform` uses an entirely different swapchain model, needs to be looked at.

Herschel:
These are the validation errors given by `colour-uniform`:


```
[2019-09-14T17:31:08Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-vkEndCommandBuffer-commandBuffer-00060 (0)] : vkEndCommandBuffer(): It is invalid to issue this call inside an active VkRenderPass 0x18[]. The Vulkan spec states: If commandBuffer is a primary command buffer, there must not be an active render pass instance (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkEndCommandBuffer-commandBuffer-00060)
object info: (type: COMMAND_BUFFER, hndl: 2946642394896)

[2019-09-14T17:31:08Z ERROR gfx_backend_vulkan]
[...]

PR #3043, opened 2019-10-14
Fix validation errors in quad example

demurgos:
PR checklist:
- [x] tested examples with the following backends: vulkan, gl (wasm)

Fix validation errors in quad example

This commit fixes the Vulkan validation errors for the `quad` example:
- The framebuffer is destroyed without waiting for the command buffer to be executed.
- The swapchain is not destroyed.

This commit also adds a small comment explaining why `main` extends the lifetime of `window` beyond its initial scope.

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -121,6 +121,7 @@ fn main() {
             .expect("Failed to create an instance!");
         let surface = instance.create_surface(&window).expect("Failed to create a surface!");
         let adapters = instance.enumerate_adapters();
+        // Return `window` so it is not dropped: dropping it invalidates `surface`.
         (window, instance, adapters, surface)
     };
     #[cfg(feature = "gl")]
@@ -619,7 +620,7 @@ where
             submission_complete_fences.push(
                 device
                     .create_fence(true)
-                    .expect("Could not create semaphore"),
+                    .expect("Could not create fence"),
             );
             cmd_buffers.push(cmd_pools[i].allocate_one(command::Level::Primary));
         }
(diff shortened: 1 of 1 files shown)


examples/quad/main.rs
```diff
@@ -877,6 +878,9 @@ where
                 Some(&self.submission_complete_semaphores[frame_idx]),
             );
 
+            self.device
+                .wait_for_fence(&self.submission_complete_fences[frame_idx], !0)
```
kvark: We can't do this. I mean, technically we can, but that would not be a good example if it waits on CPU for the GPU work to be done every frame...

examples/quad/main.rs
```diff
@@ -121,6 +121,7 @@ fn main() {
             .expect("Failed to create an instance!");
         let surface = instance.create_surface(&window).expect("Failed to create a surface!");
         let adapters = instance.enumerate_adapters();
+        // Return `window` so it is not dropped: dropping it invalidates `surface`.
```
rukai: Nice!

examples/quad/main.rs
demurgos: @kvark I introduced this to solve the following validation error. Is there a better way?

```
[2019-10-14T11:34:03Z ERROR gfx_backend_vulkan]
VALIDATION [VUID-vkDestroyFramebuffer-framebuffer-00892 (0)] : Cannot call vkDestroyFramebuffer on VkFramebuffer 0x59[] that is currently in use by a command buffer. The Vulkan spec states: All submitted commands that refer to framebuffer must have completed execution (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkDestroyFramebuffer-framebuffer-00892)
object info: (type: FRAMEBUFFER, hndl: 89)
```

kvark: I'll need to look into this closer. The new swapchain API that we introduced recently is most difficult on Vulkan. It's not too easy to investigate.

kvark: Actually, it's not our bug, see https://github.com/gfx-rs/gfx/issues/3015#issuecomment-531491026

demurgos: Okay, thank you for checking this. I'll remove the `wait_for_fence` call then.

PR #3044, opened 2019-10-14
Add screenshots for the examples that pop open a window

### LINKED ISSUE #3041: Screenshot of examples in README
Once upon a time there were screenshots of the rendered examples in their respective READMEs. It seems like this is no longer the case. I think it would be a good idea to include a README in each of the three examples with a screenshot of the desired output.

mandeep:
fixes #3041 

This PR adds screenshots for the quad and colour-uniform examples. Please check that the screenshots are rendered correctly.


examples/colour-uniform/README.md
```diff
@@ -0,0 +1,6 @@
+# Colour Uniform
+
+The following image is an output of `cargo run --bin colour-uniform --features=gl` with R set at 127,
+B set at 100, and G set at 50:
+
+![colour_uniform](colour_uniform_screenshot.png "Colour Uniform")
```
kvark: does the name need to include the example name? since it's already in the folder, it seems that we could just call it "screenshot.png" or something.

mandeep: I guess it's a matter of taste. I prefer to be explicit in naming, but I'm more than happy to change the filename to screenshot.png if necessary. :smiley:

kvark: I mean, we aren't calling the other files "colour_uniform_cargo.toml" etc :)
If you don't mind, let's rename these to just "screenshot.png" then.

mandeep: Good point! I've updated the filenames.

mandeep: Anything else needed from me?

kvark:
Bors r+

mandeep:
@kvark thanks for taking the time to review :smiley:

kvark:
Sorry, something happened to my email: I sent r+ long long ago but it never reached github, and I didn't notice that. Thank you for pinging me!
bors retry

kvark:
Bors retry

PR #3047, opened 2019-10-17
Build Metal shaders in Makefile instead of build.rs

### LINKED ISSUE #3046: Avoid building shaders in build.rs scripts
Doing so is not compatible with cross-compilation, which for example is what [Gecko try bots do](https://bugzilla.mozilla.org/show_bug.cgi?id=1586291) for testing macOS.
Instead, we should have a `make` command to update shader binaries based on the source, and both the source with the binaries are going to be versioned.

grovesNL:
Fixes #3046
PR checklist:


Makefile
```diff
@@ -82,14 +78,8 @@ quad:
...
-	#if [ -e $(SDL2_CONFIG) ]; then exit 1; fi
-	#mkdir -p $(SDL2_DEST)
-	#test -f $(SDL2_DEST)/dev.deb || curl -sLo $(SDL2_DEST)/dev.deb $(SDL2_PPA)/libsdl2-dev_2.0.3+z4~20140315-8621-1ppa1precise1_amd64.deb
-	#test -f $(SDL2_DEST)/bin.deb || curl -sLo $(SDL2_DEST)/bin.deb $(SDL2_PPA)/libsdl2_2.0.3+z4~20140315-8621-1ppa1precise1_amd64.deb
-	#dpkg-deb -x $(SDL2_DEST)/bin.deb .
-	#dpkg-deb -x $(SDL2_DEST)/dev.deb .
-	#sed -e s,/usr,$(SDL2_DEST),g $(SDL2_CONFIG) > $(SDL2_CONFIG).new
-	#mv $(SDL2_CONFIG).new $(SDL2_CONFIG)
-	#chmod a+x $(SDL2_CONFIG)
+metal-shaders:
```
kvark: do you think we could rewrite it as `make shader-binaries` that would detect the backends automatically based on the OS?

grovesNL: Sure, updated

grovesNL:
bors r=kvark

PR #3064, opened 2019-10-29
Make spirv_cross an optional dependency of gfx-auxil

### LINKED ISSUE #3063: Don't depend on spirv-cross in gfx-auxil
All backends depend on `gfx-auxil`, which drags `spirv_cross` unconditionally at the moment.

kvark:
Fixes #3063
This is a breaking revision for gfx-auxil, but we can release patched backends for 0.4 if/when we want.

PR checklist:


kvark:
Bors r=aleksijuvani

PR #3068, opened 2019-11-01
[dx12] update d3d12 to 0.3 and use explicit linking

### LINKED ISSUE #3067: Only use d3d12-rs to access DX12 functionality
According to https://github.com/gfx-rs/wgpu/issues/365#issuecomment-548474938, we should have some code that explicitly loads and probes "d3d12.dll" and "dxgi.dll". It's better to have it isolated within "d3d12-rs" instead of duplicating there.

Blocked by https://github.com/gfx-rs/d3d12-rs/issues/18 at the moment.

kvark:
Fixes #3067 (not exactly matching the wording but addresses the real blocker)
Depends on https://github.com/gfx-rs/d3d12-rs/pull/20 merged and published, doesn't need to be updated when that happens
PR checklist:
- [x] tested examples with the following backends: d3d12

Verified on a Windows 7 machine that we are gracefully failing to init DX12 now.


kvark:
bors retry

PR #3069, opened 2019-11-01
[0.4/dx12] update d3d12 to 0.3 and use explicit linking

kvark:
Sibling of #3068
Also depends on https://github.com/gfx-rs/d3d12-rs/pull/20


kvark:
bors r=msiglreith

kvark:
not sure what it's doing, hmm
bors retry

PR #3071, opened 2019-11-02
UPDATE: The getting started readme now informs macOS users that cmake…

Bergasms:
… is a requirement for one of the libraries. Without this information a clean clone will not work and the user will be presented with a compilation error when trying to run the examples.

Fixes #issue
PR checklist:

### DIFF
--- a/info/getting_started.md
+++ b/info/getting_started.md
@@ -2,7 +2,9 @@
 
 ## macOS Dependencies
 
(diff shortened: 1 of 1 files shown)


Bergasms:
@kvark Sure. Just wanted it in because the examples is generally the first port of entry so having it not work is sucky. Will update the PR.

PR #3072, opened 2019-11-04
[0.3/dx12] Fix for building with rust 1.35.0

zakorgy:
Fixes https://ci.appveyor.com/project/loki04/webrender/builds/28520384#L580
PR checklist:
- [ ] `make` succeeds (on *nix)
- [ ] `make reftests` succeeds
- [x] tested examples with the following backends: dx12
- [ ] `rustfmt` run on changed code

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/3072)

PR #3074, opened 2019-11-04
[0.4/dx12] Fix for building with rust 1.35.0

zakorgy:
The sibling of https://github.com/gfx-rs/gfx/pull/3072
PR checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/gfx-rs/gfx/3074)


kvark:
Looks like there is a fix missing, according to CI?

kvark:
perhaps we could have a short note in CHANGELOG as well (not required here, I can follow up later)

Issue #3079, opened 2019-11-04
ShaderStageFlags::ALL does not match vulkan definition


icefoxen:
Short info header:
- GFX version: 0.4.0
- OS: Linux Ubuntu 18.04
- GPU: NVidia GTX 1060, Vulkan backend.

`ShaderStageFlags::ALL` is defined as `GRAPHICS.bits | COMPUTE.bits` here: https://github.com/gfx-rs/gfx/blob/5ad5cfea6e20ab73851a4f75e2ce3b92ac0a39dd/src/hal/src/pso/mod.rs#L95 
But, the Vulkan spec explicitly says it must be 0x7FFF_FFFF : https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VkShaderStageFlagBits .  Presumably they want it to be "all" for realsies, always, no matter what extensions may get added in the future.  I'm not yet sure whether this value gets fed directly into the Vulkan backend without conversion but I'm getting validation errors that *look* related so I'm betting it is.

icefoxen:
Confirmed the value is sent straight to Vulkan, the conversion to Vulkan types is defined here: https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkan/src/conv.rs#L174-L176  and used here: https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkan/src/command.rs#L942

kvark:
Please post the validation errors for completeness. That one sounds like something we could release in a patch.

icefoxen:
Turns out the validation errors were due to something else.  Sigh!  Still technically maybe-not-correct though.

omni-viral:
IIRC this special value exists so you can specify **all stages that are valid in the context**

PR #3078, opened 2019-11-04
Implement Error trait for error types

omni-viral:
Fixes https://github.com/gfx-rs/gfx/issues/3073

Fixes #issue
PR checklist:
- [ ] `make` succeeds (on *nix)
- [ ] `make reftests` succeeds
- [ ] tested examples with the following backends:
- [ ] `rustfmt` run on changed code

PR #3080, opened 2019-11-05
[0.4/hal] Error derives and ShaderStageFlags

kvark:
Has #3078 for 0.4, also has a fix for #3079 (still need to follow-up on `master`).
PR checklist:

Reviewers - how do you feel about the flag change? Can we actually publish it as a patch?

### DIFF
--- a/src/hal/src/pso/mod.rs
+++ b/src/hal/src/pso/mod.rs
@@ -40,6 +40,17 @@ impl From<device::OutOfMemory> for CreationError {
     }
 }
 
+impl std::fmt::Display for CreationError {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CreationError::OutOfMemory(err) => write!(fmt, "Failed to create pipeline: {}", err),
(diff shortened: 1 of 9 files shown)


src/hal/src/pso/mod.rs
```diff
@@ -40,6 +40,17 @@ impl From<device::OutOfMemory> for CreationError {
     }
 }
 
+impl std::fmt::Display for CreationError {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CreationError::OutOfMemory(err) => write!(fmt, "Failed to create pipeline: {}", err),
+            CreationError::Other => write!(fmt, "Failed to create pipeline: Unsupported usage: Implementation specific error ocurred"),
```
grovesNL: Typo in "occurred" here and a few other places

grovesNL:
> Reviewers - how do you feel about the flag change? Can we actually publish it as a patch?

It seems reasonable to me to include it in a patch release – I don't think anyone should be relying on the current value directly

kvark:
Thanks @grovesNL !
Notes are addressed.
bors r=grovesNL

PR #3081, opened 2019-11-05
ShaderStageFlags fix

### LINKED ISSUE #3079: ShaderStageFlags::ALL does not match vulkan definition

kvark:
Follow-up to #3078 with pieces from #3080
Fixes #3079
PR checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
 # Change Log
 
+### hal-0.4.1 (04-11-2019)
+  - `Error` implementations
+  - fix `ShaderStageFlags::ALL`
+
 ### backend-dx12-0.4.1, backend-dx11-0.4.2 (01-11-2019)
   - switch to explicit linking of "d3d12.dll", "d3d11.dll" and "dxgi.dll"
+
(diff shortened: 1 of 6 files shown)


kvark:
bors r=grovesNL

PR #3084, opened 2019-11-09
Implement Error for pso::CreationError

BourgondAries:
Fixes #issue
PR checklist:

### DIFF
--- a/src/hal/src/device.rs
+++ b/src/hal/src/device.rs
@@ -105,6 +105,24 @@ impl From<DeviceLost> for OomOrDeviceLost {
     }
 }
 
+impl std::fmt::Display for OomOrDeviceLost {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            OomOrDeviceLost::DeviceLost(err) => write!(fmt, "Failed querying device: {}", err),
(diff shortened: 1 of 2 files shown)


src/hal/src/device.rs
```diff
@@ -105,6 +105,24 @@ impl From<DeviceLost> for OomOrDeviceLost {
     }
 }
 
+impl std::fmt::Display for OomOrDeviceLost {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            OomOrDeviceLost::DeviceLost(err) => write!(fmt, "Failed querying device: {}", err),
```
kvark: should the messages be more aligned with enum variants?

BourgondAries: Yes, but I based this off CreationError, which says `Failed to create pipeline`, while not specifying pipeline in its name, it's just used where pipelines are made, so I used the same idea here. What should it be instead, and should we change CreationError too?

kvark: > while not specifying pipeline in its name

it's specifying the pipeline in the "full name", as in "pso::CreationError", the semantics is "error creating a pipeline".

Actually, now that I look at it, you code is fine since the `impl Display` for inner `DeviceLost` and `OutOfMemory` should make it into the message produced.

PR #3094, opened 2019-11-16
[0.4/GL] Update glutin version

Gordon-F:
Fixes #issue
PR checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-gl-0.4.1 (16-11-2019)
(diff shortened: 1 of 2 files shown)


kvark:
Thank you!
Is the new `glutin` compatible with the old one? It happens to be a public dependency of the GL backend, so we'd technically have to break the version if `glutin` is breaking...
Another option would be to yank 0.4.0 version of the GL backend and leave only this 0.4.1.
@grovesNL any strong feelings on this one?

kvark:
Yep, can't do this - a breaking change (unfortunately!)

PR #3100, opened 2019-11-25
[mtl] Use a separate internal queue for mapping invalidation

### LINKED ISSUE #3099: Mapping invalidation is a hard stall on Metal
See this code - https://github.com/gfx-rs/gfx/blob/3641183231f16877d4ea2fbdb2ff208ce736d6c4/src/backend/metal/src/device.rs#L1817

I believe this is inacceptable. The discussion in VkPortability is also kicked off.
Any ideas on how to make this better are very welcome!

Note: this is affecting (negatively) our WebRender friends.

kvark:
Fixes #3099 , presumably. Need to do some testing to confirm...

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -179,9 +179,10 @@ impl VisibilityShared {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct Device {
     pub(crate) shared: Arc<Shared>,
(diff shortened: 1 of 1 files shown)


kvark:
Tested on Szeged's WR fork. Didn't benchmark specifically, but at least this doesn't appear to be breaking anything. I think we can proceed.

kvark:
bors r-
bors r=grovesNL

PR #3102, opened 2019-11-26
Refactor DescriptorType

### LINKED ISSUE #3055: Refactor DescriptorType
Related to #1372, we should turn it into a rich enum.
Something (but maybe not precisely) like this would do:

```rust
bitflags! {
    pub struct ImageFlags: u8 {
        const COMBINED_WITH_SAMPLER = 0x1;
        const MULTISAMPLED = 0x2;
    }
}

bitflags! {
    pub struct BufferFlags: u8 {
        const TEXEL_VIEW = 0x1;
        const STORAGE = 0x2;
        const DYNAMIC_OFFSET = 0x4;
    }
}

enum DescriptorType {
  Sampler,
  Image {
    flags: ImageFlags,
    kind: ViewKind,
  },
  StorageImage {
    kind: ViewKind,
  },
  Buffer {
    flags: BufferFlags,
  },
  InputAttachment,
}
```

It would be great to get in for 0.4 release, so that wgpu-rs can start playing with this and take advantage of the new semantics. However, this seems like a big API change, so I'm leaning towards postponing it to 0.5. This way we can discuss the API in more detail before proceeding.

cormac-obrien:
Fixes #3055 
PR checklist:
- [ ] `make reftests` succeeds (seems broken for DX12 on master?)
- [x] tested examples with the following backends: vulkan, dx11, dx12, gl

This PR refactors `DescriptorType`:

```rust
pub enum DescriptorType {
    /// A descriptor associated with sampler.
    Sampler,
    /// A single descriptor associated with both a sampler and an image to be
    /// used together.
    CombinedImageSampler,
    /// A descriptor associated with an image.
    Image {
        /// The specific type of this image descriptor.
        ty: ImageDescriptorType,
    },
    /// A descriptor associated with a buffer.
    Buffer {
        /// The access type of this buffer descriptor.
        access: BufferDescriptorAccess,
        /// The format of this buffer descriptor.
        format: BufferDescriptorFormat,
    },
    /// A descriptor associated with an input attachment.
    InputAttachment,
}
```

The number of top-level variants is reduced from 11 to 5 by turning `Image` and `Buffer` into struct variants, which allows e.g.:
```rust
DescriptorType::Buffer {
    access: Storage,
    format: Texel,
}
```
for what was previously `StorageTexelBuffer`.

One other note: `gfx-backend-dx11` relies on the conversion from `DescriptorType` to `u32` for sorting descriptors, so I've kept the old `DescriptorType` as `RawDescriptorType` for that purpose.

### DIFF
--- a/src/backend/dx12/src/resource.rs
+++ b/src/backend/dx12/src/resource.rs
@@ -467,10 +467,12 @@ impl From<pso::DescriptorType> for DescriptorContent {
 
         match ty {
             Dt::Sampler => Dc::SAMPLER,
-            Dt::CombinedImageSampler => Dc::SRV | Dc::SAMPLER,
             Dt::Image { ty } => match ty {
                 Idt::Storage => Dc::SRV | Dc::UAV,
-                Idt::Sampled => Dc::SRV,
+                Idt::Sampled { with_sampler } => match with_sampler {
+                    true => Dc::SRV | Dc::SAMPLER,
+                    false => Dc::SRV,
+                }
             }
             Dt::Buffer { ty, format } => match ty {
                 Bdt::Storage { .. } => match format {
--- a/src/backend/dx12/src/resource.rs
+++ b/src/backend/dx12/src/resource.rs
@@ -459,7 +459,7 @@ impl From<pso::DescriptorType> for DescriptorContent {
         use hal::pso::{
             DescriptorType as Dt,
             ImageDescriptorType as Idt,
-            BufferDescriptorAccess as Bda,
+            BufferDescriptorType as Bdt,
             BufferDescriptorFormat as Bdf
         };
 
@@ -472,14 +472,14 @@ impl From<pso::DescriptorType> for DescriptorContent {
                 Idt::Storage => Dc::SRV | Dc::UAV,
                 Idt::Sampled => Dc::SRV,
             }
-            Dt::Buffer { access, format } => match access {
-                Bda::Storage => match format {
-                    Bdf::Dynamic => Dc::SRV | Dc::UAV | Dc::DYNAMIC,
-                    Bdf::Structured | Bdf::Texel => Dc::SRV | Dc::UAV,
+            Dt::Buffer { ty, format } => match ty {
+                Bdt::Storage { .. } => match format {
+                    Bdf::Structured { dynamic_offset: true } => Dc::SRV | Dc::UAV | Dc::DYNAMIC,
+                    Bdf::Structured { dynamic_offset: false } | Bdf::Texel => Dc::SRV | Dc::UAV,
                 }
-                Bda::Uniform => match format {
-                    Bdf::Dynamic => Dc::CBV | Dc::DYNAMIC,
-                    Bdf::Structured => Dc::CBV,
+                Bdt::Uniform => match format {
+                    Bdf::Structured { dynamic_offset: true } => Dc::CBV | Dc::DYNAMIC,
+                    Bdf::Structured { dynamic_offset: false } => Dc::CBV,
                     Bdf::Texel => Dc::SRV,
                 }
             }
--- a/src/backend/dx12/src/resource.rs
+++ b/src/backend/dx12/src/resource.rs
@@ -456,21 +456,34 @@ impl DescriptorContent {
 
 impl From<pso::DescriptorType> for DescriptorContent {
     fn from(ty: pso::DescriptorType) -> Self {
-        use hal::pso::DescriptorType as Dt;
+        use hal::pso::{
+            DescriptorType as Dt,
+            ImageDescriptorType as Idt,
+            BufferDescriptorAccess as Bda,
+            BufferDescriptorFormat as Bdf
+        };
+
+        use DescriptorContent as Dc;
+
         match ty {
-            Dt::Sampler => DescriptorContent::SAMPLER,
-            Dt::CombinedImageSampler => DescriptorContent::SRV | DescriptorContent::SAMPLER,
-            Dt::SampledImage | Dt::InputAttachment | Dt::UniformTexelBuffer => {
-                DescriptorContent::SRV
+            Dt::Sampler => Dc::SAMPLER,
+            Dt::CombinedImageSampler => Dc::SRV | Dc::SAMPLER,
+            Dt::Image { ty } => match ty {
+                Idt::Storage => Dc::SRV | Dc::UAV,
+                Idt::Sampled => Dc::SRV,
             }
-            Dt::StorageImage | Dt::StorageBuffer | Dt::StorageTexelBuffer => {
-                DescriptorContent::SRV | DescriptorContent::UAV
+            Dt::Buffer { access, format } => match access {
+                Bda::Storage => match format {
+                    Bdf::Dynamic => Dc::SRV | Dc::UAV | Dc::DYNAMIC,
+                    Bdf::Structured | Bdf::Texel => Dc::SRV | Dc::UAV,
+                }
+                Bda::Uniform => match format {
+                    Bdf::Dynamic => Dc::CBV | Dc::DYNAMIC,
+                    Bdf::Structured => Dc::CBV,
+                    Bdf::Texel => Dc::SRV,
+                }
             }
-            Dt::StorageBufferDynamic => {
-                DescriptorContent::SRV | DescriptorContent::UAV | DescriptorContent::DYNAMIC
-            }
-            Dt::UniformBuffer => DescriptorContent::CBV,
-            Dt::UniformBufferDynamic => DescriptorContent::CBV | DescriptorContent::DYNAMIC,
+            Dt::InputAttachment => Dc::SRV,
         }
     }
 }
--- a/src/backend/metal/src/native.rs
+++ b/src/backend/metal/src/native.rs
@@ -814,10 +814,14 @@ impl From<pso::DescriptorType> for DescriptorContent {
     fn from(ty: pso::DescriptorType) -> Self {
         match ty {
             pso::DescriptorType::Sampler => DescriptorContent::SAMPLER,
-            pso::DescriptorType::CombinedImageSampler => {
-                DescriptorContent::TEXTURE | DescriptorContent::SAMPLER
+            pso::DescriptorType::Image { ty } => {
+                match ty {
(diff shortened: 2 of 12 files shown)


src/hal/src/pso/descriptor.rs
```diff
@@ -39,7 +39,7 @@ pub type DescriptorArrayIndex = usize;
 #[repr(C)]
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum DescriptorType {
+pub enum RawDescriptorType {
```
kvark: Let's remove it entirely :)

In Dx11 we can just rely on `Ord` implementation for our `DescriptorType`.
In Vulkan we should be having the `match` into `ash` descriptor type (which matches this one).

src/hal/src/pso/descriptor.rs
```diff
@@ -67,6 +67,91 @@ pub enum DescriptorType {
...
+    Storage,
+    /// Uniform buffers allow only load operations.
+    Uniform,
+}
+
+/// Format of a buffer.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum BufferDescriptorFormat {
+    // TODO: seems to have no effect in hal, ask about removing or documenting that it's a no-op
```
kvark: This has great effect, which we've been using for a long time, especially in wgpu.
What made you confused about this?

src/hal/src/pso/descriptor.rs
```diff
@@ -67,6 +67,91 @@ pub enum DescriptorType {
...
+    /// TODO
+    Dynamic,
+    /// The buffer is interpreted as a structure defined in a shader.
+    Structured,
```
kvark: instead of having a separate `Dynamic` variant here, let's have this as a bool in
```rs
Structured { dynamic_offset: bool },
```

src/hal/src/pso/descriptor.rs
```diff
@@ -67,6 +67,91 @@ pub enum DescriptorType {
...
+        }
+    }
+}
+
+/// Access type of a buffer.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum BufferDescriptorAccess {
+    /// Storage buffers allow load, store, and atomic operations.
+    Storage,
```
kvark: how do you feel about extending this to:
```rust
Storage { read_only: true }
```
I know this is a bit out of scope, but that's something WebGPU has and would be good to take advantage of on the gfx-rs level as well.
And since your PR reworks all the relevant code, it would be convenient to do it here. To clarify, there is no expectation that this boolean would affect anything today, you can just ignore it in the backends.

src/hal/src/pso/descriptor.rs
```diff
@@ -67,6 +67,91 @@ pub enum DescriptorType {
...
+            }
+            DescriptorType::InputAttachment => RawDescriptorType::InputAttachment,
+        }
+    }
+}
+
+/// Access type of a buffer.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum BufferDescriptorAccess {
```
kvark: let's rename it to `BufferDescriptorType` for the following reasons:
  - uniform/storage are not exactly different as read/write, since one can have a read-only storage. Uniform buffers are different types of bindings
  - consistency with `ImageDescriptorType`
  - avoid confusion with `buffer::Access` type

work/scenes/compute.ron
```diff
@@ -19,7 +22,10 @@
 			capacity: 1,
 			ranges: [
 				(
-					ty: StorageBuffer,
+					ty: Buffer(
+              access: Storage,
```
kvark: the indentation seems all broken here

src/hal/src/pso/descriptor.rs
cormac-obrien: I made this comment looking at the "legacy" backends (gl/dx11), realized that it behaved as expected in the modern frontends (vk/mtl/dx12) and promptly forgot to remove the comment 😅

src/hal/src/pso/descriptor.rs
```diff
@@ -33,38 +33,67 @@ pub type DescriptorBinding = u32;
...
+}
+
+/// Format of a buffer.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum BufferDescriptorFormat {
+    /// The buffer is interpreted as a structure defined in a shader.
+    Structured {
+        /// If true, the buffer is accessed by an additional offset specified by
+        /// `DescriptorSetWrite::array_offset`.
```
kvark: the comment is incorrect, the offset this refers to is actually coming from the last parameter of [bind_graphics_descriptor_sets](https://docs.rs/gfx-hal/0.4.1/gfx_hal/command/trait.CommandBuffer.html#tymethod.bind_graphics_descriptor_sets) (and the corresponding compute one)

src/hal/src/pso/descriptor.rs
```diff
@@ -33,38 +33,67 @@ pub type DescriptorBinding = u32;
...
+/// Specific type of a buffer.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum BufferDescriptorType {
+    /// Storage buffers allow load, store, and atomic operations.
+    Storage {
+        /// If true, store operations are not permitted on this buffer.
+        read_only: bool,
+    },
+    /// Uniform buffers allow only load operations.
```
kvark: we should rephrase this, as I don't think users are even expected to call `load` on a uniform buffer (that's what storage buffers are for!).
Maybe "Uniform buffers allow providing constant data to a shader" ?

src/backend/metal/src/native.rs
```diff
@@ -817,18 +817,18 @@ impl From<pso::DescriptorType> for DescriptorContent {
...
-            | pso::DescriptorType::InputAttachment => DescriptorContent::TEXTURE,
-            pso::DescriptorType::UniformBuffer | pso::DescriptorType::StorageBuffer => {
-                DescriptorContent::BUFFER
-            }
-            pso::DescriptorType::UniformBufferDynamic
-            | pso::DescriptorType::StorageBufferDynamic => {
-                DescriptorContent::BUFFER | DescriptorContent::DYNAMIC_BUFFER
+            pso::DescriptorType::Image { .. } => DescriptorContent::TEXTURE,
+            pso::DescriptorType::Buffer { format, .. } => match format {
+                pso::BufferDescriptorFormat::Structured { dynamic_offsets } => {
```
kvark: this doesn't compile since the field is singular: `dynamic_offset`

src/backend/dx12/src/resource.rs
```diff
@@ -456,21 +456,34 @@ impl DescriptorContent {
...
+            Dt::Image { ty } => match ty {
+                Idt::Storage => Dc::SRV | Dc::UAV,
+                Idt::Sampled => Dc::SRV,
             }
-            Dt::StorageBufferDynamic => {
-                DescriptorContent::SRV | DescriptorContent::UAV | DescriptorContent::DYNAMIC
+            Dt::Buffer { ty, format } => match ty {
+                Bdt::Storage { .. } => match format {
+                    Bdf::Structured { dynamic_offset: true } => Dc::SRV | Dc::UAV | Dc::DYNAMIC,
+                    Bdf::Structured { dynamic_offset: false }| Bdf::Texel => Dc::SRV | Dc::UAV,
```
kvark: nit: let's give "| Bdf::Texel" a bit more space

src/hal/src/pso/descriptor.rs
```diff
@@ -33,38 +33,67 @@ pub type DescriptorBinding = u32;
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum ImageDescriptorType {
+    /// A sampled image allows sampling operations.
+    Sampled,
+    /// A storage image allows load, store and atomic operations.
+    Storage,
+}
+
+/// The type of a descriptor. TODO: more specific doc
```
grovesNL: nit: we could probably just remove the TODO in this case

src/hal/src/pso/descriptor.rs
```diff
@@ -33,38 +33,67 @@ pub type DescriptorBinding = u32;
...
+    /// conversion when loaded in a shader.
+    Texel,
+}
+
+/// Specific type of an image descriptor.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum ImageDescriptorType {
+    /// A sampled image allows sampling operations.
+    Sampled,
```
kvark: Would you be open to have this as :
```rust
Sampled { with_sampler: bool },
```
Such that it replaces `CombinedImageSampler`.

The reasoning is for #3055, as the description says, we'll need to add a few fields to each image descriptor:
  - dimensionality
  - multisampling bit

It would be much more convenient for us and the users if there was a single entry point for image descriptors, even ones that come combined with samplers.

work/scenes/compute.ron
```diff
@@ -19,7 +22,10 @@
 			capacity: 1,
 			ranges: [
 				(
-					ty: StorageBuffer,
+					ty: Buffer(
+							ty: Storage(read_only: false),
```
kvark: nit: indentation here seems broken

kvark:
Ideally, we'd also squash some of the commits in the middle (e.g. one that fixes Metal), so that every one is buildable at least.
Ready when you are :)

kvark:
Beautiful, and amazing work, thank you!
bors r+

PR #3110, opened 2019-12-18
[mtl] restore the viewport after clear_attachments

### LINKED ISSUE #3109: [mtl] clear_attachments changes viewport but does not reset it
Short info header: 
- GFX version: hal-0.4
- OS: MacOS
- GPU: Intel

Calling `clear_attcahment` after the viewport is set changes the viewport but does not reset it.
I've a modified quad example for it: https://github.com/zakorgy/gfx/tree/clear-attachment (just added a `clear_attachment` to clear an area to red)

The expected result with Vulkan and DX12:
Screenshot from 2019-12-18 09-08-13

The actual result with Metal, the rendered image is flipped:
Screenshot 2019-12-18 at 9 00 42

I suspect this is caused by changing the viewport in `clear_attachments`: https://github.com/gfx-rs/gfx/blob/hal-0.4/src/backend/metal/src/command.rs#L2953 . Also I'm not sure what happens with scissor, because it's also set for the clear but doesn't seem to affect the result. 

cc @kvark

kvark:
Fixes #3109
PR checklist:

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -378,11 +378,12 @@ impl State {
         }
     }
 
-    fn make_render_commands(
-        &self,
-        aspects: Aspects,
-    ) -> impl Iterator<Item = soft::RenderCommand<&soft::Ref>> {
-        // Apply previously bound values for this command buffer
+    fn make_viewport_and_scissor_commands(
+        &self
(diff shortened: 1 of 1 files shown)


kvark:
I believe this is simple enough to merge
bors r+

Issue #3114, opened 2019-12-28
Driver throws exception when Dx11 backend tries to bind a buffer with opt-level > 0


feawilsch:
Discovered with my app, reproducible with the quad and colour-uniform examples.

Tested configurations (all with approx. identical behavior):
- GFX version: 0.4.3, latest master
- OS: Windows 10, Windows 7
- GPU: Nvidia Geforce 840M, Intel HD Graphics 4600, AMD Radeon HD 7770
- Rustc: 1.40, 1.42-Nightly 12-25
- opt-level:  1, 2, 3

kvark:
Thank you for reporting this! Did you run via MSVS to see if any relevant debug run-time messages pop up? Assuming we are clean, the only thing to suspect would be our `winapi` bindings.

feawilsch:
I'm not sure what you mean. I tried running it with debugger attached and RUST_LOG=info. 
The exceptions are along the lines of
    Unhandled exception at 0x000007FEF3EED370 (atidxx64.dll) in quad.exe: 0xC0000005: Access 
    violation reading location 0x0000000000000003

No relevant looking log messages.

kvark:
Could you post the full call stack?

feawilsch:
I noticed the call stacks are quite different depending on whether the program is called from the terminal or by double clicking.

Terminal: 

 	igd10iumd64.dll!00007ffb8b0ca2dd()	Unknown
 	igd10iumd64.dll!00007ffb8b214c09()	Unknown
 	igd10iumd64.dll!00007ffb8b213e42()	Unknown
 	igd10iumd64.dll!00007ffb8b20b35c()	Unknown
 	igd10iumd64.dll!00007ffb8b0d5388()	Unknown
 	d3d11.dll!00007ffb9232be8f()	Unknown
 	d3d11.dll!00007ffb9232b413()	Unknown
 	d3d11.dll!00007ffb9233700e()	Unknown
 	d3d11.dll!00007ffb9233d107()	Unknown
 	d3d11.dll!00007ffb9233cf73()	Unknown
 	d3d11.dll!00007ffb9233ca1c()	Unknown
 	d3d11_3SDKLayers.dll!00007ffb2fd65de5()	Unknown
 	d3d11_3SDKLayers.dll!00007ffb2fd6625a()	Unknown
 	d3d11_3SDKLayers.dll!00007ffb2fd4e366()	Unknown
[...]

PR #3118, opened 2020-01-04
[0.4/dx11] disable coherent memory, remove MemoryHeapFlags

kvark:
The coherency isn't working properly yet. It needs to be improved before we can turn it on again (see #3119).
There was also an error in buffer ranges, possibly related to #3114 
PR checklist:
- [x] tested examples with the following backends: quad


kvark:
I believe this can go in. It makes wgpu-rs actually work on DX11, in general.
cc @fkaa 
bors r+

PR #3122, opened 2020-01-05
[0.4/dx11] New binding model

### LINKED ISSUE #3121: Support read-only storage resources in DX11
In DX11 (and DX12), read-only storage becomes an SRV, while writable storage becomes an UAV. This means, we don't know what exact binding to use at the level of Vulkan pipeline layout. What we do in DX12 is having both SRV and UAV associated with a storage binding. This plays well "combined" descriptors, which also happen to have 2 different associated things with them. This task is about introducing a similarly generic model for dx11.

kvark:
Does most of the work for #3121 . Unfortunately, it *still* doesn't fix it, because SPIRV-Cross on HLSL lacks resource mapping overrides. So we are blocked on https://github.com/KhronosGroup/SPIRV-Cross/issues/1252 , but we can follow-up with a fix here once it's available, the fix is going to be small.

Status:
- [x] types and general structure
- [x] descriptor pool
- [x] binding descriptors
- [x] pipeline layout
- [x] shader decorations

Tested on wgpu-rs examples :tada: .

Compromises:
  - descriptor binding code responsible for coherent memory is removed, since it's tricky and we disabled coherent memory in #3120
  - copying descriptor sets is not implemented for now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+### backend-dx11-0.4.4, backend-dx11-0.4.5 (06-01-2020)
+  - disable coherent memory for being broken
+  - rewrite the binding model completely
+
 ### backend-metal-0.4.2 (18-11-2019)
   - fix missing iOS metallib
   - fix viewport/scissor after `clear_attachments` call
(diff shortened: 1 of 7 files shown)


kvark:
bors r+
hopefully we'll get some DX experts on the review board, some time

PR #3123, opened 2020-01-07
[dx11] new binding model

kvark:
Sibling of #3122 on master

This change *really* shines on master, because the `DescriptorType` enum is now more complex, and all the matching is concentrated in one single place instead of being spread thin over the code.


kvark:
bors r+
thanks `winit` team for releasing the stable!

PR #3124, opened 2020-01-07
[dx12] hold DXGI library alive

kvark:
Fixes the crash in `enumerate_adapters`. Thanks @jrmuizel for discovering it!
Note: a similar fix for DX11 went in #3096


kvark:
bors r=trivial

kvark:
bors retry

kvark:
bors retry

PR #3139, opened 2020-02-01
Move line width out of line polygon mode

krolli:
Line width state is not tied only to line polygon mode and storing it
there has correctness implications on Vulkan back-end. This is
problematic for gfx-portability when targeting Vulkan back-end, since it
can cause correct Vulkan application to trigger validation errors.

Prepares for fixing https://github.com/gfx-rs/portability/issues/206
PR checklist:


kvark:
(sorry for getting back late here, you caught me travelling)
bors r+

PR #3143, opened 2020-02-13
[vk] expect -1 returned by acquire_next_image

kvark:
Fixes #3141 (works around)
PR checklist:
- [ ] `make` succeeds (on *nix)
- [ ] `make reftests` succeeds
- [ ] tested examples with the following backends:
- [ ] `rustfmt` run on changed code

PR #3144, opened 2020-02-13
[0.4/vk] work around Intel swapchain bug

kvark:
Sibling of #3143


kvark:
bors r+
https://crates.io/crates/gfx-backend-vulkan/0.4.2

Issue #3133, opened 2020-01-28
OpenGL surface via surfman


kvark:
Just spoke to @pcwalton about https://github.com/pcwalton/surfman, and I'm wondering if we can use this for implementing the *true* surfaces on GL context, that don't require GL device. Seems like a direction worth exploring.

The tricky bit is that presenting the shared DXGI surface in WGL seems impossible at the moment (link to an issue missing?). I would be comfortable restricting our GL backend to non-windows platforms, or just reporting that a surface can't be presented to on Windows. We have enough of alternatives there (DX11/DX12/Vulkan), so we should be safe. Most value of OpenGL comes to Linux/Android.

Related to #1304

msiglreith:
I'm not totally sure where exactly this would fit in. Would you mind clarifying a bit.

Regarding OpenGL on Windows: Right now it should provide the same as Vulkan regarding swapchain/surface separation from devices as they are backed by different GL contexts (e.g in the quad example the `wgl` path: https://github.com/gfx-rs/gfx/blob/master/examples/quad/main.rs#L114)

kvark:
I'm not totally sure either. The way I see it that we'd use `surfman` for creating the surface  independently of `gfx_backend_gl::Device`. Then we'd glue it together with the device at `create_swapchain` and present on it. This should allow us to work with WSI with the same path we do on other backends.

msiglreith:
FWIW I just tried the `threads` example on a wayland machine but replacing `https://github.com/pcwalton/surfman/blob/master/surfman/examples/threads.rs#L98` with `Connection::new().unwrap()`, which causes the example to hang (without the change it works normally). This would simulate our use-case where the `Instance` creates at least one pure context.

The documentation seems to kinda imply that this might not work: https://github.com/pcwalton/surfman/blob/master/surfman/src/platform/unix/wayland/surface.rs#L32

pcwalton:
Note that the master branch is surfman 0.1, which is only lightly maintained. The new work is surfman 0.2 in the "multi" branch, which is a major overhaul. Once @asajeffrey finishes upgrading Servo to surfman 0.2 (a large task that has taken several weeks due to all the platforms Servo runs on), we will merge surfman 0.2 into master.

kvark:
@pcwalton does 0.2 changes anything with regards to this limitation?
> Surfaces are specific to a given context and cannot be rendered to from any context other than
the one they were created with.

What we really need is to have the surface to be rendered by a different context.

[...]

PR #3151, opened 2020-02-25
OpenGL surface via surfman

zicklag:
Fixes/Discussion: https://github.com/gfx-rs/wgpu/issues/450

PR checklist:
  - GL ( excluding compute because it isn't supported )

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -8,8 +8,7 @@ edition = "2018"
 [features]
 default = []
 metal = ["gfx-backend-metal"]
-gl = ["gfx-backend-gl", "gfx-backend-gl/glutin"]
-wgl = ["gfx-backend-gl", "gfx-backend-gl/wgl"]
+gl = ["gfx-backend-gl"]
 dx11 = ["gfx-backend-dx11"]
 dx12 = ["gfx-backend-dx12"]
 vulkan = ["gfx-backend-vulkan"]
@@ -28,20 +27,16 @@ name = "compute"
 path = "compute/main.rs"
 
 [dependencies]
-env_logger = "0.6"
 image = "0.21"
 log = "0.4"
 hal = { path = "../src/hal", version = "0.5", package = "gfx-hal" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.5" }
 winit = { version = "0.21.0", features = ["web-sys"] }
+gfx-backend-gl = { path = "../src/backend/gl", version = "0.5", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+env_logger = "0.6"
 glsl-to-spirv = "0.1.4"
-
-[dependencies.gfx-backend-gl]
-path = "../src/backend/gl"
-version = "0.5"
-optional = true
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 wasm-bindgen = "0.2.55"
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -30,9 +30,9 @@ path = "compute/main.rs"
 image = "0.21"
 log = "0.4"
 hal = { path = "../src/hal", version = "0.5", package = "gfx-hal" }
+gfx-backend-gl = { path = "../src/backend/gl", version = "0.5", optional = true }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.5" }
 winit = { version = "0.21.0", features = ["web-sys"] }
-gfx-backend-gl = { path = "../src/backend/gl", version = "0.5", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 env_logger = "0.6"
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -17,8 +17,9 @@ build = "build.rs"
 name = "gfx_backend_gl"
 
 [features]
-default = []
+default = ["wgl", "surfman", "surfman-x11"]
 wgl = []
+surfman-x11 = ["surfman/sm-x11"]
 
 [dependencies]
 arrayvec = "0.5"
@@ -32,9 +33,14 @@ parking_lot = "0.10"
 spirv_cross = { version = "0.18", features = ["glsl"] }
 lazy_static = "1"
 raw-window-handle = "0.3"
+glutin = { version = "0.23.0", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-glutin = { version = "0.23.0", optional = true }
+winit = "0.21.0"
+
+[target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
+# TODO: Update to released version when it comes out
+surfman = { git = "https://github.com/katharostech/surfman.git", branch = "gfx-support", features = ["sm-raw-window-handle"], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3.6"
@@ -65,3 +71,4 @@ winapi = { version = "0.3", features = ["libloaderapi", "windef", "wingdi", "win
 
 [build-dependencies]
 gl_generator = "0.11"
+cfg_aliases = "0.1.0"
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -37,7 +37,7 @@ glutin = { version = "0.23.0", optional = true }
 
 [target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
 # TODO: Update to released version when it comes out
-surfman = { git = "https://github.com/katharostech/surfman.git", branch = "gfx-support", features = ["sm-raw-window-handle"], optional = true }
+surfman = { git = "https://github.com/servo/surfman.git", rev = "41ac1ee", features = ["sm-raw-window-handle"], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3.6"
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -18,7 +18,7 @@ name = "gfx_backend_gl"
 
 [features]
 default = ["wgl", "surfman", "surfman-x11"]
-wgl = []
+wgl = ["winapi"]
 surfman-x11 = ["surfman/sm-x11"]
 
 [dependencies]
@@ -34,9 +34,6 @@ spirv_cross = { version = "0.18", features = ["glsl"] }
 lazy_static = "1"
 raw-window-handle = "0.3"
 glutin = { version = "0.23.0", optional = true }
-
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-winit = "0.21.0"
 
 [target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
 # TODO: Update to released version when it comes out
@@ -67,7 +64,7 @@ features = [
 ]
 
 [target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3", features = ["libloaderapi", "windef", "wingdi", "winuser"] }
+winapi = { version = "0.3", features = ["libloaderapi", "windef", "wingdi", "winuser"], optional = true }
 
 [build-dependencies]
 gl_generator = "0.11"
--- a/src/backend/gl/build.rs
+++ b/src/backend/gl/build.rs
@@ -2,11 +2,10 @@ use gl_generator::{Api, Fallbacks, Profile, Registry};
 use std::env;
 use std::fs::File;
 use std::path::PathBuf;
-use cfg_aliases::cfg_aliases;
 
 fn main() {
     // Setup cfg aliases
-    cfg_aliases! {
+    cfg_aliases::cfg_aliases! {
         // Platforms
         wasm: { target_arch = "wasm32" },
         android: { target_os = "android" },
@@ -15,8 +14,8 @@ fn main() {
         linux: { target_os = "linux" },
         // Backends
         surfman: { all(unix, feature = "surfman", not(ios)) },
-        glutin: { all(feature = "glutin", not(wasm)) },
         wgl: { all(windows, feature = "wgl") },
+        glutin: { all(feature = "glutin", not(any(wasm, surfman))) },
         dummy: { not(any(wasm, glutin, wgl, surfman)) },
     }
 
--- a/src/backend/gl/build.rs
+++ b/src/backend/gl/build.rs
@@ -2,8 +2,24 @@ use gl_generator::{Api, Fallbacks, Profile, Registry};
 use std::env;
 use std::fs::File;
 use std::path::PathBuf;
+use cfg_aliases::cfg_aliases;
 
 fn main() {
+    // Setup cfg aliases
+    cfg_aliases! {
+        // Platforms
+        wasm: { target_arch = "wasm32" },
+        android: { target_os = "android" },
+        macos: { target_os = "macos" },
+        ios: { target_os = "ios" },
+        linux: { target_os = "linux" },
+        // Backends
+        surfman: { all(unix, feature = "surfman", not(ios)) },
+        glutin: { all(feature = "glutin", not(wasm)) },
+        wgl: { all(windows, feature = "wgl") },
+        dummy: { not(any(wasm, glutin, wgl, surfman)) },
+    }
+
     let target = env::var("TARGET").unwrap();
     let dest = PathBuf::from(&env::var("OUT_DIR").unwrap());
 
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -3,10 +3,6 @@
 
 #![allow(missing_docs, missing_copy_implementations)]
 
-// Check for incompatible feature flags
-#[cfg(all(surfman, glutin))]
-compile_error!("You cannot specify both `surfman` and `glutin` features at the same time.");
-
 #[macro_use]
 extern crate bitflags;
 #[macro_use]
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -3,13 +3,18 @@
 
 #![allow(missing_docs, missing_copy_implementations)]
 
+// Check for incompatible feature flags
+#[cfg(all(surfman, glutin))]
+compile_error!("You cannot specify both `surfman` and `glutin` features at the same time.");
+
 #[macro_use]
 extern crate bitflags;
 #[macro_use]
 extern crate log;
 extern crate gfx_hal as hal;
-#[cfg(all(feature = "glutin", not(target_arch = "wasm32")))]
-pub extern crate glutin;
+
+#[cfg(surfman)]
+use parking_lot::RwLock;
 
 use std::cell::Cell;
 use std::fmt;
@@ -32,18 +37,29 @@ mod queue;
 mod state;
 mod window;
 
-#[cfg(all(not(target_arch = "wasm32"), feature = "glutin"))]
-pub use crate::window::glutin::{config_context, Headless, Surface, Swapchain};
-#[cfg(target_arch = "wasm32")]
+// Web implementation
+#[cfg(wasm)]
 pub use window::web::{Surface, Swapchain};
 
-#[cfg(all(feature = "wgl", not(target_arch = "wasm32")))]
-use window::wgl::DeviceContext;
+// Glutin implementation
+#[cfg(glutin)]
+pub use crate::window::glutin::{Instance, Surface, Swapchain};
 
-#[cfg(all(feature = "wgl", not(target_arch = "wasm32")))]
+// Surfman implementation
+#[cfg(surfman)]
+pub use crate::window::surfman::{Instance, Surface, Swapchain};
+// Helps windows detect discrete GPUs
+#[cfg(surfman)]
+surfman::declare_surfman!();
+
+// WGL implementation
+#[cfg(wgl)]
+use window::wgl::DeviceContext;
+#[cfg(wgl)]
 pub use window::wgl::{Instance, Surface, Swapchain};
 
-#[cfg(not(any(target_arch = "wasm32", feature = "glutin", feature = "wgl")))]
+// Catch-all dummy implementation
+#[cfg(dummy)]
 pub use window::dummy::{Surface, Swapchain};
 
 pub use glow::Context as GlContext;
@@ -53,15 +69,23 @@ type ColorSlot = u8;
 
 pub(crate) struct GlContainer {
     context: GlContext,
+
+    #[cfg(surfman)]
+    surfman_device: Starc<RwLock<surfman::Device>>,
+
+    #[cfg(surfman)]
+    surfman_context: Starc<RwLock<surfman::Context>>,
 }
 
 impl GlContainer {
-    #[cfg(feature = "glutin")]
+    #[cfg(not(wasm))]
     fn make_current(&self) {
-        // Unimplemented
+        // TODO:
+        // NOTE: Beware, calling this on dereference breaks the surfman backend
+        // I'm not sure if there would be similar concequences with other backends.
     }
 
-    #[cfg(not(target_arch = "wasm32"))]
+    #[cfg(any(glutin, wgl))]
     fn from_fn_proc<F>(fn_proc: F) -> GlContainer
     where
         F: FnMut(&str) -> *const std::os::raw::c_void,
@@ -70,7 +94,24 @@ impl GlContainer {
         GlContainer { context }
     }
 
-    #[cfg(target_arch = "wasm32")]
+    #[cfg(surfman)]
+    fn from_fn_proc<F>(
+        fn_proc: F,
+        surfman_device: Starc<RwLock<surfman::Device>>,
+        surfman_context: Starc<RwLock<surfman::Context>>,
+    ) -> GlContainer
+    where
+        F: FnMut(&str) -> *const std::os::raw::c_void,
+    {
+        let context = glow::Context::from_loader_function(fn_proc);
+        GlContainer {
+            context,
+            surfman_device,
+            surfman_context,
+        }
+    }
+
+    #[cfg(wasm)]
     fn from_canvas(canvas: &web_sys::HtmlCanvasElement) -> GlContainer {
         let context = {
             use wasm_bindgen::JsCast;
@@ -102,9 +143,20 @@ impl GlContainer {
 impl Deref for GlContainer {
     type Target = GlContext;
     fn deref(&self) -> &GlContext {
-        #[cfg(all(feature = "glutin", not(target_arch = "wasm32")))]
+        #[cfg(not(wasm))]
         self.make_current();
         &self.context
+    }
+}
+
+#[cfg(surfman)]
+impl Drop for GlContainer {
+    fn drop(&mut self) {
+        // Contexts must be manually destroyed to prevent a panic
+        self.surfman_device
+            .read()
+            .destroy_context(&mut self.surfman_context.write())
+            .expect("TODO");
     }
 }
 
@@ -112,13 +164,13 @@ impl Deref for GlContainer {
 pub enum Backend {}
 
 impl hal::Backend for Backend {
-    #[cfg(any(all(not(target_arch = "wasm32"), feature = "glutin"), feature = "wgl"))]
+    #[cfg(not(any(wasm, dummy)))]
     type Instance = Instance;
 
-    #[cfg(all(target_arch = "wasm32", not(feature = "wgl")))]
+    #[cfg(wasm)]
     type Instance = Surface;
 
-    #[cfg(not(any(target_arch = "wasm32", feature = "glutin", feature = "wgl")))]
+    #[cfg(dummy)]
     type Instance = DummyInstance;
 
     type PhysicalDevice = PhysicalDevice;
@@ -183,6 +235,7 @@ impl Error {
     }
 }
 
+#[cfg(not(wasm))]
 fn debug_message_callback(source: u32, gltype: u32, id: u32, severity: u32, message: &str) {
     let source_str = match source {
         glow::DEBUG_SOURCE_API => "API",
@@ -397,7 +450,7 @@ unsafe impl<T: ?Sized> Sync for Wstarc<T> {}
 #[derive(Debug)]
 pub struct PhysicalDevice(Starc<Share>);
 
-#[cfg(any(target_arch = "wasm32", not(feature = "wgl")))]
+#[cfg(not(wgl))]
 type DeviceContext = ();
 
 impl PhysicalDevice {
@@ -595,7 +648,7 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
         // initialize permanent states
         let gl = &self.0.context;
 
-        #[cfg(not(target_arch = "wasm32"))]
+        #[cfg(not(wasm))]
         {
             if cfg!(debug_assertions) && gl.supports_debug() {
                 gl.enable(glow::DEBUG_OUTPUT);
@@ -714,10 +767,10 @@ impl q::QueueFamily for QueueFamily {
     }
 }
 
-#[cfg(not(any(target_arch = "wasm32", feature = "glutin", feature = "wgl")))]
+#[cfg(dummy)]
 pub struct DummyInstance;
 
-#[cfg(not(any(target_arch = "wasm32", feature = "glutin", feature = "wgl")))]
+#[cfg(dummy)]
 impl hal::Instance<Backend> for DummyInstance {
     fn create(_: &str, _: u32) -> Result<Self, hal::UnsupportedBackend> {
         unimplemented!()
@@ -733,38 +786,6 @@ impl hal::Instance<Backend> for DummyInstance {
     }
     unsafe fn destroy_surface(&self, _surface: Surface) {
         unimplemented!()
-    }
-}
-
-#[cfg(all(feature = "glutin", not(target_arch = "wasm32")))]
-#[derive(Debug)]
-pub enum Instance {
-    Headless(Headless),
-    Surface(Surface),
-}
-
-#[cfg(all(feature = "glutin", not(target_arch = "wasm32")))]
-impl hal::Instance<Backend> for Instance {
-    fn create(name: &str, version: u32) -> Result<Instance, hal::UnsupportedBackend> {
-        Headless::create(name, version).map(Instance::Headless)
-    }
-
-    fn enumerate_adapters(&self) -> Vec<adapter::Adapter<Backend>> {
-        match self {
-            Instance::Headless(instance) => instance.enumerate_adapters(),
-            Instance::Surface(instance) => instance.enumerate_adapters(),
-        }
-    }
-
-    unsafe fn create_surface(
-        &self,
-        _: &impl raw_window_handle::HasRawWindowHandle,
-    ) -> Result<Surface, hal::window::InitError> {
-        unimplemented!()
-    }
-
-    unsafe fn destroy_surface(&self, _surface: Surface) {
-        // TODO: Implement Surface cleanup
     }
 }
 
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -40,6 +40,10 @@ pub use window::web::{Surface, Swapchain};
 // Glutin implementation
 #[cfg(glutin)]
 pub use crate::window::glutin::{Instance, Surface, Swapchain};
+#[cfg(glutin)]
+pub use glutin;
+#[cfg(glutin)]
+pub use crate::window::glutin::config_context;
 
 // Surfman implementation
 #[cfg(surfman)]
--- a/src/backend/gl/src/window/surfman.rs
+++ b/src/backend/gl/src/window/surfman.rs
@@ -59,11 +59,9 @@ impl Instance {
     fn get_default_context_attributes() -> sm::ContextAttributes {
         sm::ContextAttributes {
             version: sm::GLVersion::new(3, 3), // TODO: Figure out how to determine GL version
-            // TODO: Determine flags to provide. At least ALPH I think, but probably all of them.
-            // skipping COMPATIBILITY_PROFILE for now, because it panics with a TODO.
-            flags: sm::ContextAttributeFlags::ALPHA
-                | sm::ContextAttributeFlags::DEPTH
-                | sm::ContextAttributeFlags::STENCIL,
+            // TODO: Skipping COMPATIBILITY_PROFILE for now, because it panics with a TODO, but
+            // that is probably something we want to provide later.
+            flags: sm::ContextAttributeFlags::ALPHA,
         }
     }
 
@@ -186,10 +184,7 @@ impl hal::Instance<B> for Instance {
         Ok(self.create_surface_from_rwh(has_handle.raw_window_handle()))
     }
 
-    unsafe fn destroy_surface(&self, surface: Surface) {
-        // Surface implments Drop and will clean up the surface when it gets dropped
-        drop(surface);
-    }
+    unsafe fn destroy_surface(&self, _surface: Surface) {}
 }
 
 #[derive(Debug)]
--- a/src/backend/gl/src/window/surfman.rs
+++ b/src/backend/gl/src/window/surfman.rs
@@ -136,9 +136,8 @@ impl hal::Instance<B> for Instance {
 
         for surfman_adapter in &[
             &self.hardware_adapter,
-            // TODO: Enabling these causes a segfault for an unknown reason
-            // &self.low_power_adapter,
-            // &self.software_adapter,
+            &self.low_power_adapter,
+            &self.software_adapter,
         ] {
             // Create a surfman device
             let mut device =
@@ -151,6 +150,8 @@ impl hal::Instance<B> for Instance {
 
             // Create context
             let context = device.create_context(&context_descriptor).expect("TODO");
+            // Make context current
+            device.make_context_current(&context).expect("TODO");
 
             // Wrap in Starc<RwLock<T>>
             let context = Starc::new(RwLock::new(context));
@@ -196,6 +197,14 @@ pub struct Surface {
 }
 
 impl Surface {
+    /// Make the surface's gl context the current context
+    pub fn make_context_current(&self) {
+        self.device
+            .write()
+            .make_context_current(&self.context.read())
+            .expect("TODO");
+    }
+
     pub fn context(&self) -> Starc<RwLock<sm::Context>> {
         self.context.clone()
     }
@@ -241,7 +250,6 @@ impl window::PresentationSurface<B> for Surface {
         config: window::SwapchainConfig,
     ) -> Result<(), window::CreationError> {
         let gl = &device.share.context;
-        // let surface_info = self.device.read().surface_info(&self.surface.read());
 
         if let Some(old) = self.swapchain.take() {
             for fbo in old.fbos {
@@ -304,6 +312,7 @@ impl window::PresentationSurface<B> for Surface {
 
 impl window::Surface<B> for Surface {
     fn supports_queue_family(&self, _: &QueueFamily) -> bool {
+        self.make_context_current();
         true
     }
 
--- a/src/backend/gl/src/window/surfman.rs
+++ b/src/backend/gl/src/window/surfman.rs
@@ -1,0 +1,342 @@
+//! [Surfman](https://github.com/pcwalton/surfman)-based OpenGL backend for GFX-hal
+
+use crate::{conv, native, Backend as B, Device, GlContainer, PhysicalDevice, QueueFamily, Starc};
+use hal::{adapter::Adapter, format as f, image, window};
+
+use arrayvec::ArrayVec;
+use glow::HasContext;
+use parking_lot::RwLock;
+use surfman as sm;
+
+use std::cell::RefCell;
+use std::fmt;
+use std::iter;
+
+#[derive(Debug)]
+pub struct Swapchain {
+    // Underlying window, required for presentation
+    pub(crate) context: Starc<RwLock<sm::Context>>,
+    // Extent because the window lies
+    pub(crate) extent: window::Extent2D,
+    ///
+    pub(crate) fbos: ArrayVec<[native::RawFrameBuffer; 3]>,
+}
+
+impl window::Swapchain<B> for Swapchain {
+    unsafe fn acquire_image(
+        &mut self,
+        _timeout_ns: u64,
+        _semaphore: Option<&native::Semaphore>,
+        _fence: Option<&native::Fence>,
+    ) -> Result<(window::SwapImageIndex, Option<window::Suboptimal>), window::AcquireError> {
+        // TODO: sync
+        Ok((0, None))
+    }
+}
+
+thread_local! {
+    /// The thread-local surfman connection
+    static SM_CONN: RefCell<sm::Connection> =
+        RefCell::new(sm::Connection::new().expect("TODO"));
+}
+
+pub struct Instance {
+    hardware_adapter: sm::Adapter,
+    // TODO: We're not using these yet, but leave them here for later
+    #[allow(dead_code)]
+    low_power_adapter: sm::Adapter,
+    #[allow(dead_code)]
+    software_adapter: sm::Adapter,
+}
+
+impl fmt::Debug for Instance {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Instance").field(&["Adapter..."; 3]).finish()
+    }
+}
+
+impl Instance {
+    fn get_default_context_attributes() -> sm::ContextAttributes {
+        sm::ContextAttributes {
+            version: sm::GLVersion::new(3, 3), // TODO: Figure out how to determine GL version
+            // TODO: Determine flags to provide. At least ALPH I think, but probably all of them.
+            // skipping COMPATIBILITY_PROFILE for now, because it panics with a TODO.
+            flags: sm::ContextAttributeFlags::ALPHA
+                | sm::ContextAttributeFlags::DEPTH
+                | sm::ContextAttributeFlags::STENCIL,
+        }
+    }
+
+    pub unsafe fn create_surface_from_rwh(
+        &self,
+        raw_handle: raw_window_handle::RawWindowHandle,
+    ) -> Surface {
+        // Get context attributes
+        let context_attributes = Self::get_default_context_attributes();
+
+        // Open a device for the surface
+        // TODO: Assume hardware adapter
+        let mut device = SM_CONN
+            .with(|c| c.borrow().create_device(&self.hardware_adapter))
+            .expect("TODO");
+
+        // Create context descriptor
+        let context_descriptor = device
+            .create_context_descriptor(&context_attributes)
+            .expect("TODO");
+
+        // Create context
+        let mut context = device.create_context(&context_descriptor).expect("TODO");
+
+        // Create the surface with the context
+        let surface = device
+            .create_surface(
+                &context,
+                surfman::SurfaceAccess::GPUOnly,
+                surfman::SurfaceType::Widget {
+                    // Create a native widget for the raw window handle
+                    native_widget: SM_CONN.with(|c| {
+                        c.borrow()
+                            .create_native_widget_from_rwh(raw_handle)
+                            .expect("TODO")
+                    }),
+                },
+            )
+            .expect("TODO");
+
+        // Bind surface to context
+        device
+            .bind_surface_to_context(&mut context, surface)
+            .expect("TODO");
+
+        device.make_context_current(&context).expect("TODO");
+
+        // Create a surface with the given context
+        Surface {
+            renderbuffer: None,
+            swapchain: None,
+            context: Starc::new(RwLock::new(context)),
+            device: Starc::new(RwLock::new(device)),
+        }
+    }
+}
+
+impl hal::Instance<B> for Instance {
+    fn create(_: &str, _: u32) -> Result<Self, hal::UnsupportedBackend> {
+        Ok(Instance {
+            hardware_adapter: SM_CONN.with(|c| c.borrow().create_hardware_adapter().expect("TODO")),
+            low_power_adapter: SM_CONN
+                .with(|c| c.borrow().create_low_power_adapter().expect("TODO")),
+            software_adapter: SM_CONN.with(|c| c.borrow().create_software_adapter().expect("TODO")),
+        })
+    }
+
+    fn enumerate_adapters(&self) -> Vec<Adapter<B>> {
+        let mut adapters = Vec::with_capacity(3);
+
+        let context_attributes = Self::get_default_context_attributes();
+
+        for surfman_adapter in &[
+            &self.hardware_adapter,
+            // TODO: Enabling these causes a segfault for an unknown reason
+            // &self.low_power_adapter,
+            // &self.software_adapter,
+        ] {
+            // Create a surfman device
+            let mut device =
+                SM_CONN.with(|c| c.borrow().create_device(surfman_adapter).expect("TODO"));
+
+            // Create context descriptor
+            let context_descriptor = device
+                .create_context_descriptor(&context_attributes)
+                .expect("TODO");
+
+            // Create context
+            let context = device.create_context(&context_descriptor).expect("TODO");
+
+            // Wrap in Starc<RwLock<T>>
+            let context = Starc::new(RwLock::new(context));
+            let context_ = context.clone();
+            let device = Starc::new(RwLock::new(device));
+            let device_ = device.clone();
+
+            // Create gl container
+            let gl = GlContainer::from_fn_proc(
+                |symbol_name| {
+                    device_
+                        .write()
+                        .get_proc_address(&context_.read(), symbol_name)
+                        as *const _
+                },
+                device,
+                context,
+            );
+
+            // Create physical device
+            adapters.push(PhysicalDevice::new_adapter((), gl));
+        }
+
+        adapters
+    }
+
+    unsafe fn create_surface(
+        &self,
+        has_handle: &impl raw_window_handle::HasRawWindowHandle,
+    ) -> Result<Surface, window::InitError> {
+        Ok(self.create_surface_from_rwh(has_handle.raw_window_handle()))
+    }
+
+    unsafe fn destroy_surface(&self, surface: Surface) {
+        // Surface implments Drop and will clean up the surface when it gets dropped
+        drop(surface);
+    }
+}
+
+#[derive(Debug)]
+pub struct Surface {
+    pub(crate) swapchain: Option<Swapchain>,
+    pub(crate) context: Starc<RwLock<sm::Context>>,
+    device: Starc<RwLock<sm::Device>>,
+    renderbuffer: Option<native::Renderbuffer>,
+}
+
+impl Surface {
+    pub fn context(&self) -> Starc<RwLock<sm::Context>> {
+        self.context.clone()
+    }
+
+    fn swapchain_formats(&self) -> Vec<f::Format> {
+        // TODO: Make sure this is correct. I believe it is. Reference:
+        // https://github.com/pcwalton/surfman/blob/master/surfman/src/context.rs#L34-L37
+        vec![f::Format::Rgba8Srgb, f::Format::Bgra8Srgb]
+    }
+}
+
+impl Drop for Surface {
+    fn drop(&mut self) {
+        // Unbind and get the underlying surface from the context
+        let surface = self
+            .device
+            .read()
+            .unbind_surface_from_context(&mut self.context.write())
+            .expect("TODO");
+
+        if let Some(mut surface) = surface {
+            // Destroy the underlying surface
+            self.device
+                .read()
+                .destroy_surface(&mut self.context.write(), &mut surface)
+                .expect("TODO");
+        }
+
+        // Destroy the backing context
+        self.device
+            .read()
+            .destroy_context(&mut self.context.write())
+            .expect("TODO");
+    }
+}
+
+impl window::PresentationSurface<B> for Surface {
+    type SwapchainImage = native::ImageView;
+
+    unsafe fn configure_swapchain(
+        &mut self,
+        device: &Device,
+        config: window::SwapchainConfig,
+    ) -> Result<(), window::CreationError> {
+        let gl = &device.share.context;
+        // let surface_info = self.device.read().surface_info(&self.surface.read());
+
+        if let Some(old) = self.swapchain.take() {
+            for fbo in old.fbos {
+                gl.delete_framebuffer(fbo);
+            }
+        }
+
+        if self.renderbuffer.is_none() {
+            self.renderbuffer = Some(gl.create_renderbuffer().unwrap());
+        }
+
+        let desc = conv::describe_format(config.format).unwrap();
+        gl.bind_renderbuffer(glow::RENDERBUFFER, self.renderbuffer);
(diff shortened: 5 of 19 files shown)


src/backend/gl/src/lib.rs
```diff
@@ -736,6 +738,83 @@ pub enum Instance {
...
+impl Instance {
+    
+    pub fn create_surface_from_wayland(
+        &self,
+        display: *mut c_void,
+        surface: *mut c_void,
+    ) -> Surface {
+        log::trace!("Creating GL surface from wayland");
+        let context = unsafe {
+            glutin::ContextBuilder::new()
```
kvark: the way it would ideally work is: we create a "pure" GL context that is not associated to any surface, for `Instance`. Then physical devices and logical devices, queues, are all just references to this main context.

When a surface is created (here), we'd need to create a GL context per surface, but in a way that shares all resources with the main context. https://docs.rs/glium/0.19.0/glium/glutin/struct.ContextBuilder.html#method.with_shared_lists

This scheme assumes that we can share resources and drivers will not freak out, and that we can create a pure context for the `Instance`.

zicklag: OK, so, if I understand correctly, we want to create a global context once, that we will then pass into future context creations with `with_shared_lists(global_context)`. That way every time you create a surface, it shares the resources with the global context.

So could we do that with something like ( in concept ):

```rust
lazy_static! {
    static ref global_context: Arc<Mutex<Option<Context>>> = Arc::new(Mutex::new(None);
}

//...

fn create_surface() {
    let global_context = global_context.lock().unwrap()
    let shared_ctx = if let Some(context) = global_context {
        context
    } else {
        *global_context = Some(glutin::ContextBuilder::new()
            .with_vsync(true));
    };

    glutin::ContextBuilder::new().with_shared_lists(shared_ctx);
}
```

kvark: Yes, the global context would be a part of the `Instance`, it doesn't need to be in a static :)

zicklag: Most of the functions to create a context require an event loop. The instance could create an event loop when it is created, but there is a limitation of the wayland backend that requires that all shared contexts use the same event loop. That would be a problem because the event loop is a part of the window that is passed in to `create_surface()` and isn't available when we create the instance.

Creating a context with [`build_osmesa()`](https://docs.rs/glutin/0.23.0/glutin/platform/unix/trait.HeadlessContextExt.html#tymethod.build_osmesa) doesn't require an event loop, but I don't know if you would still have the same problem on wayland.

kvark: The *idea*, innocent in its purity and naivety, is that the root GL context (associated with the Instance) is a *pure* GL context, meaning that it's not associated to any window, and therefore it shouldn't need any events whatsoever.
When I was looking into it, I ended up prototyping support for pure GL contexts in https://github.com/rust-windowing/glutin/pull/942 . It wasn't merged, and I don't know how the situation really advanced since then. Figuring this out, I think, is one of the core puzzle pieces in the whole GL support story, although there might be other approaches too.

`build_osmesa()` isn't helpful for HW acceleration - it's a software context, it can't share resource with real ones.

One (temporary?) workaround could be to have a GL-specific way of creating an `Instance`, where the user would provide the event loop. Yes, it's an easy to miss invariant, but it would work for most cases, and could be "good enough". That would make the instance creation diverging from the other backends, but at least everything else will be the same,

zicklag: > build_osmesa() isn't helpful for HW acceleration - it's a software context, it can't share resource with real ones.

Got it, that makes sense. I always wondered what mesa was. Now I think I get it: a software implementation of of GL.

> One (temporary?) workaround could be to have a GL-specific way of creating an Instance, where the user would provide the event loop.

If we lazily initialized the global context by waiting until the first call to `create_surface()` where we *would* have an event loop from the passed in window, we wouldn't even have to change the API. I think I'll start with that and leave a doc comment explaining it and that we might want to change it.

kvark: I don't think going lazy is an option here: we'll be asked to enumerate adapters, provide limits and supported extensions, create devices and resources, etc, all before creating a surface. Heck, there are applications that don't even need a surface.

zicklag: OK, I see it now. I missed the fact that `Headless` and `Surface` had a context in them. I'll look around a little bit for a way to make a "Pure" context and if I can't come up with anything on that front I'll try a temporary alternative way to create a GL instance that requires an event loop.

zicklag: I think I see now where `surfman` could be used. I just found the examples on docs.rs, which don't seem to be on the master branch anymore.

It looks like it might actually be really simple to use `surfman` to create the global context, without requiring a window handle. I'm not sure if it does some magic to handle the context sharing on Wayland or not, but its a start, and we could ask the `surfman` maintainer about that.

I'll try it out and see if that works.

kvark: Thank you for looking into this! There is a lot of context here, and grasping the situation is the main factor of why the problem is still unsolved. You seem to understand the full picture now.

zicklag:
If surfman works do you think we remove the `glutin` feature and make it the only option? If surfman is the only way to support the desired API, we probably want to.

I'm asking because I think I might take out the glutin code just to help me think while understanding the code and avoid a bunch of feature gates, even if I add it in again afterwards.

zicklag:
Another question: for the `Headless` instance type, should `create_surface` completely ignore the window handle and create a headless surface? That's what I'm assuming the intent is, but I couldn't be sure because it still receives a raw window handle. Is that just to be compatible with the non-headless API?

kvark:
@zicklag I haven't investigated `surfman` in depth. My understanding was that it doesn't work on Windows in the way we want (no resource sharing). Also, if the `glutin` path *can* work in the way we want (with pure context for the instance), it would still be nice to keep supporting it. So all in all, I would prefer the `glutin` path to stay, even if it's only partially usable today.

> for the Headless instance type

I don't think there should be a headless instance type. Ideally, we'd expose it at the physical device level. One physical device would be "software", while another will be "hardware". When the user creates a surface, they don't know about a physical device, but the API clearly has means to express that different physical devices can work with different surfaces via [supports_queue_family](https://docs.rs/gfx-hal/0.4.1/gfx_hal/window/trait.Surface.html#tymethod.supports_queue_family) (each physical device would only expose one queue family, and you can make it so the headless physical device is only compatible with headless surfaces. I.e. the only point of divergence on the API level would be the way a surface is created: we could have `create_emulated_surface` versus another `create_surface` for accelerated surfaces.

zicklag:
> So all in all, I would prefer the glutin path to stay, even if it's only partially usable today.

Sounds good.

>  Ideally, we'd expose it at the physical device level. One physical device would be "software", while another will be "hardware".

I think I get it. I'll see about that then. I'll try to get non-headless to work first.

---

Also, what is `wgl`? Is that for windows?

kvark:
WGL is basically window system interface in Windows for OpenGL. It's what you use to set up the GL context (internally by glutin, or by other means).

zicklag:
So ideally would we want to support `glutin` and `wgl` in separate feature flags, or just rely on `glutin` to do it?

kvark:
Overall, Windows is least of a priority for GL (so just because `surfman` has problems with it doesn't mean we can't technically leave it as the only choice): since on Windows we also have D3D11, D3D12, and Vulkan. I would be totally fine if our GL backend was simply not available on Windows, or expose only compute abilities, or had only software implementation.

For your work, try to focus on Linux & Android first and foremost, then macOS/iOS, and only afterwards - Windows. No need to consider a separate WGL code path at this point :)

src/backend/gl/src/lib.rs
```diff
@@ -32,34 +41,60 @@ mod queue;
...
-
 pub(crate) struct GlContainer {
     context: GlContext,
+
+    /// In order to set the current context we must have access to the instance
+    #[cfg(feature = "surfman")]
+    surfman_data: Option<(
+        Starc<RwLock<surfman::Device>>,
+        Starc<RwLock<surfman::Context>>,
+    )>,
```
zicklag: @kvark Is there any problem, do you think, with using `RwLock`'s to access the surfman `Device` and `Context` here?

The issue is that, in order to actually make the context current, you have to have mutable access to the surfman Device. Also, in the `Instance` you need mutable access to the Device to create contexts and surfaces. Locking was the only way I could think to allow both the `GLContainer` and the `Instance` have mutable access to the `Device` and `Context`.

zicklag:
I think I have the `surfman` backend mostly in the state that I want it and it *compiles*, but I haven't yet gotten it to finish linking so I haven't been able to run it yet. I'm assuming that the problem has to do with my build environment, but I'll have to look more into it.

src/backend/gl/src/lib.rs
kvark: It should be fine, although we'll likely need `Arc` instead of `Starc` here, depending on the `Sync` guarantees of `surfman`. The most frequent operations in the API are issuing commands within render passes. Since GL doesn't have command buffers, we are recording the commands "in software", hence we don't need to access the GL context during this hot code. We'll need to access it only when doing device-level operations or submitting command buffers, which are not as frequent, and we can totally live with an `RwLock`.

kvark:
Perhaps, we need to call `set_instance` somewhere? :)

src/backend/gl/src/lib.rs
zicklag: The `surfman::Context` and `Device` are `!Sync` and `!Send` so `Arc` didn't work. Will that be an issue.

zicklag:
Oops, I had that in there earlier! I lost it in a refactor. :)

src/backend/gl/src/lib.rs
kvark: Hmm, it's certainly not as good as we want, but the main thing to address here is the divergence with the API. Having extra constraints is fine.

zicklag:
Do you have any idea what's wrong if I'm getting undefined symbols while linking X11:

```
 = note: ld: error: undefined symbol: XOpenDisplay
          >>> referenced by connection.rs:79 (/home/zicklag/git/other/surfman/surfman/src/platform/unix/x11/connection.rs:79)
          >>>               gfx_backend_gl-d96c34389fc1283e.29fwo7artum5a8yx.rcgu.o:(surfman::platform::unix::x11::connection::Connection::new::h891448e5eadb68ed) in archive /home/zicklag/git/other/wgpu-rs/target/debug/deps/libgfx_backend_gl-d96c34389fc1283e.rlib
....
```

And more like that.

I've got a weird setup right now because I'm not an admin on the system I'm using to build, but it doesn't have `libx11-dev` installed, so to remedy that I downloaded the `libx11-dev` package source and copied the headers into `~/.local/include/x11` and wrote a `x11.pc` that includes it.

Apparently I did something wrong, but what does the undefined symbol mean? Does that mean I'm not linking X11 in? Like I'm missing a `-l` flag to the linker?

kvark:
Yes definitely missing a link to x11 library.

zicklag:
Ah, thanks. Got it to by adding `-lX11 -L$HOME/.local/lib` to `RUSTFLAGS` and symlinking my `/usr/lib/libX11.so.6` to `$HOME/.local/libX11.so`. I keep forgetting that `-lsomelib` won't look at `libsomelib.so.suffix`.

I don't know why it didn't "Just Work". Maybe `surfman` doesn't use `pkg-config` or something.

Anyway, now I've got a panic in `surfman` to track down, but it's better than a linker error. :wink:

zicklag:
Hey, we're getting close. You have any idea what this means?

```
[2020-03-02T03:56:52Z ERROR gfx_backend_gl] [API/Error] ID 1 : GL_INVALID_FRAMEBUFFER_OPERATION in glBlitFramebuffer(incomplete draw/read buffers)
[2020-03-02T03:56:52Z TRACE wgpu_core::command::allocator] recycling comb submitted in 3 when 4 is lowest active
[2020-03-02T03:56:52Z TRACE wgpu_core::command::render] Encoding render pass begin in command buffer (0, 4, Gl)
[2020-03-02T03:56:52Z DEBUG gfx_backend_gl::device] create_framebuffer [Renderbuffer(1)]
thread 'main' panicked at 'Error creating FBO: InvalidFramebufferOperation for RenderPass { attachments: [Attachment { format: Some(Bgra8Srgb), samples: 1, ops: AttachmentOps { load: Clear, store: Store }, stencil_ops: AttachmentOps { load: DontCare, store: DontCare }, layouts: Undefined..Present }], subpasses: [SubpassDesc { color_attachments: [0], depth_stencil: None }] }'
```

The panic is from [here](https://github.com/gfx-rs/gfx/blob/106cb216b4fc86663783559ae7a45a88608210b0/src/backend/gl/src/device.rs#L1035).

kvark:
There is a blit happening, and it complains that one of the framebuffers is incomplete. First of all, I would assume that your code path does *not* need a blit. With your surface implementation, rendering to a swapchain should be direct, no copies/blits involved.

zicklag:
OK, I found out where the blit is:

https://github.com/gfx-rs/gfx/blob/106cb216b4fc86663783559ae7a45a88608210b0/src/backend/gl/src/queue.rs#L219-L234

It is probably erroring because I do not have a swap buffers call for `surfman` and I don't know how to do that. I haven't looked into it yet.

But, you did say that you don't think there should be any copying anyway. Is there a way around using the blit in this case?

kvark:
Here is how OpenGL allows one to render a frame: you draw framebuffer(0), and you render, done.

Here is how gfx-rs allows one to render a frame: you get the next image, then you have/create a view into it, then you create a framebuffer that includes that view. That framebuffer can include other views as well, such as your manially depth texture. Then you execute a render pass that renders into that framebuffer.

If we don't blit, we would require that if a framebuffer contains the swapchain image (in GL), it has to have nothing else. This is not what users expect... So instead, we create a render buffer internally, render into it, and then when it's time to present - we blit into the GL framebuffer(0).

That blit is costy and *highly* unfortunate. However, `surfman` approach gives us hope. Looking at https://docs.rs/surfman/0.1.3/surfman/platform/unix/x11/surface/struct.SurfaceTexture.html for example, I see that `gl_texture()` just returns a texture ID. That means we can expose it to the users as regular `Backend::Image` types, and have framebuffers mixing them with other images. Which in turn means that we can finally remove the blit.

Needless to say, you are unravelling one hack in GL after another, and it's sadly all connected. Requires heroic effort to defeat, but Rust community will be very happy :)

zicklag:
Thanks for the explanation and the help. I'm glad we've gotten this far. :)

I'll try to process that and look around and show you what I come up with.

kvark:
@zicklag how is this going? Facing any troubles?

zicklag:
Hey @kvark, thanks for checking back. I actually hadn't had a chance to look at it until last night. I haven't run into any problems so to speak, but I really don't have a grip on the OpenGL API or the Vulkan/GFX API yet. I'm a little confused at where to start.

I did have a question, though. In this code:

```rust
            gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(swapchain.fbos[index as usize]));
            gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, None);
            gl.blit_framebuffer(
                0,
                0,
                extent.width as _,
                extent.height as _,
                0,
                0,
                extent.width as _,
                extent.height as _,
                glow::COLOR_BUFFER_BIT,
                glow::LINEAR,
            );
```

The point is that you are setting our framebuffer from the swapchain as the buffer we are reading from and you set the write or "DRAW" buffer to `None` which means framebuffer(0)? Then the blit will blit from our framebuffer and to framebuffer(0)?

So how are we going to get around the blit? By creating a surface with `surfman` and then rendering to that surface instead of framebuffer(0), something that we can only do because surfman abstracts over surfaces and allows us to present surfaces and not just the result of rendering framebuffer(0)? This is a little fuzzy to me. :)

kvark:
@zicklag 

>"DRAW" buffer to None which means framebuffer(0)

In `glow`, binding `None` means the default framebuffer, so the one associated with the window.

Would be great if you could hop on "#gfx:matrix.org" room instead of here, given the amount of back-and-forth discussion.

> So how are we going to get around the blit?

I can describe what I thought could work, before @msiglreith said in https://github.com/gfx-rs/gfx/issues/3133#issuecomment-596108857 that this wouldn't work...
I thought that instead of creating casual renderbuffers and pretending they are OpenGL swapchain (see [this code](https://github.com/gfx-rs/gfx/blob/033850f6a3fe2d8399eae3f998548ee61ade5f03/src/backend/gl/src/window/glutin.rs#L181)), and then blitting from them to the default framebuffer, we could work with the GL textures representing Surfman surfaces directly and give them to the user in `acquire_image`, without any blitting to follow up.

I asked @pcwalton for clarification in https://github.com/gfx-rs/gfx/issues/3133#issuecomment-596164336 . In any way, it sounds like we should be using the latest `surfman` from git master instead of 0.1, and then switch to 0.2 when it's published. Perhaps, that would let us get further?

It may be that we still need a blit. Even if it's the case, it's not worse than what we have. The main goal of your work still stands - to align the OpenGL API with the others.

zicklag:
OK, I think I get it a bit more now. I already ended up switching to the git master `surfman` because I had a PR for it and based the PR off of master.

> Would be great if you could hop on "#gfx:matrix.org" room instead of here, given the amount of back-and-forth discussion.

I can't get on chat because of network restrictions where I'm at so unfortunately we're stuck here. Thanks for sticking with me so far, though. :grin: 

I'll probably try to see if I can fix the error we are getting when the blit *is* and also look around at `acquire_image`.

zicklag:
After trying to understand surfman a bit more I realized that because devices and contexts are thread-local, the way that I was using a `surfman::Device` created by the `Instance` meant that you wouldn't be able to create surfaces from other threads without violating the `Starc`.

I just pushed an update that makes the `Instance` have no references to any thread-local information and instead creates the devices, surfaces, and contexts, on-demand when they are requested, in the thread that they are requested in. If I did this right, this *should* mean that you can create surfaces in different threads from each-other, but you can't send them over threads or you will get a panic from the `Starc`.

Now I'm going to look into the update for `acquire_image`.

kvark:
@zicklag @pcwalton is the threading model of Surfman described anywhere? I.e. is it just the handle that's thread-local, or the actual device has to live on the same thread?
In the worst case, we'll make it so GL devices can't be moved between threads. Not the ideal situation, but I don't think the current one is better anyway :)

zicklag:
@kvark I was wondering that, too. I checked out the source for `surfman::Device` and it looks like it is just a wrapper around the connection and the desired adapter:

https://github.com/pcwalton/surfman/blob/ae14cc9afd140bc73e89f3fe8c080f729f09ccd2/surfman/src/platform/unix/x11/device.rs#L12-L18

```rust
/// A thread-local handle to a device.
///
/// Devices contain most of the relevant surface management methods.
pub struct Device {
    pub(crate) native_connection: Arc<NativeConnectionWrapper>,
    pub(crate) adapter: Adapter,
}
```

It looks like it should be cheap enough create new devices when we need them. It's just a clone of an Arc and an Enum.

```rust
pub(crate) fn new(connection: &Connection, adapter: &Adapter) -> Result<Device, Error> {
    Ok(Device {
        native_connection: connection.native_connection.clone(),
        adapter: (*adapter).clone(),
    })
}
```

`Device`'s are `!Sync` + `!Send`, `Surface`'s are `!Sync`, and `Context`'s are `!Sync` + `!Send` so it looks like all of those will have to stay confined to their threads. With the latest changes I pushed, I use a `thread_local!` `surfman::Connection` to back the `Instance` and when the users wants to `enumerate_adapters` or `create_surface` it uses the thread local connection, creating a new connection if one hasn't been created on that thread yet, and then creates a Device, Context, and Surface right there, in the same thread it was requested.

While you won't be able to transfer Surfaces over threads, at least you can still create them in multiple threads.

The thing I'm not sure about is whether or not `Device`'s created with different connections will share resources, which would be required for that setup.

kvark:
I just realized that the PR is targeting hal-0.4. Is it intended? If this is breaking the API, it should definitely go to `master`.

zicklag:
I don't know, actually. I haven't actually broken anything that was already there yet, just added a new feature and implementation for surfman, as well as a couple extra functions for `glutin`. As far as I know that shouldn't be breaking.

kvark:
Ah ok, that's fine then. We typically only target the past branches with patches, or at least small well-defined features. This one is a big and not well defined :)

zicklag:
OK, so I made some progress understanding the way some of this is put together. I finally realized that you need the framebuffer from the `surfman` `SurfaceInfo` and to `gl.bind_framebuffer()` in order to actually render to the surface. I had completely missed that. I also *think* I fixed a problem that would have occurred with the pixel format because I didn't have an alpha channel. Things are coming together in my head a bit better.

I've come to the conclusion that, right now, even though we would rather not have the blit in there, it should be working with the blit there and something else must be broken, I just haven't found out what it is yet. After finding the surfman surface framebuffer object I still don't know exactly where to put that framebuffer. In the code I put it here, but I'm not sure that is right:

https://github.com/gfx-rs/gfx/blob/4c9ab0742492800fdc13bed5c4948f367e4b76ee/src/backend/gl/src/window/surfman.rs#L250-L257

When I try running that I get a different invalid operation:

```
[2020-03-12T01:46:14Z ERROR gfx_backend_gl] [API/Error] ID 1 : GL_INVALID_OPERATION in glFramebufferRenderbuffer(window-system framebuffer)
thread 'main' panicked at 'Error allocating memory buffer InvalidOperation', /home/zicklag/git/other/gfx/src/backend/gl/src/device.rs:549:21
```

Any hints on that?

kvark:
> I've come to the conclusion that, right now, even though we would rather not have the blit in there, it should be working with the blit there and something else must be broken, I just haven't found out what it is yet

That sounds about right. Glad we are on the same page!

> After finding the surfman surface framebuffer object I still don't know exactly where to put that framebuffer

Wait, surfman gives you a framebuffer? My expectation (or wish?..) was that it gives you a GL texture, which we'd have as an attachment to our framebuffers. If it gives you a framebuffer, you can't configure it (i.e. add any attachments), you'd just be binding it to the `DRAW_FRAMEBUFFER` bind point and issuing a blit from another framebuffer (that we manage, like the current implementation does) that has an attachment texture with the frame pixels.

zicklag:
> you'd just be binding it to the DRAW_FRAMEBUFFER bind point and issuing a blit from another framebuffer (that we manage, like the current implementation does) that has an attachment texture with the frame pixels.

Ah, so in that design, the one we would prefer to avoid, we would blit on to our surface framebuffer from `surfman` instead of blitting onto framebuffer(0).

---

So I had read this:

https://github.com/pcwalton/surfman/blob/ae14cc9afd140bc73e89f3fe8c080f729f09ccd2/surfman/src/device.rs#L190-L196

```
    /// Returns various information about the surface, including the framebuffer object needed to
    /// render to this surface.
    /// 
    /// Before rendering to a surface attached to a context, you must call `glBindFramebuffer()`
    /// on the framebuffer object returned by this function. This framebuffer object may or not be
    /// 0, the default framebuffer, depending on plat
```

I was thinking you had to bind the framebuffer, so that is what I needed to get to render to the surface. It seems you can create a texture, but the textures are read only and doesn't work on the window widgets?

```
    /// Creates a surface texture from an existing generic surface for use with the given context.
    /// 
    /// The surface texture is local to the supplied context and takes ownership of the surface.
    /// Destroying the surface texture allows you to retrieve the surface again.
    /// 
    /// *The supplied context does not have to be the same context that the surface is associated
    /// with.* This allows you to render to a surface in one context and sample from that surface
    /// in another context.
    /// 
    /// Calling this method on a widget surface returns a `WidgetAttached` error.
    fn create_surface_texture(&self, context: &mut Self::Context, surface: Self::Surface)
                              -> Result<Self::SurfaceTexture, (Error, Self::Surface)>;
```

We were supposed to be writing to textures or do you write to the texture through the `Context` and only read the texture?

kvark:
From what you are quoting and saying, it does unfortunately look like we have to have our own texture to render to, and then just blit into the surfman framebuffer. Let's at least investigate why it's complaining when we do this. Have the code to reproduce it?

kvark:
Hmm, that actually contradicts https://github.com/gfx-rs/gfx/issues/3133#issuecomment-596174198 if that discussion was being understood by both parties the same way. I wonder if there is a tiny chance the code/comments you are looking at is just outdated stuff from surfman-0.1. In any way, we need to get the blit path working now, to restore some sanity, and go from there.

zicklag:
>  Have the code to reproduce it?

Yes. I'm just using the `wgpu-rs` cube and triangle examples as the test. You'll need to pull the following in addition to this PR:

  - https://github.com/gfx-rs/wgpu.git
  - https://github.com/gfx-rs/wgpu-rs/pull/182

Also my in `wgpu-rs` `Cargo.toml` file I have the following:

```
[patch.crates-io]
gfx-backend-gl = { path = "../gfx/src/backend/gl" }
gfx-hal = { path = "../gfx/src/hal" }
gfx-backend-vulkan = { path = "../gfx/src/backend/vulkan" }

[patch.'https://github.com/gfx-rs/gfx']
gfx-hal = { path = "../gfx/src/hal" }

[patch.'https://github.com/gfx-rs/wgpu']
wgpu-core = { path = "../wgpu/wgpu-core" }
wgpu-native = { path = "../wgpu/wgpu-native" }
```

To test it run `cargo run --example cube --features surfman --no-default-features`.

> I wonder if there is a tiny chance the code/comments you are looking at is just outdated stuff from surfman-0.1.

*Maybe*, but those docs are not from 0.1, they are from master, and they aren't actually on Docs.rs so they are newer than 0.1. They could still be older than the current implementation, but at the very least they are newer than the published crate.

> In any way, we need to get the blit path working now, to restore some sanity, and go from there.

:+1:

kvark:
I'm able to reproduce the problem on the gfx-rs quad example. Hooking up the whole wgpu-rs for this is a bit too early - we can work within gfx-rs until we get at least anything working. Here is the modification - https://github.com/kvark/gfx/commit/20f6e4da10f7b511d0c0e645a646c6cd79d7f8bc
Spewing the framebuffer status of `DRAW_FRAMEBUFFER`, which has surfman's FBO in it, gives me 0x8219, which stands for `GL_FRAMEBUFFER_UNDEFINED`. According to the [docs](https://www.khronos.org/opengl/wiki/Framebuffer_Object#Completeness_Rules) this would only happen if we don't really have a default framebuffer.

This is actually reasonable, since the default framebuffer is a property of the current context. What's missing here is a switch to the GL context associated with the surface we are blitting into. Once I add this, the error goes away, and the blit happens. However, there are more issues to take care of:
  1. we need to signal surfman's context that the frame is ready. I quickly looked for `swap_buffers` in surfman but didn't find any
  2. the log says we are creating a framebuffer every frame ("create_framebuffer [Renderbuffer(1)]"), which is suspicious
  3. nothing is actually rendered, the screen is black

@zicklag  Would you be able to take it further from here? Or need help looking into any of these?

zicklag:
@kvark Thanks for looking into that! I'll try to take it from here. This is a great learning experience.

I think the `surfman` version of `swap_buffers` might be [`present_surface`](https://docs.rs/surfman/0.1.4/surfman/platform/unix/wayland/device/struct.Device.html#method.present_surface) so I'll look into that.

I'm starting to get a better grip on the `threads` example for `surfman` which is confusing to me, but seems to indicate that I'm missing some essential steps such as binding the surface to a context which I recently wasn't doing.

>  we can work within gfx-rs until we get at least anything working

Sounds good.

kvark:
@zicklag where are we now?

zicklag:
I haven't had time to look into it again yet. I've been a bit busy. I might get to looking at it again in the next few days.

kvark:
There is a lot of progress here, and a great deal of investment. I just don't want it to be lost :)
Even if some stuff doesn't work as expected, we can land the code.

zicklag:
Its up to you if you want to land this and start another PR for continued work, but I'm still very interested in getting this working. :)

Glad to have interest in getting this in, though, I've had larger efforts than this that didn't end up getting into other projects before. :smiley:

zicklag:
Just an update: I did some more looking into it and after trying some stuff out, I'm still having trouble getting anything to show up on the screen. I'm going to do some more experimentation.

Also as far as the log about the framebuffer creation, it seems that the function that is logging that message is the `Device::create_framebuffer` which I don't think is called from anywhere inside the backend, which makes me think that it might be the demo's rendering code that is making that call over  and over.

Anyway, not much progress just yet, but I *think* we're close. :slightly_smiling_face:

kvark:
@zicklag great, thanks for the update! Let us know if there is anything of a concern, or a blocker.

zicklag:
So I just did a test to make sure I could get an onscreen version of surfman's offscreen example running, to make sure that I could actually get a non-black screen from surfman in the simplest situation. Everything worked fine, which is really good. The biggest thing that I proved out with that was that my `native_widget_from_rwh` ( raw window handle ) PR to surfman actually worked. Now that I'm sure what the simplest situation looks like I'm going to try to look into why things aren't getting to the screen.

There has to be a break somewhere where I'm not setting the right context to current or something like that I think. Otherwise, it is actually very simple to display the surface, it is just easy to get lost in the abstractions in the backend implementation.

Sorry for the super slow progress, but I'll keep working at it. :)

kvark:
Amazing news, @zicklag , let's get this work wrapped up!

zicklag:
Whoa! Praise the Lord, it just worked! The whole automatic "make context current" when dereferencing the `GlContainer` was making it break, probably because that context that it is making current should only really be used when evaluating the devices capabilities ( if I understand correctly ).

image

There are *all kinds* of TODO in the code all over the place, but it was kind of already a little like that. ;)

I'm going to look into that assertion that I commented out of WGPU to see if I can fix that now.

zicklag:
I made a fix for the failing assertions, but I don't know that it was technically correct. It seems to work so far. I just changed the default min storage buffer alignment to 256 instead of 1024.

kvark:
@zicklag greatness! let's make CI happy, and be sure to have surfman feature tested on Travis.
I'll re-review the code tomorrow.

zicklag:
>  let's make CI happy, and be sure to have surfman feature tested on Travis.

Good idea. I'll check out the CI failures.

A couple other tasks:

 - There is a panic that I need to fix that happens when dropping the Instance ( or surface or something like that ), but I know what the issue is there and it will be easy to fix. 
 - I don't have a wayland device to test on, or I don't know how to test Wayland anyway. I'm running PopOS and I think that it might be running an X11 protocol adapter on top of Wayand or something like that.
    - I didn't figure out how to get the surface size from a Wayland raw window handle: https://github.com/pcwalton/surfman/pull/61/files#diff-292e583dbc847cf34f3a0d0a4ba31801R189-R192

kvark:
I wish I knew the answer. 10 minutes of searching didn't reveal it :(
Calling for @msiglreith - do you know how this is supposed to work? How does one figure out the Wayland surface dimensions?

msiglreith:
@kvark there is no real size, it's determined on swapchain creation - see also the notes for the wayland extension https://renderdoc.org/vkspec_chunked/chap32.html#_wsi_surface

zicklag:
OK, did a good chunk of cleanup. I fixed the panic that was happening upon surface drop. I managed to get the `make` pristine on Linux, but it is failing in CI on windows and Mac. There must be some failure to feature gate the `surfman` and `glutin` crates out of the Windows and Mac builds. I'll have try fix that.

The examples don't work with the `glutin` backend anymore because we removed the `gl` feature specific code as the API is now properly aligned with the other backends ( yay! ). Technically I think the glutin backend still works fine if you conform to its API divergence ( I was using the `glutin` backend with `wgpu` earlier and it worked ).

I might have to look at WASM, too because that is failing CI as well.

Other than CI, is there anything else I should do?

kvark:
That's wonderful news! I think, as long as surfman path works, we can afford to (temporarily?) lose `glutin` path, given that the new path has no API hacks, and we can finally get it rolling on `wgpu` and friends. So CI is the only thing that needs to pass here, and please make sure `surfman` path is tested on it.
Also, it would be good to squash some commits. Ideally, there should be a linear progression with each step compilable by itself. If it's impossible, we should just squash everything.

zicklag:
The `surfman` path will be tested now with the examples depending on it so I *think* that should be all we need. I'll do an extra check of the Makefile to make sure I should't have any extra changes in there.

Probably almost none of those commits other than the last few actually work so I'll just squash the whole lot of them.

examples/Cargo.toml
```diff
@@ -8,8 +8,9 @@ edition = "2018"
 [features]
 default = ["debug-assertions"]
 metal = ["gfx-backend-metal"]
-gl = ["gfx-backend-gl", "gfx-backend-gl/surfman"]
-wgl = ["gfx-backend-gl", "gfx-backend-gl/wgl"]
+gl = ["gfx-backend-gl", "gfx-backend-gl/surfman"] # GL for Linux
+webgl = ["gfx-backend-gl"] # GL for web
```
grovesNL: nit: could we just disable `surfman` on `target_arch = "wasm32"` to avoid the additional `webgl` feature?

zicklag: That was actually how it was initially. I didn't realize it would come into effect here, but while I was combing through the backend implementation I thought that it didn't make any sense to compile with both the wasm32 arch and the `surfman`, `gluten`, or `wgl` features so I turned it into a compile error to combine any of those with `wasm32` arch.

This help a *lot* when it came to the cleanliness of the numerous `#[cfg()]` checks because you didn't have to add in some sort of check for `target_arch = "wasm32"` every time you needed to check for a feature. All you had to do was `feature = "surfman` and know that it was not also going to be targeting `wasm32`.

I tried to get around the extra feature by using cargo's `target` checks but apparently there is either a bug or otherwise strange behavior because if you do the following in the `Cargo.toml`, it adds both the `surfman` *and* the `wgl` feature, even to `wasm32` builds:

```toml
[target.'cfg(target_arch = "wasm32")'.dependencies.gfx-backend-gl]
path = "../src/backend/gfx-backend-gl"
optional = true

[target.'cfg(windows)'.dependencies.gfx-backend-gl]
path = "../src/backend/gfx-backend-gl"
features = ["wgl"]
optional = true

[target.'cfg(unix)'.dependencies.gfx-backend-gl]
path = "../src/backend/gfx-backend-gl"
features = ["surfman"]
optional = true
```

Its unfortunate that that doesn't work because that would allow us to automatically detect the correct feature without even having to have a separate one for `wgl`.

Anyway, I think that the extra feature is worth the readability in the code, especially because we can't get away from having separate features for `wgl` and `gl` anyway.

zicklag:
@kvark Do we need to test the `gl` backend on Mac? I don't think its ready yet, so we skip it in CI for Mac for now right?

kvark:
@zicklag what's missing in mac? I thought surfman would abstract this away.
Testing mac GL would be useful, as a backup for mac Metal.

zicklag:
At the very least it is missing an implementation for `surfman_surface_from_raw_window_handle`, which I implemented for Wayland and X11, but I can't test or implement for Mac. I'm not sure if there is anything else that isn't implemented, but actually I think you are right that `surfman` should handle the large majority of it at least.

kvark:
Is that in a surfman PR? I can try adding this and testing it.

kvark:
also, if it's on the surfman side, we don't have to block on it here

zicklag:
> Is that in a surfman PR? I can try adding this and testing it.

Yep: https://github.com/pcwalton/surfman/pull/61

> also, if it's on the surfman side, we don't have to block on it here

No, but it will fail CI until surfman works. Oh, maybe not actually because it should still compile even if it fails at runtime with `unimplemented!()`. OK, I'll see if I can get it to compile on Mac then, even if it doesn't run yet.

examples/Cargo.toml
grovesNL: Ok sounds good to me :+1:

In general I think having the extra `#cfg`s in our own implementation (even at the cost of some readability) is usually preferable to exposing additional features – mostly to make it easier for users to use. But there's not much we can do here if cargo doesn't easily support this :)

zicklag: Yeah, it would be *perfect* if cargo supported this. I opened a [forum topic](https://users.rust-lang.org/t/cargo-confusion-with-feature-and-target-specific-dependencies/40083?u=zicklag) just now to see if we can get any feedback on whether or not that could work.

Maybe its just a Cargo bug that we could fix, or maybe there is a workaround.

hecrj: This is a known issue in cargo: https://github.com/rust-lang/cargo/issues/1197

A solution seems to be available in nightly: https://github.com/rust-lang/cargo/issues/7914

zicklag: Cool, at least that means we'll be able to make that nicer after that gets stabilized. I'll subscribe to that thread and once it gets merged we could fix it if we don't mind requiring using the latest stable Rust.

zicklag:
I'm removing `glutin` from any tests for now if you are fine with that. We could bring it back later if we want.

zicklag:
It looks like surfman doesn't support iOS unfortunately. That leaves us some sort of decision to make.

Currently the big gain of the `surfman` implementation is the unified API such that we don't have to treat the GL backend differently than the other backends like we used to. The examples have already been adapted to this. If we have to use `glutin` for iOS, though, we would have specific handling for iOS, like we needed before `surfman`. I'm not sure if it is possible to make the `glutin` backend conform to the desired API similar to the `surfman` backend, but as I understand, the point of `surfman` was to give us an API advantage when threading that we wouldn't have with `glutin`.

Do we even need GL on iOS, really, though? Every iOS device is going to have metal on it right? In that case it seems like there isn't really any point to GL on iOS. It should be less performant to use GL, too.

I'm thinking we just don't test GL on iOS and target iOS with the metal backend.

kvark:
Yes, let's proceed without GL on iOS.

zicklag:
After thinking a bit about using this backend in `wgpu` I'm feeling like we should try to get the GL backend to compile without having to specify any target-specific features such as `wgl`. Otherwise I don't know that it can be compatible with `wgpu`'s concept of *not* needing to specify features for the backend. I'll look into that.

---

On another note, CI is passing! :tada:

kvark:
Correct - we should be allowing GL backend selected on platforms that support surfman, i.e. Linux X11, Wayland, Android, and macOS.

zicklag:
Got it.

With some help on the forum, I just [found a way](https://users.rust-lang.org/t/any-such-thing-as-cfg-aliases/40100/4?u=zicklag) to very nicely encapsulate `#[cfg]` checks into "aliases" that we can use in code so that we can just `#[cfg(surfman)]` instead of having to do stuff like `#[cfg(all(not(target_arch = "wasm32"), feature = "surfman", unix)))]` so I'm going to go through the code and make it so that on any platform you can include whatever features you want without hitting compiler errors. If the feature is not supported on the platform, then it will just not be used.

The perfectly addresses both mine and @grovesNL concerns about the features setup.

src/backend/gl/src/window/surfman.rs
```diff
@@ -0,0 +1,346 @@
...
+impl fmt::Debug for Instance {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Instance").field(&["Adapter..."; 3]).finish()
+    }
+}
+
+impl Instance {
+    fn get_default_context_attributes() -> sm::ContextAttributes {
+        sm::ContextAttributes {
+            version: sm::GLVersion::new(3, 3), // TODO: Figure out how to determine GL version
```
kvark: is that for surfman to create a context internally? One reasonable default is:
  - OpenGL ES 3.0 on mobile
  - OpenGL Core 3.2 on macOS
  - OpenGL 2.1 elsewhere

src/backend/gl/src/window/surfman.rs
```diff
@@ -0,0 +1,346 @@
...
+        Ok(self.create_surface_from_rwh(has_handle.raw_window_handle()))
+    }
+
+    unsafe fn destroy_surface(&self, surface: Surface) {
+        // Surface implments Drop and will clean up the surface when it gets dropped
+        drop(surface);
+    }
+}
+
+// TODO: Not sure if this TODO is relevant with surfman.
```
kvark: let's remove these TODOs, it seems irrelevant now

src/backend/gl/src/window/surfman.rs
```diff
@@ -0,0 +1,346 @@
...
+}
+
+impl Surface {
+    pub fn context(&self) -> Starc<RwLock<sm::Context>> {
+        self.context.clone()
+    }
+
+    fn swapchain_formats(&self) -> Vec<f::Format> {
+        // TODO: Make sure this is correct. I believe it is. Reference:
+        // http://docs.rs/surfman/struct.ContextAttributeFlags.html#associatedconstant.ALPHA
```
kvark: that link is dead

examples/Cargo.toml
```diff
@@ -6,13 +6,18 @@ workspace = ".."
...
+gl = ["gfx-backend-gl", "gfx-backend-gl/surfman"] # GL for Unix
+x11 = ["gfx-backend-gl/x11"]
+webgl = ["gfx-backend-gl"] # GL for web
+wgl = ["gfx-backend-gl", "gfx-backend-gl/wgl"] # GL for Windows
 dx11 = ["gfx-backend-dx11"]
 dx12 = ["gfx-backend-dx12"]
 vulkan = ["gfx-backend-vulkan"]
+# this is necessary because cargo doesn't allow using `target.'cfg(debug-assertions)'.dependencies`
+# to conditionally include crates
+debug-assertions = ["env_logger"]
```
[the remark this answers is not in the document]
zicklag: Is there already a built-in `debug-assertions` feature?

I wasn't exactly sure how to handle that, but the way it was previously setup with `target.'cfg(debug-assertions)'.dependencies` to gate the dependency was triggering a warning from Cargo that said that using the `target.cfg` feature like that wasn't supported for `debug-assertions`.

examples/Cargo.toml
```diff
@@ -33,18 +38,12 @@ log = "0.4"
 hal = { path = "../src/hal", version = "0.4", package = "gfx-hal" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.4" }
 winit = { version = "=0.20.0-alpha6", features = ["web-sys"] }
-
-[target.'cfg(debug_assertions)'.dependencies]
-env_logger = "0.6"
+env_logger = { version = "0.6", optional = true }
+gfx-backend-gl = { path = "../src/backend/gl", version = "0.4", optional = true }
```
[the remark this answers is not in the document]
zicklag: OK, sounds good.

src/backend/gl/Cargo.toml
```diff
@@ -28,13 +29,19 @@ gfx-hal = { path = "../../hal", version = "0.4" }
 auxil = { path = "../../auxil/auxil", version = "0.1", package = "gfx-auxil" }
 smallvec = "0.6"
 glow = "0.3.0-alpha3"
-parking_lot = "0.9"
+parking_lot = "0.10"
 spirv_cross = { version = "0.16", features = ["glsl"] }
 lazy_static = "1"
 raw-window-handle = "0.3"
+euclid = "0.20.7"
```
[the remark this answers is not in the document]
zicklag: I'll have to double-check that. That may have been left over from a previous iteration, but something ( maybe surfman? ) needed that to define a window/surface size.

src/backend/gl/src/window/surfman.rs
zicklag: > is that for surfman to create a context internally? 

I think so. Its uses those attributes when creating GL contexts.

I'll add those defaults you mentioned.

src/backend/gl/src/window/surfman.rs
zicklag: Ah, yeah, the docs I was referring to aren't released yet ( and I botched the link, too ).

I'll link to the doc string on GitHub for now.

examples/Cargo.toml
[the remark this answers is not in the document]
kvark: yes, in the code `cfg(debug_assertion)` is enabled in debug builds. Now that you are declaring it here, it will probably not conflict (i.e. it will be `cfg(feature = "debug_assertions")`, but I also don't understand how we'd be enabling this now.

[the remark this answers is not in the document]
zicklag: Yeah, the problem is that you can't include dependencies based on the `cfg(debug-assertions)`. The line in the cargo toml was previously attempting to only include `env_logger` when `cfg(debug-assertions)` was true, but that isn't supported in Cargo, so the only way to actually parameterize the inclusion of `env_logger` was to make a feature for it. Otherwise we could just always include `env_logger` and not make it optional. It is just an example and in a real-life situation you likely won't only be logging in debug mode anyway.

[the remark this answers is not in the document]
kvark: let's just depend on env_logger unconditionally in the examples, no big deal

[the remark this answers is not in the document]
zicklag: Sounds good.

zicklag:
I've been occupied all week, but I'll be attempting some work on this this weekend and hopefully will finish or make progress on the feature-less compilation and the rebase and other reviews.

zicklag:
With a some great help on the forum I was just able to release the [`cfg_aliases`](https://docs.rs/cfg_aliases) crate which greatly simplified our `cfg` checks throughout the code here. Now in the `build.rs` we have:

https://github.com/gfx-rs/gfx/blob/693317a043d3ee6d67be8bf0ea6db3c3269c11f3/src/backend/gl/build.rs#L8-L20

That creates aliases that we can then use in our code:

https://github.com/gfx-rs/gfx/blob/693317a043d3ee6d67be8bf0ea6db3c3269c11f3/src/backend/gl/src/lib.rs#L40-L65

That `dummy` check in particular would be rather verbose if it weren't for the alias, but now we can simply and *maintanably* manage our cfg checks without adding a bunch of verbose and confusing attributes.

Plus if we decided that we only wanted to support surfman on `linux` ( not that we would do such a thing ) we could update the cfg alias in *one* place and have it updated everywhere.

I've still got to update the examples and remove the feature checks from CI and such. I'm not 100% sure it is working right yet, but once I get CI to pass we should be in the clear. :smiley:

examples/Cargo.toml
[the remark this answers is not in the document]
zicklag: Do I rebase onto master or `hal-0.5`?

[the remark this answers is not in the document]
kvark: Let's land on master first, test it a bit, and then back-port to hal-0.5

src/backend/gl/src/window/surfman.rs
zicklag: I tested out using version 2.1 on Linux, but it panics with a shader compilation error:

    thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CompilationFailed("0:6(16): error: interface blocks with an instance name are not allowed in GLSL 1.40 (GLSL 1.50 or GLSL ES 3.00 required)\n")

It won't work until I go all the way up to 3.3.

zicklag:
@grovesNL I'm having issues with the latest `glow` for the WASM quad example because in `glow` 0.4 the [`UniformLocation`](https://docs.rs/glow/0.4.0/glow/trait.HasContext.html#associatedtype.UniformLocation) type isn't `Copy` anymore so `UniformDesc` can't implement Copy either.

https://github.com/gfx-rs/gfx/blob/5cd986dce448f20ce66ab8943484d3f74d38a1e2/src/backend/gl/src/native.rs#L315-L320

If I try to remove the `Copy` bound on `UniformDesc` I get issues because it isn't `Send` either, so it can't be put into an `RwLock` ( I'm pretty sure that's the cause, anyway ).

For now I just rolled back `glow` to 0.3-alpha4 again. `glow` 0.4 does work fine for the non-wasm backends. Do you know of anything I can do to address this?

zicklag:
@kvark @pcwalton Do you have any insight on to whether or not we should be using https://github.com/servo/surfman or https://github.com/pcwalton/surfman here?

I don't know what the relationship between the repos is, but now servo's version is 31 commits ahead of the @pcwalton's, [one of which](https://github.com/servo/surfman/commit/122eb77f17061afadf31f7ebe4551fbb66a5e979) being a method to resize a widget surface, which we are actually going to need. ( There is an open TODO in GL backend code right now ).

I know that servo recently merged @pcwalton's after having been way behind it. Is further development happening in servo's repo?

grovesNL:
@zicklag could we remove `Copy` from `UniformDesc` and try to fix the problem with `Send`? We could also wrap `UniformLocation` in `Arc` or something else if necessary (at least for now)

zicklag:
@grovesNL  OK, I've worked around it by putting it in a `Starc` that will make it `Sync + Send` but it will panic if accessed from outside of its original thread. Probably not ideal and I'm not sure if there is a better way around it, but it will work for now.

---

@kvark It looks like Mac builds are failing right now because the CI server doesn't have x11 installed. Do you know how to get X11 installed on there or does X11 not even run on Mac? According to the surfman cargo TOML it looks like it isn't excluding the X11 feature for MacOS so that seems to indicate that it *could* work.

I could add a `gl-x11` feature to the examples but it would have to default to off and users would have to enable it manually to support X11 on Linux. It would clutter up the features, which is something we are trying to avoid.

The two ways out of this I can see is to either put X11 on the CI server or tell surfman not to include X11 on MacOS unconditionally ( which if it actually *is* supported we would not want to do ).

zicklag:
OK, I think it is pretty much ready ( again :smile: ) other than the GL version selection comment above and MacOS/iOS builds which are failing because of the above-mentioned problem with X11.

kvark:
> It looks like Mac builds are failing right now because the CI server doesn't have x11 installed

We don't expect the users to have X11 installed on macOS either.
Can you make it so the x11 feature is auto-enabled for GL but only on Linux?

zicklag:
> We don't expect the users to have X11 installed on macOS either.

Yeah, good point. Unfortunately, limitations of Cargo are preventing us from changing the selected features of `surfman` depending on the target. I'll try to see if there is a way around it, though. I've got one idea I haven't tried yet.

zicklag:
@kvark I just noticed in the surfman [source code](https://github.com/pcwalton/surfman/blob/ae14cc9afd140bc73e89f3fe8c080f729f09ccd2/surfman/src/platform/unix/mod.rs#L11-L12) that the x11 module is actually gated so that it *won't* be used on `macos` even though the dependency on x11 ( which is actually what is failing, the dependencies `build.rs` ) is *not* gated. This means that all we have to do is gate the dependency as well as the code and I think we will be good.

That should be easy. I'll start on that.

---

Also @kvark what are you thinking about which surfman to use? Right now we have to use a git dependency already so that doesn't make a difference. Should I try to make a PR for servo/surfman with the changes we need because it is getting some more recent development?

Haven't heard anything from @pcwalton yet.

kvark:
Looks like servo/surfman is the upstream, and we should be using that. We've just made a gfx-hal-0.5 release, so it's fine to have some git dependencies for a while to settle up and test, until we need to release gfx-hal-0.6.

zicklag:
Almost have surfman ready. Might be done within the next few days. :crossed_fingers: .

zicklag:
@kvark @monocodus doesn't seem to be respecting the nightly-only rustfmt features such as spaces around range patterns. Which rustfmt version should I use?

kvark:
@zicklag please ignore the bot. It's fresh and we don't yet have a way to configure it properly.

hecrj:
I gave it a quick test and the `quad` example using the `gl` feature displays a black screen on my environment (Arch Linux with X11, NVIDIA drivers).

I decided to test the `master` branch of `servo/surfman` and the `threads` example seems to exhibit the same issue. I have reported the problem here: https://github.com/servo/surfman/issues/173

I am guessing both things are related, so I figured I'd mention it in case someone else decides to test this.

zicklag:
@kvark There was that unaddressed note about the default GL version that I couldn't figure out ( actually now that I say that, I can't remember if an incorrect value there might have led to a blank screen like @hecrj is getting ), but otherwise, CI just passed ( though I re-pushed after squashing commits ) and I think Its done!

src/backend/gl/src/window/surfman.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+}
+
+impl Instance {
+    fn get_default_context_attributes() -> sm::ContextAttributes {
+        sm::ContextAttributes {
+            version: sm::GLVersion::new(3, 3), // TODO: Figure out how to determine GL version
+            // TODO: Determine flags to provide. At least ALPH I think, but probably all of them.
+            // skipping COMPATIBILITY_PROFILE for now, because it panics with a TODO.
+            flags: sm::ContextAttributeFlags::ALPHA
```
kvark, line 64: let's not have DEPTH/STENCIL at all for the swapchain

src/backend/gl/src/window/surfman.rs
```diff
@@ -0,0 +1,342 @@
...
+    unsafe fn create_surface(
+        &self,
+        has_handle: &impl raw_window_handle::HasRawWindowHandle,
+    ) -> Result<Surface, window::InitError> {
+        Ok(self.create_surface_from_rwh(has_handle.raw_window_handle()))
+    }
+
+    unsafe fn destroy_surface(&self, surface: Surface) {
+        // Surface implments Drop and will clean up the surface when it gets dropped
+        drop(surface);
```
kvark, line 191: I wonder if this is idiomatic, given that you can just declare it as `_surface: Surface`

src/warden/src/lib.rs
```diff
@@ -9,47 +9,6 @@ extern crate serde;
 pub mod gpu;
 pub mod raw;
 
-#[cfg(feature = "gl")]
-pub fn init_gl_surface() -> gfx_backend_gl::Surface {
```
kvark, line 13: yay!

src/backend/gl/Cargo.toml
```diff
@@ -17,8 +17,9 @@ build = "build.rs"
 name = "gfx_backend_gl"
 
 [features]
-default = []
+default = ["wgl", "surfman", "surfman-x11"]
```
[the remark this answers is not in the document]
zicklag: It won't fail. CI passes on all platforms with all of those features enabled. That was part of point of the latest refactoring. The setup is that you could enable any features you want and it won't error on any platform because the features are properly gated. So even if you enable `wgl` on Linux or `surfman-x11` on MacOS, it will just ignore those because they aren't supported on that platform. That way we don't have to dynamically enable or disable features based on the target platform. We can just put whatever features we want to be enabled across all platforms in the default and users can happily build anywhere without specifying anything.

Because they are features though, if you want to build on Linux without X11 you can still do `cargo build --no-default-features --features surfman,wgl` and it will build it with wgl on windows, but when building on Linux it will use surfman without X11.

src/backend/gl/Cargo.toml
```diff
@@ -32,9 +33,14 @@ parking_lot = "0.10"
 spirv_cross = { version = "0.18", features = ["glsl"] }
 lazy_static = "1"
 raw-window-handle = "0.3"
+glutin = { version = "0.23.0", optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-glutin = { version = "0.23.0", optional = true }
+winit = "0.21.0"
```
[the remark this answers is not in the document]
zicklag: Good question, I'll double-check that.

src/backend/gl/build.rs
```diff
@@ -2,8 +2,24 @@ use gl_generator::{Api, Fallbacks, Profile, Registry};
 use std::env;
 use std::fs::File;
 use std::path::PathBuf;
+use cfg_aliases::cfg_aliases;
 
 fn main() {
+    // Setup cfg aliases
+    cfg_aliases! {
```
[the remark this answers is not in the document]
zicklag: No, because it actually needs to run `println!()`'s to integrate the aliases with Cargo.

src/backend/gl/src/window/surfman.rs
zicklag: I thought about that, but it seems a little strange right off because it almost seems like the function doesn't do anything until you realize that `Surface` is owned and will therefore be dropped.

I'm good going either way with it, though. Wasn't really set on one way or the other.

src/backend/gl/Cargo.toml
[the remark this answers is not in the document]
kvark: I wonder how big these dependencies are. In Gecko, we will not need to `wgl` for sure, for example, we wouldn't want to keep that in tree of dependencies.
In general, gfx-backend-gl users will unlikely be very interested in `wgl`, given the availability of the other backends on Windows. Maybe we could at least not make it a default?

src/backend/gl/src/window/surfman.rs
kvark: yeah, I think calling `drop()` explicitly has its own advantages. It's just weird to see, makes you suspicious.
The danger of doing `_surface` is that it may be tempting for someone who looks at the code and tries to refactor it, to remove the parameter if it's unused. This danger doesn't apply if we are implementing an existing trait (like we do here), so it's fine to do.

src/backend/gl/Cargo.toml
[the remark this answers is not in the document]
zicklag: If you are using `gfx-backend-gl` on Windows and you don't provide the `wgl` feature or the `glutin` feature, it will end up compiling in Dummy mode and not doing anything. If that were the case for your application I would think you would just not include `gfx-backend-gl` at all on  that platform by using target specific dependencies.

In other words if we don't default either `glutin` or `wgl`, the default operation of `gfx-backend-gl` on windows would be to do nothing which doesn't seem right to me.

src/backend/gl/Cargo.toml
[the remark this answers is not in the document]
zicklag: Actually it doesn't look like it is even used. The only reference to it is an import of a couple of items, but even when the `winit` feature is enabled the imports remain unused so I'll just take that out completely.

src/backend/gl/Cargo.toml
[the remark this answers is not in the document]
kvark: Right, I see. I think it's fine. The downside for having `wgl` is that you pay for this dependency to be included in the tree even though you aren't going to use it ever.

let's say, if there is a user who needs GL on windows and Linux, they will just enable `wgl` feature unconditionally. Others will not need to know about it.

src/backend/gl/src/window/surfman.rs
kvark: looks like SPIRV-Cross needs to be configured properly. It's fine, we can do it later :)

src/backend/gl/Cargo.toml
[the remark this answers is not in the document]
zicklag: I'm still a little confused. When would you include `gfx-backend-gl` on Windows if you weren't going to use it on Windows? The `wgl` dependencies aren't included when you are building on Linux.

Would the problem situation be when a user includes `gfx-backend-gl` for a project, only intending to use GL on Linux, and then accidentally including the `wgl` dependencies when building for Windows, even though there were actually using something like the `dx11` backend for Windows?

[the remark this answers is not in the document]
kvark: We wouldn't include gfx-backend-gl on Windows for `wgpu`, but the way your PR is set up right now, we'd still technically depend on `wgl`. Which means `wgl` and its dependencies will be erroneously vendored in Gecko third party sources.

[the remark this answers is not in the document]
zicklag: In that case I think we would just do this in `wgpu`:

```toml
[target.'cfg(unix, not(target_os = "ios")'.dependencies]
gfx-backend-gl = "0.5"
```
Because you aren't including `gfx-backend-gl` on Windows, the `wgl` dependencies ( which is actually just the `winapi` crate ) will never be included into the tree because the `winapi` crate is only brought in by `gfx-backend-gl` on windows:

```toml
[target.'cfg(windows)'.dependencies]
winapi = { version = "0.3", features = ["libloaderapi", "windef", "wingdi", "winuser"] }
```

So if you don't include `gfx-backend-gl` on windows, you won't ever include `winapi`.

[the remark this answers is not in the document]
kvark: Actually, if it's just winapi, then we are totally fine - it's already vendored on Windows.
:)

zicklag:
Unfortunately I just found that the reftests are segfaulting for GL and the makefile wasn't testing GL on CI properly.

I'll have to look into that. I just pushed a commit that addresses all of the points of the review and additionally allows you to run with both the `glutin` and the `surfman` features enabled with `surfman` taking precedence on Linux. This lets you use `glutin` on Windows and `surfman` on Linux if you preferred `glutin` for Windows over `wgl`.

---

> As the default features being our last change, it would be great to try taking the quad example from hal-0.5 branch, moving it out, and compiling with "GL" feature against your modified code. 

I'll try that out.

zicklag:
For some reason it is segfaulting right here:

https://github.com/gfx-rs/gfx/blob/99dc1164bd77bb56344dc79c2e510874ac8c9f34/src/backend/gl/src/info.rs#L126

It is when it tries to get the vendor name from the `gl` context. There it isn't happening when running the `quad` example, but it will happen if you re-order the creation of the surface and the enumeration of the adapters.

Any clue what might cause that?

kvark:
Is SetCurrentContext being called?

zicklag:
Ah, thank you, that is definitely the cause.

Apparently I was just getting lucky with the fact that it happened to be working with the order that the current `quad` and `wgpu` examples were running certain functions. Now I'm having a little bit of trouble, but I'm not necessarily stuck yet.

The problem is that when I try to run `make_current` on the context whenever the `GlContainer` is dreferenced, it just renders a blank screen for some reason. When I don't make current on dereference it errors out somewhere along the line ( unless you get lucky, like I did ealier ) and I need to figure out all of the places in the code I need to make current now. Hopefully that won't be too difficult.

zicklag:
@kvark I found out where to put the `make_context_current` call to fix the error, but I'm not 100% sure it is the best spot to put it:

https://github.com/gfx-rs/gfx/blob/c4e2c2dac4b4bdab0cfde62b36047de0cc61fda7/src/backend/gl/src/window/surfman.rs#L314-L317

It is the earliest spot that I could find after enumerating adapters that has access to the specific surface that needs to be made current. I tried sticking it in `configure_swapchain` which is where I felt that it should go, but it panics with an invalid command error.

It seems like maybe we should be making the context current on de-reference, but I still don't know how to fix the black screen that happens when we do that, and I don't know how to go about debugging it. I feel like we can leave it for now and fix it later.

---

The reftests aren't segfaulting anymore but they are hitting `unimplemented!()` for `Device::create_query_pool()`. Do we just leave that for now?

Also, CI isn't set to run the reftests, but there is a Makefile entry specifically for CI reftests. Does that need to be added to the Travis config?

zicklag:
I just tested the following steps and got the `quad` and `colour-uniform` examples working on `hal-0.5`:

1. switch to this PR's branch
1. rebase onto `hal-0.5`
1. Change the default features of `gfx-backend-gl` to `[]` like it was before the PR, to avoid making a breaking change
1. Update the `gfx-backend-gl` dependency to include the `features = ["wgl", "surfman", "surfman-x11"]`
1. Run the examples

So as long as I'm not missing something other than the default features that makes this change breaking, I think we can safely backport it to 0.5!

kvark:
@zicklag what happens if the example tries to use the `glutin` feature like it does today?

zicklag:
Ah, I get it now, we want to test the *old* examples to make sure that they still work as a validation of the fact that our changes are still backward compatible. :+1:

I tested it and, good news! The examples work without changes after the latest commit I made to re-add some "unused" imports/exports in the `glutin` implementation.

That means that the only difference that will be necessary for the backport and the master commit will be to make sure the default features and examples stay unchanged from `hal-0.5` for the backported version. :star2: Piece of :cake:.

kvark:
That's wonderful!
So where do we stand now? Are there known issues blocking the merge at this point?

zicklag:
Not that I know of. Some things such as when the context gets made current might not be technically correct or work in all cases, but that is kind of the story with the whole backend right now. :wink: 

I think its ready. If we notice anything else after we can make follow-up PRs.

kvark:
@zicklag would you want to follow-up with a PR to hal-0.5 that has this code with the default features fixed?

zicklag:
Yeah, I can probably have that PR ready by tonight.

PR #3154, opened 2020-02-26
allow specification of API version during instance creation

### LINKED ISSUE #3146: Exposing Vulkan 1.1 entry points.
### Problem

Currently, gfx-backend-vulkan only exposes Vulkan 1.0 to the user. See [here](https://github.com/gfx-rs/gfx/blob/0525c1748c04d2fa878af99ddf424db635a31934/src/backend/vulkan/src/lib.rs#L18) and [here](https://github.com/gfx-rs/gfx/blob/0525c1748c04d2fa878af99ddf424db635a31934/src/backend/vulkan/src/lib.rs#L351) in particular.

Vulkan 1.1 has exposed [subgroup operations](https://www.khronos.org/blog/vulkan-subgroup-tutorial) (see also this [presentation](https://www.khronos.org/assets/uploads/developers/library/2018-vulkan-devday/06-subgroups.pdf)), which are very nice for compute applications in particular. A gfx-hal user cannot currently access these, due to the hardcoded Vulkan 1.0 usage set by gfx-backend-vulkan. 

### Solution

I do not have enough API design knowledge to suggest how one can expose Vulkan 1.1 to the user. My best guess would be using feature flags, so that feature flag `vulkan` loads Vulkan 1.0, while feature flag `vulkan1.1` loads Vulkan 1.1. Then we could have for instance:

```
#[cfg(feature="vulkan")]
use ash::version::{DeviceV1_0, EntryV1_0, InstanceV1_0};
#[cfg(feature="vulkan1.1")]
use ash::version::{DeviceV1_1, EntryV1_1, InstanceV1_1};
```

(and similarly a `cfg` flag to set `vk_make_version!(1, 1, 0)` if `vulkan1.1` is enabled?)

### Conclusion

I am worried that this might be much more complicated than I am imagining it to be, but I am happy to implement to any solutions that are decided upon through discussion here.

bzm3r:
Can I have input on:

1) the approach I have taken to solve the issue? In particular, see the `ApiVersion` struct defined in `src/hal/lib.rs`

2) the documentation I have provided. 

Fixes #3146
PR checklist:
- [x] `rustfmt` run on changed code (**except** for `backend/gl`, which produces an error)

`cargo fmt` error when it is run on `backend/gl`:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\tools\rustfmt\src\lists.rs:678:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\tools\rustfmt\src\lists.rs:678:17
```

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -597,7 +597,8 @@ fn create_backend(
 ) -> BackendState<back::Backend> {
     let window = wb.build(event_loop).unwrap();
     let instance =
-        back::Instance::create("gfx-rs colour-uniform", 1).expect("Failed to create an instance!");
+        back::Instance::create("gfx-rs colour-uniform", 1, hal::ApiVersion::new(1, 1, 1))
+            .expect("Failed to create an instance!");
     let surface = unsafe {
         instance
             .create_surface(&window)
--- a/examples/compute/main.rs
+++ b/examples/compute/main.rs
@@ -41,8 +41,8 @@ fn main() {
         .collect();
     let stride = std::mem::size_of::<u32>() as u64;
 
-    let instance =
-        back::Instance::create("gfx-rs compute", 1).expect("Failed to create an instance!");
+    let instance = back::Instance::create("gfx-rs compute", 1, hal::ApiVersion::new(1, 1, 1))
+        .expect("Failed to create an instance!");
 
     let adapter = instance
         .enumerate_adapters()
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -115,8 +115,8 @@ fn main() {
     #[cfg(not(feature = "gl"))]
     let (_window, instance, mut adapters, surface) = {
         let window = wb.build(&event_loop).unwrap();
-        let instance =
-            back::Instance::create("gfx-rs quad", 1).expect("Failed to create an instance!");
+        let instance = back::Instance::create("gfx-rs quad", 1, hal::ApiVersion::new(1, 1, 1))
+            .expect("Failed to create an instance!");
         let surface = unsafe {
             instance
                 .create_surface(&window)
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -270,7 +270,7 @@ fn get_format_properties(
 }
 
 impl hal::Instance<Backend> for Instance {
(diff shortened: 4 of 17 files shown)


kvark:
I'm a bit concerned about going this way. Isn't Vulkan 1.1. essentially just a Vulkan 1.0 plus a few extensions? In this case, we should be getting it in the same way as we are getting extensions. Speaking of which, we don't have any ability to specify *instance* extensions.

So, perhaps, we can approach it in the following way. Have a bitflags type for every bit associated with an Instance "extension". Mechanically, that would be similar to requesting extensions when instantiating a logical device, but at the instance level. So we'd have a static associated function/constant to an Instance that would return the bitflags of *supported* extensions. So the user would get that bitflags, AND it with whatever they need, and pass it to `Instance::create`.

Then, we could have a bitflags alias for all the things included in Vulkan 1.1.
@bzm3r @msiglreith @grovesNL does that sound feasible to any extent?

grovesNL:
@kvark that approach sounds reasonable to me. I think the alias would work well in the gfx-portability implementation too.

bzm3r:
@kvark It is not clear to me that Vulkan 1.1 is just Vulkan 1.0 + a few extensions. According to the [press release](https://www.khronos.org/news/press/khronos-group-releases-vulkan-1-1), while Vulkan 1.1 does standardize many extensions, it seems to be more than that?

For example, in [this presentation](https://www.khronos.org/assets/uploads/developers/library/overview/Vulkan-1-1-Presentation_Mar18.pdf), there is a slide titled "New Functionality in Vulkan 1.1" and a slide titled "Proven Extensions Now in Vulkan 1.1 Core". Subgroup operations and protected content are in "New Functionality", and not "Proven Extensions"?

kvark:
Thank you for the great links @bzm3r ! You are right, Vulkan 1.1 may be a bit more than just a couple of extensions being promoted. I don't think it changes our approach though:
```
const EXTENSION_1 = 0x01;
const EXTENSION_2 = 0x02;
const VULKAN_1_1_EXTRA = 0x4; // doesn't need to be public
const VULKAN_1_1 = EXTENSION_1 | EXTENSION_2 | VULKAN_1_1_EXTRA;
```

msiglreith:
Sounds ok to me, even though just following Vulkan might be the easiest :p
Vulkan extension revisions are a possible inconvenience.

PR #3155, opened 2020-02-26
[vk] supress into_iter autoref coercion warning

bzm3r:
Currently, the following warning is emitted during building:

```
warning: this method call currently resolves to `<&[T; N] as IntoIterator>::into_iter` (due to autoref coercions), but that might change in the future when `IntoIterator` impls for arrays are added.

   --> src/backend/vulkan/src/lib.rs:296:51

    |

296 |         for (info_label, info) in additional_info.into_iter() {

    |                                                   ^^^^^^^^^ help: use `.iter()` instead of `.into_iter()` to avoid ambiguity: `iter`

    |

    = note: `#[warn(array_into_iter)]` on by default

    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!

    = note: for more information, see issue #66145 <https://github.com/rust-lang/rust/issues/66145>
```

This PR wonders if it would be okay to do what the warning suggests we do.

Fixes #issue (not applicable, no known issue)
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -293,6 +293,7 @@ unsafe extern "system" fn debug_utils_messenger_callback(
             message
         );
 
+        #[allow(array_into_iter)]
         for (info_label, info) in additional_info.into_iter() {
             match info {
                 Some(data) => {


kvark:
I'm confused by this error. It says "this was previously accepted by the compiler but is being phased out", yet the arrays *will* support `IntoIterator`, just with a different meaning. That new meaning is exactly what we want here, so I don't think anything needs to be changed on our side, other than maybe silencing the warning in this case?

bzm3r:
@kvark ok, I think I have suppressed the warning

bzm3r:
Any idea why the Travis CI build is failing?

kvark:
Network problems?
bors r+

grovesNL:
bors r=kvark

LukasKalbertodt:
Hi there! I'd really advice against silencing this warning. 

>  That new meaning is exactly what we want here, so I don't think anything needs to be changed on our side, other than maybe silencing the warning in this case?

The problem is that in the future the expression `additional_info.into_iter()` will return an iterator over the elements *by value* whereas today an iterator over *references* to those values is returned. This very easily leads to broken builds, as you can imagine. In your specific case it actually doesn't matter as you only print the values and due to match ergonomics. However, that's an exception. And I'd still advice you to just replace `into_iter()` with `iter()`.

If you have further questions, just ask me. Sorry for bothering you with future compatibility lints and thanks for referencing the lint tracking issue so that I was able to catch this potential problem.

kvark:
@LukasKalbertodt we do want it to be by value, hence I said "exactly what we need here". Breaking the build would be unfortunate though, of course. Changing them to `iter()` now and then back to `into_iter()` once it's changed would probably be the safest option, still.

LukasKalbertodt:
Yeah, for most people the *new* behavior of `array.into_iter()` is what they want. But all the current code has been written with the *current* behavior (iterator over references). And that code probably won't compile anymore once the behavior changes. 

But yeah, it might be fine here. The worst case -- if you keep suppressing the warning -- would be someone (who doesn't know about this) changing this piece of code so that it suddenly depends on the variables being references (and not values) and then the code breaks later. But that seems unlikely. 

I just wanted to make sure the full situation is clear to everyone. The decision is yours of course.

PR #3157, opened 2020-03-02
check if addDebugMarker from Metal is supported

### LINKED ISSUE #3098: MTLIGAccelBuffer addDebugMarker:range:]: unrecognized selector
I receive this error message:

```
   Compiling wgpu v0.4.0 (/Users/da1/Sources/rust/wgpu-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 6.35s
     Running `target/debug/examples/cube`
2019-11-25 14:56:16.620 cube[16426:135436] -[MTLIGAccelBuffer addDebugMarker:range:]: unrecognized selector sent to instance 0x7fdfe35a14c0
[2019-11-25T13:56:16Z ERROR relevant] Values of this type can't be dropped!
[1]    16426 segmentation fault  cargo run --example cube
```

when running the cube example or the skybox one.

I am on Mac OS "El Capitan", MacBook Pro (Retina, 13-inch, Mid 2014).
GPU Intel Iris 1536 MB.

counterfactual-regret-minimization:
Fixes #3098 fixes gfx-rs/wgpu-rs/issues/138
PR checklist:
- [ ] ~~`make reftests` succeeds~~
it complains about `Testing Metal` "metal-0.17.1/src/capturemanager.rs:50:25": `Scene 'transfer': thread 'main' panicked at 'Class with name MTLCaptureManager could not be found'` because [that object doesn't exist on my macOS 10.11 machine](https://developer.apple.com/documentation/metal/mtlcapturemanager)
Metal

### DIFF
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -114,10 +114,9 @@ fn get_final_function(
     })?;
 
     if !function_specialization {
-        assert!(
-            specialization.data.is_empty() && specialization.constants.is_empty(),
-            "platform does not support specialization",
-        );
+        if !specialization.data.is_empty() || !specialization.constants.is_empty() {
+            error!("platform does not support specialization");
+        }
         return Ok(mtl_function);
     }
 
@@ -905,7 +904,7 @@ impl Device {
                 msl::ComponentSwizzle::Identity,
                 msl::ComponentSwizzle::Identity,
                 msl::ComponentSwizzle::Identity,
-                msl::ComponentSwizzle::Identity
+                msl::ComponentSwizzle::Identity,
             ],
             ycbcr_conversion_enable: false,
             ycbcr_model: msl::SamplerYCbCrModelConversion::RgbIdentity,
@@ -2374,7 +2373,7 @@ impl hal::device::Device<Backend> for Device {
                 let options = conv::resource_options_from_storage_and_cache(storage, cache);
                 if offset == 0x0 && size == cpu_buffer.length() {
                     cpu_buffer.set_label(name);
-                } else {
+                } else if self.shared.private_caps.supports_debug_markers {
                     cpu_buffer.add_debug_marker(
                         name,
                         NSRange {
@@ -2708,7 +2707,7 @@ impl hal::device::Device<Backend> for Device {
                     assert_eq!(mip_sizes.len(), 1);
                     if offset == 0x0 && cpu_buffer.length() == mip_sizes[0] {
                         cpu_buffer.set_label(name);
-                    } else {
+                    } else if self.shared.private_caps.supports_debug_markers {
                         cpu_buffer.add_debug_marker(
                             name,
                             NSRange {
@@ -3062,13 +3061,15 @@ impl hal::device::Device<Backend> for Device {
             n::Buffer::Bound {
                 ref raw, ref range, ..
             } => {
-                raw.add_debug_marker(
-                    name,
-                    NSRange {
-                        location: range.start,
-                        length: range.end - range.start,
-                    },
-                );
+                if self.shared.private_caps.supports_debug_markers {
+                    raw.add_debug_marker(
+                        name,
+                        NSRange {
+                            location: range.start,
+                            length: range.end - range.start,
+                        },
+                    );
+                }
             }
         }
     }
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -713,6 +713,7 @@ struct PrivateCapabilities {
     max_texture_layers: u64,
     max_fragment_input_components: u64,
     sample_count_mask: u8,
+    supports_debug_markers: bool,
 }
 
 impl PrivateCapabilities {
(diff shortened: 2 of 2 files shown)


src/backend/metal/src/device.rs
```diff
@@ -114,10 +115,14 @@ fn get_final_function(
     })?;
 
     if !function_specialization {
-        assert!(
-            specialization.data.is_empty() && specialization.constants.is_empty(),
-            "platform does not support specialization",
-        );
+        if caps_support_debug_markers {
```
kvark: the specialization is not related to the debug markers

src/backend/metal/src/lib.rs
```diff
@@ -958,6 +959,11 @@ impl PrivateCapabilities {
             max_texture_layers: 2048,
             max_fragment_input_components: if os_is_mac { 128 } else { 60 },
             sample_count_mask,
+            supports_debug_markers: if os_is_mac {
```
kvark: could we rewrite that to be consistent with the other checks, which are using `supports_any`?

counterfactual-regret-minimization: Not with confidence -- i'm not clear how `MTLFeatureSet` works: if i test for [macOS_GPUFamily1_v2](https://developer.apple.com/documentation/metal/mtlfeatureset/macos_gpufamily1_v2), will that check also be successful for later OS versions, or do i need to explicitly list them too?

src/backend/metal/src/device.rs
counterfactual-regret-minimization: So how do i check if specialization is supported? That assertion fails on my machine, but i don't think it should even be being tested.

kvark: The capability of checking for specialization support is already passed in as another parameter - `function_specialization`. Perhaps, all we want to do here is to change the `assert` into `error!` as you did, just without involving `caps_support_debug_markers`.

src/backend/metal/src/lib.rs
kvark: IIRC, the versions are backwards compatible, but GPU families are not. So you'd list the pairs of (OS, GPU Family) -> minimal supported version, if any.

src/backend/metal/src/device.rs
counterfactual-regret-minimization: Ah, that would work; the problem was that that assert was panicking on my machine, which doesn't support specialization apparently, but the examples work if i comment it out.

src/backend/metal/src/lib.rs
counterfactual-regret-minimization: Explicitly listing the GPU families where the function exists seems like the wrong thing to do, because when future ones come out the check will incorrectly think they don't support the function?

kvark: I agree this is a tricky thing, but I don't see a better alternative. When a new GPU family comes out (and they do come out), we don't know if a particular feature is supported. So I think it's safest to assume no feature until we inspect the updated spec and update these listings with the new family.

src/backend/metal/src/device.rs
```diff
@@ -104,7 +104,7 @@ fn get_final_function(
     library: &metal::LibraryRef,
     entry: &str,
     specialization: &pso::Specialization,
-    function_specialization: bool,
+    function_specialization: bool
```
kvark: leftover change? trailing comma is good

src/backend/metal/src/device.rs
```diff
@@ -114,10 +114,9 @@ fn get_final_function(
     })?;
 
     if !function_specialization {
-        assert!(
-            specialization.data.is_empty() && specialization.constants.is_empty(),
-            "platform does not support specialization",
-        );
+        if specialization.data.is_empty() && specialization.constants.is_empty() {
```
kvark: the condition should be the opposite: if specialization has anything, we should be erroring out

kvark: Could also return a `FunctionError` btw, which is slightly better, but I'm fine either way, since it's not the focus of your PR

src/backend/metal/src/device.rs
counterfactual-regret-minimization: I *think* i got the squashing correct, have never done that before. Also it took me ages to realise i should be running `rustfmt +nightly *` so i may have misformatted things before.

src/backend/metal/src/lib.rs
```diff
@@ -713,7 +713,11 @@ struct PrivateCapabilities {
     max_texture_layers: u64,
     max_fragment_input_components: u64,
     sample_count_mask: u8,
+<<<<<<< HEAD
```
kvark: that's an unresolved merge conflict :)

kvark:
Oh uh, many commits here now. The way I typically squash is:
```git rebase -i <base_branch>``` then selecting the commits for squashing

src/backend/metal/src/lib.rs
counterfactual-regret-minimization: I have no idea how to do that 'squashing' thing :(

kvark:
Would you mind joining `#gfx:matrix.org` chat (on the Matrix)? It would be easier to discuss Git practices there than here.

src/backend/metal/src/device.rs
kvark: hmm, the trailing comma is still missing here?

PR #3159, opened 2020-03-06
[0.4] Cache surface formats to reduce overhead with swapchain recreation

aloucks:
Some device/drivers have significant overhead when querying supported surface formats. This can result in choppy window resizing due to slow swapchain re-creation.

By caching the formats, we can achieve extremely smooth window resizing.

Fixes #issue
PR checklist:

The `colour-uniform` example didn't seem to benefit too much, but you could also check out this example that now resizes very smoothly:

https://github.com/aloucks/imgui-glfw-support/blob/master/examples/demo_window_smooth_resize.rs

```toml
[patch.crates-io]
gfx-hal = { path = "../gfx/src/hal" }
gfx-backend-vulkan = { path = "../gfx/src/backend/vulkan" }
```

### DIFF
--- a/src/backend/vulkan/src/window.rs
+++ b/src/backend/vulkan/src/window.rs
@@ -1,5 +1,6 @@
 use std::{
     borrow::Borrow,
+    collections::HashMap,
     fmt,
     hash,
     os::raw::c_void,
@@ -94,6 +95,7 @@ pub struct RawSurface {
     pub(crate) handle: vk::SurfaceKHR,
     pub(crate) functor: khr::Surface,
     pub(crate) instance: Arc<RawInstance>,
+    formats_cache: Mutex<HashMap<vk::PhysicalDevice, Option<Vec<Format>>>>,
 }
 
 impl Instance {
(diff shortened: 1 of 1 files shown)


aloucks:
I'm hoping to have WGPU benefit from this. I guess it could be implemented there instead if you're open to that. However, these values can't change at runtime and the memory required to cache them is pretty small. Caching it at this lower levels allows the larger ecosystem to benefit from it (e.g. Rendy).

In regards to WGPU, this is abstracted away by the library and the user application doesn't have the ability to cache and reuse this data when re-creating the swapchain.

kvark:
Conceptually putting this here is not the right thing to do. gfx-rs aims to provide the building blocks, not convenience. If Vulkan function is slow, we are just exposing it, we aren't trying to make it faster magically.

> these values can't change at runtime

Are you sure about this? I thought that you may have a situation where a window is moved to a different monitor, and out of a sudden there are different formats to present.

> In regards to WGPU, this is abstracted away by the library and the user application doesn't have the ability to cache and reuse this data when re-creating the swapchain.

I'd be totally happy to have this handled in WGPU :)

aloucks:
This seems to be an Nivida driver bug that's been fixed in the latest version `442.50`.

https://devtalk.nvidia.com/default/topic/1046926/vulkan/execution-time-of-vkgetphysicaldevicesurfaceformatskhr-increased-by-20-000-with-driver-411-63/

PR #3160, opened 2020-03-06
Cache surface formats to reduce overhead with swapchain recreation

aloucks:
Some device/drivers have significant overhead when querying supported surface formats. This can 
 esult in choppy window resizing due to slow swapchain re-creation.

By caching the formats, we can achieve extremely smoothe window resizing.

This is the master branch companion to: https://github.com/gfx-rs/gfx/pull/3159


kvark:
(Being discussed in #3160)

PR #3165, opened 2020-03-18
Introduce buffer SubRange

### LINKED ISSUE #3166: Replace RangeArg by SubRange
gfx-rs API is overloaded with generics. It's also not fully focused on *why* generics are used. I.e. the iterators and borrowing allow for zero-copy abstractions and wrappers, so they are there for performance. But `RangeArg` is not about performance. I believe we should remove it and just use `SubRange` from #3165 consistently instead.

kvark:
Fixes #3166 

Having `Range<Option<X>>` was not very good, since it doesn't make sense for the offset to be optional (instead, it's obvious `0` can be used).

The PR also merges the `XxxTexelBuffer` descriptors, since having them different didn't make sense in the world where descriptor types are different (i.e. all the difference is in the type, not in the descriptor itself).

PR checklist:


kvark:
@msiglreith thank you for taking a look!
bors r=msiglreith

PR #3168, opened 2020-03-20
Read-only image descriptors

### LINKED ISSUE #3126: Fix Warden to work without timer queries
Currently, Warden can't be used for reftesting and benchmarking on many platforms because it assumes the support for time queries. We need to make it more flexible. This will open the road to enable it on Travis for CI headless.

In order to repro, either run `make reftests` if you are not on Windows, or specifically `cd src/warden && cargo run --bin reftest --features gl -- ci` if you are on Windows.

### LINKED ISSUE #3164: Features for read-only storage descriptors
Knowing if something is read-only is a good way for us to do less work. Both direct users of gfx-rs and wgpu *do* know that, and would like to communicate it to us. However, Vulkan Portability doesn't know about that and would like us to operate in the blind.

I wonder if we can support this by having a "READONLY_STORAGE" feature that VkPI will not enable.

Another way to address this would be having a "PORTABILITY" feature - this has the benefit of the "default" being more restricted and closer to WebGPU.

We could also consider refactoring the descriptor types to add:
  - read-only image storage
  - comparison samplers

kvark:
Fixes #3164
Fixes #3126
PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -3193,9 +3193,12 @@ impl From<pso::DescriptorType> for DescriptorContent {
                 ty: Idt::Sampled {
                     with_sampler: false,
                 },
-            }
-            | Dt::InputAttachment => DescriptorContent::SRV,
-            Dt::Image { ty: Idt::Storage } => DescriptorContent::SRV | DescriptorContent::UAV,
+            } |
+            Dt::Image { ty: Idt::Storage { read_only: true } } |
+            Dt::InputAttachment => DescriptorContent::SRV,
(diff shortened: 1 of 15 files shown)


kvark:
Damn, looks like Serde doesn't like 128-bit formats?
>thread 'main' panicked at 'failed to open/parse the suite: "/Users/dmalyshau/Code/gfx/src/warden/../../work/reftests/local.ron": Message("u128 is not supported")', src/libcore/result.rs:1188:5

kvark:
bors retry

kvark:
Bors retry

PR #3169, opened 2020-03-21
Basic debug markers API

kvark:
Addresses the "Section/call marking" API portion of #1508 (only Vulkan implementation so far)
PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -2288,6 +2288,16 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
     {
         unimplemented!()
     }
+
+    unsafe fn insert_debug_marker(&mut self, _name: &str, _color: u32) {
+        //TODO
+    }
+    unsafe fn begin_debug_marker(&mut self, _name: &str, _color: u32) {
+        //TODO
+    }
+    unsafe fn end_debug_marker(&mut self) {
+        //TODO
+    }
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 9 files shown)


kvark:
Question to @msiglreith and @grovesNL : should we work with `&CStr` for all the labels and markers? This was suggested by @aloucks in https://github.com/gfx-rs/wgpu/pull/528#discussion_r395944840

msiglreith:
Having a `CStr` variant as well sounds fine to me - wgpu can use `CStr` and others might want to go with `&str`

kvark:
@msiglreith you are saying - have both `CStr` and `str`? Or something like https://github.com/gfx-rs/wgpu/pull/528#discussion_r395941964 ?

msiglreith:
@kvark one function (`CStr`) for the backends to implement and an additional which takes a `&str` and forwards it to the first function

kvark:
@msiglreith so for Metal accessed from Rust code, it would mean 2 conversions:
  1. from `str` to `CStr`
  2. from `CStr` to `NSString`

kvark:
bors try

kvark:
Let's keep thinking about this. There doesn't appear to be an optimal solution yet.
bors r+

kvark:
what's going on with CI? both Travis and bors appear to go crazy, even though it's weekend time.
bors retry

aloucks:
All the builds have passed.. just not on a single run. It's always one of the mac builds that times out.

kvark:
bors retry

grovesNL:
Other users have the same issue in https://travis-ci.community/t/macos-build-an-error-occurred-while-generating-the-build-script/7684

kvark:
bors retry

kvark:
Bors retry

PR #3175, opened 2020-03-23
Pass Extend by mutable reference

kvark:
Fixes the use of Extend. The last version was wrong because we can't pass Extend by reference if a value was expected.
Also has a commit for formatting pass: having the code formatted will help aligning the hal-0.5 with master in the future.
PR checklist:


kvark:
I believe this is not controversial.
bors r+

kvark:
bors retry

PR #3178, opened 2020-03-25
[gl] add missing R/Rg/Rgba16Unorm format conversions

simonrepp:
Hi gfx team!

In my project I'm using `gfx_hal::format::Rg16Unorm`, passing it to `device.create_image(...)` as a format at some point, however as I discovered there is not yet any conversion for that format defined in the gl backend in `conv.rs` (which later leads to a crash at runtime):

https://github.com/gfx-rs/gfx/blob/fdfe887f5afab7ca057e3e824817d0bcdcf794ba/src/backend/gl/src/conv.rs#L95-L101

I made a naive attempt to add a mapping myself, based on deductions from the other entries in the mapping table and eventually experimenting with different combinations, the last one I arrived at being the one in this draft PR:

```rust
 Rg16Unorm => FormatDescription::new(glow::RG16, glow::RG_INTEGER, glow::UNSIGNED_SHORT, 2, Float),
```

However every variant I tried leads to an `Unsupported uniform datatype!` panic at a later point, right here:

https://github.com/gfx-rs/gfx/blob/fdfe887f5afab7ca057e3e824817d0bcdcf794ba/src/backend/gl/src/queue.rs#L938-L945

I'd love to make myself useful here and provide a fully working patch but right now I fear I'm stumbling around in the dark a bit, so glad for any pointers how I could continue this!

---

PR checklist:

This currently fails elsewhere (in the Vulkan backend) so I can't tell right now:
```
Testing Vulkan:
	Scene 'vertex-offset':
thread 'main' panicked at 'attempt to shift right with overflow', src/backend/vulkan/src/command.rs:36:15
```

Not working yet, see above, I'm testing this against a project that uses gfx-hal (https://codeberg.org/suchbliss/good-form), the examples won't likely cover this as this is about unimplemented features.

### DIFF
--- a/src/backend/gl/src/conv.rs
+++ b/src/backend/gl/src/conv.rs
@@ -151,6 +151,7 @@ pub fn describe_format(format: Format) -> Option<FormatDescription> {
         ),
         R16Sint => FormatDescription::new(glow::R16I, glow::RED_INTEGER, glow::SHORT, 1, Integer),
         R16Sfloat => FormatDescription::new(glow::R16, glow::RED, glow::HALF_FLOAT, 1, Float),
+        R16Unorm => FormatDescription::new(glow::R16, glow::RED, glow::UNSIGNED_SHORT, 1, Float),
         Rg16Uint => FormatDescription::new(
             glow::RG16UI,
             glow::RG_INTEGER,
@@ -159,6 +160,7 @@ pub fn describe_format(format: Format) -> Option<FormatDescription> {
             Integer,
         ),
         Rg16Sint => FormatDescription::new(glow::RG16I, glow::RG_INTEGER, glow::SHORT, 2, Integer),
+        Rg16Unorm => FormatDescription::new(glow::RG16, glow::RG, glow::UNSIGNED_SHORT, 2, Float),
         Rg16Sfloat => FormatDescription::new(glow::RG16, glow::RG, glow::HALF_FLOAT, 2, Float),
         Rgba16Uint => FormatDescription::new(
             glow::RGBA16UI,
@@ -172,6 +174,9 @@ pub fn describe_format(format: Format) -> Option<FormatDescription> {
         }
         Rgba16Sfloat => {
             FormatDescription::new(glow::RGBA16, glow::RGBA, glow::HALF_FLOAT, 4, Float)
+        }
+        Rgba16Unorm => {
+            FormatDescription::new(glow::RGBA16, glow::RGBA, glow::UNSIGNED_SHORT, 4, Float)
         }
         R32Uint => FormatDescription::new(
             glow::R32UI,


src/backend/gl/src/conv.rs
```diff
@@ -159,6 +159,7 @@ pub fn describe_format(format: Format) -> Option<FormatDescription> {
             Integer,
         ),
         Rg16Sint => FormatDescription::new(glow::RG16I, glow::RG_INTEGER, glow::SHORT, 2, Integer),
+        Rg16Unorm => FormatDescription::new(glow::RG16, glow::RG_INTEGER, glow::UNSIGNED_SHORT, 2, Float),
```
kvark: this should use `glow::RG` instead, it's not an integer format

simonrepp: Hey thanks for the quick feeback, appreciate it!

I assume you're referring to the second parameter (`tex_external`), however no matter which of the two was meant, with either change I run into the second, later issue I described above. I've tried quite a few combinations and I'm unsure if it's really just this line that has an issue or if there's other code that needs adapting (there are only conversions for *8Unorms currently, I was wondering if that is maybe because for the other variants - 16/32 there would have to be more changes elsewhere?)

If your feeling is that this line in itself should be fine to enable usage or Rg16Unorm data, I'll hunt the other issue down the stack some more, because maybe in the end that other one is completely unrelated and I'm just wasting our time, which I really don't want to do. :) ... Thanks for your help!

kvark:
wow the reftests fail for you on Vulkan, please file an issue about this!

src/backend/gl/src/conv.rs
kvark: Are you referring to "Unsupported uniform datatype!" panic? How do we get there?
Do you have a repro case that I could try as well?

simonrepp: Yes that one! I can look into the exact code path up to there a bit later (however as I said maybe this is really unrelated after all, might have been to quick in assuming a connection).

Re. repro: You can check out the `gl` branch of https://codeberg.org/suchbliss/good-form and run with `cargo run --features=gl`, which should start and then immediately crash with the mentioned error. (It's not exactly a minimal repro though I'm afraid ;))

simonrepp:
Will gladly file an issue on the failing reftests later! :)

src/backend/gl/src/conv.rs
simonrepp: Oh re. repro l just realized with all the audio dependencies involved and whatnot ... I'm not sure the repro procedure on the whole project will be approachable depending on your system, I'll definitely look into it later anyhow!

simonrepp:
- I added the other two obvious (R/Rgba16Unorm) conversions to make this PR worth our while :)
- I further investigated the "Unsupported uniform datatype" issue and it turns out this is apparently really something else, I would open another issue for that if you concur with that (to me it looks like either a bug or maybe something about the DescriptorType changes in 0.5.0 that's not entirely clear to me yet)

simonrepp:
Opened an issue for the reftests failing at #3183

src/backend/gl/src/conv.rs
```diff
@@ -151,6 +151,7 @@ pub fn describe_format(format: Format) -> Option<FormatDescription> {
         ),
         R16Sint => FormatDescription::new(glow::R16I, glow::RED_INTEGER, glow::SHORT, 1, Integer),
         R16Sfloat => FormatDescription::new(glow::R16, glow::RED, glow::HALF_FLOAT, 1, Float),
+        R16Unorm => FormatDescription::new(glow::R16, glow::RED, glow::UNSIGNED_SHORT, 2, Float),
```
msiglreith: should be `1` component

simonrepp: Thanks for checking, fixed!

kvark:
bors r=kvark,msiglreith

PR #3198, opened 2020-04-02
[0.5/vk] fix handling of AMD viewport negation

kvark:
Addresses https://github.com/gfx-rs/wgpu-rs/issues/227
We are now taking MAINTENANCE1 whenever it's supported. This code path is supposed to be unaffected. However, if it's not supported, and NDC_Y_FLIP is requested, and we end up using AMD_NEGATIVE_VIEWPORT_HEIGHT, then we don't shift the Y of the viewport.

PR checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # Change Log
 
-### backend-metal-0.5.1 (26-02-2020)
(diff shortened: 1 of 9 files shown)


kvark:
bors r=wgpu

PR #3205, opened 2020-04-07
Fix clearing of GL depth buffer

### LINKED ISSUE #3202: GL depth buffer not being cleared
Short info header:
- GFX version: 0.5.0
- OS: Windows 10
- GPU: GTX 1070

When `glDepthMask` is set to false, clears do not work on depth buffer. This happens when last rendered item on the pipeline was transparent and `depth.write` was set to false. This prevents next frame from clearing the depth buffer. This does not happen on vulkan and depth buffer is cleared correctly.

I guess the easy fix is to just set depth mask to true in https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/queue.rs#L530 before clear? Restoring previous value is not neccessary as render pass will set it anyway

chemicstry:
Fixes #3202 

It seems that the same issue could apply to the stencil and color buffers too (they have their own `stencil_mask` and `color_mask` functions which are not reset), but it would take some time for me to test that.

PR checklist:

### DIFF
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -77,9 +77,7 @@ pub enum Command {
     BindRasterizer {
         rasterizer: pso::Rasterizer,
     },
-    BindDepth {
-        depth: Option<pso::DepthTest>,
-    },
+    BindDepth(Option<pso::Comparison>),
     SetViewports {
         first_viewport: u32,
         viewport_ptr: BufferSlice,
@@ -108,8 +106,8 @@ pub enum Command {
     SetDrawColorBuffers(usize),
     SetPatchSize(i32),
     BindProgram(<GlContext as glow::HasContext>::Program),
-    SetBlend(pso::ColorBlendDesc),
-    SetBlendSlot(ColorSlot, pso::ColorBlendDesc),
+    SetBlend(Option<pso::BlendState>),
+    SetBlendSlot(ColorSlot, Option<pso::BlendState>),
     BindAttribute(n::AttributeDesc, n::RawBuffer, i32, u32),
     //UnbindAttribute(n::AttributeDesc),
     CopyBufferToBuffer(n::RawBuffer, n::RawBuffer, command::BufferCopy),
@@ -148,6 +146,11 @@ pub enum Command {
     BindTexture(u32, n::Texture, n::TextureTarget),
     BindSampler(u32, n::Sampler),
     SetTextureSamplerSettings(u32, n::Texture, n::TextureTarget, image::SamplerDesc),
+
+    SetColorMask(Option<DrawBuffer>, pso::ColorMask),
+    SetDepthMask(bool),
+    SetStencilMask(pso::StencilValue),
+    SetStencilMaskSeparate(pso::Sided<pso::StencilValue>),
 }
 
 pub type FrameBufferTarget = u32;
@@ -198,6 +201,10 @@ struct Cache {
     attributes: Vec<n::AttributeDesc>,
     // Active uniforms
     uniforms: Vec<n::UniformDesc>,
+    // Current depth mask
+    depth_mask: Option<bool>,
+    // Current stencil mask
+    stencil_mask: Option<pso::Sided<pso::StencilValue>>,
 }
 
 impl Cache {
@@ -216,6 +223,8 @@ impl Cache {
             vertex_buffer_descs: Vec::new(),
             attributes: Vec::new(),
             uniforms: Vec::new(),
+            depth_mask: None,
+            stencil_mask: None,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -521,28 +521,20 @@ impl CommandQueue {
...
+                let gl = &self.share.context;
                 match (depth, stencil) {
                     (Some(depth), Some(stencil)) => {
-                        self.share.context.clear_buffer_depth_stencil(
-                            glow::DEPTH_STENCIL,
-                            0,
-                            depth,
-                            stencil as _,
-                        );
+                        gl.depth_mask(true);
```
kvark, line 527: are you sure we don't need to reset this back? gfx-rs API has 3 kind of clears, just like Vulkan:
  - explicit command outside of a render pass - you don't need to restore the mask in this case
  - load operation on an attachment when starting a render pass - you may need to restore the mask according to the actively bound pipeline
  - explicit attachment clear command inside a render pass - you do need to restore the mask

chemicstry: I don't have a good understandting of the whole API yet so thanks for pointing this out

kvark: if you can hop on #gfx:matrix.org, we'll be happy to explain in a more interactive way :)

src/backend/gl/src/command.rs
```diff
@@ -96,7 +96,12 @@ pub enum Command {
     /// Clear signed integer color drawbuffer of bound framebuffer.
     ClearBufferColorI(DrawBuffer, [i32; 4]),
     /// Clear depth-stencil drawbuffer of bound framebuffer.
-    ClearBufferDepthStencil(Option<pso::DepthValue>, Option<pso::StencilValue>),
+    ClearBufferDepthStencil {
+        depth: Option<pso::DepthValue>,
+        depth_mask: Option<bool>,
```
kvark, line 101: do we really need to have it here? Could we instead have separate commands for setting the depth mask and stencil masks (i.e. `SetDepthMask`, `SetStencilMask`), so that the command encoder would insert them if necessary into the command stream?

src/backend/gl/src/state.rs
```diff
@@ -87,7 +87,7 @@ pub(crate) fn bind_stencil(gl: &GlContainer, stencil: &Option<pso::StencilTest>,
                 );
                 if let pso::State::Static(values) = stencil.write_masks {
                     unsafe {
-                        gl.stencil_mask_separate(glow::FRONT, values.back);
+                        gl.stencil_mask_separate(glow::BACK, values.back);
```
kvark, line 90: nice!

src/backend/gl/src/command.rs
```diff
@@ -480,84 +480,121 @@ impl CommandBuffer {
...
+                assert_eq!(attachment.ops.load, pass::AttachmentLoadOp::Clear);
+
+                let channel = view_format.base_format().1;
+                let index = clear.index - glow::COLOR_ATTACHMENT0;
+
+                // Temporarily reset color mask if it was not ColorMask::ALL
+                let color_mask = self
+                    .cache
+                    .blend_targets
+                    .get(index as usize).map(Option::as_ref).flatten()
```
kvark, line 520: nit: `flat_map`?

src/backend/gl/src/command.rs
```diff
@@ -480,84 +480,121 @@ impl CommandBuffer {
...
-                    }
+                match channel {
+                    ChannelType::Unorm
+                    | ChannelType::Snorm
+                    | ChannelType::Ufloat
+                    | ChannelType::Sfloat
+                    | ChannelType::Srgb
+                    | ChannelType::Uscaled
+                    | ChannelType::Sscaled => clear_cmds
+                        .push(Command::ClearBufferColorF(index, unsafe {
```
kvark, line 535: let's refactor this as `let cmd = match channel {...}; clear_cmds.push(cmd);`

src/backend/gl/src/command.rs
```diff
@@ -480,84 +480,121 @@ impl CommandBuffer {
...
+
+                // Restore masks if they were reset
+                if let Some(mask) = depth_mask {
+                    clear_cmds.push(Command::SetDepthMask(mask));
+                }
+                if let Some(mask) = stencil_mask {
+                    clear_cmds.push(Command::SetStencilMaskSeparate(mask));
+                }
+            }
+        }
```
kvark, line 597: can we reorder the code in such a way that we could just live-add the clear commands (with `self.push_cmd`) instead of collecting them and the flushing them? I.e. can we move the `BindFrameBuffer` and `DrawBuffers` before this giant block?

src/backend/gl/src/command.rs
```diff
@@ -757,6 +794,17 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
...
+                    .cache
+                    .blend_targets
+                    .get(0).map(Option::as_ref).flatten()
+                    .map(|blend_target| blend_target.mask)
+                    .filter(|mask| *mask != pso::ColorMask::ALL);
+                if color_mask.is_some() {
+                    self.push_cmd(Command::SetColorMask(Some(0), pso::ColorMask::ALL));
+                }
+
                 match image.channel {
```
kvark, line 808: similarly, let's refactor this to do `let cmd = match {...};`

src/backend/gl/src/command.rs
```diff
@@ -154,6 +149,11 @@ pub enum Command {
     BindTexture(u32, n::Texture, n::TextureTarget),
     BindSampler(u32, n::Sampler),
     SetTextureSamplerSettings(u32, n::Texture, n::TextureTarget, image::SamplerDesc),
+
+    SetColorMask(Option<DrawBuffer>, pso::ColorMask),
```
kvark, line 153: could we also use them for implementing `bind_graphics_pipeline`? this would make it a single-responsibility path

chemicstry: sure!

src/backend/gl/src/command.rs
chemicstry: I was about to do that, but you can't call `self.push_cmd` as it requires `&mut self` and self was already borrowed to access `self.pass_cache`.

kvark: do you need to borrow the cache though?

chemicstry:
Wow this bot spams, could just notify about forgotten `cargo fmt`.

Sorry for another rewrite, but to fix the requested non-queued `push_cmd` I had to refactor `push_cmd` itself. I separated buffer storage into a separate struct so `push_cmd` does not need to borrow entire `CommandBuffer`. This also removes the ugly `push_cmd_internal` used to avoid borrow errors. If this is too much I can just revert to queueing commands into temporary vec.

src/backend/gl/src/command.rs
```diff
@@ -235,17 +244,94 @@ impl From<hal::Limits> for Limits {
     }
 }
 
-/// A command buffer abstraction for OpenGL.
-///
-/// If you want to display your rendered results to a framebuffer created externally, see the
-/// `display_fb` field.
 #[derive(Debug)]
-pub struct CommandBuffer {
+pub struct BufferStorage {
```
kvark, line 248: let's rename it to something to avoid confusion with buffer storage descriptors

src/backend/gl/src/command.rs
```diff
@@ -235,17 +244,94 @@ impl From<hal::Limits> for Limits {
...
+
+impl BufferStorage {
+    fn push_cmd(&mut self, cmd: Command) {
+        let mut memory = self
+            .memory
+            .try_lock()
+            .expect("Trying to record a command buffers, while memory is in-use.");
+
+        let cmd_buffer = match *memory {
+            BufferMemory::Linear(ref mut buffer) => &mut buffer.commands,
```
kvark, line 264: if you just do `=> buffer` here and `=> storage.get_mut(&self.id).unwrap()` in the other arm, it will be less verbose

src/backend/gl/src/command.rs
```diff
@@ -235,17 +244,94 @@ impl From<hal::Limits> for Limits {
...
+    }
+}
+
+/// A command buffer abstraction for OpenGL.
+///
+/// If you want to display your rendered results to a framebuffer created externally, see the
+/// `display_fb` field.
+#[derive(Debug)]
+pub struct CommandBuffer {
+    pub(crate) buffer: BufferStorage,
```
kvark, line 334: maybe something like `data: CommandStorage` or something?

chemicstry: sure, I was a bit undecisive of name. Will make it `CommandStorage`

chemicstry:
Had a hiccup with squash but should be ok now

kvark:
bors r-
actually, Travis looks really sad

chemicstry:
My bad, some errors appeard due to merge with master. Travis succeeded now.

PR #3212, opened 2020-04-12
[0.4] parking lot update for Metal and DX11

kvark:
Also required for a smooth update in gecko...
When dependencies change, Gecko engineering wants it to happen across all crates, preferably without functional changes


kvark:
bors r=gecko

PR #3214, opened 2020-04-14
Fix index out of bounds access when using object labels longer than 64 characters

### LINKED ISSUE #3213: Out of bounds array access when passing a buffer name with a length >= 64
Passing a label name (for e.g. a buffer) with a length greater or equal to 64 will cause the code to try to write to an index outside the length of the array. This is fine in debug mode when bounds checks are enabled, but in release mode this will cause undefined behaviour.

I'm not sure what should happen in this case, either it could silently truncate the string to 64 characters or it could panic.

https://github.com/gfx-rs/gfx/blob/deeb135553dccbc654ce105c042aa07b4cdbdd23/src/backend/vulkan/src/device.rs#L2284

HalfVoxel:
Fixes #3213.

I have not been able to test this change unfortunately since I have been unable to find a combination of wgpu, wgpu-rs and gfx that work without errors.

PR checklist:

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -2274,14 +2274,30 @@ impl Device {
     unsafe fn set_object_name(&self, object_type: vk::ObjectType, object_handle: u64, name: &str) {
         let instance = &self.shared.instance;
         if let Some(DebugMessenger::Utils(ref debug_utils_ext, _)) = instance.1 {
-            // Append a null terminator to the string while avoiding allocating memory
-            static mut NAME_BUF: [u8; 64] = [0u8; 64];
-            std::ptr::copy_nonoverlapping(
-                name.as_ptr(),
-                &mut NAME_BUF[0],
-                name.len().min(NAME_BUF.len()),
-            );
-            NAME_BUF[name.len()] = 0;
+            // Keep variables outside the if-else block to ensure they do not
+            // go out of scope while we hold a pointer to them
+            let mut buffer: [u8; 64] = [0u8; 64];
+            let mut buffer_vec: Vec<u8>;
+
+            // Append a null terminator to the string
+            let name_ptr = if name.len() < 64 {
+                // Common case, string is very small. Allocate a copy on the stack.
+                std::ptr::copy_nonoverlapping(name.as_ptr(), buffer.as_mut_ptr(), name.len());
+                // Add null terminator
+                buffer[name.len()] = 0;
+                buffer.as_mut_ptr()
+            } else {
+                // Less common case, the string is large.
+                // This requires a heap allocation.
+                buffer_vec = name
+                    .as_bytes()
+                    .iter()
+                    .cloned()
+                    .chain(std::iter::once(0))
+                    .collect::<Vec<u8>>();
+                buffer_vec.as_mut_ptr()
+            };
+
             let _result = debug_utils_ext.debug_utils_set_object_name(
                 self.shared.raw.handle(),
                 &vk::DebugUtilsObjectNameInfoEXT {
@@ -2289,7 +2305,7 @@ impl Device {
                     p_next: std::ptr::null_mut(),
                     object_type,
                     object_handle,
-                    p_object_name: NAME_BUF.as_ptr() as *mut _,
+                    p_object_name: name_ptr as *const i8,
                 },
             );
         }


src/backend/vulkan/src/device.rs
```diff
@@ -2274,24 +2274,34 @@ impl Device {
...
+                        p_next: std::ptr::null_mut(),
+                        object_type,
+                        object_handle,
+                        p_object_name: name_ptr as *const i8,
+                    },
+                );
+            };
+
+            // Append a null terminator to the string
+            if name.len() < 64 {
```
kvark, line 2291: let's do `let name_ptr = if ... else ...;` so that we don't have to have a helper function.
Note that you'd need to move the declaration out of if/else:
```rust
let mut buf_vect: Vec<u8>;
if ... {...} else {
buffer = (name.as_bytes().iter().cloned().chain(std::iter::once(0)))
                    .collec();
    buf_vect.as_mut_ptr()
};
```

src/backend/vulkan/src/device.rs
```diff
@@ -2274,24 +2274,34 @@ impl Device {
...
+                // Common case, string is very small. Allocate a copy on the stack.
+                let mut buffer: [u8; 64] = [0u8; 64];
+                std::ptr::copy_nonoverlapping(name.as_ptr(), &mut buffer[0], name.len());
```
kvark, line 2294: nit: can we use `buffer.as_mut_ptr()` instead?

src/backend/vulkan/src/device.rs
HalfVoxel: Yeah that's probably better.

src/backend/vulkan/src/device.rs
HalfVoxel: Yup.

src/backend/vulkan/src/device.rs
```diff
@@ -2274,22 +2274,38 @@ impl Device {
...
-                name.len().min(NAME_BUF.len()),
-            );
-            NAME_BUF[name.len()] = 0;
+            // Keep variables outside the if-else block to ensure they do not
+            // go out of scope while we hold a pointer to them
+            let mut buffer: [u8; 64] = [0u8; 64];
+            let mut buffer_vec: Vec<u8>;
+
+            // Append a null terminator to the string
+            let name_ptr = if name.len() < 64 {
```
kvark, line 2283: nit: we could do `if name.len() < buffer.len()` to avoid repetition

src/backend/vulkan/src/device.rs
```diff
@@ -2274,22 +2274,38 @@ impl Device {
...
-            static mut NAME_BUF: [u8; 64] = [0u8; 64];
-            std::ptr::copy_nonoverlapping(
-                name.as_ptr(),
-                &mut NAME_BUF[0],
-                name.len().min(NAME_BUF.len()),
-            );
-            NAME_BUF[name.len()] = 0;
+            // Keep variables outside the if-else block to ensure they do not
+            // go out of scope while we hold a pointer to them
+            let mut buffer: [u8; 64] = [0u8; 64];
```
kvark, line 2279: type can be inferred here, not needed

src/backend/vulkan/src/device.rs
```diff
@@ -2274,22 +2274,38 @@ impl Device {
...
+                buffer.as_mut_ptr()
+            } else {
+                // Less common case, the string is large.
+                // This requires a heap allocation.
+                buffer_vec = name
+                    .as_bytes()
+                    .iter()
+                    .cloned()
+                    .chain(std::iter::once(0))
+                    .collect::<Vec<u8>>();
```
kvark, line 2297: It just ocurred to be that there is already a primitive for what we are doing here (on-stack | heap allocation), it's called `SmallVec`. So perhaps, we can skip most of this code and just do
```rust
let label_vec = name
                    .as_bytes()
                    .iter()
                    .cloned()
                    .chain(std::iter::once(0))
                    .collect::<SmallVec<[u8; 64]>>();
```
and be done with it?

HalfVoxel:
I thought about SmallVec too, however it seems likely that it would be a lot slower. I'm thinking it would first try to copy it to a stack allocated 64 byte buffer and then when it realizes that this will not work it will copy it to the heap allocated buffer. So it might require two copies instead of 1.
Maybe it is smart enough to be compiled to the same thing though. I'll try to see what rustc generates.

HalfVoxel:
The performance difference is even larger than I thought. For small strings (10 characters or so) the custom buffer is about 10 times faster than SmallVec. For large buffers (65 characters) its about 5 times faster.

```
smallvec_small          time:   38.203 ns
smallvec_large          time:   77.385 ns
custom_buffer_small     time:   2.9266 ns
custom_buffer_large     time:   16.082 ns
```

Here is the testing code I used: https://pastebin.com/eRtdpMy9

Even smallvec isn't really *slow* though, so maybe it's just premature optimization.

kvark:
Wow interesting. Where you measuring with Release?
What I'd ideally expect to happen is that `as_bytes().iter()` returns an iterator that knows about its length, and so it can hint to `SmallVec` that the heap allocation is needed before any bytes are copied. Perhaps, it's just something that's missing from `SmallVec` atm.

HalfVoxel:
> Wow interesting. Where you measuring with Release?
Yup. I used the `criterion` library which does all kinds of warmups and stuff to reduce noise too.

Yeah that's what I would expect in the best case too, given rust's goal of no-overhead-iterators. But it seems like it is not supported by some part of the stack.

kvark:
We need to back-port this to hal-0.5 branch, to address https://github.com/gfx-rs/wgpu-rs/issues/289

PR #3217, opened 2020-04-16
Add Missing glutin export

zicklag:
Fixes #issue
PR checklist:
- [ ] `make reftests` succeeds ( can't test because I don't have Vulkan )
- [x] tested examples with the following backends: `gl`

Review the latest commit alone if you don't want to see all the rustfmt clutter. :wink:

This adds an extra export to the `glutin` backend for GL which was accidentally removed with the surfman update. This will keep code that is still using the `glutin` backend working.

### DIFF
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -39,11 +39,9 @@ pub use window::web::{Surface, Swapchain};
 
 // Glutin implementation
 #[cfg(glutin)]
-pub use crate::window::glutin::{Instance, Surface, Swapchain};
+pub use crate::window::glutin::{Headless, Instance, Surface, Swapchain, config_context};
 #[cfg(glutin)]
 pub use glutin;
-#[cfg(glutin)]
-pub use crate::window::glutin::config_context;
 
 // Surfman implementation
 #[cfg(surfman)]


src/backend/gl/src/queue.rs
```diff
@@ -313,11 +322,12 @@ impl CommandQueue {
             };
         } else if self.state.num_viewports > 1 {
             // 16 viewports is a common limit set in drivers.
-            let viewports: SmallVec<[[f32; 4]; 16]> = (0..self.state.num_viewports)
+            let viewports: SmallVec<[[f32; 4]; 16]> = (0 .. self.state.num_viewports)
```
kvark, line 325: why are we changing this back? I like the spaces, but rustfmt stable doesn't :/

zicklag: I was just using nightly and rustfmt so that it matched what was in the `rustfmt.toml`. We could use stable or we could take out the nightly-only formatting options from the toml so that they both yield the same formatting ( which would probably be best ).

kvark: we should take out the nightly options, yeah (unfortunately)

zicklag: I'll ad that to this PR then.

kvark: Ideally, to a different PR, if possible.
This one just doesn't need to touch the formatting at all :)

zicklag: Ah, OK. No problem. :)

zicklag:
This is ready now.

PR #3229, opened 2020-04-26
Improve documentation

GabrielMajeri:
This PR improves the docs for various components of the public API. I also improved the crate-level documentation to give an overview of the device initialization flow.

The objective for me was to document everything on the "critical path". This means every item someone with no prior `gfx-hal` or Vulkan experience would require in order to do something simple, such as [rendering a triangle on screen](https://www.falseidolfactory.com/2020/04/01/intro-to-gfx-hal-part-1-drawing-a-triangle.html).

### DIFF
--- a/src/hal/src/adapter.rs
+++ b/src/hal/src/adapter.rs
@@ -1,10 +1,12 @@
-//! Physical devices and adapters.
+//! Physical graphics devices.
 //!
-//! The `PhysicalDevice` trait specifies the API a backend must provide for dealing with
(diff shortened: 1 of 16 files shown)


GabrielMajeri:
> There is a minor suggestion to not use `graphics` as explicitly, but I think we can go ahead with this.

Would saying something like "graphics and _compute_ devices" be better?

kvark:
Possibly? Or just a "GPU"

Issue #3027, opened 2019-10-01
wgpu backend


kvark:
We need a backend developed to use https://github.com/gfx-rs/wgpu-rs , should be called `gfx-backend-wgpu`.

The purpose of this backend would solely be targeting the web: native capabilities of wgpu are not beneficial here. Even though today wgpu-rs doesn't target the web, we can already establish the gfx-rs backend so that we are ready when the time comes.

In terms of difficulty, I expect the logic to be similar to gfx-backend-metal. Of course, we don't expect it to perform as well as targeting the backends directly.

kvark:
Part of the task here is figuring out how to test this. Being able to run [gfx_ocean](https://github.com/msiglreith/gfx_ocean) is one of the goals.

kvark:
Copied from https://github.com/gfx-rs/gfx/pull/3231#issuecomment-619543057 :

So the main goal of this backend is to target the Web. This means, we don't care in particular about wgpu-core or even wgpu-native (even though that one could go through emscripten to the Web). Instead, we care about the path that wgpu-rs uses for the Web backend, which is web-sys.

Now, if you just try to link to wgpu-rs, you'll be forced to deal with the fact that this is meant to be used by the end users and not intermediate layers. The API can be restrictive, making this backend impossible. Therefore, I believe the only sensible path for gfx-backend-wgpu is to link to web-sys directly and calling the web methods, without linking to any of the wgpu ecosystem crates at all.

kvark:
Also see https://github.com/webgpu-native/webgpu-headers/issues/47 as highly relevant to the way we get async functions.

PR #3231, opened 2020-04-26
Add WebGPU backend

GabrielMajeri:
This PR adds a WebGPU backend for `gfx-hal`, making some progress towards #3027.

I've currently gotten the backend to compile. Further additions will come in future PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = [
     "src/backend/gl",
     "src/backend/metal",
     "src/backend/vulkan",
+    "src/backend/webgpu",
     "src/hal",
     "src/warden",
     "examples",
--- a/src/backend/webgpu/Cargo.toml
+++ b/src/backend/webgpu/Cargo.toml
@@ -1,0 +1,113 @@
+[package]
+name = "gfx-backend-webgpu"
+version = "0.5.0"
+description = "WebGPU backend for gfx-rs"
+homepage = "https://github.com/gfx-rs/gfx"
+repository = "https://github.com/gfx-rs/gfx"
+keywords = ["graphics", "gamedev"]
+license = "MIT OR Apache-2.0"
+authors = ["The Gfx-rs Developers"]
+readme = "README.md"
+documentation = "https://docs.rs/gfx-backend-webgpu"
+workspace = "../../.."
+edition = "2018"
+
+[features]
+default = []
+
+[lib]
+name = "gfx_backend_webgpu"
+
+[dependencies]
+hal = { path = "../../hal", version = "0.5", package = "gfx-hal" }
+js-sys = "0.3.37"
+raw-window-handle = "0.3"
+wasm-bindgen = "0.2.60"
+wasm-bindgen-futures = "0.4.10"
+web-sys = { version = "0.3.37", features = [
+    "Document",
+    "Navigator",
+    "Node",
+    "NodeList",
+    "Gpu",
+    "GpuAdapter",
+    "GpuAddressMode",
+    "GpuBindGroup",
+    "GpuBindGroupEntry",
+    "GpuBindGroupDescriptor",
+    "GpuBindGroupLayout",
+    "GpuBindGroupLayoutEntry",
+    "GpuBindGroupLayoutDescriptor",
+    "GpuBlendDescriptor",
+    "GpuBlendFactor",
+    "GpuBlendOperation",
+    "GpuBindingType",
+    "GpuBuffer",
+    "GpuBufferBinding",
+    "GpuBufferCopyView",
+    "GpuBufferDescriptor",
+    "GpuCanvasContext",
+    "GpuColorDict",
+    "GpuColorStateDescriptor",
+    "GpuCommandBuffer",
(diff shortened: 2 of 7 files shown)


kvark:
Oh wow, you are attacking this task, this is great news!
Please note that #3027 doesn't explain clearly how this needs to be done. We didn't understand all the details ourselves. So let me at least try to do a better job here.

So the main goal of this backend is to target the Web. This means, we don't care in particular about `wgpu-core` or even `wgpu-native` (even though that one could go through emscripten to the Web). Instead, we care about the path that `wgpu-rs` uses for the Web backend, which is `web-sys`.

Now, if you just try to link to `wgpu-rs`, you'll be forced to deal with the fact that this is meant to be used by the end users and not intermediate layers. The API can be restrictive, making this backend impossible. Therefore, I believe the only sensible path for gfx-backend-wgpu is to link to `web-sys` directly and calling the web methods, without linking to any of the `wgpu` ecosystem crates at all.

Does that make sense? @grovesNL can correct if I said this wrong.

GabrielMajeri:
Yeah, makes sense. I will look into switching to `web-sys`. 

The only problem I've encountered so far, is that some of the `WebGPU` APIs are `async`. How do we expose that through the `gfx-hal` API?

kvark:
That's one of the architectural blockers to resolve, it could be a serious issue.
I wonder if we can just have an executor blocking on that? I.e. to start, just have an executor being a part of the backend, and call [block_on](https://docs.rs/futures/0.3.4/futures/executor/fn.block_on.html) on the `wasm_bindgen_futures::JsFuture` that is supposed to be async.
@grovesNL ^

grovesNL:
@kvark we won't be able to use `block_on` on the web (e.g. https://github.com/rustwasm/wasm-bindgen/issues/1733#issuecomment-552038681)

GabrielMajeri:
Does this mean we'll have to modify the `gfx-hal` API to be `async`?

kvark:
No, we aren't going to modify gfx-hal API here.
We need to figure out what to...

grovesNL:
I'm not sure of the best way to handle this, some options could be:
- modify gfx-hal's API
- expose async functions as a feature specific to WebGPU, so we'd have separate entry points specific to WebGPU
- work with promises/futures internally in most functions/types, so functions are implemented behind promise chains (e.g. `device_promise.then(device => device.xyz())`)
- try to find some way to yield the browser temporarily at the Rust level (like async/await itself does internally), similar to Emscripten's Asyncify https://emscripten.org/docs/porting/emscripten-runtime-environment.html#using-asyncify-to-yield-to-the-browser

Maybe there are other ways to handle this?

GabrielMajeri:
The easiest solution for now seems to be to expose separate entry points specific for WebGPU (at least temporarily, until a better solution is proposed).

I could make the `open` function panic when called, telling people to instead use `open_async` or something like that on WASM.

Would that be acceptable?

kvark:
@GabrielMajeri this would let you to proceed and get something working. Similar to how we have OpenGL API somewhat working but requiring application changes for a while (but no more). When we find a proper solution for async, we wouldn't need to scrap a lot of code, so there is nothing to lose if we go this direction here.

GabrielMajeri:
@kvark Alright, I've started work on that. Would it be OK to get this merged as-is (since its just boilerplate code), and add instance creation in the next PR, for easier review?

kvark:
Yeah, as long as CI is happy :)

GabrielMajeri:
Ok! I've looked at the logs for the failing build. It seems to be an ICE on nightly:

```rust
error: internal compiler error: src/librustc_mir/interpret/validity.rs:396: Unexpected error during ptr inbounds test: cannot read from foreign (extern) static DefId(0:40 ~ dispatch[5c52]::ffi[0]::[0]::_dispatch_queue_attr_concurrent[0])
```

I'll see if it's already reported.

kvark:
@GabrielMajeri also, let's please rename this to `gfx-backend-webgpu` in folder "webgpu". This matches "vulkan" and "metal" better, since they aren't "vk" and "mtl". We also don't want to confuse people with the idea that this is going through `wgpu` library, since we are going to target the Web directly.

GabrielMajeri:
@kvark Noted, renamed everything from `wgpu` to `webgpu`.

kvark:
I'm disabling monocodus warnings in #3234. Sorry about the mess it makes!

kvark:
bors try

kvark:
Looks like the nightly failure will be fixed in the next nightly...

kvark:
bors try

kvark:
@GabrielMajeri sorry, looks like it needs to be updated now that Mesh shaders have landed. It should be fairly straightforward though. Would you mind rebasing and updating it?
```
error[E0046]: not all trait items implemented, missing: `draw_mesh_tasks`, `draw_mesh_tasks_indirect`, `draw_mesh_tasks_indirect_count`
   --> src/backend/webgpu/src/command.rs:134:1
```

GabrielMajeri:
No problem, fixed now.

PR #3232, opened 2020-04-27
[0.5/vk] gracefully handle the case where the drive supports it but hardware does not

kvark:
Required for https://bugzilla.mozilla.org/show_bug.cgi?id=1633472
PR checklist:


kvark:
I think this is straightforward.
bors r+

kvark:
this works, it seems
bors r+

PR #3236, opened 2020-04-28
WIP: Initial stub for mesh shaders

MatusT:
This is my attempt at adding support for mesh shaders. 

This (first) PR only stubs the API. From my investigation, the extension is relatively simple(it effectively only adds new shaders and 3 draw commands mirroring the original ones) and seems to be the same across Direct3D and Vulkan. 

Even though the Vulkan only has NV extension so far, based on announcements of RDNA2 and DirectX Ultimate we can expect KHR version and I don't see how it could diverge in any way. Thus, both Vulkan and DX12 implementations should be trivial.

I know nothing about Metal and I am not aware of information about support for mesh shaders. At worst, they could be emulated using compute shaders but It would probably be so slow that It would be practically worthless.

Tasks:
- [x] Stub the API
- [x] Implement empty backend
- [x] Implement non-functional Vulkan backend
- [x] Implement non-functional DirectX 12 backend
- [x] Implement non-functional DirectX 11 backend
- [x] Implement non-functional Metal backend
- [x] Implement non-functional OpenGL backend
- [x] Fix examples compilation errors
- [x] Fix warden compilation errors

PR checklist:
- [ ] tested examples with the following backends
  - [x] DX12
  - [x] Vulkan
  - [x] OpenGL
  - [ ] Metal (have no way of testing)

Resources:
[Vulkan specification of mesh shaders](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap46.html#VK_NV_mesh_shader)
[DirectX specification of mesh shader](https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html)

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -1255,51 +1255,59 @@ impl<B: Backend> PipelineState<B> {
                     },
                 );
 
-                let shader_entries = pso::GraphicsShaderSet {
-                    vertex: vs_entry,
-                    hull: None,
-                    domain: None,
-                    geometry: None,
-                    fragment: Some(fs_entry),
-                };
-
                 let subpass = pass::Subpass {
                     index: 0,
                     main_pass: render_pass,
                 };
 
+                let vertex_buffers = vec![
+                    pso::VertexBufferDesc {
+                        binding: 0,
+                        stride: size_of::<Vertex>() as u32,
+                        rate: pso::VertexInputRate::Vertex,
+                    }
+                ];
+
+                let attributes = vec![
+                    pso::AttributeDesc {
+                        location: 0,
+                        binding: 0,
+                        element: pso::Element {
+                            format: f::Format::Rg32Sfloat,
+                            offset: 0,
+                        },
+                    },
+                    pso::AttributeDesc {
+                        location: 1,
+                        binding: 0,
+                        element: pso::Element {
+                            format: f::Format::Rg32Sfloat,
+                            offset: 8,
+                        },
+                    },
+                ];
+
                 let mut pipeline_desc = pso::GraphicsPipelineDesc::new(
-                    shader_entries,
-                    pso::Primitive::TriangleList,
+                    pso::PrimitiveAssembler::Vertex {
+                        buffers: vertex_buffers,
+                        attributes,
+                        input_assembler: pso::InputAssemblerDesc {
+                            primitive: pso::Primitive::TriangleList,
+                            with_adjacency: false,
+                            restart_index: None,
+                        },
+                        vertex: vs_entry,
+                        geometry: None,
+                        tessellation: None,
+                    },
                     pso::Rasterizer::FILL,
+                    Some(fs_entry),
                     &pipeline_layout,
                     subpass,
                 );
                 pipeline_desc.blender.targets.push(pso::ColorBlendDesc {
                     mask: pso::ColorMask::ALL,
                     blend: Some(pso::BlendState::ALPHA),
-                });
-                pipeline_desc.vertex_buffers.push(pso::VertexBufferDesc {
-                    binding: 0,
-                    stride: size_of::<Vertex>() as u32,
-                    rate: pso::VertexInputRate::Vertex,
-                });
-
-                pipeline_desc.attributes.push(pso::AttributeDesc {
-                    location: 0,
-                    binding: 0,
-                    element: pso::Element {
-                        format: f::Format::Rg32Sfloat,
-                        offset: 0,
-                    },
-                });
-                pipeline_desc.attributes.push(pso::AttributeDesc {
-                    location: 1,
-                    binding: 0,
-                    element: pso::Element {
-                        format: f::Format::Rg32Sfloat,
-                        offset: 8,
-                    },
                 });
 
                 device.create_graphics_pipeline(&pipeline_desc, None)
--- a/src/backend/dx11/src/conv.rs
+++ b/src/backend/dx11/src/conv.rs
@@ -785,6 +785,7 @@ pub fn map_stage(stage: Stage) -> spirv::ExecutionModel {
         Stage::Compute => spirv::ExecutionModel::GlCompute,
         Stage::Hull => spirv::ExecutionModel::TessellationControl,
         Stage::Domain => spirv::ExecutionModel::TessellationEvaluation,
+        Stage::Task | Stage::Mesh => panic!("{:?} shader is not supported in DirectX 11", stage),
     }
 }
 
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -919,43 +919,52 @@ impl device::Device<Backend> for Device {
                 .map_err(|err| pso::CreationError::Shader(err))
         };
 
-        let vs = build_shader(pso::Stage::Vertex, Some(&desc.shaders.vertex))?.unwrap();
-        let ps = build_shader(pso::Stage::Fragment, desc.shaders.fragment.as_ref())?;
-        let gs = build_shader(pso::Stage::Geometry, desc.shaders.geometry.as_ref())?;
-        let ds = build_shader(pso::Stage::Domain, desc.shaders.domain.as_ref())?;
-        let hs = build_shader(pso::Stage::Hull, desc.shaders.hull.as_ref())?;
+        let (layout, vs, gs, hs, ds) = match desc.primitive_assembler {
+            pso::PrimitiveAssembler::Vertex {
+                ref buffers,
+                ref attributes,
+                ref input_assembler,
+                ref vertex,
+                ref tessellation,
+                ref geometry,
+            } => {               
+                let vs = build_shader(pso::Stage::Vertex, Some(&vertex))?.unwrap();
+                let gs = build_shader(pso::Stage::Geometry, geometry.as_ref())?;  
 
-        let layout = self.create_input_layout(
-            vs.clone(),
-            &desc.vertex_buffers,
-            &desc.attributes,
-            &desc.input_assembler,
-        )?;
-        let rasterizer_state = self.create_rasterizer_state(&desc.rasterizer)?;
-        let blend_state = self.create_blend_state(&desc.blender)?;
-        let depth_stencil_state = Some(self.create_depth_stencil_state(&desc.depth_stencil)?);
+                let layout = self.create_input_layout(vs.clone(), buffers, attributes, input_assembler)?;
 
-        let vs = self.create_vertex_shader(vs)?;
+                let vs = self.create_vertex_shader(vs)?;
+                let gs = if let Some(blob) = gs {
+                    Some(self.create_geometry_shader(blob)?)
+                } else {
+                    None
+                };
+                let (hs, ds) = if let Some(ts) = tessellation {
+                    let hs = build_shader(pso::Stage::Hull, Some(&ts.0))?.unwrap();
+                    let ds = build_shader(pso::Stage::Domain, Some(&ts.1))?.unwrap();
+
+                    (Some(self.create_hull_shader(hs)?), Some(self.create_domain_shader(ds)?))
+                } else {
+                    (None, None)
+                };
+
+                (layout, vs, gs, hs, ds)
+            },
+            pso::PrimitiveAssembler::Mesh { .. } => {
+                return Err(pso::CreationError::UnsupportedPipeline)
+            }
+        };
+
+        let ps = build_shader(pso::Stage::Fragment, desc.fragment.as_ref())?;
         let ps = if let Some(blob) = ps {
             Some(self.create_pixel_shader(blob)?)
         } else {
             None
-        };
-        let gs = if let Some(blob) = gs {
-            Some(self.create_geometry_shader(blob)?)
-        } else {
-            None
-        };
-        let ds = if let Some(blob) = ds {
-            Some(self.create_domain_shader(blob)?)
-        } else {
-            None
-        };
-        let hs = if let Some(blob) = hs {
-            Some(self.create_hull_shader(blob)?)
-        } else {
-            None
-        };
+        };  
+
+        let rasterizer_state = self.create_rasterizer_state(&desc.rasterizer)?;
+        let blend_state = self.create_blend_state(&desc.blender)?;
+        let depth_stencil_state = Some(self.create_depth_stencil_state(&desc.depth_stencil)?);              
 
         Ok(GraphicsPipeline {
             vs,
--- a/src/backend/dx12/src/conv.rs
+++ b/src/backend/dx12/src/conv.rs
@@ -641,5 +641,6 @@ pub fn map_stage(stage: pso::Stage) -> spirv::ExecutionModel {
         pso::Stage::Compute => spirv::ExecutionModel::GlCompute,
         pso::Stage::Hull => spirv::ExecutionModel::TessellationControl,
         pso::Stage::Domain => spirv::ExecutionModel::TessellationEvaluation,
+        pso::Stage::Task | pso::Stage::Mesh => panic!("{:?} shader is not yet implemented in SPIRV-Cross", stage),
     }
 }
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -328,13 +328,12 @@ impl Device {
                 )
                 .map_err(gen_unexpected_error)?;
             }
-            if !layout.elements[set as usize].mutable_bindings.contains(&binding) {
-                ast.set_decoration(
-                    storage_buffer.id,
-                    spirv::Decoration::NonWritable,
-                    0,
-                )
-                .map_err(gen_unexpected_error)?
+            if !layout.elements[set as usize]
+                .mutable_bindings
+                .contains(&binding)
+            {
+                ast.set_decoration(storage_buffer.id, spirv::Decoration::NonWritable, 0)
+                    .map_err(gen_unexpected_error)?
             }
         }
 
@@ -353,13 +352,12 @@ impl Device {
                 )
                 .map_err(gen_unexpected_error)?;
             }
-            if !layout.elements[set as usize].mutable_bindings.contains(&binding) {
-                ast.set_decoration(
-                    image.id,
-                    spirv::Decoration::NonWritable,
-                    0,
-                )
-                .map_err(gen_unexpected_error)?
+            if !layout.elements[set as usize]
+                .mutable_bindings
+                .contains(&binding)
+            {
+                ast.set_decoration(image.id, spirv::Decoration::NonWritable, 0)
+                    .map_err(gen_unexpected_error)?
             }
         }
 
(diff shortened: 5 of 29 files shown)


kvark:
Thank you, that's amazing to see! I'll do a proper review shortly.
Just wanted to comment on Metal and platforms without the mesh shaders.

IIRC from WebGPU discussions, the whole idea of the mesh shaders was kicked off by some research work of doing them on compute, long before it became available in hardware/drivers. Therefore, compute path should be possible, even if not as efficient. I'd like us to have that path available at some point. We have `Hints` specifically to tell the users whether something is natively supported versus emulated.

kvark:
> there be MeshPipelineDesc as counterpart to GraphicsPipelineDesc?

Only if Vulkan or D3D12 define a separate pipeline kind for this. As far as I see from a brief look, it's all the same graphics pipeline, just a bunch of stages are replaced. So let's keep it `GraphicsPipelineDesc`

> GraphicsPipelineDesc silently ignores input_assembler and contents of vertex_buffers and attributes

Can we have an enum?
```rust
enum PrimitiveAssemer {
  Vertex(InputAssemer, VertexBuffers),
  Mesh(...),
}
```

MatusT:
> 
> Can we have an enum?
> 
> ```rust
> enum PrimitiveAssemer {
>   Vertex(InputAssemer, VertexBuffers),
>   Mesh(...),
> }
> ```

I will try to summarise here what options I see. 

Firstly, just the struct itself:
1. Follow what Vulkan/DX12 does and silently ignore unnecessary attributes.
  - Pros:
    - Backwards compatible API, since members are public. I see this as a very important attribute.
  - Cons:
    - `Vec<VertexBufferDesc> `and  `Vec<AttributeDesc>` must be set empty. Not that annoying.
    - `InputAssemblerDesc` must be filled with some values, which does not really make sense and seems annoying. **However**, I would personally circumvent this by implementing Default for it + Default for GraphicsPipelineDesc itself so that a user could create the struct with only necessary attributes and just `.. Default::default()` the rest.
2. Make an enum
  - Pros:
    - You now have only one member and it kind of makes sense.
  - Cons:
    - Breaking change. 
    - Enum must contain 3 whole members inside one variant and absolutely none in the second:
    ```
    enum PrimitiveAssembler {
      Vertex(Vec<VertexBufferDesc>, Vec<AttributeDesc>, InputAssemblerDesc),
      Mesh, // literally nothing
    }
    ```
    You can of course have any variation: make those 3 members optional or make only `InputAssemblerDesc` optional

Secondly, there is a question of `GraphicsPipelineDesc::new()` function. Similar questions with respect to API apply as above:
- Make `primitive` and `rasterizer` optional? Remove them altogether from the new function?
- Make a new `new_mesh` function? I am personally in favour of this option.

Since I have started writing about breakage of the API, I have realised that the API of `GraphicsShaderSet` will be broken in any case because vertex becomes optional and task+mesh must be filled. I don't know how many projects rely on the stability of gfx, so I am unable to asses the potential disruption in the ecosystem.

As you are the maintainer, you should choose the course of action and I will modify the code accordingly.

kvark:
> the API of GraphicsShaderSet will be broken in any case because vertex becomes optional

Right, I don't think we should try too hard to avoid API breakage here. Consider this to be experimental, we'll adjust as we go, and `master` is where experimentation happens.

> Mesh, // literally nothing

I think we can do better than that. We can exclude the task and mesh stages from the graphics shader set and put them here:
```rust
enum PrimitiveAssember<'a, B> {
  Vertex {
  },
  Mesh {
    task_shader: EntryPoint<'a, B>,
    mesh_shader: EntryPoint<'a, B>,
  },
}
```

msiglreith:
+1 for the enum variant - totally replacing the GraphicsShaderSet might be an option here and move all old pipeline stages into the Vertex variant. Fragment shader entrypoint could live then as field in the GraphicsPipelineDesc or be part of some other sub desc.

MatusT:
I updated the API with the Enum idea. I must admit that the solution is slowly growing on me.

I also added 2 new errors to the pipeline CreationError, one for when the mesh shader is not supported and the other one for when the vertex stage is missing in vertex primitive assembly.

examples/colour-uniform/main.rs
```diff
@@ -1256,7 +1256,7 @@ impl<B: Backend> PipelineState<B> {
                 );
 
                 let shader_entries = pso::GraphicsShaderSet {
-                    vertex: vs_entry,
+                    vertex: Some(vs_entry),
```
kvark, line 1259: I think we should change the way we use this structure. The fragment stage could be moved outside, straight into the `GraphicsPipelineDesc`. And the structure with all the other stages here could be a part of the enum variant for non-mesh shaders. What do you think?

examples/colour-uniform/main.rs
```diff
@@ -1302,6 +1294,26 @@ impl<B: Backend> PipelineState<B> {
                     },
                 });
 
+                let mut pipeline_desc = pso::GraphicsPipelineDesc::new(
+                    shader_entries,
+                    pso::PrimitiveAssembler::Vertex {
+                        buffers: vertex_buffers,
```
kvark, line 1300: would be easier to just do `vec![..]` in here and provide everything in place

examples/colour-uniform/main.rs
```diff
@@ -1302,6 +1294,26 @@ impl<B: Backend> PipelineState<B> {
...
+                        attributes,
```
kvark, line 1301: same for attributes

src/backend/dx11/src/conv.rs
```diff
@@ -785,6 +785,8 @@ pub fn map_stage(stage: Stage) -> spirv::ExecutionModel {
         Stage::Compute => spirv::ExecutionModel::GlCompute,
         Stage::Hull => spirv::ExecutionModel::TessellationControl,
         Stage::Domain => spirv::ExecutionModel::TessellationEvaluation,
+        Stage::Task => panic!("Task shader not yet implemented in the SPIR-V Cross."),
+        Stage::Mesh => panic!("Mesh shader not yet implemented in the SPIR-V Cross."),
```
kvark, line 789: nit: could do
```
Stage::Task | Stage::Mesh => panic!("{:?} shader is not yet implemented in SPIRV-Cross", stage),
```

src/backend/dx11/src/device.rs
```diff
@@ -919,18 +919,26 @@ impl device::Device<Backend> for Device {
...
         let ds = build_shader(pso::Stage::Domain, desc.shaders.domain.as_ref())?;
         let hs = build_shader(pso::Stage::Hull, desc.shaders.hull.as_ref())?;
 
-        let layout = self.create_input_layout(
-            vs.clone(),
-            &desc.vertex_buffers,
-            &desc.attributes,
-            &desc.input_assembler,
-        )?;
+        let (vertex_buffers, attributes, input_assembler) = match &desc.primitive_assembler {
```
kvark, line 929: could make this easier if you combine it with the next statement, i.e.
```
let layout = match desc.primitive_assembler {...};
```

src/backend/dx11/src/device.rs
```diff
@@ -919,18 +919,26 @@ impl device::Device<Backend> for Device {
...
+            &pso::PrimitiveAssembler::Vertex {
```
kvark, line 930: please don't match on the reference, match on the thing itself

src/backend/dx12/src/conv.rs
```diff
@@ -641,5 +641,7 @@ pub fn map_stage(stage: pso::Stage) -> spirv::ExecutionModel {
         pso::Stage::Compute => spirv::ExecutionModel::GlCompute,
         pso::Stage::Hull => spirv::ExecutionModel::TessellationControl,
         pso::Stage::Domain => spirv::ExecutionModel::TessellationEvaluation,
+        pso::Stage::Task => panic!("Task shader not yet implemented in the SPIR-V Cross."),
```
kvark, line 644: same as with the other backend, could be slightly simpler

src/backend/dx12/src/device.rs
```diff
@@ -1792,250 +1790,265 @@ impl d::Device<B> for Device {
...
+                        vertex_bindings,
+                        baked_states,
+                    })
+                } else {
+                    Err(pso::CreationError::Other)
+                }
+            }
+            pso::PrimitiveAssembler::Mesh { .. } => {
+                // TODO: Implement mesh shaders for DX12
+                Err(pso::CreationError::UnsupportedPipeline)
```
kvark, line 2050: it's somewhat unfortunate to indent the whole multi-page block above because of this. Maybe we can have it as early return for now?

src/backend/gl/src/command.rs
```diff
@@ -1484,6 +1484,31 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
         unimplemented!()
     }
 
+    unsafe fn draw_mesh_tasks(&mut self, _: u32, _: u32) {
```
kvark, line 1487: we have some defines for these, could we use them here, or add a new one if needed?
 I.e. look at how other draws use `hal::DrawCount`, `hal::IndexCount`, etc

src/backend/gl/src/device.rs
```diff
@@ -758,12 +758,30 @@ impl d::Device<B> for Device {
             }
         };
 
+        let (desc_vertex_buffers, desc_attributes, desc_input_assembler) =
+            match &desc.primitive_assembler {
```
kvark, line 762: please don't match on a reference

src/backend/gl/src/device.rs
```diff
@@ -846,13 +864,13 @@ impl d::Device<B> for Device {
...
 
-        let patch_size = match desc.input_assembler.primitive {
+        let patch_size = match desc_input_assembler.primitive {
             pso::Primitive::PatchList(size) => Some(size as _),
             _ => None,
         };
 
         let mut vertex_buffers = Vec::new();
-        for vb in &desc.vertex_buffers {
+        for vb in desc_vertex_buffers {
```
kvark, line 873: nit: we should probably move that into the match arm, could be done as a follow-up

src/backend/gl/src/queue.rs
```diff
@@ -200,7 +209,7 @@ impl CommandQueue {
     /// Return a reference to a stored data object.
     fn get_raw(data: &[u8], ptr: com::BufferSlice) -> &[u8] {
         assert!(data.len() >= (ptr.offset + ptr.size) as usize);
-        &data[ptr.offset as usize..(ptr.offset + ptr.size) as usize]
+        &data[ptr.offset as usize .. (ptr.offset + ptr.size) as usize]
```
kvark, line 212: Are you using rustfmt nightly? We used that previously, but now we want to switch fully to the stable. That requires cleaning up the rustfmt.toml a bit. The point is - this is an unrelated task, It would be good if your PR didn't attempt to do it partially. Reviewing only non-formatting changes is easier.

src/backend/vulkan/src/device.rs
```diff
@@ -102,70 +102,100 @@ impl GraphicsPipelineInfoBuf {
...
-        }
-        // Domain stage
-        if let Some(ref entry) = desc.shaders.domain {
-            this.add_stage(vk::ShaderStageFlags::TESSELLATION_EVALUATION, entry);
-        }
-        // Hull stage
-        if let Some(ref entry) = desc.shaders.hull {
-            this.add_stage(vk::ShaderStageFlags::TESSELLATION_CONTROL, entry);
-        }
+        match &desc.primitive_assembler {
```
kvark, line 105: please no match of references :)

examples/colour-uniform/main.rs
MatusT: I agree. Although I would just let shaders be freestanding inside `PrimitiveAssembler` and remove `GraphicsShaderSet`. So the final look would be:
```
pub enum PrimitiveAssembler<'a, B: Backend> {
    /// Vertex based pipeline
    Vertex {
        buffers: Vec<VertexBufferDesc>,
        attributes: Vec<AttributeDesc>,
        input_assembler: InputAssemblerDesc,
        vertex: EntryPoint<'a, B>,
        hull: Option<EntryPoint<'a, B>>,
        domain: Option<EntryPoint<'a, B>>,
        geometry: Option<EntryPoint<'a, B>>,
    },
    /// Mesh shading pipeline
    Mesh {
        task_shader: Option<EntryPoint<'a, B>>,
        mesh_shader: EntryPoint<'a, B>,
    },
}
```
Or it would be too much nesting IMHO.

src/backend/gl/src/queue.rs
MatusT: I normally use stable, but the errors I received hinted that I should use nightly for the gfx. 
This slipped accidentally. Should I just format manually until rustfmt.toml is resolved?

examples/colour-uniform/main.rs
kvark: yes, that would be fine by me as well. Perhaps, the hull and domain should be merged into a tuple:
```rust
tessellation: Option<(EntryPoint, EntryPoint)>,`
```

src/backend/gl/src/queue.rs
kvark: yes, please, just don't bother formatting too hard

examples/colour-uniform/main.rs
MatusT: That's a good idea. I think this completes the API bikeshedding! :)

src/backend/gl/src/command.rs
MatusT: I created TaskCount and TaskOffset.

src/backend/dx12/src/device.rs
MatusT: So the new version is the best I could do without moving large indented blocks inside match and also satisfy Rust borrow checker. 

There are probably a lot of variants It could be rewritten. I might need some further advice on this if this is not satisfactory.

src/backend/dx11/src/device.rs
```diff
@@ -919,43 +919,62 @@ impl device::Device<Backend> for Device {
...
+            pso::PrimitiveAssembler::Vertex {
+                ref buffers,
+                ref attributes,
+                ref input_assembler,
+                ref vertex,
+                ref tessellation,
+                ref geometry,
+            } => {
+                let (hs, ds) = if let Some(ts) = tessellation {
+                    (Some(&ts.0), Some(&ts.1))
```
kvark, line 932: looks like we could have combined this match with the later one

src/hal/src/lib.rs
```diff
@@ -90,6 +90,10 @@ pub type InstanceCount = u32;
 pub type DrawCount = u32;
 /// Number of work groups.
 pub type WorkGroupCount = [u32; 3];
+///
+pub type TaskCount = u32;
+///
+pub type TaskOffset = u32;
```
kvark, line 96: We have a separate `VertexOffset` because it has a different type, while the index and instance are just using "count" for both the count and the offset. Would be good to make this somewhat consistent. Perhaps, we could use `TaskCount` for both?

src/hal/src/pso/graphics.rs
```diff
@@ -101,22 +69,63 @@ pub struct BakedStates {
     /// Static depth bounds.
     pub depth_bounds: Option<Range<f32>>,
 }
-
+#[derive(Debug)]
+/// Primitive Assembler describes how input data are fetched in the pipeline and formed into primitives before being sent into the fragment shader.
+pub enum PrimitiveAssembler<'a, B: Backend> {
```
kvark, line 74: This structure looks beautiful now, thank you!

src/hal/src/pso/mod.rs
```diff
@@ -100,6 +104,10 @@ bitflags!(
         /// Read/Write access from host.
         /// (Not a real pipeline stage)
         const HOST = 0x4000;
+        /// Task shader stage.
+        const TASK = 0x80000;
```
kvark, line 108: should be `TASK_SHADER` and `MESH_SHADER` for consistency with other flags

src/backend/dx11/src/conv.rs
```diff
@@ -785,6 +785,7 @@ pub fn map_stage(stage: Stage) -> spirv::ExecutionModel {
         Stage::Compute => spirv::ExecutionModel::GlCompute,
         Stage::Hull => spirv::ExecutionModel::TessellationControl,
         Stage::Domain => spirv::ExecutionModel::TessellationEvaluation,
+        Stage::Task | Stage::Mesh => panic!("{:?} shader is not supported in DirectX 11", stage),
```
kvark, line 788: I just realized we don't need `Stage` at all now. Filed https://github.com/gfx-rs/gfx/issues/3241 to pursue separately...

MatusT: Should I just remove It for now and wait until #3241 is resolved?

src/backend/dx11/src/device.rs
MatusT: Done

src/hal/src/lib.rs
MatusT: Done.

src/hal/src/pso/graphics.rs
MatusT: Thanks! :)

src/hal/src/pso/mod.rs
MatusT: Done.

MatusT:
We have one final big problem to resolve that I just realized.

There is one major difference between DX12 and Vulkan.
Vulkan only allows 1D Tasks https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdDrawMeshTasksNV.html.
While DX12 allows specifying all 3 dimensions https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html#dispatchmesh-api. In addition, It does not have offsets(`first_task`).

I modelled the API according to Vulkan, which is currently the lowest common denominator.

Alternative could be:
- allowing all 3 dimensions and reporting maximum sizes of Y, Z as 1 on the Vulkan backend
- maybe patch the shader to transform 1D to 3D?
 
Maybe this will change with KHR version of the extension and we should wait until then. Unless, NVIDIA/Khronos/AMD could give use some more formation.

src/backend/dx11/src/conv.rs
kvark: only if you want to help resolving #3241. Otherwise, we'll remove this code later.

kvark:
Let's model this after Vulkan for now, and then reconsider when KHR extension comes out.
Are you in Khronos by any chance? We could use some early insights on what the API will be, if it happens at all :)

src/backend/dx11/src/conv.rs
MatusT: I am fraid I don't have that much time currently :(. Have to work on my diploma thesis. So If that is alright, I will leave It for now.

MatusT:
> Let's model this after Vulkan for now, and then reconsider when KHR extension comes out.

I will leave It as is It is then.

> Are you in Khronos by any chance? We could use some early insights on what the API will be, if it happens at all :)

I thought you were part of Khronos on behalf of Mozilla :D. What I would give for being part of Khronos. No, I am merely a master's student who is using WebGPU for my diploma thesis and I wanted to use mesh shaders for that.

kvark:
I am a part of Khronos, just having more than I can chew on my plate. I can dive and check the archives though to see where things are going, at least.
CI mac is experiencing strange failures - https://travis-ci.org/github/gfx-rs/gfx/jobs/682350838#L3665

kvark:
bors try

kvark:
Could you rebase and squash please?

MatusT:
> Could you rebase and squash please?

Done.

kvark:
bors try

kvark:
I'm confused as to why some PRs are landing fine, and others aren't :/
Anyhow, this nightly error appears to be fixed upstream, just going to wait for the next nightly.

MatusT:
@kvark CI finally succeeded!

PR #3239, opened 2020-05-01
DX12 waitable swap chain

kvark:
Fixes #3193
Fixes #3088
Tested on wgpu-rs examples.

PR #3240, opened 2020-05-01
[0.5/dx12] waitable swapchain

kvark:
Sibling of #3239


kvark:
I'm going to go ahead and merge this. If any follow-ups are taken from #3239, we'll back port them here.
bors r+

PR #3244, opened 2020-05-03
d3d12: Fix offset calculation for root descriptors that was causing root signature data to be overwritten

### LINKED ISSUE #3243: `SetGraphicsRootDescriptorTable` calls do not happen when using dynamic offsets with a buffer descriptor
- GFX version: 0.5.0, dx backend 0.5.0
- OS: Windows 10
- GPU: NVIDIA Quadro M1200

I have a `DescriptorType::Buffer` of `BufferDescriptorType::Uniform`, with the format `BufferDescriptorFormat::Structured`. When I set `dynamic_offsets` to true, and supply a dynamic offset, nothing is rendered. I checked it out in renderdoc:

Without dynamic offsets:
debug-1

With dynamic offsets:
debug-2

It looks like the `SetGraphicsRootDescriptorTable` never gets called when using dynamic offsets. Looked in the code and found the loop where this (doesn't) happen: https://github.com/gfx-rs/gfx/blob/015fa95b171e1a708e467c47fe843ea1c5f6cc3b/src/backend/dx12/src/command.rs#L821

Also tested on Metal and Vulkan, both seem to work and render the image properly.

I'm willing to work on a fix if somebody could point me in the right direction - I don't know a lot about the d3d12 API but it seems like `table_update` should be called in addition to `descriptor_cbv_update`.

Dooskington:
Fixes #3243

Root descriptors were always given offset 0, regardless of any other descriptors that were present. This caused whatever user data that was at offset 0 and 1 to get overwritten (in my case, it was a SRV descriptor with an image, then a sampler descriptor, resulting in no image being rendered). 

`bind_descriptor_sets` expects a certain ordering for the descriptors within each set: 
SRV/CBV/UAV descriptors, in one table
Sampler descriptors, in one table
Root descriptors

`create_pipeline_layout` didn't follow this contract (looks like it was meant to, though), resulting in the offset mixup. I just re-arranged it a bit to ensure that tables are added first, so that the `root_offset` was correct by the end of the method and the `RootDescriptor` could be created with it.

I also touched up some documentation, hope that is ok. I am new to directx, did a lot of digging in the [documentation ](https://docs.microsoft.com/en-us/windows/win32/direct3d12/example-root-signatures) but it's possible I screwed something up big time here. Let me know.

PR checklist:
- [x] tested examples with the following backends: `quad`, `compute` on directX. Although these don't use dynamic descriptor sets, so I have just been using my own game project as a test bed.
- [ ] `rustfmt` run on changed code (I did this and it changed a ton of already existing formatting, is this still desired?)

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -1518,6 +1518,8 @@ impl d::Device<B> for Device {
         // Each descriptor set layout will be one table entry of the root signature.
         // We have the additional restriction that SRV/CBV/UAV and samplers need to be
         // separated, so each set layout will actually occupy up to 2 entries!
+        // SRV/CBV/UAV tables are added to the signature first, then Sampler tables,
+        // and finally dynamic uniform descriptors.
         //
         // Dynamic uniform buffers are implemented as root descriptors.
         // This allows to handle the dynamic offsets properly, which would not be feasible
@@ -1526,10 +1528,11 @@ impl d::Device<B> for Device {
         // Root signature layout:
         //     Root Constants: Register: Offest/4, Space: 0
         //     ...
-        //     DescriptorTable0: Space: 1 (+1) (SrvCbvUav)
-        //     Root Descriptors
-        //     DescriptorTable0: Space: 2 (+1) (Sampler)
-        //     DescriptorTable1: Space: 3 (+1) (SrvCbvUav)
+        // DescriptorTable0: Space: 1 (SrvCbvUav)
+        // DescriptorTable0: Space: 1 (Sampler)
+        // Root Descriptors 0
+        // DescriptorTable1: Space: 2 (SrvCbvUav)
+        // Root Descriptors 1
         //     ...
 
         let sets = sets.into_iter().collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


Dooskington:
`SetGraphicsRootDescriptorTable` and `SetGraphicsRootConstantBufferView` are now being called together correctly:

image

kvark:
@msiglreith would you have a chance to have a look here?

src/backend/dx12/src/device.rs
```diff
@@ -1528,8 +1530,9 @@ impl d::Device<B> for Device {
         //     ...
         //     DescriptorTable0: Space: 1 (+1) (SrvCbvUav)
         //     Root Descriptors
-        //     DescriptorTable0: Space: 2 (+1) (Sampler)
-        //     DescriptorTable1: Space: 3 (+1) (SrvCbvUav)
+        //     DescriptorTable1: Space: 2 (+1) (SrvCbvUav)
+        //     DescriptorTable0: Space: 3 (+1) (Sampler)
+        //     Root Descriptors
```
msiglreith, line 1535: should be slightly different given the current layout (incl. changes to sampler space assignment which hasn't be reflected so far)
```
DescriptorTable0: Space: 1 (SrvCbvUav)
DescriptorTable0: Space: 1 (Sampler)
Root Descriptors 0
DescriptorTable1: Space: 2 (SrvCbvUav)
RootDescriptors 1
```

Dooskington: Thanks! I'll update that comment.

Dooskington:
Anything else to be done to get this merged? 
I tried to run rustfmt but it touches a lot of code, as if the lib has never been formatted. So is this still a requirement?

I can attempt the `make` stuff today.

kvark:
Could you squash it please? Our bors can't squash.

Dooskington:
@kvark sure, done.

kvark:
Do you need that fix back-ported to 0.5? We'd need the same PR but for hal-0.5 branch, that also has a commit of updating CHANGELOG and bumping the patch version in the backend.

Dooskington:
I don't need it backported right this second, but if you think it will be a while for it to otherwise get shipped out, maybe it's a good idea. What's the process like - bump the dx backend to 0.5.2 and then update CHANGELOG with the following?:

### backend-dx12-0.5.2 (05-04-2020)
  - fix offset calculation for root descriptors

kvark:
Yes, that changelog entry looks great. You'd basically checkout the branch, branch-off, cherry-pick that commit you made in master, then make an extra commit with the version bump and the changelog entry, file a PR. And thank you! I think it would be worthwhile to do.

PR #3251, opened 2020-05-07
[0.5] Update spirv-cross to version 0.20

grovesNL:
Backport #3249 to 0.5

PR checklist:


src/auxil/auxil/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-auxil"
-version = "0.3.0"
+version = "0.3.1"
```
kvark, line 3: I think spirv_cross is a public dependency of this crate, so it has to jump to 0.4.0

grovesNL: Good catch, thanks!

PR #3252, opened 2020-05-07
[0.5/mtl] fix issue with fill_buffer when using offset and size

kvark:
Port of #3248


kvark:
@grovesNL let's get this also before releasing

PR #3253, opened 2020-05-09
Mesh shaders: implement for Vulkan backend

MatusT:
This also includes an example.

To discuss\resolve:
- [x] `mesh_fn` added to `RawDevice`. Is it ok? Will It stay ok being extended with many more extensions?
- [ ] Limits: Unfortunately, Khronos made a slight in mistake in 1.0 by not assigning `pNext` member to some structs that should have had one. This includes VkPhysicalProperties. Without Vulkan 1.1(or KHR extension) where this was remedied, we are unable to query for limits of the mesh shader extension that are inside [VkPhysicalDeviceMeshShaderPropertiesNV](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VkPhysicalDeviceMeshShaderPropertiesNV) struct.

PR checklist:


examples/mesh-shading/data/index.html
```diff
@@ -0,0 +1,17 @@
+<html>
+  <head>
+    <meta content="text/html;charset=utf-8" http-equiv="Content-Type"/>
```
kvark, line 3: why do you need an HTML here? is this targeting JS?

src/backend/vulkan/src/command.rs
```diff
@@ -836,30 +836,42 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
             .cmd_draw_indexed_indirect(self.raw, buffer.raw, offset, draw_count, stride)
     }
 
-    unsafe fn draw_mesh_tasks(&mut self, _: TaskCount, _: TaskCount) {
-        unimplemented!()
+    unsafe fn draw_mesh_tasks(&mut self, task_count: TaskCount, first_task: TaskCount) {
+        if let Some(mesh_fn) = &self.device.mesh_fn {
```
kvark, line 840: let's just do `expect` with this message

src/backend/vulkan/src/lib.rs
```diff
@@ -1268,6 +1296,7 @@ pub struct RawDevice {
     raw: ash::Device,
     features: Features,
     instance: Arc<RawInstance>,
+    mesh_fn: Option<MeshShader>,
```
kvark, line 1299: this looks good

src/backend/vulkan/src/lib.rs
```diff
@@ -687,17 +692,29 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
 
+            let mesh_shader_features = vk::PhysicalDeviceMeshShaderFeaturesNV::builder()
+                .task_shader(requested_extensions.contains(Features::TASK_SHADER))
+                .mesh_shader(requested_extensions.contains(Features::MESH_SHADER))
+                .build();
+
             let info = vk::DeviceCreateInfo {
                 s_type: vk::StructureType::DEVICE_CREATE_INFO,
-                p_next: ptr::null(),
+                p_next: &mesh_shader_features as *const vk::PhysicalDeviceMeshShaderFeaturesNV as *const std::ffi::c_void,
```
kvark, line 717: just to clarify, will this confuse vulkan implementations that don't support or know about mesh shaders?

src/backend/vulkan/src/lib.rs
```diff
@@ -663,6 +665,9 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
             })
             .collect::<Vec<_>>();
 
+        let requested_extensions = requested_features & Features::EXTENSIONS_MASK;
```
kvark, line 668: why is this separation needed? what happens if you just use `requsted_features` for everything?

MatusT: I see now that It is not. This was only a confusion on my side of understanding gfx-hal. Sorry, I am still grappling this huge codebase :). I though It was going directly to a Vulkan structure and missed the conversion.

src/backend/vulkan/src/lib.rs
MatusT: Yeah, It might. I have reworked the code quite a lot now and I hope you like It. So now It uses the builder pattern from ash to be much cleaner and also assigns the p_next only if neccessary. I think this code structure will also make it much cleaner for any future extension.

src/backend/vulkan/src/lib.rs
MatusT: Good :)

src/backend/vulkan/src/command.rs
MatusT: Done.

examples/mesh-shading/data/index.html
MatusT: It is a leftover. As you correctly guessed the mesh example is based on the quad example. It is basically just a copy paste with removed image and few modifications here and there.

I would like to achieve support of mesh shaders in wgpu(-rs) so I left the html there as It could possibly run in the Firefox then. But that might be too much in a future so I am removing It.

src/backend/vulkan/src/command.rs
```diff
@@ -836,30 +836,36 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
             .cmd_draw_indexed_indirect(self.raw, buffer.raw, offset, draw_count, stride)
     }
 
-    unsafe fn draw_mesh_tasks(&mut self, _: TaskCount, _: TaskCount) {
-        unimplemented!()
+    unsafe fn draw_mesh_tasks(&mut self, task_count: TaskCount, first_task: TaskCount) {
+        self.device.mesh_fn.as_ref()
+            .expect("Draw command not supported. You must request feature MESH_SHADING_PIPELINE.")
```
kvark, line 841: our feature(s) is named differently, right? We should use the proper name here, like `TASK_SHADER`

src/backend/vulkan/src/command.rs
```diff
@@ -836,30 +836,36 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
...
-        _: hal::DrawCount,
-        _: u32,
+        buffer: &n::Buffer,
+        offset: buffer::Offset,
+        draw_count: hal::DrawCount,
+        stride: u32,
     ) {
-        unimplemented!()
+        self.device.mesh_fn.as_ref()
+            .expect("Draw command not supported. You must request feature MESH_SHADING_PIPELINE.")
```
kvark, line 853: same here

src/backend/vulkan/src/command.rs
```diff
@@ -836,30 +836,36 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
...
+        buffer: &n::Buffer,
+        offset: buffer::Offset,
+        count_buffer: &n::Buffer,
+        count_buffer_offset: buffer::Offset,
+        max_draw_count: DrawCount,
+        stride: u32,
     ) {
-        unimplemented!()
+        self.device.mesh_fn.as_ref()
+            .expect("Draw command not supported. You must request feature MESH_SHADING_PIPELINE.")
```
kvark, line 867: and here

src/backend/vulkan/src/lib.rs
```diff
@@ -687,26 +689,34 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
                 0 => None,
                 1 => Some(*KHR_MAINTENANCE1),
                 _ => unreachable!(),
-            });
+            })
+            .chain(
+                if requested_features.contains(Features::TASK_SHADER) | requested_features.contains(Features::MESH_SHADER) {
```
kvark, line 694: let's do `requested_features.intersects(Features::TASK_SHADER | Features::MESH_SHADER)`

src/backend/vulkan/src/lib.rs
```diff
@@ -1081,6 +1098,9 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
         if features.inherited_queries != 0 {
             bits |= Features::INHERITED_QUERIES;
         }
+        if self.supports_extension(*MESH_SHADER) {
+            bits |= Features::MESH_SHADER;
```
kvark, line 1102: what about `TASK_SHADER` bit?

src/backend/vulkan/src/lib.rs
```diff
@@ -687,26 +689,34 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+                .queue_create_infos(&family_infos)
+                .enabled_extension_names(&str_pointers)
+                .enabled_features(&enabled_features);
+
+            let mut mesh_shader_features = vk::PhysicalDeviceMeshShaderFeaturesNV::builder()
+                .task_shader(requested_features.contains(Features::TASK_SHADER))
+                .mesh_shader(requested_features.contains(Features::MESH_SHADER))
+                .build();
+
+            if requested_features.contains(Features::TASK_SHADER) | requested_features.contains(Features::MESH_SHADER) {
```
kvark, line 717: similarly, let's do `intersects(A | B)`

src/backend/vulkan/src/lib.rs
```diff
@@ -687,26 +689,34 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
-                flags: vk::DeviceCreateFlags::empty(),
-                queue_create_info_count: family_infos.len() as u32,
-                p_queue_create_infos: family_infos.as_ptr(),
-                enabled_layer_count: 0,
-                pp_enabled_layer_names: ptr::null(),
-                enabled_extension_count: str_pointers.len() as u32,
-                pp_enabled_extension_names: str_pointers.as_ptr(),
-                p_enabled_features: &enabled_features,
-            };
+            let mut info = vk::DeviceCreateInfo::builder()
```
kvark, line 707: this is cool!

MatusT: Thanks! :)

src/backend/vulkan/src/lib.rs
MatusT: Done.

src/backend/vulkan/src/lib.rs
MatusT: Added. Can't be separated without VkPhysicalFeatures2. Most probably inconsequential as DirectX 12 requires both shaders to be supported despite Task being optional in the pipeline as well as in the Vulkan extension itself.

src/backend/vulkan/src/lib.rs
MatusT: Done.

src/backend/vulkan/src/command.rs
MatusT: Done.

src/backend/vulkan/src/command.rs
MatusT: Done.

src/backend/vulkan/src/command.rs
MatusT: Done.

PR #3255, opened 2020-05-15
[webgpu] Instance creation and device enumeration

GabrielMajeri:
Add basic instance creation to the WebGPU backend. Also allows enumerating physical devices.

I've currently excluded the `WebGPU` backend from the `test` target, and added it manually to `check`.

PR checklist:


GabrielMajeri:
I've added a separate `check-wasm` Makefile target for those who want to test WASM (similar to `quad-wasm`).

This should not break make for systems without the `wasm32-unknown-unknown` target installed.

GabrielMajeri:
CI failed, but due to an unrelated problem:
```
error: failed to run custom build command for `glsl-to-spirv v0.1.7`
```

kvark:
Looks like we need to stop testing the examples (with `glsl-to-spirv`) on iOS

kvark:
@GabrielMajeri please rebase

PR #3266, opened 2020-06-01
Move pso::Stage to auxil::ShaderStage

### LINKED ISSUE #3241: Consider removing pso::Stage from the API
It looks like it's only used for one error variant and a bit in the backends. It's not really needed in the API itself.

kvark:
Fixes #3241
PR checklist:


kvark:
bors r=straightforward

PR #3269, opened 2020-06-09
Add Texture Descriptor Array Feature

cwfitzgerald:
This adds `Feature::TEXTURE_DESCRIPTOR_ARRAY` which is unconditionally present on DX12, DX11, and Vulkan, > MTL 2.0 on Metal, and unavailable on OpenGL. Strictly speaking that feature might be supported on opengl, but I'd want to make sure the backend supported it first.

A slightly concerning discovery is that Limits can't be extended without a breaking change, so descriptor indexing might have some issues. Thankfully this feature doesn't change the available limits.


cwfitzgerald:
> I'm surprised that this is supported on DX11, tbh. Do you know how this is resolved?

Spriv-cross happily spits out
```hlsl
Texture2D<float4> u_Textures[2] : register(t0);
SamplerState u_Sampler : register(s1);
```
for
```glsl
layout(set = 0, binding = 0) uniform texture2D u_Textures[2];
layout(set = 0, binding = 1) uniform sampler u_Sampler;
```
all the way down to shader model 4.0. Tbh it's not really that surprising, as it has to be compile time indexed, so it's no different for hardware than having two textures in a row.

kvark:
I don't think we should be exposing this feature in backends that don't yet support it. I.e. DX11 may be calling spirv_cross correctly, but the binding side will likely break before it gets to the shader. Are you testing this on anything?

cwfitzgerald:
I just finished my example, and it doesn't seem to work on DX11, so I have no idea if that's an issue with the DX11 backend, or what. I'll disable it on DX11 as I have no personal need for that backend.

PR #3271, opened 2020-06-12
[0.5] Implement Descriptor Indexing

cwfitzgerald:
This implements the gfx-hal component of supporting descriptor indexing.

Main actions:
 - Add features to hal
 - Enable appropriate features on metal
 - Enable features on DX12 with additional compiler flag if needed
 - Wire up enabling extensions to Vulkan.

Remaining Questions:
 - This PR isn't finished, but my main concern is how to get the struct of function pointers from `Instance::create`, through `PhysicalDevice` so it can get to `PhysicalDevice::features` where it needs to be. I sorta just stuck it in the structs, but if you have a better idea, please do suggest.
 - Any other vulkan advice/tips appreciated. 
 - I need to properly detect when optional instance extensions aren't supported.
 - Remove the horrible duplication of the extension name... somehow.

### DIFF
--- a/src/backend/dx12/src/device.rs
+++ b/src/backend/dx12/src/device.rs
@@ -117,6 +117,7 @@ pub(crate) enum CommandSignature {
 pub(crate) fn compile_shader(
     stage: pso::Stage,
     shader_model: hlsl::ShaderModel,
+    features: &hal::Features,
     entry: &str,
     code: &[u8],
 ) -> Result<native::Blob, d::ShaderError> {
@@ -137,6 +138,13 @@ pub(crate) fn compile_shader(
     let mut shader_data = native::Blob::null();
     let mut error = native::Blob::null();
     let entry = ffi::CString::new(entry).unwrap();
+    let mut compile_flags = d3dcompiler::D3DCOMPILE_ENABLE_STRICTNESS;
+    if cfg!(debug_assertions) {
+        compile_flags |= d3dcompiler::D3DCOMPILE_DEBUG;
+    }
+    if features.contains(hal::Features::UNSIZED_DESCRIPTOR_ARRAY) {
+        compile_flags |= d3dcompiler::D3DCOMPILE_ENABLE_UNBOUNDED_DESCRIPTOR_TABLES;
+    }
     let hr = unsafe {
         d3dcompiler::D3DCompile(
             code.as_ptr() as *const _,
@@ -146,7 +154,7 @@ pub(crate) fn compile_shader(
             ptr::null_mut(),
             entry.as_ptr() as *const _,
             full_stage.as_ptr() as *const i8,
-            1,
+            compile_flags,
             0,
             shader_data.mut_void() as *mut *mut _,
             error.mut_void() as *mut *mut _,
@@ -487,6 +495,7 @@ impl Device {
                         let shader = compile_shader(
                             stage,
                             shader_model,
+                            features,
                             &entry_point.name,
                             shader_code.as_bytes(),
                         )?;
@@ -505,7 +514,7 @@ impl Device {
         code: &[u8],
     ) -> Result<r::ShaderModule, d::ShaderError> {
         let mut shader_map = BTreeMap::new();
-        let blob = compile_shader(stage, hlsl::ShaderModel::V5_1, hlsl_entry, code)?;
+        let blob = compile_shader(stage, hlsl::ShaderModel::V5_1, &self.features, hlsl_entry, code)?;
         shader_map.insert(entry_point.into(), blob);
         Ok(r::ShaderModule::Compiled(shader_map))
     }
--- a/src/backend/vulkan/src/conv.rs
+++ b/src/backend/vulkan/src/conv.rs
@@ -389,84 +389,99 @@ pub fn map_buffer_features(features: vk::FormatFeatureFlags) -> format::BufferFe
     format::BufferFeature::from_bits_truncate(features.as_raw())
 }
 
-pub fn map_device_features(features: Features) -> vk::PhysicalDeviceFeatures {
-    // vk::PhysicalDeviceFeatures is a struct composed of Bool32's while
-    // Features is a bitfield so we need to map everything manually
-    vk::PhysicalDeviceFeatures::builder()
-        .robust_buffer_access(features.contains(Features::ROBUST_BUFFER_ACCESS))
-        .full_draw_index_uint32(features.contains(Features::FULL_DRAW_INDEX_U32))
-        .image_cube_array(features.contains(Features::IMAGE_CUBE_ARRAY))
-        .independent_blend(features.contains(Features::INDEPENDENT_BLENDING))
-        .geometry_shader(features.contains(Features::GEOMETRY_SHADER))
-        .tessellation_shader(features.contains(Features::TESSELLATION_SHADER))
-        .sample_rate_shading(features.contains(Features::SAMPLE_RATE_SHADING))
-        .dual_src_blend(features.contains(Features::DUAL_SRC_BLENDING))
-        .logic_op(features.contains(Features::LOGIC_OP))
-        .multi_draw_indirect(features.contains(Features::MULTI_DRAW_INDIRECT))
-        .draw_indirect_first_instance(features.contains(Features::DRAW_INDIRECT_FIRST_INSTANCE))
-        .depth_clamp(features.contains(Features::DEPTH_CLAMP))
-        .depth_bias_clamp(features.contains(Features::DEPTH_BIAS_CLAMP))
-        .fill_mode_non_solid(features.contains(Features::NON_FILL_POLYGON_MODE))
-        .depth_bounds(features.contains(Features::DEPTH_BOUNDS))
-        .wide_lines(features.contains(Features::LINE_WIDTH))
-        .large_points(features.contains(Features::POINT_SIZE))
-        .alpha_to_one(features.contains(Features::ALPHA_TO_ONE))
-        .multi_viewport(features.contains(Features::MULTI_VIEWPORTS))
-        .sampler_anisotropy(features.contains(Features::SAMPLER_ANISOTROPY))
-        .texture_compression_etc2(features.contains(Features::FORMAT_ETC2))
-        .texture_compression_astc_ldr(features.contains(Features::FORMAT_ASTC_LDR))
-        .texture_compression_bc(features.contains(Features::FORMAT_BC))
-        .occlusion_query_precise(features.contains(Features::PRECISE_OCCLUSION_QUERY))
-        .pipeline_statistics_query(features.contains(Features::PIPELINE_STATISTICS_QUERY))
-        .vertex_pipeline_stores_and_atomics(features.contains(Features::VERTEX_STORES_AND_ATOMICS))
-        .fragment_stores_and_atomics(features.contains(Features::FRAGMENT_STORES_AND_ATOMICS))
-        .shader_tessellation_and_geometry_point_size(
-            features.contains(Features::SHADER_TESSELLATION_AND_GEOMETRY_POINT_SIZE),
-        )
-        .shader_image_gather_extended(features.contains(Features::SHADER_IMAGE_GATHER_EXTENDED))
-        .shader_storage_image_extended_formats(
-            features.contains(Features::SHADER_STORAGE_IMAGE_EXTENDED_FORMATS),
-        )
-        .shader_storage_image_multisample(
-            features.contains(Features::SHADER_STORAGE_IMAGE_MULTISAMPLE),
-        )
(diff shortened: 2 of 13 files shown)


src/backend/dx12/src/device.rs
```diff
@@ -146,7 +147,11 @@ pub(crate) fn compile_shader(
             ptr::null_mut(),
             entry.as_ptr() as *const _,
             full_stage.as_ptr() as *const i8,
-            1,
+            d3dcompiler::D3DCOMPILE_DEBUG | if features.contains(hal::Features::UNSIZED_DESCRIPTOR_ARRAY) {
```
kvark, line 150: This changes the default value from 1 to 0.
Let's move this out and add a bit of logic here, something like:
```rust
let mut compile_flags = d3dcompiler::D3DCOMPILE_ENABLE_STRICTNESS;
if cfg!(debug_assertions) {
  compile_flags |= d3dcompiler::D3DCOMPILE_DEBUG;
}
if features.contains(hal::Features::UNSIZED_DESCRIPTOR_ARRAY) {
  compile_flags |= d3dcompiler::D3DCOMPILE_ENABLE_UNBOUNDED_DESCRIPTOR_TABLES;
}
```

src/backend/vulkan/src/conv.rs
```diff
@@ -466,7 +466,25 @@ pub fn map_device_features(features: Features) -> vk::PhysicalDeviceFeatures {
         .sparse_residency_aliased(features.contains(Features::SPARSE_RESIDENCY_ALIASED))
         .variable_multisample_rate(features.contains(Features::VARIABLE_MULTISAMPLE_RATE))
         .inherited_queries(features.contains(Features::INHERITED_QUERIES))
-        .build()
+        .build();
+
+    let extensions = crate::DeviceCreationExtensions {
+        descriptor_indexing: if features.intersects(
```
kvark, line 472: hah, you got this right! :)

src/backend/vulkan/src/lib.rs
```diff
@@ -424,6 +429,10 @@ impl hal::Instance<Backend> for Instance {
             })?
         };
 
+        let get_physical_device_properties = vk::KhrGetPhysicalDeviceProperties2Fn::load(|name| unsafe {
```
kvark, line 432: we need to only do this if `instance_extensions` contains your extension

src/backend/vulkan/src/lib.rs
```diff
@@ -510,6 +520,7 @@ impl hal::Instance<Backend> for Instance {
                     handle: device,
                     extensions,
                     properties,
+                    get_physical_device_properties: self.get_physical_device_properties.clone(),
```
kvark, line 523: since it's needed in multiple places, let's just put it into `RawInstance` struct, alongside the debug messenger and stuff

src/backend/vulkan/src/lib.rs
```diff
@@ -689,17 +705,15 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
-                p_next: ptr::null(),
-                flags: vk::DeviceCreateFlags::empty(),
-                queue_create_info_count: family_infos.len() as u32,
-                p_queue_create_infos: family_infos.as_ptr(),
-                enabled_layer_count: 0,
-                pp_enabled_layer_names: ptr::null(),
-                enabled_extension_count: str_pointers.len() as u32,
-                pp_enabled_extension_names: str_pointers.as_ptr(),
-                p_enabled_features: &enabled_features,
+            let info = vk::DeviceCreateInfo::builder()
```
kvark, line 708: oh wow, fancy builders!

src/backend/vulkan/src/lib.rs
```diff
@@ -61,9 +61,10 @@ lazy_static! {
         vec![
             DebugUtils::name(),
             DebugReport::name(),
+            CStr::from_bytes_with_nul(b"VK_KHR_get_physical_device_properties2\0").unwrap(),
         ]
     } else {
-        vec![]
+        vec![CStr::from_bytes_with_nul(b"VK_KHR_get_physical_device_properties2\0").unwrap()]
```
kvark, line 67: let's put this string as a constant alongside the other extension names (e.g.`AMD_NEGATIVE_VIEWPORT_HEIGHT`) and re-use in here

src/backend/vulkan/src/conv.rs
```diff
@@ -466,7 +466,25 @@ pub fn map_device_features(features: Features) -> vk::PhysicalDeviceFeatures {
         .sparse_residency_aliased(features.contains(Features::SPARSE_RESIDENCY_ALIASED))
         .variable_multisample_rate(features.contains(Features::VARIABLE_MULTISAMPLE_RATE))
         .inherited_queries(features.contains(Features::INHERITED_QUERIES))
-        .build()
+        .build();
+
+    let extensions = crate::DeviceCreationExtensions {
```
kvark, line 471: Would it make sense to have this structure containing the `core` as well? Returning a tuple from this function doesn't really give us anything :)

src/hal/src/lib.rs
```diff
@@ -201,7 +201,13 @@ bitflags! {
         const SAMPLER_MIP_LOD_BIAS = 0x0008 << 64;
 
         /// Make the NDC coordinate system pointing Y up, to match D3D and Metal.
-        const NDC_Y_UP = 0x01 << 80;
+        const NDC_Y_UP = 0x0001 << 80;
+        /// Allow indexing sampled texture descriptor arrays with dynamically non-uniform data
+        const SAMPLED_TEXTURE_DESCRIPTOR_INDEXING = 0x0002 << 80;
```
kvark, line 206: these should be in the `CORE_MASK`, actually, since they correspond directly to Vulkan extensions/features.
To clarify, the "core" is slightly misleading here. It was meant to be a mask of things that come from Vulkan.

src/backend/dx12/src/device.rs
cwfitzgerald: `D3DCOMPILE_DEBUG` is defined as `1 << 0` so that would seem to be the value of 1 it was set to before. I agree with the general refactor.

src/backend/vulkan/src/conv.rs
cwfitzgerald: For once... :)

src/backend/vulkan/src/lib.rs
cwfitzgerald: I tried that but was getting some inscrutable type error, need to try again when I'm not frustrated with vulkan's existance.

src/backend/vulkan/src/conv.rs
cwfitzgerald: Definitely

src/backend/vulkan/src/lib.rs
```diff
@@ -895,7 +910,23 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+                descriptor_indexing_features = Some(vk::PhysicalDeviceDescriptorIndexingFeaturesEXT::builder().build());
+                if let Some(ref mut mut_ref) = descriptor_indexing_features {
+                    mut_ref.p_next = mem::replace(&mut features2.p_next, mut_ref as *mut _ as *mut _);
+                }
+            }
+            unsafe { get_device_properties.get_physical_device_features2_khr(self.handle, &mut features2 as *mut _); }
+            features2.features
+        } else {
+            unsafe { self.instance.0.get_physical_device_features(self.handle) }
+        };
```
cwfitzgerald, line 927: This code definitely needs to be audited.

src/backend/vulkan/src/lib.rs
```diff
@@ -109,7 +114,7 @@ lazy_static! {
         );
 }
 
-pub struct RawInstance(ash::Instance, Option<DebugMessenger>);
+pub struct RawInstance(ash::Instance, Option<DebugMessenger>, Option<vk::KhrGetPhysicalDeviceProperties2Fn>);
```
kvark, line 117: at this point we should name the fields (can follow up, not strictly required her)

src/backend/vulkan/src/lib.rs
```diff
@@ -424,6 +429,13 @@ impl hal::Instance<Backend> for Instance {
             })?
         };
 
+        let get_device_physical_properties_found = extensions.iter().find(|&&ext| ext == *KHR_GET_PHYSICAL_DEVICE_PROPERTIES2);
+        let get_physical_device_properties = get_device_physical_properties_found.map(|_| {
```
kvark, line 433: nit: any reason not to write this down as a single expression?`get_device_physical_properties_found` isn't used anywhere else

src/backend/vulkan/src/lib.rs
```diff
@@ -895,7 +910,23 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
 
-        let features = unsafe { self.instance.0.get_physical_device_features(self.handle) };
+        let mut descriptor_indexing_features = None;
+        let features = if let Some(ref get_device_properties) = self.instance.2 {
+            let features = vk::PhysicalDeviceFeatures::builder().build();
+            let mut features2 = vk::PhysicalDeviceFeatures2KHR::builder().features(features).build();
+            if self.supports_extension(*EXT_DESCRIPTOR_INDEXING) {
+                descriptor_indexing_features = Some(vk::PhysicalDeviceDescriptorIndexingFeaturesEXT::builder().build());
+                if let Some(ref mut mut_ref) = descriptor_indexing_features {
+                    mut_ref.p_next = mem::replace(&mut features2.p_next, mut_ref as *mut _ as *mut _);
```
kvark, line 920: is `push_next` missing from Ash in this case? let's file a small issue for this

src/backend/vulkan/src/lib.rs
```diff
@@ -895,7 +910,23 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+                }
+            }
+            unsafe { get_device_properties.get_physical_device_features2_khr(self.handle, &mut features2 as *mut _); }
```
kvark, line 923: can we write it down in a comment how the sequence looks like?
I see `PhysicalDeviceFeatures2KHR` -> `PhysicalDeviceDescriptorIndexingFeaturesEXT` -> `PhysicalDeviceFeatures`. Is that correct?
The code could have been a little simpler if it didn't use `mem::replace`. SInce you know what the current tail is, you should be able to just do `p_next = ...` there.

src/backend/vulkan/src/lib.rs
```diff
@@ -895,7 +910,23 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
                     == info::intel::DEVICE_SKY_LAKE_MASK);
 
-        let features = unsafe { self.instance.0.get_physical_device_features(self.handle) };
+        let mut descriptor_indexing_features = None;
+        let features = if let Some(ref get_device_properties) = self.instance.2 {
+            let features = vk::PhysicalDeviceFeatures::builder().build();
+            let mut features2 = vk::PhysicalDeviceFeatures2KHR::builder().features(features).build();
+            if self.supports_extension(*EXT_DESCRIPTOR_INDEXING) {
+                descriptor_indexing_features = Some(vk::PhysicalDeviceDescriptorIndexingFeaturesEXT::builder().build());
+                if let Some(ref mut mut_ref) = descriptor_indexing_features {
```
kvark, line 919: since we know this to be `Some` at this point, we should not do `if`. Instead, let's do `descriptor_indexing_features.as_mut().unwrap()` or similar

src/backend/vulkan/src/lib.rs
kvark: agreed, see new comments :)

src/backend/vulkan/src/lib.rs
cwfitzgerald: Just a artifact of how I developed the code, will combine.

src/backend/vulkan/src/lib.rs
cwfitzgerald: Yeah, only some things get push_next for some reason.

src/backend/vulkan/src/lib.rs
cwfitzgerald: I thought that would be rejected by the borrow checker, but now I think about it, idk why...

src/backend/vulkan/src/lib.rs
cwfitzgerald: https://github.com/MaikKlein/ash/issues/229 someone beat us to the punch

cwfitzgerald:
I believe that is everything addressed

cwfitzgerald:
Everything should be bumped and all, I'll work on forward-porting this to master.

kvark:
bors retry

PR #3279, opened 2020-06-26
[0.5] Add draw_indirect_count feature

cwfitzgerald:
gfx-hal work for https://github.com/gfx-rs/wgpu/issues/742. I would not merge this until I've finished the wgpu work and prove that it works on both dx12 and vulkan, but it can be reviewed now :)

This adds two methods to a gfx-hal trait, so I added them with a default `unimplemented!()` implementation so it is a non-breaking change. When this is ported forward to master, the functions will just exist as is.

This adds support for this feature on DX12 and Vulkan (through `VK_KHR_draw_indirect_count`). Metal is possible as discussed in the wgpu issue, but I don't have access to a development mac, so I couldn't do the work on it, at least right now. GL 4.6 also seems to support this though `glMultiDrawIndexedIndirectCount`, but that's a problem to solve once the GL backend is sorted.

### DIFF
--- a/src/backend/vulkan/Cargo.toml
+++ b/src/backend/vulkan/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-backend-vulkan"
-version = "0.5.8"
+version = "0.5.9"
 description = "Vulkan API backend for gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
@@ -26,7 +26,7 @@ log = { version = "0.4" }
 lazy_static = "1"
 shared_library = { version = "0.1.9", optional = true }
 ash = "0.31"
-hal = { path = "../../hal", version = "0.5.2", package = "gfx-hal" }
+hal = { path = "../../hal", version = "0.5.3", package = "gfx-hal" }
 smallvec = "1.0"
 raw-window-handle = "0.3"
 
--- a/src/backend/vulkan/Cargo.toml
+++ b/src/backend/vulkan/Cargo.toml
@@ -25,7 +25,7 @@ byteorder = "1"
 log = { version = "0.4" }
 lazy_static = "1"
 shared_library = { version = "0.1.9", optional = true }
-ash = "0.30"
+ash = "0.31"
 hal = { path = "../../hal", version = "0.5.2", package = "gfx-hal" }
 smallvec = "1.0"
 raw-window-handle = "0.3"
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -12,6 +12,7 @@ extern crate objc;
 use ash::extensions::{
     self,
     ext::{DebugReport, DebugUtils},
+    khr::DrawIndirectCount,
 };
 use ash::version::{DeviceV1_0, EntryV1_0, InstanceV1_0};
 use ash::vk;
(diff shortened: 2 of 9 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -1294,11 +1318,16 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
     }
 }
 
+struct DeviceExtensionFunctions {
+    draw_indirect_count: Option<vk::KhrDrawIndirectCountFn>,
```
kvark, line 1322: why don't we use the slightly more convenient https://docs.rs/ash/0.31.0/ash/extensions/khr/struct.DrawIndirectCount.html ?

src/hal/src/command/mod.rs
```diff
@@ -480,6 +480,46 @@ pub trait CommandBuffer<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    unsafe fn draw_indirect_count(
+        &mut self,
+        _buffer: &B::Buffer,
+        _offset: buffer::Offset,
+        _count_buffer: &B::Buffer,
+        _count_buffer_offset: buffer::Offset,
+        _max_draw_count: u32,
+        _stride: u32
+    ) {
+        unimplemented!("Backend doesn't support draw_indirect_count");
```
kvark, line 500: when porting this to master, let's not forget to remove this default impl

src/backend/vulkan/src/lib.rs
cwfitzgerald: Damn, missed that, yes definitely. Always have trouble finding things in ash.

src/hal/src/command/mod.rs
cwfitzgerald: Agreed

src/backend/vulkan/src/lib.rs
kvark: I see that you pushed some code after the review, but this point still stands

cwfitzgerald:
Alright, should be upgraded. I also changed the feature name from (DRAW_INDIRECT_COUNT to MULTI_DRAW_INDIRECT_COUNT) to be more consistent with the other MULTI_DRAW features.

src/backend/vulkan/Cargo.toml
```diff
@@ -25,7 +25,7 @@ byteorder = "1"
 log = { version = "0.4" }
 lazy_static = "1"
 shared_library = { version = "0.1.9", optional = true }
-ash = "0.30"
+ash = "0.31"
 hal = { path = "../../hal", version = "0.5.2", package = "gfx-hal" }
```
kvark, line 29: sorry, one last thing - the `hal` requirement for affected backends needs to be updated

cwfitzgerald: Whoops, bad merge resolution

kvark:
> MULTI_DRAW_INDIRECT_COUNT

I think it would make sense to keep it `DRAW_INDIRECT_COUNT` as you had originally, since that matches Vulkan, which gfx-hal is close to. We can reconsider the naming at wgpu-core level, but here we have the strong Vulkan affinity.

kvark:
There is some breakage upstream, unfortunately. Filed https://github.com/servo/core-foundation-rs/issues/393 for the CI issue... We'll have to proceed in blind upon merging.

cwfitzgerald:
Alright, that should be all the changes needed. I have tested this using a modified version of the wgpu-rs texture-array example to try both multi-draw-indirect and multi-draw-indirect-count, so I think we're good to go.

kvark:
Bors r+

kvark:
Please don't forget to make one for `master`!

Issue #3278, opened 2020-06-26
Transform the empty backend into a mock


kvark:
Doing nothing is often better than panicking. For example, it would allow us to test the internal logic of gfx-memory: we could randomly spam allocations and check for consistency of the allocator internal state.

GabrielMajeri:
Alright, this issue is now partially solved and we've got fuzz tests for `gfx-memory`.

What would be the next step? Mocking the descriptor pool API to fuzz `gfx-descriptor`?

kvark:
Yes, that would be very useful as well!

PR #2764, opened 2019-04-30
Start turning empty backend into basic mock

zakarumych:
Few types at a time.

PR checklist:
- [ ] `make` succeeds (on *nix)
- [ ] `make reftests` succeeds
- [ ] tested examples with the following backends:
- [ ] `rustfmt` run on changed code

PR #3280, opened 2020-06-27
Start mocking the empty backend

GabrielMajeri:
Starts work on #3278 

I've stubbed enough functions to be able to get the `quad` example to compile, run, and do nothing in a loop.

Supersedes PR #2764 (which seems to not be progressing)

PR checklist:

### DIFF
--- a/src/backend/empty/src/lib.rs
+++ b/src/backend/empty/src/lib.rs
@@ -1,4 +1,4 @@
-//! Dummy backend implementation to test the code for compile errors
+//! Mock backend implementation to test the code for compile errors
 //! outside of the graphics development environment.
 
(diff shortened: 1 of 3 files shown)


src/backend/empty/src/lib.rs
```diff
@@ -427,19 +485,20 @@ impl device::Device<Backend> for Device {
     }
 
     unsafe fn map_memory(&self, _: &(), _: memory::Segment) -> Result<*mut u8, device::MapError> {
-        panic!(DO_NOT_USE_MESSAGE)
+        // 640 KB ought to be enough
```
kvark, line 488: that's awesome! would be good to have some sort of safe guards in place for this to not be overrun, still

PR #3281, opened 2020-06-27
Implement DRAW_INDIRECT_COUNT features

cwfitzgerald:
Forward-ports #3279 to master. The default impl in the hal was removed in favor of stub impls in all the backends. Addtionally the mesh shader function machinery was combined with the machinery #3279 adds to get extension pointers.


kvark:
Thank you!
CI is mad, otherwise good to go :)

PR #3283, opened 2020-06-28
Mock memory-related functions in the empty backend

GabrielMajeri:
Mock even more functions in the empty backend.
Should be enough to start fuzzing `gfx-memory`.

- The first commit only replaces the `panic`s with `unimplemented`s.
- Second commit adds the `Memory`, `Buffer` and `Image` types.

PR checklist:

### DIFF
--- a/src/backend/empty/src/memory.rs
+++ b/src/backend/empty/src/memory.rs
@@ -1,0 +1,58 @@
+use hal::{
+    device::{AllocationError, MapError, OutOfMemory},
+    memory::Segment,
+    MemoryTypeId,
+};
+use std::cell::UnsafeCell;
+use std::convert::TryInto;
+
+/// A memory allocation.
+#[derive(Debug)]
+pub struct Memory {
+    /// The type of memory this block was allocated from.
+    memory_type: MemoryTypeId,
+    /// Size of the allocated memory block.
+    size: u64,
+    /// The backing memory allocation.
+    data: UnsafeCell<Box<[u8]>>,
+}
+
+impl Memory {
+    pub fn allocate(memory_type: MemoryTypeId, size: u64) -> Result<Self, AllocationError> {
+        assert_eq!(memory_type.0, 0, "We only support one memory type");
+
+        // Allocate a buffer in RAM
+        let data = {
+            let size = size
+                .try_into()
+                // If we're on 32-bit and the given size is greater than 2^32,
+                // we certainly can't allocate it.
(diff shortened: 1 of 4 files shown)


src/backend/empty/src/memory.rs
```diff
@@ -0,0 +1,140 @@
...
+        }
+        if let Some(size) = segment.size {
+            if segment.offset + size > self.size {
+                return Err(MapError::OutOfBounds);
+            }
+        }
+
+        let order = Ordering::SeqCst;
+        self.mapped
+            .compare_exchange(false, true, order, order)
```
GabrielMajeri, line 61: I wanted to ensure the memory isn't mapped twice so I tried using an `AtomicBool`. It *seems* to achieve the intended behavior?

src/backend/empty/src/memory.rs
```diff
@@ -0,0 +1,140 @@
...
+                .compare_exchange(false, false, order, order)
+                .expect("Cannot free memory if still mapped");
+        }
+    }
+}
+
+unsafe impl Sync for Memory {}
+
+#[cfg(test)]
+mod tests {
```
kvark, line 90: we don't actually need to test this. Gfx-hal backends are inherently unsafe, they are free to assume the valid usage rules of Vulkan are followed.

src/backend/empty/src/memory.rs
kvark: as said in another comment, we don't even need to check if it's mapped. We need to assume the user knows what they are doing

rukai: Isnt it useful for a mock backend to check its usage?
A real backend will have vulkan validation layers (or similar) so this is not that different?
Or is it just not worth the complexity?

kvark: You can't validate *all* usage, it's a huge task that we can't accomplish.
So if we are to draw the line, it would be very close towards "not validating anything".

It also is a question of what the backend will be used for. Previously, we only used it to check if something builds.
Next, we want to use it for gfx-extras testing. This doesn't require any validation, it basically just need to not panic.

Are you suggesting that people use the backend at some later point to validate their workflows?

rukai: Honestly im not suggesting anything, I was just curious as to your reasoning here, and your response was enlightening so thankyou. :)

src/backend/empty/src/memory.rs
GabrielMajeri: Yep, I understand, removed them now

PR #3285, opened 2020-06-29
[0.5/dx11] Bump libloading to 0.6

cwfitzgerald:
This bumps libloading in dx11 to 0.6, this isn't a breaking change as it is all internal.

We now directly depend on both libloading 0.6 (dx11) and 0.5 (d3d12-rs), however we have indirectly depended on both (ash uses 0.6) for a while, so this is no worse. Once d3d12 gets a new release, this issue will go away.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-dx11-0.5.2 (29-07-2020)
+  - update libloading to 0.6
+
 ### backend-vulkan-0.5.11 (22-07-2020)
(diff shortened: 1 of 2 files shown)


kvark:
Blocked on d3d12-rs to update first

kvark:
@cwfitzgerald d3d12 is ready, could you rebase this and prepare for landing?

cwfitzgerald:
Only took me a month, but we're rebased and ready to go. d3d12-rs was a minor version change and this isn't _required_, so I left its dependency/version as is.

PR #3290, opened 2020-07-05
[0.5] Mock empty backend descriptor set functions

GabrielMajeri:
More work on #3278, with the intent to follow up with fuzz testing for the `gfx-descriptor` crate.

### DIFF
--- a/src/backend/empty/src/buffer.rs
+++ b/src/backend/empty/src/buffer.rs
@@ -1,7 +1,7 @@
 #[derive(Debug)]
 pub struct Buffer {
     /// Size of this buffer
-    pub size: u64,
+    pub(crate) size: u64,
 }
 
 impl Buffer {
--- a/src/backend/empty/src/descriptor.rs
+++ b/src/backend/empty/src/descriptor.rs
@@ -1,0 +1,42 @@
+use hal::pso;
+use log::debug;
+
+/// Dummy descriptor pool.
+#[derive(Debug)]
+pub struct DescriptorPool;
+
+impl pso::DescriptorPool<crate::Backend> for DescriptorPool {
+    unsafe fn allocate_set(
+        &mut self,
+        _layout: &DescriptorSetLayout,
+    ) -> Result<DescriptorSet, pso::AllocationError> {
+        Ok(DescriptorSet {
+            name: String::new(),
+        })
+    }
+
+    unsafe fn free_sets<I>(&mut self, descriptor_sets: I)
+    where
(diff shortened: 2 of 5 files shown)


src/backend/empty/src/buffer.rs
```diff
@@ -1,7 +1,7 @@
 #[derive(Debug)]
 pub struct Buffer {
     /// Size of this buffer
-    pub size: u64,
+    pub(crate) size: u64,
```
GabrielMajeri, line 4: I've also added this fix here because I've realized these types are publicly exported and these fields could otherwise be modified.

src/backend/empty/src/descriptor.rs
```diff
@@ -0,0 +1,42 @@
...
+        Ok(DescriptorSet {
+            name: String::new(),
+        })
+    }
+
+    unsafe fn free_sets<I>(&mut self, descriptor_sets: I)
+    where
+        I: IntoIterator<Item = DescriptorSet>,
+    {
+        descriptor_sets.into_iter().for_each(std::mem::drop);
```
kvark, line 22: it's a bit weird to call `drop` on the explicitly if we can just iterate them and let them be dropped naturally, i.e.
```rust
for _ in descriptor_sets {}
```

src/backend/empty/src/descriptor.rs
```diff
@@ -0,0 +1,42 @@
...
+        debug!("Resetting descriptor pool");
+    }
+}
+
+#[derive(Debug)]
+pub struct DescriptorSetLayout {
+    /// User-defined name for this descriptor set layout
+    pub(crate) name: String,
+    /// Bindings contained in this layout
+    pub(crate) bindings: Vec<pso::DescriptorSetLayoutBinding>,
```
kvark, line 35: do we need these fields?

GabrielMajeri: I thought it would add some "weight" to the structure, to simulate the fact that a real implementation would also use _some_ memory to retain this kind of data.

GabrielMajeri: I've removed the `bindings` field, but I'll keep the name since it might be useful for debugging (through `set_descriptor_set_layout_name`) and also adds some "weight".

kvark:
bors retry

kvark:
Trying to fix the CI in https://github.com/gfx-rs/gfx/pull/3293

kvark:
borsr retry

kvark:
@GabrielMajeri it's published!

PR #3291, opened 2020-07-05
[master] Mock empty backend descriptor set functions

GabrielMajeri:
Mirror of #3290 but for `master` branch.


kvark:
bors retry

PR #3297, opened 2020-07-10
[0.5/vk] skip unknown memory properties

kvark:
Should address https://github.com/gfx-rs/wgpu/issues/718
~~It's a bit risky, but that's the recommendation from Vulkan so far.~~

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -531,6 +531,12 @@ impl hal::Instance<Backend> for Instance {
                     handle: device,
                     extensions,
                     properties,
+                    known_memory_flags:
+                        vk::MemoryPropertyFlags::DEVICE_LOCAL |
+                        vk::MemoryPropertyFlags::HOST_VISIBLE |
+                        vk::MemoryPropertyFlags::HOST_COHERENT |
+                        vk::MemoryPropertyFlags::HOST_CACHED |
+                        vk::MemoryPropertyFlags::LAZILY_ALLOCATED,
                 };
                 let queue_families = unsafe {
                     self.raw
@@ -646,6 +652,7 @@ pub struct PhysicalDevice {
     handle: vk::PhysicalDevice,
     extensions: Vec<vk::ExtensionProperties>,
     properties: vk::PhysicalDeviceProperties,
+    known_memory_flags: vk::MemoryPropertyFlags,
 }
 
 impl PhysicalDevice {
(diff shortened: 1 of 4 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -872,50 +886,43 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+            if vk_flags.take(vk::MemoryPropertyFlags::LAZILY_ALLOCATED) {
+                type_flags |= Properties::LAZILY_ALLOCATED;
+            }
+
+            if !vk_flags.is_empty() {
+                warn!("Stopping memory type iteration on unknown flags {:?}", vk_flags);
+                // Vulkan guarantees that all the fancy memory types come after
+                // the non-fancy ones, so we can stop here.
+                break
+            }
```
Ralith, line 919: This condition is incorrect. [§10.2](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-device):

> There is no ordering requirement between [any two elements of memoryTypes]  for the case their propertyFlags members are not in a subset relation.

An unknown memory property flag may appear at any position in the sequence, so long as it does so in a type whose properties are not a superset of those of a type that appears later. For example, the very first memory type could be a type with exactly one unknown property, causing this loop to unconditionally fail.

Instead, how about `.iter().filter(|x| SUPPORTED_FLAGS.contains(x.property_flags))`?

kvark: I see. I was hoping we can get away with just taking the first N. Since once we start skipping elements, we'll need to introduce the bidirectional mapping inside a device, to know how user IDs map to Vulkan IDs and visa versa

Ralith: I don't have much context for what you're trying to do here, but `VK_MAX_MEMORY_TYPES` is only 32, so a bidirectional lookup table shouldn't be a big deal.

kvark: Yeah, it's not really a big deal. Just something I was hoping to avoid :)

kvark:
@Ralith please take another look, this is now rewritten

src/backend/vulkan/src/lib.rs
```diff
@@ -646,6 +652,7 @@ pub struct PhysicalDevice {
     handle: vk::PhysicalDevice,
     extensions: Vec<vk::ExtensionProperties>,
     properties: vk::PhysicalDeviceProperties,
+    known_memory_flags: vk::MemoryPropertyFlags,
```
Ralith, line 655: Could this be a `const`?

kvark: nope, ash's bitflags are not const or-able

Ralith: You could convert to raw and back, though that's a bit ugly.

kvark:
bors r=Ralith

PR #3299, opened 2020-07-11
[vk] filter out known memory types

### LINKED ISSUE #3296: Skip unknown memory types returned by Vulkan
See discussion in https://github.com/KhronosGroup/Vulkan-Docs/issues/1298 and https://github.com/gfx-rs/wgpu/issues/718
gfx-hal needs to only expose memory types that it can work with, or reason about. Therefore, we can't just pass through the information about them between the app and the driver. We need to have some sort of mapping in gfx-hal to convert between exposed memory types and whatnot.

kvark:
Fixes #3296
Sibling of #3297 for master
PR checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Change Log
 
+### backend-vulkan-0.5.10 (10-07-2020)
+  - skip unknown memory types
+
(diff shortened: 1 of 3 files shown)


kvark:
bors r=Ralith

PR #3304, opened 2020-07-14
replaced PrimitiveAssembler vecs with slices

### LINKED ISSUE #3259: Avoid heap allocations in GraphicsPipelineDesc
It currently has a lot of `Vec` arguments. This should be unnecessary.

khodzha:
Fixes #3259
PR checklist:
- [X] tested examples with the following backends: gl vulkan (compilation only)


khodzha:
not sure whether this is what was needed :shrug: but yeah

kvark:
That's an improvement, thank you!
Generally, we go even further and make the container generic. But I don't think we need to require it here.
bors r+

PR #3307, opened 2020-07-16
replaced heap allocated vecs with stack allocated arrays for vulkan backend

### LINKED ISSUE #3201: Use inplace_it in Vulkan for alloca
We have this performance issue in Vulkan backend that gfx-rs API is based on iterators, and Ash expects slices, so we need some extremely lightweight way of producing a slice, sometimes without knowing anything about the number of elements. Basically, we want `alloca`, but it's not available in Rust.

However there is a simple crate that does something similar - https://github.com/NotIntMan/inplace_it . We should try it out and see if we can remove all the heap allocations from hot paths of the Vulkan backend, and possibly in other backends too.

khodzha:
Fixes #3201

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -559,10 +559,13 @@ impl d::Device<B> for Device {
     where
         IA: IntoIterator,
         IA::Item: Borrow<pass::Attachment>,
+        IA::IntoIter: ExactSizeIterator,
         IS: IntoIterator,
         IS::Item: Borrow<pass::SubpassDesc<'a>>,
+        IS::IntoIter: ExactSizeIterator,
         ID: IntoIterator,
         ID::Item: Borrow<pass::SubpassDependency>,
+        ID::IntoIter: ExactSizeIterator,
     {
         let attachments = attachments
             .into_iter()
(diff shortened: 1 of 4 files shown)


khodzha:
PR checklist:
- [X] tested examples with the following backends: vulkan

kvark:
The inplace PR got merged and published 🎉  Please change the required version to 0.3.1 and use the new function to simplify this implementation.

khodzha:
@kvark is it okay to use stack for vecs here:
https://github.com/gfx-rs/gfx/blob/c8cce08ea21450543f096109e7281b89b49b0814/src/backend/vulkan/src/device.rs#L1611-L1614
https://github.com/gfx-rs/gfx/blob/c8cce08ea21450543f096109e7281b89b49b0814/src/backend/vulkan/src/device.rs#L1617-L1622
https://github.com/gfx-rs/gfx/blob/c8cce08ea21450543f096109e7281b89b49b0814/src/backend/vulkan/src/device.rs#L1624-L1639

kvark:
> is it okay to use stack for vecs here?

Yeah, I think we want to use the stack there as well.

src/backend/vulkan/src/device.rs
```diff
@@ -666,32 +605,103 @@ impl d::Device<B> for Device {
...
+                    p_attachments: attachments.as_ptr(),
+                    subpass_count: subpasses.len() as u32,
+                    p_subpasses: subpasses.as_ptr(),
+                    dependency_count: dependencies.len() as u32,
+                    p_dependencies: dependencies.as_ptr(),
+                };
+
+                let result = self.shared.raw.create_render_pass(&info, None);
+
+                match result {
```
kvark, line 694: let's try to move things out of `inplace_or_alloc_array` where possible. In this case, for example, we should do
```rust
let result = inplace_it::inplace_or_alloc_array(dependencies.len(), |uninit_guard| {
  ..
});
match result {
  ...
}
```

src/backend/vulkan/src/device.rs
```diff
@@ -721,27 +730,38 @@ impl d::Device<B> for Device {
...
+                    flags: vk::PipelineLayoutCreateFlags::empty(),
+                    set_layout_count: set_layouts.len() as u32,
+                    p_set_layouts: set_layouts.as_ptr(),
+                    push_constant_range_count: push_constant_ranges.len() as u32,
+                    p_push_constant_ranges: push_constant_ranges.as_ptr(),
+                };
+
+                let result = self.shared.raw.create_pipeline_layout(&info, None);
+
+                match result {
```
kvark, line 757: same here, we should move that out

src/backend/vulkan/src/device.rs
```diff
@@ -798,24 +818,29 @@ impl d::Device<B> for Device {
...
+            let caches = uninit_guard.init_with_iter(caches);
+
+            let result = self.shared.raw.fp_v1_0().merge_pipeline_caches(
+                self.shared.raw.handle(),
+                target.raw,
+                caches.len() as u32,
+                caches.as_ptr(),
+            );
+
+            match result {
```
kvark, line 837: an here (and everywhere else)

src/backend/vulkan/src/device.rs
```diff
@@ -1621,34 +1652,42 @@ impl d::Device<B> for Device {
                 .collect::<Vec<_>>(),
```
khodzha, line 1652: still `Vec` here since `DescriptorSetLayout` wants `Arc<Vec>` :thinking:

kvark: that's fine, creating descriptor set layouts is basically init time

src/backend/vulkan/src/device.rs
```diff
@@ -1920,16 +1967,23 @@ impl d::Device<B> for Device {
...
-            d::WaitFor::Any => false,
-            d::WaitFor::All => true,
-        };
-        let result = self.shared.raw.wait_for_fences(&fences, all, timeout_ns);
+            .map(|fence| fence.borrow().0);
+
+        let result = inplace_it::inplace_or_alloc_array(fences.len(), |uninit_guard| {
+            let fences = uninit_guard.init_with_iter(fences);
+
+            let all = match wait {
```
kvark, line 1979: we should keep it out of the closure

khodzha: agreed, fixed

kvark:
bors retry

kvark:
bors retry

PR #3310, opened 2020-07-18
Bump cocoa and core-graphics deps for metal backend

elrnv:
This fixes build issues on macOS caused by multiple versions of cocoa
dependencies. core-graphics is also updated to be consistent with cocoa v0.22,
which removes a duplication.

The build passes and test_send_sync passes with this PR on my macOS 10.15.5 machine.

PR #3311, opened 2020-07-18
Bump cocoa and core-graphics deps for metal backend

elrnv:
This fixes build issues on macOS caused by multiple versions of cocoa
deps. core-graphics is also updated to be consistent with cocoa v0.22,
which removes a duplication.

hal-0.5 version of #3310 

PR checklist:
- [ ] `make reftests` succeeds (fails on my machine, but this perhaps should be filed in another issue)
- [x] tested examples with the following backends: `metal`

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -29,7 +29,7 @@ bitflags = "1.0"
 copyless = "0.1.4"
 log = { version = "0.4" }
 dispatch = { version = "0.2", optional = true }
-metal = { version = "0.18", features = ["private"] }
+metal = { version = "0.19", features = ["private"] }
 foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-backend-metal"
-version = "0.5.4"
+version = "0.5.5"
 description = "Metal API backend for gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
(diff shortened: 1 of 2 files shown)


src/backend/metal/Cargo.toml
```diff
@@ -33,8 +33,8 @@ metal = { version = "0.18", features = ["private"] }
 foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
-cocoa = "0.20"
-core-graphics = "0.19"
+cocoa = "0.22"
```
kvark, line 36: let's also switch `metal` dependecy to `"0.18.1"

parasyte:
This issue is breaking macOS builds all the way down to crates depending on `wgpu`. Glad to see fixes are in progress, but this issue is quite critical.

jrmuizel:
What's the build error? Why wasn't it caught by CI?

elrnv:
I guess we can wait until 0.18.2?

elrnv:
@jrmuizel The specific build error looks like
```
error[E0308]: mismatched types
    --> src/backend/metal/src/command.rs:1808:17
     |
1808 | /                 NSRange {
1809 | |                     location: range.start,
1810 | |                     length: range.end - range.start,
1811 | |                 },
     | |_________________^ expected struct `cocoa::foundation::NSRange`, found a different struct `cocoa::foundation::NSRange`
     |
     = note: perhaps two different versions of crate `cocoa` are being used?
```

as helpfully noted by the compiler, this is due to the interaction of code from two different cocoa versions.
I believe the reason is because we didn't have a patch version on the metal dependency, which updated its cocoa dependency in 0.18.1, which was a breaking change I guess. So I think the reason why it wasn't caught by CI earlier is because it's a semver violation on a dependency.
Please correct me if I'm wrong.

kvark:
> I guess we can wait until ~0.18.2~ 0.19?

yes, we'll need to get 0.19 rolling tonight. You can make this PR to rely on 0.19, so that no changes are required when it's ready.

I wonder if we get away from depending on `core-graphics` entirely... (like metal-rs does now) but this could be a different story.

PR #3312, opened 2020-07-20
Remove core-graphics deps

Gordon-F:
Same as https://github.com/gfx-rs/metal-rs/pull/157
PR checklist:

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -29,12 +29,11 @@ bitflags = "1.0"
 copyless = "0.1.4"
 log = { version = "0.4" }
 dispatch = { version = "0.2", optional = true }
-metal = { version = "0.18", features = ["private"] }
+metal = { version = "0.19", features = ["private"] }
 foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
 cocoa = "0.22"
-core-graphics = "0.21"
 smallvec = "1"
 spirv_cross = { version = "0.20", features = ["msl"] }
 parking_lot = "0.10"
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -65,14 +65,13 @@ use hal::{
 use range_alloc::RangeAllocator;
 
 use cocoa::foundation::NSInteger;
-use core_graphics::base::CGFloat;
-use core_graphics::geometry::CGRect;
 #[cfg(feature = "dispatch")]
 use dispatch;
 use foreign_types::ForeignTypeRef;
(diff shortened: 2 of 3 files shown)


src/backend/metal/Cargo.toml
```diff
@@ -29,12 +29,11 @@ bitflags = "1.0"
 copyless = "0.1.4"
 log = { version = "0.4" }
 dispatch = { version = "0.2", optional = true }
-metal = { version = "0.18", features = ["private"] }
+metal = { version = "=0.18.0", features = ["private"] }
```
kvark, line 32: I just published 0.19, let's use it

src/backend/metal/Cargo.toml
```diff
@@ -29,12 +29,11 @@ bitflags = "1.0"
...
 foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
-cocoa = "0.22"
-core-graphics = "0.21"
+cocoa = "0.20"
```
kvark, line 36: this shouldn't revert back the cocoa version

src/backend/metal/src/lib.rs
```diff
@@ -102,6 +100,59 @@ pub type GraphicsCommandPool = CommandPool;
 /// A type representing Metal binding's resource index.
 type ResourceIndex = u32;
 
+#[cfg(target_pointer_width = "64")]
+pub type CGFloat = f64;
```
kvark, line 104: Since `metal-0.19` publicly exports `CGFloat` and `CGSize`, let's use them

src/backend/metal/src/lib.rs
```diff
@@ -102,6 +100,59 @@ pub type GraphicsCommandPool = CommandPool;
...
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default)]
+pub struct CGRect {
+    pub origin: CGPoint,
+    pub size: CGSize
+}
+
+impl CGRect {
+    #[inline]
+    pub fn new(origin: &CGPoint, size: &CGSize) -> CGRect {
```
kvark, line 148: we shouldn't get those by reference if all we do is copying them. Let's accept them by value

jrmuizel:
FWIW, I just landed a change to create a new `core-graphics-types` crate that should cover this usecase:
https://github.com/servo/core-foundation-rs/tree/master/core-graphics-types

kvark:
Alright, let's use it! Thanks @jrmuizel

kvark:
We need the same PR now against hal-0.5 branch

jrmuizel:
It's on crates.io now

PR #3318, opened 2020-07-22
use stack allocated arrays in CommandBuffer in vulkan backend

### LINKED ISSUE #3309: Use ExactSizeIterator bound in command encoding methods
As a follow-up to https://github.com/gfx-rs/gfx/pull/3307 we'd want to put `ExactSizeIterator` bound on the methods that are called often, methods of `trait CommandBuffer`.
cc @khodzha

khodzha:
Fixes #3309

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -4140,6 +4140,7 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
     ) where
         T: IntoIterator,
         T::Item: Borrow<com::ImageCopy>,
+        T::IntoIter: ExactSizeIterator,
     {
         match (&src.like, &dst.like) {
             (&native::ImageLike::Unbound { .. }, _) | (_, &native::ImageLike::Unbound { .. }) => {
@@ -4244,6 +4245,7 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
     ) where
         T: IntoIterator,
         T::Item: Borrow<com::BufferImageCopy>,
+        T::IntoIter: ExactSizeIterator,
     {
         match dst.like {
             native::ImageLike::Unbound { .. } => {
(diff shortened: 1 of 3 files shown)


khodzha:
PR checklist:
- [X] tested examples with the following backends (gl, vulkan compilation only)

khodzha:
there is this `SmallVec` left:
https://github.com/gfx-rs/gfx/blob/210b54f225a460628d700b4562a86331c6d22409/src/backend/vulkan/src/command.rs#L49

but I couldnt figure out the return type without it

it should be something like `Map<T, SomeClosure>`, but idk how the closure is typed
rustc says something like:
> found struct `std::iter::Map<_, [closure@src/backend/vulkan/src/command.rs:59:14: 70:10]>`

which doesnt help :shrug:

kvark:
@khodzha oh, this one is simple, actually. We don't need to return a stack-allocated vector from this function, we want to return an iterator:
```rust
fn map_buffer_image_regions<T>(_image: &n::Image, regions: T) -> impl Iterator<item = vk::BufferImageCopy>
```
And then after invocation we'd do stack allocation for the results (based on the length of the original iterator), to pass down a slice to Ash

khodzha:
there are `smallvec`s in `BarrierSet` and other places still

khodzha:
also there is this `bind_vertex_buffers` which does 
https://github.com/gfx-rs/gfx/blob/210b54f225a460628d700b4562a86331c6d22409/src/backend/vulkan/src/command.rs#L539-L543

and unzip() automatically collects here :shrug:

kvark:
ok, it sounds like we can follow-up with more changes then :)

khodzha:
is there something to be done about smallvecs embedded into structs? or we just leave them be?

kvark:
uh, it's not clear to me how we can resolve `fn destructure_barriers` gracefully

khodzha:
there is also this struct https://github.com/gfx-rs/gfx/blob/210b54f225a460628d700b4562a86331c6d22409/src/backend/vulkan/src/device.rs#L462-L466

Issue #1725, opened 2018-01-06
Descriptor set copies


msiglreith:
We currently only support descriptor set updates via writes. Updating via copying is currently missing.

kvark:
Note that copies are now implemented on Vulkan and D3D12.

PR #3322, opened 2020-08-01
Implement descriptor copy for opengl

GabrielMajeri:
Implements #1725 for the OpenGL backend.

PR checklist:
- [X] `make` succeeds
- [X] tested a modified example which copies between two descriptors with the OpenGL backend

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -1627,8 +1627,32 @@ impl d::Device<B> for Device {
         I: IntoIterator,
         I::Item: Borrow<pso::DescriptorSetCopy<'a, B>>,
     {
-        for _copy in copies {
-            unimplemented!()
+        for copy in copies {
+            let copy = copy.borrow();
+
+            let src_set = &copy.src_set;
+            let dst_set = &copy.dst_set;
+            if std::ptr::eq(src_set, dst_set) {
+                panic!("copying within same descriptor set is not currently supported");
+            }
+
+            let src_bindings = src_set.bindings.lock();
+            let mut dst_bindings = dst_set.bindings.lock();
+
+            let count = copy.count;
+
+            // TODO: add support for array bindings when the OpenGL backend gets them
+            let src_start = copy.src_binding as usize;
+            let src_end = src_start + count;
+            assert!(src_end <= src_bindings.len());
+
+            let src_slice = &src_bindings[src_start..src_end];
+
+            let dst_start = copy.dst_binding as usize;
+            let dst_end = dst_start + count;
+            assert!(dst_end <= dst_bindings.len());
+
+            dst_bindings[dst_start..dst_end].clone_from_slice(src_slice);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/backend/gl/src/device.rs
```diff
@@ -1627,8 +1627,31 @@ impl d::Device<B> for Device {
...
-        for _copy in copies {
-            unimplemented!()
+        for copy in copies {
+            let copy = copy.borrow();
+
+            let src_set = &copy.src_set;
+            let dst_set = &copy.dst_set;
+            if std::ptr::eq(src_set, dst_set) {
+                panic!("copying within same descriptor set is not currently supported");
+            }
```
GabrielMajeri, line 1637: The Vulkan API allows these in theory (as long as the ranges do not overlap), however the `gfx-hal` docs do not specify whether this is permitted.

GabrielMajeri:
The CI build seems to be unrelated, Travis failed to checkout the code

src/backend/gl/src/device.rs
```diff
@@ -1627,8 +1627,31 @@ impl d::Device<B> for Device {
...
+
+            let src_bindings = src_set.bindings.lock();
+            let mut dst_bindings = dst_set.bindings.lock();
+
+            let count = copy.count;
+
+            let src_start = copy.src_binding as usize + copy.src_array_offset;
```
kvark, line 1644: this seems wrong. If all the bindings are laid out with all the array elements flattened, then we'd need to sum up all the array counts for descriptor bindings `0 .. copy.src_binding`

src/backend/gl/src/device.rs
```diff
@@ -1627,8 +1627,31 @@ impl d::Device<B> for Device {
...
+            let src_end = src_start + count;
+            assert!(src_end <= src_bindings.len());
+
+            let src_slice = &src_bindings[src_start..src_end];
+
+            let dst_start = copy.dst_binding as usize + copy.dst_array_offset;
```
kvark, line 1650: same here

src/hal/src/pso/descriptor.rs
```diff
@@ -275,17 +280,34 @@ pub enum Descriptor<'a, B: Backend> {
...
+    /// *Note*: when there are more descriptors required than
+    /// array elements left in the specified binding starting
+    /// at the specified offset, the updates are taken from
+    /// the next binding (starting with offset 0), and so on.
     pub src_binding: DescriptorBinding,
+    /// Offset into the descriptor array to start copying from.
     pub src_array_offset: DescriptorArrayIndex,
+    /// Descriptor set to copy to.
     pub dst_set: &'a B::DescriptorSet,
+    /// Binding to copy to.
```
kvark, line 301: wow, thank you for all the docs!

src/backend/gl/src/device.rs
GabrielMajeri: From what I can understand, these are the types of bindings OpenGL supports:
```rust
pub(crate) enum DescSetBindings {
    Buffer { /* ... */ },
    Texture(/* ... */),
    Sampler(/* ... */),
    SamplerDesc(/* ... */),
}
```
It seems OpenGL doesn't support array bindings?

kvark: In that case, `src_array_offset` has to be zero, and we don't need to check it. We can have a "TODO: support array bindings" line

GabrielMajeri: Ok, I've updated the PR to add a `// TODO` for when GL gets support for array bindings.

PR #3323, opened 2020-08-02
WIP: Feature multiview

### LINKED ISSUE #3303: Add support for Multiview (Multi Viewport)
From the [Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.2-khr-extensions/html/chap44.html#_new_spir_v_capabilities_7):
# VK_KHR_multiview
This extension has the same goal as the OpenGL ES GL_OVR_multiview extension - it enables rendering to multiple “views” by recording a single set of commands to be executed with slightly different behavior for each view. It includes a concise way to declare a render pass with multiple views, and gives implementations freedom to render the views in the most efficient way possible.

---
I am not familiar with how/if other backend APIs provide the same functionality.

Limeth:
Once finished, Fixes #3303.
Hopefully the API usage is correct. An example demonstrating the Multiview feature would be desirable. Feedback is appreciated.

PR checklist:


src/backend/vulkan/src/lib.rs
```diff
@@ -650,11 +652,65 @@ impl queue::QueueFamily for QueueFamily {
     }
 }
 
+pub struct PhysicalDeviceMultiviewProperties {
+    pub max_multiview_view_count: u32,
+    pub max_multiview_instance_index: u32,
+}
```
Limeth, line 658: Substitute to `vk::PhysicalDeviceMultiviewProperties` without any pointers that would prevent `PhysicalDeviceProperties` below from being `Sync`.

src/hal/src/lib.rs
```diff
@@ -242,21 +242,28 @@ bitflags! {
...
 
         /// Supports task shader stage.
-        const TASK_SHADER = 0x0001 << 96;
+        const TASK_SHADER = 0x0000_0001_0000_0000 << 64;
         /// Supports mesh shader stage.
-        const MESH_SHADER = 0x0002 << 96;
+        const MESH_SHADER = 0x0000_0002_0000_0000 << 64;
+
+        /// Supports multiview
+        const MULTIVIEW = 0x0001_0000_0000_0000 << 64;
```
kvark, line 262: this should be `0x0000_0004_0000_0000` and so on. No need to leave such a big gap with the mesh shader

src/backend/vulkan/src/lib.rs
```diff
@@ -650,11 +652,65 @@ impl queue::QueueFamily for QueueFamily {
...
+                .any(|ep| unsafe { CStr::from_ptr(ep.extension_name.as_ptr()) } == extension)
+        }
+
+        if let Some(ref get_device_properties) = instance.raw.get_physical_device_properties {
+            let mut properties2_builder = vk::PhysicalDeviceProperties2KHR::builder()
+                .properties(vk::PhysicalDeviceProperties::builder().build());
+            let mut multiview = None;
+
+            // Add extension infos to the p_next chain
+            if supports_extension(extensions, *KHR_MULTIVIEW) {
```
kvark, line 687: something is not right here. Why are we checking for this extension in two places?

src/backend/vulkan/src/lib.rs
```diff
@@ -650,11 +652,65 @@ impl queue::QueueFamily for QueueFamily {
...
+
+                properties2_builder = properties2_builder.push_next(multiview.as_mut().unwrap());
+            }
+
+            let mut properties2 = properties2_builder.build();
+            unsafe { get_device_properties.get_physical_device_properties2_khr(*device, &mut properties2 as *mut _); }
+
+            Self {
+                properties: properties2.properties,
+                multiview: multiview.map(|multiview| multiview.into()),
```
kvark, line 698: nit: could write this as `muiltiview.map(PhysicalDeviceMultiviewProperties::from)`

src/backend/vulkan/src/lib.rs
Limeth: This check is performed only once during the initialization of `PhysicalDevice`s, I have now replaced the same check within `PhysicalDevice::features()`: https://github.com/gfx-rs/gfx/pull/3323/files#diff-44198a1b71ecb41e695db8b6967cd689R1028

Limeth:
The @monocodus bot seems to have been recommending odd style changes, such as incorrect indentation, so I did not apply their suggestions this time.
I have tried creating a Multiview example derived from the `quad` example, but I got stuck when implementing the synchronization logic, as I do not feel confident in my understanding of it and how it is structured in gfx.

kvark:
@Limeth we can deal with monocodus later.
The synchronization model is exactly what Vulkan has.

P-ignatovich:
@Limeth Hey, I noticed you're having some problem with monocodus. I've tried to run rustfmt locally on a given code to figure out which part of the suggestion is wrong but it seems like it gives the same result as listed in the suggestion. Could you please point out the wrong indentation in the suggestion?

Also please feel free to use our [issue tracker](https://github.com/monocodus/issues-tracker/issues/new/choose) to describe what exactly is wrong, so that our team can resolve the issue in a timely manner.

Limeth:
@P-ignatovich Please see the suggested changes I approved previously: https://github.com/gfx-rs/gfx/pull/3323/commits/2bfbd4d4efaa4bc127e8392f679cf51cfc2e2c43

P-ignatovich:
@Limeth Thank you for your answer!
I took a closer look at these changes and found the issues you mentioned. They're caused by incorrect processing of huge formatting fixes by monocodus. We'll see what we can do about it.

For now, I've disabled rustfmt in out default config so users don't get confused with its behavior.

Thank you very much for your report! Hope we'll make it work better soon :)

kvark:
@Limeth are you going to come back to this PR?

Limeth:
I would like to, but I've been hesitant because there's some kind of UB making it crash on the example, and I couldn't figure out what it was caused by.

Issue #3319, opened 2020-07-25
Stack overflow running make reftests


hgallagher1993:
Short info header:
- GFX version: Master
- OS: Linux Mint 18.3
- GPU: 

Output of `make reftests`:

```
Parsing test suite 'local'...
Testing Vulkan:
	Scene 'compute':

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Aborted
Makefile:72: recipe for target 'reftests' failed
make: *** [reftests] Error 134
```

Did a small bit of investigating with git bisect and it seems to be #3307 that's causing it.

khodzha:
what i found is that it `SIGABRT`s here:

https://github.com/gfx-rs/gfx/blob/3fce1bd0550110f0565018fd2b4b74a4da151c62/src/backend/vulkan/src/device.rs#L1674

while entering this function (i mean the call itself fails):
https://github.com/NotIntMan/inplace_it/blob/03fb87daf35e6f0d792d52702246e12ac90fd950/src/fixed_array.rs#L104-L105

but then my debugging falls short since i have no idea how to inspect/debug stack sizes for closures(? is that even proper thing to do? :shrug: )

maybe nesting `inplace_it::inplace_or_alloc_array` was doomed from the start? :shrug:

kvark:
that's scary... but easy to check in a standalone example. Maybe file an issue to their repo?

khodzha:
I couldn't reproduce it with simple Vecs of i32. But digging into asm this is what i found (take it with a grain of salt since im not fluent in all this native stackframe shenanigans):

`__rust_probestack` after each call walks whole stackframe (in steps of 1000 bytes i think) to check whether it will cause segfault.
Idk where the actual size of each stackframe comes from exactly but:
- it's requested to be `19_106_544` before we segfault and on my machine stacksize is 8mbs
- it's somehow related to `inplace` macro in https://github.com/NotIntMan/inplace_it/blob/03fb87daf35e6f0d792d52702246e12ac90fd950/src/fixed_array.rs#L104


If i change this macro to allocate at most 1632 elems the `quad` example starts working fine again. I suppose this depends on element type & bytesize
https://github.com/NotIntMan/inplace_it/blob/03fb87daf35e6f0d792d52702246e12ac90fd950/src/fixed_array.rs#L239

And if i change the macro to 128 elems max all `make reftests` pass for vulkan without stack overflows.

[...]

PR #3326, opened 2020-08-03
update inplace_it

khodzha:
fixes excessive stack consumption in inplace_it impl that was leading to stack overflow crashes


khodzha:
Fixes #3319
PR checklist:
- [X] `make reftests` succeeds (with vulkan backend!)
- [X] tested examples with the following backends: quad.rs with vulkan backend

PR #3332, opened 2020-08-14
Unbounded subresource range

### LINKED ISSUE #3200: Unbound image subresources
We should remove the use of `Range` in `image::SubresourceLayers` and `image::SubresourceRange`, and instead pass count of layers/mipmaps as `Option<NonZero>`.  This would correspond to `VK_REMAINING_ARRAY_LAYERS` and `VK_REMAINING_MIP_LEVELS` in vulkan. We could also take this opportunity to refine the terms used for fields and comments.

GabrielMajeri:
Fixes #3200 

I've only updated the GL and Vulkan backends, since those are the only ones I have access to.

PR checklist:
- [X] tested examples with the following backends: GL, Vulkan


src/hal/src/image.rs
```diff
@@ -659,10 +676,20 @@ pub struct SubresourceLayers {
...
+    /// A value of `None` indicates the subresource contains
+    /// all of the remaining levels.
+    pub level_count: Option<NonZeroU8>,
+    /// First layer in this subresource
+    pub base_layer: Layer,
+    /// Number of sequential layers in this subresource.
+    ///
+    /// A value of `None` indicates the subresource contains
+    /// all of the remaining layers.
+    pub layer_count: Option<NonZeroU16>,
```
GabrielMajeri, line 692: These are `U8`/`U16` to match the types `Level` and `Layer`

src/backend/gl/src/device.rs
```diff
@@ -1474,40 +1474,42 @@ impl d::Device<B> for Device {
         range: i::SubresourceRange,
     ) -> Result<n::ImageView, i::ViewCreationError> {
         //TODO: check if `layers.end` covers all the layers
-        let level = range.levels.start;
-        assert_eq!(level + 1, range.levels.end);
+        let level = range.base_level;
+        assert_eq!(1, range.level_count.unwrap().get());
```
GabrielMajeri, line 1478: I don't know if I can do anything except unwrap here

kvark:
Alright, I've spent quite a bit of time converting this (pushing to this PR), and I have opinions now.

First and foremost, it turns out that Vulkan *does* require the exact layer count for `SubresourceLayers`. That means, only `SubresourceRanges` has this semantic of "remaining" stuff, and it's only used in a couple of places (notably, image creation and barriers).

Secondly, `wgpu` doesn't care about this, since it has to know and validate all the ranges anyway.
And direct users of the API unlikely care either. So this is more or less a "nice to have" thing rather than necessity.

As a result, I'm thinking of just redoing this from the ground up, again, in a separate PR.
Sorry for dragging you @GabrielMajeri into this, your help is still hugely appreciated!
This issue though I'll have to take care myself, since it requires building on all platforms.

GabrielMajeri:
It wasn't wasted time, since we learned about how we should redesign this API in the future 🙂

PR #3333, opened 2020-08-15
Support unbound subresource ranges

### LINKED ISSUE #3200: Unbound image subresources

kvark:
Fixes #3200
PR checklist:
- [x] tested examples with the following backends: Metal, DX12

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -101,12 +101,6 @@ const QUAD: [Vertex; 6] = [
     },
 ];
 
-const COLOR_RANGE: i::SubresourceRange = i::SubresourceRange {
-    aspects: f::Aspects::COLOR,
-    levels: 0 .. 1,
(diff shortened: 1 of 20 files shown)


kvark:
We do a bit of cleanup on names in the `image` module.
Notably, `num_levels` is no longer a footgun: it looks like it gives you how many mipmaps an image has, but in fact it compute the maximum of what it could have. So now it's renamed to `compute_num_levels`.
Reviewing this must be fun! Please free to do so as it lands and afterwards :)
bors r+

PR #3336, opened 2020-08-15
Dependency updates, mainly parking_lot

kvark:
Just preparing for the release.
Note that GL's windowing features are totally disabled. That is to allow us to introduce them later without breaking.
Currently GL backend is considered broken.


kvark:
cc @zicklag 
bors r+

zicklag:
Oh, good idea. I'm going to be working as much as I can to get the OpenGL stuff ready over this weekend, but I'm not sure how far I'll get. I've got to get this over with before I just don't do it. :slightly_smiling_face:

PR #3340, opened 2020-08-17
GL Final Showdown

zicklag:
WIP getting GL finally working again with `surfman`.

PR checklist:
- [ ] `make reftests` succeeds ( Fails because of things unimplemented in GL backend. Out of scope for this PR )
    - vulkan
    - gl

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -35,7 +35,6 @@ image = "0.23"
 log = "0.4"
 hal = { path = "../src/hal", version = "0.6", package = "gfx-hal" }
 auxil = { path = "../src/auxil/auxil", version = "0.5", package = "gfx-auxil" }
-gfx-backend-gl = { path = "../src/backend/gl", version = "0.6", optional = true }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.6" }
 winit = { version = "0.22", features = ["web-sys"] }
 
@@ -54,6 +53,12 @@ features = [ "console", "Document", "Element", "HtmlElement", "Node", "Window" ]
 
 [dependencies.gfx-backend-vulkan]
 path = "../src/backend/vulkan"
+version = "0.6"
+features = ["x11"]
+optional = true
+
+[target.'cfg(unix)'.dependencies.gfx-backend-gl]
+path = "../src/backend/gl"
 version = "0.6"
 features = ["x11"]
 optional = true
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -17,10 +17,8 @@ build = "build.rs"
 name = "gfx_backend_gl"
 
 [features]
-#default = ["wgl", "surfman", "surfman-x11"]
-default = []
-wgl = ["winapi"]
-#surfman-x11 = ["surfman/sm-x11"]
+default = ["surfman"]
+x11 = ["surfman/sm-x11"]
 
 [dependencies]
 arrayvec = "0.5"
@@ -34,12 +32,12 @@ parking_lot = "0.11"
 spirv_cross = { version = "0.20", features = ["glsl"] }
 lazy_static = "1"
 raw-window-handle = "0.3"
-# disable this for now, the versions of glutin are jumping too fast
+# disable this for now, the versions of glutin are jumping too fast.
+# Also, surfman is the new default backend for Linux.
 #glutin = { version = "0.24", optional = true }
 
-#[target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
-# TODO: Update to released version when it comes out
-#surfman = { git = "https://github.com/servo/surfman.git", rev = "41ac1ee", features = ["sm-raw-window-handle"], optional = true }
+[target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
+surfman = { version = "0.3", features = ["sm-raw-window-handle"], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3.6"
@@ -64,9 +62,6 @@ features = [
     "WebGlTexture",
     "Window",
 ]
-
-[target.'cfg(windows)'.dependencies]
-winapi = { version = "0.3", features = ["libloaderapi", "windef", "wingdi", "winuser"], optional = true }
 
 [build-dependencies]
 gl_generator = "0.11"
--- a/src/backend/gl/src/lib.rs
+++ b/src/backend/gl/src/lib.rs
@@ -78,9 +78,11 @@ pub(crate) struct GlContainer {
 impl GlContainer {
     #[cfg(not(wasm))]
     fn make_current(&self) {
-        // TODO:
-        // NOTE: Beware, calling this on dereference breaks the surfman backend
-        // I'm not sure if there would be similar concequences with other backends.
+        #[cfg(surfman)]
+        self.surfman_device
+            .write()
+            .make_context_current(&self.surfman_context.read())
+            .expect("TODO");
     }
 
     #[cfg(any(glutin, wgl))]
--- a/src/backend/gl/src/window/surfman.rs
+++ b/src/backend/gl/src/window/surfman.rs
@@ -29,12 +29,9 @@ thread_local! {
 }
 
 pub struct Instance {
-    hardware_adapter: sm::Adapter,
-    // TODO: We're not using these yet, but leave them here for later
-    #[allow(dead_code)]
-    low_power_adapter: sm::Adapter,
-    #[allow(dead_code)]
-    software_adapter: sm::Adapter,
+    // The root context with which all other context share resources
+    root_context: Starc<RwLock<sm::Context>>,
+    device: Starc<RwLock<sm::Device>>,
 }
 
 impl fmt::Debug for Instance {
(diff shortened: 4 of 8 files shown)


zicklag:
@kvark I've got a quick question about EGL, which is used by `surfman` for X11. I'm using a container for my dev environment because I'm not root on my build machine, and the surfman examples run fine *inside* the container. Outside of the container, though, the examples bomb out when creating the surfman X11 connection with an error: `egl function was not loaded`.

Do you know what needs to be installed to get EGL to load?

kvark:
Probably `libegl1-mesa` package?

zicklag:
Strangely, I have that and it still won't work anyway. Oh, well, thanks for the hint. I'm not going to worry too much about it . I just needed it for the debugger to work, but I'll be fine without it for now.

---

OK, so I've gotten your changes to surfman pulled locally and I've gotten all setup to work on the `gfx-backend-gl`. I just wanted to make sure that this sounds right on how we organize the sharing between contexts because I can't remember exactly how we were going to put it together:

- When you create the `Instance` ( currently actually when we enumerate adapters )  we will want to create essentially a "root" context.
- The root context is the one that we create [here](https://github.com/katharostech/gfx/blob/d282b15d99361e2df198f9c3932926540c8a7a83/src/backend/gl/src/window/surfman.rs#L146). That is the context that we use to create the `gl` container with the `from_fn_proc` function.
- When surfaces are created for that `Instance`, each surface will create its own context that will have the "root" context in the `share_with` of its `ContextAttributes`.

Is that along the right lines?

kvark:
> OK, so I've gotten your changes to surfman pulled locally

My changes were merged into surfman proper in https://github.com/servo/surfman/pull/185, you shouldn't need my branch, just use `master`.

> Is that along the right lines?

Correct!

Also, the adapters may need to create their own GL contexts, depending on exposed/requested features. If you expose only a single adapter, you can just re-use the context from the instance, that's simpler.

zicklag:
Haha! Finally got this working I think!

image

I haven't switched to using the official `servo/surfman` yet. I'm using a fork of @kvark's old fork 'cause I got stuck without internet for a few days and it was what I had on my machine. 😉

I also started work on porting WGPU to the latest GFX and including support for OpenGL and it is going very well so far. Hope to have this finished soon!

src/backend/gl/src/window/surfman.rs
```diff
@@ -58,12 +57,15 @@ impl Instance {
...
-        let context_attributes = Self::get_default_context_attributes();
+        let mut context_attributes = Self::get_default_context_attributes();
+
+        // Share the context with the root context
+        let root_ctx = &*self.root_context.read();
+        context_attributes.share_with = Some(root_ctx);
 
         // Open a device for the surface
-        // TODO: Assume hardware adapter
         let mut device = SM_CONN
```
kvark, line 67: why are we creating the device here? I thought we'd be using the one we have on the instance already

src/backend/gl/src/window/surfman.rs
```diff
@@ -44,12 +42,13 @@ impl fmt::Debug for Instance {
...
 
 impl Instance {
-    fn get_default_context_attributes() -> sm::ContextAttributes {
+    fn get_default_context_attributes() -> sm::ContextAttributes<'static> {
         sm::ContextAttributes {
             version: sm::GLVersion::new(3, 3), // TODO: Figure out how to determine GL version
             // TODO: Skipping COMPATIBILITY_PROFILE for now, because it panics with a TODO, but
             // that is probably something we want to provide later.
             flags: sm::ContextAttributeFlags::ALPHA,
+            share_with: None,
```
kvark, line 51: please switch to the upstream surfman, it has a better API from what I made in my fork initially

src/backend/gl/src/window/surfman.rs
zicklag: Ah, we probably don't need to. I'll change that.

src/backend/gl/src/window/surfman.rs
zicklag: :+1: Sounds good.

src/backend/gl/src/window/surfman.rs
```diff
@@ -29,12 +29,10 @@ thread_local! {
...
 
 pub struct Instance {
-    hardware_adapter: sm::Adapter,
-    // TODO: We're not using these yet, but leave them here for later
-    #[allow(dead_code)]
-    low_power_adapter: sm::Adapter,
-    #[allow(dead_code)]
-    software_adapter: sm::Adapter,
+    // The root context with which all other context share resources
+    root_context: Starc<RwLock<sm::Context>>,
```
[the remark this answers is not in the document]
zicklag: I'll try that out real quick. That would be *really* nice.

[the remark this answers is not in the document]
zicklag: I tried it real quick and got errors because `Device` and `Context` are `!Sync + !Send`.

zicklag:
OK, I think this is good now.

src/backend/gl/Cargo.toml
```diff
@@ -17,10 +17,9 @@ build = "build.rs"
 name = "gfx_backend_gl"
 
 [features]
-#default = ["wgl", "surfman", "surfman-x11"]
-default = []
+default = ["wgl", "surfman", "surfman-x11"]
 wgl = ["winapi"]
-#surfman-x11 = ["surfman/sm-x11"]
+surfman-x11 = ["surfman/sm-x11"]
```
kvark, line 22: We are trying to make default deps as uncontroversial as possible. X11, for example, is not enabled by default in the Vulkan backend - https://github.com/gfx-rs/gfx/blob/379ace36c8ae78c1f2a7e4c5af4e9715b204c1ea/src/backend/vulkan/Cargo.toml#L41

I believe it shouldn't be default here either.
Also, we should call this feature just `x11`, assuming that `surfman` is the main path.

src/backend/gl/Cargo.toml
```diff
@@ -34,12 +33,13 @@ parking_lot = "0.11"
...
-# disable this for now, the versions of glutin are jumping too fast
+# disable this for now, the versions of glutin are jumping too fast.
+# Also, surfman is the new default backend for Linux.
 #glutin = { version = "0.24", optional = true }
 
-#[target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
+[target.'cfg(all(unix, not(target_os = "ios")))'.dependencies]
 # TODO: Update to released version when it comes out
-#surfman = { git = "https://github.com/servo/surfman.git", rev = "41ac1ee", features = ["sm-raw-window-handle"], optional = true }
+surfman = { version = "0.3.0", features = ["sm-raw-window-handle"], optional = true }
```
kvark, line 42: super nit: just "0.3"

src/backend/gl/src/lib.rs
```diff
@@ -78,9 +78,13 @@ pub(crate) struct GlContainer {
 impl GlContainer {
     #[cfg(not(wasm))]
     fn make_current(&self) {
-        // TODO:
-        // NOTE: Beware, calling this on dereference breaks the surfman backend
-        // I'm not sure if there would be similar concequences with other backends.
+        #[cfg(surfman)]
+        {
```
kvark, line 82: nit: are the curvy braces needed here?

src/backend/gl/src/window/surfman.rs
```diff
@@ -95,75 +94,69 @@ impl Instance {
...
+            .make_context_current(&context)
+            .expect("TODO");
 
         // Create a surface with the given context
         Surface {
             renderbuffer: None,
             swapchain: None,
             context: Starc::new(RwLock::new(context)),
-            device: Starc::new(RwLock::new(device)),
+            device: self.device.clone(),
```
kvark, line 106: Does the surface really need to get a hold of the device?

kvark:
Seems to be working on macOS. Although, we'd need to test the wgpu-rs samples before publishing the gfx backend.

src/backend/gl/src/window/surfman.rs
zicklag: Yes because the drop implementation needs a device handle in order to clean up the surface and the context.

zicklag:
> Seems to be working on macOS.

Oh, sweet.

> Although, we'd need to test the wgpu-rs samples before publishing the gfx backend.

Yeah, sounds like a good idea. That was the issue last time: we didn't find the issues with the backend until we tried hooking it up to WGPU.

src/backend/gl/src/window/surfman.rs
kvark: We now have `Instance::destroy_surface`, where this logic needs to be moved into. The Drop shouldn't even be there.

zicklag: Ah, OK. One second and I'll fix that.

zicklag: There you go.

kvark:
Nice to have a near-instant CI now, isn't it? :)

examples/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2018"
 default = []
 metal = ["gfx-backend-metal"]
 gl = ["gfx-backend-gl"]
```
kvark, line 11: Examples don't need to try to cut out the dependencies. Example project can just always enable x11 like it does for vulkan, no need to show this out in the interface.

zicklag:
Uh, yeah, that's pretty nice. :)

examples/Cargo.toml
zicklag: That makes sense. Done!

zicklag:
If I've estimated correctly I should be able to test this with WGPU very soon. I'll have a PR on WGPU and we'll figure out how well this works with the WGPU examples.

kvark:
Please rebase and squash

zicklag:
Done. And, hey, we deleted more lines of code than we added. :tada:

zicklag:
Ah, oops, I made the GL backend only included in the examples for `cfg(unix)` platforms and now CI is failing for windows. Do we want to completely disable OpenGL for windows, or do we leave it? I don't think I broke GL on windows with the surfman stuff, but I never tested it before or after so I don't know if it's working.

kvark:
Let's disable it on Windows completely

kvark:
that also means we don't need to enable WGL dependency

zicklag:
Yay! :tada: Thanks so much for your help @kvark. Couldn't have done it without your guidance. :smile:

Next up, WGPU!

PR #3342, opened 2020-08-19
[0.6/dx12] refactor pipeline layouts and descriptor binding

kvark:
Fixes the current breakage of the wgpu-rs shadow example on dx12.

The problem was the last fix being incomplete. We switched from parameter iteration into root offset iteration in `flush*`.
But we didn't advance the root offset on bindings that are non-dirty (as well as invalid bindings).

This PR moves the common parts about pipelines and layouts into a separate structure. It also explicitly stores a root offset per parameter, as well as the total number of root slots. That allows us to make the `flush*` function robust, and also moves it to be a method of `PipelineCache`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-dx12-0.6.1 (18-08-2020)
+  - fix descriptor binding
+
 ### backend-vulkan-0.6.1 (17-08-2020)
   - fix Android build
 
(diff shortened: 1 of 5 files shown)


kvark:
Thank you for taking a look!

Bors r=cwfitzgerald

kvark:
We are losing travis... :(

kvark:
We are losing travis... :(
bors retry

kvark:
Merged in #3343

PR #3343, opened 2020-08-19
[Dx12] object labels and multisampling

kvark:
based on #3342, which is already published, but stuck on CI
related to https://github.com/gfx-rs/wgpu-rs/issues/47
tested on wgpu-rs examples


msiglreith (approved):
lgtm
bors r+

kvark:
RIP travis.org
Going to switch us to GHA now

PR #3356, opened 2020-08-31
Update spirv_cross to 0.21

grovesNL:
spirv_cross 0.21 exposes `force_zero_initialized_variables` for GLSL/HLSL/MSL and `force_native_arrays` for MSL. I think we can probably enable both of these options unconditionally in gfx and wgpu.

(I'll backport this into 0.6 once we accept these changes)

PR checklist:
- [ ] `make` succeeds (on *nix)
- [ ] `make reftests` succeeds
- [X] tested examples with the following backends: metal, gl

PR #3358, opened 2020-09-02
[0.6] Backport spirv_cross update

grovesNL:
Backport #3356 to update spirv_cross:
- spirv_cross 0.21 enables us to use zero initialization for shader variables and use of native arrays in MSL
- increase crate versions for the dx12/dx11/metal backends and auxil

### DIFF
--- a/src/auxil/auxil/Cargo.toml
+++ b/src/auxil/auxil/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-auxil"
-version = "0.5.0"
+version = "0.6.0"
 description = "Implementation details shared between gfx-rs backends"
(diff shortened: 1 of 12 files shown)


src/auxil/auxil/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-auxil"
-version = "0.5.0"
+version = "0.5.1"
```
kvark, line 3: this should probably be `0.6.0`, since spirv-cross is the public dependency

grovesNL: Good point – we'll need this change on `master` too then

grovesNL:
bors r=kvark

PR #3360, opened 2020-09-02
[dx12] fix root signature indexing

kvark:
Fixes a critical problem reported in https://github.com/gfx-rs/wgpu/issues/849#issuecomment-685151553
It appears to be a regression from #3342, where I accidentally started using the root offsets where the root indices are expected. We might need to change their types to not hit this any more times (maybe in d3d12-rs even? cc @msiglreith ).
PR checklist:
- [x] tested examples with the following backends: D3D12

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -295,13 +295,13 @@ impl PipelineCache {
             None => return,
         };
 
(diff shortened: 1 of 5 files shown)


kvark:
Given that it's just an accidental mistake, I'll just go ahead and merge this.
bors r+

PR #3362, opened 2020-09-02
[0.6/dx11] dx11 internal copy refactor

kvark:
Makes the internal code better scale-able (shifting from fields to maps).
Fixes the race condition on the internal constant buffer (yikes!)
Enables uploading R8,RG8, and RGBA8 texture copies from buffers.
PR checklist:

Note: this is missing a commit that bumps the patch version and adds a changelog entry


msiglreith (approved):
looks fine to me 👍 
bors r+

PR #3363, opened 2020-09-02
[0.6/dx11] fix read-only buffer SRV

kvark:
Follow-up to #3362
Also fixes the read-only buffer SRVs.
Gets DX11 the the point of running vange-rs :tada:


kvark:
bors r=followup

PR #3366, opened 2020-09-04
[0.6/dx11] fix cpu-visible memory

kvark:
Fixes running anything in release.

Tested on vange-rs 🎉 

One problem was that every memory was allocated as if it was cpu-visible, even if it wasn't supposed to be. And we used memory that was uninitialized...

Another, we had a few occasions of this evil pattern:
```rust
let some_ptr = if let Some(something) { &something } else {ptr::null()};
```
fun thing happens here because the `something` is dead right after that expression, and the pointer gets used still...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+### backend-dx11-0.6.3 (04-09-2020)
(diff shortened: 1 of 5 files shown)


msiglreith (approved):
bors r+

PR #3368, opened 2020-09-07
[0.6/dx11] fix memory flushes and presentation

kvark:
The flush ranges were computed incorrectly.
Tested on vange-rs

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -32,7 +32,9 @@ use range_alloc::RangeAllocator;
 
 use winapi::{
     shared::{
-        dxgi::{IDXGIAdapter, IDXGIFactory, IDXGISwapChain},
+        dxgi::{
+            IDXGIAdapter, IDXGIFactory, IDXGISwapChain, DXGI_PRESENT_ALLOW_TEARING,
+        },
         dxgiformat,
         minwindef::{FALSE, HMODULE, UINT},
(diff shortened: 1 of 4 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -978,12 +986,17 @@ impl queue::CommandQueue<Backend> for CommandQueue {
...
+        let mut presentation = surface.presentation.as_mut().unwrap();
+        let (interval, flags) = match presentation.mode {
+            window::PresentMode::IMMEDIATE => (0, DXGI_PRESENT_ALLOW_TEARING),
+            //Note: this ends up not presenting anything for some reason
+            //window::PresentMode::MAILBOX if !presentation.is_init => (1, DXGI_PRESENT_DO_NOT_SEQUENCE),
+            window::PresentMode::MAILBOX |
+            window::PresentMode::FIFO => (1, 0),
+            _ => (0, 0),
+        };
+        presentation.is_init = false;
```
msiglreith, line 998: this is only for the sequencing part?

msiglreith (approved):

src/backend/dx11/src/lib.rs
kvark: this is related to the `Note` above, for some reason I wasn't able to make it work
with the line commented out, the `is_init` isn't used, but it's fine

PR #3372, opened 2020-09-08
Document command::Level variants

hgallagher1993:
I just took each of the explanations directly from https://vulkan-tutorial.com/Drawing_a_triangle/Drawing/Command_buffers#page_Command-buffer-allocation


kvark (approved):
bors r+

PR #3373, opened 2020-09-15
[mtl] Enable debug info for internal shaders

kvark:
Closes #2398
My understanding is that these options do not reduce the optimization level, so there is no performance concern in just shipping the MTL libraries with debug info.


grovesNL (approved):
Thanks, sounds good to me!
bors r+

Issue #3353, opened 2020-08-27
WGPU GL Rendering Artifacts


zicklag:
I'm opening this issue to continue discusion of the rendering artifacts in the WIP gl backend for WGPU: https://github.com/gfx-rs/wgpu/issues/450#issuecomment-681312225.

---

> @kvark :
> So we should experiment. For example, try to bind the swapchain texture as glBindTexture on the surface context before blitting.

Got it, but, what texture am I supposed to bind? It doesn't seem that there are any textures associated to the surface. Maybe that's the problem? The only thing associated to the context seems to be the [framebuffers](https://github.com/gfx-rs/gfx/blob/master/src/backend/gl/src/window/surfman.rs#L16-L23):

```rust
pub struct Swapchain {
    // Underlying window, required for presentation
    pub(crate) context: Starc<RwLock<sm::Context>>,
    // Extent because the window lies
    pub(crate) extent: window::Extent2D,
    ///
    pub(crate) fbos: ArrayVec<[native::RawFrameBuffer; 3]>,
}
```

kvark:
One way, is to get the texture object (temporarily) via https://docs.rs/surfman/0.3.0/surfman/device/trait.Device.html#tymethod.surface_texture_object
Another way to try is, instead of doing `glBindTexture`, do `glBindFramebuffer(GL_READ_FRAMEBUFFER, ...)` on that framebuffer, which should tell the driver to do the same thing.

zicklag:
@kvark I haven't tried any workarounds yet, I'm trying that next, but does it mean anything that the triangle demo works fine?

![image](https://user-images.githubusercontent.com/25393315/91626092-b118f300-e971-11ea-89a0-9caa2c3ad2b4.png)

Does that give us any pointers?

zicklag:
OK, and another oddity. All I did was change the clear color in the cube demo from this:

```rust
ops: wgpu::Operations {
                        load: wgpu::LoadOp::Clear(wgpu::Color {
                            r: 0.1,
                            g: 0.2,
                            b: 0.3,
                            a: 1.0,
                        }),
                        store: true,
                    },
```

To this:

```rust
ops: wgpu::Operations {
                        load: wgpu::LoadOp::Clear(wgpu::Color {
                            r: 0.,
                            g: 0.,
                            b: 1.,
                            a: 1.0,
                        }),
                        store: true,
                    },
```

And now it works:

[...]

PR #3375, opened 2020-09-17
Enable GL backend via Surfman

zicklag:
( Will,  hopefully ) fix issue #3353 

PR checklist:

### DIFF
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -27,7 +27,7 @@ log = { version = "0.4" }
 gfx-hal = { path = "../../hal", version = "0.6" }
 auxil = { path = "../../auxil/auxil", version = "0.5", package = "gfx-auxil", features = ["spirv_cross"] }
 smallvec = "1.0"
-glow = "0.4"
+glow = "0.6.1"
 parking_lot = "0.11"
 spirv_cross = { version = "0.22", features = ["glsl"] }
 lazy_static = "1"
--- a/src/backend/gl/src/queue.rs
+++ b/src/backend/gl/src/queue.rs
@@ -2,11 +2,12 @@ use std::borrow::Borrow;
 use std::{mem, slice};
 
 use glow::HasContext;
+use parking_lot::Mutex;
 use smallvec::SmallVec;
 
 use crate::{
-    command as com, device, info::LegacyFeatures, native, state, Backend, GlContext, Share, Starc,
-    Surface, Swapchain,
+    command as com, device, info::LegacyFeatures, native, state, Backend, Error, GlContext, Share,
+    Starc, Surface, Swapchain,
 };
 
 // State caching system for command queue.
@@ -62,6 +63,7 @@ pub struct CommandQueue {
     features: hal::Features,
     vao: Option<native::VertexArray>,
     state: State,
+    presentation_fence: Starc<<glow::Context as glow::HasContext>::Fence>,
 }
 
 impl CommandQueue {
@@ -76,6 +78,7 @@ impl CommandQueue {
             features,
             vao,
             state: State::new(),
+            presentation_fence: Starc::new(std::ptr::null_mut()),
         }
     }
 
@@ -203,9 +206,14 @@ impl CommandQueue {
         &data[ptr.offset as usize..(ptr.offset + ptr.size) as usize]
     }
 
-    fn present_by_copy(&self, swapchain: &Swapchain, index: hal::window::SwapImageIndex) {
+    fn present_by_copy(&self, swapchain: &Swapchain, _index: hal::window::SwapImageIndex) {
         let gl = &self.share.context;
         let extent = swapchain.extent;
+
+        // Wait for rendering to finish
+        unsafe {
+            gl.wait_sync(*self.presentation_fence, 0, glow::TIMEOUT_IGNORED);
+        }
 
         #[cfg(wgl)]
         swapchain.make_current();
@@ -218,7 +226,7 @@ impl CommandQueue {
 
         // Use the framebuffer from the surfman context
         #[cfg(surfman)]
-        let fbo = gl
+        let draw_fbo = gl
             .surfman_device
             .read()
             .context_surface_info(&swapchain.context.read())
@@ -227,12 +235,20 @@ impl CommandQueue {
             .framebuffer_object;
 
         unsafe {
+            let tmp_read_fbo = gl.context.create_framebuffer().expect("TODO");
             gl.context
-                .bind_framebuffer(glow::READ_FRAMEBUFFER, Some(swapchain.fbos[index as usize]));
+                .bind_framebuffer(glow::READ_FRAMEBUFFER, Some(tmp_read_fbo));
+            gl.context.framebuffer_renderbuffer(
+                glow::READ_FRAMEBUFFER,
+                glow::COLOR_ATTACHMENT0,
+                glow::RENDERBUFFER,
+                Some(swapchain.renderbuffer),
+            );
+
             gl.context.bind_framebuffer(
                 glow::DRAW_FRAMEBUFFER,
                 #[cfg(surfman)]
-                match fbo {
+                match draw_fbo {
                     0 => None,
                     other => Some(other),
                 },
@@ -251,6 +267,8 @@ impl CommandQueue {
                 glow::COLOR_BUFFER_BIT,
                 glow::LINEAR,
             );
+
+            gl.context.delete_framebuffer(tmp_read_fbo);
         }
 
         // Present the surfman surface
@@ -703,7 +721,7 @@ impl CommandQueue {
                 match texture_target {
                     glow::TEXTURE_2D => {
                         gl.bind_texture(glow::TEXTURE_2D, Some(dst_texture));
-                        gl.tex_sub_image_2d_pixel_buffer_offset(
+                        gl.tex_sub_image_2d(
                             glow::TEXTURE_2D,
                             data.image_layers.level as _,
                             data.image_offset.x,
@@ -712,12 +730,12 @@ impl CommandQueue {
                             data.image_extent.height as _,
                             texture_format,
                             pixel_type,
-                            data.buffer_offset as i32,
+                            glow::PixelUnpackData::BufferOffset(data.buffer_offset as u32),
                         );
                     }
                     glow::TEXTURE_2D_ARRAY => {
                         gl.bind_texture(glow::TEXTURE_2D_ARRAY, Some(dst_texture));
-                        gl.tex_sub_image_3d_pixel_buffer_offset(
+                        gl.tex_sub_image_3d(
                             glow::TEXTURE_2D_ARRAY,
                             data.image_layers.level as _,
                             data.image_offset.x,
@@ -729,7 +747,7 @@ impl CommandQueue {
                                 - data.image_layers.layers.start as i32,
                             texture_format,
                             pixel_type,
-                            data.buffer_offset as i32,
+                            glow::PixelUnpackData::BufferOffset(data.buffer_offset as u32),
                         );
                     }
                     _ => unimplemented!(),
@@ -756,7 +774,7 @@ impl CommandQueue {
                 gl.active_texture(glow::TEXTURE0);
                 gl.bind_buffer(glow::PIXEL_PACK_BUFFER, Some(dst_buffer));
                 gl.bind_texture(glow::TEXTURE_2D, Some(src_texture));
-                gl.get_tex_image_pixel_buffer_offset(
+                gl.get_tex_image(
                     glow::TEXTURE_2D,
                     data.image_layers.level as _,
                     //data.image_offset.x,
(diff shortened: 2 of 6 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -8,6 +8,7 @@ use crate::{
     command as com, device, info::LegacyFeatures, native, state, Backend, GlContext, Share, Starc,
     Surface, Swapchain,
 };
+use parking_lot::Mutex;
```
kvark, line 11: super nit: the use statements are currently grouped as: standard library, then external dependencies, then local dependencies. So `parking_lot` should go near `glow` and `smallvec` ones.

src/backend/gl/src/queue.rs
```diff
@@ -227,11 +233,19 @@ impl CommandQueue {
             .framebuffer_object;
 
         unsafe {
+            let tmp_read_fbo = gl.context.create_framebuffer().expect("TODO");
```
kvark, line 236: Problem is that this code doesn't really use `tmp_read_fbo` anywhere. All it does is creating it and then destroying it. Perhaps, you forgot to fix the next line with `bind_framebuffer`?

src/backend/gl/src/queue.rs
```diff
@@ -1195,6 +1211,14 @@ impl hal::queue::CommandQueue<Backend> for CommandQueue {
             }
         }
 
+        // Create a fence used to synchronize the finish of the rendering and
+        // the blit used to present on the surface.
+        self.presentation_fence = self
```
kvark, line 1216: We are creating the fence but never waiting on it. In present() there should be a call to `wait_sync` on that fence.

kvark (changes requested):
Thank you for trying these ideas out!
Fortunately, I found a few places that would explain why it doesn't work yet :)

src/backend/gl/src/queue.rs
zicklag: Ah yes! Whoops.

src/backend/gl/src/queue.rs
zicklag: Ah yes, I like doing that with my use statements too.

src/backend/gl/src/queue.rs
zicklag: Ah, right again. I think I took that out and forgot to put it back in.

zicklag (commented):
Sweet. I'll come back and fix those probably later tonight. Thanks!

zicklag:
I fixed the issues you pointed out and I'm still getting the jagged edges. :confused: 

image

Any other ideas. :slightly_smiling_face:

src/backend/gl/src/queue.rs
```diff
@@ -207,6 +213,22 @@ impl CommandQueue {
         let gl = &self.share.context;
         let extent = swapchain.extent;
 
+        // Wait for rendering to finish
+        unsafe {
+            let wait_result = gl.client_wait_sync(
```
kvark, line 218: this actually needs to be `wait_sync`, not `client_wait_sync`

zicklag: Interesting, glow doesn't actually have a `wait_sync`. :astonished: 

https://docs.rs/glow/0.6.0/glow/?search=wait_sync

Should I open an issue for glow on that? Is it maybe because it isn't supported on GL ES or something? Maybe that is just a simple PR to glow I could make?

kvark: Yes, I recommend implementing it quickly in a fork, making a PR (it's a non-breaking change), and using your fork for the dependency here in the meantime.

zicklag: OK, updated glow with the extra binding and submitted a PR and updated this branch with the changes, but still no change in rendering.

kvark:
What if, after we render to the renderbuffer on the main context (in submit), we bind it for reading to a framebuffer, and actually issue a blit to some other FBO? Just to replicate the same scenario that we have on surface context (in present). This would ensure that the image goes through the necessary memory barriers and layout changes under the hood. Of course we don't want to ship that, but it would confirm the theory that there is a workaround, at least.

zicklag:
> What if, after we render to the renderbuffer on the main context (in submit), we bind it for reading to a framebuffer, and actually issue a blit to some other FBO? 

I'll try that next, but I just found something interesting. I just made a new commit that gets rid of context sharing completely and I'm still getting the same rendering artifacts.

Also, I'm trying to figure out why we need separate contexts for the surface and the root context. Is that because in multi-threading you cannot use the same context across different threads? Being that we can't share anything across extra threads right now anyway because of `Starc` and such maybe we shouldn't even be sharing contexts at all?

Not that it is helping in this case it seems.

kvark:
>I just made a new commit that gets rid of context sharing completely and I'm still getting the same rendering artifacts.

That's an interesting experiment! We should double-check if the sharing is actually requested at the low level (by debugging through context creation).

>I'm not actually sure how to try what you suggested in submit because the framebuffer I would need to bind as the read FBO would be the swapchain FBO which we don't have access to in the submit function. Does the test I already did by getting rid of context sharing cover that test?

Let's do it in present() then. To clarify: this operation is purely artificial and makes no practical sense, it's just to convince the driver to transition the renderbuffer into "readable" state. So, in present() before switching the current context, you could create an FBO and attach the renderbuffer to it, bind the FBO for reading. Then maybe create another temporary RBO, another temporary FBO, attach them, and bind from drawing. Then do a blit.

After that, we can proceed as usual by switch the contexts and doing a *real* blit.

kvark:
@zicklag have you tried to make further progress here?

zicklag:
Haven't had the chance to try anything else yet.

kvark:
What platform are you on? Can you reproduce the rendering issue on the quad example?

kvark:
I tested this PR without the last commit on both macOS and Linux with a modified quad example and got no issues...

zicklag:
I'm on Linux, but the quad example did work before this PR: only the WGPU Cube example actually exhibits the problem so far.

kvark:
I think the difference with the `cube` example is that the latter has a second pass, where it draws edges only, and enables blending. So this is what confusing the driver in our case, and it would be trivial to replicate on gfx's quad example.

> Especially with Bevy getting an OpenGL renderer backend which would have been one of my larger use-cases

I'd expect them to happily throw it out if gfx-rs had a working GL backend.
There is also Veloren, which is essentially blocked for wgpu migration on our GL backend.

zicklag:
> I think the difference with the cube example is that the latter has a second pass, where it draws edges only, and enables blending. So this is what confusing the driver in our case, and it would be trivial to replicate on gfx's quad example.

Hmm, that would be good to do.

> I'd expect them to happily throw it out if gfx-rs had a working GL backend.
> There is also Veloren, which is essentially blocked for wgpu migration on our GL backend.

Ahgh, I know. You're going to convince me to keep trying this when I have time. LOL :smiley:

---

OK, well I'll try to get back to this eventually, then. Right now all my time is going to getting out an 0.2 release of my [Arsenal](https://github.com/katharostech/arsenal) game engine ( which is built on Bevy ). GL support is definitely something we really want for Arsenal so this will probably take priority at some point, but the first goal is for me to get Arsenal working at all.

kvark:
Sorry, didn't mean to try convincing you. Looks like you got enough on your plate at the moment :) Just wanted to argue/elaborate on why GL backend would still be a strategic goal for us, even if Bevy doesn't use it.

zicklag:
Yeah, no problem man. :D I really do want to get this figured out. We'll figure out it sometime. :+1: It would be _amazing_ to support all of the graphics APIs through one interface.

kvark:
Bummer. I added wireframe rendering to the quad example, and it still doesn't exhibit the issue.

kvark:
Do you see artifacts in the quad example with https://github.com/kvark/gfx/tree/gl-test ?
I tried running wgpu's examples on GL with the fixes of your PRs (except for the last commit in this PR, which is temporary), and it's all fine on Linux. Going to do the same on macOS now...
Would be good to resurrect and land this, at least on master. Are you interested to do this? Otherwise, I can pick it up.

zicklag:
> Do you see artifacts in the quad example with https://github.com/kvark/gfx/tree/gl-test ?

Just ran it and it looks fine.

> I tried running wgpu's examples on GL with the fixes of your PRs (except for the last commit in this PR, which is temporary), and it's all fine on Linux. Going to do the same on macOS now...

Oh, interesting. I'll try it again.

> Would be good to resurrect and land this, at least on master. Are you interested to do this? Otherwise, I can pick it up.

I should be able to get it rebased. I'll see if I can do that now, I've got ~20 mins. It shouldn't be too hard.

src/backend/gl/Cargo.toml
```diff
@@ -27,7 +27,7 @@ log = { version = "0.4" }
 gfx-hal = { path = "../../hal", version = "0.6" }
 auxil = { path = "../../auxil/auxil", version = "0.5", package = "gfx-auxil", features = ["spirv_cross"] }
 smallvec = "1.0"
-glow = "0.4"
+glow = { git = "https://github.com/katharostech/glow.git", branch = "gl-wait-sync" }
```
kvark, line 30: should just use "0.6.1"

src/backend/gl/src/queue.rs
```diff
@@ -62,8 +63,12 @@ pub struct CommandQueue {
     features: hal::Features,
     vao: Option<native::VertexArray>,
     state: State,
+    presentation_fence: <glow::Context as glow::HasContext>::Fence,
```
kvark, line 66: let's use `Starc` as we do with other native stuff here, for consistency, and to avoid `unsafe impl` that follow

src/backend/gl/src/queue.rs
kvark: looks like this still needs to be done?

kvark:
Unable to test the wgpu's cube example on macOS, getting
> thread 'main' panicked at 'Error InvalidOperation executing command: CopyBufferToBuffer(2, 1, BufferCopy { src: 0, dst: 0, size: 580 })', /Users/dmalyshau/Code/gfx/src/backend/gl/src/queue.rs:1184:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

This is unrelated to your PR but unfortunate. Will look into it, not a blocker.

src/backend/dx12/src/command.rs
```diff
@@ -389,63 +389,66 @@ pub struct CommandBuffer {
     pool_shared: Arc<PoolShared>,
     begin_flags: com::CommandBufferFlags,
 
-    // Cache renderpasses for graphics operations
+    /// Cache renderpasses for graphics operations
```
kvark: looks like something went totally wrong with the rebase? These changes aren't supposed to be here

zicklag: Fixed. :+1:

kvark (approved):
Alright, looks like everything is in place!
bors r+

zicklag:
> looks like this still needs to be done?

Sorry, it won't let me reply to our comment inline. What still needs to be done?

kvark:
> > looks like this still needs to be done?
> 
> Sorry, it won't let me reply to our comment inline. What still needs to be done?

It was using `client_wait_sync` instead of `wait_sync`, but it looks like you fixed this now :+1:

kvark:
Could you make a similar PR towards hal-0.6 branch?
I'd be comfortable publishing GL backend with this. Remember that we didn't publish it since hal-0.6 got out. Even if it's partially broken, still more useful than no backend at all.

zicklag:
> It was using client_wait_sync instead of wait_sync, but it looks like you fixed this now +1

Ah, cool.

> Alright, looks like everything is in place!

Sweet! :smiley:

I haven't tested the WGPU examples yet, but I'll try those out in a sec ( if there's no major conflicts to resolve ). If those don't work for me, it might just be that my driver's goofy. Also it's running the Mesa driver, but I've got a hybrid GPU setup on my machine, so I can choose to run it with the GPU which I didn't realize earlier so I'm not sure if that is the reason for the weirdness. I'll have to try both.

> Could you make a similar PR towards hal-0.6 branch?
> I'd be comfortable publishing GL backend with this. Remember that we didn't publish it since hal-0.6 got out. Even if it's partially  > broken, still more useful than no backend at all.

Sounds good. I'll do that. :+1:

zicklag:
Oh, yeah, I forgot about that extra framebuffer was only added for `surfman`. I'll have to gate the code off for the windows version and do that differently I guess.

kvark:
But we don't need Surfman (and GL backend) on windows at all, since it's not able to support WGL in the long run anyway

zicklag:
Ah, then no reason to pollute the code with extra `#[cfg]` checks. What do we want to do here? Just disable CI for GL on windows?

kvark:
Yes, rip out GL on windows completely :)

zicklag:
There that should make CI pass now.

kvark:
I was worried that the GL error I'm seeing on macOS is really about something broken with the presentation, and just manifesting itself on the next command. But it turns out to be a genuine error of buffer copying. Not sure what's going on there, but we can proceed.
bors r+

zicklag:
Just about out of time now, I'll try to do the hal-0.6 PR and a test of the WGPU examples tomorrow! :slightly_smiling_face:

kvark:
Thank you! I'll do some more cleanup of the GL backend in master in the meantime.

PR #3376, opened 2020-09-17
[dx12] fix resource tier-1 handling of render targets

kvark:
Fixes a case where we request `TRANSFER_DST` without `COLOR_ATTACHMENT` on an image that has a non-renderable format, when running on Resource Tier-1 hardware.


msiglreith (approved):
bors r+

PR #3387, opened 2020-09-27
Metal optimizations and fixes

### LINKED ISSUE #3381: Avoid redundant depth-stencil states in Metal
> file:///private/var/folders/_s/63j0dt2d54l9vgyvl3v0n26r0000gn/T/com.apple.dt.XcodeBU5JKT/dota2.gputrace/: runtime: API: Encoder 'Render 0 0x60000b3203c0' has 15 redundant DepthStencilState bindings

The code appears to bind it unconditionally for the most part, without checking for redundancy.

There is also an issue with scissor-rect, could more to a separate issue.

kvark:
Fixes a bunch of things, except performance on Dota :(
Fixes #3381

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -10,8 +10,7 @@ use hal::{
     buffer, command as com,
     device::OutOfMemory,
     format::{Aspects, FormatDesc},
-    image::{Extent, Filter, Layout, Level, NumSamples, SubresourceRange},
-    memory,
+    image as i, memory,
     pass::AttachmentLoadOp,
     pso, query,
     window::{PresentError, Suboptimal},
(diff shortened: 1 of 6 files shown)


src/backend/metal/src/command.rs
```diff
@@ -1218,6 +1219,7 @@ impl CommandSink {
     }
 
     /// Switch the active encoder to render by starting a render pass.
+    /// Second returned value is `true` if the switch has just happened.
```
grovesNL, line 1250: Looks like a leftover comment? I don't see a second return value here

grovesNL (approved):
Looks great, thank you!

I really like the clear split between hal vs. Metal state, it definitely makes the scissor changes easier to follow

src/backend/metal/src/command.rs
kvark: oh right, let me fix that!

kvark:
> I really like the clear split between hal vs. Metal state, it definitely makes the scissor changes easier to follow

I've taken a few tries to separate it at the type level by putting into different structs, really wasted a ton of time on experimenting with it. In the end, there wasn't a good path forward. Mostly, because states are generally not 100% Metal or 100% HAL. Only some states (like `resource_vs`) are clear-cut like that, but most are somewhat in the middle.

bors r=grovesNL

PR #3390, opened 2020-09-30
[vk/0.6] Properly enable VK_KHR_maintenance3 when VK_EXT_descriptor_indexing is enabled

cwfitzgerald:
Fixes https://github.com/gfx-rs/wgpu/issues/946. DI requires maintence3, so we shouldn't need to check if it exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-vulkan-0.6.3 (30-09-2020)
+  - enable VK_KHR_maintenance3 when VK_EXT_descriptor_indexing is enabled
+
 ### backend-dx12-0.6.4 backend-vulkan-0.6.2 backend-metal-0.6.3 (23-09-2020)
   - fix descriptor indexing features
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-vulkan-0.6.3 (23-09-2020)
```
kvark, line 3: let's fix the date here

src/backend/vulkan/src/lib.rs
```diff
@@ -732,9 +734,9 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
                         | Features::STORAGE_TEXTURE_DESCRIPTOR_INDEXING
                         | Features::UNSIZED_DESCRIPTOR_ARRAY
                 ) {
-                    Some(*EXT_DESCRIPTOR_INDEXING)
+                    vec![*KHR_MAINTENANCE3, *EXT_DESCRIPTOR_INDEXING]
```
kvark, line 737: what about the place where we *expose* the features for descriptor indexing? These should also check for `KHR_MAINTENANCE3`

kvark (changes requested):

src/backend/vulkan/src/lib.rs
cwfitzgerald: Discussed on matrix; unneeded

kvark (approved):

PR #3392, opened 2020-10-01
Remove unused dependencies

vickenty:
Even when built with x11 and xcb features enabled on Linux,
Vulkan backend does not use anything from x11 or xcb crates.

Fixes #issue
PR checklist:


kvark (commented):
Wow, this is pretty slick!
How about we remove the features entirely?

kvark (approved):
Thank you!
This change would make some of the clients life much nicer, e.g. wgpu-native and stuff.
bors r+

Issue #2652, opened 2019-02-25
dx12: Descriptor Pool limitations


msiglreith:
Descriptor pools map 1:1 to a slice in the global descriptor heap (1M descriptors, 2k samplers).
Overallocation would be troublesome, but Vulkan allows to have more. Even some CTS tests relay on it as well as some real world applications.
Emulating this is quite costly and I would love to avoid this. At the end this depends on what Khronos Working Group decides on.

kvark:
FTR, Khronos TSG didn't decide to change these in any way since we said that it's possible on D3D12. I thought we had a rough plan on supporting multiple pools in play, and it's more a matter of implementation complexity than resulting performance.

msiglreith:
well, yes, it's doable but requires a lot of descriptor juggling. Performance impact is hard to estimate I would say. At the end, exposing a feature to restrict stuff might give the user the best of both worlds?

kvark:
Do you have an idea of how this could sound in the spec? Saying something straightforward like "at no point should a draw call use bindings from multiple descriptor pools, unless there is one for samplers specifically" :)

msiglreith:
Exposing an upper limit of concurrently allocated descriptors types by descriptor pools coupled with error on descriptor pool allocating due to fragmentation etc.
This would allow to hide the heap as implementation detail. Maybe in far future Vulkan 3 will move towards Mantle definition of descriptor sets :P

kvark:
Interestingly, the number of samplers is already capped in Vulkan: `maxSamplerAllocationCount`

msiglreith:
right, nonetheless, ordered array of samplers as found in descriptor sets are more of an issue. Even though we can create 2k samplers easily (still less than 4k guarenteed by vulkan) and would be able to store them in gpu heaps we still need to deal with the issue of the user being able to arbitrarily combine them.

kvark:
Sampler limitations are addressed in  #3393

PR #3393, opened 2020-10-02
[dx12] Dota-related fixes and improvements

### LINKED ISSUE #1949: Dynamic buffer descriptors

### LINKED ISSUE #2140: dx12: Indivial command buffer reset

### LINKED ISSUE #2661: dx12: Concurrent cmd list recording
See https://github.com/gfx-rs/gfx/issues/2140#issuecomment-396446497

Only one cmd list can be recorded against one command allocator at the same time. When recording multiple lists we need separate command allocators.

### LINKED ISSUE #2668: dx12: command buffer lifecycle
Some parts of the vulkan cmd buffer lifecycle are not implemented correctly (e.g reset -> begin)

kvark:
Fixes #2661
Fixes #2668
Fixes #2140
Fixes the sampler part of #2652
Fixes #1949 (I think? it enables dynamic storage buffers)
PR checklist:

### DIFF
--- a/src/backend/dx12/Cargo.toml
+++ b/src/backend/dx12/Cargo.toml
@@ -26,6 +26,7 @@ range-alloc = { path = "../../auxil/range-alloc", version = "0.1.1" }
 bitflags = "1"
 native = { package = "d3d12", version = "0.3", features = ["libloading"] }
 log = "0.4"
+parking_lot = "0.11"
 smallvec = "1"
 spirv_cross = { version = "0.21", features = ["hlsl"] }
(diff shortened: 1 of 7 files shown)


msiglreith (approved):
looks reasonable to me so far

kvark:
I ran gfx-ocean just fine. Just stumbled upon a d3d12 validation error due to https://github.com/gfx-rs/gfx/issues/3401
bors r=msiglreith

PR #3398, opened 2020-10-05
[0.6/dx12] fix alpha blend factors

kvark:
Fixes https://github.com/gfx-rs/wgpu/issues/953
This is a part of Vulkan semantics that needs mapping. In D3D12 it's not legal to use the COLOR factors on alpha, but in Vulkan 
it's fine.


msiglreith (approved):
bors r+

PR #3399, opened 2020-10-05
[dx12] map color blend factors to alpha

kvark:
Sibling of  #3398 for master


kvark:
bors r=msiglreith

kvark:
bors r=msiglreith

PR #3403, opened 2020-10-08
[dx12] fix and refactor the samplers

kvark:
This PR follows #3393. It combines a few closely related things around samplers:
  1. It turns out we can't copy GPU-visible descriptors into each other. We used to create samplers in the GPU-visible heap, and track the origin indices. Now we create them in a separate CPU-only pool, and only copy from there into the GPU heap.
  2. The previous PR was missing a piece of logic that would update the sampler data on a descriptor inside `copy_descriptor_sets`.
  3. From the early days, we had code duplication between descriptor copies and writes. This PR introduces a structure that manages those copy ranges and flushes them.
  4. Using copy ranges more aggressively, and non-symmetrically between the source and destination.

### DIFF
--- a/src/backend/dx12/src/descriptors_cpu.rs
+++ b/src/backend/dx12/src/descriptors_cpu.rs
@@ -161,3 +161,64 @@ impl DescriptorCpuPool {
         }
     }
 }
+
+#[derive(Default)]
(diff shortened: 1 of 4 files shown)


msiglreith (approved):
bors r+

PR #3405, opened 2020-10-12
[0.6/dx12] get proper compressed textures support

kvark:
Fixes https://github.com/gfx-rs/wgpu/issues/956
It never actually worked, so the example broke as soon as we switched it to Bc1 format.


cwfitzgerald (approved):
Nice!

kvark:
bors r=cwfitzgerald

Issue #3219, opened 2020-04-19
Support for OpenXR


lachlansneff:
OpenXR is pretty new, but there are basic bindings for it in rust already https://docs.rs/openxr. Eventually, supporting this would enable applications that use `gfx-hal` directly or libraries that use `gfx-hal`, like `wgpu`, to transparently support most vr and ar headsets on most architectures.

I'm proposing that support gets added to gfx-hal. I can contribute to this, but I don't know where to start.

matthew-russo:
I'm beginning to experiment with OpenXR integration on my own and curious to see where this goes. I'm happy to help as well but by far not an expert with gfx or graphics in general and don't have a solid grasp on all the concepts to propose a clean abstraction between the two.

I think there would need to be some clear definition of what support would mean. It could range from providing structures and functions bridge the gap and leave the rest to the user all the way to embedding OpenXR in gfx-hal so that things like the swapchain creation and management are pure gfx-hal and it doesn't appear to the user as an external API.

Curious to get input from the gfx maintainers on this and see if there has been any other efforts in to this.

lachlansneff:
Yes, I think making it mostly transparent for the user is the right way to go. OpenXR will take some vulkan or opengl instance data and present you will swapchains to render onto. Hiding all the configuration will be complex, as there's a lot of various vr-specific things, like positioning in space and whatnot that probably should not be hidden.

Maybe it'd be a good idea to port WebXR over to native as a wrapper over openxr, just as WebGPU has been ported to native.

jdm:
Servo has developed a webxr crate with an openxr backend, if that helps at all: https://github.com/servo/webxr/

str4d:
The OpenXR backend in `webxr` hard-codes `D3D11` as the graphics backend, so that would need to be generalised before it could be used here.

[...]

PR #3409, opened 2020-10-15
[vk] Migrate ash structs to builders

str4d:
I didn't call builder methods for values that were clearly being set to their defaults in the struct. I did leave e.g. calls to `*Flags::empty()` and other non-obvious defaults, as well as any values that were default but marked as `TODO`.

PR checklist:
  - WSL doesn't have graphics drivers.
  - I don't have `make` on Windows, so I manually ran `cd src/warden && cargo run --bin reftest --features vulkan -- local`
  - Except `mesh-shading` (I didn't have that feature).

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -14,7 +14,7 @@ use hal::{
 };
 
 use std::borrow::Borrow;
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::ops::Range;
 use std::pin::Pin;
 use std::sync::Arc;
@@ -59,8 +59,8 @@ impl GraphicsPipelineInfoBuf {
         source: &pso::EntryPoint<'a, B>,
     ) {
         let string = CString::new(source.entry).unwrap();
-        let p_name = string.as_ptr();
         self.c_strings.push(string);
+        let name = self.c_strings.last().unwrap().as_c_str();
 
         self.specialization_entries.push(
             source
@@ -83,15 +83,15 @@ impl GraphicsPipelineInfoBuf {
             p_data: source.specialization.data.as_ptr() as _,
         });
 
-        self.stages.push(vk::PipelineShaderStageCreateInfo {
-            s_type: vk::StructureType::PIPELINE_SHADER_STAGE_CREATE_INFO,
-            p_next: ptr::null(),
-            flags: vk::PipelineShaderStageCreateFlags::empty(),
-            stage,
-            module: source.module.raw,
-            p_name,
-            p_specialization_info: self.specializations.last().unwrap(),
-        })
+        self.stages.push(
+            vk::PipelineShaderStageCreateInfo::builder()
+                .flags(vk::PipelineShaderStageCreateFlags::empty())
+                .stage(stage)
+                .module(source.module.raw)
+                .name(name)
+                .specialization_info(self.specializations.last().unwrap())
+                .build(),
+        )
     }
 
     unsafe fn initialize<'a>(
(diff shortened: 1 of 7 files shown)


src/backend/vulkan/src/device.rs
```diff
@@ -212,54 +203,45 @@ impl GraphicsPipelineInfoBuf {
...
 
-        this.rasterization_state = vk::PipelineRasterizationStateCreateInfo {
-            s_type: vk::StructureType::PIPELINE_RASTERIZATION_STATE_CREATE_INFO,
-            p_next: ptr::null(),
-            flags: vk::PipelineRasterizationStateCreateFlags::empty(),
-            depth_clamp_enable: if desc.rasterizer.depth_clamping {
+        this.rasterization_state = vk::PipelineRasterizationStateCreateInfo::builder()
+            .flags(vk::PipelineRasterizationStateCreateFlags::empty())
+            .depth_clamp_enable(if desc.rasterizer.depth_clamping {
                 if device.shared.features.contains(Features::DEPTH_CLAMP) {
```
kvark, line 209: we don't really need to check for this, can just call `depth_clamp_enable(desc.rasterizer.depth_clamping)`

kvark (commented):
That's an interesting pull request. Thank you for making it!
As usual with non-trivial PRs not linked to any issues, there comes a question of - what is the motivation for doing this?

One thing that's clearly a win is that we don't need to write `vk::StructureType::` things in all the structs. This is an improvement.
On the other hand, builders hide the fields that we'd potentially want to set but not doing right now. Seeing `let builder = ... builder.something();` in a few places is also technically worse than just modifying the structs.

In general, I think builders work best when used by a *higher* level logic. For example, by the user directly, or by an engine. In this case, they know what they need, and they leave everything else to be defaults. However, in gfx-rs case we pretty much want *everything*, so the builder makes less sense to use.

I just wanted to sync up on what you expect the outcomes to be from merging this.

str4d:
> what is the motivation for doing this?

I'm attempting to add OpenXR support to the Vulkan backend (#3219), and I'm currently running into a problem at the top of `Instance::enumerate_adapters` where the OpenXR runtime is returning an `RUNTIME_FAILURE` error in the integration, but doesn't in the supposedly-equivalent code in the `openxr` crate's Vulkan example. I made these changes to `gfx_backend_vulkan::Instance::create` to reduce the differences between the two codebases (to make it easier to hunt down the issue).

I figured that the changes might be independently of interest, since I saw some existing places that `gfx-backend-vulkan` _is_ using the `ash` builders (added in the last few months IIRC), and assumed that the reason for not using them was that they didn't exist three years ago when most of the code I touched was written. I don't have any particular motivation to get these changes merged; take what makes sense 🙂

kvark:
Ok, I want to hear out if other members have opinions on this. Otherwise, we'll proceed (let's say, tomorrow)

str4d:
Ooh, interesting development: my OpenXR diff crashes when based on master, but does _not_ crash when rebased on this branch. So there's maybe an actual bug being fixed by this PR...

src/backend/vulkan/src/lib.rs
```diff
@@ -417,16 +414,11 @@ impl hal::Instance<Backend> for Instance {
...
             let str_pointers = cstrings.iter().map(|s| s.as_ptr()).collect::<Vec<_>>();
 
-            let create_info = vk::InstanceCreateInfo {
-                s_type: vk::StructureType::INSTANCE_CREATE_INFO,
-                p_next: ptr::null(),
-                flags: vk::InstanceCreateFlags::empty(),
-                p_application_info: &app_info,
-                enabled_layer_count: layers.len() as _,
-                pp_enabled_layer_names: str_pointers.as_ptr(),
-                enabled_extension_count: extensions.len() as _,
```
str4d, line 427: False alarm. The bug was here: I was combining `extensions` with the set of extensions required by the OpenXR runtime, and didn't spot that the wrong length was being used here. The builder API "fixed" the bug because it takes a slice instead of a length+pointer, avoiding this kind of problem entirely.

grovesNL:
@kvark using the builders sounds alright to me :+1: I don't have much of a preference either way since the trade-offs seem pretty minor

kvark (approved):
bors r+

PR #3410, opened 2020-10-15
[vk] allow different descriptors in a single write

### LINKED ISSUE #3408: Redesign descriptor writes
We currently have an API where each descriptor is provided as an `enum`, but they all have to be of the same type within a descriptor write, according to Vulkan spec. This turned out to be quite confusing, and I hit this in `wgpu` a few times.

Possible ways to proceed:
  1. leave everything unchanged
  2. make it closer to Vulkan by providing separate arrays of buffers, textures, and samplers
  3. make it nicer to use by moving the descriptor set out of the write, and allowing write descriptors to be anything. I may be mistaken, but I think only Vulkan backend would need to do anything about this. VkPi would also not care much.

I think my preference is (3), going to experiment more with it.

kvark:
Fixes #3408
PR checklist:
- [x] tested examples with the following backends: vulkan, metal, dx12

The quad example now uses this semantics to update all the descriptors in a single write.

### DIFF
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -451,23 +451,15 @@ where
         );
 
         unsafe {
-            device.write_descriptor_sets(vec![
-                pso::DescriptorSetWrite {
-                    set: &desc_set,
-                    binding: 0,
-                    array_offset: 0,
-                    descriptors: Some(pso::Descriptor::Image(
-                        &*image_srv,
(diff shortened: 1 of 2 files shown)


msiglreith (approved):
ci is failing

kvark:
bors r=msiglreith

PR #3411, opened 2020-10-15
[0.6/vk] allow different descriptors in a single write

kvark:
Sibling of #3410 for `hal-0.6` branch, to be used by wgpu


kvark:
bors r=msiglreith

Issue #3308, opened 2020-07-17
Fill in Limits Required by Wgpu


cwfitzgerald:
Now that wgpu has support for limits, the various gfx-hal backends need to make sure they are filling in at least all the limits required by wgpu. As of right now we max the limits with the default to force a minimum, but this isn't accurate to the hardware.

There is a decent amount of research that has to go into figuring out the exact values for all the backends as some backends (like dx12) some things don't have limits like vulkan does.

kvark:
I just realized that this issue doesn't actually tell which limits are missing.

cwfitzgerald:
So at this point, both DX11 and DX12 are updated, so unless you think metal needs more work, we can close this, at least until GL is re-enabled.

PR #3413, opened 2020-10-15
[0.6/dx12] Improve D3D12 limits

cwfitzgerald:
Helps progress #3308 towards completion. This improves the limits to better reflect reality. There are still some corner cases not covered by this, but it is much more accurate.

To derive these limits I used:
 - https://docs.microsoft.com/en-us/windows/win32/direct3d12/hardware-support
 - Various d3d constants.
 - Poking people on the directx discord to understand what should go where :)

Not ready to merge as I didn't do a version bump in case we wanted to roll this and any possible bug fixes to my bug into a single release.

### DIFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -1047,6 +1047,23 @@ impl hal::Instance<Backend> for Instance {
             //TODO: find a way to get a tighter bound?
             let sample_count_mask = 0x3F;
 
+            // Theoretically vram limited, but in practice 2^20 is the limit
+            let tier3_practical_descriptor_limit = 1 << 20;
+
+            let full_heap_count = match features.ResourceBindingTier {
+                d3d12::D3D12_RESOURCE_BINDING_TIER_1 => d3d12::D3D12_MAX_SHADER_VISIBLE_DESCRIPTOR_HEAP_SIZE_TIER_1,
+                d3d12::D3D12_RESOURCE_BINDING_TIER_2 => d3d12::D3D12_MAX_SHADER_VISIBLE_DESCRIPTOR_HEAP_SIZE_TIER_2,
+                d3d12::D3D12_RESOURCE_BINDING_TIER_3 => tier3_practical_descriptor_limit,
+                _ => unreachable!(),
+            } as _;
+
+            let uav_limit = match features.ResourceBindingTier {
+                d3d12::D3D12_RESOURCE_BINDING_TIER_1 => 8, // conservative, is 64 on feature level 11.1
+                d3d12::D3D12_RESOURCE_BINDING_TIER_2 => 64,
+                d3d12::D3D12_RESOURCE_BINDING_TIER_3 => tier3_practical_descriptor_limit,
+                _ => unreachable!(),
+            } as _;
+
             let physical_device = PhysicalDevice {
                 library: Arc::clone(&self.library),
                 adapter,
@@ -1063,6 +1080,7 @@ impl hal::Instance<Backend> for Instance {
                     Features::MULTI_DRAW_INDIRECT |
                     Features::FORMAT_BC |
                     Features::INSTANCE_RATE |
+                    Features::DEPTH_CLAMP |
                     Features::SAMPLER_MIP_LOD_BIAS |
                     Features::SAMPLER_ANISOTROPY |
                     Features::TEXTURE_DESCRIPTOR_ARRAY |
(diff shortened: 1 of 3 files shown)


src/backend/dx12/src/lib.rs
```diff
@@ -1047,6 +1047,23 @@ impl hal::Instance<Backend> for Instance {
             //TODO: find a way to get a tighter bound?
             let sample_count_mask = 0x3F;
 
+            // Theoretically vram limited, but in practice 2^20 is the limit
+            let tier3_practical_descriptor_limit = 1 << 20;
```
cwfitzgerald, line 1051: This particular bit was derived from "word of mouth" issues, not aware of a hard cite.

src/backend/dx12/src/lib.rs
```diff
@@ -1047,6 +1047,23 @@ impl hal::Instance<Backend> for Instance {
...
+
+            let full_heap_count = match features.ResourceBindingTier {
+                d3d12::D3D12_RESOURCE_BINDING_TIER_1 => d3d12::D3D12_MAX_SHADER_VISIBLE_DESCRIPTOR_HEAP_SIZE_TIER_1,
+                d3d12::D3D12_RESOURCE_BINDING_TIER_2 => d3d12::D3D12_MAX_SHADER_VISIBLE_DESCRIPTOR_HEAP_SIZE_TIER_2,
+                d3d12::D3D12_RESOURCE_BINDING_TIER_3 => tier3_practical_descriptor_limit,
+                _ => unreachable!(),
```
kvark, line 1057: this isn't really unreachable, i.e. a later DX version may report TIER_4

src/backend/dx12/src/lib.rs
```diff
@@ -1081,13 +1098,34 @@ impl hal::Instance<Backend> for Instance {
...
-                    max_per_stage_descriptor_samplers: d3d12::D3D12_COMMONSHADER_SAMPLER_REGISTER_COUNT as _,
-                    max_per_stage_descriptor_storage_buffers: 4,
-                    max_per_stage_descriptor_storage_images: 4,
-                    max_per_stage_descriptor_uniform_buffers: d3d12::D3D12_COMMONSHADER_CONSTANT_BUFFER_API_SLOT_COUNT as _,
+                    max_descriptor_set_sampled_images: full_heap_count,
+                    max_descriptor_set_storage_buffers: uav_limit,
+                    max_descriptor_set_storage_images: uav_limit,
+                    max_descriptor_set_uniform_buffers: full_heap_count,
+                    max_descriptor_set_samplers: d3d12::D3D12_MAX_SHADER_VISIBLE_SAMPLER_HEAP_SIZE as _,
+                    max_per_stage_descriptor_sampled_images: match features.ResourceBindingTier {
```
kvark, line 1106: why is this not `d3d12::D3D12_COMMONSHADER_INPUT_RESOURCE_REGISTER_COUNT` ?

src/backend/dx12/src/lib.rs
```diff
@@ -1081,13 +1098,34 @@ impl hal::Instance<Backend> for Instance {
...
+                        d3d12::D3D12_RESOURCE_BINDING_TIER_1 => 128,
+                        d3d12::D3D12_RESOURCE_BINDING_TIER_2
+                        | d3d12::D3D12_RESOURCE_BINDING_TIER_3 => full_heap_count,
+                        _ => unreachable!(),
+                    } as _,
+                    max_per_stage_descriptor_samplers: match features.ResourceBindingTier {
```
kvark, line 1112: why is this not `d3d12::D3D12_COMMONSHADER_SAMPLER_REGISTER_COUNT` ?

src/backend/dx12/src/lib.rs
```diff
@@ -1081,13 +1098,34 @@ impl hal::Instance<Backend> for Instance {
...
+                        d3d12::D3D12_RESOURCE_BINDING_TIER_1 => 16,
+                        d3d12::D3D12_RESOURCE_BINDING_TIER_2
+                        | d3d12::D3D12_RESOURCE_BINDING_TIER_3 => d3d12::D3D12_MAX_SHADER_VISIBLE_SAMPLER_HEAP_SIZE,
+                        _ => unreachable!(),
+                    } as _,
+                    max_per_stage_descriptor_storage_buffers: uav_limit,
```
kvark, line 1118: that seems rather high, i.e. it's higher than `D3D12_UAV_SLOT_COUNT`

src/backend/dx12/src/lib.rs
```diff
@@ -1104,15 +1142,16 @@ impl hal::Instance<Backend> for Instance {
                         depth: 1,
                     },
                     max_compute_work_group_count: [
+                        d3d12::D3D12_CS_DISPATCH_MAX_THREAD_GROUPS_PER_DIMENSION,
```
kvark, line 1145: wow, nice fixes!

src/backend/dx12/src/lib.rs
cwfitzgerald: Alright, I can default it to be the same as 3 for now.

src/backend/dx12/src/lib.rs
cwfitzgerald: With both this and `max_per_stage_descriptor_samplers`, you have the problem of descriptor arrays. `Texture2D<float> array[10000]` counts 10k against this limit, but only uses one register. As far as I'm aware there is no way to express "max_descriptor_registers" in this limit set as vulkan doesn't seem to have a direct analogue.

src/backend/dx12/src/lib.rs
cwfitzgerald: The table says that tier 3 gets the "full heap", so I would imagine this would be okay.

src/backend/dx12/src/lib.rs
kvark: wow, that's unfortunate! so many limits, yet here there is a big problem now for D3D12 and Metal backends :/

kvark (approved):

cwfitzgerald:
Updated with comments. I also added the DEPTH_CLAMP feature which is guaranteed supported and is properly wired up.

kvark:
we need a separate commit with the bumped patch version and a changelog entry

cwfitzgerald:
Added and rebased.

kvark (approved):
bors r+

PR #3414, opened 2020-10-16
Implement std::error::Error for UnsupportedBackend

Neo-Zhixing:
```rust
let instance = gfx_backend::Instance::create("cs535", 1)?;
```
gives compiling error
```
error[E0277]: the trait bound `UnsupportedBackend: std::error::Error` is not satisfied
  --> src/hardware.rs:18:65
   |
18 |         let instance = gfx_backend::Instance::create("cs535", 1)?;
   |                                                                 ^ the trait `std::error::Error` is not implemented for `UnsupportedBackend`
   |
   = note: required because of the requirements on the impl of `From<UnsupportedBackend>` for `Box<dyn std::error::Error>`
   = note: required by `from`
```

PR checklist:

### DIFF
--- a/src/hal/src/lib.rs
+++ b/src/hal/src/lib.rs
@@ -481,6 +481,15 @@ pub enum IndexType {
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub struct UnsupportedBackend;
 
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!
bors r+

PR #3416, opened 2020-10-17
[0.6/dx11] Assorted Fixes for WGPU Examples

cwfitzgerald:
Further improvements on #3308 

This PR should be read commit by commit and each commit is a free-standing change.

Changes (in order):
 - Shaders weren't properly cleared when using a pipeline with no shader after one with a shader.
   - This partially fixes the shadow example. There are now shadows, but in the wrong places.
   - I have added the dirty bits so that the interaction with builtin shaders can be properly tracked,
     and I don't try to set geometry, hull, or domain shaders when they are potentially unsupported.
 - Copied #3405 into DX11.
   - Fixes compressed shader use in the skybox example.
 - DX11 doesn't allow copying all layers of a 2DArray texture all at once. Added a loop to copy individually.
   - Fixes skybox entirely.
 - Gave the feature-and-limit treatment to DX11, minding feature levels as much as I could. 
   - https://docs.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-devices-downlevel-intro#feature-support-for-feature-levels-12_1-through-9_3 was my primary reference
   - As descriptor indexing isn't possible in DX11, I ommitted the descriptor array limit change, despite DX11 supporting texture descriptor arrays.
   - I ommitted any feature that the backend doesn't actually support yet (such as indirect).
   - I did _not_ ommit feeatures that might be buggy :)
   - WGPU no longer complains about missing features.


src/backend/dx11/src/lib.rs
```diff
@@ -1305,17 +1334,41 @@ impl CommandBufferState {
...
 
                 context.VSSetShader(pipeline.vs.as_raw(), ptr::null_mut(), 0);
-                if let Some(ref ps) = pipeline.ps {
-                    context.PSSetShader(ps.as_raw(), ptr::null_mut(), 0);
+
+                if self.dirty_flag.contains(DirtyStateFlag::PIPELINE_PS) {
+                    if let Some(ref ps) = pipeline.ps {
+                        context.PSSetShader(ps.as_raw(), ptr::null_mut(), 0);
+                    } else {
+                        context.PSSetShader(ptr::null_mut(), ptr::null_mut(), 0);
```
kvark, line 1417: Could we reduce the branching here and in similar places? i.e.
```rust
let res = pipeline.ps.as_ref().map_or(ptr::null_mut(), |ps| ps.as_raw());
context.PSSetShader(res, ptr::null_mut(), 0);
```

src/backend/dx11/src/lib.rs
```diff
@@ -1082,8 +1082,12 @@ bitflags! {
         const RENDER_TARGETS = (1 << 1);
         const VERTEX_BUFFERS = (1 << 2);
         const GRAPHICS_PIPELINE = (1 << 3);
```
kvark, line 1159: what's the semantic of `GRAPHICS_PIPELINE` now that you have each of the stages in a separate flag?

src/backend/dx11/src/lib.rs
```diff
@@ -1034,7 +1034,7 @@ impl RenderPassCache {
 
         cache
             .dirty_flag
-            .insert(DirtyStateFlag::GRAPHICS_PIPELINE | DirtyStateFlag::VIEWPORTS);
+            .insert(DirtyStateFlag::GRAPHICS_PIPELINE | DirtyStateFlag::PIPELINE_PS | DirtyStateFlag::VIEWPORTS);
```
kvark, line 1112: why does a subpass change `PIPELINE_PS` and not other stages?

src/backend/dx11/src/internal.rs
```diff
@@ -854,34 +854,37 @@ impl Internal {
...
+                for layer in info.image_layers.layers.clone() {
+                    let layer_offset = layer - info.image_layers.layers.start;
+                    unsafe {
+                        context.UpdateSubresource(
+                            dst.internal.raw,
+                            dst.calc_subresource(
+                                info.image_layers.level as _,
+                                layer as _,
+                            ),
+                            &d3d11::D3D11_BOX {
```
kvark, line 873: nit: we should probably construct this outside of the loop for clarity

src/backend/dx11/src/lib.rs
cwfitzgerald: This change to the dirty flag is because the following clear_attachments invalidates VS, PS, etc

src/backend/dx11/src/lib.rs
cwfitzgerald: This says that _something_ in the pipeline changed. Be it any of the individual flags, the vertex shader, raster state, etc.

PR #3417, opened 2020-10-17
[0.6/dx11] Fix Read-Only Depth Stencil

cwfitzgerald:
Depends on #3416. Commits will be rebased away once this is merged.

Fixes the water example under DX11.

### DIFF
--- a/src/backend/dx11/src/conv.rs
+++ b/src/backend/dx11/src/conv.rs
@@ -691,7 +691,7 @@ fn map_stencil_side(side: &StencilFace) -> D3D11_DEPTH_STENCILOP_DESC {
 
 pub(crate) fn map_depth_stencil_desc(
     desc: &DepthStencilDesc,
-) -> (D3D11_DEPTH_STENCIL_DESC, State<StencilValue>) {
+) -> (D3D11_DEPTH_STENCIL_DESC, State<StencilValue>, bool) {
     let (depth_on, depth_write, depth_func) = match desc.depth {
         Some(ref depth) => (TRUE, depth.write, map_comparison(depth.fun)),
         None => unsafe { mem::zeroed() },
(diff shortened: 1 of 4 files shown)


src/backend/dx11/src/conv.rs
```diff
@@ -730,6 +730,15 @@ pub(crate) fn map_depth_stencil_desc(
         None => unsafe { mem::zeroed() },
     };
 
+    let stencil_read_only = write_mask == 0 ||
```
kvark, line 733: I wonder if we should make those conditions even more precise (at some point). I.e. technically speaking, if your stencil function is ALWAYS, then it doesn't matter what `StencilFailOp` is for any of the sides.

src/backend/dx11/src/device.rs
```diff
@@ -633,6 +642,13 @@ impl Device {
         let mut desc: d3d11::D3D11_DEPTH_STENCIL_VIEW_DESC = unsafe { mem::zeroed() };
         desc.Format = info.format;
 
+        if let Some(stencil) = read_only_stencil {
+            desc.Flags = match stencil {
+                true => d3d11::D3D11_DSV_READ_ONLY_DEPTH | d3d11::D3D11_DSV_READ_ONLY_STENCIL,
```
kvark, line 647: there is never a case where we end up with just `d3d11::D3D11_DSV_READ_ONLY_STENCIL`? what if the format is stencil-only, and it's read-only?

kvark (commented):
Once concern here. Great stuff overall!

src/backend/dx11/src/device.rs
cwfitzgerald: DX11 doesn't seem to support stencil only formats, at least as far as I can tell.

src/backend/dx11/src/conv.rs
cwfitzgerald: Good point, shouldn't be too hard to add.

PR #3422, opened 2020-10-18
[0.6/dx11] Support Dynamic Offsets into Uniform Buffers

cwfitzgerald:
This fixes the wgpu shadow example.

Couple code notes:
- There is a duplicated function, but due to the different return values, it's not possible to easily abstract it without jumping through tons of hoops.
- The bind descriptor set functions had one cast too many on their pointers, so I removed them.

Updated changelog. I think next version bump should be once I finish fixing all the examples (only msaa-line is left afaik).


src/backend/dx11/src/lib.rs
```diff
@@ -2037,49 +2156,84 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
...
-            // TODO: offsets
+            let (vs_offsets, fs_offsets) = generate_graphics_dynamic_constant_buffer_offsets(
+                &*set.layout.bindings,
+                &mut offset_iter,
+                self.context1.is_some()
+            );
+
+            // TODO: This should be the actual buffer length for RBA purposes,
+            //       but that information isn't easily accessible here.
+            let vs_counts = vec![4096_u32; vs_offsets.len()];
```
kvark, line 2167: can we have less dynamic allocations in hot functions like this one?
I.e. `vs_offsets`, `vs_counts`, `fs_offsets`, `fs_counts` - all of them could be in some `struct Temp` within the command buffer, just re-used.

src/backend/dx11/src/lib.rs
```diff
@@ -2148,35 +2304,56 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
...
-            // TODO: offsets
+            let cs_offsets = generate_compute_dynamic_constant_buffer_offsets(
+                &*set.layout.bindings,
+                &mut offset_iter,
+                self.context1.is_some()
+            );
+
+            // TODO: This should be the actual buffer length for RBA purposes,
+            //       but that information isn't easily accessible here.
+            let cs_counts = vec![4096_u32; cs_offsets.len()];
```
kvark, line 2315: same concern here

kvark (commented):
This is great stuff!
My only concern here is heap allocations in hot code.

src/backend/dx11/src/lib.rs
```diff
@@ -1517,9 +1514,128 @@ impl CommandBufferState {
...
+    }
+
+    (vs_offsets, fs_offsets)
+}
+
+fn generate_compute_dynamic_constant_buffer_offsets<'a>(
+    bindings: impl IntoIterator<Item = &'a pso::DescriptorSetLayoutBinding>,
+    offset_iter: &mut impl Iterator<Item = u32>,
+    context1_some: bool,
+) -> ArrayVec<[UINT; d3d11::D3D11_COMMONSHADER_CONSTANT_BUFFER_API_SLOT_COUNT as _]> {
```
kvark, line 1584: can we has a typedef for this array vec? :)

kvark (approved):
looks awesome! One small nit, and we are good to go

src/backend/dx11/src/lib.rs
cwfitzgerald: Done, the name of which is infinitely bikeshed-able 😄

cwfitzgerald:
bors r=kvark

kvark:
bors retry

cwfitzgerald:
All wgpu examples besides msaa-line and texture-arrays pass, as I expected, so we should be good to go.

bors r=kvark

PR #3423, opened 2020-10-19
[0.6/dx11] Fix multisampling

cwfitzgerald:
Fixes the `msaa-line` wgpu example.

Now only remains the texture-array example, and we're golden :)

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -92,9 +92,10 @@ impl Device {
     fn create_rasterizer_state(
         &self,
         rasterizer_desc: &pso::Rasterizer,
+        multisampling_desc: &Option<pso::Multisampling>
     ) -> Result<ComPtr<d3d11::ID3D11RasterizerState>, pso::CreationError> {
         let mut rasterizer = ptr::null_mut();
-        let desc = conv::map_rasterizer_desc(rasterizer_desc);
+        let desc = conv::map_rasterizer_desc(rasterizer_desc, multisampling_desc);
 
         let hr = unsafe {
             self.raw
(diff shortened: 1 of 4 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -588,15 +589,36 @@ impl Device {
                 }
             }
             image::ViewKind::D2 => {
```
cwfitzgerald, line 591: Interestingly, even for a 2D texture, the dx11 backend seems to only generate `2DTextureArray` views, so this code doesn't get used even in the `msaa-level` example.

kvark: yeah, IIRC there was a good reason for it :)

src/backend/dx11/src/lib.rs
```diff
@@ -226,6 +226,19 @@ fn get_limits(feature_level: d3dcommon::D3D_FEATURE_LEVEL) -> hal::Limits {
...
     };
 
+    // https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11device-checkmultisamplequalitylevels#remarks
+    // for more information.
+    let max_samples = match feature_level {
+        d3dcommon::D3D_FEATURE_LEVEL_9_1
+        | d3dcommon::D3D_FEATURE_LEVEL_9_2
+        | d3dcommon::D3D_FEATURE_LEVEL_9_3
+        | d3dcommon::D3D_FEATURE_LEVEL_10_0 => 0b0000_0001, // Conservative, MSAA isn't required.
+        d3dcommon::D3D_FEATURE_LEVEL_10_1 => 0b0000_1111, // Optimistic, 4xMSAA is required on all formats _but_ RGBA32.
```
kvark, line 236: why are we optimistic here?
Also, is this the correct mask? I often get it wrong, but you can check the Metal example for the reference. Maybe this should be `1 | 4` instead of this value

src/backend/dx11/src/device.rs
cwfitzgerald: I shall defer to experience :)

kvark (approved):
bors r+

PR #3425, opened 2020-10-20
[0.6/dx11] Fix writes of descriptor arrays

cwfitzgerald:
Depends on #3423.

Fixes the `texture-array` wgpu example. Once this is merged, all examples should at least run on DX11, and as such I have bumped versions.

This is a surprisingly small change for how long it took me to understand the code :)

I have added some comments to help explain a bit more how the binding situation works, especially now with arrays involved.

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -1775,7 +1775,7 @@ impl device::Device<Backend> for Device {
 
         for binding in bindings.iter() {
             let content = DescriptorContent::from(binding.ty);
-            total.add_content(content, binding.stage_flags);
+            total.add_content_many(content, binding.stage_flags, binding.count as _);
         }
 
         bindings.sort_by_key(|a| a.binding);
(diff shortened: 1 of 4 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -1797,27 +1797,47 @@ impl device::Device<Backend> for Device {
...
-                mapping.add_content(content, binding.stage_flags);
+                mapping.add_content_many(content, binding.stage_flags, binding.count as _);
             }
 
-            for (binding, descriptor) in write.set.layout.bindings[binding_start..]
-                .iter()
-                .zip(write.descriptors)
-            {
+            // We start at the given binding index and array index
+            let mut binding_index = write.binding as usize;
```
kvark, line 1823: this should be using `binding_start`, not `write.binding`. These are different things!

kvark (changes requested):
Looks great! Just one thing to fix

kvark (approved):
bors r+

cwfitzgerald:
didn't even touch that

bors retry

PR #3427, opened 2020-10-21
[0.6/dx11] Add Debug Names and More

cwfitzgerald:
Does some general cleanup, supports alpha-to-coverage, and adds debug names to various resoures.

Debug spans don't work right for an unknown reason, for investigation later.

### DIFF
--- a/src/backend/dx11/src/debug.rs
+++ b/src/backend/dx11/src/debug.rs
@@ -1,41 +1,15 @@
-use winapi::um::d3d11;
+use winapi::um::{d3d11, d3d11_1, d3dcommon};
 
 use wio::{com::ComPtr, wide::ToWide};
 
 use std::{env, ffi::OsStr, fmt};
 
-// TODO: replace with new winapi version when available
-#[allow(bad_style, unused)]
-mod temp {
-    use winapi::{
-        shared::minwindef::{BOOL, INT},
-        um::{
-            unknwnbase::{IUnknown, IUnknownVtbl},
-            winnt::LPCWSTR,
-        },
-    };
-
-    RIDL! {#[uuid(0xb2daad8b, 0x03d4, 0x4dbf, 0x95, 0xeb, 0x32, 0xab, 0x4b, 0x63, 0xd0, 0xab)]
-    interface ID3DUserDefinedAnnotation(ID3DUserDefinedAnnotationVtbl):
-        IUnknown(IUnknownVtbl) {
-        fn BeginEvent(
-            Name: LPCWSTR,
-        ) -> INT,
-        fn EndEvent() -> INT,
-        fn SetMarker(
-            Name: LPCWSTR,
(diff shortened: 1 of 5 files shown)


src/backend/dx11/src/debug.rs
```diff
@@ -84,11 +57,44 @@ pub fn debug_marker(context: &ComPtr<d3d11::ID3D11DeviceContext>, args: fmt::Arg
...
+    }
+    res
+}
+
+/// Set the debug name of a resource.
+///
+/// Must be ASCII.
+///
+/// SetPrivateData will copy the data internally so the data doesn't need to live.
+pub fn debug_name(resource: &d3d11::ID3D11DeviceChild, name: &str) {
```
kvark, line 82: let's call those as actions, i.e. `set_debug_name`

src/backend/dx11/src/device.rs
```diff
@@ -1519,10 +1538,15 @@ impl device::Device<Backend> for Device {
...
-                unordered_access_views.push(
-                    self.view_image_as_unordered_access(&view)
-                        .map_err(|_| device::BindError::WrongMemory)?,
-                );
+                let uav = self
+                    .view_image_as_unordered_access(&view)
+                    .map_err(|_| device::BindError::WrongMemory)?;
+
+                if let Some(ref mut name) = image.internal.debug_name {
+                    debug_name_with_suffix(&uav, name, &format!(" -- UAV Mip {}", mip));
```
kvark, line 1546: if we are going with `format!`, let's use the baseline `debug_name` for it. There is no point to pretend that `debug_name_with_suffix` helps here :)

src/backend/dx11/src/lib.rs
```diff
@@ -20,7 +20,6 @@ of continuous descriptor ranges (per type, per shader stage).
 extern crate bitflags;
 #[macro_use]
 extern crate log;
-#[macro_use]
 extern crate winapi;
```
kvark, line 23: no need for `extern crate` declarations in Rust2018 if you aren't importing macros

src/backend/dx11/src/debug.rs
```diff
@@ -73,7 +47,6 @@ impl Drop for DebugScope {
     }
 }
 
-#[cfg(debug_assertions)]
 pub fn debug_marker(context: &ComPtr<d3d11::ID3D11DeviceContext>, args: fmt::Arguments) {
```
kvark, line 50: this should probably just get an `&str`. No point to receive the arguments if all we do is just creating a `String` for them, anyway

kvark (commented):
A few minor notes. Overall looks great!

cwfitzgerald:
Comments should be addressed, I squashed the commits as keeping them separate was too difficult.

kvark:
thanks!
bors r+

PR #3433, opened 2020-10-23
[0.6/dx11] Fix two memory leaks

cwfitzgerald:
This fixes memory leaks in various wgpu examples, as well as leaks on swapchain resize.

I have used `thunderdome` as my arena which is a simple, limited-unsafe, no-deps, implementation of a generational arena. It's on the newer end, but I've been seeing people use it successfully in the community, combined to the lack of unsafe, I think it should be perfectly stable for use here. 

Will rollup master after this passes review.

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -2913,9 +2913,6 @@ pub struct Memory {
 
     // list of all buffers bound to this memory
     local_buffers: Arc<RwLock<LocalResourceArena<InternalBuffer>>>,
-
-    // list of all images bound to this memory
-    local_images: Arc<RwLock<LocalResourceArena<InternalImage>>>,
 }
 
 impl fmt::Debug for Memory {
@@ -3169,6 +3166,17 @@ pub struct InternalImage {
     render_target_views: Vec<ComPtr<d3d11::ID3D11RenderTargetView>>,
 
     debug_name: Option<String>
+}
+
+impl InternalImage {
+    unsafe fn release_resources(&mut self) {
+        (&*self.raw).Release();
+        self.copy_srv = None;
+        self.srv = None;
+        self.unordered_access_views.clear();
+        self.depth_stencil_views.clear();
+        self.render_target_views.clear();
+    }
 }
 
 impl fmt::Debug for InternalImage {
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -43,9 +43,9 @@ use winapi::{shared::{
 use wio::com::ComPtr;
 
 use arrayvec::ArrayVec;
-use parking_lot::{Condvar, Mutex};
+use parking_lot::{Condvar, Mutex, RwLock};
 
-use std::{borrow::Borrow, cell::RefCell, fmt, mem, ops::Range, os::raw::c_void, ptr, sync::Arc};
+use std::{borrow::Borrow, cell::RefCell, fmt, mem, ops::Range, os::raw::c_void, ptr, sync::{Arc, Weak}};
 
 macro_rules! debug_scope {
     ($context:expr, $($arg:tt)+) => ({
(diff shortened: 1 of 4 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -3088,12 +3091,33 @@ pub struct InternalBuffer {
...
+    /// If accessed from a Buffer, once memory is bound to a buffer, this should never be None.
+    memory_index: Option<thunderdome::Index>,
     srv: Option<*mut d3d11::ID3D11ShaderResourceView>,
     uav: Option<*mut d3d11::ID3D11UnorderedAccessView>,
     usage: buffer::Usage,
     debug_name: Option<String>,
 }
 
+impl InternalBuffer {
+    unsafe fn release_resources(&self) {
```
kvark, line 3113: let's have it `&mut self` and zeroing out the internal fields, just for ease of debugging later on

src/backend/dx11/src/lib.rs
```diff
@@ -3088,12 +3091,33 @@ pub struct InternalBuffer {
...
+    /// The copy of the InternalBuffer inside of the arena will have an empty Weak.
+    ///
+    /// If accessed from a Buffer, once memory is bound to a buffer, this should never be None.
+    local_memory_arena: Weak<RwLock<LocalResourceArena<InternalBuffer>>>,
+    /// Index into the above memory arena.
+    ///
+    /// The copy of the InternalBuffer inside of the arena will have None.
+    ///
+    /// If accessed from a Buffer, once memory is bound to a buffer, this should never be None.
+    memory_index: Option<thunderdome::Index>,
```
kvark, line 3105: as discussed on Matrix, let's move this out of `InternalBuffer`

src/backend/dx11/src/lib.rs
```diff
@@ -3150,6 +3174,17 @@ pub struct InternalImage {
     debug_name: Option<String>
 }
 
+impl InternalImage {
+    unsafe fn release_resources(&mut self) {
+        (&*self.raw).Release();
+        self.copy_srv = None;
+        self.srv = None;
+        self.unordered_access_views.clear();
```
kvark, line 3182: we aren't calling `Release` on each view? This seems different from the `InternalBuffer` logic

cwfitzgerald: They're owned; they're `Option<ComPtr<Whatever>>`

src/backend/dx11/src/lib.rs
```diff
@@ -3094,11 +3094,29 @@ pub struct InternalBuffer {
     debug_name: Option<String>,
 }
 
+impl InternalBuffer {
+    unsafe fn release_resources(&mut self) {
+        (&*self.raw).Release();
+        self.raw = ptr::null_mut();
+        self.disjoint_cb.take().map(|cb| (&*cb).Release());
```
kvark, line 3101: I'm wary of using functional style (`map`) where you express clear imperative steps, like here. You aren't really mapping a value. if this was written as `if let Some(xx) = yy.take() { do_zz() }` it would look nicer, have less sugar, and be easier to debug, at the cost of slight increase in verbosity

kvark (approved):
bors r+

src/backend/dx11/src/lib.rs
cwfitzgerald: I was actually surprised there didn't seem to be a `for_each` equivilant for `Option`. I personally like the functional part of it, but I agree that the verb is inaccurate considering what is happening.

That being said, there is definitely the argument that the if-let will have better debug perf. Will keep in mind going forward.

kvark: I've just had to debug a bunch of `xxx.do()?.map(|| ...).map_err(|| ...)` and realize how wasteful all the sugar is compared to an old good `match` :)

PR #3434, opened 2020-10-24
[0.6/dx12] Improve descriptor handling

kvark:
Can be reviewed commit-by-commit. Actually gets us to release the descriptors.
It also fixes a mistake I made in one of the previous PRs that would panic in view creation, unnecessarily.

Also works around https://github.com/gfx-rs/wgpu/issues/1002 in the following way:
  - we rotate the temporary descriptor heaps instead of resetting them every time
  - once we reach the head during rotation, we insert a new heap in

That is not a guarantee that we'll never hit this WARP issue again, but there is a high chance most apps will not notice. If there is a better workaround, I'll be happy to scrap this one.

I think the old logic was plain wrong, too: the `update_pool_index ` wasn't actually being used...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+### backend-dx12-0.6.10 (24-10-2020)
+  - actually free descriptor sets
+  - work around descriptor update issue on WARP
+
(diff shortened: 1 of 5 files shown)


msiglreith (approved):
bors r+

PR #3437, opened 2020-10-25
[0.6/auxil/gl/dx11/dx12/metal] Select entry point when multiple exist

### LINKED ISSUE #3435: Ensure that shaders are generated once per entry point
rust-gpu outputs vertex and fragment shaders in a single SPIR-V file, but we haven't been testing this path much yet because we usually receive shaders as separate files. This appears to cause problems for the MSL generation path (possibly HLSL and GLSL too but I haven't checked them yet).

Currently it looks like we convert SPIR-V to MSL for one entry point only and then try to reuse the cache, but this fails with missing entry points when they exist in a single file. We can probably include the entry point as part of the cache key to avoid reusing the cache across different entry points for now (i.e. until this is supported in SPIRV-Cross/naga).

For example, if a shader has `main_vs` and `main_fs`, we might compile only one of these paths. So if we compiled/cached the fragment shader entry point, we might end up with `Device::create_render_pipeline)(no module): Function retrieval error "Function \'main_vs\' does not exist"` later, because the `main_vs` entry doesn't exist.

grovesNL:
Fixes #3435
PR checklist:
- [X] tested examples with the following backends: dx11, dx12, metal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-metal-0.6.4, backend-dx12-0.6.11, backend-dx11-0.6.9 (26-10-2020)
+  - fix entry point selection to choose the correct entry point (instead of a default) when multiple entry points exist
+
 ### backend-dx12-0.6.10 (24-10-2020)
   - actually free descriptor sets
   - work around descriptor update issue on WARP
--- a/src/auxil/auxil/Cargo.toml
+++ b/src/auxil/auxil/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-auxil"
-version = "0.6.0"
+version = "0.7.0"
 description = "Implementation details shared between gfx-rs backends"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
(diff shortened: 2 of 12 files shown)


src/auxil/auxil/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-auxil"
-version = "0.6.0"
+version = "0.6.1"
```
kvark, line 3: this isn't backwards compatible, should be 0.7.0 now

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-metal-0.6.4, backend-dx12-0.6.11, backend-dx11-0.6.9 (26-10-2020)
+  - specify entry point when shaders contain multiple entry points
```
kvark, line 4: would be nice to tell what user impact this brings, e.g. "fixes a case where something something"

src/auxil/auxil/Cargo.toml
grovesNL: Oops, nice catch :)

CHANGELOG.md
grovesNL: :+1: it should be clearer now

kvark (approved):
Thank you!
bors r+

PR #3440, opened 2020-10-26
[dx12] fix command list reuse

### LINKED ISSUE #3439: D3D12 conditions for command list re-use
Getting this with the updated D3D12 code on Dota2 via gfx-portability:

> D3D12 ERROR: ID3D12CommandQueue::ExecuteCommandLists: Failed to execute a command list 0x0000016B33A28A40:'Unnamed ID3D12GraphicsCommandList Object' because the command queue fence has not advanced past previous executions of the command list. [ EXECUTION ERROR #553: COMMAND_LIST_SYNC]

kvark:
Fixes #3439
We used to instantly return a command list into the pool upon submit() if we knew it's a one-time submit.
However, it could happen that the user would get that command list, record it, and submit before the last one has finished.
As indicated in the issue, this can't be allowed.

So this PR makes it so we keep a hold of the command list until the reset() time.

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -13,7 +13,7 @@ use winapi::{
     Interface,
 };
 
-use std::{borrow::Borrow, cell::Cell, cmp, fmt, iter, mem, ops::Range, ptr, sync::Arc};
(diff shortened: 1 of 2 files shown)


cwfitzgerald:
bors retry

PR #3442, opened 2020-10-26
[0.6/dx11] Fix UAV bindings 

cwfitzgerald:
Fixes https://github.com/gfx-rs/wgpu/issues/997 and assorted issues discovered when fixing that.

This PR should be reviewed commit by commit and each commit is freestanding.

A note about the general arch of the UAV changes. This moves us more towards a "lazy state changes" model. ~~While vulkan discorages this, the DX11 driver is likely doing this anyway. We currently do a lot of redundant state changes (and the calculations to figure them out) due to the differences in binding model, so moving lazy should help eliminate this. More major changes are out of scope of this PR.~~ Discussed on matrix pros/cons of this approach, and this part is no longer relevant.

I will actually PR master after this PR this time ;)

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -561,6 +561,8 @@ impl Device {
         if winerror::SUCCEEDED(hr) {
             Ok(unsafe { ComPtr::from_raw(uav) })
         } else {
+            error!("CreateUnorderedAccessView failed: 0x{:x}", hr);
+
             Err(image::ViewCreationError::Unsupported)
         }
     }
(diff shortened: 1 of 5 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -677,15 +681,38 @@ impl Device {
 
         match info.view_kind {
             image::ViewKind::D2 => {
-                desc.ViewDimension = d3d11::D3D11_DSV_DIMENSION_TEXTURE2D;
-                *unsafe { desc.u.Texture2D_mut() } = d3d11::D3D11_TEX2D_DSV { MipSlice }
+                match info.kind {
+                    image::Kind::D2(_, _, _, 1) => {
```
kvark, line 685: seems unnecessary to repeat this match, it could be simpler to just do something like `if info.kind.num_samples() > 1 {..} else {...}` in both places

kvark (commented):
(review commit 1)

kvark (approved):
(review commit 2)

src/backend/dx11/src/device.rs
```diff
@@ -1889,7 +1889,18 @@ impl device::Device<Backend> for Device {
 
         for binding in bindings.iter() {
             let content = DescriptorContent::from(binding.ty);
-            total.add_content_many(content, binding.stage_flags, binding.count as _);
+            // If this binding is used by the graphics pipeline and is a UAV, it belongs to the "Output Merger"
+            // stage, so we only put them in the fragment stage to save redundant descriptor allocations.
+            let stage_flags =
```
kvark, line 1894: could be simpler to just write this as
```rust
if content.contains(DescriptorContent::UAV) {
  stage_flags &= !pso::ShaderStageFlags::VERTEX;
}
```
looks to be equivalent to what you wrote?

src/backend/dx11/src/shader.rs
```diff
@@ -191,12 +191,28 @@ fn patch_spirv_resources(
...
+        // In order to find out of the buffer is read-only, we must ask if the first member of its base type is read only.
+        let read_only = ast
+            .get_member_decoration(storage_buffer.base_type_id, 0, spirv::Decoration::NonWritable)
+            .map_err(gen_query_error)? != 0;
         let (_content, res_index) = layout.sets[set].find_register(stage, binding);
 
+        let index = if read_only {
+            // Read only storage buffers are SRVs
+            res_index.t as u32
+        } else {
```
kvark, line 203: nit: can use `else if`

kvark (approved):
(review commit 3)

src/backend/dx11/src/device.rs
cwfitzgerald: ~~This does seem equivalent if you only consider [VS, PS, COMP] as your shader set, but the dx11 backend could very well support tess/geo, so I'm trying to include them whenever I write logic like this.~~

This behaves differently if _only_ vertex is set; it will just remove VERTEX without also setting FRAGMENT.

Musing over if there's a way to simplify this...

kvark (approved):
Thank you!
Please publish when you get to it :)
bors r+

PR #3444, opened 2020-10-27
[0.6/dx11] Implement and Fix Assorted Issues

cwfitzgerald:
Fixes https://github.com/gfx-rs/wgpu/issues/1009.

Fixes assorted smaller issues. Implements push constants and draw indirect.

This PR should be reviewed commit by commit and each commit is freestanding.

After this PR lands, I think the DX11 backend _should_ have everything wgpu needs implemented in some form, so we can start telling people to try DX11 on their project as it might actually work (and I can start collecting bugs) 🎉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Change Log
 
+### backend-dx11-0.6.11 (27-10-2020)
+  - implement push constants
+  - implement indirect rendering
+  - fix error messages when performing unsupported draws
+  - fix memory invalidation and mapping
+  - actually set stencil reference value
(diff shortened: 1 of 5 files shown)


kvark (approved):
This is a whole bag of amazing work!
My only concern is that creating (and keeping it forever) a buffer for every command buffer is more expensive than it could be, especially for WebGPU where push constants are not even in the core spec.
Ideally, we'd create this buffer lazily upon the first pipeline set that has a layout with defined constant ranges, and free it in `reset()`. I don't think this is a blocker though, we can proceed and then keep an eye on making this better.

bors r+

Issue #71, opened 2014-07-05
Find a solution for shader language support


cmr:
Different graphics APIs have their own shading languages or ideas for what a shading language should be. In our quest to support multiple backend APIs, we need a way to unify these.

kvark:
Shaders become one of the major usability bottleneck of gfx-rs. Let's continue this discussion, even if the code for it will end up in a separate library.

kvark:
This might be worth looking at: https://github.com/kmcallister/glassful
Also, cc @csherratt

bjz:
Something to also think about is whether we want to support recompiling of shaders at runtime.

kmcallister:
I would very much like that feature.

kvark:
Inspired by Glassful and @csherratt's [example code](https://www.reddit.com/r/rust/comments/2ufeqf/rustlike_syntax_for_opengl_shading_language/co7z1tw) from Reddit, I tried to develop it further. Here is an example of what I'd like to use (very roughly):
```rust
/// uniforms
struct MyParams {
    color: vec4,
}
/// vertex buffer 1
struct MyBuf1 {
    pos: vec2,
}
/// vertex buffer 2
struct MyBuf2 {
    tc: vec2,
}
let technique = glassful! {
    struct Varying {
        tc: vec2,
    }
    // @Position is a special vertex output, not sure if we can make it more sound
    fn vertex(_params: MyParams, buf1: MyBuf1, buf2: MyBuf2) -> (@Position, Varying) {
        (vec4(buf1.pos, 0.0, 1.0), Varying { tc: buf2.tc })
    }
    fn fragment(params: MyParams, _var: Varying) -> vec4 {
        params.color
    }
    static pass0 = pass(vertex, fragment);
};
...
let program = device.link_program(&technique.pass0).unwrap();
```
The implementation doesn't seem trivial. Here are the biggest problems that I see:
  1. How to bring some struct definitions (i.e. `MyParams`, `MyBuf1`) into the macro, or outside of it?
  2. How to extract only the relevant items from the whole program when compiling a particular stage? I.e. a fragment program should not contain definitions of vertex buffers.
  3. How to translate from Rust functional style into GLSL/whatever imperative one? Just printing out the text while traversing the AST doesn't seem enough.
  4. How to support the run-time compilation, where the shader parameter struct and vertex formats are not known in advance in compile time?

Any ideas, feedback, suggestions - please let me know.

csherratt:
[...]

PR #3445, opened 2020-10-28
[mtl] experimental Naga support

kvark:
First humble steps towards #71 ...

Naga is an optional github dependency (obviously will need to be switched to crates upon release).
If it fails, we fall back to spirv-cross. Extra steps are taking to ensure that shader modules compiled by Naga and SPIRV-Cross can still link together.

PR checklist:


grovesNL (approved):
This is exciting, looks great so far! :+1:

PR #3446, opened 2020-10-28
[0.6/dx] Workaround spirv-cross Limitation

cwfitzgerald:
If a shader has a matrix as a vertex input (for things like per-instance data), spirv-cross will generate different semantics than we expect (see https://github.com/KhronosGroup/SPIRV-Cross/issues/1512). This works around the problem on both DX11 and DX12 backends. It is not implemented as efficiently as it possibly could be, but this was the easiest way to do this without really changing some functions around. Hopefully this will be resolved soon enough and this change can be removed.

I didn't bump versions as I have more fixes for both backends coming shortly.

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -165,6 +165,7 @@ impl Device {
         vertex_buffers: &[pso::VertexBufferDesc],
         attributes: &[pso::AttributeDesc],
         input_assembler: &pso::InputAssemblerDesc,
+        vertex_semantic_remapping: auxil::FastHashMap<u32, Option<(u32, u32)>>,
     ) -> Result<InputLayout, pso::CreationError> {
         let mut layout = ptr::null_mut();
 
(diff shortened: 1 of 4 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -182,9 +183,26 @@ impl Device {
             vertex_strides[buffer.binding as usize] = buffer.stride;
         }
 
+        // See [`shader::introspect_spirv_vertex_semantic_remapping`] for details of why this is needed.
+        let semantics: Vec<_> = attributes.iter().map(|attrib| {
```
kvark, line 187: it looks like constructing the `semantics :Vec` is not necessary here. Can't we just derive the semantic for each attribute on the spot within the `.filter_map(|attrib| {}` body?

src/backend/dx11/src/shader.rs
```diff
@@ -32,6 +32,54 @@ fn gen_query_error(err: SpirvErrorCode) -> device::ShaderError {
...
+/// The returned hashmap has attribute location as a key and an Optional remapping to a two part semantic.
+///
+/// eg.
+/// `2 -> None` means use default semantic `TEXCOORD2`
+/// `2 -> Some((0, 2))` means use two part semantic `TEXCOORD0_2`. This is how matrices are represented by spirv-cross.
+///
+/// This is a temporary workaround for https://github.com/KhronosGroup/SPIRV-Cross/issues/1512.
+///
+/// This workaround also exists under the same name in the DX12 backend.
+pub(crate) fn introspect_spirv_vertex_semantic_remapping(raw_data: &[u32]) -> Result<auxil::FastHashMap<u32, Option<(u32, u32)>>, device::ShaderError> {
```
kvark, line 46: If this code is to stay, we'd want it in the `auxil` crate.
If it's really temporary, then it's fine to keep it duplicated for now.

src/backend/dx11/src/shader.rs
```diff
@@ -32,6 +32,54 @@ fn gen_query_error(err: SpirvErrorCode) -> device::ShaderError {
...
+
+    let mut map = auxil::FastHashMap::default();
+
+    let inputs = ast.get_shader_resources().map_err(gen_query_error)?.stage_inputs;
+    for input in inputs {
+        let idx = ast.get_decoration(input.id, spirv::Decoration::Location).map_err(gen_query_error)?;
+
+        let ty = ast.get_type(input.type_id).map_err(gen_query_error)?;
+
+        match ty {
```
kvark, line 59: can we do `let count = match ty {...};` so that we can follow with a single loop of `map.insert` over that count?

src/backend/dx12/src/device.rs
```diff
@@ -1877,10 +1938,34 @@ impl d::Device<B> for Device {
             }
         }
 
+        // See [`introspect_spirv_vertex_semantic_remapping`] for details of why this is needed.
+        let semantics: Vec<_> = attributes.iter().map(|attrib| {
```
kvark, line 1942: similarly here, why don't we compute this from within the other loop, instead of collecting into `semantics` vector?

kvark:
Agreed to proceed as-is on matrix.
bors r+

PR #3447, opened 2020-10-29
dx12: fix entry point selection

msiglreith:
SPIRV-Cross translates the selected entrypoint to `main` always.
`.find(|entry_point| entry_point.name == real_name)` will fail for non-`main` entrypoints of the source spirv module. Actually we don't need to query for all the entry points as we already selected the one we want.

Tested it by modfiying one of the spirv modules in the quad examples.

PR checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### backend-dx12-unreleased
   - fix matrix vertex inputs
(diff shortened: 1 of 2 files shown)


kvark (approved):
Nice, thank you!
bors r+

kvark:
@msiglreith can we get that into `master` as well, please?

PR #3448, opened 2020-10-30
[0.6/dx] Assorted Fixes to Make BVE-Reborn Work Under DX11

cwfitzgerald:
Fixes https://github.com/gfx-rs/wgpu/issues/1005

PR is commit by commit etc.

Notes:
 - I have disabled `MAILBOX` right now as it was acting like FIFO. Until we can figure out a way to make it work without limiting frame rate, we can re-enable.
 - An extension of the swapchain changes will be enabling the FLIP modes on windows 8 or newer. This will also allow us to bring back ALLOW_TEARING.
 - The spirv-cross workaround had a bug where there is a panic while there were more vertex attributes than inputs to the shader.
 - I have included changes that reduce the amount of clones we make to `ImageView`s where we didn't actually need them.
 - I have chosen this method of doing borrowed image views (forgetting smart pointers) as removing the smart pointers will make a lot of code more complicated (due to helpers like `clone`, `cast`, direct method calls, etc) than raw pointers. I am not inherently opposed to switching to raw pointers, but thought this was the easier solution.

Have a couple DX12 bugs to attend to following this, so not bumping version on that yet.

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -1792,7 +1792,7 @@ impl device::Device<Backend> for Device {
                     set_debug_name_with_suffix(&srv, name, " -- SRV");
                 }
 
-                Some(srv)
+                Some(srv.into_raw())
             } else {
                 None
             },
@@ -1803,7 +1803,7 @@ impl device::Device<Backend> for Device {
                     set_debug_name_with_suffix(&rtv, name, " -- RTV");
                 }
 
-                Some(rtv)
+                Some(rtv.into_raw())
             } else {
                 None
             },
@@ -1814,7 +1814,7 @@ impl device::Device<Backend> for Device {
                     set_debug_name_with_suffix(&uav, name, " -- UAV");
                 }
 
-                Some(uav)
+                Some(uav.into_raw())
             } else {
                 None
             },
@@ -1825,7 +1825,7 @@ impl device::Device<Backend> for Device {
                     set_debug_name_with_suffix(&dsv, name, " -- DSV");
                 }
 
-                Some(dsv)
+                Some(dsv.into_raw())
             } else {
                 None
             },
@@ -1837,10 +1837,11 @@ impl device::Device<Backend> for Device {
                     set_debug_name_with_suffix(&rodsv, name, " -- DSV");
                 }
 
-                Some(rodsv)
+                Some(rodsv.into_raw())
             } else {
                 None
             },
+            owned: true,
         })
     }
 
@@ -2015,12 +2016,10 @@ impl device::Device<Backend> for Device {
                         c: ptr::null_mut(),
                         t: image
                             .srv_handle
-                            .clone()
-                            .map_or(ptr::null_mut(), |h| h.as_raw() as *mut _),
+                            .map_or(ptr::null_mut(), |h| h as *mut _),
                         u: image
                             .uav_handle
-                            .clone()
-                            .map_or(ptr::null_mut(), |h| h.as_raw() as *mut _),
+                            .map_or(ptr::null_mut(), |h| h as *mut _),
                         s: ptr::null_mut(),
                     },
                     pso::Descriptor::Sampler(sampler) => RegisterData {
@@ -2034,12 +2033,10 @@ impl device::Device<Backend> for Device {
                             c: ptr::null_mut(),
                             t: image
                                 .srv_handle
-                                .clone()
-                                .map_or(ptr::null_mut(), |h| h.as_raw() as *mut _),
+                                .map_or(ptr::null_mut(), |h| h as *mut _),
                             u: image
                                 .uav_handle
-                                .clone()
-                                .map_or(ptr::null_mut(), |h| h.as_raw() as *mut _),
+                                .map_or(ptr::null_mut(), |h| h as *mut _),
                             s: sampler.sampler_handle.as_raw() as *mut _,
                         }
                     }
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -456,11 +456,24 @@ impl Device {
                     ArraySize,
                 }
             }
+            image::ViewKind::D2 if info.kind.num_samples() > 1 => {
+                desc.ViewDimension = d3dcommon::D3D11_SRV_DIMENSION_TEXTURE2DMS;
+                *unsafe { desc.u.Texture2DMS_mut() } = d3d11::D3D11_TEX2DMS_SRV {
+                    UnusedField_NothingToDefine: 0,
+                }
+            }
             image::ViewKind::D2 => {
                 desc.ViewDimension = d3dcommon::D3D11_SRV_DIMENSION_TEXTURE2D;
                 *unsafe { desc.u.Texture2D_mut() } = d3d11::D3D11_TEX2D_SRV {
                     MostDetailedMip,
                     MipLevels,
+                }
+            }
+            image::ViewKind::D2Array if info.kind.num_samples() > 1 => {
+                desc.ViewDimension = d3dcommon::D3D11_SRV_DIMENSION_TEXTURE2DMSARRAY;
+                *unsafe { desc.u.Texture2DMSArray_mut() } = d3d11::D3D11_TEX2DMS_ARRAY_SRV {
+                    FirstArraySlice,
+                    ArraySize,
                 }
             }
             image::ViewKind::D2Array => {
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -185,13 +185,13 @@ impl Device {
 
         // See [`shader::introspect_spirv_vertex_semantic_remapping`] for details of why this is needed.
         let semantics: Vec<_> = attributes.iter().map(|attrib| {
-            match vertex_semantic_remapping.get(&attrib.location).unwrap() {
-                Some((major, minor)) => {
+            match vertex_semantic_remapping.get(&attrib.location) {
+                Some(Some((major, minor))) => {
                     let name = std::borrow::Cow::Owned(format!("TEXCOORD{}_\0", major));
                     let location = *minor;
                     (name, location)
                 }
-                None => {
+                _ => {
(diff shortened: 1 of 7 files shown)


src/backend/dx11/src/device.rs
```diff
@@ -185,13 +185,13 @@ impl Device {
 
         // See [`shader::introspect_spirv_vertex_semantic_remapping`] for details of why this is needed.
         let semantics: Vec<_> = attributes.iter().map(|attrib| {
-            match vertex_semantic_remapping.get(&attrib.location).unwrap() {
```
kvark, line 188: in what case would the remapping *not* contain the attrib location? It seems like `unwrap()` could be well suited here

src/backend/dx11/src/device.rs
```diff
@@ -457,19 +457,34 @@ impl Device {
                 }
             }
             image::ViewKind::D2 => {
-                desc.ViewDimension = d3dcommon::D3D11_SRV_DIMENSION_TEXTURE2D;
-                *unsafe { desc.u.Texture2D_mut() } = d3d11::D3D11_TEX2D_SRV {
-                    MostDetailedMip,
-                    MipLevels,
+                if info.kind.num_samples() > 1 {
```
kvark, line 460: we should probably lift that condition to the match arm, i.e.
```rust
image::ViewKind::D2 if info.kind.num_samples() > 1 => {...}
image::ViewKind::D2 => {...}
```
The benefit is that all match arms become very similar and nested in the same way

src/backend/dx11/src/device.rs
```diff
@@ -776,11 +791,7 @@ impl Device {
             Windowed: TRUE,
             // TODO:
             SwapEffect: dxgi::DXGI_SWAP_EFFECT_DISCARD,
-            Flags: if config.present_mode == window::PresentMode::IMMEDIATE {
-                dxgi::DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING
```
kvark, line 780: why are we no longer allowing tearing?

src/backend/dx11/src/lib.rs
```diff
@@ -3421,6 +3423,51 @@ pub struct ImageView {
...
+    }
+}
+
+impl ImageView {
+    fn clone_borrowed(&self) -> Self {
+        unsafe {
+                Self {
+                subresource: self.subresource,
+                format: self.format,
+                rtv_handle: self.rtv_handle.as_ref().map(|p| ComPtr::from_raw(p.as_raw())),
```
kvark, line 3450: I don't understand why we are doing this? Looks like we have a `ComPtr` and we are constructing a new `ComPtr` from the same pointer without actually cloning?

src/backend/dx11/src/lib.rs
```diff
@@ -3421,6 +3423,51 @@ pub struct ImageView {
...
+                borrowed: true
+            }
+        }
+    }
+}
+
+impl Drop for ImageView {
+    fn drop(&mut self) {
+        if self.borrowed {
+            mem::forget(self.rtv_handle.take());
```
kvark, line 3464: ok, it sounds very much like we should just got with raw pointers
If they are too annoying to deal with, we can define a wrapper `struct RawPtr<T>(*mut T)` or something like that and implement the necessary properties there

src/backend/dx11/src/lib.rs
```diff
@@ -3869,25 +3916,14 @@ impl From<pso::DescriptorType> for DescriptorContent {
                 ty: Bdt::Uniform, ..
             } => DescriptorContent::CBV,
             Dt::Buffer {
-                ty: Bdt::Storage { read_only: true },
```
kvark, line 3872: can you remind me, what's the problem with binding read-only storage as SRV?
I thought your issue was only the case where we wanted to bind the same resource as both UAV and SRV. But in this case, it would be either one but not both.

src/backend/dx12/src/device.rs
```diff
@@ -1943,17 +1943,15 @@ impl d::Device<B> for Device {
             let semantics = vertex_semantic_remapping
```
kvark, line 1943: similar question here - why would we expect `vertex_semantic_remapping` to not have our attribute location?

src/backend/dx11/src/device.rs
cwfitzgerald: That flag doesn't apply to `DXGI_SWAP_EFFECT_*` swap effects, only `DXGI_SWAP_EFFECT_FLIP*`. If used with non-flip effects, raises INVALID_ARG. This was the cause of gfx-rs/wgpu#1005

src/backend/dx11/src/lib.rs
cwfitzgerald: Alright, sounds reasonable.

src/backend/dx11/src/device.rs
cwfitzgerald: If a pipeline describes `0 => Float4, 1 => Float4`, but then the vertex shader only has:
```glsl
layout(location = 0) in vec4 input;
```
then there is no entry for location 1, causing a panic.

kvark: ok, sounds good

src/backend/dx11/src/lib.rs
kvark: I think there is a better solution hiding here. We have a place where we connect the shader AST with the binding layout, it's called `patch_spirv_resources` in DX11 backend. There, if we see that the shader has `NonWritable` but the layout isn't read-only, we could just remove the `NonWritable` declaration. The benefit of doing so is that we'd still go a more efficient SRV path for truly read-only bindings, which WebGPU would use.

src/backend/dx11/src/shader.rs
```diff
@@ -239,18 +239,29 @@ fn patch_spirv_resources(
...
+                ..
+            } => {
+                read_only
+            }
+            _ => unreachable!()
+        };
+
+        // If the binding is read/write, we need to generate a UAV here.
+        if !read_only {
+            ast.set_member_decoration(storage_buffer.type_id, 0, spirv::Decoration::NonWritable, 0)
```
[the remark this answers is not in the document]
cwfitzgerald: As far as I can tell I need to treat this like a bool, so to disable it I set its value to 0.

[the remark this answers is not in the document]
kvark: You are correct!

cwfitzgerald:
Alrighty, should be good to go here, confirmed it still works.

src/backend/dx11/src/lib.rs
```diff
@@ -3412,15 +3409,82 @@ impl Image {
...
+    uav_handle: Option<*mut d3d11::ID3D11UnorderedAccessView>,
+    borrowed: bool,
+}
+
+impl Clone for ImageView {
+    fn clone(&self) -> Self {
+        Self {
+            subresource: self.subresource,
+            format: self.format,
+            rtv_handle: self.rtv_handle.as_ref().map(|&p| {
```
kvark, line 3428: Why are we adding the references here? Or rather, the question is, why would we ever want to clone it while bumping the refcounts? Perhaps, there is only one clone implementation that needs to exist

src/backend/dx11/src/lib.rs
```diff
@@ -3412,15 +3409,82 @@ impl Image {
...
-    srv_handle: Option<ComPtr<d3d11::ID3D11ShaderResourceView>>,
-    dsv_handle: Option<ComPtr<d3d11::ID3D11DepthStencilView>>,
-    rodsv_handle: Option<ComPtr<d3d11::ID3D11DepthStencilView>>,
-    uav_handle: Option<ComPtr<d3d11::ID3D11UnorderedAccessView>>,
+    rtv_handle: Option<*mut d3d11::ID3D11RenderTargetView>,
+    srv_handle: Option<*mut d3d11::ID3D11ShaderResourceView>,
+    dsv_handle: Option<*mut d3d11::ID3D11DepthStencilView>,
+    rodsv_handle: Option<*mut d3d11::ID3D11DepthStencilView>,
+    uav_handle: Option<*mut d3d11::ID3D11UnorderedAccessView>,
+    borrowed: bool,
```
kvark, line 3420: nit: perhaps turning this around and making this an `owned` flag would make sense, since it has additional semantics of acquiring/releasing ownership

src/backend/dx11/src/lib.rs
cwfitzgerald: For a similar reason as why `Cow` turns owned when you clone it and `ComPtr` des; I want the default action to take references, so we don't end up with dangling pointers if at all possible. Additionally, if we do own it, we need to clone to get another owned `ImageView`.

That is a long-winded way of saying: Clone not taking references would be surprising, and I don't want to be surprising with the potential for dangling references.

kvark: These are raw pointers (unlike Cow/ComPtr). When you Clone or Copy a raw pointer, you don't get surprised. You can implement this as just a `copy_raw()` method instead of `impl Clone` if that makes it cleaner.
The main question still stands - why would we need both borrowing copy and owning copy?

cwfitzgerald: Let me see what happens if I make clone only borrowing and if nothing explodes that'll be fine with me.

cwfitzgerald:
Both suggestions implemented :)

src/backend/dx11/src/lib.rs
```diff
@@ -3412,15 +3409,52 @@ impl Image {
...
+    fn clone(&self) -> Self {
+        Self {
+            subresource: self.subresource,
+            format: self.format,
+            rtv_handle: self.rtv_handle.clone(),
+            srv_handle: self.srv_handle.clone(),
+            dsv_handle: self.dsv_handle.clone(),
+            rodsv_handle: self.rodsv_handle.clone(),
+            uav_handle: self.uav_handle.clone(),
+            owned: true
```
kvark, line 3433: this should be `false`

kvark (commented):
Beautiful! Just one last fix is needed!

kvark (approved):
bors r+

cwfitzgerald:
arg, hate random failures

bors retry

PR #3449, opened 2020-10-30
[0.6/dx12] Fix Heap Selection for Non-Renderable Textures

cwfitzgerald:
This fixes the issues with BC7 images being created on the wrong heap. This only moves to the target heap if the DX12 image flags actually demand it be there.


kvark (approved):
great, thank you!
bors r+

Issue #2503, opened 2018-12-08
dx12: Resource transition


msiglreith:
For exclusive resource we require to issue pipeline barriers for transfer the ownership onto different queues.
In dx12 only the queue type is relevant (Graphics/Compute vs Copy), in between no resource state transition is required.

As first implementation step we should relax the current implementation to treat the cases of transition between the same queue type. Requires a bit of logic regarding the queue family id but nothing too fancy.

The next step would be to analyze the requirements on vulkan side and see how it interacts with dx12. In the ideal case it would fit within our current dx12 barrier bounds and we basically have 'free' resource transitions. Otherwise we require additional resource state transitions to deal with the Copy <-> Graphics/Compute transitions.

kvark:
It looks like we are at the stage where specific image layouts map OK to D3D12 resource states. There are some subtle complexities with copies and resolves, but they seem to be under control.

There is a big difference between the `image::Layout::General` layout and [D3D12_RESOURCE_STATE_COMMON](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_states), which can be considered to be the corresponding state in DX12:

> VK_IMAGE_LAYOUT_GENERAL supports all types of device access.

In contrast, DX12 requires a resource to be in a specific state for each usage.

### Rough Proposal

Always map `General` to `COMMON`. When recording command buffers, always transition the state to a specific state, do operation, and then transition back to `COMMON`. This is unfortunate, but well within the expected bahavior. It's good that many functions accept a list of jobs (i.e. multiple blits, multiple flushes, etc), so that we can combine the transitions together between multiple resources.

PR #3451, opened 2020-11-01
[dx12] refactor resource transitions, exclude swapchain images

kvark:
Functionally, the biggest change here is excluding swapchain images from the explicit pipeline barriers.
Plus, a whole bag of small refactor.
Note: this is related to #2503 but doesn't actually solve it.


msiglreith (commented):
bors r+

PR #3452, opened 2020-11-01
[0.6/dx11] Fix depth to depth copies

cwfitzgerald:
This fixes depth-stencil copies on DX11. It also adds some asserts so unsupported copies properly blows up instead of silently dropping calls. This will also tell us when we may need to add some extra compute copies for things that can be copied between in vk but not DX11.

### DIFF
--- a/src/backend/dx11/src/internal.rs
+++ b/src/backend/dx11/src/internal.rs
@@ -708,7 +708,45 @@ impl Internal {
         } else {
             // Default copy path
             for region in regions.into_iter() {
-                let info = region.borrow();
+                let info: &command::ImageCopy = region.borrow();
+
+                assert_eq!(
+                    src.decomposed_format.typeless,
+                    dst.decomposed_format.typeless,
+                    "DX11 backend cannot copy between underlying image formats: {} to {}.",
+                    src.decomposed_format.typeless,
+                    dst.decomposed_format.typeless,
+                );
+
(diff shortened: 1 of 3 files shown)


src/backend/dx11/src/internal.rs
```diff
@@ -708,7 +708,33 @@ impl Internal {
...
+                    src.decomposed_format.typeless,
+                    dst.decomposed_format.typeless,
+                );
+
+                let src_depth_stencil = src.format.is_depth() || src.format.is_stencil();
+                let dst_depth_stencil = dst.format.is_depth() || dst.format.is_stencil();
+
+                assert_eq!(src_depth_stencil, dst_depth_stencil, "Both images in a copy must be depth stencil textures");
+
+                let copy_box = if src_depth_stencil || dst_depth_stencil {
```
kvark, line 726: since they are equal, no need to "||" both

kvark, line 726: It would be great to issue a warning here if the given `src_offset and `extent` don't cover the whole image. Otherwise, somebody in the future may waste time trying to figure out why it's not respected for the copy.

kvark (commented):
Nice find! A few minor things

cwfitzgerald:
Updated, also added check for multisampled resources as they also need full resource only copies.

kvark (approved):
Nice, thank you!
bors r+

cwfitzgerald:
bors retry

PR #3456, opened 2020-11-02
[0.6/dx11] Assorted Fixes

cwfitzgerald:
These are fixes exposed by a user's project.

Fixes:
 - Treat push constant offset as bytes not words
 - Properly find register numbers for UAVs in non-fragment shaders.

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -2858,7 +2858,7 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
         offset: u32,
         constants: &[u32],
     ) {
-        let start = offset as usize;
+        let start = (offset / 4) as usize;
         let end = start + constants.len();
 
         self.push_constant_data[start..end].copy_from_slice(constants);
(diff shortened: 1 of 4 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -3748,6 +3748,36 @@ impl DescriptorSetInfo {
...
+            stage
+        } else {
+            ShaderStage::Fragment
+        };
+
+        let mut res_offsets = self
+            .registers
+            .map_register(|info| info.res_index as DescriptorIndex)
+            .select(register_stage);
+        for binding in self.bindings.iter() {
```
kvark, line 3768: nit: if `self.bindings` is sorted by binding index, we could use binary search to find the match, and then check the `UAV` flag.

kvark (approved):
Nice work!
You are setting a record on how much a released backend can be patched :)
bors r+

src/backend/dx11/src/lib.rs
cwfitzgerald: That wouldn't work as we need to accumulate through all UAV bindings to get the proper offset.

kvark: Hmm. That mapping (between bindings and resource indices) is precisely the responsibility of a pipeline layout. So I'm surprised that we have to recompute it here.

PR #3458, opened 2020-11-02
[0.6/dx11] Fix buffer to image copies of textures of <4 bytes/pixel.

cwfitzgerald:
We had some nasty corruption as a result of alignment throwing some calculations off. This fixes the copies for all copies to textures that have less than 4 bytes per pixel.

As a future extension, a lot of the copies in that file aren't using full warps, so adjusting them to use full warps likely will be useful.

### DIFF
--- a/src/backend/dx11/shaders/copy.hlsl
+++ b/src/backend/dx11/shaders/copy.hlsl
@@ -456,8 +456,14 @@ void cs_copy_buffer_image2d_r16(uint3 dispatch_thread_id : SV_DispatchThreadID) 
     uint src_idx = GetBufferSrc16(dispatch_thread_id);
     uint2 data = Uint32ToUint16x2(BufferCopySrc.Load(src_idx));
 
-    Image2CopyDstR[dst_idx                 ] = data.x;
-    Image2CopyDstR[dst_idx + uint3(1, 0, 0)] = data.y;
+    uint remaining_x = bounds.x - dst_idx.x;
+
+    if (remaining_x >= 2) {
+        Image2CopyDstR[dst_idx + uint3(1, 0, 0)] = data.y;
+    } 
+    if (remaining_x >= 1) {
+        Image2CopyDstR[dst_idx + uint3(0, 0, 0)] = data.x;
+    }
 }
 
 [numthreads(COPY_2D_NUM_THREAD_X, COPY_2D_NUM_THREAD_Y, 1)]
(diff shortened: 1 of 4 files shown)


src/backend/dx11/shaders/copy.hlsl
```diff
@@ -552,10 +599,28 @@ void cs_copy_buffer_image2d_r8(uint3 dispatch_thread_id : SV_DispatchThreadID) {
...
-    Image2CopyDstR[dst_idx + uint3(0, 0, 0)] = data.x;
-    Image2CopyDstR[dst_idx + uint3(1, 0, 0)] = data.y;
-    Image2CopyDstR[dst_idx + uint3(2, 0, 0)] = data.z;
-    Image2CopyDstR[dst_idx + uint3(3, 0, 0)] = data.w;
+    uint remaining_x = bounds.x - dst_idx.x;
+
+    switch (remaining_x) {
+        default:
+            Image2CopyDstR[dst_idx + uint3(0, 0, 0)] = data.x;
+            Image2CopyDstR[dst_idx + uint3(1, 0, 0)] = data.y;
```
kvark, line 607: would it be simpler to avoid repeating ourselves and just adding a condition for each of the assignments?
```cpp
if (remaining_x > 0) {
  Image2CopyDstR[dst_idx + uint3(0, 0, 0)] = data.x;
}
if (remaining_x > 1) {
  Image2CopyDstR[dst_idx + uint3(1, 0, 0)] = data.y;
}
...
```
We could even just loop there

src/backend/dx11/src/lib.rs
```diff
@@ -294,7 +294,7 @@ fn get_limits(feature_level: d3dcommon::D3D_FEATURE_LEVEL) -> hal::Limits {
         non_coherent_atom_size: 1, // TODO
         max_sampler_anisotropy: 16.,
         optimal_buffer_copy_offset_alignment: 1, // TODO
-        optimal_buffer_copy_pitch_alignment: 1,  // TODO
+        optimal_buffer_copy_pitch_alignment: 4,
```
kvark, line 297: maybe we could have a small comment explaining why it's 4?

src/backend/dx11/shaders/copy.hlsl
cwfitzgerald: I don't think we can loop due to data being in a uvec4 (now yes we could put it in an array, but meh). If statements though should actually be a bit more coherent, so it should be a win.

kvark: you should be able to index `uvec4` as an array, no need to put it out

kvark (approved):
Thank you!
bors r+

PR #3460, opened 2020-11-03
Add Cargo Deny Configuration

zicklag:
Fixes #issue
PR checklist:

---

@kvark I'll add this to CI in a second, but it won't actually deny duplicate dependencies it will just display them in the log and deny crates with unallowed licenses. Is how we'd want it?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -41,6 +41,8 @@ jobs:
   build:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    # Prevent sudden announcement of a new advisory from failing ci:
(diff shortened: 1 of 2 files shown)


kvark (approved):
Let's see if this proves useful :)
bors r+

zicklag:
Bummer, looks like the GitHub action only runs on Linux. That should be fine for now I think, I'll fix it.

zicklag:
Hmm, what happened? Bors got stuck or something it looks like.

kvark:
bors retry

grovesNL:
I think bors isn't able to merge certain changes to GitHub Actions, so that's probably why it's failing

PR #3461, opened 2020-11-03
[0.6/dx12] fix respecting the read-only storage flags

kvark:
Fixes https://github.com/gfx-rs/wgpu/issues/1013
It could be a regression from spirv-cross update, although the decoration fixes imply that it shouldn't have ever worked 🤔 
PR checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,15 @@
 # Change Log
 
-### backend-gl-0.6.0 (2-11-2020)
-  - re-enable the GL backend for Linux, Mac, and Web
-  - migrate GL backend to use context sharing in `surfman`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,14 +1,15 @@
 # Change Log
 
+### backend-dx12-0.6.12 (2-11-2020)
```
cwfitzgerald, line 3: nit: missing leading zero in day

cwfitzgerald (approved):
Like all good bugfixes, it uncovered some more bugs, but https://github.com/gfx-rs/wgpu/issues/1013 no longer shows up, so looks good to me! One minor nit.

kvark:
bors r=cwfitzgerald

PR #3462, opened 2020-11-03
[dx12] fix the respect of readonly storage flags

kvark:
Sibling of  #3461 for master


kvark:
bors r=cwfitzgerald

kvark:
bors retry

PR #3463, opened 2020-11-03
[0.6/gl] avoid wait sync if it's not available

grovesNL:
glWaitSync isn't always available (e.g. on the web), so we could probably skip it when it's not supported for now.

### DIFF
--- a/src/backend/gl/src/queue.rs
+++ b/src/backend/gl/src/queue.rs
@@ -187,8 +187,10 @@ impl CommandQueue {
         let extent = swapchain.extent;
 
         // Wait for rendering to finish
-        unsafe {
-            gl.wait_sync(self.presentation_fence.unwrap(), 0, glow::TIMEOUT_IGNORED);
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/queue.rs
```diff
@@ -187,8 +187,10 @@ impl CommandQueue {
         let extent = swapchain.extent;
 
         // Wait for rendering to finish
-        unsafe {
-            gl.wait_sync(self.presentation_fence.unwrap(), 0, glow::TIMEOUT_IGNORED);
+        if self.share.private_caps.sync {
```
kvark, line 190: can we refactor this a bit in the following way: if `private_caps.sync` is `false`, we should not create the `presentation_fence`, hence leaving it `None`. Then here we'd just do `if let Some(fence) = self.presentation_fence { ... }`

kvark (approved):
Thank you!
bors r+

PR #3466, opened 2020-11-05
[0.6/GL] descriptor write fixes with version bump

kvark:
Makes us run the `cube` and `mipmap` examples somewhat...
cc @zicklag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+### backend-gl-0.6.1 (05-11-2020)
+  - fix descriptor writes
+  - fall back from persistent to emulated mapping
+  - support multi-sampled renderbuffers, multi-level views
+
 ### backend-dx12-0.6.13 (03-11-2020)
   - fix panic when creating unused multi-mip storage view
 
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gfx-backend-gl"
-version = "0.6.0"
+version = "0.6.1"
 description = "OpenGL backend for gfx-rs"
 homepage = "https://github.com/gfx-rs/gfx"
 repository = "https://github.com/gfx-rs/gfx"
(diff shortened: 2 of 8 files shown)


zicklag:
Cube runs for me now:

image

Still get the artifacts, but at this point maybe don't take my machine as a reference. I'm not the admin on my system so I can't change this at the moment, but I have to run the demo in a container because my host is missing some package so that it complains that the `egl function could not be loaded`. Also I'm running a hybrid GPU setup, so programs run on the CPU with a Mesa driver by default, but can optionally be started with the full Nvidia GPU. But I don't know that I can get the Nvidia GPU to work inside of the container so I can only really test how it runs with the Mesa driver right now, assuming I'm understanding the situation properly.

Anyway, looks like the fixes helped!

kvark:
@zicklag ok, good, so at least we can reproduce the issue again.
On my macOS, the behavior of presentation is very strange: image shows up for a few seconds, then the screen is black for another few seconds, then it shows up again. It's strangely stable, I'm just confused to see the change at the second-scale as opposed to frame-scale or something.

zicklag:
Ooh, mipmap does work too. Still with the jagged edges, but cool!

image

kvark:
Do you want to review the PR btw? :)

zicklag:
> Do you want to review the PR btw? :)

No, but thanks. :) I've got some other stuff I've got to focus on now, but it was great to be able to help with all of this. Hopefully with OpenGL all connected to WGPU now it will be a lot easier for others to get into contributing and testing the GL backend. Can't wait to see how it progresses!

kvark:
Could you file a bug about the jagged edges, specifying your configuration clearly? I know we had that in a few places, but these discussions were so long it's now difficult to recap.

zicklag:
Would it be appropriate just to update the original post for this issue: https://github.com/gfx-rs/gfx/issues/3353?

kvark:
Actually, yes, let's just work with #3353
Just to double-check, you weren't able to reproduce the issue on the quad example? I wonder if the clearing color takes part in this. Could you try changing wgpu-rs example to clear to a simple color (like black)? Could you try changing the quad example to clear to the same color as the cube does?

grovesNL (approved):
bors r+

Issue #3469, opened 2020-11-07
Copy on present in OpenGL backend


kvark:
Currently, the GL backend is built in such a way that we copy the full frame on present(). This is a waste of fill-rate and performance that we've been trying to eliminate. The plan was to modify `surfman` in a mystical way that would let us do this - https://github.com/servo/surfman/issues/202. However, digging into EGL as well as asking around on Khronos slack proved this to be impossible. The best we can do is - avoid context sharing in GL. So the problem of copy on present remains unsolved.

I've been thinking that perhaps in most cases where the swapchain is used, there should be no need for a copy at all. Most applications would render to the swapchain image in a single pass, and the depth is discarded after it. So what if we create the EGL surface with all of color, depth, stencil, just in case, and try to use it aggressively? Given the render pass boundaries in this API, it seems like it could cover most cases:

### Cases

Case A: rendering to color only.
GL backend sees that an application only renders to the color, it binds the main framebuffer, and the copy is avoided.

Case B: rendering to color and depth, the depth is cleared and discarded at the end.
We see the clear+discard on the depth, and we figure out that we don't actually need to consider the user-provided depth image. So we just use the default framebuffer again, and the copy is avoided.

Case C: rendering to color and depth, depth is test-only, loaded from a depth pre-pass.
We see that the depth is discarded, so we can use the default framebuffer. However, we also see that the depth is loaded, so we issue a copy from the user-provided depth into the main FBO depth, first, before starting a pass.

Case D: rendering to color and depth, and depth is stored (not discarded).
We proceed with the logic of B/C, but after the pass we issue a copy from the framebuffer depth to the user-provided depth.

Case X: rendering with other colors alongside the swapchain. I don't see how we can reasonably support that in GL backend. We'll have to document and assert on that.

Case Y: user-provided depth is not a 24-bit depth. I don't think we can deal with that, unfortunately.

### Analysis

This requires some logic to be implemented, for tight integration between EGL and rendering. The benefit is avoiding the copy of color in all cases. However, another cost is that some cases would involve a copy of the depth, which roughly have the same size...

PR #3470, opened 2020-11-09
[gl] replace Surfman by EGL

### LINKED ISSUE #3468: Idiomatic OpenGL context directly on EGL
## Intro

Looking more at our current issues with GL presentation (#3353), at `surfman` code (https://github.com/servo/surfman/issues/202), and around, I think we need to approach this from the ground up again.

First of all, `surfman` (used in #3375) brings a lot of friction:
  - it's a big 3rd party dependency
  - it has a questionable code base (strange choice of features, many dependencies, hugely outdated code style)
  - assumptions don't always match (#3467)
  - close to be unmaintained now
  - needs to be rewritten largely, due to https://github.com/servo/surfman/issues/202

We contributed a bit to it, but it takes a lot of effort to truly become fluent enough to land the changes we need. But we know what OpenGL backend of gfx-rs needs. What if we could reduce the scope of work here?

On Windows, we have Vulkan, D3D12 and D3D11. There is no need for OpenGL there. On macOS, OpenGL has been deprecated for a while already, and Metal is widely supported. This leaves us with only non-Apple unix systems, including Android. All of them, if I understand correctly, are designed to support EGL as the main context/window manager for GL.

## Proposal

Therefore, what if we can just directly depend on EGL unconditionally (on non-wasm targets)? No need for optional `glutin`, `surfman`, etc, features. Just make it so our GL backend uses EGL for context creation.

- `Instance` = `egl::Display`, created by `eglGetDisplay`. We could choose the default one - I think this is the most tricky part in the proposal so far.
- `Surface` = `egl::Surface`, created by `eglCreateWindowSurface`
- `Adapter` = `egl::Config`. When `enumerate_adapters` is called, we could get all the contexts supported by the display, and then query their attributes to find out which are software renderers, which support presentation, etc
- `Device` = `egl::Context`, created by `eglCreateContext`.

Note: this elaboration doesn't include WebGL target. I don't know how to properly fit it in here, just yet.

kvark:
Fixes #3468

This is a BIG change. It re-architects WSI of the GL backend to unconditionally use EGL (via `khronos-egl` crate atm, but we can change that). This means the backend is *only* supported on Linux/BSD/Android, it's no longer supported on Apple and Microsoft platforms.

One of the consequences - we throw Surfman away. There was too much complication and too little benefit from it anyway.

Another consequence - we are locked to GLES, given that I experienced issues trying to get a desktop GL context via EGL. We can revisit this, but really I think going with GLES 3.1 makes total sense for the backend, unconditionally. if the target platform is powerful above what GLES 3.1 offers, it should just support Vulkan.

Most importantly, it solves GL context sharing, so presentation is much more robust. However, it doesn't solve the extra copy - #3469. For this, we'll need to follow-up by directly using platform extensions, such as [EGL_WL_create_wayland_buffer_from_image](https://www.khronos.org/registry/EGL/extensions/WL/EGL_WL_create_wayland_buffer_from_image.txt).

PR checklist:
- [x] tested examples with the following backends: Linux


grovesNL (approved):
This looks great, thank you!

I think this is a good direction :+1: I think we can probably use a similar path for WebGL by creating contexts ahead of time and later reusing them.

Only supporting OpenGL ES sounds fine for now, we can always look into the issues with desktop OpenGL later if it's needed by anyone.

bors r+

grovesNL:
Oh I didn't notice the CI failures, we need to add the egl package on Linux and disable gl on macOS.

I'm not sure about the Android failure but maybe we could disable CI on gl there temporarily:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CrossCompilation', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/khronos-egl-2.1.1/build.rs:7:10

https://github.com/gfx-rs/gfx/pull/3470/checks?check_run_id=1372587404#step:5:83

kvark:
Filed https://github.com/timothee-haudebourg/khronos-egl/issues/6, will disable Android here for now

VincentFTS:
> This looks great, thank you!
> 
> I think this is a good direction 👍 I think we can probably use a similar path for WebGL by creating contexts ahead of time and later reusing them.
> 
> Only supporting OpenGL ES sounds fine for now, we can always look into the issues with desktop OpenGL later if it's needed by anyone.
> 
> bors r+

It would be very useful here to retrieve WebGL support. I have no dev background in neither WebGL nor OpenGL, but I may help somehow.

kvark:
We are definitely going to keep WebGL support, although I haven't tested it in a while. You can try following the instructions in ccf789ddcb45dd528ddef83e164d82a5788ad4ce on master to see where we stand now? I hope it's not broken, or otherwise it should have been caught by CI.

VincentFTS:
gfx-backend-gl doesn’t build with wasm32-unknown-unknown target since `Surface` struct doesn’t implement `present` method called [here](https://github.com/gfx-rs/gfx/blob/c726674506d4255fd447d367dfe04a7d06b5de9f/src/backend/gl/src/queue.rs#L1122)

kvark:
Ah ok. So you can definitely help by fixing this and re-enabling the WASM build on CI!

PR #3465, opened 2020-11-03
[0.6/dx12] Fix Panic When Creating Texture View

cwfitzgerald:
Fixes last issue preventing bve-reborn from running on DX12.

I'll forward port my pending changes after this merges.

PR #3472, opened 2020-11-09
[dx/master] Rollup of DX Fixes

cwfitzgerald:
Got a little behind here; rolls up #3444 #3446 #3448 #3452 #3456 #3458 and #3465. 

#3449 seems to have been obsoleted by a previous change in master: https://github.com/gfx-rs/gfx/blame/fa323527758d35a3913815e62bcb0e08511f7384/src/backend/dx12/src/device.rs#L2443-L2449. Could you just validate that this change does indeed serve the same purpose?

### DIFF
--- a/src/backend/dx11/shaders/copy.hlsl
+++ b/src/backend/dx11/shaders/copy.hlsl
@@ -456,8 +456,14 @@ void cs_copy_buffer_image2d_r16(uint3 dispatch_thread_id : SV_DispatchThreadID) 
     uint src_idx = GetBufferSrc16(dispatch_thread_id);
     uint2 data = Uint32ToUint16x2(BufferCopySrc.Load(src_idx));
 
-    Image2CopyDstR[dst_idx                 ] = data.x;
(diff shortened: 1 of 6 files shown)


kvark:
thats... an impressive list!
bors r+

kvark:
> Could you just validate that this change does indeed serve the same purpose?

the blame points to lines from 2017, so looks like the link is wrong

kvark (approved):

kvark:
bors retry

cwfitzgerald:
> 
> 
> > Could you just validate that this change does indeed serve the same purpose?
> 
> the blame points to lines from 2017, so looks like the link is wrong

It was the line in the middle from two months ago: more precise blame: https://github.com/gfx-rs/gfx/blame/fa323527758d35a3913815e62bcb0e08511f7384/src/backend/dx12/src/device.rs#L2445

kvark:
My change didn't check the target format features, so yours seems to still be needed.

cwfitzgerald:
Alright, I'll follow up with a proper merge of that, need to actually think about it.

PR #3473, opened 2020-11-09
Fix cargo-deny checks

kvark:
Fixes a fallout of #3460, and also unconfuses bors...


kvark:
bors r+
hold on poor GH Action bot, we are on the way to save you!

kvark:
Going to do a bit of force-pushing here, since GHA doesn't appear to run this change for some reason.

PR #3475, opened 2020-11-14
re-enable WebGL support

MarcAntoine-Arnaud:
Mentionned in https://github.com/gfx-rs/gfx/pull/3470
Maybe help issue #2749

PR checklist:
- [x] re-enable WebGL rendering
- [x] use published version of `spirv_cross` dependency
- [x] re-enable CI
- [x] check and update examples

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -43,7 +43,7 @@ env_logger = "0.7"
 glsl-to-spirv = "0.1.4"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-wasm-bindgen = "0.2.55"
+wasm-bindgen = "0.2"
 console_error_panic_hook = "0.1.6"
 console_log = "0.1.2"
 
@@ -60,6 +60,11 @@ optional = true
 path = "../src/backend/gl"
 version = "0.6"
 features = ["x11"]
+optional = true
+
+[target.'cfg(all(target_arch = "wasm32"))'.dependencies.gfx-backend-gl]
+path = "../src/backend/gl"
+version = "0.6"
 optional = true
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies.gfx-backend-metal]
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -69,6 +69,7 @@ jobs:
             MacOS Nightly,
             Ubuntu Stable,
             Ubuntu Nightly,
+            Web Assembly,
             Windows Stable,
             Windows Nightly,
           ]
@@ -88,6 +89,11 @@ jobs:
           - os: ubuntu-18.04
             name: Ubuntu Nightly
             channel: nightly
+            additional_command:
+          - os: ubuntu-18.04
+            name: Web Assembly
+            channel: nightly
+            target: wasm32-unknown-unknown
             additional_command:
           - os: windows-2019
             name: Windows Stable
--- a/examples/quad/data/index.html
+++ b/examples/quad/data/index.html
@@ -15,3 +15,4 @@
     </script>
   </body>
 </html>
+
--- a/examples/quad/main.rs
+++ b/examples/quad/main.rs
@@ -352,7 +352,7 @@ where
         };
 
         // Image
-        let img_data = include_bytes!("data/logo.png");
+        let img_data = include_bytes!("./data/logo.png");
 
         let img = image::load(Cursor::new(&img_data[..]), image::ImageFormat::Png)
             .unwrap()
@@ -634,7 +634,7 @@ where
         let pipeline = {
             let vs_module = {
                 let spirv =
-                    auxil::read_spirv(Cursor::new(&include_bytes!("data/quad.vert.spv")[..]))
+                    auxil::read_spirv(Cursor::new(&include_bytes!("./data/quad.vert.spv")[..]))
                         .unwrap();
                 unsafe { device.create_shader_module(&spirv) }.unwrap()
             };
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -27,13 +27,13 @@ auxil = { path = "../../auxil/auxil", version = "0.5", package = "gfx-auxil", fe
 smallvec = "1.0"
 glow = "0.6.1"
 parking_lot = "0.11"
-spirv_cross = { version = "0.22", features = ["glsl"] }
+spirv_cross = { version = "0.22", features = ["glsl"]}
 lazy_static = "1"
 raw-window-handle = "0.3"
-x11 = { version = "2", features = ["xlib"], optional = true }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 egl = { package = "khronos-egl", version = "2.1.0" }
+x11 = { version = "2", features = ["xlib"], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3.6"
--- a/src/backend/gl/src/window/web.rs
+++ b/src/backend/gl/src/window/web.rs
@@ -12,7 +12,7 @@ pub struct Swapchain {
     pub(crate) extent: window::Extent2D,
     pub(crate) channel: f::ChannelType,
     pub(crate) raw_format: native::TextureFormat,
-    pub(crate) fbos: ArrayVec<[native::RawFrameBuffer; 3]>,
+    pub(crate) frame_buffers: ArrayVec<[native::RawFrameBuffer; 3]>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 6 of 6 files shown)


Cargo.toml
```diff
@@ -13,4 +13,5 @@ members = [
     "src/hal",
     "src/warden",
     "examples",
+    "examples/quad"
```
kvark, line 16: nit: trailing comma is missing

examples/Cargo.toml
```diff
@@ -43,9 +43,10 @@ env_logger = "0.7"
 glsl-to-spirv = "0.1.4"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-wasm-bindgen = "0.2.55"
+wasm-bindgen = "0.2"
 console_error_panic_hook = "0.1.6"
 console_log = "0.1.2"
+stdweb = "0.4"
```
kvark, line 49: isn't it an alternative to wasm-bindgen? why do we need to depend on both?

examples/quad/Cargo.toml
```diff
@@ -0,0 +1,72 @@
+[package]
```
kvark, line 1: Why do we need the quad example under a different cargo crate?

src/backend/metal/Cargo.toml
```diff
@@ -34,7 +34,7 @@ foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
 cocoa-foundation = "0.1"
-spirv_cross = { version = "0.22", features = ["msl"] }
+spirv_cross = { version = "0.22", features = ["msl"], git = "https://github.com/grovesNL/spirv_cross", rev = "61d3b8c" }
```
kvark, line 37: this seems unnecessary

examples/Cargo.toml
MarcAntoine-Arnaud: True, not needed. It was added for some tests.

examples/quad/Cargo.toml
MarcAntoine-Arnaud: I cannot find a configuration to use the same for native and web.
Nothing was explained to run web examples.
If you find something I'm very interested.

src/backend/metal/Cargo.toml
MarcAntoine-Arnaud: without these changes I cannot compile with the latest version of spirv_cross.
`Ast` don't has the `Parse` and `Compile` traits implemented.

examples/quad/Cargo.toml
kvark: You are right, the README of `/examples` needs to explain how to build for the web. @grovesNL didn't we have that somewhere?

src/backend/metal/Cargo.toml
kvark: That's weird. We build this on CI for every revision, so it compiles.
I just double-checked on my macOS, running `make` in the root of latest master works fine.

examples/quad/Cargo.toml
grovesNL: I'm on my phone so I can't get a link to the header, but it's under the examples README

MarcAntoine-Arnaud: But for me the `examples/Cargo.toml` cannot provide the right target.
In native mode it generate a binary.
For Web, the crate requires to be a library...

src/backend/metal/Cargo.toml
MarcAntoine-Arnaud: Yes it works because you don't build for web, and the master branch of gfx don't support web.
And it's fixed in only [1 commit](https://github.com/grovesNL/spirv_cross/commit/cf025136927c89070b82a1b0f4e9d3d7df806b9e#diff-dc54eb49a875a14243b80fbefe0fb34ddcd9576f674edd0f3e6d1eef3f380f2e) ...

examples/quad/Cargo.toml
grovesNL: It's possible to build binary targets or examples with wasm-bindgen using the instructions there - we don't use wasm-pack or stdweb

src/backend/metal/Cargo.toml
kvark: This is the Metal backend. We aren't supposed to build it on WASM at all.

examples/quad/Cargo.toml
MarcAntoine-Arnaud: I found nothing to run the code using `wasm-bindgen` in development mode.
Nothing mentioned here: https://github.com/gfx-rs/gfx#example

And [paint example](https://github.com/rustwasm/wasm-bindgen/tree/master/examples/paint) that use le `#[wasm_bindgen(start)]` requires to add:
- In the `Cargo.toml`
```rust
[lib]
crate-type = ["cdylib"]
```
- start webpack (using `npm run serve`)

I have't found something to run `wasm-bindgen` without WebPack ...
That's why for the native, the `bin` works. But for Web example it's not enough.

src/backend/metal/Cargo.toml
MarcAntoine-Arnaud: Yes sorry.
I need the last commit for GL backend to fix the Wasm target.
But GL Backend depends on the auxil, that requires to be updated like other backend ...

To limit changes, I can update with:
- last commit of `spirv-cross` for `backend-gl`
- last commit of `spirv-cross` for `auxil` only for `gl` feature

But is it really something relevant ... I'm not 100% convinced.

MarcAntoine-Arnaud: that's why I have added the check on the issue `use published version of spirv_cross dependency`
So maybe we can wait for the new release of spirv before merging that PR.

examples/quad/Cargo.toml
grovesNL: The instructions are located at https://github.com/gfx-rs/gfx/blob/master/examples/README.md#running-quad-with-webglwebassembly

MarcAntoine-Arnaud: Thank you for the link.
I think it's not a real strait forward example setup...
Lot of operations are required to simply open the render.
And for development purpose it's quasi impossible to run and copy file every times ...

A minima I think it can be awesome to add a mention in the [root README](https://github.com/gfx-rs/gfx#example) to redirect to the URL.

I have opened an [issue](https://github.com/rustwasm/wasm-bindgen/issues/2359) on `wasm-bindgen` projet.

examples/quad/src/main.rs
```diff
@@ -21,7 +21,7 @@ extern crate gfx_backend_vulkan as back;
 use wasm_bindgen::prelude::*;
 
 #[cfg(target_arch = "wasm32")]
-#[wasm_bindgen(start)]
+#[wasm_bindgen]
```
grovesNL, line 24: I think we still want `wasm_main` to be the entry point, so `start` should be correct

examples/quad/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
grovesNL, line 1: It would be simpler if we could avoid the need for `package.json` here

examples/quad/webpack.config.js
```diff
@@ -0,0 +1,28 @@
+const path = require('path');
```
grovesNL, line 1: It would be simpler if we could avoid having any webpack configuration

examples/quad/Cargo.toml
MarcAntoine-Arnaud: Ok so I will revert example changes.

src/backend/metal/Cargo.toml
MarcAntoine-Arnaud: Hi @grovesNL, @kvark,
 I have tested locally and personally I'm not 100% sure.
We have 2 options to work with patch:
```toml
[dependencies]
# ...
spirv_cross = { version = "0.22", features = ["glsl"]}
# ...

[patch.'https://github.com/grovesNL/spirv_cross']
spirv_cross = { git = 'https://github.com/grovesNL/spirv_cross', rev = '61d3b8c' }
```

Or that I have used:
```toml
spirv_cross = { version = "0.22", features = ["glsl"], git = 'https://github.com/grovesNL/spirv_cross', rev = '61d3b8c'}
```

Personnaly I'm not a very fan of `[patch....]` syntax.
Do you think to a better solution ?

And in addition, that don't works:
```toml
[target.'cfg(target_arch = "wasm32")'.patch.'https://github.com/grovesNL/spirv_cross']
spirv_cross = { git = 'https://github.com/grovesNL/spirv_cross', rev = '61d3b8c' }
```

So every backend and auxil has to have the patch information...

grovesNL: Yeah that's right, I just meant it's possible to use `patch` locally to test the fix out, instead of replacing all of the dependencies.

I [published a new version of spirv_cross](https://crates.io/crates/spirv_cross) so we can leave it as `version = "0.22"` for now (`cargo update` should automatically update to `0.22.1`)

grovesNL (approved):
Looks great, thank you! Could you please squash all commits before we merge this?

PR #3479, opened 2020-11-17
[dx] Disable automatic Alt+Enter handling in DXGI

### LINKED ISSUE #3477: DX11/12 backends automagically install an Alt+Enter fullscreen handler
Currently, the DX11/DX12 backends will allow a user to toggle to fullscreen by pressing Alt+Enter. This is inconsistent with the other backends, which provide no such behavior. You can see this in the gfx-rs examples:
```
cargo run --bin quad --features dx11
```
and pressing Alt+Enter, which will fullscreen even though there is no code in the app specifically handling this.

This behavior seems to be automatically installed by DXGI, and is controlled by this method:
https://docs.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgifactory-makewindowassociation
I guess this is enabled by default and needs to be explicitly disabled by calling with `DXGI_MWA_NO_ALT_ENTER `.

Ideally this would be disabled by default for the DX backends to maintain consistent behavior, or at least provided as an option.

Short info header:
- GFX version: https://github.com/gfx-rs/gfx/commit/87a24bf390ddd0ee6bc34393d879f2d823aee67f
- OS: Windows 10 Pro 64-bit
- GPU: Nvidia Geforce 2080 Ti

Herschel:
Fixes #3477.

Uses [IDXGIFactory::MakeWindowAssociation](https://docs.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgifactory-makewindowassociation) to disable Alt+Enter behavior on dx11/dx12 backends.

 * winapi-rs does not have the `DXGI_MWA` constants (opened PR https://github.com/retep998/winapi-rs/pull/960). Not sure where to cram these -- for now, I put them right beside the function call.
 * Using both `DXGI_MWA_NO_WINDOW_CHANGES | DXGI_MWA_NO_ALT_ENTER`. Either of these seems to disable the Alt+Enter behavior; according to the docs, `DXGI_MWA_NO_WINDOW_CHANGES` seems more general, but I guess this is desired (DXGI shouldn't fiddle with anything automatically for consistent behavior among backends?).

PR checklist:
  - dx11, dx12, vulkan

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -972,6 +972,11 @@ impl window::PresentationSurface<Backend> for Surface {
             }
         };
 
+        // Disable automatic Alt+Enter handling by DXGI.
+        const DXGI_MWA_NO_WINDOW_CHANGES: u32 = 1;
+        const DXGI_MWA_NO_ALT_ENTER: u32 = 2;
+        self.factory.MakeWindowAssociation(self.wnd_handle, DXGI_MWA_NO_WINDOW_CHANGES | DXGI_MWA_NO_ALT_ENTER);
+
         let mut resource: *mut d3d11::ID3D11Resource = ptr::null_mut();
         assert_eq!(
             winerror::S_OK,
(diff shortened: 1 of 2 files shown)


kvark (approved):
Thank you!
bors r+

PR #3485, opened 2020-11-25
Enable vulkan debug markers in release builds

### LINKED ISSUE #3482: Debug markers in release build
Right now, debug markers only available if the debug_assertions feature is set. This is true at least on the [Vulkan](https://github.com/gfx-rs/gfx/blob/f09b062c5159b4a9dd8879dff3c0a3ac1bd81ea7/src/backend/vulkan/src/lib.rs#L440) and [DirectX 12](https://github.com/gfx-rs/gfx/blob/f09b062c5159b4a9dd8879dff3c0a3ac1bd81ea7/src/backend/dx12/src/lib.rs#L751) backends, but I'd imagine the others as well. It would be nice if there were a feature flag or other way to enable debug markers in release builds as well.

fintelia:
Fixes #3482 (at least for the Vulkan backend)
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -62,7 +62,10 @@ lazy_static! {
             *KHR_GET_PHYSICAL_DEVICE_PROPERTIES2,
         ]
     } else {
-        vec![*KHR_GET_PHYSICAL_DEVICE_PROPERTIES2]
+        vec![
+            DebugUtils::name(),
+            *KHR_GET_PHYSICAL_DEVICE_PROPERTIES2,
+	]
     };
     static ref DEVICE_EXTENSIONS: Vec<&'static CStr> = vec![extensions::khr::Swapchain::name()];
     static ref SURFACE_EXTENSIONS: Vec<&'static CStr> = vec![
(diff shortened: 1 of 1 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -437,7 +433,6 @@ impl hal::Instance<Backend> for Instance {
                 })
             });
 
-        #[cfg(debug_assertions)]
         let debug_messenger = {
             // make sure VK_EXT_debug_utils is available
             if instance_extensions.iter().any(|props| unsafe {
```
kvark, line 438: there is a change missing here. We don't want to request the `VK_EXT_debug_utils` or `DebugReport::name()` if running in release. So these conditions should look like:
```rust
if cfg!(debug_assertions) && instance_extensions.iter().any(...) {
  ...
} else if cfg!(debug_assertions) && instance_extensions.iter().any(..) {
  ...
} else {
  None
}
```

kvark (changes requested):

src/backend/vulkan/src/lib.rs
fintelia: If we don't request debug utils, then [insert_debug_marker will be a no-op.](https://github.com/gfx-rs/gfx/blob/e99e71d0022f72fe4f558ce9f528e258131af2a5/src/backend/vulkan/src/command.rs#L1121)

kvark: Oh snap, you are right. In this case, we should at least gate the `DebugReport::name()` by `debug_assertions`?

kvark (approved):
Thank you!
bors r+

kvark:
Uh, Rust nightly bug :(
Could you comment out the macOS nightly build from bors.toml for now?

kvark:
bors retry

kvark:
bors retry

kvark:
bors retry

kvark:
bors retry

kvark:
bors retry

PR #3486, opened 2020-11-27
[mtl] fix render pass compatibility check wrt sample count

kvark:
Addresses https://github.com/gfx-rs/wgpu/issues/1044 on master
PR checklist:


kvark:
Actually no longer sure it's the right fix. Need to think about it with a fresh head

PR #3487, opened 2020-11-27
[0.6/dx11] Increase max bound descriptor set count

cwfitzgerald:
Allows the latest rend3 version to run on DX11. As the DX11 backend has no actual limit to how many descriptor sets can be bound, I just set it to the max value.


kvark:
Please rebase! (sorry about breaking this for you)

kvark (approved):
bors r+

cwfitzgerald:
Thanks for the review!

bors retry

Issue #3489, opened 2020-11-28
Integrated GPUs are Incorrectly Exposed as Discrete under DX11


cwfitzgerald:
Seems to be similar to #3261, but we need a way to detect this under DX11. Asked on the discord, will see what they say.

cwfitzgerald:
This is resolved as reasonably as it can by #3490

PR #3490, opened 2020-11-28
Determine GPU type on DX11 when possible

Diggsey:
Partially addresses #3489

From what I can tell there's no way to discover this via the DX11 APIs, so guess based on the vendor and device IDs. This can be expanded on later, but initially I've just set it to guess "integrated" for Intel GPUs and "discrete" for everything else, which probably covers >90% of the cases where this applies...

### DIFF
--- a/src/backend/dx11/src/dxgi.rs
+++ b/src/backend/dx11/src/dxgi.rs
@@ -6,7 +6,7 @@ use winapi::{
         guiddef::{GUID, REFIID},
         winerror,
     },
-    um::unknwnbase::IUnknown,
+    um::{d3d11, unknwnbase::IUnknown},
     Interface,
 };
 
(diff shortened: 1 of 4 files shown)


kvark (approved):
Nicely done!
Could you add a commit that bumps the backend version and adds a note to CHANGELOG, please?

Diggsey:
Added 👍

src/backend/dx11/src/dxgi.rs
```diff
@@ -140,30 +140,79 @@ fn enum_adapters1(
...
+            }
+        } else if video_mem > 512_000_000 {
+            DeviceType::DiscreteGpu
+        } else {
+            // Fall back to guessing based on the vendor ID.
+            match (vendor_id, device_id) {
+                // Intel
+                (0x8086, _) => DeviceType::IntegratedGpu,
+                // AMD/ATI
+                (0x1002, _) => DeviceType::IntegratedGpu,
```
cwfitzgerald, line 185: AMD seems to use this identifier for both their discrete and integrated gpus. Additionally, I would rather all and chips be labeled as discrete than vis versa.

src/backend/dx11/src/dxgi.rs
```diff
@@ -140,30 +140,79 @@ fn enum_adapters1(
...
+        name.to_string_lossy().into_owned()
+    };
+
+    AdapterInfo {
+        name: device_name,
+        vendor: vendor_id as usize,
+        device: device_id as usize,
+        device_type: if (flags & dxgi::DXGI_ADAPTER_FLAG_SOFTWARE) != 0 {
+            DeviceType::VirtualGpu
+        } else if winerror::SUCCEEDED(unsafe {
```
cwfitzgerald, line 165: Could we factor this check out into its own line for readability's sake

src/backend/dx11/src/dxgi.rs
```diff
@@ -140,30 +140,79 @@ fn enum_adapters1(
...
+                &mut features_options2 as *mut _ as *mut _,
+                mem::size_of::<d3d11::D3D11_FEATURE_DATA_D3D11_OPTIONS2>() as _,
+            )
+        }) {
+            if features_options2.UnifiedMemoryArchitecture == 1 {
+                DeviceType::IntegratedGpu
+            } else {
+                DeviceType::DiscreteGpu
+            }
+        } else if video_mem > 512_000_000 {
```
cwfitzgerald, line 177: This check should be inverted. If we don't know if it's UMA or not, we should assume <= 512 is integrated.

In point of fact, I don't think we need the vendor based stuff at all. If none of these other checks succeed, we're Discrete.

cwfitzgerald (changes requested):
Looking good! A couple of small logic issues but we're almost there :)

Diggsey:
Updated

src/backend/dx11/src/dxgi.rs
```diff
@@ -140,30 +140,71 @@ fn enum_adapters1(
...
+        device: device_id as usize,
+        device_type: if (flags & dxgi::DXGI_ADAPTER_FLAG_SOFTWARE) != 0 {
+            DeviceType::VirtualGpu
+        } else if discovered_gpu_type {
+            if features_options2.UnifiedMemoryArchitecture == 1 {
+                DeviceType::IntegratedGpu
+            } else {
+                DeviceType::DiscreteGpu
+            }
+        } else if video_mem < 512_000_000 {
```
cwfitzgerald, line 178: nit: `<=`

cwfitzgerald (approved):
One more nit then we're good to go!

cwfitzgerald:
Thanks!

bors r+

kvark:
@Diggsey please make the same PR against `master`, just exclude commit that had a version bump and changelog updates

PR #3493, opened 2020-12-01
[master/dx11] Increase max bound descriptor set count

cwfitzgerald:
Rolls up #3487 now that wgpu is targeting master! Should be the last major rollup I need to do! Woooo!

I looked at #3449 again but I don't think master needs that change, it looks correct.


kvark (approved):
bors r+

kvark:
bors retry

cwfitzgerald:
bors retry

kvark:
bors retry

PR #3494, opened 2020-12-01
Minor OpenGL improvements

### LINKED ISSUE #3495: OpenGL fails to natively render to a window
Short info header:
- GFX version: latest
- OS: Solus OS (Linux)
- GPU: NVidia (default display on a machine with Intel + NV)

It appears that `egl::NATIVE_RENDERABLE` is 0 for all the exposed contexts. Internet appears to be full of people complaining about NVidia specifically, likely related. In the worst case we can expose an adapter that doesn't support any presentation.

kvark:
Fixes #3495
Still doesn't show anything on my other Linux laptop, investigating...
I'm seeing that `NATIVE_RENDERABLE = TRUE`  results in no FB configurations found. I believe this is required for our stuff to work :/


kvark:
Confirmed that it still works on the other machine.
bors r+

kvark:
bors retry

PR #3496, opened 2020-12-02
Use thiserror and refactor errors

kvark:
Makes error variants a bit more sane and structured.
Also takes advantage of `thiserror`, since that's what we settled on in `wgpu` land.
PR checklist:

New code (HAL only):
> Finished dev [unoptimized + debuginfo] target(s) in 17.65s

Old code (HAL only):
> Finished dev [unoptimized + debuginfo] target(s) in 7.73s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+### hal-unreleased
+  - use `thiserror` for errors
+  - error variants and a few names are refactored
+
 ### backend-dx12-unreleased
   - fix SPIR-V entry point selection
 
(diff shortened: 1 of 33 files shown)


kvark:
New timing -  [cargo-timing-20201202T054953Z.html.zip](https://github.com/gfx-rs/gfx/files/5627237/cargo-timing-20201202T054953Z.html.zip)
Old timing -  [cargo-timing-20201202T055308Z.html.zip](https://github.com/gfx-rs/gfx/files/5627242/cargo-timing-20201202T055308Z.html.zip)

It's clear that `gfx` itself doesn't take more time to build. In fact, even less.
But in total the time now has to include the `thiserror` dependencies, including `syn`, `quote`, and `proc-macro`.

So it's a question of whether we expect the clients to already depend on `thiserror` (in which case it's free), or not.
The benefit is clear - our code becomes more manageable (+338 −865).

kvark:
I believe the most important (current and future) consumers of gfx-hal are:
  - `wgpu` - already uses `thiserror`
  - `amethyst` - uses `thiserror` transitively through `cpal`, also depends on a lot of other crates that drag `syn` and friends in
  - `webrender` - also not a concern for it

So I think we should bite the bullet and go for it.

grovesNL (approved):
Thanks, looks good to me!

I agree that the additional compilation time could be a concern, but also agree that most consumers of gfx-hal will already have these dependencies somewhere in their dependency tree.

PR #3498, opened 2020-12-02
Add Naga dependency

kvark:
By exposing Naga in gfx-hal we'll be able to have wgpu and gfx communicate the shaders without going through SPIR-V.
PR checklist:

The controversial part here is that Naga is not optional. I tried to think of pros and cons.
Cons:
- Naga needs more work anyway, it's not ready yet
- Naga in gfx-hal isn't needed for any libraries that only depend on the gfx-hal, and do not care about Naga

Pros:
- Naga *will* need to be ready at some point, replacing SPIRV-Cross. And it will need to be available for wgpu interop, so we might as well skip the stage where it's optional in the gfx-hal.
- Naga by itself without features is super lightweight, it shouldn't affect the libraries too much. All the frontends and backends are gated by features.
- The end user app will need to depend on gfx-hal backends, and probably not the Vulkan backend alone. So in the end those backends would require Naga anyway to do their translation (currently, only Metal backend supports this, optionally).
- It's annoying in Cargo to say that `naga` dep is optional in a backend, in addition to requiring `naga` dep in the gfx-hal. This would require renaming the feature in the backend to be something else.

Unresolved question: how do we nicely specify Naga revision? It's needed everywhere: in gfx-hal, in each backend, in wgpu, etc, and all of them need to agree. We can't simply re-export Naga from gfx-hal, since every client will need to enable some of the features (i.e. Metal backend enables "mtl-out", `wgpu` enables "wgsl-in", etc).

One way to proceed is to just use the revisions. It's a bit clunky, since for the reader a revision is just a random set of symbols.
Another interesting way is to have tags in the Naga repo, and depend on those tags. E.g. it could be called `gfx-1`, `gfx-2`, etc, monotonically, as we update Naga dependency through the stack. I think that could work fine, at least better than with revisions only.

### DIFF
--- a/src/hal/src/device.rs
+++ b/src/hal/src/device.rs
@@ -396,6 +396,14 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
     /// [graphics pipelines][crate::pso::GraphicsPipelineDesc].
     unsafe fn create_shader_module(&self, spirv: &[u32]) -> Result<B::ShaderModule, ShaderError>;
 
+    /// Create a new shader module from the `naga` module.
+    unsafe fn create_shader_module_from_naga(
+        &self,
+        _module: naga::Module,
+    ) -> Result<B::ShaderModule, ShaderError> {
+        Err(ShaderError::Unsupported)
+    }
+
     /// Destroy a shader module module
     ///
     /// A shader module can be destroyed while pipelines created using its shaders are still in use.
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -74,32 +74,25 @@ jobs:
           - os: macos-10.15
             name: MacOS Stable
             channel: stable
-            additional_command: cd src/backend/metal && cargo check --all-features
           - os: macos-10.15
             name: MacOS Nightly
             channel: nightly
-            additional_command:
           - os: ubuntu-18.04
             name: Ubuntu Stable
             channel: stable
-            additional_command:
           - os: ubuntu-18.04
             name: Ubuntu Nightly
             channel: nightly
-            additional_command:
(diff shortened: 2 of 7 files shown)


grovesNL:
Using tags for now sounds good to me.

In the longer term, the Naga public API will stabilize which should allow us to release frequently anyway. This might allow us to switch off of the git dependency anyway, and still use `patch` to test fixes between Naga releases.

src/hal/src/device.rs
```diff
@@ -396,6 +396,14 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
     /// [graphics pipelines][crate::pso::GraphicsPipelineDesc].
     unsafe fn create_shader_module(&self, spirv: &[u32]) -> Result<B::ShaderModule, ShaderError>;
 
+    /// Create a new shader module from the `naga` module.
+    unsafe fn create_shader_module_from_naga(
```
grovesNL, line 400: Could Naga still be optional if we accepted an enum for `create_shader_module` (e.g. a `SpirV` variant and a feature-gated `Naga` variant)?

grovesNL (approved):

src/hal/src/device.rs
kvark: I think Naga being optional is a bit orthogonal to whether we have an `enum` or not. In both cases we have a choice, the enum just requires `#[non_exhaustive]` on it. In fact, I tried to prototype it originally and then went back to a different entry point.
Now I don't have a strong preference one way or the other.

kvark:
Discussed on the Matrix. Proceeding as is. Will keep in mind the option to 1) make it optional, and 2) to merge the entry points.
bors r=grovesNL

PR #3499, opened 2020-12-02
[dx12] Fix `set_command_buffer_name`

yzsolt:
`CommandBuffer.raw` is null when the command buffer is not recording, so
we have to store the name and set it later in `CommandBuffer::begin`.

Fixes water and texture-arrays `wgpu-rs` examples: https://github.com/gfx-rs/wgpu/issues/1059
PR checklist:
- [ ] `make` succeeds (on *nix) - not applicable
- [x] tested examples with the following backends: water and texture-arrays from wgpu-rs with dx12 backend

### DIFF
--- a/src/backend/dx12/src/command.rs
+++ b/src/backend/dx12/src/command.rs
@@ -449,6 +449,8 @@ pub struct CommandBuffer {
 
     /// Temporary transition barriers.
     barriers: Vec<d3d12::D3D12_RESOURCE_BARRIER>,
+    /// Name of the underlying raw `GraphicsCommandList` object.
+    pub(crate) raw_name: Vec<u16>,
 }
 
 impl fmt::Debug for CommandBuffer {
@@ -496,6 +498,7 @@ impl CommandBuffer {
             retained_resources: Vec::new(),
             temp_marker: Vec::new(),
             barriers: Vec::new(),
+            raw_name: Vec::new(),
         }
     }
 
@@ -1205,6 +1208,10 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
         assert_eq!(self.raw, native::GraphicsCommandList::null());
         self.allocator = Some(allocator);
         self.raw = list;
+
+        if !self.raw_name.is_empty() {
+            self.raw.SetName(self.raw_name.as_ptr());
+        }
     }
 
     unsafe fn finish(&mut self) {
@@ -1220,6 +1227,10 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
             self.raw.close();
         }
         if self.phase != Phase::Initial {
+            // Reset the name so it won't get used later for an unnamed `CommandBuffer`.
+            const EMPTY_NAME: u16 = 0;
+            self.raw.SetName(&EMPTY_NAME);
+
             self.pool_shared.release_list(self.raw);
             self.raw = native::GraphicsCommandList::null();
         }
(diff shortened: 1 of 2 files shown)


yzsolt:
I haven't checked if the name appears correctly in graphics debuggers. If they sample the CommandList object's name separately for every command buffer recording session, it should be ok. But even if not, at least we don't crash now.

src/backend/dx12/src/device.rs
```diff
@@ -3615,8 +3615,7 @@ impl d::Device<B> for Device {
     }
 
     unsafe fn set_command_buffer_name(&self, command_buffer: &mut cmd::CommandBuffer, name: &str) {
-        let cwstr = wide_cstr(name);
-        command_buffer.raw.SetName(cwstr.as_ptr());
+        command_buffer.raw_name = wide_cstr(name);
```
kvark, line 3618: we still want to assign it immediately if possible (i.e. if  `raw` is not NULL)

src/backend/dx12/src/command.rs
```diff
@@ -1205,6 +1208,15 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
         assert_eq!(self.raw, native::GraphicsCommandList::null());
         self.allocator = Some(allocator);
         self.raw = list;
+
+        if self.raw_name.is_empty() {
+            // Reset the name so a previously named raw GraphicsCommandList's 
+            // name won't be used for an unnamed CommandBuffer.
+            const EMPTY_NAME: u16 = 0;
+            self.raw.SetName(&EMPTY_NAME);
```
kvark, line 1216: instead of doing this in `begin()`, did you consider resetting the name unconditionally when the `raw` gets returned to the `pool_shared`? It seems like that could be an elegant solution.

kvark (commented):
Looks great, thank you!
Just a couple of small things before we land.

src/backend/dx12/src/command.rs
yzsolt: I was afraid that graphics debuggers may not pick up the name if it gets cleared "early", but now that I fiddled around a bit with RenderDoc and PIX, concluded that they don't care about `ID3D12GraphicsCommandList` object names at all. Even tried with modified C++ DX12 samples, on multiple vendors, still no name is shown for command list objects. Other object (e.g. `CommandQueue`, `CommandAllocator`) names are shown properly.

So setting the name seems to be useless, as it has no effect. Of course it doesn't hurt doing it anyway.

src/backend/dx12/src/device.rs
```diff
@@ -3616,7 +3616,11 @@ impl d::Device<B> for Device {
 
     unsafe fn set_command_buffer_name(&self, command_buffer: &mut cmd::CommandBuffer, name: &str) {
         let cwstr = wide_cstr(name);
-        command_buffer.raw.SetName(cwstr.as_ptr());
+        if !command_buffer.raw.is_null() {
+            command_buffer.raw.SetName(cwstr.as_ptr());
```
kvark, line 3620: if we set the name, we still want to remember it in `raw_name` as well, so that we can set it again upon the buffer being re-recorded

kvark (commented):
Thank you for updating! It looks great, just one last small thing

src/backend/dx12/src/command.rs
kvark: Would you want to file an issue to `RenderDoc` about this?

yzsolt: Sure! Probably not an issue, since SetName is explicitly skipped in the injected wrapper code for CommandLists, but maybe the author(s) can shed some light on the reasons.

kvark (approved):
Wonderful, thank you!
bors r+

PR #3503, opened 2020-12-03
Fix Metal entry points reflected from Naga, extend to Vulkan backend

kvark:
Depends on https://github.com/gfx-rs/naga/pull/299 . Will need to change the tag once that lands.
Paves more way towards the pervasive Naga usage across the stack.


grovesNL (approved):

kvark:
bors r=grovesNL

PR #3506, opened 2020-12-03
[gl] Add Instance struct in web version to align API with other backends.

VincentFTS:
PR checklist:
- [X] `make` succeeds (on macOS)
- [X] tested examples with the following backends: quad example with gl backend with wasm target arch (no windows nor linux available to test gl backend there)

### DIFF
--- a/src/backend/gl/src/window/web.rs
+++ b/src/backend/gl/src/window/web.rs
@@ -4,7 +4,7 @@ use crate::{
 use arrayvec::ArrayVec;
 use glow::HasContext;
 use hal::{adapter::Adapter, format as f, image, window};
-use std::iter;
+use std::{iter, sync::Mutex};
 use wasm_bindgen::JsCast;
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


VincentFTS:
I did not protect `create_surface()` from creating one more surface when one already exists since no Error seems to be relevant.

src/backend/gl/src/window/web.rs
```diff
@@ -15,38 +15,68 @@ pub struct Swapchain {
...
+                swapchain: None,
+                renderbuffer: None,
+            })
         } else {
             unreachable!()
         }
     }
 
+    unsafe fn destroy_surface(&self, _surface: Surface) {
+        // TODO: Implement Surface cleanup
```
kvark, line 68: can just make the canvas to `None`?

kvark (commented):
This looks great! It's the last piece of the puzzle to get the API aligned between all the backends. Just a few notes before we go.

kvark (approved):
Well done, thank you!
bors r+

VincentFTS:
My pleasure !

PR #3507, opened 2020-12-03
[gl] Remove WebGL hardcoded min_storage_buffer_offset_alignment value.

VincentFTS:
PR checklist:
- [X] `make` succeeds (on macOS)
- [x] tested examples with the following backends: quad example with gl backend with wasm target arch (no windows nor linux available to test gl backend there)


kvark (approved):
Thank you!
bors r+

PR #3520, opened 2020-12-07
OpenGL Naga integration

### LINKED ISSUE #3519: Replace GL's x11 dependency by hand-written code
We don't need much from the x11 library. Maybe it's an overkill to hook it up, and we can get away with something that we just hack in?

kvark:
Closes #3519
The meat of the issue here is to get the proper resource mapping plumbed between gfx and Naga glsl backend, and this isn't done in the PR. However, very primitive shaders that use no resources (e.g. hello-triangle) should already work.

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -66,11 +66,11 @@ impl Device {
         }
     }
 
-    pub fn create_shader_module_from_source(
+    fn create_shader_module_raw(
         &self,
         shader: &str,
         stage: ShaderStage,
-    ) -> Result<n::ShaderModule, d::ShaderError> {
+    ) -> Result<n::Shader, d::ShaderError> {
         let gl = &self.share.context;
 
         let can_compute = self.share.limits.max_compute_work_group_count[0] != 0;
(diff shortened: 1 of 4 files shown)


src/backend/gl/src/device.rs
```diff
@@ -396,13 +405,14 @@ impl Device {
         desc_remap_data: &mut n::DescRemapData,
         name_binding_map: &mut FastHashMap<String, (n::BindingTypes, pso::DescriptorBinding)>,
     ) -> n::Shader {
-        assert_eq!(point.entry, "main");
         match *point.module {
             n::ShaderModule::Raw(raw) => {
+                assert_eq!(point.entry, "main");
```
grovesNL, line 410: FWIW we probably don't need these because we appear to be passing `point.entry` through to naga/spirv-cross

grovesNL (approved):

src/backend/gl/src/device.rs
kvark: Ok, I'll remove these in the follow-up.

kvark:
bors r=grovesNL

PR #3527, opened 2020-12-10
Return None for unsupported formats in image_format_properties

ElArtista:
Fixes image_format_properties method for gfx-backend-gl to return None for unsupported/unmapped image formats.
Up until now clients where tricked into believing that all formats where available and subsequent calls to image_create failed.


grovesNL (approved):
Thank you!
bors r+

PR #3529, opened 2020-12-12
Support xcb window handles in egl create_surface

ElArtista:


src/backend/gl/src/window/egl.rs
```diff
@@ -158,6 +158,8 @@ impl hal::Instance<crate::Backend> for Instance {
         let mut native_window = match has_handle.raw_window_handle() {
             #[cfg(not(target_os = "android"))]
             Rwh::Xlib(handle) => handle.window,
+            #[cfg(not(target_os = "android"))]
+            Rwh::Xcb(handle) => handle.window as _,
```
kvark, line 162: This may need be a little more complicated than just this. Currently, under x11 feature we are using the XLib handle in `get_platform_display`. With XCB, I guess the `Instance::create()` needs to be changed a bit as well? I.e. it could try xlib, then fall back to xcb, or something like that (not really sure)

kvark (commented):
does it work?

src/backend/gl/src/window/egl.rs
ElArtista: For some reason though, it does not fail on egl::get_platform_display with the vkcube XCB example I'm running now, although I'll investigate into it more

ElArtista: I'm pretty sure that the window handles are interoperable, given that they are only an Id in the X11 system:
https://www.x.org/releases/X11R7.7/doc/libxcb/tutorial/index.html#helloworld

ElArtista: I can also verify this, because multiple window manipulation tools (xprop, xdotool) and also my window manager (bspwm) reference the windows by their id too and can use the same id between them

kvark (approved):
bors r+

PR #3530, opened 2020-12-12
Renamed ImageKind to ImageType in opengl backend

ElArtista:
This will allow us to store a kind: hal::image::Kind field in the near future


kvark (approved):
thank you!
bors r+

PR #3531, opened 2020-12-13
Implement get_image_subresource_footprint in opengl backend

ElArtista:
Implementation similar to metal backend


kvark (approved):
Note that the exact logic of the footprint would only matter for mappable textures, which I don't think the backend supports just yet.
bors r+

PR #3532, opened 2020-12-14
Picking correct depth/stencil binding point relevant to attachment aspect

ElArtista:


ElArtista:
PR has dependency to #3531, will rebase it after it gets merged.

src/backend/gl/src/device.rs
```diff
@@ -1114,12 +1114,22 @@ impl d::Device<B> for Device {
             }
 
             if let Some(depth_stencil) = subpass.depth_stencil {
+                let aspects = match &attachments[depth_stencil] {
```
kvark, line 1117: A little pet peeve of mine (sorry!) - could we please not match on a reference?
There is no need to take advantage of the extra sugar here, we'd rather want explicit code, i.e. `match attachments[depth_stencil] {...}`.

src/backend/gl/src/device.rs
```diff
@@ -1114,12 +1114,22 @@ impl d::Device<B> for Device {
...
+                    n::ImageView::Texture { sub, .. } => &sub.aspects,
+                    n::ImageView::Renderbuffer { aspects, .. } => aspects,
+                };
+
+                let attachment = [
```
kvark, line 1122: let's not try to be too fancy here
```rust
if aspects == Aspects::DEPTH { glow::DEPTH_ATTACHMENT }
else if aspects == Aspects::STENCIL { glow::STENCIL_ATTACHMENT }
else { glow::DEPTH_STENCIL_ATTACHMENT }
```

src/backend/gl/src/native.rs
```diff
@@ -253,7 +256,10 @@ impl SwapchainImage {
                 num_levels: 1,
                 num_layers: 1,
             },
-            view: ImageView::Renderbuffer(renderbuffer),
+            view: ImageView::Renderbuffer {
+                raw: renderbuffer,
+                aspects: format::Aspects::empty(),
```
kvark, line 261: any reason to not use `::COLOR` here?

ElArtista: I'm matching the aspect above in the format :)

src/backend/gl/src/device.rs
ElArtista: Done.

src/backend/gl/src/device.rs
ElArtista: Ok!

src/backend/gl/src/native.rs
kvark: oh my!

kvark (approved):
bors r+

PR #3533, opened 2020-12-15
Rewrite the GL binding model

### LINKED ISSUE #2784: OpenGL binding model
Related to the implementation in #2071 and general limitations of #1619.

## Problem
GLSL has this super annoying limitation of only having combined image samplers. API itself has seen the addition of separate "sampler object" concept, but the shading language never caught up to it :( What this means for us is basically that this Vulkan GLSL code is not compatible with OpenGL:
```glsl
layout(set = 0, binding = 0) uniform sampler s_PaletteSampler;
layout(set = 1, binding = 0) uniform utexture1D t_ColorTable;
layout(set = 1, binding = 1) uniform texture1D t_Palette;
layout(set = 1, binding = 2) uniform sampler s_ColorTableSampler;
layout(set = 1, binding = 3) uniform sampler s_OtherSampler;
void main() {
    uvec2 color_params = texelFetch(usampler1D(t_ColorTable, s_ColorTableSampler), 5, 0).xy;
    uvec2 other_params = texelFetch(usampler1D(t_ColorTable, s_OtherSampler), 6, 0).xy;
    vec4 main_color = texelFetch(sampler1D(t_Palette, s_PaletteSampler), int(color_params[0]), 0);
    vec4 other_color = texelFetch(sampler1D(t_Palette, s_OtherSampler), int(other_params[0]), 0);
}
```
On the level of the binding model we have 2 images and 2 samplers, but OpenGL-style GLSL would have to use 4 different "sampler" objects (which are combined), and this number can grow as the number of combinations of images used with samplers grows in a program...

Note: this problem applies equally to gfx-rs, Vulkan Portability, and WebGPU.
cc @msiglreith @ZeGentzy @grovesNL @Kangz

## Solution
There is no good way to solve it, but we need a strict rule defined that could be:
  1. efficiently implemented by the backend without diverging too much from the Vulkan binding model (in particular, without introducing too much lazyness in binding and pipeline creation).
  2. clearly communicated to the clients and verified on our side as a part of the portability contract, while still permitting typical use-cases to work unchanged.

My proposal is to have the following rule: each image needs to be used by a shader with *no more than one sampler*, known at pipeline creation time. This means that the binding model would only have a single slot for this image, and that it could persist through the changing pipelines as a command buffer is recorded.

In terms of implementation, when a new pipeline is set, it needs to bind the GL sampler objects to the active texture slots used by the shader, corresponding to that constant association of the images with samplers in the shader (that pipeline knows about).

## Challenges
The idea isn't fully thought through or tested yet. I imagine one of the difficulties would be extracting the image-sampler association from the SPIR-V. Some analysis is needed that would take into account uniform control flow and specialization constants...

Another problem is what do we do about client code that doesn't fit into the rule. It could be useful to provide an ability for the user to say, at the pipeline layout (or descriptor set layout) creation time, that a particular image will be used with more than one sampler, without compromising the general effectiveness of the binding model.

### LINKED ISSUE #2858: [gl] Describe backend internals
See https://github.com/gfx-rs/gfx/pull/2857/files#diff-6e80d01aaeb75f01d28d8338b0a64797R2 for a rough start.
cc @grovesNL @ZeGentzy @kyren 

Doing this would help maintaining the backend in the long run.

kvark:
Closes #2858
Closes #2784
Depends on https://github.com/gfx-rs/naga/pull/333

I was having a hard time trying to map our Naga semantics to the old binding model of gfx-backend-gl. I also couldn't stand the fact that the pipeline layouts were internally mutable, and mapping could grow as more pipelines are created. That seemed very out of line with regards to how other backends are written, and I didn't fully understand that code...

So the binding model is totally rewritten here. It becomes somewhat similar to Metal or D3D11, but with special logic to update the texture binding state instead of having true sampler objects.
The new code likely hosts bugs, I wasn't able to test it a lot, especially given the overall state of the backend. I think this rewrite is still a good investment into the future of the backend, and we should cope with the bugs coming our way.

The thing I'm most interesting in getting reviewed here is the general approach, the architecture of the binding model, which is described a bit in the crate `lib.rs` comment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,6 @@ members = [
     "src/warden",
     "examples",
 ]
+
+[patch."https://github.com/gfx-rs/naga"]
+#naga = { path = "../naga" }
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -36,7 +36,7 @@ libloading = "0.6"
 
 [dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
-tag = "gfx-4"
+tag = "gfx-5"
 #TODO: remove `spv-out` once spirv-cross is deprecated
 features = ["spv-out", "glsl-out"]
 optional = true
--- a/src/backend/gl/src/command.rs
+++ b/src/backend/gl/src/command.rs
@@ -1,6 +1,6 @@
 #![allow(missing_docs)]
 
-use crate::GlContext;
+use crate::{GlContext, MAX_SAMPLERS};
 
 use hal::{
     self, buffer, command,
@@ -148,7 +148,7 @@ pub enum Command {
     BindBufferRange(u32, u32, n::RawBuffer, i32, i32),
     BindTexture(u32, n::Texture, n::TextureTarget),
     BindSampler(u32, n::Sampler),
-    SetTextureSamplerSettings(u32, n::Texture, n::TextureTarget, image::SamplerDesc),
+    SetTextureSamplerSettings(u32, n::TextureTarget, image::SamplerDesc),
 
     SetColorMask(Option<DrawBuffer>, pso::ColorMask),
     SetDepthMask(bool),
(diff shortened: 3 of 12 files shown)


kvark:
Rebased now, also added a bit of logic to reflect Naga's buffer resources.
bors r+

PR #3536, opened 2020-12-16
Remove optional dependency on mint

tangmi:
It seems that `mint` is no longer used anywhere in `gfx`. This change removes the listing from the manifest file.


kvark (approved):
Thank you!
bors r+

Issue #2418, opened 2018-09-21
RTX Extensions for DX12, Vulkan and OpenGL


novacrazy:
I'm not sure what the protocol is for extension support, but with Nvidia's newest GPUs they have also helped to add ray-tracing extensions to DirectX 12, Vulkan ~~and even OpenGL~~. These three extensions are very similar, so adding them to the HAL would be nice to have.

Quick links:
[VK_NV_raytracing](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VK_NV_ray_tracing)
[GLSL_NV_raytracing](https://github.com/KhronosGroup/GLSL/blob/master/extensions/nv/GLSL_NV_ray_tracing.txt)
[SPV_NV_raytracing](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/NV/SPV_NV_ray_tracing.html)

Additionally, I found that Metal supports some ray-tracing acceleration via the [Metal Performance Shaders](https://developer.apple.com/documentation/metalperformanceshaders).

anchpop:
Any progress on this?

kvark:
Whatever we expose will likely be based on Khronos efforts, and there isn't anything yet to target, see https://github.com/KhronosGroup/Vulkan-Docs/issues/686

dgkoch:
> Nvidia's newest GPUs they have also helped to add ray-tracing extensions to DirectX 12, Vulkan and even OpenGL

Actually there is no OpenGL support.  The VK/GLSL/SPV are all part of the Vulkan extension (which is now called VK_NV_ray_tracing btw).

novacrazy:
Ah, I see. Slightly unfortunate but understandable. I've updated the links as well, which seemed to have rotted due to the name change.

helixd-2k18:
> and even OpenGL
Can be opted by OptiX, but may have some problems with shaders compilation. Can be used Vulkan interop (but better use Vulkan itself).
For DirectX 12 can be used DXR. For Vulkan API can be used `VK_NV_ray_tracing`. For shading language can be used HLSL with DXC.

kvark:
[...]

PR #3544, opened 2020-12-19
WIP acceleration structure API proposal

tangmi:
Initial API proposal for just the acceleration structure part of adding ray tracing support. Once the API design is set, we can implement (and smoke test) and then move on to adding ray tracing pipelines and ray query feature detection.

There's a lot of open questions I have in the comments and I intend for this draft to be a place to track discussion. @kvark: I intend to drive discussion on individual pieces at `#gfx:matrix.org`, referring to this PR as a reference, so don't worry about reviewing all at once (I'll make sure we get through all of it)!

**The main theme of nearly all my questions is about how far can we deviate from `VK_KHR_acceleration_structure` to better fit `gfx` (and potentially gain some API safety).**

Here's some notes on implementation differences I expect between DX12 and Vulkan, which may be interesting to discuss how we'll solve:
- `B::AccelerationStructure`
	- VK: `vk::AccelerationStructureKHR`
	- DX: Proabaly `()`, since there's no handle associated with an acceleration structure--they're just packed into buffers.
- Serialize/Deserialize
	- VK:
		- Get serialized size: `ash::extensions::khr::AccelerationStructure::write_acceleration_structures_properties` with `ash::vk::QueryType::ACCELERATION_STRUCTURE_COMPACTED_SIZE_KHR`
	- DX:
		- Get serialized size: `ID3D12GraphicsCommandList4::EmitRaytracingAccelerationStructurePostbuildInfo` (or `BuildRaytracingAccelerationStructure`), but it writes to a buffer and not to CPU mem?
- Host operations
	- VK: `vk::PhysicalDeviceAccelerationStructureFeaturesKHR::acceleration_structure_host_commands` + methods on `Device`
	- DX: unsupported natively, could be emulated in gfx
- Specifically laid out structs for GPU buffers:
	- Transform matrix: 3x4 row matrix that corresponds to `mint::RowMatrix3x4`. Metal uses a 4x4 matrix (`matrix_float4x4`), which I believe maps to `mint::RowMatrix4`, given [this "Working with Matrices" doc](https://developer.apple.com/documentation/accelerate/working_with_matrices).
	- AABB positions: Both APIs use a `float3` min, then max to specify the AABB. Metal puts a bounding box on every acceleration structure and doesn't have an AABB-specific variant.
	- TLAS instance struct:
		- VK: [`VkAccelerationStructureInstanceKHR`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureInstanceKHR.html)
		- DX: [`D3D12_RAYTRACING_INSTANCE_DESC`](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_raytracing_instance_desc)
		- These two structs have the same layout ([not it Metal, though](https://developer.apple.com/documentation/metalperformanceshaders/mpsinstanceaccelerationstructure)) and are probably intended to be written from compute shaders, which probably users probably need to know about it (either by docs or querying for it).
- Building acceleration structures
	- Primitive counts
		- VK: Separates the primitive counts and the geometry data in `VkAccelerationStructureGeometryKHR`  and `VkAccelerationStructureBuildRangeInfoKHR` at descriptor and build time, respectively
		- DX: Requires the primitive count at descriptor time when creating `D3D12_RAYTRACING_GEOMETRY_TRIANGLES_DESC` or `D3D12_RAYTRACING_GEOMETRY_AABBS_DESC`, and doesn't require any additional data at build time

Related: #2418

### DIFF
--- a/src/hal/src/acceleration_structure.rs
+++ b/src/hal/src/acceleration_structure.rs
@@ -1,0 +1,232 @@
+//! TODO docs
+
+// TODO remove and add docs
+#![allow(missing_docs)]
+
+use std::ops::Range;
+
+use crate::{buffer::Offset, format::Format, Backend, IndexType};
+
+/// Denotes the type of acceleration structure.
+#[derive(Debug)]
+pub enum AccelerationStructureType {
+    ///
+    TopLevel,
+    ///
+    BottomLevel,
+    // TODO vulkan supports "generic level" (where the concrete build type is specified), but discourages its use for applications "written directly for Vulkan" since it "could affect capabilities or performance in the future" (https://www.khronos.org/blog/vulkan-ray-tracing-final-specification-release). Perhaps this is to better support `vkd3d-proton`, but we probably don't want it exposed in gfx?
+}
+
+/// A description of the data needed to build an acceleration structure.
+#[derive(Debug)]
+pub struct AccelerationStructureDesc<'a, B: Backend> {
+    /// TODO: document lifetime required for this buffer
+    pub buffer: &'a B::Buffer,
+    /// TODO: the final gpu address on DX12 needs to be D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BYTE_ALIGNMENT (256), but vulkan only requires this offset to be 256 (as defined by https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureCreateInfoKHR.html)
+    pub buffer_offset: Offset,
+
+    /// The type of acceleration structure to build.
+    pub ty: AccelerationStructureType,
+    // /// currently only has `accelerationStructureCaptureReplay`
+    // create_flags: VkAccelerationStructureCreateFlagsKHR,
+    // /// used for `accelerationStructureCaptureReplay`
+    // device_address: VkDeviceAddress,
+}
+
+bitflags! {
+    /// https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkBuildAccelerationStructureFlagBitsKHR.html
+    pub struct AccelerationStructureFlags: u32 {
+        /// indicates that the specified acceleration structure can be updated with update of VK_TRUE in vkCmdBuildAccelerationStructuresKHR or vkCmdBuildAccelerationStructureNV .
+        const ALLOW_UPDATE = 0x1;
+        ///  indicates that the specified acceleration structure can act as the source for a copy acceleration structure command with mode of VK_COPY_ACCELERATION_STRUCTURE_MODE_COMPACT_KHR to produce a compacted acceleration structure.
+        const ALLOW_COMPACTION = 0x2;
+        /// indicates that the given acceleration structure build should prioritize trace performance over build time.
+        const PREFER_FAST_TRACE = 0x4;
+        ///  indicates that the given acceleration structure build should prioritize build time over trace performance.
+        const PREFER_FAST_BUILD = 0x8;
+        /// indicates that this acceleration structure should minimize the size of the scratch memory and the final result build, potentially at the expense of build time or trace performance.
+        const LOW_MEMORY = 0x10;
+    }
+}
+
+/// TODO docs
+#[derive(Debug)]
+pub enum AccelerationStructureCreateMode {
+    /// specifies that the destination acceleration structure will be built using the specified geometries.
+    Build,
+    /// specifies that the destination acceleration structure will be built using data in a source acceleration structure, updated by the specified geometries.
+    /// Note there's constraints on update: https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#acceleration-structure-update-constraints
+    Update,
+}
+
+/// A description of the geometry data needed to populate an acceleration structure.
+///
+/// TODO: there's something here that smells w/ what fields are needed to get the required build size vs what fields are needed to actually build. Also, the top/bottom levels having different requirements on which fields are valid.
+#[derive(Debug)]
+pub struct AccelerationStructureGeometryDesc<'a, B: Backend> {
+    pub flags: AccelerationStructureFlags,
+    /// The type of acceleration structure to build.
+    pub ty: AccelerationStructureType,
+    pub mode: AccelerationStructureCreateMode,
+
+    pub src: Option<B::AccelerationStructure>,
+    pub dst: Option<B::AccelerationStructure>,
+
+    // TODO: We could enforce the following the type system?
+    // - in both vulkan (via `VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03792`) and DX12 (via type system), all of the structs here must be the same variant.
+    // - blas must be triangles or aabbs, tlas must be instances
+    pub geometries: &'a [AccelerationStructureGeometry<'a, B>],
+    // Both APIs support "array" and "array of pointers", presumably to allow for cheap reuse of the geometry descriptors.
+    // pgeometries: &'a [&'a GeometryDesc<'a, B>],
+    pub scratch: Option<(&'a B::Buffer, Offset)>,
+}
+
+bitflags! {
+    /// https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkGeometryFlagBitsKHR.html
+    pub struct AccelerationStructureGeometryFlags: u32 {
+        /// indicates that this geometry does not invoke the any-hit shaders even if present in a hit group.
+        const OPAQUE = 0x1;
+        /// indicates that the implementation must only call the any-hit shader a single time for each primitive in this geometry. If this bit is absent an implementation may invoke the any-hit shader more than once for this geometry.
+        const NO_DUPLICATE_ANY_HIT_INVOCATION = 0x2;
+    }
+}
+
+/// TODO docs
+#[derive(Debug)]
+pub struct AccelerationStructureGeometry<'a, B: Backend> {
+    pub flags: AccelerationStructureGeometryFlags,
+    pub geometry: Geometry<'a, B>,
+}
+
+/// TODO docs
+#[derive(Debug)]
+pub enum Geometry<'a, B: Backend> {
+    /// TODO docs
+    Triangles(GeometryTriangles<'a, B>),
+
+    /// TODO docs
+    /// TODO bikeshed capitalization of AABBs.
+    Aabbs(GeometryAabbs<'a, B>),
+
+    // TODO
+    /// TODO docs
+    Instances(GeometryInstances<'a, B>),
+}
+
+// https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureGeometryTrianglesDataKHR.html
+// for docs: Memory Safety (https://nvpro-samples.github.io/vk_raytracing_tutorial_KHR/#accelerationstructure)
+// TODO: I don't like that this is reused for the "get build sizes" and "do actual build" with complex rules on what fields are ignored when. Perhaps we could use DX12's model for `D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC` and `D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS`?
+#[derive(Debug)]
+pub struct GeometryTriangles<'a, B: Backend> {
+    /// Both APIs require support for at least the following:
+    /// - `(R32_G32, Float)`
+    /// - `(R32_G32_B32, Float)`
+    /// - `(R16_G16, Float)`
+    /// - `(R16_G16_B16_A16, Float)`
+    /// - `(R16_G16, Inorm)`
+    /// - `(R16_G16_B16_A16, Inorm)`
+    /// VK could support more by querying `VK_FORMAT_FEATURE_ACCELERATION_STRUCTURE_VERTEX_BUFFER_BIT_KHR`, DX12 is not queryable? Note [the DX12 ray tracing spec](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#d3d12_raytracing_geometry_triangles_desc) says it supports more than [the Win32 docs](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_raytracing_geometry_triangles_desc).
+    pub vertex_format: Format,
+    pub vertex_buffer: &'a B::Buffer,
+    pub vertex_buffer_offset: Offset,
+    pub vertex_buffer_stride: Offset,
+
+    /// aka "vertex count"?
+    pub max_vertex: Offset,
+
+    // index format must be DXGI_FORMAT_R32_UINT, DXGI_FORMAT_R16_UINT
+    // can also be VK_INDEX_TYPE_NONE_KHR/DXGI_FORMAT_UNKNOWN
+    pub index_buffer: Option<(&'a B::Buffer, Offset, IndexType)>,
+
+    /// 3x4 matrix
+    /// TODO enum for cpu repr?
+    /// Must point to an address containing `TransformMatrix`
+    pub transform: Option<(&'a B::Buffer, Offset)>,
+}
+
+// VkTransformMatrixKHR
+#[derive(Debug)]
+pub struct TransformMatrix {
+    /// 3x4 row-major affine transformation matrix. Use `mint::RowMatrix3x4` if available.
+    pub matrix: [[f32; 4]; 3],
+}
+
+#[derive(Debug)]
+pub struct GeometryAabbs<'a, B: Backend> {
+    /// Must point to a buffer with buffer with an array of `AabbPositions`s
+    /// TODO: document lifetime required for this buffer
+    pub buffer: &'a B::Buffer,
+    pub buffer_offset: Offset,
+    pub buffer_stride: Offset,
+}
+
+// TODO doc
+#[derive(Debug)]
+pub struct AabbPositions {
+    /// Use `mint::Vector3` if available.
+    pub min: [f32; 3],
+    pub max: [f32; 3],
+}
+
+/// https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureGeometryInstancesDataKHR.html
+#[derive(Debug)]
+pub struct GeometryInstances<'a, B: Backend> {
+    /// Must point to a buffer with buffer with an array of `Instance`s
+    /// TODO: document lifetime required for this buffer
+    /// TODO this struct also allows passing an array of pointers, idk if that makes sense outside the host operations case
+    pub buffer: &'a B::Buffer,
+    pub buffer_offset: Offset,
+}
+
+bitflags! {
+    /// https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkGeometryInstanceFlagBitsKHR.html
+    pub struct InstanceFlags: u32 {
+        const TRIANGLE_FACING_CULL_DISABLE_BIT = 0x1;
+        const TRIANGLE_FRONT_COUNTERCLOCKWISE_BIT = 0x2;
+        const FORCE_OPAQUE_BIT = 0x4;
+        const FORCE_NO_OPAQUE_BIT = 0x8;
+    }
+}
+
+// TODO AFAIK rust doesn't have custom sized fields, so we'll need some binary writer wrapper to actually support this at an API level.
+#[derive(Debug)]
+pub struct Instance {
+    transform: TransformMatrix,
+    /// 24 bits
+    instance_custom_index: u32,
+    /// 8 bits visibility mask for the geometry. The instance may only be hit if rayMask & instance.mask != 0
+    mask: u32,
+    /// 24 bit
+    instance_shader_binding_table_record_offset: u32,
+    /// 8 bits
+    flags: InstanceFlags,
+
+    /// either B::AccelerationStructure (host operations) or GPU address (buffer + offset?)
+    acceleration_structure_reference: u64,
+}
+
+#[derive(Debug)]
+pub struct AccelerationStructureSizeRequirements {
+    pub acceleration_structure_size: u64,
+    pub update_scratch_size: u64,
+    pub build_scratch_size: u64,
+}
+
+#[derive(Debug)]
+pub enum AccelerationStructureCopyMode {
+    Clone,
+    Compact,
+
+    // these are subject to the device supporting serialization--also (at least on DX12) mainly used for debug tooling and not load perf--I'm unsure what an end-user use case looks like.
+    Serialize,
+    Deserialize,
+}
+
+#[derive(Debug)]
+pub struct AccelerationStructureBuildRangeDesc {
+    pub primitive: Range<u32>,
+    pub first_vertex: u32,
+    /// "defines an offset in bytes into the memory where a transform matrix is defined.""
+    /// TODO: this is documented as only used for `Geometry::Triangles`, but I have no idea why. GeometryTriangles::transform is implied to point to exactly one transform matrix, but I guess this pattern would allow for passing a list of transforms to reuse the same geometry (but not go through the TLAS' Instance::transform...). DX12 doesn't require any additional info at build time and thus doesn't have this concept--passing 0 here in the vulkan case would probably meet parity with DX12
+    pub transform_offset: u32,
+}
--- a/src/hal/src/command/mod.rs
+++ b/src/hal/src/command/mod.rs
@@ -17,7 +17,7 @@ mod clear;
 mod structs;
 
 use crate::{
-    buffer,
+    acceleration_structure, buffer,
     image::{Filter, Layout, SubresourceRange},
     memory::{Barrier, Dependencies},
     pass, pso, query, Backend, DrawCount, IndexCount, IndexType, InstanceCount, TaskCount,
@@ -603,6 +603,46 @@ pub trait CommandBuffer<B: Backend>: fmt::Debug + Any + Send + Sync {
 
     /// Requests a timestamp to be written.
     unsafe fn write_timestamp(&mut self, stage: pso::PipelineStage, query: query::Query<B>);
+
+    /// TODO docs
+    /// `build_range_infos` must be the same length as `infos` and each element must have a length equal to the parallel `info` entry's `geometries.len()`. there's probably a way to make this safer without mangling the api shape too much...
+    unsafe fn build_acceleration_structures(
+        &self,
+        infos: &[acceleration_structure::AccelerationStructureGeometryDesc<B>],
+        build_range_infos: &[&[acceleration_structure::AccelerationStructureBuildRangeDesc]],
+    );
+
+    /// TODO docs
+    /// `indirect_device_addresses` is an array of infoCount buffer device addresses which point to infos[i]->geometryCount VkAccelerationStructureBuildRangeInfoKHR structures defining dynamic offsets to the addresses where geometry data is stored, as defined by infos[i].
+    unsafe fn build_acceleration_structures_indirect(
+        &self,
+        infos: &[acceleration_structure::AccelerationStructureGeometryDesc<B>],
+
+        indirect_device_addresses: &[(B::Buffer, buffer::Offset)],
+        indirect_strides: &[u32],
+        max_primitive_counts: &[&u32],
+    );
+
+    /// TODO docs
+    unsafe fn copy_acceleration_structure(
+        &self,
+        src: B::AccelerationStructure,
+        dst: B::AccelerationStructure,
+        mode: acceleration_structure::AccelerationStructureCopyMode,
+    );
+
+    // for host ops?
+    // - copy_acceleration_structure_to_memory
+    // - copy_memory_to_acceleration_structure
+
+    /// TODO docs
+    unsafe fn write_acceleration_structures_properties(
+        &self,
+        structures: &[B::AccelerationStructure],
+        query_type: query::Type,
+        pool: &B::QueryPool,
+        first_query: u32,
+    );
 
     /// Modify constant data in a graphics pipeline. Push constants are intended to modify data in a
     /// pipeline more quickly than a updating the values inside a descriptor set.
--- a/src/hal/src/device.rs
+++ b/src/hal/src/device.rs
@@ -12,7 +12,7 @@
 //! and is used to actually do things.
 
 use crate::{
-    buffer, format, image,
+    acceleration_structure, buffer, format, image,
     memory::{Requirements, Segment},
     pass,
     pool::CommandPoolCreateFlags,
@@ -734,6 +734,38 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
         flags: query::ResultFlags,
     ) -> Result<bool, WaitError>;
 
+    /// Create an acceleration structure object.
+    unsafe fn create_acceleration_structure(
+        &self,
+        desc: &acceleration_structure::AccelerationStructureDesc<B>,
+    ) -> Result<B::AccelerationStructure, OutOfMemory>;
+
+    /// Destroy an acceleration structure object.
+    unsafe fn destroy_acceleration_structure(
+        &self,
+        acceleration_structure: B::AccelerationStructure,
+    );
+
+    /// Get the size requirements for the buffers needed to build an acceleration structure.
+    unsafe fn get_acceleration_structure_build_requirements(
+        &self,
+        // used for host operations
+        // build_type: acceleration_structure::HostOrDevice,
+        build_info: &acceleration_structure::AccelerationStructureGeometryDesc<B>,
+        // must be a parallel array to `build_info.geometries` containing the primitive counts for each geometry.
+        max_primitives_count: &[u32],
+    ) -> acceleration_structure::AccelerationStructureSizeRequirements;
+
+    // TODO these are used for host operations, which DX12 does not support.
+    // - build_acceleration_structures
+    // - copy_acceleration_structure
+    // - copy_acceleration_structure_to_memory
+    // - copy_memory_to_acceleration_structure
+    // - write_acceleration_structures_properties
+
+    // TODO for checking if a serialized blob is valid with the current driver version. DX12 docs say this is for PIX tooling and building a as from scratch is "likely to be faster than loading one from disk" (https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_raytracing_acceleration_structure_copy_mode). The Vulkan spec/literature doesn't mention perf implications of serialization.
+    // - get_device_acceleration_structure_compatibility
+
     /// Wait for all queues associated with this device to idle.
     ///
     /// Host access to all queues needs to be **externally** sycnhronized!
@@ -774,4 +806,11 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
     /// Associate a name with a pipeline layout, for easier debugging in external tools or with
     /// validation layers that can print a friendly name when referring to objects in error messages
     unsafe fn set_pipeline_layout_name(&self, pipeline_layout: &mut B::PipelineLayout, name: &str);
+    /// Associate a name with an acceleration structure, for easier debugging in external tools or with
+    /// validation layers that can print a friendly name when referring to objects in error messages
+    unsafe fn set_acceleration_structure_name(
+        &self,
+        acceleration_structure: &mut B::AccelerationStructure,
+        name: &str,
+    );
 }
--- a/src/hal/src/lib.rs
+++ b/src/hal/src/lib.rs
@@ -48,6 +48,7 @@ extern crate serde;
 
 use std::{any::Any, fmt, hash::Hash};
 
+pub mod acceleration_structure;
 pub mod adapter;
 pub mod buffer;
 pub mod command;
@@ -261,6 +262,19 @@ bitflags! {
         const TASK_SHADER = 0x0001 << 96;
         /// Supports mesh shader stage.
         const MESH_SHADER = 0x0002 << 96;
+
+        /// Supports acceleration structures.
+        const ACCELERATION_STRUCTURE = 0x0001 << 112;
+        // TODO Is this the right place to put these features?
+
+        // TODO "capture replay" is a vulkan term, which seems to be covered by PIX and AS (de)serialization on the DX side... should this just be an impl detail of gfx and not be exposed to end-users?
+        // const ACCELERATION_STRUCTURE_CAPTURE_REPLAY = 0x0001 << 112;
+        // TODO this is probably a good idea and supported on both APIs
+        // const ACCELERATION_STRUCTURE_CAPTURE_INDIRECT_BUILD = 0x0001 << 112;
+        // TODO this could be a way to gate this feature
+        // const ACCELERATION_STRUCTURE_CAPTURE_HOST_OPERATIONS = 0x0001 << 112;
+        // TODO this is not supported for other resource types yet
+        // const ACCELERATION_STRUCTURE_UPDATE_AFTER_BIND = 0x0001 << 112;
     }
 }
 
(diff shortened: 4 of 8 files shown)


src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,232 @@
...
+    BottomLevel,
+    // TODO vulkan supports "generic level" (where the concrete build type is specified), but discourages its use for applications "written directly for Vulkan" since it "could affect capabilities or performance in the future" (https://www.khronos.org/blog/vulkan-ray-tracing-final-specification-release). Perhaps this is to better support `vkd3d-proton`, but we probably don't want it exposed in gfx?
+}
+
+/// A description of the data needed to build an acceleration structure.
+#[derive(Debug)]
+pub struct AccelerationStructureDesc<'a, B: Backend> {
+    /// TODO: document lifetime required for this buffer
+    pub buffer: &'a B::Buffer,
+    /// TODO: the final gpu address on DX12 needs to be D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BYTE_ALIGNMENT (256), but vulkan only requires this offset to be 256 (as defined by https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureCreateInfoKHR.html)
```
tangmi, line 25: `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BYTE_ALIGNMENT` is a `#define`, is it reasonable to assume this (and the 256 from the Vulkan spec) is a "forever constant" that we can inherit as a static value in gfx?

src/hal/src/device.rs
```diff
@@ -734,6 +734,38 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
         flags: query::ResultFlags,
     ) -> Result<bool, WaitError>;
 
+    /// Create an acceleration structure object.
```
tangmi, line 737: Open question: does gfx have a plan on how to handle API extensions?

I currently plan on implementing this like the mesh shaders are implemented (i.e. have an implicit requirement that the `Feature` flag is available to be able to successfully call relevant methods). In my prototyping I realize this could become a maintenance burden (e.g. all non-supported backends need to stub out the methods and it's not extremely obvious which types and functions on device/commandlist are relevant to which extensions).

I can see a possibility with the extension trait pattern (unsupported backends would still need to stub out implementations, but it'd be very clear that e.g. ray tracing is not supported), or have the device/command list have a function to optionally return an extension-specific object with the relevant functions on it (unsupported backends can just return `None`, but I suspect implementing/using an API like this will be more cumbersome).

src/hal/src/device.rs
```diff
@@ -734,6 +734,38 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+
+    // TODO these are used for host operations, which DX12 does not support.
+    // - build_acceleration_structures
+    // - copy_acceleration_structure
+    // - copy_acceleration_structure_to_memory
+    // - copy_memory_to_acceleration_structure
+    // - write_acceleration_structures_properties
+
+    // TODO for checking if a serialized blob is valid with the current driver version. DX12 docs say this is for PIX tooling and building a as from scratch is "likely to be faster than loading one from disk" (https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_raytracing_acceleration_structure_copy_mode). The Vulkan spec/literature doesn't mention perf implications of serialization.
+    // - get_device_acceleration_structure_compatibility
```
tangmi, line 767: This function operates basically the same in Vulkan (VkGetDeviceAccelerationStructureCompatibility) and DX (CheckDriverMatchingIdentifier). Both take a pointer to the serialized data and read two GUIDs (32 bytes) to see if it matches the current driver (DX can signal slightly more info than VK).

On the comment in `AccelerationStructureCopyMode`, I wonder if serializing an acceleration structure is something we need to expose as a part of the API. I can't find any literature on why one would want to do this in Vulkan, but as said [mentioned link](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_raytracing_acceleration_structure_copy_mode), DX mainly uses serialization for PIX and not to save on build time. If this is also true for Vulkan (like RenderDoc wants it or something), we may be able to leave this off the API and handle the debug tooling case as a backend implementation detail (i.e. I don't think people are writing debug tooling targetting the gfx api yet)

src/hal/src/device.rs
```diff
@@ -734,6 +734,38 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    unsafe fn get_acceleration_structure_build_requirements(
+        &self,
+        // used for host operations
+        // build_type: acceleration_structure::HostOrDevice,
+        build_info: &acceleration_structure::AccelerationStructureGeometryDesc<B>,
+        // must be a parallel array to `build_info.geometries` containing the primitive counts for each geometry.
+        max_primitives_count: &[u32],
+    ) -> acceleration_structure::AccelerationStructureSizeRequirements;
+
+    // TODO these are used for host operations, which DX12 does not support.
```
tangmi, line 759: Note these functions (in vulkan) optionally take a deferral (`VK_KHR_deferred_host_operations`), which I don't think gfx has a concept in its API for..?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkGeometryInstanceFlagBitsKHR.html
+    pub struct InstanceFlags: u32 {
+        const TRIANGLE_FACING_CULL_DISABLE_BIT = 0x1;
+        const TRIANGLE_FRONT_COUNTERCLOCKWISE_BIT = 0x2;
+        const FORCE_OPAQUE_BIT = 0x4;
+        const FORCE_NO_OPAQUE_BIT = 0x8;
+    }
+}
+
+// TODO AFAIK rust doesn't have custom sized fields, so we'll need some binary writer wrapper to actually support this at an API level.
```
tangmi, line 191: Maybe [`packed_struct`](https://docs.rs/packed_struct/0.4.0/packed_struct/) or [Pre-RFC: Generic integers (uint<N> and int<N>)](https://internals.rust-lang.org/t/pre-rfc-generic-integers-uint-n-and-int-n/7641) could be applicable here?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,232 @@
...
+#[derive(Debug)]
+pub struct AccelerationStructureSizeRequirements {
+    pub acceleration_structure_size: u64,
+    pub update_scratch_size: u64,
+    pub build_scratch_size: u64,
+}
+
+#[derive(Debug)]
+pub enum AccelerationStructureCopyMode {
+    Clone,
```
tangmi, line 217: bikeshed: "clone" is the vk terminology, but I think in Rust it implies a deep copy, which I don't believe is what the acceleration structure copy command does.

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,232 @@
...
+    Serialize,
+    Deserialize,
+}
+
+#[derive(Debug)]
+pub struct AccelerationStructureBuildRangeDesc {
+    pub primitive: Range<u32>,
+    pub first_vertex: u32,
+    /// "defines an offset in bytes into the memory where a transform matrix is defined.""
+    /// TODO: this is documented as only used for `Geometry::Triangles`, but I have no idea why. GeometryTriangles::transform is implied to point to exactly one transform matrix, but I guess this pattern would allow for passing a list of transforms to reuse the same geometry (but not go through the TLAS' Instance::transform...). DX12 doesn't require any additional info at build time and thus doesn't have this concept--passing 0 here in the vulkan case would probably meet parity with DX12
```
tangmi, line 230: This is a case where my gut says we should follow DX's model of knowing the vertex/index counts upfront (during Device::create_acceleration_structure vs CommandList::build_acceleration_structure) and just not exposing vulkan's `transform_offset`. I'm open to ideas on how to implement this nicely in DX, however!

src/hal/src/command/mod.rs
```diff
@@ -604,6 +604,46 @@ pub trait CommandBuffer<B: Backend>: fmt::Debug + Any + Send + Sync {
     /// Requests a timestamp to be written.
     unsafe fn write_timestamp(&mut self, stage: pso::PipelineStage, query: query::Query<B>);
 
+    /// TODO docs
+    /// `build_range_infos` must be the same length as `infos` and each element must have a length equal to the parallel `info` entry's `geometries.len()`. there's probably a way to make this safer without mangling the api shape too much...
```
tangmi, line 608: These are some spicy requirements, but the API is unsafe, so I guess it could be argued to be reasonable?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    pub src: Option<B::AccelerationStructure>,
+    pub dst: Option<B::AccelerationStructure>,
+
+    // TODO: We could enforce the following the type system?
+    // - in both vulkan (via `VUID-VkAccelerationStructureBuildGeometryInfoKHR-type-03792`) and DX12 (via type system), all of the structs here must be the same variant.
+    // - blas must be triangles or aabbs, tlas must be instances
+    pub geometries: &'a [AccelerationStructureGeometry<'a, B>],
+    // Both APIs support "array" and "array of pointers", presumably to allow for cheap reuse of the geometry descriptors.
+    // pgeometries: &'a [&'a GeometryDesc<'a, B>],
```
tangmi, line 80: Should I have something like

```rust
pub enum GeometriesList<'a, B: Backend> {
   Array(&'a [AccelerationStructureGeometry<'a, B>]),
   ArrayOfRef(&'a [&'a AccelerationStructureGeometry<'a, B>]),
}
```

to denote this difference?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+    Build,
+    /// specifies that the destination acceleration structure will be built using data in a source acceleration structure, updated by the specified geometries.
+    /// Note there's constraints on update: https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#acceleration-structure-update-constraints
+    Update,
+}
+
+/// A description of the geometry data needed to populate an acceleration structure.
+///
+/// TODO: there's something here that smells w/ what fields are needed to get the required build size vs what fields are needed to actually build. Also, the top/bottom levels having different requirements on which fields are valid.
+// TODO: I don't like that this is reused for the "get build sizes" and "do actual build" with complex rules on what fields are ignored when. Perhaps we could use DX12's model for `D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC` and `D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS` and move at least the src/dst fields onto the actual "build acceleration structure" command.?
```
tangmi, line 64: This is the struct I'm most interested in feedback for. Mentioned in the many TODO comments on and around it, this is a kind of central type which most examples I've seen just keep around a mutable version of this and update fields as needed by methods.

I think we can move the src/dst/scratch values onto `build_acceleration_structures*` or have another type that has src/dst/scratch and a `AccelerationStructureGeometryDesc`.

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+use crate::{buffer::Offset, format::Format, Backend, IndexType};
+
+/// Denotes the type of acceleration structure.
+#[derive(Debug)]
+pub enum AccelerationStructureType {
+    ///
+    TopLevel,
+    ///
+    BottomLevel,
+    // TODO vulkan supports "generic level" (where the concrete build type is specified), but discourages its use for applications "written directly for Vulkan" since it "could affect capabilities or performance in the future" (https://www.khronos.org/blog/vulkan-ray-tracing-final-specification-release). Perhaps this is to better support `vkd3d-proton`, but we probably don't want it exposed in gfx?
```
kvark, line 17: Technically, if there is a possibility of running vkd3d over gfx-portability, then we should try to make it possible.
For example, people could use it to run windows games on macOS.
In this case, we can make a call to not support other levels at least for now, just leaving a comment about this omission.

src/hal/src/acceleration_structure.rs
kvark: Sure! Let's just make it a constant here.

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+// TODO remove and add docs
+#![allow(missing_docs)]
+
+use std::ops::Range;
+
+use crate::{buffer::Offset, format::Format, Backend, IndexType};
+
+/// Denotes the type of acceleration structure.
+#[derive(Debug)]
+pub enum AccelerationStructureType {
```
kvark, line 12: if we are in the `acceleration_structure` module, we don't have to prepend all the types with `AcceleratedStructure` prefix. I know it can be useful for doing things like `use acceleration_structure::{A, B}` without renaming `A` and `B`, but so far in gfx codebase we preferred a style where the users would just do `use hal::acceleration_structure as accel` and access the contents as `accel::Type`, `accel::Descriptor`, etc.
So I think these prefixes should be removed.

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+/// A description of the geometry data needed to populate an acceleration structure.
+///
+/// TODO: there's something here that smells w/ what fields are needed to get the required build size vs what fields are needed to actually build. Also, the top/bottom levels having different requirements on which fields are valid.
+// TODO: I don't like that this is reused for the "get build sizes" and "do actual build" with complex rules on what fields are ignored when. Perhaps we could use DX12's model for `D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC` and `D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS` and move at least the src/dst fields onto the actual "build acceleration structure" command.?
+#[derive(Debug)]
+pub struct AccelerationStructureGeometryDesc<'a, B: Backend> {
+    pub flags: AccelerationStructureFlags,
+    /// The type of acceleration structure to build.
+    pub ty: AccelerationStructureType,
+    pub mode: AccelerationStructureCreateMode,
```
kvark, line 70: Could we remove that? If the `src` is `Some`, it has to be an update.

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    pub src: Option<B::AccelerationStructure>,
+    pub dst: Option<B::AccelerationStructure>,
```
kvark, line 73: shouldn't this be mandatory?

src/hal/src/acceleration_structure.rs
kvark: That sounds appropriate, let's try that!
I assume the Vulkan backend would be able to build this stuff internally with no issues if we separate it and follow DX12 style here?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    // TODO: We could enforce the following the type system?
```
kvark, line 75: Seems worth a try to see how this would look.

A small tale from a similar issue with had with descriptor updates. Vulkan requires all the descriptors within a "write" to be of the same type (and other parameters). The backends didn't end up taking advantage of this though, and `wgpu-core` as the main user didn't enjoy this either. So the restriction only put unnecessary friction between `wgpu-core` and the backends, making the API not do its function best for this path.

In the end, we just relaxed the rules in gfx-hal itself and allowed different types of descriptors to be passed in. This required almost no extra logic in Vulkan backend, and it resolved a concern on whether we should be type-enforcing this.

src/hal/src/acceleration_structure.rs
kvark: Hmm. That seems a bit weird, although I don't know yet how `AccelerationStructureGeometry` will be stored in practice.
Just a shot in the dark - would `Cow<'a, AccelerationStructureGeometry>` be possible here?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+pub struct AccelerationStructureGeometry<'a, B: Backend> {
+    pub flags: AccelerationStructureGeometryFlags,
+    pub geometry: Geometry<'a, B>,
+}
+
+#[derive(Debug)]
+pub enum Geometry<'a, B: Backend> {
+    Triangles(GeometryTriangles<'a, B>),
+
+    /// TODO bikeshed capitalization of AABBs.
```
kvark, line 104: Your choice matches Rust guidelines, I think

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub vertex_format: Format,
+    pub vertex_buffer: &'a B::Buffer,
+    pub vertex_buffer_offset: Offset,
+    pub vertex_buffer_stride: Offset,
+
+    /// aka "vertex count"?
+    pub max_vertex: Offset,
+
+    // index format must be DXGI_FORMAT_R32_UINT, DXGI_FORMAT_R16_UINT
+    // can also be VK_INDEX_TYPE_NONE_KHR/DXGI_FORMAT_UNKNOWN
```
kvark, line 131: Should it be `Option<IndexType>` then? Or is the whole `index_buffer: None` supposed to match `VK_INDEX_TYPE_NONE_KHR`?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub index_buffer: Option<(&'a B::Buffer, Offset, IndexType)>,
+
+    /// 3x4 matrix
+    /// TODO enum for cpu repr?
```
kvark, line 135: only Metal has this different? We can work around that by doing a copy, I think (either a transfer, or in compute)

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// Must point to a buffer with buffer with an array of `AabbPositions`s
+    /// TODO: document lifetime required for this buffer
+    pub buffer: &'a B::Buffer,
+    pub buffer_offset: Offset,
+    pub buffer_stride: Offset,
+}
+
+#[derive(Debug)]
+pub struct AabbPositions {
+    /// Use `mint::Vector3` if available.
```
kvark, line 158: we aren't using `mint` for gfx-hal at this point

src/hal/src/acceleration_structure.rs
kvark: gfx-hal is a low-level API. It's fine to cut some corners and just document what bits are expected within `u32`.
Extra dependencies are discouraged

src/hal/src/acceleration_structure.rs
kvark: Indeed, let's rename. `Shallow`? `ShallowClone`?

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    // these are subject to the device supporting serialization--also (at least on DX12) mainly used for debug tooling and not load perf--I'm unsure what an end-user use case looks like.
+    Serialize,
+    Deserialize,
+}
+
+#[derive(Debug)]
+pub struct AccelerationStructureBuildRangeDesc {
+    pub primitive: Range<u32>,
+    pub first_vertex: u32,
```
kvark, line 220: could use `VertexCount` typedef that we have?

src/hal/src/acceleration_structure.rs
kvark: if we go with this, would we be able to implement `gfx-portability`'s side of the KHR raytracing support?

src/hal/src/command/mod.rs
kvark: If they need to be the same length, and we aren't passing the pointers to these arrays directly in any of the backends anyway, why not make this one array? I.e.
```rust
infos: &[(acceleration_structure::AccelerationStructureGeometryDesc<B>, &[acceleration_structure::AccelerationStructureBuildRangeDesc])],
```

kvark: Also, we should probably make this an iterator, like in other commands (see `wait_events`)

src/hal/src/command/mod.rs
```diff
@@ -604,6 +604,46 @@ pub trait CommandBuffer<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    unsafe fn build_acceleration_structures(
+        &self,
+        infos: &[acceleration_structure::AccelerationStructureGeometryDesc<B>],
+        build_range_infos: &[&[acceleration_structure::AccelerationStructureBuildRangeDesc]],
+    );
+
+    /// TODO docs
+    /// `indirect_device_addresses` is an array of infoCount buffer device addresses which point to infos[i]->geometryCount VkAccelerationStructureBuildRangeInfoKHR structures defining dynamic offsets to the addresses where geometry data is stored, as defined by infos[i].
```
kvark, line 616: similarly, can we merge all the things that are of the same length into a structure, so that it's an iterator of this structure that the user provides?

src/hal/src/command/mod.rs
```diff
@@ -604,6 +604,46 @@ pub trait CommandBuffer<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    unsafe fn build_acceleration_structures_indirect(
```
kvark, line 617: we may consider just having default implementations here with `unimplemented!` given how few backends would actually care to implement this

src/hal/src/command/mod.rs
```diff
@@ -604,6 +604,46 @@ pub trait CommandBuffer<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    /// TODO docs
+    unsafe fn copy_acceleration_structure(
+        &self,
+        src: B::AccelerationStructure,
+        dst: B::AccelerationStructure,
+        mode: acceleration_structure::AccelerationStructureCopyMode,
+    );
+
+    // for host ops?
+    // - copy_acceleration_structure_to_memory
```
kvark, line 635: names look fine to me!

src/hal/src/device.rs
kvark: Having traits would seem like a very nice quality of life improvement, but in practice for anything that tries to abstract over gfx-rs backends (i.e. `wgpu-core`) it would be a huge pain. Run-time flags are just more versatile and predictable for such a low-level library.

We could have the default implementations (doing `unimplemented!()`) to save on the boilerplate.

src/hal/src/device.rs
kvark: correct, we don't have anything like this yet...

src/hal/src/device.rs
kvark: Wouldn't you want a ray-traced scene to load faster if the data like this is loaded from disk?
I'd be totally fine just postponing any serialization stuff if you don't consider it useful.

src/hal/src/lib.rs
```diff
@@ -261,6 +262,20 @@ bitflags! {
         const TASK_SHADER = 0x0001 << 96;
         /// Supports mesh shader stage.
         const MESH_SHADER = 0x0002 << 96;
+
+        /// Supports acceleration structures.
+        const ACCELERATION_STRUCTURE = 0x0001 << 112;
+
+        // TODO Is this the right place to put these features? These are derived from `VkPhysicalDeviceAccelerationStructureFeaturesKHR`.
```
kvark, line 269: right place, yes

src/hal/src/lib.rs
```diff
@@ -631,4 +662,8 @@ pub trait Backend: 'static + Sized + Eq + Clone + Hash + fmt::Debug + Any + Send
     type Event: fmt::Debug + Any + Send + Sync;
     /// The corresponding query pool type for this backend.
     type QueryPool: fmt::Debug + Any + Send + Sync;
+
+    /// The corresponding acceleration structure type for this backend.
+    // TODO: This probably corresponds to `()` on DX12, since there's no handle for acceleration structures on DX12
```
kvark, line 667: I thought it would be a buffer on DX12?

src/hal/src/acceleration_structure.rs
tangmi: I like this style you've described, I did notice (in the Vulkan API) my eyes start to glaze over with everything with the same (long) prefix.

src/hal/src/acceleration_structure.rs
tangmi: Nice catch, I like this more

src/hal/src/acceleration_structure.rs
tangmi: This field is ignored during the "get sizes" command (and also not possible to have an AS before one queries for the required size). Per the struct-level discussion, I'll move this to the build method so it won't be an issue anymore here.

src/hal/src/acceleration_structure.rs
tangmi: That should be correct.

Tangentially related: I'd wager we won't be able to reuse vulkan or dx structs across calls that effectively (i.e. the user will have a gfx *Desc, rather than a vulkan or dx *Desc struct), so I assume that we're okay with recreating these descriptor structs as needed?

src/hal/src/acceleration_structure.rs
tangmi: I totally get that. I believe this is a case where all the backends will have this same requirement and it could be impractical to implement a relaxed version (we'd potentially need to partition an acceleration structure into multiple based on geometry type?). Note that nothing prevents vulkan or dx from being implemented as it's listed currently.

I suspect in this case we may be able to do it without much (any?) "clever" type system trickery. For reference, this is how DX12 enforces it in a C API: [D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS ](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_build_raytracing_acceleration_structure_inputs), basically moving the union tag outside of the list rather than on each element.

At the very least, without safety at the type-level, I want to make sure we get docs onto this in gfx to make sure users know about this restriction.

src/hal/src/acceleration_structure.rs
tangmi: It could be Cow time!

Now that I think about this, this distinction probably doesn't actually matter, since our `AccelerationStructureGeometry` definitely won't be exactly the same as both the VK and DX versions and this Array/ArrayOfRef option is definitely meant to avoid unnecessary copies of these structs. Since we're exposing these descriptor structs as rust structs, we'll probably have to copy anyways... 

We probably want `geometries: &'a [&'a AccelerationStructureGeometry<'a, B>]` here on the rust side.

src/hal/src/acceleration_structure.rs
tangmi: I have no strong opinions on naming. I'll leave as `Aabbs` but am happy to change in the future.

src/hal/src/acceleration_structure.rs
tangmi: I believe `VK_INDEX_TYPE_NONE_KHR` is meant to signal that there's no index data (and it's just a triangles list), since `VkBuffer` doesn't have a null value to achieve the same result.

The valid usage requirements for `vkCmdBuildAccelerationStructuresKHR` hints at this but doesn't say it explicitly.
- https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03768
- https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03806
- https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03807
- https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03712

src/hal/src/acceleration_structure.rs
tangmi: I'll reword all mint references to describe the vector kind

src/hal/src/acceleration_structure.rs
tangmi: Any strong opinions against just `Copy`? I do think simpler is probably better, but I don't think even `Clone` would confuse many people familiar with gfx/vulkan...

src/hal/src/acceleration_structure.rs
tangmi: My concern here is that we probably want this struct to be bit-for-bit compatible with [`VkAccelerationStructureInstanceKHR`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureInstanceKHR.html) and [`D3D12_RAYTRACING_INSTANCE_DESC`](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_raytracing_instance_desc) so users can create an array of `acceleration_structure::Instance` and upload it directly to a GPU buffer. Otherwise, we need to settle for describing this carefully and letting users BYO packed struct impl... thoughts?

The Metal backend will probably have to do something special to repack into its own special format, however

src/hal/src/acceleration_structure.rs
tangmi: I believe this `transform` field is only read for presence during the "get sizes" phase and only actually read during the "build" phase. We should be in a command list at that point, so it's feasible to emulate the vulkan behavior like you say.

FYI, I think the table looks something like this:

| | Vulkan | DX12 | Metal |
| - | - | - | - |
| CPU repr | Y |  |  |
| GPU repr | Y | Y | Y* |

VK: [`VkDeviceOrHostAddressConstKHR transformData`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureGeometryTrianglesDataKHR.html)
DX: [`D3D12_GPU_VIRTUAL_ADDRESS Tansform3x4`](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_raytracing_geometry_triangles_desc)
MTL: *only on the "instance" type: [`transformBuffer: MTLBuffer?`](https://developer.apple.com/documentation/metalperformanceshaders/mpsinstanceaccelerationstructure)

Note in the vulkan link that it allows for all pointers in the `AccelerationStructureGeometryDesc` (i.e. `VkAccelerationStructureBuildGeometryInfoKHR`) hierarchy to be `VkDeviceOrHostAddressConstKHR`.

Do you think this is worth doing now or could I follow-up after a DX impl (I suspect this feature will be a bit of work there)?

src/hal/src/acceleration_structure.rs
tangmi: Hm... I guess we should support this for `gfx-portability`... I'll leave this alone for now, but we'll have to come up with a way to implement this on dx12. Maybe we can look at `vkd3d-proton` and take reverse inspiration?

src/hal/src/lib.rs
tangmi: You're absolutely right, I think this was a 3am brain fart

src/hal/src/command/mod.rs
tangmi: Any suggestions on the extra requirement (also for the indirect variant) of:

```rust
for (info, build_range_infos) in infos.iter().zip(build_range_infos.iter()) {
    assert_eq!(info.geometries.len(), build_range_infos.len());
}
```

src/hal/src/acceleration_structure.rs
tangmi: I'll add this back. I wonder if there's a nicer way to handle `Generic`, as it's allowed in the "get sizes" and "create" functions, but not during the "build" commands.

src/hal/src/acceleration_structure.rs
tangmi: Oops, I may need to backpedal on that thumbs up... This struct's layout is depended on by `build_acceleration_structures_indirect` so I'll prefer to specify it explicitly here (and flatten `Range<u32>`). I don't expect `VertexCount` to change sizes, but if it does it'll likely cause strange problems.

src/hal/src/command/mod.rs
tangmi: That's a good idea. I think I'll defer figuring out which ones to later!

src/hal/src/command/mod.rs
tangmi: I initially had these commented out since these potentially depend on the [`VkDeviceOrHostAddressConstKHR`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkDeviceOrHostAddressConstKHR.html) type (which I've called `cpu-repr` in the comments), but I think I can support the GPU buffer variant of it for now

src/hal/src/device.rs
tangmi: Sounds good. I'll try and keep the `wgpu-core` use-case you mentioned in mind

src/hal/src/device.rs
tangmi: AFAIK that extension is only used by acceleration structure host commands, but regardless I've made a PR in ash https://github.com/MaikKlein/ash/pull/348 so hopefully we'll be able to use it by the next release of ash

src/hal/src/acceleration_structure.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// The buffer containing the AABB data.
+    ///
+    /// The buffer must contain a list of `AabbPositions`.
+    pub buffer: &'a B::Buffer,
+
+    /// The offset into `buffer`.
+    pub buffer_offset: Offset,
+
+    /// The stride of the AABB data in `buffer`.
+    pub buffer_stride: Offset,
```
kvark, line 176: now we can use your new typedef!

kvark (commented):
looks like github is eating the comments :(

kvark (approved):

src/hal/src/device.rs
tangmi: Do you know of a vulkan resource where we can ask about the perf implications of serialization (or just general "best practices")? The DX docs (for `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_COPY_MODE_DESERIALIZE`) seem to imply this isn't a perf win:

> This mode is only intended for tools such as PIX, though nothing stops any app from using it, but this mode is only permitted when developer mode is enabled in the OS. This copy operation is not intended to be used for caching acceleration structures, because running a full acceleration structure build is likely to be faster than loading one from disk.

src/hal/src/acceleration_structure.rs
kvark: That's correct, yes. We'll be recreating them all the time.

src/hal/src/command/mod.rs
kvark: an `assert_eq` sounds ok, for now at least.

src/hal/src/acceleration_structure.rs
kvark: No strong opinion, `Copy` is fine :)

src/hal/src/acceleration_structure.rs
kvark: I did mean preserving the bit-by-bit if this needs to go into the driver. Just instead of doing bitfields, we could have raw unsigned integers with documented layouts - that's what I meant by cutting the corners.

src/hal/src/device.rs
kvark: Khronos Slack would be the best place to ask

examples/Cargo.toml
```diff
@@ -30,13 +30,18 @@ path = "compute/main.rs"
...
+path = "ray-tracing/main.rs"
+
 [dependencies]
 image = "0.23.12"
 log = "0.4"
 hal = { path = "../src/hal", version = "0.6", package = "gfx-hal" }
 auxil = { path = "../src/auxil/auxil", version = "0.5", package = "gfx-auxil" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.6" }
 winit = { version = "0.23", features = ["web-sys"] }
+wavefront_obj = "9.0"
```
kvark, line 44: would you mind using https://crates.io/crates/obj instead? it's something we have more control of

src/backend/dx12/src/command.rs
```diff
@@ -2829,6 +2829,75 @@ impl com::CommandBuffer<Backend> for CommandBuffer {
         );
     }
 
+    unsafe fn build_acceleration_structures<'a, I>(&self, _descs: I)
```
kvark, line 2832: are you sure you don't want to just make this a default implementation in the trait, to avoid all the boilerplate in backends that don't support it?

examples/Cargo.toml
tangmi: I don't technically need to load an `obj` file at all for this change, this is just an easy way to get a bunch of data into the API to (hopefully) smoke test it a bit better than just a cube.

Ideally I'll do like a cornell box or something, which seems reasonable to do programmatically and should show off ray tracing pretty well (or maybe just a triangle?).

src/backend/dx12/src/command.rs
tangmi: I've undone the backend stubs for now and will make a default "unimplemented" implementation in `gfx-hal`.

examples/Cargo.toml
kvark: That may end up being a whole lot of code. I.e. you'd need something like `genmesh`, and debugging would be even more difficult. I think loading an `obj` is perfectly reasonable here.

tangmi: Sounds good. Above all else, I want it to be understandable 😁

kvark:
> Merge remote-tracking branch 'upstream/master' into acceleration-stru… 

Please try to rebase instead of merging, to keep history linear

tangmi:
> 
> 
> > Merge remote-tracking branch 'upstream/master' into acceleration-stru…
> 
> Please try to rebase instead of merging, to keep history linear

I'll do this when I clean up this draft, please call me out if I forget then!

tangmi:
@kvark the changes in `hal` are ready for review. Please note there's no rush since the backends depend on new releases of upstream deps and the usefulness of this API depends on ray-tracing-pipelines or ray-query being also defined and implemented.

Some things on my mind:
- I'm wondering if it's worth it to commit the API first, then the implementations or hold off until we have at least the Vulkan backend (which requires a new `ash` release `>=0.32`).
  - Note that I'm also not confident in the quality of what I've hacked together here for Vulkan.
- I will turn the leftover TODO comments into separate issues (i.e. work for host operations, capture replay)
- I'll definitely clean up the history (rebase) before the API is merged.
- I'm working on the ray-tracing-pipelines part of the API in a branch: https://github.com/tangmi/gfx/tree/ray-tracing-api
  - I don't have hardware to test ray-query, but I imagine there's very little work in `hal` to support it.

kvark:
Thank you for moving this forward!

First of all, we've been trying to wrap up hal-0.7 release for the last month. I'd like to do this before merging the changes here. Also note how a lot of type signatures are simplified now: `Borrow`, `ExactSizeIterator`, and `IntoIterator` are all gone! If you haven't already, please consider updating this PR to reflect that change with the new API.

> Please note there's no rush since the backends depend on new releases of upstream deps

We are totally cool with depending on github revisions from master. We only switch to releases when necessary, i.e. during a release. So this is not a blocker to merging the PR.

kvark:
@tangmi what's the status of this work?

tangmi:
> 
> 
> @tangmi what's the status of this work?

The current state is that I have an API shape I'm pretty happy with and am in the middle of fixing some issues in the Vulkan impl (I think the last thing I was working on was overthinking mapping lists-of-lists of structs such that all the lifetimes agree).

Unfortunately, life and work has taken me away from this for the last couple months. Also, I (and everyone else) have been having a real hard time getting a hold of a GPU that can test ray query--although if ray tracing pipelines work, it's very likely that ray query will also work. I'm definitely still eager to get back into gfx-rs.

On the plus side, I believe `ash` as since released so I think we don't need to point at some random revision anymore!

kvark:
@tangmi it would be good if gfx-rs organization created a "hardware fund" for cases like this...

tangmi:
@kvark I'd happily contribute to a fund like that--sadly, I just can't find any ray tracing cards for sale at all this year!

(I know there's scalpers around but I feel icky supporting that!)

kvark:
@tangmi now with Vulkan's RT support, aren't there any AMD cards on the market? Sorry, I'm not generally shopping for these.

kvark:
It would be great to chat on #gfx:matrix.org (or a different platform, if you have a preference?) and see if we can unblock your progress.

PR #3545, opened 2020-12-20
Wayland Support for the OpenGL Backend

nemosupremo:
*~~This PR isn't meant to be merged in the current state, but to illustrate what changes were need to get it working.~~*

I've been trying to get the OpenGL backend working with wayland. The current device I'm targeting doesn't currently support Vulkan so I'm trying to get the wayland renderer working. During my tests I'm hitting some blockers with the `hal::Instance` API and I'm wondering if I'm doing something wrong, or if the API just wasn't developed with this in mind.

I'm hitting the following roadblocks:

1. I need a pointer to this display to pass to `get_platform_display`. Both creating a new connection and using `egl::DEFAULT_DISPLAY` results in the window not working or rendering in the `quad` example. To get around this I'm passing a pointer to the create function via the `version` parameter.
2. ~~To create `wl_egl_window_create` I need to pass in a window size. I don't think this is a show stopper as long as there was some way get resize events so I could call `wl_egl_window_resize`.~~ I *think* this can be done by calling `wl_egl_window_resize` in `PresentationSurface::configure_swapchain`

Another thing I had to remove the `GL_COLORSPACE`. This attribute is only supported in EGL 1.5, but currently are only testing 1.4. Seeing how the default value is `GL_COLORSPACE_SRGB`, then I just removed this attribute entirely.

Is there a proper way to overcome these hurdles without drastically modifying gfx's api?

### DIFF
--- a/src/backend/gl/src/window/egl.rs
+++ b/src/backend/gl/src/window/egl.rs
@@ -17,7 +17,6 @@ pub struct Swapchain {
 
 #[derive(Debug)]
 pub struct Instance {
-    egl: Starc<egl::DynamicInstance>,
     wsi_library: Option<libloading::Library>,
     inner: Mutex<Inner>,
 }
@@ -62,6 +61,31 @@ type WlEglWindowResizeFun = unsafe extern "system" fn(
 );
 
 type WlEglWindowDestroyFun = unsafe extern "system" fn(window: *const raw::c_void);
+
+fn open_x_display() -> Option<(ptr::NonNull<raw::c_void>, libloading::Library)> {
+    log::info!("Loading X11 library to get the current display");
+    let library = libloading::Library::new("libX11.so").ok()?;
+    let func: libloading::Symbol<XOpenDisplayFun> =
+        unsafe { library.get(b"XOpenDisplay").unwrap() };
+    let result = unsafe { func(ptr::null()) };
+    ptr::NonNull::new(result).map(|ptr| (ptr, library))
+}
+
+fn test_wayland_display() -> Option<libloading::Library> {
+    /* We try to connect and disconnect here to simply ensure there
+     * is an active wayland display available.
+     */
+    log::info!("Loading Wayland library to get the current display");
+    let client_library = libloading::Library::new("libwayland-client.so").ok()?;
+    let wl_display_connect: libloading::Symbol<WlDisplayConnectFun> =
+        unsafe { client_library.get(b"wl_display_connect").unwrap() };
+    let wl_display_disconnect: libloading::Symbol<WlDisplayDisconnectFun> =
+        unsafe { client_library.get(b"wl_display_disconnect").unwrap() };
+    let display = ptr::NonNull::new(unsafe { wl_display_connect(ptr::null()) })?;
+    unsafe { wl_display_disconnect(display.as_ptr()) };
+    let library = libloading::Library::new("libwayland-egl.so").ok()?;
+    Some(library)
+}
 
 impl Inner {
     fn create(
@@ -210,80 +234,45 @@ impl hal::Instance<crate::Backend> for Instance {
         let mut wsi_library = None;
 
         let wayland_display = if client_ext_list.contains(&"EGL_EXT_platform_wayland") {
-            log::info!("Loading Wayland library to get the current display");
-            if let Ok(library) = libloading::Library::new("libwayland-client.so") {
-                /* We try to connect and disconnect here to simply ensure there
-                 * is an active wayland display available.
-                 */
-                let wl_display_connect: libloading::Symbol<WlDisplayConnectFun> =
-                    unsafe { library.get(b"wl_display_connect").unwrap() };
-                let wl_display_disconnect: libloading::Symbol<WlDisplayDisconnectFun> =
-                    unsafe { library.get(b"wl_display_disconnect").unwrap() };
-                if let Some(display) = ptr::NonNull::new(unsafe { wl_display_connect(ptr::null()) })
-                {
-                    unsafe { wl_display_disconnect(display.as_ptr()) }
-                    if let Ok(library) = libloading::Library::new("libwayland-egl.so") {
-                        Some(((), library))
-                    } else {
-                        None
-                    }
-                } else {
-                    None
-                }
-            } else {
-                None
-            }
+            test_wayland_display()
         } else {
             None
         };
 
         let x11_display = if client_ext_list.contains(&"EGL_EXT_platform_x11") {
-            log::info!("Loading X11 library to get the current display");
-            if let Ok(library) = libloading::Library::new("libX11.so") {
-                let func: libloading::Symbol<XOpenDisplayFun> =
-                    unsafe { library.get(b"XOpenDisplay").unwrap() };
-                let result = unsafe { func(ptr::null()) };
-                ptr::NonNull::new(result).map(|ptr| (ptr, library))
-            } else {
-                None
-            }
+            open_x_display()
         } else {
             None
         };
 
-        let display = match (wayland_display, x11_display) {
-            (Some((_, library)), _) => {
-                log::info!("Using Wayland platform");
-                let display_attributes = [egl::ATTRIB_NONE];
-                wsi_library = Some(library);
-                egl.get_platform_display(
-                    EGL_PLATFORM_WAYLAND_KHR,
-                    egl::DEFAULT_DISPLAY,
-                    &display_attributes,
-                )
-                .unwrap()
-            }
-            (_, Some((x11_display, library))) => {
-                log::info!("Using X11 platform");
-                let display_attributes = [egl::ATTRIB_NONE];
-                wsi_library = Some(library);
-                egl.get_platform_display(
-                    EGL_PLATFORM_X11_KHR,
-                    x11_display.as_ptr(),
-                    &display_attributes,
-                )
-                .unwrap()
-            }
-            _ => {
-                log::info!("Using default platform");
-                egl.get_display(egl::DEFAULT_DISPLAY).unwrap()
-            }
+        let display = if let Some(library) = wayland_display {
+            log::info!("Using Wayland platform");
+            let display_attributes = [egl::ATTRIB_NONE];
+            wsi_library = Some(library);
+            egl.get_platform_display(
+                EGL_PLATFORM_WAYLAND_KHR,
+                egl::DEFAULT_DISPLAY,
+                &display_attributes,
+            )
+            .unwrap()
+        } else if let Some((x11_display, library)) = x11_display {
+            log::info!("Using X11 platform");
+            let display_attributes = [egl::ATTRIB_NONE];
+            wsi_library = Some(library);
+            egl.get_platform_display(
+                EGL_PLATFORM_X11_KHR,
+                x11_display.as_ptr(),
+                &display_attributes,
+            )
+            .unwrap()
+        } else {
+            log::info!("Using default platform");
+            egl.get_display(egl::DEFAULT_DISPLAY).unwrap()
         };
 
         let inner = Inner::create(egl.clone(), display, wsi_library.as_ref())?;
 
         Ok(Instance {
-            egl,
             inner: Mutex::new(inner),
             wsi_library,
         })
@@ -291,12 +280,13 @@ impl hal::Instance<crate::Backend> for Instance {
 
     fn enumerate_adapters(&self) -> Vec<hal::adapter::Adapter<crate::Backend>> {
         let inner = self.inner.lock().unwrap();
-        self.egl
+        inner
+            .egl
             .make_current(inner.display, None, None, Some(inner.context))
             .unwrap();
         let context = unsafe {
             glow::Context::from_loader_function(|name| {
-                self.egl.get_proc_address(name).unwrap() as *const _
+                inner.egl.get_proc_address(name).unwrap() as *const _
             })
         };
         // Create physical device
@@ -332,7 +322,7 @@ impl hal::Instance<crate::Backend> for Instance {
                 {
                     use std::ops::DerefMut;
                     let display_attributes = [egl::ATTRIB_NONE];
-                    let display = self
+                    let display = inner
                         .egl
                         .get_platform_display(
                             EGL_PLATFORM_WAYLAND_KHR,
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/window/egl.rs
```diff
@@ -38,35 +39,58 @@ fn get_x11_open_display() -> Option<ptr::NonNull<raw::c_void>> {
...
+
+        let display = match (wl_display, x11_display) {
+            (Some(wl_display), _) => {
+                // Use egl::DEFAULT_DISPLAY so we can cleanup the connection
+                // we opened.
+                wayland::display_disconnect(wl_display.as_ptr() as _);
+                log::info!("Using Wayland platform");
+                const EGL_PLATFORM_WAYLAND_KHR: u32 = 0x31D8;
+                let display_attributes = [egl::ATTRIB_NONE];
+                egl::get_platform_display(EGL_PLATFORM_WAYLAND_KHR, ptr as _, &display_attributes)
```
kvark, line 70: I don't understand this part. The docs on `EGL_PLATFORM_WAYLAND_KHR` say that a `wl_display` is needed here for the second parameter. And we have that right here, but instead of using it the code closes the connection and passes in the pointer that was given to `Instance::new`.

We shouldn't be modifying the signature of `Instance::new`.

src/backend/gl/src/window/egl.rs
```diff
@@ -195,20 +219,26 @@ impl hal::Instance<crate::Backend> for Instance {
...
+                let window = wayland::egl_window_create(handle.surface, 1024, 768);
+                window.expect("unsupported window").as_ptr() as _
+                // handle.surface as _
+            }
             other => panic!("Unsupported window: {:?}", other),
         };
         let attributes = [
             egl::RENDER_BUFFER as usize,
             egl::SINGLE_BUFFER as usize,
-            // Always enable sRGB
```
kvark, line 203: if this is only available on EGL 1.5, we need to add a check for 1.5, not just remove them

src/backend/gl/src/window/egl.rs
```diff
@@ -195,20 +219,26 @@ impl hal::Instance<crate::Backend> for Instance {
             Rwh::Xcb(handle) => handle.window as _,
             #[cfg(target_os = "android")]
             Rwh::Android(handle) => handle.a_native_window,
+            #[cfg(not(target_os = "android"))]
+            Rwh::Wayland(handle) => {
+                let window = wayland::egl_window_create(handle.surface, 1024, 768);
```
kvark, line 224: this looks quite fishy to me. We aren't supposed to create a window here. We are supposed to get a window and initialize the GL context on it. `Rwh::Wayland` has the WL surface, and that's actually what we need here, not WL window being created. See:

> To obtain an on-screen rendering surface from a Wayland window, call
    eglCreatePlatformWindowSurface with a <dpy> that belongs to Wayland and
    a <native_window> that points to a `struct wl_egl_surface`.

examples/Cargo.toml
```diff
@@ -37,6 +37,7 @@ hal = { path = "../src/hal", version = "0.6", package = "gfx-hal" }
 auxil = { path = "../src/auxil/auxil", version = "0.5", package = "gfx-auxil" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.6" }
 winit = { version = "0.23", features = ["web-sys"] }
+raw-window-handle = "0.3"
```
kvark, line 40: once the `Instance` API is back, we'll need to remove it

kvark (commented):
Thank you for taking a stab at this!
I think I know why it doesn't work. See comments.

nemosupremo:
So, quickly, the code *works* (as in it actually renders quad.rs), but as you can imagine it hacks the API.

src/backend/gl/src/window/egl.rs
nemosupremo: The problem is the `display` that is passed to `get_platform_display` is created by winit. If I create my own, then quad does not render at all. I understand modifying `Instance::new` here is undesired - I just did it prove it works. What I need is a handle to `wl_display` here that was created by winit.

kvark: Would it be a different `wl_display`, necessarily, than what `wayland::display_connect()` returns?

src/backend/gl/src/window/egl.rs
nemosupremo: I saw that documentation as well and I think it's outdated. This is not creating a window. The documentation here is pretty sparse but you need to create a wayland egl window (https://jan.newmarch.name/Wayland/EGL/)

> An EGL window needs to be created from a Wayland surface, by the Wayland call wl_egl_window_create. This is then turned into an EGL drawing surface by the EGL call eglCreateWindowSurface. Drawing into this surface is generally done by an API such as OpenGL, and the choice of this has been set in the egl_context. This is set for the drawing surfaces by eglMakeCurrent.

src/backend/gl/src/window/egl.rs
nemosupremo: I'm not sure what is happening, but if I pass in `wl_display` that is returned from `wayland::display_connect()` then no window is created on screen. Same for passing `egl::DEFAULT_DISPLAY`.

src/backend/gl/src/window/egl.rs
nemosupremo: Here is a more explanatory article I found (https://ppaalanen.blogspot.com/2012/03/what-does-egl-do-in-wayland-stack.html): 

> The client passes the wl_display object to eglGetDisplay() and receives an EGLDisplay to be used with EGL calls. Then comes the trick that is denoted by the double-arrowed blue line from Wayland client to Mesa EGL in the diagram. The client calls the wayland-egl API (implemented in Mesa) function wl_egl_window_create() to get the native window handle. Normally you would just use the "real" native window object wl_surface (or an X11 Window if you were using X). The native window handle is used to create the EGLSurface EGL handle. **Wayland has this extra step and the wayland-egl API because a wl_surface carries no information of its size. When the EGL library allocates buffers, it needs to know the size, and wayland-egl API is the only way to tell that.**

src/backend/gl/src/window/egl.rs
kvark: Sorry, how come no window is created? The window is created by winit, it has to be there.
Do you mean that no rendering is done on it? And no errors occurring here?

src/backend/gl/src/window/egl.rs
kvark: This link says "This project hasn't been updated since mid-2017."

> An EGL window needs to be created from a Wayland surface, by the Wayland call wl_egl_window_create

My expectation is that Winit calls that, or does something equivalent to this. gfx-rs shouldn't be creating any windows.

kvark: That other link is even older, from 2012. It does sound peculiar indeed, but between the link and the actual spec for the specification we are using, I'd rather trust the spec. Note that the spec expects `wl_egl_surface`, but that function `wl_egl_window_create` returns `wl_egl_window`, so it's a clear mismatch.

nemosupremo: I'll look into this then - I saw the spec and when I tried that I got a segfault. And despite what the spec says, Weston, Wayland reference compositor uses `wl_egl_window_create` as well. I'll see if there are any other projects on github that pass in the surface directly.

https://github.com/wayland-project/weston/blob/master/clients/simple-egl.c#L365

kvark: I'm having trouble finding `wl_egl_surface` at all. Perhaps, the spec is indeed outdated, and this is now called `wl_egl_window`? :/

nemosupremo: I actually missed that - my brain read `wl_egl_surface` and just `wl_surface` - and likewise I've never seen `wl_egl_surface` in [wayland-project/wayland](https://github.com/wayland-project/wayland).

What I can say for sure is most other projects I've used as a reference use `egl_window_create` on a surface. I think the name is misleading however, like the ppaalanen's blogpost article states a `wl_egl_window` is just a surface with size information (you can see [here](https://github.com/wayland-project/wayland/blob/master/egl/wayland-egl.c#L55) that the function to create a  wl_egl_window just wraps a surface with some width/height metadata).

I think, conceptually this code is correct. The bigger problem seems to be that the raw_window_handle that I have, that was created by `winit` has a `wl_surface` and a `wl_display`. The `wl_display` I get from the `raw_window_handle` is created by `winit`, not by `Instance::create`, and when I try to use this surface in `create_platform_surface_window`, wayland doesn't like this (logs from weston 8.0.0):

```
[21:01:54.017] libwayland: invalid object (6), type (zwp_relative_pointer_manager_v1), message attach(?oii)
[21:01:54.025] libwayland: error in client communication (pid 10906)
```

AFAICT, my options are:

1. Reinitialize the glcontext with this new `display`/window handle. This repeats the create function, and also could require an API change (`&mut self`).
2. Pass in the window handle during `create`. This is what I'm doing, via the version parameter, now. Again this breaks the API
3. Somehow figure this out without needing interaction with winit.

kvark: > [21:01:54.017] libwayland: invalid object (6), type (zwp_relative_pointer_manager_v1), message attach(?oii)

This sounds like the object is bad. Is there a way to check if it's a good `wl_display`? I wonder if it's just a bug on `winit` side.

kvark:
Alright, so we have 2 major issues so far:
  1. The `wl_display` from `winit` is not compatible with `wl_display` that we could create in `Instance::new`. We need to figure out why exactly this is the case. If it just happens to be that your platform has the default WL display different from winit window, we should just make it so the surface (created from this window) can't be rendered to by the physical adapter from our Instance, so we'll never need to stitch the two `wl_display` things together.
  2. EGL needs the window size in order to create a surface. The main question to resolve (in order to make progress) here is to find out why X11 path doesn't require that. I suspect one possible answer is that X11 window handle already has that information associated with it, but WL surface doesn't. In this case, I suggest this to be a problem with `RawWindowHandle`, and we should file an issue accordingly and discuss this.

nemosupremo:
wrt (2), I don't think this is an issue. You can call `egl_window_create` with some dummy resolution (I just call it with VGA  640×480), then when you configure the swap chain, you call `egl_window_resize`.

(1) seems to be the tricky one - I suspect wayland's design doesn't support intermixing of resources by two different `wl_displays`, so when we get our `wl_surface` from `winit` and try to use it `libwayland` hits an error. I don't think this has anything to do with the physical display (after all I only have one display), but something inherent with wayland's "secure by design" architecture. I'm going to try and take a deep dive into what the Vulkan instance does.

kvark:
Digging up a bit... here is where `winit` establishes a Wayland display - https://docs.rs/wayland-client/0.28.2/wayland_client/struct.Display.html#method.connect_to_env

nemosupremo:
> For (1) if we fail to use the default wl_display, we could also write the Wayland window integration in a way similar to the Web:

This seems like the most successful path forward, but because we don't have a mutable reference to self, it would require wrapping everything in a mutex to get interior mutability. My plan was to leave eveerything as is (so `enumerate_adapters` still works) but then recreate the context once `create_surface` was called with a display other than our own. The downside of this would be calls to `create_surface`, `enumerate_adapters` and `destroy_surface` would be mutex protected even in X11 - would that cause an undue performance hit?

kvark:
There is no performance concern about mutexing this :)  I was wondering if the logic can all be tight in `src/window/wayland.rs` without affecting the main `window/egl.rs` module, but in the end you see this more clearly since you are writing the code. So I defer to your judgement on how to best roll this out.

fooishbar:
> The wl_display from winit is not compatible with wl_display that we could create in Instance::new. We need to figure out why exactly this is the case. If it just happens to be that your platform has the default WL display different from winit window, we should just make it so the surface (created from this window) can't be rendered to by the physical adapter from our Instance, so we'll never need to stitch the two wl_display things together.

You must use the same `wl_display`; objects such as `wl_surface` are _not_ shareable between different instances of `wl_display`. The ID namespace is completely private to each client connection.

nemosupremo:
Ok, taking into the account the design of wayland with the current API, I've gone the route of interior mutability. What I've done is move the `create` function largely into an `Inner` struct - that way it can easily be dropped/created in `create_surface` once we detect a new display.   Under wayland when are asked to create a surface, we drop the previous context and recreate it. The X11 implementation is largely untouched from a functionality standpoint.

nemosupremo:
The OpenGL renderer seems to work on wgpu using the gl renderer on a pi under wayland

wayland-screenshot-2020-12-22_00-21-01

kvark (commented):
Looking very close to a mergeable state now! Just a couple of things I noticed

src/backend/gl/src/window/egl.rs
```diff
@@ -177,20 +167,132 @@ impl hal::Instance<crate::Backend> for Instance {
...
+            .to_string_lossy();
+        log::info!("Client extensions: {:?}", client_extensions);
+        let client_ext_list = client_extensions.split_whitespace().collect::<Vec<_>>();
+
+        let mut wsi_library = None;
+
+        let wayland_display = if client_ext_list.contains(&"EGL_EXT_platform_wayland") {
+            log::info!("Loading Wayland library to get the current display");
+            if let Ok(library) = libloading::Library::new("libwayland-client.so") {
+                /* We try to connect and disconnect here to simply ensure there
```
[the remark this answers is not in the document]
nemosupremo: No the reason why we open and close the connection is to ensure that a wayland display server is actually running.

If you had wayland installed but you weren’t actually using it, we need to ensure that there is someway to fall back to X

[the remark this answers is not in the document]
kvark: I see, that makes sense.

src/backend/gl/src/window/egl.rs
```diff
@@ -3,7 +3,7 @@
 use crate::{conv, native, GlContainer, PhysicalDevice, Starc};
 use glow::HasContext;
 use hal::{image, window as w};
-use std::{os::raw, ptr};
+use std::{os::raw, ptr, sync::Mutex};
```
kvark, line 6: oh, one last thing: let's use `parking_lot::Mutex` just to be consistent

kvark (approved):

kvark (approved):
Awesome!
bors r+

kvark:
bors retry

PR #3546, opened 2020-12-20
Add typedef for buffer stride

tangmi:
Adds the following:

```rust
/// An stride between elements inside a buffer, in bytes.
pub type Stride = u32;
```

Note that `CommandBuffer::copy_query_pool_results`'s and `Device::get_query_pool_results`'s `stride` parameters have changed from `u64` to `u32`. I'm unsure of the implications of this, but I don't imagine nobody has a buffer with a stride > `u32::max_value()` (can backends even support buffers that large?). Please let me know if I should change this back!

PR checklist:
- [x] tested examples with the following backends: Vulkan, DX11, DX12

### DIFF
--- a/examples/compute/main.rs
+++ b/examples/compute/main.rs
@@ -42,7 +42,7 @@ fn main() {
         .skip(1)
         .map(|s| u32::from_str(&s).expect("You must pass a list of positive integers!"))
         .collect();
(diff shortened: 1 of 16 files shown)


kvark (approved):
Holy cow, that's a lot of types!
bors r+

PR #3548, opened 2020-12-21
[dx11] Fixes regressions from 0.6 to Master

cwfitzgerald:
This PR plus the gpu-alloc fixes all the DX11 specific issues in https://github.com/gfx-rs/wgpu/issues/1059.

These are all refcount issues, one was a double free of images, the other was some resources not being freed on the swapchain.


kvark (approved):
bors r+

PR #3549, opened 2020-12-22
[vk] Remove lazy_static and (most) hard-coded strings for layer and extension names

tangmi:
The `ash` generated methods `*::name()` return a `&'static CStr`, which means we probably don't need to use `lazy_static` to keep track of extension names.


kvark (approved):
Amazing cleanup, thank you!
bors r+

PR #3555, opened 2020-12-27
Check for GL program linking before using it

kvark:
Fixes an issue where there is a panic because the program failed to link, but no actual error reported.


grovesNL (approved):
bors r+

PR #3556, opened 2020-12-28
Fix WebGPU build to use buffer::Stride

tangmi:
Fixing the parts of the WebGPU build I broke in #3546, since it's not included in the `make` build.


kvark (approved):
We need to check it on CI... or disable completely
bors r+

PR #3557, opened 2020-12-28
Add WebGPU backend build to CI

tangmi:
Add an optional CI task to build the WebGPU backend on the `wasm32-unknown-unknown` target.

Note: This is just a test to see if we can at least see the status of the WebGPU build in the Ci pipeline. It may not be worth merging until the WebGPU backend is more mature, since it may just add noise for now.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,6 +39,17 @@ jobs:
       - run: rustup target add wasm32-unknown-unknown
       - run: cargo build --manifest-path examples/Cargo.toml --features gl --target wasm32-unknown-unknown --bin quad
 
+  webgpu_backend_build:
+    name: WebGPU
+    runs-on: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


kvark:
Looks trivially fixable so far?

tangmi:
I'm happy to clean up this build if we'd like to add a webgpu build to the CI requirements. I'm not sure how mature that backend is, but I suppose ensuring it keeps building can't hurt...

kvark:
It's not mature at all. In fact, it's more of a burden right now... We need to make a decision - either support it on CI, or just carry the code as dead weight (for now!) without any builds/checks. I think the latter is easier, and it's ok until there starts to be any logic in it.

tangmi:
I'm going to close this for now--when the WebGPU backend becomes more mature, this PR should be trivially reproducible.

For now, I'll remove the extra noise this PR causes.

PR #3559, opened 2020-12-29
Distinguish STORAGE from STORAGE_READ_WRITE image feature

Wumpf:
Implemented for Vulkan/DX12/DX11.
In Vulkan, storage image implies simultaneous read/write access.
Other APIs in contrast may allow STORAGE writeonly while disallowing reading.

Related to https://github.com/gfx-rs/wgpu/pull/1109 - native extension for for read+write storage textures in wgpu native; following comments on the PR I'm redoing it and would need to add this to the hal api.

PR checklist:
    - Vulkan/DX12/DX11
    - did a dump of storage/storage-atomic/storage-rw flags each and eyeballed it as a sanity check

### DIFF
--- a/src/hal/src/format.rs
+++ b/src/hal/src/format.rs
@@ -148,26 +148,27 @@ bitflags!(
     pub struct ImageFeature: u32 {
         /// Image view can be sampled.
         const SAMPLED = 0x1;
+        /// Image can be sampled with a (mipmap) linear sampler or as blit source with linear sampling.
+        const SAMPLED_LINEAR = 0x10000 | ImageFeature::SAMPLED.bits() | ImageFeature::BLIT_SRC.bits();
+
         /// Image view can be used as storage image with exclusive read & write access.
-        const STORAGE = 0x2;
-        /// Image view can be used as storage image with atomics.
-        const STORAGE_ATOMIC = 0x4;
+        const STORAGE = 0x10;
         /// Image view can be used as storage image with simultaneous read/write access.
-        const STORAGE_READ_WRITE = 0x8;
+        const STORAGE_READ_WRITE = 0x20 | ImageFeature::STORAGE.bits();
+        /// Image view can be used as storage image with atomics.
+        const STORAGE_ATOMIC = 0x40 | ImageFeature::STORAGE_READ_WRITE.bits();
+
         /// Image view can be used as color and input attachment.
-        const COLOR_ATTACHMENT = 0x80;
+        const COLOR_ATTACHMENT = 0x100;
         /// Image view can be used as color (with blending) and input attachment.
-        const COLOR_ATTACHMENT_BLEND = 0x100;
+        const COLOR_ATTACHMENT_BLEND = 0x200 | ImageFeature::COLOR_ATTACHMENT.bits();
         /// Image view can be used as depth-stencil and input attachment.
-        const DEPTH_STENCIL_ATTACHMENT = 0x200;
+        const DEPTH_STENCIL_ATTACHMENT = 0x400;
+
         /// Image can be used as source for blit commands.
-        const BLIT_SRC = 0x400;
+        const BLIT_SRC = 0x1000;
         /// Image can be used as destination for blit commands.
-        const BLIT_DST = 0x800;
-        /// Image can be sampled with a (mipmap) linear sampler or as blit source
-        /// with linear sampling.
-        /// Requires `SAMPLED` or `BLIT_SRC` flag.
-        const SAMPLED_LINEAR = 0x1000;
+        const BLIT_DST = 0x2000;
     }
 );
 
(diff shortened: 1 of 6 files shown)


src/hal/src/format.rs
```diff
@@ -148,10 +148,12 @@ bitflags!(
...
         /// Image view can be sampled.
         const SAMPLED = 0x1;
-        /// Image view can be used as storage image.
+        /// Image view can be used as storage image with exclusive read & write access.
         const STORAGE = 0x2;
-        /// Image view can be used as storage image (with atomics).
+        /// Image view can be used as storage image with atomics.
         const STORAGE_ATOMIC = 0x4;
+        /// Image view can be used as storage image with simultaneous read/write access.
+        const STORAGE_READ_WRITE = 0x8;
```
kvark, line 156: what about Metal? can we have some logic to check when this flag is available?

kvark, line 156: We need to ensure that Vulkan flags don't intersect our flags, since as you see the Vulkan backend just casts the value raw... So we should either figure out a scheme where gfx-specific values have a weird mask of 0x1000XXX (or something like that), or totally detach our `ImageFeatures` from Vulkan's, and do the more elaborate translation in the Vulkan backend.

kvark (commented):
Thank you for doing this, much needed for good wgpu support!

src/hal/src/format.rs
Wumpf: Yes, absolutely. We need to check OSX_ReadWriteTextureTier / MTLReadWriteTextureTier and then grind through [this table](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf).
I haven't done anything yet though for OpenGL and Metal, since the `format_properties` properties aren't implemented there to begin with, so there needs to be a follow up to this PR eventually to fill them out.

(got this all via one of these amazing writeups the WebGPU specification effort produces https://github.com/gpuweb/gpuweb/issues/513)

Wumpf: Went for detaching it in 8db3137
Took the freedom to express relationships between the flags in code, makes things a little bit more sane imho :)

src/hal/src/format.rs
```diff
@@ -148,24 +148,27 @@ bitflags!(
...
-        const STORAGE_ATOMIC = 0x4;
+        /// Image can be sampled with a (mipmap) linear sampler or as blit source with linear sampling.
+        const SAMPLED_LINEAR = 0x10000 | ImageFeature::SAMPLED.bits() | ImageFeature::BLIT_SRC.bits();
+
+        /// Image view can be used as storage image with exclusive read & write access.
+        const STORAGE = 0x10;
+        /// Image view can be used as storage image with simultaneous read/write access.
+        const STORAGE_READ_WRITE = 0x20 | ImageFeature::STORAGE.bits();
+        /// Image view can be used as storage image with atomics.
+        const STORAGE_ATOMIC = 0x40 | ImageFeature::STORAGE_READ_WRITE.bits();
```
kvark, line 159: We should probably not do that. More specifically, having the constant flags that are neither combinations of existing constants, or single-bit flags. It becomes a bit confusing, and there isn't a strong reason to go this route.

Interestingly, this looks more like an enum to me:
```rust
enum StorageSupport {
  Basic,
  ReadWrite,
  Atomic,
}
```
Of course I'm not suggesting you to make one, just thought about it as an interesting direction in the future.

src/hal/src/format.rs
kvark: That's not true. `format_properties` are implemented. Maybe you overlooked it?

src/hal/src/format.rs
Wumpf: I'll remove the combinations then 🤷 
Agreed on the enum, that would be cleaner. But yeah let's leave that for another time as this would ripple out much more then.

src/hal/src/format.rs
Wumpf: ouch, ofc you're right. Apologies, I must have looked at the gl version twice.

Turns out @bsecura added MTLReadWriteTextureTier to metal-rs quite recently, perfect timing \o/
Only means that we need to update to a newer version now. So I went and put in the git repo in cargo.toml for testing. Had a bit of ripple effects of things that need to be updated - want me to put that commit into a separate PR?

Ofc we can't land this like this. I saw you did some stuff in metal-rs yourself recently and already bumped the version number to 0.20.1, mind publishing to crates.io timely so this here can move on? :)


Sidenote: I found that there's also such a feature in the gpufamily->features pdf which metal-rs also implements, but in the pdf this comes with footnotes attached one should please check on the texture tier, a bit odd situation, seems the TextureTier enum is simply newer.

kvark: #3561 already switches to metal-rs master. I'll get it in now.

Wumpf: ah didn't see that. I'll rebase then!

Wumpf: done.

kvark (approved):
Looks great!
bors r+

PR #3327, opened 2020-08-05
Remove the old swapchain model

kvark:
Closes #3180 
PR checklist:
- [x] `make` succeeds (on *nix)
- [x] `make reftests` succeeds
- [x] tested examples with the following backends:
- [ ] `rustfmt` run on changed code

PR #3560, opened 2020-12-30
WIP remove framebuffer object

### LINKED ISSUE #3552: Remove Framebuffers from the API
This was a topic that was brought up in the matrix chat. I was asked to create an issue.

Essentially, now that the swapchain was removed from the API, the control over swap images (and thus framebuffers which contain swap images) was taken away from the user. We must reacquire a swap image every frame, and we must recreate a framebuffer that uses it every frame. To compensate, framebuffers are cached internally by the vulkan backend, but only if they contain a swap image.

It feels unintuitive and inconsistent that this caching behavior exists for swapimage framebuffers and only for swapimage framebuffers. It'd be nice if it extended to all framebuffers. But then, what is the point of exposing framebuffers in the first place? I imagine the reason why framebuffers are exposed in Vulkan is to take that responsibility away from the device/driver and give it to the user, but currently the user of gfx-rs does not have full control over framebuffers (because of the caching behavior and the removal of the swapchain). Perhaps the next logical step is just to remove framebuffers from the API entirely and extend the internal caching behavior to support all framebuffers.

kvark:
Closes #3552

### History and motivation

In the new swapchain model, which became the one and only model as of #3327, users are not able to pre-create framebuffers for every swapchain image in advance. We settled on the requirement for the user to create a framebuffer for every frame they render, and then remove it after presenting. This didn't fit into the low-level API too well, and Vulkan backend had to implement the smarts of tracking the use of swapchain-related framebuffers. See more detail by @thorjelly in #3552.

### Status
This PR removes framebuffer as a concept, leaving it to be an implementation detail of the backends.
Currently it's done for HAL, Vulkan, and Metal.
Biggest concerns:
  - OpenGL backend complexity/consistency

### Analysis

Effect on the API: great, simpler API. Especially the new `RenderAttachmentInfo` struct. We already had a slight divergence with Vulkan on the clear values in `begin_render_pass`, and this PR makes it look very solid and concise.

Effect on the user: great, less things to manage, especially the fact that we don't have "special" framebuffers versus the regular ones any more.

Effect on performance:
  - Vulkan: minimal.
    - It's now easier to internally reuse those framebuffers.
    - We have anecdotal evidence that of all the Vulkan implementors only one mobile vendor can actually do anything useful with the framebuffer object, and others don't use it.
  - OpenGL: negative, we can no longer create FBOs in advance
  - other: no effect

Effect on the code complexity:
  - Vulkan: positive, there is no special-casing any more, no managing of the "owner" of framebuffers.
  - Metal: positive, also less special-casing.
  - OpenGL: negative. Unlike Vulkan, we can't just create an FBO on the spot while recording, since it's done off-line. So we'll have to have internal commands elevated to higher level, i.e. `SetRenderTargets` instead of `BindFramebuffer`.

Finally, this helps streamlining the path from wgpu to the hardware, as WebGPU doesn't have framebuffers either.

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -342,18 +342,6 @@ impl<B: Backend> RendererState<B> {
             }
         };
 
-        let framebuffer = unsafe {
-            self.device
-                .borrow()
-                .device
-                .create_framebuffer(
-                    self.render_pass.render_pass.as_ref().unwrap(),
-                    iter::once(std::borrow::Borrow::borrow(&surface_image)),
-                    self.swapchain.extent,
-                )
-                .unwrap()
-        };
-
         let frame_idx = (self.swapchain.frame_index % self.swapchain.frame_queue_size) as usize;
         self.swapchain.frame_index += 1;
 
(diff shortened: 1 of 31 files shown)


msiglreith (approved):

kvark:
Closing in favor of #3571

PR #3566, opened 2021-01-07
Fixed incorrect depth format conversions in GL backend

ElArtista:
Changed icorrect format parameters from glow::DEPTH to glow::DEPTH_FORMAT


kvark (approved):
thank you!
bors r+

PR #3571, opened 2021-01-08
Image-less framebuffer API

### LINKED ISSUE #3568: Switch API to image-less framebuffers
I've been re-evaluating our swapchain API, trying to make it match Vulkan better. After toying with multiple ideas, the most promising one appears to be - to switch the API of our framebuffers according to `VK_KHR_imageless_framebuffer`. It seems to be widely supported on fresh Vulkan drivers, so we'd be able to map it exactly 1:1 with Vulkan.

Pros: it works great for our new swapchain model, since we no longer need to cache anything internally, when the actual Vulkan adapter supports this. Users would enjoy it quite a bit, since there would be no requirement to constantly create and destroy framebuffers every frame.

Wierds: Vulkan backend to have 2 code paths: caching and non-caching. This is only marginally worse than what we have today. I think the lifetimes are cleaner though: the cache is owned by the API framebuffer object and dies with it.

Cons: OpenGL backend would still need to be changed in non-trivial way, similar to #3552

kvark:
Fixes #3568
It will also solve the annoying validation error about Vulkan framebuffers being destroyed - #3015. Even though we still rely on the "acquire" to actually make the image available, the framebuffers in Vulkan are going to be more solid and persistent. No creation/removal each frame any more! 🎉 🚀

### DIFF
--- a/src/backend/gl/src/window/web.rs
+++ b/src/backend/gl/src/window/web.rs
@@ -1,11 +1,10 @@
 use crate::{
     conv, device::Device, native, Backend as B, GlContainer, PhysicalDevice, QueueFamily, Starc,
 };
-use arrayvec::ArrayVec;
 use glow::HasContext;
 use hal::{adapter::Adapter, format as f, image, window};
 use parking_lot::Mutex;
-use std::{iter, sync::Arc};
+use std::sync::Arc;
 use wasm_bindgen::JsCast;
 
 #[derive(Clone, Debug)]
@@ -13,7 +12,7 @@ pub struct Swapchain {
     pub(crate) extent: window::Extent2D,
     pub(crate) channel: f::ChannelType,
     pub(crate) raw_format: native::TextureFormat,
-    pub(crate) frame_buffers: ArrayVec<[native::RawFrameBuffer; 3]>,
+    pub(crate) framebuffer: native::RawFramebuffer,
 }
 
 #[derive(Debug)]
@@ -122,9 +121,8 @@ impl Surface {
     ) -> Result<Option<window::Suboptimal>, window::PresentError> {
         let swapchain = self.swapchain.as_ref().unwrap();
 
-        let frame_buffer = swapchain.frame_buffers.first().unwrap();
         gl.bind_framebuffer(glow::DRAW_FRAMEBUFFER, None);
-        gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(*frame_buffer));
+        gl.bind_framebuffer(glow::READ_FRAMEBUFFER, Some(swapchain.framebuffer));
         gl.blit_framebuffer(
             0,
             0,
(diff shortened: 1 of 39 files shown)


kvark:
I think this is ready for review/landing!
Interestingly, the refactor on GL ended up without as much boilerplate as I feared. I figured there is no need to maintain any caches of framebuffers, since we can mutate existing ones. That nails down pretty much the only concern I had with the change, and I'm very happy to see what it came to be, after so much experimentation and wasted code!

src/backend/gl/src/window/web.rs
```diff
@@ -13,7 +13,7 @@ pub struct Swapchain {
     pub(crate) extent: window::Extent2D,
     pub(crate) channel: f::ChannelType,
     pub(crate) raw_format: native::TextureFormat,
-    pub(crate) frame_buffers: ArrayVec<[native::RawFrameBuffer; 3]>,
+    pub(crate) frame_buffers: ArrayVec<[native::RawFramebuffer; 3]>,
```
manugildev, line 16: -  `frame_buffers` -> `framebuffers`. Just to keep consistency with the rest.
- Do we need both ArrayVec and SmallVec in the codebase? Could we remove one of them?
- Why `3`? We are only using the first one.

src/backend/vulkan/src/native.rs
```diff
@@ -64,13 +63,19 @@ pub struct Sampler(pub vk::Sampler);
...
     pub raw: vk::RenderPass,
-    pub clear_attachments_mask: u64,
+    pub attachment_count: usize,
 }
 
-#[derive(Debug, Hash)]
-pub struct Framebuffer {
-    pub(crate) raw: vk::Framebuffer,
-    pub(crate) owned: bool,
+pub type FramebufferKey = SmallVec<[vk::ImageView; 5]>;
```
manugildev, line 69: There might be some value in the future of not hardcoding the `5`, maybe by getting `surface_capabilities.max_image_count/min_image_count`.
Hardcoded also in command.rs -> line: 252:
 `let mut raw_clear_values = SmallVec::<[vk::ClearValue; 5]>::new();`

src/backend/gl/src/window/web.rs
kvark: > Do we need both ArrayVec and SmallVec in the codebase? Could we remove one of them?

That's something I direly wanted to do. In particular, removing `SmallVec`. But it turns out that it serves a distinct use case that we still need in a few places.

src/backend/vulkan/src/native.rs
kvark: Added a constant for this. We may come back to it and set up a real maximum bound, so that we can use `ArrayVec` instead of `SmallVec`.

kvark:
@manugildev thank you for taking a look!
bors r=manugildev

kvark:
bors r=manugildev

PR #3573, opened 2021-01-09
Remove lazy_static dependency from Vulkan

### LINKED ISSUE #3572: Remove lazy_static dependency from Vulkan
Instead of storing it statically, we should just make it a part of our `struct Instance` in the Vulkan backend.

manugildev:
Fixes #3572

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -2,8 +2,6 @@
 
 #[macro_use]
 extern crate log;
-#[macro_use]
-extern crate lazy_static;
 
 #[cfg(target_os = "macos")]
 #[macro_use]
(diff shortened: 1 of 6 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -106,6 +85,12 @@ pub struct Instance {
 
     /// Supported extensions of this instance.
     pub extensions: Vec<&'static CStr>,
+
+    #[cfg(not(feature = "use-rtld-next"))]
+    pub entry: Entry,
```
kvark, line 90: I wonder if this could be something like `dyn EntryTrait` that works for both? Doesn't matter much, just an option if we can remove this cfg branch

src/backend/vulkan/src/lib.rs
```diff
@@ -303,8 +288,16 @@ unsafe extern "system" fn debug_report_callback(
 
 impl hal::Instance<Backend> for Instance {
     fn create(name: &str, version: u32) -> Result<Self, hal::UnsupportedBackend> {
-        // TODO: return errors instead of panic
-        let entry = VK_ENTRY.as_ref().map_err(|e| {
+        #[cfg(not(feature = "use-rtld-next"))]
+        let vk_entry = Entry::new();
```
kvark, line 292: could we just do a `match Entry::new()` here instead of following with `map_err` later?

kvark (approved):
Looks beautiful, thank you!
Just a few nits, happy to land this when you are done.

kvark (approved):
Awesome!
bors r+

PR #3576, opened 2021-01-11
Add image features for transfers and min/max sampling

### LINKED ISSUE #3575: Add transfer src/dst format features
These have been added in one of the VK_Maintenance extensions, and are a part of Vulkan 1.1.
See `VK_FORMAT_FEATURE_TRANSFER_SRC_BIT` and `VK_FORMAT_FEATURE_TRANSFER_DST_BIT` in https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkFormatFeatureFlagBits.html
Requirement for #3146

kvark:
Fixes #3575 
Having distinct `TRANSFER_SRC` and `TRANSFER_DST` flags follows Vulkan's maintenance-1 extension and allows for better separation between "blit" and "transfer" capabilities.

PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -343,87 +343,91 @@ fn get_format_properties(
                 mem::size_of::<d3d11::D3D11_FEATURE_DATA_FORMAT_SUPPORT>() as UINT,
             )
         };
+        if hr != winerror::S_OK {
+            warn!("Format {:?} can't check the features-1: 0x{:x}", format, hr);
+            continue;
+        }
 
-        if hr == winerror::S_OK {
-            let can_buffer = 0 != support.OutFormatSupport & d3d11::D3D11_FORMAT_SUPPORT_BUFFER;
-            let can_image = 0
-                != support.OutFormatSupport
-                    & (d3d11::D3D11_FORMAT_SUPPORT_TEXTURE1D
-                        | d3d11::D3D11_FORMAT_SUPPORT_TEXTURE2D
-                        | d3d11::D3D11_FORMAT_SUPPORT_TEXTURE3D
-                        | d3d11::D3D11_FORMAT_SUPPORT_TEXTURECUBE);
-            let can_linear = can_image && !format.surface_desc().is_compressed();
-            if can_image {
-                props.optimal_tiling |=
-                    format::ImageFeature::SAMPLED | format::ImageFeature::BLIT_SRC;
-            }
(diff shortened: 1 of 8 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -343,87 +343,91 @@ fn get_format_properties(
...
+        let can_buffer = 0 != support.OutFormatSupport & d3d11::D3D11_FORMAT_SUPPORT_BUFFER;
+        let can_image = 0
+            != support.OutFormatSupport
+                & (d3d11::D3D11_FORMAT_SUPPORT_TEXTURE1D
+                    | d3d11::D3D11_FORMAT_SUPPORT_TEXTURE2D
+                    | d3d11::D3D11_FORMAT_SUPPORT_TEXTURE3D
+                    | d3d11::D3D11_FORMAT_SUPPORT_TEXTURECUBE);
+        let can_linear = can_image && !format.surface_desc().is_compressed();
+        if can_image {
+            props.optimal_tiling |= format::ImageFeature::TRANSFER_SRC
```
Wumpf, line 360: Could easily also query Min/max filter capabilities here. Same goes for Dx12. Or is there something else preventing support?

https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_feature_data_d3d11_options1

Wumpf (commented):
didn't go in-depth on the metal diff - github diff tool isn't happy with the refactor - but looked good to me on a quick glimpse. Also didn't find any way to support MinMax samplers there either.

Nice to have TRANSFER as a feature exposed, makes it more straight forward not only in wgpu land! :)

src/backend/dx11/src/lib.rs
kvark: I think that flag specifies if min/max is supported in general, but here we need to know if it's supported for a particular format. It's doable, but it's more work, and we can follow-up with this considering that the reduction sampling parameters are not exposed in this PR anyway.

kvark:
Thank you!
bors r=Wumpf

src/backend/dx11/src/lib.rs
Wumpf: I actually got the impression that if that flag is enabled it is supported wherever linear filter is supported - there's nothing indicating that this has more fine-grained granularity per format

kvark: Yeah, this part is weird. I wasn't able to find out which formats can support it in D3D, see https://github.com/gpuweb/gpuweb/pull/1283
In Vulkan, it's exposed on a per-format basis, plus a feature flag for a set of common formats. But it's definitely not on all of them.

PR #3578, opened 2021-01-11
Expose dynamic state capabilities in the backends

### LINKED ISSUE #3441: Non-dynamic pipeline states in DX12
In DX12, the following pipeline states can only be static:
  - depth bias
  - stencil read/write masks

We need to encode this limitation in the flags we expose. It doesn't seem to be easy to work around.

kvark:
Fixes  #3441
Reverses the semantic of "hints", also renaming them to "performance_caveats", and placing inside a new struct called `Capabilities`. Unlike features, this struct represent properties of the adapter that do not need to be enabled.
PR checklist:

### DIFF
--- a/src/hal/src/adapter.rs
+++ b/src/hal/src/adapter.rs
@@ -11,7 +11,7 @@
 use crate::{
     device, format, image, memory,
     queue::{QueueGroup, QueuePriority},
-    Backend, Features, Hints, Limits,
+    Backend, Capabilities, Features, Limits,
 };
 
 use std::{any::Any, fmt};
@@ -116,11 +116,12 @@ pub trait PhysicalDevice<B: Backend>: fmt::Debug + Any + Send + Sync {
     fn memory_properties(&self) -> MemoryProperties;
 
     /// Returns the features of this `PhysicalDevice`. This usually depends on the graphics API being
-    /// used.
+    /// used, as well as the actual platform underneath.
     fn features(&self) -> Features;
 
-    /// Returns the performance hints of this `PhysicalDevice`.
-    fn hints(&self) -> Hints;
+    /// Returns the capabilities of this `PhysicalDevice`. Similarly to `Features`, they
+    // depend on the platform, but unlike features, these are immutable and can't be switched on.
+    fn capabilities(&self) -> Capabilities;
 
     /// Returns the resource limits of this `PhysicalDevice`.
     fn limits(&self) -> Limits;
--- a/src/hal/src/lib.rs
+++ b/src/hal/src/lib.rs
@@ -265,12 +265,52 @@ bitflags! {
 }
 
 bitflags! {
-    /// Features that the device supports natively, but is able to emulate.
+    /// Features that the device doesn't support natively, but is able to emulate
+    /// at some performance cost.
+    #[derive(Default)]
     #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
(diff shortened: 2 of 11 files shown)


src/hal/src/adapter.rs
```diff
@@ -116,11 +116,12 @@ pub trait PhysicalDevice<B: Backend>: fmt::Debug + Any + Send + Sync {
...
 
     /// Returns the features of this `PhysicalDevice`. This usually depends on the graphics API being
-    /// used.
+    /// used, as well as the actual platform underneath.
     fn features(&self) -> Features;
 
-    /// Returns the performance hints of this `PhysicalDevice`.
-    fn hints(&self) -> Hints;
+    /// Returns the capabilities of this `PhysicalDevice`. Similarly to `Features`, they
+    // depend on the platform, but unlike features, these are immutable and can't be swithed on.
```
grovesNL, line 123: Typo in switched

src/hal/src/adapter.rs
```diff
@@ -116,11 +116,12 @@ pub trait PhysicalDevice<B: Backend>: fmt::Debug + Any + Send + Sync {
     fn memory_properties(&self) -> MemoryProperties;
 
     /// Returns the features of this `PhysicalDevice`. This usually depends on the graphics API being
-    /// used.
+    /// used, as well as the actual platform underneath.
     fn features(&self) -> Features;
 
-    /// Returns the performance hints of this `PhysicalDevice`.
-    fn hints(&self) -> Hints;
+    /// Returns the capabilities of this `PhysicalDevice`. Similarly to `Features`, they
```
grovesNL, line 122: I wonder if it would make sense to collapse `Features` and `Capabilities` together so there's only one place to look. e.g. a user could still request these dynamic states if we exposed it as a feature instead, and we'd just fail in that case

grovesNL (approved):
Looks good to me!

src/hal/src/lib.rs
```diff
@@ -265,14 +265,52 @@ bitflags! {
...
+        /// Supports `StencilTest::read_masks == State::Dynamic(_)`
+        const STENCIL_READ_MASK = 0x0100;
+        /// Supports `StencilTest::write_masks == State::Dynamic(_)`
+        const STENCIL_WRITE_MASK = 0x0200;
+        /// Supports `StencilTest::reference_values == State::Dynamic(_)`
+        const STENCIL_REFERENCE = 0x0400;
+    }
+}
+
+/// Capabilities of physical devices that are exposed but do not nee
```
grovesNL, line 304: Missing comment

src/hal/src/adapter.rs
kvark: The idea I wanted to realize here is a structure with stuff that does *not* need to be enabled. Currently, it's either in Features, which have to be enabled explicity, or Limits, which are 99% just numbers. So putting the dynamic state in either was not a good option. But we had those `Hints`, so I replaced them by a structure that is more scalable. Technically, we could move `Limits` in there as well. Do you think it's worth doing? I don't have a strong feeling here.

kvark:
Thank you for the review!
bors r=grovesNL

src/hal/src/adapter.rs
grovesNL: Right, I don’t have a strong preference either. I was just thinking it’s a little unfortunate that the users have to look in multiple places to check which features/capabilities they can use, but maybe that’s fine

PR #3579, opened 2021-01-11
[Metal] Remove lazy_static

### LINKED ISSUE #3574: Remove lazy_static dependency from Metal
Following #3572 we should remove it on Metal as well. It's only used in one place:
```rust
lazy_static! {
    static ref GFX_MANAGED_METAL_LAYER_DELEGATE_CLASS: &'static Class = unsafe {
        let mut decl = ClassDecl::new("GfxManagedMetalLayerDelegate", class!(NSObject)).unwrap();
        decl.add_method(
            sel!(layer:shouldInheritContentsScale:fromWindow:),
            layer_should_inherit_contents_scale_from_window
                as extern "C" fn(&Object, Sel, *mut Object, CGFloat, *mut Object) -> BOOL,
        );
        decl.register()
    };
}
```
Since it's a part of the `Instance`, we should be able to avoid the static, throwing the dependency out.

Gordon-F:
Fixes #3574
PR checklist:

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -34,12 +34,10 @@ metal = { git = "https://github.com/gfx-rs/metal-rs", rev="ba08f5f98c70ab941020b
 foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
-#TODO: remove this, only used for `Id`
 cocoa-foundation = "0.1"
 spirv_cross = { version = "0.22", features = ["msl"] }
 parking_lot = "0.11"
 storage-map = "0.3"
-lazy_static = "1"
 raw-window-handle = "0.3"
 
 [dependencies.naga]
--- a/src/backend/metal/src/device.rs
+++ b/src/backend/metal/src/device.rs
@@ -2132,7 +2132,7 @@ impl hal::device::Device<Backend> for Device {
                 data: msl::SamplerData,
                 binding: pso::DescriptorBinding,
                 array_index: pso::DescriptorArrayIndex,
-            };
+            }
             let mut immutable_sampler_iter = immutable_samplers.into_iter();
             let mut tmp_samplers = Vec::new();
             let mut desc_layouts = Vec::new();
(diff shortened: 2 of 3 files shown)


src/backend/metal/Cargo.toml
```diff
@@ -34,12 +34,9 @@ metal = { git = "https://github.com/gfx-rs/metal-rs", rev="ba08f5f98c70ab941020b
 foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
-#TODO: remove this, only used for `Id`
-cocoa-foundation = "0.1"
```
kvark, line 38: Now that I'm looking at this again, we should probably keep using it. My comment got you fixing this, sorry!
The justification is that `metal-rs` relies on cocoa-foundation, and as long as it's there, we gain nothing from trying to re-implement the types from it in `gfx-backend-metal`.

kvark (commented):
Beautiful!
I'd be happy to land this, but want to check if the comment make sense, and if we should roll back the cocoa-foundation changes (and delete my erroneous comment).

src/backend/metal/Cargo.toml
Gordon-F: Done :smile:

kvark (approved):
Thank you!
bors r+

kvark:
Strangely, I'm seeing this now when running wgpu-rs (patched to use latest gfx):
> objc[3495]: Class GfxManagedMetalLayerDelegate is implemented in both ?? (0x7fa83f40ee00) and ?? (0x7fa83f504520). One of the two will be used. Which one is undefined.
objc[3495]: Class GfxManagedMetalLayerDelegate is implemented in both ?? (0x7fa83f40ee00) and ?? (0x7fa83f5046c0). One of the two will be used. Which one is undefined.

kvark:
Ah, snap, that may be a real issue here. If anybody creates multiple instances, we are in trouble :/
Perhaps, the instance code could check if there is an existing declaration somehow?

kvark:
nvm, fixing this in #3585 :)

PR #3586, opened 2021-01-13
Remove all the Borrows

### LINKED ISSUE #3569: Don't use Borrow with lifetimes in the API
We have a lot of patterns like this:
```rust
where
        T: IntoIterator,
        T::Item: Borrow<hal::memory::Barrier<'a, Backend>>,
```

Where the borrowed type is defined as:
```rust
pub enum Barrier<'a, B: Backend> {
    Buffer {
        target: &'a B::Buffer,
        ...
    },
    Image {
        target: &'a B::Image,
        ...
    },
    ...
}
```

The main motivation for `Borrow` was to accommodate for cases where the user has an iterator containing our data, but Rust doesn't allow us to go from `Iterator<Item=T>` to `Iterator<Item=&T>`, unfortunately.

I think there is a bit of contradiction here, since there is `Borrow`, but still we require `&'a B::Something` inside. So this would only work for cases where the user has an iterator over something that contains `Barrier` to start with. This seems like a weird use case. It's more likely that the user stores `B::Buffer` and `B::Image` in some structures that are iterated over, and the user constructs `Barrier` on the fly. In this case the `Borrow` bound doesn't help at all.

Suggestion: remove `Borrow` from all the API where items depend on `'a`. Instead, we can work with just `IntoIterator<Item = Something>`.

kvark:
Fixes #3569 
PR checklist:

Tested with:
- https://github.com/gfx-rs/wgpu/pull/1156
- https://github.com/gfx-rs/portability/pull/240


grovesNL (approved):

kvark:
bors r=grovesNL

PR #3592, opened 2021-01-15
[GL] Android emulator support

Gordon-F:
PR checklist:
- [x] tested examples with the following backends: OpenGL ES

Android emulator doesn't support EGL_KHR_surfaceless_context.

### DIFF
--- a/src/backend/gl/src/window/egl.rs
+++ b/src/backend/gl/src/window/egl.rs
@@ -30,6 +30,7 @@ pub struct Inner {
     display: egl::Display,
     config: egl::Config,
     context: egl::Context,
+    pbuffer: egl::Surface,
     wl_display: Option<*mut raw::c_void>,
 }
 
@@ -113,13 +114,6 @@ impl Inner {
             log::error!("EGL supported version is only {:?}", version);
             return Err(hal::UnsupportedBackend);
         }
-        let display_ext_list = display_extensions.split_whitespace().collect::<Vec<_>>();
-        let required_display_extensions = ["EGL_KHR_create_context", "EGL_KHR_surfaceless_context"];
-        for required_ext in required_display_extensions.iter() {
-            if !display_ext_list.contains(required_ext) {
-                log::warn!("{} is not present", required_ext);
-            }
-        }
 
         if log::max_level() >= log::LevelFilter::Trace {
             log::trace!("Configurations:");
@@ -195,6 +189,15 @@ impl Inner {
             }
         };
 
+        let pbuffer = {
+            let attributes = [egl::WIDTH, 1, egl::HEIGHT, 1, egl::NONE];
+            egl.create_pbuffer_surface(display, config, &attributes)
+                .map_err(|e| {
+                    log::warn!("Error in create_pbuffer_surface: {:?}", e);
+                    hal::UnsupportedBackend
+                })?
+        };
+
         Ok(Self {
             egl,
             display,
@@ -202,6 +205,7 @@ impl Inner {
             supports_native_window,
             config,
             context,
+            pbuffer,
             wl_display: None,
         })
     }
@@ -287,8 +291,14 @@ impl hal::Instance<crate::Backend> for Instance {
         let inner = self.inner.lock();
         inner
             .egl
-            .make_current(inner.display, None, None, Some(inner.context))
+            .make_current(
+                inner.display,
+                Some(inner.pbuffer),
+                Some(inner.pbuffer),
+                Some(inner.context),
+            )
             .unwrap();
+
         let context = unsafe {
             glow::Context::from_loader_function(|name| {
                 inner
@@ -419,6 +429,7 @@ impl hal::Instance<crate::Backend> for Instance {
             display: inner.display,
             context: inner.context,
             presentable: inner.supports_native_window,
+            pbuffer: inner.pbuffer,
             wl_window,
             swapchain: None,
         })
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/window/egl.rs
```diff
@@ -596,9 +613,6 @@ impl Surface {
         gl.bind_framebuffer(glow::READ_FRAMEBUFFER, None);
 
         self.egl.swap_buffers(self.display, self.raw).unwrap();
-        self.egl
```
Gordon-F, line 599: I'm not sure about these changes. I don't know why we should make the current context without surface every frame?

src/backend/gl/src/window/egl.rs
```diff
@@ -114,11 +115,9 @@ impl Inner {
             return Err(hal::UnsupportedBackend);
         }
         let display_ext_list = display_extensions.split_whitespace().collect::<Vec<_>>();
-        let required_display_extensions = ["EGL_KHR_create_context", "EGL_KHR_surfaceless_context"];
-        for required_ext in required_display_extensions.iter() {
-            if !display_ext_list.contains(required_ext) {
-                log::warn!("{} is not present", required_ext);
-            }
+        if !display_ext_list.contains(&"EGL_KHR_create_context") {
```
kvark, line 118: do we actually require this? My impression was that we could get away even without it

src/backend/gl/src/window/egl.rs
```diff
@@ -285,10 +285,27 @@ impl hal::Instance<crate::Backend> for Instance {
...
+        if inner.supports_surfaceless_context {
+            inner
+                .egl
+                .make_current(inner.display, None, None, Some(inner.context))
+                .unwrap();
+        } else {
+            let attributes = [egl::WIDTH, 1, egl::HEIGHT, 1, egl::NONE];
+            let surface = inner
+                .egl
+                .create_pbuffer_surface(inner.display, inner.config, &attributes)
```
kvark, line 297: Why do we need a PBuffer surface? We request the surface-less context only because we can't use the provided surfaces efficiently. If the context has to have them, it's ok, we shouldn't need to do anything, still

src/backend/gl/src/window/egl.rs
kvark: It's just returning everything back to what it is before the `present()` is called. We could technically leave the surface bound, but it could introduce accidental code errors when something is rendering to the default framebuffer, and we aren't catching a GL error on that (instead corrupting one of the swapchains). Note that the `self.context` can work with different surfaces, not just one.

kvark (commented):
Thank you for the PR!

src/backend/gl/src/window/egl.rs
Gordon-F: You mean `create_window_surface` and destroy after `make_current` ?

src/backend/gl/src/window/egl.rs
Gordon-F: Any ideas how we can get around this limitation?

src/backend/gl/src/window/egl.rs
kvark: Sorry, I don't understand. Let me try again.

So this is creating a PBuffer because otherwise we can't call `make_current`, correct?
In this case, could we perhaps make it a 1x1 buffer instead of having the full width/height?

Also, we need to keep it around, so that we can call `make_current` later again (see the other comment we are discussing)

src/backend/gl/src/window/egl.rs
kvark: Yep, we can keep the `PBuffer` around and use it here when calling `make_current`

src/backend/gl/src/window/egl.rs
Gordon-F: > So this is creating a PBuffer because otherwise we can't call make_current, correct?

Correct.

> In this case, could we perhaps make it a 1x1 buffer instead of having the full width/height?

```rust
let attributes = [egl::WIDTH, 1, egl::HEIGHT, 1, egl::NONE];
```

It should be 1x1.

Gordon-F:
@kvark I think I fixed all suggestions. Thank you for reviewing!

src/backend/gl/src/window/egl.rs
```diff
@@ -285,10 +298,23 @@ impl hal::Instance<crate::Backend> for Instance {
 
     fn enumerate_adapters(&self) -> Vec<hal::adapter::Adapter<crate::Backend>> {
         let inner = self.inner.lock();
-        inner
-            .egl
-            .make_current(inner.display, None, None, Some(inner.context))
-            .unwrap();
+        if let Some(pbuffer) = inner.pbuffer {
+            inner
```
kvark, line 302: do we actually need this `if`? couldn't we just pass `make_current(inner.display, inner.pbuffer, inner.pbuffer, ..)` right away?

src/backend/gl/src/window/egl.rs
```diff
@@ -596,9 +625,21 @@ impl Surface {
         gl.bind_framebuffer(glow::READ_FRAMEBUFFER, None);
 
         self.egl.swap_buffers(self.display, self.raw).unwrap();
-        self.egl
-            .make_current(self.display, None, None, Some(self.context))
-            .unwrap();
+
+        if let Some(pbuffer) = self.pbuffer {
```
kvark, line 629: same here, perhaps we can remove the `if`?

kvark (commented):
Looks wonderful, just one small nit left!

src/backend/gl/src/window/egl.rs
kvark: Sorry, I just missed it. The code was good!

src/backend/gl/src/window/egl.rs
Gordon-F: Removed `if` and checked on the emulator and device. Everything is working fine.

kvark:
Sweet!
Bors r+

PR #3595, opened 2021-01-16
[vk] don't forget to enable maintenance-1

kvark:
Apparently, enabling maintenance-2 doesn't auto-enable maintenance-1. Who'd new? That manifests in our NDC inversion going broke:
> [1.285790 ERROR]()(no module): 
VALIDATION [VUID-VkViewport-height-01772 (-1596201395)] : Validation Error: [ VUID-VkViewport-height-01772 ] Object 0: handle = 0x559394238118, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xa0dbe64d | vkCmdSetViewport: pViewports[0].height (=-600.000000) is not greater 0.0. The Vulkan spec states: height must be greater than 0.0 (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkViewport-height-01772)
object info: (type: COMMAND_BUFFER, hndl: 94092333908248)

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -36,7 +36,7 @@ log = "0.4"
 hal = { path = "../src/hal", version = "0.6", package = "gfx-hal" }
 auxil = { path = "../src/auxil/auxil", version = "0.5", package = "gfx-auxil" }
 gfx-backend-empty = { path = "../src/backend/empty", version = "0.6" }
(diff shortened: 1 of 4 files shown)


kvark:
one tiny fix is added about GLSL versions
bors r+

PR #3596, opened 2021-01-16
[GL] Workaround WebGL's Lack of Per-Target Color Masks

### LINKED ISSUE #3593: Draw buffer color masks are not supported with WebGL
Trying to launch wgpu-rs' hello-triangle example on WebGL leads to a panic telling : `Draw buffer color masks are not supported`

cwfitzgerald:
Fixes #3593

This works around the issue. To avoid redundant state changes in the future, as mentioned in the comment, we should coalesce on the command-generator side to a single global call.


kvark (approved):
Great stuff!
bors r+

PR #3599, opened 2021-01-18
Add support for compute on GLES

kvark:
PR checklist:

### DIFF
--- a/src/backend/gl/src/info.rs
+++ b/src/backend/gl/src/info.rs
@@ -399,7 +399,10 @@ pub(crate) fn query_all(
         limits.max_viewports = get_usize(gl, glow::MAX_VIEWPORTS).unwrap_or(0);
     }
 
-    if info.is_supported(&[Core(4, 3), Ext("GL_ARB_compute_shader")]) {
+    //TODO: technically compute is exposed in Es(3, 1), but GLES requires 3.2
+    // for any storage buffers. We need to investigate if this requirement
+    // can be lowered.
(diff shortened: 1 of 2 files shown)


src/backend/gl/src/info.rs
```diff
@@ -399,7 +399,7 @@ pub(crate) fn query_all(
         limits.max_viewports = get_usize(gl, glow::MAX_VIEWPORTS).unwrap_or(0);
     }
 
-    if info.is_supported(&[Core(4, 3), Ext("GL_ARB_compute_shader")]) {
+    if info.is_supported(&[Core(4, 3), Es(3, 2), Ext("GL_ARB_compute_shader")]) {
```
Gordon-F, line 402: Compute shaders have been available from `OpenGL ES 3.1`. If we require `OpenGL ES 3.0` context, should we add LegacyFeature for compute shaders?

kvark: You are correct, compute is technically available on 3.1. However, the SSBOs require 3.2. Unless we can change the latter, I don't see a point in pretending we have compute shaders on 3.1. Am I missing anything?

Gordon-F: It would be great if we have a comment in code with an explanation of why we should have `OpenGL ES 3.2` for compute shaders 😄

kvark: agreed, will do now

kvark:
bors r=Gordon-F

PR #3603, opened 2021-01-18
Fix RangeAllocator.allocated_ranges returning empty iterator for fully allocated RangeAllocator

Wumpf:
Noticed when trying to use range-alloc for purpose of buffer zero init tracking in wgpu.
(I don't know if this led to other issues down the line)


kvark (approved):
Very nice find, thank you!
bors r+

PR #3604, opened 2021-01-19
[GL] Testing with the Android Emulator on CI

Gordon-F:
Just a proof-of-concept of how we can run a `quad example` on CI with the Android Emulator.

All other jobs temporary removed (for speed up).

TODO:
- [ ] Rollback all CI jobs
- [ ] Investigate why half of builds panicked with error: 
```thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/ndk-glue-0.2.1/src/lib.rs:255:34``` https://github.com/rust-windowing/android-ndk-rs/issues/116
- [ ] Investigate how we can run quad on Android without code duplication
- [ ] Add PR comments with results screenshot and logs

### DIFF
--- a/examples/quad_android/Cargo.toml
+++ b/examples/quad_android/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+name = "quad_android"
+version = "0.1.0"
+publish = false
+workspace = "../.."
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[features]
+default = ["gl"]
+gl = ["gfx-backend-gl"]
+
+[target.'cfg(target_os = "android")'.dependencies]
+android_logger = "0.9"
+image = "0.23.12"
+log = "0.4"
+hal = { path = "../../src/hal", package = "gfx-hal" }
+auxil = { path = "../../src/auxil/auxil", package = "gfx-auxil" }
+gfx-backend-empty = { path = "../../src/backend/empty" }
+winit = "0.24"
+ndk-glue = "0.2"
+
+[target.'cfg(target_os = "android")'.dependencies.gfx-backend-gl]
+path = "../../src/backend/gl"
+optional = true
+
+[package.metadata.android]
+package_name = "com.rust.game"
+label = "GfxQuad"
+# Vulkan supports on devices running Android API level 24 or higher.
+target_sdk_version = 29
+min_sdk_version = 29
+fullscreen = true
+opengles_version = [3, 0]
+orientation = "sensorLandscape"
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: examples/quad_android/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,935 @@
+#[cfg(not(any(
```
kvark, line 1: does this have to be a separate example? could we make it a part of the existing `quad` instead?

kvark (commented):
Exciting stuff!

examples/quad_android/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+publish = false
+workspace = "../.."
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[features]
+default = ["gl"]
+gl = ["gfx-backend-gl"]
```
kvark, line 13: can we have Vulkan as well?

examples/quad_android/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
```
kvark, line 1: another option, not necessarily a better one, is to have `examples/src/lib.rs`, so that the project is a `cdylib` library with binaries. The library would only be doing stuff on Android at this moment, but maybe we can find more use for this later?

kvark (commented):
I think this is starting to look great!

examples/quad_android/Cargo.toml
Gordon-F: I can try to run Vulkan as well after GL stabilization on CI. `reactivecircus/android-emulator-runner@v2` running an emulator on `MacOS host`. We can't easily run the Android emulator on Linux hosts ([based on README info](https://github.com/ReactiveCircus/android-emulator-runner#can-i-use-this-action-on-linux-vms)).

kvark: Sounds quite bizarre tbh that an android emulator would run worse on Linux than on macOS. Oh well...

kvark (approved):

examples/quad_android/Cargo.toml
Gordon-F: `cargo apk` allows run apk file only from cargo workspace root. But cargo doesn't allow pass features in cargo workspace root.

```bash
error: --features is not allowed in the root of a virtual workspace
```

Gordon-F: This is error already fixed in `cargo-apk`. I will try to remove more code duplication with `examples/src/lib.rs`.

Gordon-F:
I'm going to close this PR, since `gfx-hal` is deprecated. I will try to run this PR on top of `wgpu` or `wgpu-hal` when I had more free time.

Issue #3389, opened 2020-09-28
Shader compile times


kvark:
Using spirv-cross appears to be showing up in profiles. We are doing a bit of heap allocation and copying on the way to it and from it today, but also the logic itself may be slow. Hopefully, we can address it by moving to https://github.com/gfx-rs/naga

Here is some input I gathered from a Dota2 run: 
[dota-metal-shader-times.txt](https://github.com/gfx-rs/gfx/files/5294241/dota-metal-shader-times.txt)

All times are in micro-seconds. Here are some observations:
  - spirv-cross time just dominates all the other metrics. It's really a big deal to replace/optimize
  - pipeline creation is comparable to `MTLLibrary` creation of a single stage. It's not much at all in the total picture.

mkeeter:
It turns out there was [accidentally exponential behavior in `SPIRV-Cross`](https://github.com/KhronosGroup/SPIRV-Cross/pull/1594) – not sure if this was being hit, but it's a possibility!

The `spirv_cross` crate has updated to the latest upstream in [0.23.0](https://github.com/grovesNL/spirv_cross/pull/166).  Depending on when `naga` will be ready, maybe it's worth bumping the `spirv_cross` dependency for a point release?  I'd be happy to make a PR and see if anything breaks, though I don't have a full set of cross-platform hardware for testing.

(I'm biased, because I'm running into this behavior in an app using `wgpu-native`, so I'm trying to push the fix all the way downstream 😄)

grovesNL:
Accidentally closed

PR #3610, opened 2021-01-21
Update spirv-cross to version 0.23

grovesNL:
Fix the comment in https://github.com/gfx-rs/gfx/issues/3389#issuecomment-763983048


kvark (approved):
Thank you!
bors r+

PR #3611, opened 2021-01-21
[gl] storage memory barriers

kvark:
Fixes the issue discussed on Matrix - the compute example on GL doesn't always show the correct result.
PR checklist:


kvark:
Not a confirmed fix, but let's land this anyway
bors r+

PR #3614, opened 2021-01-26
Remove ExactSizeIterator bounds, update inplace_it usage

### LINKED ISSUE #3609: Consider removing ExactSizeIterator bounds
See https://github.com/NotIntMan/inplace_it/issues/5
Thanks @zakarumych for suggestion!
I think it would be good: it's more convenient to not require this, and in some cases the size hint can be better then `ExactSizeIterator`. For example, `repeat()` doesn't implement it, so `repeat().take()` is unusable with the exact size, but it does in fact have the exact size, and the `size_hint()` reflects that nicely.

kvark:
Fixes #3609
This PR is a product of wonderful collaboration with @zakarumych and @NotIntMan. Thank you guys!

PR checklist:

Tested on wgpu-rs examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change Log
 
 ### hal-unreleased
+  - `Borrow` and `ExactSizeIterator` bounds are removed from the iterators
   - error improvements:
     - use `thiserror` for errors
     - variants and a few names are refactored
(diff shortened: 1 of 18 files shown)


kvark:
fairly confident in this change, also well tested
bors r+

PR #3615, opened 2021-01-26
Be explicit about Iterator in the API

kvark:
Doesn't fix anything, but makes our API a little bit more explicit.

Original motivation for `IntoIterator` was driven by general Rust best practices, including the idea that users would already pass containers to us, such as `Vec`, or `Option`. However, at this point I believe the minor convenience is not worth the additional type layer of `IntoIterator`. In most cases we don't expect actual containers, we expect users to store the data in their own containers, and then iterator to extract it and provide to us. Overall, in gfx-hal we are trying to be explicit about the costs, and seeing an iterator on the call side helps this a bit.

Finally, as an interesting fact, in most cases in Warden the fix was to just... remove `collect()`. Generally, we don't want to build an API that promotes passing `Vec` around. We should discourage this instead.

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -192,7 +192,8 @@ impl<B: Backend> RendererState<B> {
                         ty: pso::DescriptorType::Sampler,
                         count: 1,
                     },
-                ],
+                ]
+                .into_iter(),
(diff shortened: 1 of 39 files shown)


kvark:
Let's just proceed. The main consumers don't care that much, and simplification is worth it.
bors r+

PR #3617, opened 2021-01-27
Use non-ARB functions in bind_blend_slot

### LINKED ISSUE #2685: Mali-T760 OpenGL ES 3.2 has no BlendFuncSeparateiARB and BlendEquationSeparateiARB
Short info header:
- GFX version: 0.18
- OS: Armbian Linux ARMv7
- GPU: Mali-T760
- GL_VERSION: GL_VERSION: OpenGL ES 3.2 v1.r14p0-01rel0-git(966ed26).eedaf0c5244add8e249cecdd3c721329
- OpenGL ES 3.2

Mali does not Support BlendFuncSeparateiARB but has BlendFuncSeparatei
and does not Support BlendEquationSeparateiARB but has BlendEquationSeparatei.

after switch to BlendEquationSeparatei and BlendFuncSeparatei Program runs.

https://github.com/gfx-rs/gfx/blob/e3199acf367eea00bcb38aca6eb9ebfae8d1e116/src/backend/gl/src/state.rs#L258

alula:
Fixes #2685

PR checklist:
- gl

### DIFF
--- a/src/backend/gl/src/state.rs
+++ b/src/backend/gl/src/state.rs
@@ -253,13 +253,12 @@ pub fn bind_blend_slot(gl: &gl::Gl, slot: ColorSlot, color: s::Color) {
     let buf = slot as gl::types::GLuint;
     match color.blend {
         Some(b) => unsafe {
-            //Note: using ARB functions as they are more compatible
             gl.Enablei(gl::BLEND, buf);
-            gl.BlendEquationSeparateiARB(buf,
+            gl.BlendEquationSeparatei(buf,
                 map_equation(b.color.equation),
                 map_equation(b.alpha.equation)
             );
-            gl.BlendFuncSeparateiARB(buf,
+            gl.BlendFuncSeparatei(buf,
                 map_factor(b.color.source),
                 map_factor(b.color.destination),
                 map_factor(b.alpha.source),


alula:
is pre-ll branch even maintained somewhere?

kvark:
@alula it's not really maintained any more. We have enough on our plates beside that. If you are interested, we can arrange you the power to do that.

kvark:
I think what makes a difference here is the GLES vs GL. On GL, the ARB extensions are more compatible, since they will be there all the time, and the core versions may not. But on GLES, the core is relatively capable, and at times no extensions even are developed for something in core. So for GLES, we'd use the non-ARB, and for GL we'd use ARB.

Anyhow, the CI is sad (not because of this PR, but still), and we have this situation with "pre-ll" unmaintained, so unable to merge it.

alula:
why don't move the pre-ll branch to another repo tho?

kvark:
> why don't move the pre-ll branch to another repo tho?

What would that change? The problem is not where it lives, the problem is that we moved to gfx-hal long ago (3 years by now?), and we don't want to maintain the old stuff any more. If there are people who are interested, we can chat (care to jump on `#gfx:matrix:org`? or anywhere else, really), and we'll figure out how to let you maintain it.

PR #3621, opened 2021-01-29
[DX11] Fix debug device creation

### LINKED ISSUE #3112: Direct3D 11 unrecognised in Windows 8.1
Short info header:
- GFX version: 0.4.1
- OS: Windows 8.1
- GPU: Should be the one that comes with the notebook SAMSUNG ELECTRONICS CO., LTD. 300E5M (however I lost the original OS and thus cannot fetch which is the video card type).

Running `cargo run --bin quad --features dx11` produces:

```
AdapterInfo { name: "Microsoft Basic Render Driver", vendor: 5140, device: 140, device_type: DiscreteGpu }
AdapterInfo { name: "Microsoft Basic Render Driver", vendor: 5140, device: 140, device_type: VirtualGpu }
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InitializationFailed', src\libcore\result.rs:1165:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
error: process didn't exit successfully: `C:\Users\PC User\Documents\gfx\target\
debug\quad.exe` (exit code: 101)
```

The host OS is installed from the Microsoft site (Windows 8.1 Single Language). The tool `dxdiag` relates Direct3D 11 is the current DirectX version. Before, this laptop had OpenGL 3+ driver, but changed to this OS and thus now only Direct3D 11 can be used to render.

Some softwares detect Direct3D 11 in the current OS and render with it, including Flash Player. I've seen this with a Stage3D library (Starling).

Gordon-F:
Fixes #3112
PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -699,7 +699,32 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
             // type is not unknown; or if debug device is requested but not
             // present
             if !winerror::SUCCEEDED(hr) {
-                return Err(hal::device::CreationError::InitializationFailed);
+                if cfg!(debug_assertions) {
+                    log::warn!(
+                        "Unable to create a debug device. Trying to recreate a device without D3D11_CREATE_DEVICE_DEBUG flag. More info:\n{}\n{}", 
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!
bors r+

PR #3624, opened 2021-01-30
[vk] Support Vulkan 1.1 and 1.2 and refactor extension loading

### LINKED ISSUE #3302: Investigate using Vulkan api version 1.1+
Currently the Vulkan backend is hardcoded to use version 1.0.0. Changing it to use the newest available version is easy:
```
        let api_version_default = vk::make_version(1, 0, 0);
        let api_version = entry
            .try_enumerate_instance_version()
            .unwrap_or(Some(api_version_default))
            .unwrap_or(api_version_default);
        trace!(
            "Using Vulkan api version {}.{}.{}",
            vk::version_major(api_version),
            vk::version_minor(api_version),
            vk::version_patch(api_version)
        );
```
(and pass this to `vk::ApplicationInfo`)

However, it is currently unknown if this would break any existing API: Many inputs to the gfx-hal api are just passed on to Vulkan as-is. Need to map out changes to Vulkan core in 1.1/1.2/.. and figure out if anything could behave differently in newer versions.

tangmi:
This PR includes the following changes:
- Use `vkEnumerateInstanceVersion` to query for the Vulkan driver version and use `min(driver_version, 1.2)` to create the instance.
  - This does not expose any additional 1.1+ functionality in `gfx-hal`, but should allow for later adding features like descriptor indexing and ray tracing.
- Refactor the device extension code to include an extension resolver that correctly includes and excludes requested extensions based on extension promotion status.
- `PhysicalDevice::open` on Vulkan can now fail with `DeviceCreationError::MissingExtension` if a required extension is missing.

Note that while this change can create 1.1 and 1.2 instances, we must opt into 1.3+ whenever that is released.

Some things to consider moving forward:
- API version minor increments should be backwards compatible, per the spec, so we shouldn't need to worry about forward compatibility there.
- We must continue using runtime flags to gate features not available in core Vulkan 1.0.
- When adding an extension that was promoted to core, we must include both its original functionality and any changed functionality after its promotion.
- When adding non-KHR-vendored extensions, we must implement support for all possible versions of Vulkan (since backwards-compatibility is not guaranteed).

Fixes #3302
Related #3146

PR checklist:
  - [x] +succeeds on Windows 10

### DIFF
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -104,7 +104,8 @@ impl Drop for RawInstance {
 }
 
 /// Helper wrapper around `vk::make_version`.
-#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[repr(transparent)]
 pub struct Version(u32);
 
 impl Version {
@@ -112,16 +113,28 @@ impl Version {
     pub const V1_1: Version = Self(vk::make_version(1, 1, 0));
     pub const V1_2: Version = Self(vk::make_version(1, 2, 0));
 
-    pub fn major(self) -> u32 {
+    pub fn make_version(major: u32, minor: u32, patch: u32) -> Self {
+        assert!(major < (1 << 10));
+        assert!(minor < (1 << 10));
+        assert!(patch < (1 << 12));
+        Self(vk::make_version(major, minor, patch))
+    }
+
+    pub const fn major(self) -> u32 {
         vk::version_major(self.0)
     }
 
-    pub fn minor(self) -> u32 {
+    pub const fn minor(self) -> u32 {
         vk::version_minor(self.0)
     }
 
-    pub fn patch(self) -> u32 {
+    pub const fn patch(self) -> u32 {
         vk::version_patch(self.0)
+    }
+
+    /// Return a `Version` with the major and minor components copied and the patch component set to 0.
+    pub const fn without_patch(self) -> Self {
+        Self(vk::make_version(self.major(), self.minor(), 0))
     }
 }
 
@@ -945,15 +958,8 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
                     mesh_shaders: mesh_fn,
                     draw_indirect_count: indirect_count_fn,
                 },
-                supports_y_flip: {
-                    if self.api_version == Version::V1_0 {
-                        self.supports_extension(vk::KhrMaintenance1Fn::name())
-                            || self.supports_extension(vk::KhrMaintenance2Fn::name())
-                            || self.supports_extension(vk::KhrMaintenance3Fn::name())
-                    } else {
-                        true
-                    }
-                },
+                flip_y_requires_shift: self.api_version >= Version::V1_1
+                    || self.supports_extension(vk::KhrMaintenance1Fn::name()),
                 imageless_framebuffers,
                 timestamp_period: self.properties.limits.timestamp_period,
             }),
@@ -1530,7 +1536,10 @@ pub struct RawDevice {
     features: Features,
     instance: Arc<RawInstance>,
     extension_fns: DeviceExtensionFunctions,
-    supports_y_flip: bool,
+    /// The `hal::Features::NDC_Y_UP` flag is implemented with either `VK_AMD_negative_viewport_height` or `VK_KHR_maintenance1`/1.1+. The AMD extension for negative viewport height does not require a Y shift.
+    ///
+    /// This flag is `true` if the device has `VK_KHR_maintenance1`/1.1+ and `false` otherwise (i.e. in the case of `VK_AMD_negative_viewport_height`).
+    flip_y_requires_shift: bool,
     imageless_framebuffers: bool,
     timestamp_period: f32,
 }
@@ -1555,7 +1564,7 @@ impl RawDevice {
 
     fn map_viewport(&self, rect: &hal::pso::Viewport) -> vk::Viewport {
         let flip_y = self.features.contains(hal::Features::NDC_Y_UP);
-        let shift_y = flip_y && self.supports_y_flip;
+        let shift_y = flip_y && self.flip_y_requires_shift;
         conv::map_viewport(rect, flip_y, shift_y)
     }
 
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -52,6 +52,7 @@ use hal::{
 
 use std::{
     borrow::Cow,
+    cmp,
     ffi::{CStr, CString},
     fmt, mem, slice,
     sync::Arc,
@@ -379,8 +380,7 @@ impl hal::Instance<Backend> for Instance {
                 .unwrap_or(std::ptr::null_mut())
         });
 
-        // Pick the latest version of Vulkan available.
-        let api_version = match entry.try_enumerate_instance_version() {
+        let driver_api_version = match entry.try_enumerate_instance_version() {
             // Vulkan 1.1+
             Ok(Some(version)) => version.into(),
 
@@ -402,7 +402,22 @@ impl hal::Instance<Backend> for Instance {
             .application_version(version)
             .engine_name(CStr::from_bytes_with_nul(b"gfx-rs\0").unwrap())
             .engine_version(1)
-            .api_version(api_version.into());
+            .api_version({
+                // Pick the latest API version available, but don't go later than the SDK version used by `gfx_backend_vulkan`.
+                cmp::min(driver_api_version, {
+                    // This is the max Vulkan API version supported by `gfx_backend_vulkan`.
+                    //
+                    // If we want to increment this, there are some things that must be done first:
+                    //  - Audit the behavioral differences between the previous and new API versions.
+                    //  - Audit all extensions used by this backend:
+                    //    - If any were promoted in the new API version and the behavior has changed, we must handle the new behavior in addition to the old behavior.
+                    //    - If any were obsoleted in the new API version, we must implement a fallback for the new API version
+                    //
+                    // TODO: This should be replaced by `vk::HEADER_VERSION_COMPLETE` (added in `ash@6f488cd`) and this comment moved to either `README.md` or `Cargo.toml`.
+                    Version::V1_2
+                })
+                .into()
+            });
 
         let instance_extensions = entry
             .enumerate_instance_extension_properties()
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -52,7 +52,6 @@ use hal::{
 
 use std::{
     borrow::Cow,
-    collections::{HashMap, HashSet},
     ffi::{CStr, CString},
     fmt, mem, slice,
     sync::Arc,
@@ -67,6 +66,7 @@ use shared_library::dynamic_library::{DynamicLibrary, SpecialHandles};
 mod command;
 mod conv;
 mod device;
+mod extensions_resolver;
 mod info;
 mod native;
 mod pool;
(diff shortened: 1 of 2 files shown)


src/backend/vulkan/src/lib.rs
```diff
@@ -327,7 +327,21 @@ impl hal::Instance<Backend> for Instance {
...
+            .api_version(
+                // Pick the latest version of Vulkan available.
+                match entry
+                    .try_enumerate_instance_version()
+                    // Ignore the possible `VK_ERROR_OUT_OF_HOST_MEMORY`.
+                    .unwrap()
+                {
+                    // Vulkan 1.1+
+                    Some(version) => {
+                        vk::make_version(vk::version_major(version), vk::version_minor(version), 0)
```
kvark, line 339: We have code that tracks the maintenance version, and other code that checks for it. Both currently assume 1.0 baseline, and they obviously need to be adjusted if running on 1.1. I think what we'll do is only patch the first half: code that determines what maintenance level to request would associate Vulkan-1.1 with level 3, and Vulkan-1.1 with level 4. The code that checks for it would then be unchanged.

src/backend/vulkan/src/lib.rs
```diff
@@ -327,7 +327,21 @@ impl hal::Instance<Backend> for Instance {
             .application_version(version)
             .engine_name(CStr::from_bytes_with_nul(b"gfx-rs\0").unwrap())
             .engine_version(1)
-            .api_version(vk::make_version(1, 0, 0));
+            .api_version(
+                // Pick the latest version of Vulkan available.
+                match entry
+                    .try_enumerate_instance_version()
+                    // Ignore the possible `VK_ERROR_OUT_OF_HOST_MEMORY`.
+                    .unwrap()
```
kvark, line 335: you can just include this into the patterns instead, i.e. `Ok(Some(version)) => {...}`

kvark (commented):
Exciting stuff!

src/backend/vulkan/src/lib.rs
tangmi: Is it an error if I request maintenence1 and version 1.1? If that's the case we should also try and tackle the extension dependency graph issue in this PR as well.

Is it fair to assume that we want to always enable the highest maintenance level available level available to us (for our given api version)?

kvark: > Is it an error if I request maintenence1 and version 1.1?

It is technically deprecated in Vk1.1, so there is no guarantee it's even going to be exposed in Vk1.1 drivers. I don't think we should try requesting it.

> Are you suggesting using maintenance_level: u8 as a value to represent "api 1.0, maintenance 1, 2, 3, api 1.1, and 1.2, etc"?

For all the code past initialization, `maintenance_level` defines what we can and what we can't ask the Vulkan driver. So this code can probably for the most part not care about the instance version. On Instances 1.0 and 1.1 we could just set the `maintenance_level = 3` (without requesting the actual levels) and the rest of the code will be fine.

tangmi: I'm working on a little bit of code that will resolve the requested extensions list given an api version, the extensions dependencies, and the deprecation state. The goal is to make it straightforward to add more extensions later.

tangmi: @kvark I've pushed what I came up with (here's the specific commit https://github.com/gfx-rs/gfx/pull/3624/commits/c749e50d99f0518b3423ccaa1a3f4fd3e5e9263c). This is starting to smell like "overengineering," so please sanity check me on whether or not it is indeed an improvement!

src/backend/vulkan/src/lib.rs
```diff
@@ -321,13 +355,33 @@ impl hal::Instance<Backend> for Instance {
...
+        // Pick the latest version of Vulkan available.
+        let api_version = match entry.try_enumerate_instance_version() {
+            // Vulkan 1.1+
+            Ok(Some(version)) => Version(version),
+
+            // Vulkan 1.0
+            Ok(None) => Version::V1_0,
+
+            // Ignore out of memory since it's unlikely to happen and `Instance::create` doesn't have a way to express it in the return value.
+            Err(err) if err == vk::Result::ERROR_OUT_OF_HOST_MEMORY => {
```
kvark, line 367: panicking here doesn't sounds like "ignore" :)
Should we just return `hal::UnsupportedBackend` instead on error? We can log the actual error in `warn!`

src/backend/vulkan/src/lib.rs
```diff
@@ -321,13 +355,33 @@ impl hal::Instance<Backend> for Instance {
...
+        };
+
         let app_name = CString::new(name).unwrap();
         let app_info = vk::ApplicationInfo::builder()
             .application_name(app_name.as_c_str())
             .application_version(version)
             .engine_name(CStr::from_bytes_with_nul(b"gfx-rs\0").unwrap())
             .engine_version(1)
-            .api_version(vk::make_version(1, 0, 0));
+            .api_version(vk::make_version(
```
kvark, line 380: any reason this isn't just `api_version`? are we trying to strip the patch index?

src/backend/vulkan/src/lib.rs
```diff
@@ -680,6 +735,152 @@ pub struct DeviceCreationFeatures {
...
+                .into_iter()
+                .map(|extension| (extension.name, extension))
+                .collect(),
+        }
+    }
+
+    /// Resolve `requested_extensions` into the full list of transitive dependencies and filter out any extensions that are not supported or are no longer explicitly needed.
+    ///
+    /// In general, `requested_extensions` should include the extension list needed to support the oldest version of Vulkan.
+    // TODO: Should we signal if an extension that was request could not be added?
```
kvark, line 804: yeah, looks like we need to define the interface here a bit stronger. Does this function guarantee the resolve? or are some of the requested extensions optional?

src/backend/vulkan/src/lib.rs
```diff
@@ -680,6 +735,152 @@ pub struct DeviceCreationFeatures {
...
+                            extension.name.to_string_lossy()
+                        );
+                    }
+                }
+
+                // Ensure we don't walk the dependency graph more than once.
+                if marked.contains(extension.name) {
+                    continue;
+                }
+                marked.insert(extension.name);
```
kvark, line 869: I think `insert` should return a `bool` that you can use instead of checking for `.contains()`

src/backend/vulkan/src/lib.rs
```diff
@@ -680,6 +735,152 @@ pub struct DeviceCreationFeatures {
...
+
+                for &dependency in &extension.dependencies {
+                    remaining.push(dependency);
```
kvark, line 872: should this be `remaining.extend_from_slice(&extension.dependencies)`?

src/backend/vulkan/src/lib.rs
```diff
@@ -701,64 +902,93 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
             return Err(DeviceCreationError::MissingFeature);
         }
 
+        let extensions_registry = ExtensionsRegistry::new(vec![
```
kvark, line 905: we should probably move all the version/extension stuff into a module

src/backend/vulkan/src/lib.rs
```diff
@@ -701,64 +902,93 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+            Extension::new(vk::AmdNegativeViewportHeightFn::name(), Version::V1_0)
+                .promoted_to(Version::V1_1),
+            // https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_descriptor_indexing.html
+            // Depends on `VK_KHR_get_physical_device_properties2`, which is an instance extension
+            Extension::new(vk::ExtDescriptorIndexingFn::name(), Version::V1_0)
+                .with_dependencies(vec![vk::KhrMaintenance3Fn::name()])
+                .promoted_to(Version::V1_2),
+            // https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_sampler_filter_minmax.html
+            // Depends on `VK_KHR_get_physical_device_properties2`, which is an instance extension
+            Extension::new(vk::ExtSamplerFilterMinmaxFn::name(), Version::V1_0)
```
kvark, line 916: it sounds to me that all this information is supposed to be a part of the spec, so it could be upstreamed to `Ash`. Would you want to file a small issue to them to check if they are interested?

src/backend/vulkan/src/lib.rs
```diff
@@ -1424,7 +1665,8 @@ pub struct RawDevice {
     features: Features,
     instance: Arc<RawInstance>,
     extension_fns: DeviceExtensionFunctions,
-    maintenance_level: u8,
+    /// Indicates this device is running on Vulkan 1.0 with no maintenance extensions.
+    is_core_1_0: bool,
```
kvark, line 1669: it seems confusingly general, while the real purpose is to check for Y-flip. We should make it either more versatile (to allow other pieces to rely on it), like the `maintenance_level` was, or more specific, i.e. `supports_y_flip: bool`

kvark (commented):
Looks very nice! Just a few notes before we proceed

src/backend/vulkan/src/lib.rs
tangmi: You're totally right. I remember reading here (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap5.html#VkApplicationInfo) that the patch value is ignored, but I don't see any reason why we should strip it out. I'll just pass the `api_version` as is.

src/backend/vulkan/src/lib.rs
tangmi: I'll make it return a `Result` that indicates an expected extension didn't make it. In `PhysicalDevice::open` it'll get mapped to `DeviceCreationError::MissingExtension`.

src/backend/vulkan/src/lib.rs
tangmi: Doing C++ std for my day job sometimes pushes parts of Rust std out of my brain 😁

src/backend/vulkan/src/lib.rs
tangmi: I'll file an issue and resolve this with a link once I do

src/backend/vulkan/src/lib.rs
tangmi: I'll do `supports_y_flip: bool`. Is there a case in `map_viewport` where `flip_y` can be true but `shift_y` isn't? I suspect I've added a bug since I don't think I understand this bit of code.

Is it that vulkan 1.0 (+ amd negative viewport) didn't require shifting the viewport rect to flip it?

src/backend/vulkan/src/lib.rs
tangmi: https://github.com/MaikKlein/ash/issues/367

src/backend/vulkan/src/lib.rs
kvark: It's the AMD extension that didn't require the shifting, yeah, but the normal stuff (i.e. maintenance or vk1.1) does need the shift.

tangmi: I've named it `flip_y_requires_shift` and added a comment to the field.

src/backend/vulkan/src/lib.rs
tangmi: @kvark I took a look at a couple Vulkan engines and it seems most just flatten dependencies and hard code the list.

Some examples:
- The Forge: https://github.com/ConfettiFX/The-Forge/blob/2edbb394883b971ba30498f9175c00bf834db9c6/Common_3/Renderer/Vulkan/Vulkan.cpp#L225
- Diligent: https://github.com/DiligentGraphics/DiligentCore/blob/fa345f59de694ac249c14e987032ef3e341faaa7/Graphics/GraphicsEngineVulkan/src/EngineFactoryVk.cpp#L240

I'm tempted to remove this "extension resolver" code and just hard code the list like before, thoughts?

kvark:
you rock!
bors r+

PR #3626, opened 2021-02-01
[hal] Merge `Limits` and `Capabilities` as `PhysicalDeviceProperties`

tangmi:
This PR:
- Merges `hal::Limits` and `hal::Capabilities` as `PhysicalDeviceProperties`.
  - The name was chosen to be more similar to Vulkan.
- Adds properties structs for mesh shaders and descriptor indexing.
- Adds a `DescriptorLimits` struct to hold descriptor limits and makes the values `u32`.

PR checklist:
  - [x] bench: dx12, vulkan
  - [x] colour-uniform: dx12, vulkan
  - [x] compute: dx12, vulkan
  - [x] quad: dx12, vulkan

### DIFF
--- a/src/hal/src/lib.rs
+++ b/src/hal/src/lib.rs
@@ -105,11 +105,13 @@ bitflags! {
         /// Bit mask of Vulkan Core/Extension features.
         const CORE_MASK = 0xFFFF_FFFF_FFFF_FFFF;
         /// Bit mask of Vulkan Portability features.
-        const PORTABILITY_MASK  = 0x0000_FFFF_0000_0000_0000_0000;
+        const PORTABILITY_MASK  = 0xFFFF << 64;
         /// Bit mask for extra WebGPU features.
-        const WEBGPU_MASK = 0xFFFF_0000_0000_0000_0000_0000;
+        const WEBGPU_MASK = 0xFFFF << 80;
         /// Bit mask for all extensions.
-        const EXTENSIONS_MASK = 0xFFFF_FFFF_0000_0000_0000_0000_0000_0000;
+        const EXTENSIONS_MASK = 0xFFFF_FFFF << 96;
+
+        // Bits for Vulkan Core/Extension features
 
         /// Support for robust buffer access.
         /// Buffer access by SPIR-V shaders is checked against the buffer/image boundaries.
@@ -238,8 +240,13 @@ bitflags! {
         const STORAGE_TEXTURE_DESCRIPTOR_INDEXING = 0x0400_0000_0000_0000;
         /// Allow descriptor arrays to be unsized in shaders
         const UNSIZED_DESCRIPTOR_ARRAY = 0x0800_0000_0000_0000;
+        /// Mask for all the features associated with descriptor indexing.
+        const DESCRIPTOR_INDEXING_MASK = Features::SAMPLED_TEXTURE_DESCRIPTOR_INDEXING.bits | Features::STORAGE_TEXTURE_DESCRIPTOR_INDEXING.bits | Features::UNSIZED_DESCRIPTOR_ARRAY.bits;
+
         /// Enable draw_indirect_count and draw_indexed_indirect_count
         const DRAW_INDIRECT_COUNT = 0x1000_0000_0000_0000;
+
+        // Bits for Vulkan Portability features
 
         /// Support triangle fan primitive topology.
         const TRIANGLE_FAN = 0x0001 << 64;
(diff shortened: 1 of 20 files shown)


src/hal/src/lib.rs
```diff
@@ -469,7 +488,64 @@ pub struct Limits {
 
     /// The alignment of the vertex buffer stride.
     pub min_vertex_input_binding_stride_alignment: buffer::Offset,
+}
 
+/// Resource limits related to Descriptor Indexing.
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct DescriptorIndexingCapabilities {
```
tangmi, line 496: This is a direct mapping from `vkPhysicalDeviceDescriptorIndexingPropertiesEXT`, which is probably more than what we want for `gfx`.

src/hal/src/lib.rs
```diff
@@ -467,7 +486,64 @@ pub struct Limits {
...
+    ///
+    pub max_per_stage_descriptor_update_after_bind_input_attachments: u32,
+    ///
+    pub max_per_stage_update_after_bind_resources: u32,
+    ///
+    pub max_descriptor_set_update_after_bind_samplers: u32,
+    ///
+    pub max_descriptor_set_update_after_bind_uniform_buffers: u32,
+    ///
+    pub max_descriptor_set_update_after_bind_uniform_buffers_dynamic: u32,
```
kvark, line 530: we have quite a few places for `max_XXX_descriptor_XXX_some_stage`. I wonder if we can make this nicer? I.e.
```rust
struct BindLimits {
  max_uniform_buffers: u32,
  max_samplers: u32,
  ...
}
pub per_stage_descriptor_update_limits: BindLimits,
pub descriptor set_update_limits: BindLimits,
```
We'd probably use it in the `Limits` too!

kvark (approved):
Very nice!

src/hal/src/lib.rs
tangmi: I think we'll need to get rid of the `*_update_after_bind_*` entries here, since `gfx` doesn't (yet?) support that.

I also noticed most of the core `Limits` are `usize` when the actual Vulkan value is `u32`, is this intentional? And should I follow that pattern for mesh shader and descriptor indexing? The best reason I can come up with is that these values will likely be used in calculations to size rust vecs or arrays, which would operate on `usize`.

kvark: Not intentional, we should probably change them to `u32`.

kvark: So are we removing the `_update_after_bind_` for now?

kvark (approved):
LGTM, thank you for doing this!
Just one small question

src/hal/src/lib.rs
tangmi: Admittedly I'm not sure what we'd need for this feature. I've removed the update-after-bind stuff because gfx itself doesn't support that feature. I believe update-after-bind is pretty important for descriptor indexing--although, again, I could be wrong here.

@cwfitzgerald I recall we needed descriptor indexing to handle a bug in Intel drivers for WebGPU... can I defer to you for what is important to expose as a part of `hal` for descriptor indexing? Were you imagining this being a mostly "private capabilities"-type feature or something publicly exposed?

kvark: this is the stuff from here:
> VK_DESCRIPTOR_BINDING_UPDATE_AFTER_BIND_BIT indicates that if descriptors in this binding are updated between when the descriptor set is bound in a command buffer and when that command buffer is submitted to a queue, then the submission will use the most recently set descriptors for this binding and the updates do not invalidate the command buffer. Descriptor bindings created with this flag are also partially exempt from the external synchronization requirement in vkUpdateDescriptorSetWithTemplateKHR and vkUpdateDescriptorSets. Multiple descriptors with this flag set can be updated concurrently in different threads, though the same descriptor must not be updated concurrently by two threads. Descriptors with this flag set can be updated concurrently with the set being bound to a command buffer in another thread, but not concurrently with the set being reset or freed.

We don't have that supported yet, so we can leave this off PR.
However, it would be nice to at least group those descriptor limits together in a struct, so that if/when we need to add another set of limits for the "update_after_bind", we can just re-use the same struct type.

tangmi: I'll do another pass to see what I can group together... I remember the last time I looked at it (briefly!) every group of limits were just slightly different from another and didn't meaningfully have an abstraction to pull out.

I'd also like to wait on @cwfitzgerald just to make sure this is sufficient for the Intel/WebGPU work?

tangmi: Also removed the _update_after_bind_ fields

kvark (approved):
Could we squash some of the commits (that don't build independently) before merging?

src/hal/src/lib.rs
cwfitzgerald: So for update after bind, the idea is that it actually introduces no more requirements on the API to just use update after bind descriptors like normal descriptors. i.e. update after bind descriptors are a superset of non-update after bind descriptors, so we should just implicitly use them to raise the limits on intel.

kvark:
@tangmi Let's rebase and merge

kvark (approved):
We'll need to add a changelog entry at some point, but everything looks good!
bors r+

PR #3633, opened 2021-02-09
Update libloading to 0.7

yzsolt:
Just a (hopefully) harmless update. Updated in `d2d12-rs` too: https://github.com/gfx-rs/d3d12-rs/pull/28

PR checklist:
- [x] tested examples with the following backends: GL


kvark (commented):
Thank you!
Could we switch the `d3d12` dependency to the latest revision on github in this PR, so that libloading is updated consistently?

kvark (approved):
bors r+

PR #3634, opened 2021-02-09
Use libc for the use-rtld-next Vulkan backend feature

### LINKED ISSUE #3540: Replace shared_library by libloading in Vulkan backend
We use libloading on other backends.

yzsolt:
`RTLD_NEXT` is a GNU-specific flag which isn't supported currently by libloading, so we can't replace shared_library with it. However, we can simply use `dlsym` directly, and since libc is already present in the dependency chain, in practice it still means building one less dependency.

Closes #3540

PR checklist:

### DIFF
--- a/src/backend/vulkan/Cargo.toml
+++ b/src/backend/vulkan/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 
 [features]
 default = []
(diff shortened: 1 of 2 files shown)


kvark (approved):
Nice!
bors r+

PR #3635, opened 2021-02-11
[vk] Replace extension resolver with hard-coded extensions list

tangmi:
PR checklist:
- [x] tested examples with the following backends: colour-uniform, quad, compute, bench on vulkan


kvark (approved):
As discussed on the matrix, this seems like a good thing.
Thank you!
bors r+

PR #3638, opened 2021-02-13
[vk] Print IDs as hex in DebugUtil output

tangmi:
The Vulkan validation layer outputs messages that reference object handles and message IDs in a hexadecimal format and this change makes `gfx`'s custom `DebugUtil` output consistent with that.


kvark (approved):
Thank you!
bors r+

Issue #3117, opened 2020-01-02
Working around `spirv-cross` by pre-compiling shaders / pre-preparing pipelines


mitchmindtree:
I recently asked in the gfx riot channel about the feasibility of pre-compiling shaders for each target as an alternative to depending on `spirv-cross`:

> Hi folks! Is it possible to opt-out of using the spirv-cross dependency in favour of pre-compiling the shaders for each target?

@kvark provided some insight:

> Basically, there are many factors that affect how the native shader is generated: pipeline layout, specialization, capabilities of the target platform, and other things. The best we can do if you want to run without spirv_cross is: pre-polulate pipeline caches and ship them, add a mode that just returns an error if you are creating a pipeline that is not in the given cache.
That path would require a few changes, mostly getting the pipeline cache to a good state (currently, DX12/11 don't implement it).

I've opened this issue so that it's easier to track/discuss this. While I'm at it, I thought I'd add some of our motivation for working around `spirv-cross` as we transition to `gfx` for [`nannou`](https://github.com/nannou-org/nannou)'s graphics:

We've had a few people testing our `conrod-rendy` backend on macOS run into some compilation issues that required updating their xcode toolchain to build `spirv-cross` properly. We're aiming to remove the need for any of the 3rd party C deps as they (particularly `shaderc`) have been the source of the vast majority of our issues at the nannou repo over the past couple years and working around them are often the biggest blockers in our creative coding workshops. So far we're managing to avoid `shaderc` and `spirv-reflect` by pre-compiling the spirv and specifying layout manually, I think `spirv-cross` is that last we want to work around before we should have a theoretically smooth story for building on all platforms.

The [javelin project](https://gfx-rs.github.io/2019/07/13/javelin.html) looks like a very exciting long-term solution to this, but I wonder if it's worth spending time enabling some feature-gated workaround in the meantime? I'm not sure I'd want to ask to sacrifice time that could have been spent on `javelin` though :)

Skepfyr:
It's going to be a while before javelin is ready (but it's great to hear its needed), and I think that precompiled shaders could be useful even with it, as it can't be faster than that.

Interestingly even with spir-v cross enabled it would be nice to allow a cache that falls back to the slow way if it misses.

freesig:
[...]

Issue #2520, opened 2018-12-20
error: failed to run custom build command for `spirv_cross v0.12.0`


oncealong:
Short info header:
- GFX version: master
- OS: macos 10.14
- GPU: Radeon Pro 580 8192 MB

just `git clone` and `cd gfx/examples`, and run `cargo run --bin quad --features metal`.

backtrace:
```
➜  examples git:(master) cargo run --bin quad --features metal
   Compiling spirv_cross v0.12.0
error: failed to run custom build command for `spirv_cross v0.12.0`
process didn't exit successfully: `/Users/along/workplace/rust/gfx/target/debug/build/spirv_cross-bb9726a46787fef7/build-script-build` (exit code: 101)
--- stdout
OPT_LEVEL = Some("0")
TARGET = Some("x86_64-apple-darwin")
HOST = Some("x86_64-apple-darwin")
CXX_x86_64-apple-darwin = None
CXX_x86_64_apple_darwin = None
HOST_CXX = None
CXX = None
CXXFLAGS_x86_64-apple-darwin = None
CXXFLAGS_x86_64_apple_darwin = None
HOST_CXXFLAGS = None
CXXFLAGS = None
DEBUG = Some("true")
CXX_x86_64-apple-darwin = None
CXX_x86_64_apple_darwin = None
HOST_CXX = None
CXX = None
CXXFLAGS_x86_64-apple-darwin = None
CXXFLAGS_x86_64_apple_darwin = None
HOST_CXXFLAGS = None
CXXFLAGS = None
running: "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-Wall" "-Wextra" "-o" "/Users/along/workplace/rust/gfx/target/debug/build/spirv_cross-cf2c702c55917d00/out/src/wrapper.o" "-c" "src/wrapper.cpp"
cargo:warning=src/wrapper.cpp:1:9: warning: unknown pragma ignored [-Wunknown-pragmas]
cargo:warning=#pragma warning(disable : 4996 4101)
cargo:warning=        ^
cargo:warning=In file included from src/wrapper.cpp:2:
cargo:warning=In file included from src/vendor/SPIRV-Cross/spirv_cross_util.hpp:20:
cargo:warning=In file included from src/vendor/SPIRV-Cross/spirv_cross.hpp:21:
[...]

PR #3641, opened 2021-02-15
Make spirv-cross optional

kvark:
SPIRV-Cross is a big foreign dependency, which often cause issues, such as https://github.com/denoland/rusty_v8/issues/465 and #3389, #3117, #2520, and many others.
Previously in #3445 and #3520 we implemented Naga code path as an alternative, but hidden behind an optional feature. Now it's time to flip the table and say that Naga is permanent, but SPIRV-Cross is optional until completely phased out.

Most of the changes here are from re-routing the main shader creation logic from using SPIRV-Cross types to using Naga types. The "cross" stuff is just getting derived from Naga types if necessary.

The change only applies to Metal and OpenGL backends so far. D3D will come on board once https://github.com/gfx-rs/naga/pull/408 is operational.

Note: it's going to be a long while for Naga to mature, most users need to keep the "cross" dependency enabled for now! Even the included examples don't run on Naga yet, since the SPIR-V generated by glslang is quite specific, and Naga's SPV frontend doesn't like it yet.

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -3,10 +3,10 @@ use crate::{
     info::LegacyFeatures,
     native as n,
     pool::{BufferMemory, CommandPool, OwnedBuffer},
-    state, Backend as B, GlContainer, GlContext, MemoryUsage, Share, Starc, MAX_TEXTURE_SLOTS,
+    state, Backend as B, FastHashMap, GlContainer, GlContext, MemoryUsage, Share, Starc,
+    MAX_TEXTURE_SLOTS,
 };
 
-use auxil::{spirv_cross_specialize_ast, FastHashMap, ShaderStage};
 use hal::{
     buffer, device as d,
     format::{Format, Swizzle},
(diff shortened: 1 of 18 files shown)


src/backend/gl/src/device.rs
```diff
@@ -519,133 +513,137 @@ impl Device {
...
+            self.combine_separate_images_and_samplers(&mut ast, context.reborrow());
+            self.set_push_const_layout(&mut ast).unwrap();
+
+            let glsl = self
+                .translate_spirv_cross(&mut ast, stage, ep.entry)
+                .unwrap();
+            debug!("SPIRV-Cross generated shader:\n{}", glsl);
+            result = Self::create_shader_module_raw(&self.share.context, &glsl, stage);
+        }
+        if !ep.module.prefer_naga {
```
Gordon-F, line 636: @kvark Awesome work! Thank you!

Maybe you mean:
```rust
if ep.module.prefer_naga {
```

Otherwise, we still get naga errors with the enabled `cross` feature.

kvark: The point here is to try Naga in case it was not preferred originally. I was indeed missing a check for the result here, thank you for spotting!

grovesNL (approved):
Thanks! It was a bit hard to follow the MSL changes in detail because of things moving around, but overall everything seems alright with me

src/backend/metal/Cargo.toml
```diff
@@ -16,36 +16,45 @@ edition = "2018"
 default = []
```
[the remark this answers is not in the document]
kvark: That's an interesting idea! I'm generally opposed to default features. If you need it, it's a one time change to add them. If you don't, it's a whole mess out of a sudden.

Given that we are changing gfx-rs on git here, and the issue will likely be discovered instantly if anybody faces it, it seems to me that the risk to cause trouble with this change is very low.

kvark:
Thank you for the review!
Let's keep the option to enable it by default on the table. I'm still thinking about this.
bors r=grovesNL

PR #3654, opened 2021-02-26
[vk] Device creation refactor

tangmi:
This PR is a refactor of the logical device creation code. The intention is to make it simpler and less error-prone to add feature/extension usage to the Vulkan backend by consolidating and re-using some of the logic.

- Moves `PhysicalDevice` into its own file
- Moves all the extension/feature related code into a new type, `PhysicalDeviceFeatures`
  - This type is reused both to query the device capabilities and as input to `VkDeviceCreateInfo`
  - Adds support for `VkPhysicalDeviceVulkan12Features` to test for and enable features in 1.2 when the extension may not be available.
  - Exclude extension-specific `VkPhysicalDevice*Features` structs if we're not requesting the corresponding extensions.
  - Request `VK_KHR_sampler_mirror_clamp_to_edge` when applicable. Closes #3651 
- Pulls the `VkPhysicalDeviceProperties` and `supported_extensions` list into a new type, `PhysicalDeviceInfo`

PR checklist:
- [x] `make` succeeds (on *nix)
- [x] `make reftests` succeeds
- [x] tested examples with the following backends: bench, colour-uniform, compute, quad.
  - Also ran device initialization with `VK_LAYER_LUNARG_device_simulation` with a selection of device reports ranging across the different api versions.

PR #3655, opened 2021-02-27
[vk] Fix some issues caused by #3654

tangmi:
Fixes a couple issues caused by #3654
- `Features::NDC_Y_UP` was not advertised correctly for 1.1
- `Features::SAMPLER_MIRROR_CLAMP_EDGE` could be missed in 1.2 if the driver does not explicitly support `VK_KHR_sampler_mirror_clamp_edge`.
- `Features::DRAW_INDIRECT_COUNT` could be missed in 1.2 if the driver does not explicitly support `VK_KHR_draw_indirect_count`.
- `VK_KHR_draw_indirect_count` was not correctly excluded in 1.2
- `cmd_draw_indirect_count` and `cmd_draw_indexed_indirect_count` were not correctly being loaded in 1.2
  - I'm unsure if there's a way in `ash` to load both the extension function pointer and the core promoted function pointer into the same Rust function, but that would certainly make this code simpler.

### DIFF
--- a/src/backend/vulkan/src/command.rs
+++ b/src/backend/vulkan/src/command.rs
@@ -1,10 +1,15 @@
-use ash::{version::DeviceV1_0, vk};
+use ash::{
+    version::{DeviceV1_0, DeviceV1_2},
+    vk,
+};
 use smallvec::SmallVec;
 use std::{collections::hash_map::Entry, ffi::CString, mem, ops::Range, slice, sync::Arc};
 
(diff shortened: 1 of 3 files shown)


kvark (approved):
That's a scary list of fixes!
Also, an interesting approach with this `enum` thing 👍 
Thank you for taking care of this! We'd be wasting hours hitting the issues in wgpu-rs side.
bors r+

PR #3658, opened 2021-03-04
[dx11] Implement Fill Buffer and Fix Panic

cwfitzgerald:
This fixes the issues on wgpu's shadow example. Now DX11 runs all examples flawlessly.

Works On My Machine™

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -2710,8 +2710,42 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
         unimplemented!()
     }
 
-    unsafe fn fill_buffer(&mut self, _buffer: &Buffer, _sub: buffer::SubRange, _data: u32) {
-        unimplemented!()
+    unsafe fn fill_buffer(&mut self, buffer: &Buffer, sub: buffer::SubRange, data: u32) {
+        let mut device: *mut d3d11::ID3D11Device = mem::zeroed();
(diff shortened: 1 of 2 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -2710,8 +2710,42 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
...
+        desc.Format = dxgiformat::DXGI_FORMAT_R32_TYPELESS;
+        desc.ViewDimension = d3d11::D3D11_UAV_DIMENSION_BUFFER;
+        *desc.u.Buffer_mut() = d3d11::D3D11_BUFFER_UAV {
+            FirstElement: sub.offset as u32 / 4,
+            NumElements: sub.size.unwrap_or(buffer.requirements.size) as u32 / 4,
+            Flags: d3d11::D3D11_BUFFER_UAV_FLAG_RAW,
+        };
+
+        let mut uav: *mut d3d11::ID3D11UnorderedAccessView = ptr::null_mut();
+        let hr = device.CreateUnorderedAccessView(
```
kvark, line 2737: we need to enforce the `D3D11_BIND_UNORDERED_ACCESS` flag at buffer creation in order to be able to do this

kvark: nvm, we already do this

kvark:
The remaining nit isn't that important, we can proceed.
Thank you for the fixes!
bors r+

PR #3660, opened 2021-03-06
[dx11] Support FL10 and expose a beginning set of support flags

cwfitzgerald:
Makes progress towards https://github.com/gfx-rs/wgpu/issues/1244.

I'm not terribly happy with the downlevel feature flags as they are, but I hope to improve them/make them more concrete in a future PR, so I'm not worried about that too much.

Diff isn't very pretty as I didn't split things up.

### DIFF
--- a/src/hal/src/lib.rs
+++ b/src/hal/src/lib.rs
@@ -282,9 +282,9 @@ bitflags! {
         // Bits for Extensions
 
         /// Supports task shader stage.
-        const TASK_SHADER = 0x0000_0001 << 96;
+        const TASK_SHADER = 0x0001 << 96;
         /// Supports mesh shader stage.
-        const MESH_SHADER = 0x0000_0002 << 96;
+        const MESH_SHADER = 0x0002 << 96;
         /// Mask for all the features associated with mesh shader stages.
         const MESH_SHADER_MASK = Features::TASK_SHADER.bits | Features::MESH_SHADER.bits;
     }
@@ -340,6 +340,8 @@ pub struct PhysicalDeviceProperties {
     pub descriptor_indexing: DescriptorIndexingProperties,
     /// Mesh Shader properties.
     pub mesh_shader: MeshShaderProperties,
+    /// Downlevel properties.
+    pub downlevel: DownlevelProperties,
     /// Performance caveats.
     pub performance_caveats: PerformanceCaveats,
     /// Dynamic pipeline states.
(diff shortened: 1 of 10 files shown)


src/hal/src/lib.rs
```diff
@@ -271,11 +271,53 @@ bitflags! {
...
+        const MESH_SHADER = 0x0002 << 96;
         /// Mask for all the features associated with mesh shader stages.
         const MESH_SHADER_MASK = Features::TASK_SHADER.bits | Features::MESH_SHADER.bits;
+
+        // Bits for downlevel hardware and apis
+
+        /// Supports compute shaders
+        const COMPUTE_SHADER = 0x0001 << 112;
+        /// Supports FL5 compute shaders
+        const COMPUTE_SHADER_FL5 = 0x0002 << 112;
```
kvark, line 285: what does this mean in the non-DX context?

src/hal/src/lib.rs
```diff
@@ -271,11 +271,53 @@ bitflags! {
...
+        /// R16G16
+        /// R16G16B16A16
+        /// R8
+        /// R8G8
+        /// R8G8B8A8_UINT
+        /// R8G8B8A8_UNORM
+        ///
+        /// Image -> Buffer
+        /// All copies
+        const DX11_COMPUTE_COPIES = 0x0010 << 112;
```
kvark, line 316: We should not have backend-specific flags like this.
can we instead just say something like `DEVICE_LOCAL_BUFFER_TO_IMAGE_COPIES`? I think the idea of the limit is coming from the fact that DX11 doesn't have buffer-image copies on GPU, but it still can do buffer-image if the buffer is in CPU memory (i.e. UpdateSubresource).

kvark (changes requested):
This looks very solid. Amazing work on DX11 side!
I have a concern about the API side though that we'd need to address.

cwfitzgerald:
Alright, I have touched up the various issues, as well as improved the d3d11 limits a bit. As mentioned on the matrix, I think we should use D3D's concept of shader models as they are very well defined and we can easily map other versioning schemes onto them. I have also de-backend'd the other flags. I haven't included a description of shader what shader modules support but will include a better description once this bubbles up to wgpu.

src/hal/src/lib.rs
```diff
@@ -282,11 +282,30 @@ bitflags! {
...
         /// Supports task shader stage.
-        const TASK_SHADER = 0x0000_0001 << 96;
+        const TASK_SHADER = 0x0001 << 96;
         /// Supports mesh shader stage.
-        const MESH_SHADER = 0x0000_0002 << 96;
+        const MESH_SHADER = 0x0002 << 96;
         /// Mask for all the features associated with mesh shader stages.
         const MESH_SHADER_MASK = Features::TASK_SHADER.bits | Features::MESH_SHADER.bits;
+
+        // Bits for downlevel hardware and apis
```
kvark, line 291: We are sort of tight on flags right now, and want to start migrating things into better places. If I understand correctly, these are all flags, and they don't need to be "enabled" like regular Vulkan features. In this case, let's create a `struct DownlevelProperties`, and make it a member of `struct PhysicalDeviceProperties`. You can see there is already a few things in there :)

We'll probably want to make the `shader_model: u8` or an enum there instead of the flag, too.

kvark (changes requested):
Sorry to bug you again on this, I think we should move this to the new properties API.

cwfitzgerald:
Alright, comments should be addressed, assuming the compile gods are nice to me. You should briefly re-review as some logic changed.

kvark (approved):
Looks good, thank you!
bors r+

PR #3661, opened 2021-03-07
[GL] Improve WebGL version and feature checking

Gordon-F:
Fixes #issue
PR checklist:
- [x] tested examples with the following backends: OpenGL/WebGL

### DIFF
--- a/src/backend/gl/src/info.rs
+++ b/src/backend/gl/src/info.rs
@@ -55,28 +55,38 @@ impl Version {
     /// Note that this function is intentionally lenient in regards to parsing,
     /// and will try to recover at least the first two version numbers without
     /// resulting in an `Err`.
+    /// # Notes
+    /// `WebGL 2` version returned as `OpenGL ES 3.0`
     pub fn parse(mut src: &str) -> Result<Version, &str> {
-        // TODO: Parse version and optional vendor
         let webgl_sig = "WebGL ";
-        let is_webgl = src.contains(webgl_sig);
-        if is_webgl {
-            return Ok(Version {
-                major: 2,
-                minor: 0,
-                is_embedded: true,
-                revision: None,
-                vendor_info: "".to_string(),
-            });
-        }
+        // According to the WebGL specification
+        // VERSION	WebGL<space>1.0<space><vendor-specific information>
+        // SHADING_LANGUAGE_VERSION	WebGL<space>GLSL<space>ES<space>1.0<space><vendor-specific information>
+        let is_webgl = src.starts_with(webgl_sig);
+        let is_es = if is_webgl {
+            let pos = src.rfind(webgl_sig).unwrap_or(0);
+            src = &src[pos + webgl_sig.len()..];
+            true
+        } else {
+            let es_sig = " ES ";
+            match src.rfind(es_sig) {
+                Some(pos) => {
+                    src = &src[pos + es_sig.len()..];
+                    true
+                }
+                None => false,
+            }
+        };
 
-        let es_sig = " ES ";
-        let is_es = match src.rfind(es_sig) {
+        let glsl_es_sig = "GLSL ES ";
+        let is_glsl = match src.find(glsl_es_sig) {
             Some(pos) => {
-                src = &src[pos + es_sig.len()..];
+                src = &src[pos + glsl_es_sig.len()..];
                 true
             }
             None => false,
         };
+
         let (version, vendor_info) = match src.find(' ') {
             Some(i) => (&src[..i], src[i + 1..].to_string()),
             None => (src, String::new()),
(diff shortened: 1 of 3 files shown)


kvark:
Is there a particular reason we want to treat WebGL as a distinct target for the purposes of the compatibility checks?

Gordon-F:
`WebGL` not always equals `OpenGL ES`. These changes should help properly check WebGL combability info. If we don't have `WebGL` as a distinct target we should set current `Version::new_embedded` (3.0/2.0) as a `WebGL` (1.0/2.0) somehow.

Also this changes can be a part of https://github.com/gfx-rs/wgpu/issues/1244.

kvark:
> WebGL not always equals OpenGL ES

I know this is true, but the question is whether these differences are important for our GL backend. Could you give an example of where this distinction would make sense?

> If we don't have WebGL as a distinct target we should set current Version::new_embedded (3.0/2.0) as a WebGL (1.0/2.0) somehow.

Right, I don't think this is going to be too hard.

Gordon-F:
> I know this is true, but the question is whether these differences are important for our GL backend. Could you give an example of where this distinction would make sense?

Ok, Agree with you. They are only theoretically different.

If other code in PR is ok, my plan:
1. Remove `new_webgl` from Version
2. Make `WebGL1` == `OpenGL ES 2.0` and `WebGL2` == `OpenGL ES 3.0` internally

Gordon-F:
Temporarily convert PR to draft because latest changes (`WebGL 2.0` == `OpenGL ES 3.0`) panic in `cube` example.

Gordon-F:
Code panic here: https://github.com/gfx-rs/gfx/blob/4b63f30807204b22f1ba52bb3356e69e19fb896a/src/backend/gl/src/lib.rs#L552-L559

I don't know how we can enable `FRAMEBUFFER_SRGB` for WebGL target 🤔

Gordon-F:
Ok, we don't need to enable `FRAMEBUFFER_SRGB` on embedded targets. Should be ready to merge.

src/backend/gl/src/info.rs
```diff
@@ -55,28 +55,43 @@ impl Version {
...
-        // TODO: Parse version and optional vendor
         let webgl_sig = "WebGL ";
-        let is_webgl = src.contains(webgl_sig);
+        // According to the WebGL specification
+        // VERSION	WebGL<space>1.0<space><vendor-specific information>
+        // SHADING_LANGUAGE_VERSION	WebGL<space>GLSL<space>ES<space>1.0<space><vendor-specific information>
+        let is_webgl = src.starts_with(webgl_sig);
+        let is_es;
+
         if is_webgl {
```
kvark, line 68: nit: would be cleaner to do `let is_es = if is_webgl {... } else {..};`

src/backend/gl/src/queue.rs
```diff
@@ -987,7 +987,7 @@ impl Queue {
                     _ => unsafe { gl.disable(gl_offset) },
                 }
 
-                if !self.share.info.is_webgl() && !self.share.info.version.is_embedded {
+                if !crate::is_webgl() && !self.share.info.version.is_embedded {
```
kvark, line 990: looks like `is_webgl` here is redundant. If it's not ES, then it's not WebGL for sure

src/backend/gl/src/lib.rs
```diff
@@ -671,3 +672,7 @@ fn resolve_sub_range(
     let end = sub.size.map_or(whole.end, |s| whole.start + sub.offset + s);
     whole.start + sub.offset..end
 }
+
+pub(crate) const fn is_webgl() -> bool {
```
kvark, line 676: no need for `pub(crate)` here. Private functions are automatically visible to child modules

src/backend/gl/src/lib.rs
```diff
@@ -553,6 +553,7 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
             .0
             .legacy_features
             .contains(info::LegacyFeatures::SRGB_COLOR)
```
kvark, line 555: can the `LevacyFeatures::SRGB_COLOR` bit already take into account that it's not available on ES? seems strange that we have this `is_embedded` check in addition to it.
Also, something tells me that ES really does support SRGB. Maybe not all ES?

src/backend/gl/src/info.rs
```diff
@@ -55,28 +55,43 @@ impl Version {
...
+                }
+                None => false,
+            };
         }
 
-        let es_sig = " ES ";
-        let is_es = match src.rfind(es_sig) {
+        let glsl_es_sig = "GLSL ES ";
+        let is_glsl;
+        match src.find(glsl_es_sig) {
```
kvark, line 85: same here, let's do it more functional style: `let is_glsl = if {... } else {..};`

kvark (commented):
Looks great overall, thank you for updating!
There is a few small things we can revisit before landing.

src/backend/gl/src/lib.rs
Gordon-F: Based on info in [this extension](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_sRGB_write_control.txt), `FRAMEBUFFER_SRGB` is always enabled with `OpenGL ES 3.0`. Added extra comment to code.

PR #3521, opened 2020-12-10
Sparse (tiled) resources for Vulkan and DirectX 12

deprilula28:
Progress in implementing sparse resources, listed on issue #1610

This PR adds the following:
- A `bind_sparse` method in CommandQueue based on `submit`.
- A Vulkan implementation for `bind_sparse`, making a [vkQueueBindSparse](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkQueueBindSparse.html) call.
- A DirectX12 implementation for `bind_sparse` for images, creating a virtual resource with [CreateReservedResource](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-createreservedresource) if unbound and mapping it to memory with [UpdateTileMappings](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12commandqueue-updatetilemappings).
- `SPARSE_BINDING`, `SPARSE_RESIDENCY` and `SPARSE_ALIASED` view capabilities when creating images. Also sets the texture layout to `D3D12_TEXTURE_LAYOUT_64KB_UNDEFINED_SWIZZLE` when creating images on DX12.
- A `sparselybound` feature on the `quad` example which binds the logo texture using `bind_sparse` instead of `bind_image_memory`.

Remaining additions:
- [ ] DirectX 11 implementation.
- [ ] Buffer and opaque textures in the DirectX 12 implementation.
- [ ] Setting sparse flags in buffer resources. (This will have to be a breaking change or a different buffer create method)

PR checklist:
- [x] `make` succeeds (on *nix)
- [x] `make reftests` succeeds
- [x] tested examples with the following backends: Vulkan, DirectX 12, DirectX 11

PR #3665, opened 2021-03-10
Refactoring & adding SPARSE_BINDING to QueueType

Neo-Zhixing:
We now have sparse binding thanks to #3521 but currently 'SPARSE_BINDING' is missing from 'QueueType'. This PR adds the 'SPARSE_BINDING' bit to QueueType by converting it from a enum to a bitflags struct.

Reference: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkQueueFlagBits.html

PR checklist:
- [x] tested examples with the following backends: Metal, DX12, DX11, Vulkan

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -1135,8 +1135,8 @@ impl window::PresentationSurface<Backend> for Surface {
 pub struct QueueFamily;
 
 impl queue::QueueFamily for QueueFamily {
-    fn queue_type(&self) -> queue::QueueType {
-        queue::QueueType::GRAPHICS | queue::QueueType::COMPUTE | queue::QueueType::TRANSFER
+    fn queue_flags(&self) -> queue::QueueFlags {
+        queue::QueueFlags::all()
     }
     fn max_queues(&self) -> usize {
         1
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -1136,7 +1136,7 @@ pub struct QueueFamily;
 
 impl queue::QueueFamily for QueueFamily {
     fn queue_type(&self) -> queue::QueueType {
-        queue::QueueType::General
+        queue::QueueType::GRAPHICS | queue::QueueType::COMPUTE | queue::QueueType::TRANSFER
     }
     fn max_queues(&self) -> usize {
         1
--- a/src/backend/empty/src/lib.rs
+++ b/src/backend/empty/src/lib.rs
@@ -574,8 +574,8 @@ impl device::Device<Backend> for Device {
 #[derive(Debug)]
 pub struct QueueFamily;
 impl queue::QueueFamily for QueueFamily {
-    fn queue_type(&self) -> queue::QueueType {
-        queue::QueueType::all()
+    fn queue_flags(&self) -> queue::QueueFlags {
+        queue::QueueFlags::all()
     }
     fn max_queues(&self) -> usize {
         1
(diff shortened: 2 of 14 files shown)


src/backend/empty/src/lib.rs
```diff
@@ -575,7 +575,7 @@ impl device::Device<Backend> for Device {
 pub struct QueueFamily;
 impl queue::QueueFamily for QueueFamily {
     fn queue_type(&self) -> queue::QueueType {
```
kvark, line 577: this type and function should no longer be called "queue type". It's a set of flags now, so `QueueFlags`?

src/hal/src/queue/mod.rs
```diff
@@ -19,38 +19,40 @@ use std::{any::Any, fmt};
...
+        const GRAPHICS = 0b1;
+        /// Queue supports compute operations
+        const COMPUTE = 0b10;
+        /// Queue supports transfer operations
+        const TRANSFER = 0b100;
+        /// Queue supports sparse resource memory management operations
+        const SPARSE_BINDING = 0b1000;
+
+        /// Queue supports both graphics operations and transfer operations
+        const GRAPHICS_TRANSFER = Self::GRAPHICS.bits | Self::TRANSFER.bits;
```
kvark, line 36: This shouldn't be needed. All graphics queues are required to support transfers.

src/hal/src/queue/mod.rs
```diff
@@ -19,38 +19,40 @@ use std::{any::Any, fmt};
...
+
+        /// Queue supports both compute operations and transfer operations
+        const COMPUTE_TRANSFER = Self::GRAPHICS.bits | Self::COMPUTE.bits;
```
kvark, line 39: same here, all compute queues are required to support transfers

kvark (commented):
Thank you for contributing!
Do we really need to make these to be flags? Alternatively, we could make `supports_sparse_binding` function to `trait QueueFamily`. That would be less intrusive.

src/backend/empty/src/lib.rs
Neo-Zhixing: name changed

Neo-Zhixing:
I agree that simply adding a `supports_sparse_binding` function could be easier. However this is a better match to the original Vulkan APIs. I'm not sure which one has a higher priority - API stability or matching Vulkan APIs?

src/backend/dx11/src/lib.rs
```diff
@@ -1135,8 +1135,8 @@ impl window::PresentationSurface<Backend> for Surface {
 pub struct QueueFamily;
 
 impl queue::QueueFamily for QueueFamily {
-    fn queue_type(&self) -> queue::QueueType {
-        queue::QueueType::General
+    fn queue_flags(&self) -> queue::QueueFlags {
```
kvark, line 1138: maybe just `fn flags(&self)`?

kvark (commented):
This is a tough choice, but also a relatively small piece of the API.
I'm good with proceeding. Let's just fix the CI (failures on Windows) and maybe rename `queue_flags` to just `flags()`?

Neo-Zhixing:
I've opened another PR (#3667) that simply added `supports_sparse_binding` function to `trait QueueFamily` as suggested. Please merge that PR and close this one if preferred.

kvark:
Thank you! This is very nice of you to provide the alternative. I'm not feeling strongly about this decision, but I merged #3667 as less of an intrusive solution.
Philosophically, we wanted to follow Vulkan were possible, so that we can get Vulkan Portability nicely, and it's easy to find answers about the guarantees by reading the Vulkan spec. In practice, both of these needs dim compared to the need to have `wgpu` to run efficiently on gfx, so some parts of gfx get slightly adjusted to make this path fast and efficient (like the swapchain API). So it's fine to have the queue types stay.

PR #3672, opened 2021-03-13
basic conservative rasterization support for Vulkan & DX12

### LINKED ISSUE #3637: [vk] Implement conservative rasterization
This issue is meant to track the Vulkan implementation of conservative rasterization. Conservative rasterization is currently supported in the DX12 backend.

Vulkan has support via an extension: [`VK_EXT_conservative_rasterization`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_conservative_rasterization.html) (GL does as well, as does a higher feature level of DX11)

Since support for the feature is not required by all (or any) backend, it may be worth adding it to either feature flags or the device capabilities. A single "supports conservative rasterization" flag would work, but the feature can also be broken up into "supports overestimation", "supports underestimation", "supports the `FullyCoveredEXT` shader builtin", etc. DX12 signals the presence of conservative rasterization support via [`D3D12_CONSERVATIVE_RASTERIZATION_TIER`](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_conservative_rasterization_tier)/[`D3D12_FEATURE_DATA_D3D12_OPTIONS`](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_feature_data_d3d12_options).

Wumpf:
Sorta fixes #3637 

Only provides the very basic option of enabling/disabling overestimation rasterization. This is almost everything DX12 has to offer anyways except the optional `SV_InnerCoverage` support, but leaves out even more Vulkan goodies (e.g. underestimation, various device properties, configurable extra-over-estimation, conservative raster of lines...).
Availability of the coverage variable is the most important missing thing but could later be exposed as an additional feature flag (debatable, since this is not how Vulkan does it, so more of a device property?)

Probably more controversial about this PR: `validate_creation_desc_support`
I didn't want to repeat feature check code over all devices so I created a function on the desc that validates against features (and made unallowed use of DEPTH_CLAMP an error since this was close by..). Imho this is the only sane way forward for stuff like this since more often than not there needs to be a check _somewhere_ but the philosophy of having this on this level and in that particular place might be questionable.

Tested this with wgpu - I already have wgpu and wgpu-rs branches ready with a sample that can swap back and forth between enabled and disabled conservative rasterization. Going to look like this:

image

PR checklist:
- [ ] `make reftests` succeeds -- failed in alloc, seems unrelated `thread 'main' panicked at 'removal index (is 0) should be < len (is 0)', library/alloc/src/vec.rs:1066:13`

### DIFF
--- a/src/backend/dx12/src/conv.rs
+++ b/src/backend/dx12/src/conv.rs
@@ -293,7 +293,6 @@ pub fn map_rasterizer(rasterizer: &pso::Rasterizer, multisample: bool) -> D3D12_
         ForcedSampleCount: 0,         // TODO: currently not supported
         AntialiasedLineEnable: FALSE, // TODO: currently not supported
         ConservativeRaster: if rasterizer.conservative {
-            // TODO: check support
             D3D12_CONSERVATIVE_RASTERIZATION_MODE_ON
         } else {
             D3D12_CONSERVATIVE_RASTERIZATION_MODE_OFF
--- a/src/backend/dx12/src/lib.rs
+++ b/src/backend/dx12/src/lib.rs
@@ -1310,6 +1310,14 @@ impl hal::Instance<Backend> for Instance {
                 tiled_resource_features |= Features::SPARSE_RESIDENCY_IMAGE_3D;
             }
 
+            let conservative_faster_features = if features.ConservativeRasterizationTier
+                == d3d12::D3D12_CONSERVATIVE_RASTERIZATION_TIER_NOT_SUPPORTED
+            {
+                Features::empty()
+            } else {
+                Features::CONSERVATIVE_RASTERIZATION
+            };
+
             let physical_device = PhysicalDevice {
                 library: Arc::clone(&self.library),
                 adapter,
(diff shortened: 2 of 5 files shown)


kvark:
Looks great @Wumpf , thank you for adding this!

I share the concern about `validate_creation_desc_support`. The problem is - how do we consistently draw the line between what we validate and what not? A loooot of things *can* be validated. In some places the cost is nothing, in another the cost can be visible. Right now the balance is - validate the minimum, i.e. just the feature set on device creation, that's pretty much it! Everything else is just "oops, unexpected stuff came", it's not really validation. Adding the check for conservative and depth clamping, while leaving all the other 100 pipeline states, seems like a wrong balance to me.

So that "check somewhere" should be in `wgpu`, which is safe. Would you mind removing the `validate_creation_desc_support` check from the PR?

Wumpf:
Alright, makes sense. Done!

kvark:
Bors r+

PR #3676, opened 2021-03-14
[gl] make dummy pbuffer optional

versusvoid:
My system has `GL_OES_surfaceless_context`, but unable to create pbuffer surfaces (It's mesa, so I doubt I'm the only one)

As I see it correct way would be to:
1) See if we have the extension
2) Try to choose config with `PBUFFER_BIT` if not
3) Notify user if both fails

But to check for extension we need to already have context and consequently a chosen config, so it's a bit recursive.
I simply try to bind newly created context and proceed to make pbuffer only if it fails.

### DIFF
--- a/src/backend/gl/src/window/egl.rs
+++ b/src/backend/gl/src/window/egl.rs
@@ -32,7 +32,7 @@ pub struct Inner {
     context: egl::Context,
     /// Dummy pbuffer (1x1).
     /// Required for `eglMakeCurrent` on platforms that doesn't supports `EGL_KHR_surfaceless_context`.
-    pbuffer: egl::Surface,
+    pbuffer: Option<egl::Surface>,
     wl_display: Option<*mut raw::c_void>,
 }
 
@@ -196,13 +196,20 @@ impl Inner {
             }
         };
 
-        let pbuffer = {
+        // Testing if context can be binded without surface
+        // and creating dummy pbuffer surface if not.
+        let pbuffer = if egl.upcast::<egl::EGL1_5>().is_none()
+            && !display_extensions.contains("EGL_KHR_surfaceless_context")
+        {
             let attributes = [egl::WIDTH, 1, egl::HEIGHT, 1, egl::NONE];
             egl.create_pbuffer_surface(display, config, &attributes)
+                .map(Some)
                 .map_err(|e| {
                     log::warn!("Error in create_pbuffer_surface: {:?}", e);
                     hal::UnsupportedBackend
                 })?
+        } else {
+            None
         };
 
         Ok(Self {
@@ -300,8 +307,8 @@ impl hal::Instance<crate::Backend> for Instance {
             .egl
             .make_current(
                 inner.display,
-                Some(inner.pbuffer),
-                Some(inner.pbuffer),
+                inner.pbuffer,
+                inner.pbuffer,
                 Some(inner.context),
             )
             .unwrap();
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/window/egl.rs
```diff
@@ -196,13 +196,19 @@ impl Inner {
...
         };
 
-        let pbuffer = {
-            let attributes = [egl::WIDTH, 1, egl::HEIGHT, 1, egl::NONE];
-            egl.create_pbuffer_surface(display, config, &attributes)
-                .map_err(|e| {
-                    log::warn!("Error in create_pbuffer_surface: {:?}", e);
-                    hal::UnsupportedBackend
-                })?
+        // Testing if context can be binded without surface
```
kvark, line 199: We have the `display_extensions` list, let's just check if the extension is present instead of trying to create it.

kvark (changes requested):
Thank you, this looks great!
I think we need one small change to land.

src/backend/gl/src/window/egl.rs
versusvoid: Well, you see, it's the problem I wrote about in description.
`display_extensions` are EGL extensions:

```
[2021-03-14T04:52:36Z INFO  gfx_backend_gl::window::egl] Display vendor "Mesa Project", 
version (1, 4), extensions: "EGL_ANDROID_blob_cache 
EGL_ANDROID_native_fence_sync 
EGL_EXT_buffer_age                       
EGL_EXT_create_context_robustness 
EGL_EXT_image_dma_buf_import 
EGL_EXT_image_dma_buf_import_modifiers 
EGL_EXT_swap_buffers_with_damage 
EGL_KHR_config_attribs 
EGL_KHR_create_context                               
EGL_KHR_create_context_no_error 
EGL_KHR_fence_sync 
EGL_KHR_get_all_proc_addresses 
EGL_KHR_gl_colorspace 
EGL_KHR_gl_renderbuffer_image 
EGL_KHR_gl_texture_2D_image 
EGL_KHR_gl_texture_3D_image                      
EGL_KHR_gl_texture_cubemap_image 
EGL_KHR_image_base 
EGL_KHR_no_config_context 
EGL_KHR_reusable_sync 
EGL_KHR_surfaceless_context 
EGL_KHR_swap_buffers_with_damage 
EGL_EXT_pixel_format_float 
EGL_KHR_wait_sync      
EGL_MESA_configless_context 
EGL_MESA_drm_image 
EGL_MESA_image_dma_buf_export 
EGL_MESA_query_driver 
EGL_WL_bind_wayland_display 
EGL_WL_create_wayland_buffer_from_image "
```

While GL/GLES extensions, containing `GL_OES_surfaceless_context` will only appear here:
https://github.com/gfx-rs/gfx/blob/1a111d7080d1beb433041f34a87c7f7473a57cc5/src/backend/gl/src/lib.rs#L369-L372

kvark: Sorry, you are right, "GL_OES_xx" is not going to show up there. However, "EGL_KHR_surfaceless_context" should be there, unless your context is really EGL 1.5.
So I think our detection should try `egl.upcast::<egl::EGL1_5>()` and check for `EGL_KHR_surfaceless_context` if this fails.

versusvoid: Oh, EGL also has surfaceless extension :disappointed: 
I've found one in glxinfo output and just assumed it's the one
Sure, I'll check for it

kvark (approved):
Beautiful, thank you!
bors r+

PR #3680, opened 2021-03-16
[dx12] reset command allocators periodically

### LINKED ISSUE #3505: Memory leak in the examples with DX12 backend
The `quad` example leaks memory heavily (10-20 MB/sec, depending on the framerate) if modified slightly to render continously. `wgpu-rs` examples are also leaking now that `wgpu-rs` is using `gfx` master (https://github.com/gfx-rs/wgpu/issues/1059).

The leak was introduced by https://github.com/gfx-rs/gfx/pull/3393

The amount of leaked memory doesn't seem to change if the window is resized, so it's probably not a texture leak. Maybe it's command buffer allocation related?

- GFX version: master starting from ee90ac192f97f98813e1cbd03f87f3f7cffec89c
- OS: Windows 10
- GPU: all

kvark:
Fixes #3505
PR checklist:

### DIFF
--- a/src/backend/dx12/Cargo.toml
+++ b/src/backend/dx12/Cargo.toml
@@ -30,6 +30,7 @@ log = "0.4"
 parking_lot = "0.11"
 smallvec = "1"
 spirv_cross = { version = "0.23", features = ["hlsl"] }
+thunderdome = "0.3"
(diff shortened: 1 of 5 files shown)


cwfitzgerald (approved):
LGTM!

PR #3685, opened 2021-03-17
[vk] Only enable sampler_filter_minmax and imageless_framebuffer on 1.2 when device supports it

### LINKED ISSUE #3677: samplerFilterMinmax enabled by default
Short info header:
- GFX version: gfx-backend-vulkan 0.7.0
- OS: Arch Linux
- GPU: GeForce GT 730
Gfx and wgpu-rs examples fails with "Unknown device creation error: ERROR_FEATURE_NOT_PRESENT".
Short investigation showed me that samplerFilterMinmax feature enabled by default [here](https://github.com/gfx-rs/gfx/blob/master/src/backend/vulkan/src/physical_device.rs#L168)
but for my GPU this feature is not supported.
> vulkaninfo cut:
VkPhysicalDeviceVulkan12Properties:
	driverID                                             = DRIVER_ID_NVIDIA_PROPRIETARY
	driverName                                           = NVIDIA
	driverInfo                                           = 460.56
	conformanceVersion                                   = 1.2.4.0
VkPhysicalDeviceVulkan12Features:
        ...
	runtimeDescriptorArray                             = true
	**samplerFilterMinmax                                = false**
	scalarBlockLayout                                  = true
        ...

tangmi:
Clients with a device that is running Vulkan 1.2 but does not support `sampler_filter_minmax` would crash on physical device creation. This can happen as well with `imageless_framebuffer` on e.g. Android devices.

This change makes it so the features are only enabled if they are available.

Fixes #3677

Note I can't effectively test this outside of the device simulator because I don't have real hardware in this configuration.

### DIFF
--- a/src/backend/vulkan/src/physical_device.rs
+++ b/src/backend/vulkan/src/physical_device.rs
@@ -64,6 +64,7 @@ impl PhysicalDeviceFeatures {
         api_version: Version,
         enabled_extensions: &[&'static CStr],
         requested_features: Features,
+        supports_vulkan12_sampler_filter_minmax: bool,
     ) -> PhysicalDeviceFeatures {
         // This must follow the "Valid Usage" requirements of [`VkDeviceCreateInfo`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkDeviceCreateInfo.html).
         let features = requested_features;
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!
bors r+

PR #3686, opened 2021-03-17
Add a texture format filter.

### LINKED ISSUE #3683: OpenGL backend lists all known formats as supported
This issue describes a problem with the OpenGL backend (`gfx-backend-gl`), in which inaccurate reporting of image format properties can easily lead to a crash when creating images in those formats.

## Rationale
The main way one would go about checking for valid image creation parameters in the HAL, for use in `Device::create_image`, is by calling `PhysicalDevice::image_format_properties` (Whose Vulkan equivalent is `vkGetPhysicalDeviceImageFormatProperties`). One calls this function with parameters very close to those of `Device::create_image` and gets in return an `Option<ImageFormatProperties>`, which encodes in itself both whether the given set of parameters is supported or not and the ranges for which image parameters in the given format are valid.

However, the implementation of  `PhysicalDevice::image_format_properties` in the OpenGL backend reports every format known to `gfx_backend_gl::conv::describe_format` as supported, which is definitely not the case for quite a number of those known formats in OpenGL ES. This leads to the use of formats like `Format::Bgra8Unorm`, crashing the application as a `GL_INVALID_OPERATION` is triggered by some drivers upon the backend calling `glTexImage2D`.

## How to reproduce

I came across this problem in the `Mesa 20.2.6` implementation of OpenGL ES 3.2, with the `Mesa Intel(R) UHD Graphics 620 (KBL GT2)` renderer. Keep in mind that this issue may not happen in more lenient implementations.

In order to verify and trigger this bug, one only needs to create an image through `Device::create_image` with parameter combinations not specified in [the OpenGL ES 3 documentation for `glTexImage2D`](https://www.khronos.org/registry/OpenGL-Refpages/es3/html/glTexImage2D.xhtml), that `PhysicalDevice::image_format_properties` indicates as supported.

The following log is from my program, which I believe is easy enough to understand in context.
```
[2021-03-16T20:52:47Z TRACE creeper::graphics] [12/768] Probing image with usage 0x00000004, format of Bgra8Unorm and tiling of Optimal
[2021-03-16T20:52:47Z TRACE creeper::graphics] Image format properties: FormatProperties { max_extent: Extent { width: 4294967295, height: 4294967295, depth: 4294967295 }, max_levels: 255, max_layers: 65535, sample_count_mask: 127, max_resource_size: 18446744073709551615 }
Mesa: User error: GL_INVALID_OPERATION in glTexImage2D(format = GL_BGRA, type = GL_UNSIGNED_BYTE, internalformat = GL_RGBA8)
[2021-03-16T20:52:47Z ERROR gfx_backend_gl] [API/Error] ID 1 : GL_INVALID_OPERATION in glTexImage2D(format = GL_BGRA, type = GL_UNSIGNED_BYTE, internalformat = GL_RGBA8)
thread 'main' panicked at 'Error creating image: InvalidOperation for kind D2(10, 10, 1, 1) of Bgra8Unorm', /home/mbr/.cargo/registry/src/github.com-1ecc6299db9ec823/gfx-backend-gl-0.7.1/src/device.rs:1548:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a601302ff0217b91589b5a7310a8a23adb843fdc/library/std/src/panicking.rs:495:5
   1: std::panicking::begin_panic_fmt
             at /rustc/a601302ff0217b91589b5a7310a8a23adb843fdc/library/std/src/panicking.rs:437:5
   2: <gfx_backend_gl::device::Device as gfx_hal::device::Device<gfx_backend_gl::Backend>>::create_image
             at /home/mbr/.cargo/registry/src/github.com-1ecc6299db9ec823/gfx-backend-gl-0.7.1/src/device.rs:1548:13
   3: creeper::graphics::ImageProfile::for_image
   4: creeper::graphics::Adapter::probe_images
   5: creeper::graphics::Adapter::collect
   6: creeper::graphics::Graphics::collect::{{closure}}
             at ./creeper/src/graphics.rs:29:19
   7: core::iter::adapters::map_try_fold::{{closure}}
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/mod.rs:912:28
   8: core::iter::traits::iterator::Iterator::try_fold
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:1888:21
   9: <core::iter::adapters::Map<I,F> as core::iter::traits::iterator::Iterator>::try_fold
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/mod.rs:938:9
  10: core::iter::traits::iterator::Iterator::find_map
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:2263:9
  11: <core::iter::adapters::FilterMap<I,F> as core::iter::traits::iterator::Iterator>::next
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/adapters/mod.rs:1245:9
  12: <alloc::vec::Vec<T> as alloc::vec::SpecFromIterNested<T,I>>::from_iter
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec.rs:2082:32
  13: <alloc::vec::Vec<T> as alloc::vec::SpecFromIter<T,I>>::from_iter
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec.rs:2226:20
  14: <alloc::vec::Vec<T> as core::iter::traits::collect::FromIterator<T>>::from_iter
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec.rs:1959:9
  15: core::iter::traits::iterator::Iterator::collect
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:1670:9
  16: creeper::graphics::Graphics::collect
             at ./creeper/src/graphics.rs:27:18
  17: creeper_dump::backends::unix::GraphicsBackends::collect
  18: creeper_dump::main
             at ./creeper/dump/src/main.rs:28:17
  19: core::ops::function::FnOnce::call_once
             at /home/mbr/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

## Proposed Solution

My proposed solution for this problem, which I plan to implement in a follow-up to this issue is to add a filter to `PhysicalDevice::image_format_properties` that removes input parameter combinations forbidden in the OpenGL ES specification, and to selectively lift some of these filters based on a combination of version and possibly relevant extensions.

Using version information to gate features is not a new thing to the OpenGL backend, as shown by functions such as `gfx_backend_gl::PhysicalDevice::new_adapter`. However, those are not as restrictive as a proper image format properties filter would be, which might pose a challenge when it comes to existing code targetting the HAL that relies on specific formats always being present.

DarkRyu550:
## General
This commit adds a structure called `info::TextureFormatFilter` and changes the
code in `info::query_all` so that, when detecting OpenGL ES it can select a
set of valid texture property combinations.

This provides `PhysicalDevice::image_format_properties` with a way of more
accurately filtering out formats that aren't valid for textures and whose
use would lead to a crash.

Fixes #3683
PR checklist:
1. OpenGL
2. Vulkan

## Remarks
One thing to keep in mind about this PR is that the current filter implementation overblocks formats. This is due to a limitation with the definition of `PhysicalDevice::image_format_properties`. Since the image creation function picks the underlying OpenGL function based on not just the format and usage, as `image_format_properties` assumes, but also based on the mipmap levels. There's no way for us to figure out for certain which function is gonna be picked for an image from `image_format_properties`. Thus, overblocking is the only viable solution I could think of that doesn't involve major changes to `image_format_properties`.

Other important thing to note is that this PR does not back port this change to `hal-0.7`, which I think is also important. I have already got that back port implemented in my own fork, but I'll leave it for another PR.

### DIFF
--- a/src/backend/gl/src/info.rs
+++ b/src/backend/gl/src/info.rs
@@ -358,6 +358,183 @@ impl Info {
     }
 }
 
+/// This structure checks whether a given image format is whitelisted to be used
+/// or not in the backend.
+///
+/// Unlike with Vulkan, OpenGL gives its users no easy way to query for valid
+/// image formats. Instead, it relies heavily on semantics, which are laid out
+/// in its specifications and function documentations.
+///
+/// This structure is intended to condense all of the supported formats into a
+/// single queryable location, acquired once when the adapter is created.
+#[derive(Debug)]
+pub struct TextureFormatFilter {
+    whitelist: HashSet<(u32, u32, u32)>
+}
+impl TextureFormatFilter {
+    /// This is the fixed, spec-defined list of all triplets in the form
+    /// `(Internal Format, Format, Type)` that are supported by OpenGL ES 3 and
+    /// WebGL for image creation via the `glTexImage` family of functions.
+    ///
+    /// Combinations of these parameters other than the ones in this table are
+    /// not supported by the core specifications. Though that does not mean they
+    /// can't be used or can't work under any circumstance. What it does mean is
+    /// that using anything outside this list is relying on permissive drivers
+    /// to not trigger undefined behavior.
+    ///
+    /// This is a one to one copy of the tables provided to us in
+    /// [the documentation of `glTexImage2D`], which is the same table as the
+    /// for `glTexImage3D`. The contents of the table are copied pretty much
+    /// verbatim in order to facilitate maintenance.
+    ///
+    /// [the documentation of `glTexImage2D`]: https://www.khronos.org/registry/OpenGL-Refpages/es3/html/glTexImage2D.xhtml
+    const ES3_TABLE: &'static [(u32, u32, u32)] = &[
+        /* Taken from Table 1. Unsized Internal Formats. */
+        (glow::RGB,             glow::RGB,             glow::UNSIGNED_BYTE),
+        (glow::RGB,             glow::RGB,             glow::UNSIGNED_SHORT_5_6_5),
+        (glow::RGBA,            glow::RGBA,            glow::UNSIGNED_BYTE),
+        (glow::RGBA,            glow::RGBA,            glow::UNSIGNED_SHORT_4_4_4_4),
+        (glow::RGBA,            glow::RGBA,            glow::UNSIGNED_SHORT_5_5_5_1),
+        (glow::LUMINANCE_ALPHA, glow::LUMINANCE_ALPHA, glow::UNSIGNED_BYTE),
+        (glow::LUMINANCE,       glow::LUMINANCE,       glow::UNSIGNED_BYTE),
+        (glow::ALPHA,           glow::ALPHA,           glow::UNSIGNED_BYTE),
+
+        /* Taken from Table 2. Sized Internal Formats. */
+        (glow::R8,             glow::RED,          glow::UNSIGNED_BYTE),
+        (glow::R8_SNORM,       glow::RED,          glow::BYTE),
+        (glow::R16F,           glow::RED,          glow::HALF_FLOAT),
+        (glow::R16F,           glow::RED,          glow::FLOAT),
+        (glow::R32F,           glow::RED,          glow::FLOAT),
+        (glow::R8UI,           glow::RED_INTEGER,  glow::UNSIGNED_BYTE),
+        (glow::R8I,            glow::RED_INTEGER,  glow::BYTE),
+        (glow::R16UI,          glow::RED_INTEGER,  glow::UNSIGNED_SHORT),
+        (glow::R16I,           glow::RED_INTEGER,  glow::SHORT),
+        (glow::R32UI,          glow::RED_INTEGER,  glow::UNSIGNED_INT),
+        (glow::R32I,           glow::RED_INTEGER,  glow::INT),
+        (glow::RG8,            glow::RG,           glow::UNSIGNED_BYTE),
+        (glow::RG8_SNORM,      glow::RG,           glow::BYTE),
+        (glow::RG16F,          glow::RG,           glow::HALF_FLOAT),
+        (glow::RG16F,          glow::RG,           glow::FLOAT),
+        (glow::RG32F,          glow::RG,           glow::FLOAT),
+        (glow::RG8UI,          glow::RG_INTEGER,   glow::UNSIGNED_BYTE),
+        (glow::RG8I,           glow::RG_INTEGER,   glow::BYTE),
+        (glow::RG16UI,         glow::RG_INTEGER,   glow::UNSIGNED_SHORT),
+        (glow::RG16I,          glow::RG_INTEGER,   glow::SHORT),
+        (glow::RG32UI,         glow::RG_INTEGER,   glow::UNSIGNED_INT),
+        (glow::RG32I,          glow::RG_INTEGER,   glow::INT),
+        (glow::RGB8,           glow::RGB,          glow::UNSIGNED_BYTE),
+        (glow::SRGB8,          glow::RGB,          glow::UNSIGNED_BYTE),
+        (glow::RGB565,         glow::RGB,          glow::UNSIGNED_BYTE),
+        (glow::RGB565,         glow::RGB,          glow::UNSIGNED_SHORT_5_6_5),
+        (glow::RGB8_SNORM,     glow::RGB,          glow::BYTE),
+        (glow::R11F_G11F_B10F, glow::RGB,          glow::UNSIGNED_INT_10F_11F_11F_REV),
+        (glow::R11F_G11F_B10F, glow::RGB,          glow::HALF_FLOAT),
+        (glow::R11F_G11F_B10F, glow::RGB,          glow::FLOAT),
+        (glow::RGB9_E5,        glow::RGB,          glow::UNSIGNED_INT_5_9_9_9_REV),
+        (glow::RGB9_E5,        glow::RGB,          glow::HALF_FLOAT),
+        (glow::RGB9_E5,        glow::RGB,          glow::FLOAT),
+        (glow::RGB16F,         glow::RGB,          glow::HALF_FLOAT),
+        (glow::RGB16F,         glow::RGB,          glow::FLOAT),
+        (glow::RGB32F,         glow::RGB,          glow::FLOAT),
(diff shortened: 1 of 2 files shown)


kvark:
> Other important thing to note is that this PR does not back port this change to hal-0.7, which I think is also important. I have already got that back port implemented in my own fork, but I'll leave it for another PR.

probably unnecessary at this point. GL is rough right now, it's fine to expect GL adopters to run on latest master.

src/backend/gl/src/info.rs
```diff
@@ -358,6 +358,183 @@ impl Info {
...
+        (glow::RGBA,            glow::RGBA,            glow::UNSIGNED_SHORT_5_5_5_1),
+        (glow::LUMINANCE,       glow::LUMINANCE,       glow::UNSIGNED_BYTE),
+        (glow::LUMINANCE_ALPHA, glow::LUMINANCE_ALPHA, glow::UNSIGNED_BYTE)
+    ];
+
+    /// Creates a new filter for OpenGL ES 3.
+    fn new_es3() -> Self {
+        Self {
+            whitelist: {
+                let mut whitelist = HashSet::<(u32, u32, u32)>::default();
```
kvark, line 499: can also do `Self::ES3_TABLE.iter().cloned().collect()`

src/backend/gl/src/lib.rs
```diff
@@ -607,7 +610,22 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+            ..
+        } = conv::describe_format(format)?;
+
+        if !self.0.texture_format_filter.check(
+            tex_internal,
+            tex_external,
+            data_type) {
+
+            /* This format is not supported. */
+            return None
```
kvark, line 626: nit: lets not have an early return here, since the rest of the code doesn't have any logic. The if/else should just result in an expression

kvark (approved):
Looks good, thank you for contributing!
I have a few small suggestions, but they are minor.
We can proceed.
bors r+

src/backend/gl/src/info.rs
```diff
@@ -566,7 +744,20 @@ pub(crate) fn query_all(
...
+        /* Use the OpenGL ES 3 format filter. */
+        TextureFormatFilter::new_es3()
+    } else if info.is_supported(&[Es(1, 0)]) {
+        /* Use the OpenGL ES 1 and OpenGL ES 2 format filter. */
+        TextureFormatFilter::new_es1_es2()
+    } else {
+        /* We're using the core specification. We can assume all of the
+         * combinations are valid, provided the OpenGL enums values are also
+         * valid for textures. */
+        TextureFormatFilter::new()
```
kvark, line 757: this is an empty list, isn't it? So for the core spec we'd end up not exposing any formats at all, which is not the intent here.

kvark (changes requested):

src/backend/gl/src/info.rs
DarkRyu550: Yeah, I got that wrong

DarkRyu550: Corrected it in [5c88051](https://github.com/gfx-rs/gfx/pull/3686/commits/5c880518d7acbb6c8b2acb726fdba46ec3315b72).

Gordon-F:
Maybe we can reuse `Requirement` code here?

For example:
```rust
pub fn describe_format(format: Format) -> Option<FormatDescription> {
    use crate::native::VertexAttribFunction::*;
    use hal::format::Format::*;
    let _ = Double; //mark as used

    // TODO: Add more formats and error handling for `None`
    let format_desc = match format {
        R8Uint => FormatDescription::new(
            glow::R8UI,
            glow::RED_INTEGER,
            glow::UNSIGNED_BYTE,
            1,
            Integer,
            &[Es(1, 0)] // If None - Supports everywhere
        ),
        _ => return None,
    };

    if let Some(requirements) = format_desc.requirements {
        if crate::is_supported(requirements) {
            format_desc
        }
    }

    None
}
```

It also should save from code duplication problem (`FormatDescription` table and `TextureFormatFilter` table).

DarkRyu550:
> 
> It also should save from code duplication problem (`FormatDescription` table and `TextureFormatFilter` table).
>

I'd originally thought about implementing the filter in that way, but ultimately decided against it for two main reasons: maintainability and modularity. 

First, for maintainability. As far as I know, the only way to know whether or not a given set of texture parameters is supported or not in OpenGL ES is to take it right from the documentation. This goes for both deriving this list for the first time and for keeping it up to date. That list is, as of OpenGL ES 3.2, 76 entries long, and probably bound to grow in the future. 

So, if you were to implement this in the filter you'd together with the logic in `describe_format`, you'd have to add at least 76 extra branches and keep those up to date with the table in the documentation. Having it as a separate table of `(u32, u32, u32)` lets us keep a structure that is effectively identical to that of the original table, which should take much less effort to verify and update, whenever needed. Then, having a separate type built around it leaves the way these tables are actually specified as an implementation detal.

Second, for modularity. Because I planned for the whitelist to be expanded based on available extensions, it would only make the branching in `describe_format` more complicated if we were to also check for it there, rather than in a separate, composable set.

kvark:
@Gordon-F does your concern still stand?

kvark:
Ok, we can iterate on it. It's good to land at least for addressing the problem we have now.
bors r+

PR #3687, opened 2021-03-17
Add pso ShaderCreationError

Gordon-F:
As discussed in the matrix. Add new pso creation error - ShaderCreationError.

PR checklist:
- [x] tested examples with the following backends: GL (CI should check all other errors)

### DIFF
--- a/src/hal/src/pso/mod.rs
+++ b/src/hal/src/pso/mod.rs
@@ -21,6 +21,9 @@ pub enum CreationError {
     /// Unknown other error.
     #[error("Implementation specific error occurred")]
     Other,
+    /// Shader module creation error.
+    #[error("{0:?} shader creation failed: {1:}")]
+    ShaderCreationError(ShaderStageFlags, String),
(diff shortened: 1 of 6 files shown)


src/hal/src/pso/mod.rs
```diff
@@ -21,6 +21,9 @@ pub enum CreationError {
     /// Unknown other error.
     #[error("Implementation specific error occurred")]
     Other,
+    /// Shader module creation error.
+    #[error("Shader creation failed: {0:}")]
+    ShaderCreationError(String),
```
kvark, line 26: I think we always know the shader stage when we issue this error, don't we? In this case, we should make it `ShaderCreationError(Stage, String)`. Would you be open to try it out?

kvark (commented):
Thank you for doing this! I have one concern to address

Gordon-F:
CI is finally happy. Should be ready to merge.

kvark (approved):
superb, thank you for addressing this!
bors r+

PR #3689, opened 2021-03-20
descriptive assertions in push_graphics_constants

hclarke:
push_graphics_constants can fail if the stages don't match the pipeline layout

i changed some `unwrap`s to `expect`s to make it a bit easier to see what's going wrong


kvark (approved):
bors r+

PR #3690, opened 2021-03-20
Fix incorrect gl texture internal formats

dlunch:
Fixes some incorrect internal texture format. [glTexImage2D](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glTexImage2D.xhtml) [Ext BGRA888](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_format_BGRA8888.txt)

PR checklist:

### DIFF
--- a/src/backend/gl/src/conv.rs
+++ b/src/backend/gl/src/conv.rs
@@ -132,9 +132,7 @@ pub fn describe_format(format: Format) -> Option<FormatDescription> {
             4,
             Float,
         ),
-        Bgra8Unorm => {
-            FormatDescription::new(glow::RGBA8, glow::BGRA, glow::UNSIGNED_BYTE, 4, Float)
-        }
+        Bgra8Unorm => FormatDescription::new(glow::BGRA, glow::BGRA, glow::UNSIGNED_BYTE, 4, Float),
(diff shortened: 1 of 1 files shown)


src/backend/gl/src/conv.rs
```diff
@@ -132,9 +132,7 @@ pub fn describe_format(format: Format) -> Option<FormatDescription> {
             4,
             Float,
         ),
-        Bgra8Unorm => {
-            FormatDescription::new(glow::RGBA8, glow::BGRA, glow::UNSIGNED_BYTE, 4, Float)
-        }
+        Bgra8Unorm => FormatDescription::new(glow::BGRA, glow::BGRA, glow::UNSIGNED_BYTE, 4, Float),
```
kvark, line 135: looks like `Bgra8Srgb` is still a problem

kvark (approved):
Thank you!
bors r+

src/backend/gl/src/conv.rs
dlunch: Yes, but I'm not sure which internal format is correct for Bgra8Srgb.

PR #3691, opened 2021-03-21
[Vulkan] Fixed vulkan present optimality

Neo-Zhixing:
The vulkan implementation of Queue::present mixed up the return value of ash's queue_present.

queue_present "returns whether the swapchain is suboptimal for the surface". Therefore when it is true, we should return Some(Suboptimal) and when it's false we return None.


kvark (approved):
Oh wow, this is quite counter-intuitive.
Thank you for fixing!
bors r+

PR #3692, opened 2021-03-23
[vk] Don't advertise `format::ImageFeature::SAMPLED_MINMAX` if the device does not support it.

tangmi:
Follow-up to #3677
PR checklist:


kvark (approved):
Thank you!
bors r+

PR #3693, opened 2021-03-23
[GL] Fix EGL context creation and set setBuffersGeometry for Android

Gordon-F:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -30,7 +30,7 @@ parking_lot = "0.11"
 raw-window-handle = "0.3"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-egl = { package = "khronos-egl", version = "4", features = ["dynamic"] }
+egl = { package = "khronos-egl", version = "4.1", features = ["dynamic"] }
 libloading = "0.7"
 
 [dependencies.auxil]
--- a/src/backend/gl/src/window/egl.rs
+++ b/src/backend/gl/src/window/egl.rs
@@ -67,6 +67,16 @@ type WlEglWindowResizeFun = unsafe extern "system" fn(
 
 type WlEglWindowDestroyFun = unsafe extern "system" fn(window: *const raw::c_void);
 
+#[cfg(target_os = "android")]
+extern "C" {
+    pub fn ANativeWindow_setBuffersGeometry(
+        window: *mut raw::c_void,
+        width: i32,
+        height: i32,
+        format: i32,
+    ) -> i32;
+}
+
 fn open_x_display() -> Option<(ptr::NonNull<raw::c_void>, libloading::Library)> {
     log::info!("Loading X11 library to get the current display");
     unsafe {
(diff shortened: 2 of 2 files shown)


src/backend/gl/Cargo.toml
```diff
@@ -30,7 +30,7 @@ parking_lot = "0.11"
 raw-window-handle = "0.3"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-egl = { package = "khronos-egl", version = "4", features = ["dynamic"] }
+egl = { package = "khronos-egl", version = "4.1", features = ["dynamic"] }
```
Gordon-F, line 33: Includes fix https://github.com/timothee-haudebourg/khronos-egl/pull/19

Without this patch we can't use `gfx-backend-gl` on Android devices.

src/backend/gl/src/window/egl.rs
```diff
@@ -198,9 +208,8 @@ impl Inner {
 
         // Testing if context can be binded without surface
         // and creating dummy pbuffer surface if not.
-        let pbuffer = if egl.upcast::<egl::EGL1_5>().is_none()
-            && !display_extensions.contains("EGL_KHR_surfaceless_context")
-        {
+        // Some EGL 1.5 devices does not support EGL_KHR_surfaceless_context (Android Emulator, for example)
```
kvark, line 211: are you sure? The extension is a [part of the 1.5 core](https://docs.imgtec.com/Architecture_Guides/Supported_Extensions_OpenGL_ES_EGL/topics/d0e1207.html):
> This functionality is core to EGL 1.5, so the extension is no longer needed.

kvark: I'm worried about the opposite situation: we'd be running on an EGL-1.5 device that simply doesn't expose the extension (because it's in core).

kvark (commented):
Just one concern, looks great otherwise!

src/backend/gl/src/window/egl.rs
Gordon-F: Nice catch, thank you! We need a runtime version check here.

Android Emulator running on EGL 1.4, but 1.5 are also available on the system.

kvark (approved):
Thank you!
bors r+

src/backend/gl/src/window/egl.rs
kvark: I'm a bit confused. So it happily gives you an 1.5 context even if you initialized 1.4?
That seems like a serious footgun. Perhaps, we could raise this with khronos-egl crate author in an issue?
Could you file it, please?

Gordon-F: > So it happily gives you an 1.5 context even if you initialized 1.4?

No, it gives me 1.4 context, but previous check `egl.upcast::<egl::EGL1_5>()` is true, because we have EGL 1.5 in Android 10+ systems.

kvark: why are we not using 1.5 if it's available?

Gordon-F: `eglInitialize` return 1.4 version.

PR #3695, opened 2021-03-24
Update naga to gfx-18

Gordon-F:
When trying to update my dev environment to the latest `naga`, I saw that api changes a little bit. This PR update to latest available naga (a0b5717fed0bd1c8536fb2204f308d6db39d295f)

Hope this will be helpful 😄 

PR checklist:

### DIFF
--- a/src/backend/gl/Cargo.toml
+++ b/src/backend/gl/Cargo.toml
@@ -47,7 +47,7 @@ optional = true
 
 [dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
-tag = "gfx-17"
+tag = "gfx-18"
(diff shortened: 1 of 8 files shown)


kvark:
Thank you!
Let's release a naga tag after https://github.com/gfx-rs/naga/pull/608 merges in.
For your code, I think we need to disable all validation in gfx's use of Naga. gfx-rs is meant to be totally unsafe, so there is no need to do extra work at this stage. In all the places where you use `ValidationFlags`, it should be `ValidationFlags::empty()`.

kvark:
@Gordon-F gfx-18 tag is published now, please use it

kvark (approved):

PR #3699, opened 2021-03-25
[WebGL] Return only RGBA swapchain formats

Gordon-F:
WebGL doesn't supports BGRA texture formats

PR checklist:
- [x] tested examples with the following backends: WebGL


kvark (approved):
Looks reasonable, thank you!
bors r+

PR #3708, opened 2021-04-01
added capture

adamnemecek:
Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -2134,7 +2134,7 @@ impl d::Device<B> for Device {
         unimplemented!()
     }
 
-   fn stop_capture(&self) {
+    fn stop_capture(&self) {
         unimplemented!()
     }
 }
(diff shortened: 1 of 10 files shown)


src/backend/gl/src/device.rs
```diff
@@ -2129,4 +2129,12 @@ impl d::Device<B> for Device {
     ) {
         // TODO
     }
+
+    fn start_capture(&self) {
+        unimplemented!()
+    }
+
+   fn stop_capture(&self) {
```
kvark, line 2137: please run `cargo fmt`

src/backend/metal/src/device.rs
```diff
@@ -150,11 +150,8 @@ impl Drop for Device {
     fn drop(&mut self) {
         if cfg!(feature = "auto-capture") {
```
kvark, line 151: let's remove the `auto-capture` feature entirely now, please.
If the user wants it, they can control it now

kvark (commented):
Great start!
Just a few things before we can land

kvark (approved):
Great, thank you!

PR #3713, opened 2021-04-04
avoid panic & error propagation on DSV creation

### LINKED ISSUE #3711: Unimplemented path in Dx11 backend when constructing cube views of depth textures
Short info header:
- GFX version: d002952
- OS: Windows 10
- GPU: AMD Radeon(TM) Vega 3 Graphics

```
PanicInfo: panicked at 'not implemented', D:\veloren\gfx\src\backend\dx11\src\device.rs:771:18
```

I believe this is similar to the issue that was fixed in dx12: https://github.com/gfx-rs/gfx/issues/3702

I'm currently attempting to mimic the patch used there.

Imberflur:
Fixes #3711 

Tries to mimic the fix for the Dx12 backend. This works for me, but I'm not sure how ideal this is because it pretty much ignores any error from `view_image_as_depth_stencil` when creating the image view, maybe the ignored errors should be logged?

PR checklist:

### DIFF
--- a/src/backend/dx11/src/device.rs
+++ b/src/backend/dx11/src/device.rs
@@ -768,7 +768,13 @@ impl Device {
                     }
                 }
             }
-            _ => unimplemented!(),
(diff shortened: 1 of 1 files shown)


kvark (approved):
Thank you!
bors r+

PR #3715, opened 2021-04-05
Fix buffer descriptor indexing on vulkan1.2

csnewman:
Fixes descriptor indexing not being enabled on vulkan1.2

PR checklist:


kvark (approved):
Nice, thank you!
bors r+

PR #3717, opened 2021-04-06
[vk] only free non-zero number of commmand buffers or descriptors

kvark:
Fixes the validation error in https://github.com/gfx-rs/wgpu-rs/issues/837#issuecomment-813790158


kvark:
bors retry

Issue #3716, opened 2021-04-05
Pipeline cache serialization/deserialization investigation


expenses:
## Goal

It would be neat and useful to have an implementation of [`get_pipeline_cache_data`](https://github.com/gfx-rs/gfx/blob/2a93d52661aafcbd6441ea83e739c8ced906cd21/src/hal/src/device.rs#L281) on all modern platforms (Vulkan, DX12, Metal). Along with the corresponding code in [`create_pipeline_cache`](https://github.com/gfx-rs/gfx/blob/2a93d52661aafcbd6441ea83e739c8ced906cd21/src/hal/src/device.rs#L277), this would allow for being able to cache the pipelines to disk on all backends, giving a good performance boost when a lot of pipelines are used.

### Vulkan

The Vulkan API has this `get_pipeline_cache_data` function [built in](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPipelineCacheData.html).

### Metal

The Metal backend has a pipeline cache:
https://github.com/gfx-rs/gfx/blob/2a93d52661aafcbd6441ea83e739c8ced906cd21/src/backend/metal/src/native.rs#L207-L211

However there is no way to serialize or deserialize it at present.

The key blocker for this is that the `ModuleInfo` struct stores a `metal::Library`:

https://github.com/gfx-rs/gfx/blob/2a93d52661aafcbd6441ea83e739c8ced906cd21/src/backend/metal/src/native.rs#L200-L205

While there is no way in the Metal API to serialize a [`MTLLibrary`](https://developer.apple.com/documentation/metal/mtllibrary) (the underlying type), there is a [`serialize`](https://developer.apple.com/documentation/metal/mtldynamiclibrary/3553982-serialize) function for [`MTLDynamicLibrary`](https://developer.apple.com/documentation/metal/mtldynamiclibrary) which I believe we could convert into. It serializes directly into a file though, which is pretty gross. Presumably we'd then have to read back from this file.

### DX12

The DirectX 12 backend doesn't have a pipeline cache. However, there is an issue that lays out how one could be created: https://github.com/gfx-rs/gfx/issues/2877, similar to what the Metal backend does.

kvark:
Thank you for filing this!
About the Metal backend, the pipeline caching path is the old stuff we use with SPIRV-Cross.
I tried to adjust it for Naga, but it wasn't easy. So for all the purposes, consider there *not* to be an implementation on Metal right now (since Naga is the future).

[...]

PR #3719, opened 2021-04-07
Use Binary Archives as a pipeline cache on the metal backend

expenses:
See #3716. This PR builds on-top of https://github.com/gfx-rs/metal-rs/pull/207 to add an implementation of `get_pipeline_cache_data` for the metal backend. This PR is not ready for reviews.

Todo:

- [x] Get rid of the Mutex around the pipeline cache and replace `is_empty` with an `AtomicBool`, as `unsafe impl Sync` is probably fine for the binary archive.
- [x] Check the availability of binary archives on the target with `device.supports_binary_archive()`.
- [x] Potentially use a second cache to store SPIR-V -> MSL transformations.
- [x] Come up with a test suite to test pipeline caches. Something with hundreds/thousands of pipelines (maybe using shaders from shadertoy?) would be nice.

PR checklist:
- [x] tested examples with the following backends: `metal`

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -15,7 +15,8 @@ edition = "2018"
 [features]
 default = []
 signpost = []
-cross = ["spirv_cross", "auxil", "naga/spv-out"]
+cross = ["spirv_cross", "auxil", "naga/spv-out", "serde-cache"]
+serde-cache = ["serde", "naga/serialize", "naga/deserialize", "bincode"]
 
 [lib]
 name = "gfx_backend_metal"
@@ -38,6 +39,8 @@ parking_lot = "0.11"
 storage-map = "0.3"
 raw-window-handle = "0.3"
 tempfile = "3.2"
+serde = { version = "1", features = ["serde_derive"], optional = true }
+bincode = { version = "1", optional = true }
 
 [dependencies.auxil]
 package = "gfx-auxil"
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -15,8 +15,7 @@ edition = "2018"
 [features]
 default = []
 signpost = []
-cross = ["spirv_cross", "auxil", "naga/spv-out", "serde-cache"]
-serde-cache = ["naga/serialize", "naga/deserialize"]
+cross = ["spirv_cross", "auxil"]
 
 [lib]
 name = "gfx_backend_metal"
@@ -56,7 +55,7 @@ optional = true
 
 [dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
-features = ["spv-in", "msl-out"]
+features = ["spv-in", "msl-out", "spv-out", "serialize", "deserialize"]
 
 # This forces docs.rs to build the crate on mac, otherwise the build fails
 # and we get no docs at all.
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2018"
 default = []
 signpost = []
 cross = ["spirv_cross", "auxil", "naga/spv-out", "serde-cache"]
-serde-cache = ["serde", "naga/serialize", "naga/deserialize", "bincode", "spirv_cross/serialize"]
+serde-cache = ["naga/serialize", "naga/deserialize", "spirv_cross/serialize"]
 
 [lib]
 name = "gfx_backend_metal"
@@ -30,7 +30,7 @@ copyless = "0.1.4"
 fxhash = "0.2.1"
 log = { version = "0.4" }
 dispatch = { version = "0.2", optional = true }
-metal = { git = "https://github.com/expenses/metal-rs", branch="binary-archive", features = ["private"] }
+metal = { git = "https://github.com/gfx-rs/metal-rs", rev="78f632d194c7c16d18b71d7373c4080847d110b0", features = ["private"] }
 foreign-types = "0.3"
 objc = "0.2.5"
 block = "0.1"
@@ -39,8 +39,8 @@ parking_lot = "0.11"
 storage-map = "0.3"
 raw-window-handle = "0.3"
 tempfile = "3.2"
-serde = { version = "1", features = ["serde_derive"], optional = true }
-bincode = { version = "1", optional = true }
+serde = { version = "1", features = ["serde_derive"] }
+bincode = "1"
 
 [dependencies.auxil]
 package = "gfx-auxil"
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -57,7 +57,7 @@ optional = true
 
 [dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
-# tag = "gfx-21"
+tag = "gfx-22"
 features = ["spv-in", "msl-out"]
 
 # This forces docs.rs to build the crate on mac, otherwise the build fails
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -56,7 +56,7 @@ optional = true
 
 [dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
-tag = "gfx-21"
+# tag = "gfx-21"
 features = ["spv-in", "msl-out"]
 
 # This forces docs.rs to build the crate on mac, otherwise the build fails
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2018"
 default = []
 signpost = []
 cross = ["spirv_cross", "auxil", "naga/spv-out", "serde-cache"]
-serde-cache = ["serde", "naga/serialize", "naga/deserialize", "bincode"]
+serde-cache = ["serde", "naga/serialize", "naga/deserialize", "bincode", "spirv_cross/serialize"]
 
 [lib]
 name = "gfx_backend_metal"
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -15,7 +15,8 @@ edition = "2018"
 [features]
 default = []
 signpost = []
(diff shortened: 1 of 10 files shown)


expenses:
I've added most of the features that I want for this now; what's left is a bit of a clean up, optimization as far as serialization is concerned and to merge https://github.com/gfx-rs/naga/pull/686 and https://github.com/grovesNL/spirv_cross/compare/master...expenses:serialize-feature?expand=1.

expenses:
Paused for a little bit because I'd really like the SPV->MSL cache to be done for naga, but testing this requires naga compilation of the shadertoy shaders to work a little better. https://github.com/gfx-rs/naga/pull/693 fixes most of the issues.

kvark:
This is also going to help gfx-portability. In particular, as we are finishing up the Naga fixes needed to get Dota2 running, we'll be benchmarking again, and having the working shader cache will boost it significantly.

src/backend/metal/src/device.rs
```diff
@@ -1386,69 +1383,117 @@ impl hal::device::Device<Backend> for Device {
...
+            let bytes = std::fs::read(&temp_path).unwrap();
+
+            Ok(bytes)
+        };
+
+        Ok(bincode::serialize(&n::SerializablePipelineCache {
+            binary_archive: &binary_archive()?,
+            spv_to_msl: n::serialize_spv_to_msl_cache(&cache.spv_to_msl),
+        })
+        .unwrap())
```
expenses, line 1481: This could be made faster by not using `bincode` and making it zero-copy.

src/backend/metal/src/device.rs
```diff
@@ -1386,69 +1383,117 @@ impl hal::device::Device<Backend> for Device {
...
+            }
+        };
+
+        if let Some(data) = data.filter(|data| !data.is_empty()) {
+            let pipeline_cache: n::SerializablePipelineCache = bincode::deserialize(data).unwrap();
+
+            Ok(n::PipelineCache {
+                binary_archive: create_binary_archive(&pipeline_cache.binary_archive)?,
+                spv_to_msl: n::load_spv_to_msl_cache(pipeline_cache.spv_to_msl),
+            })
```
expenses, line 1437: This could be made faster by not using `bincode` and making it zero-copy.

expenses:
I think this is about ready for review now. https://github.com/gfx-rs/naga/pull/716 needs to be merged and a `gfx-22` tag of naga needs to be made before merging.

expenses:
We're still parsing and validating the SPIR-V code with naga unnecessarily, as we should be able to skip over that if the SPIR-V is in the cache. Parsing and validating only seems to take about 40ms for 8000 modules though, so I don't think this is a huge problem.

expenses:
> Parsing and validating only seems to take about 40ms for 8000 modules though, so I don't think this is a huge problem.

Writing out the MSL shaders with a fresh pipeline cache also only takes 50ms for 8000 modules, so it appears that naga is fast enough that there's not _too_ much benefit to having a SPV->MSL cache in the first place.

expenses:
> Writing out the MSL shaders with a fresh pipeline cache also only takes 50ms for 8000 modules, so it appears that naga is fast enough that there's not _too_ much benefit to having a SPV->MSL cache in the first place.

Here are some timings:
```
cleared spv->msl cache: 8.50 5.87 5.44 5.46 6.42
hot spv->msl cache: 5.48 4.89 4.78 4.78 4.64
```

src/backend/metal/src/device.rs
```diff
@@ -1386,69 +1383,117 @@ impl hal::device::Device<Backend> for Device {
...
+        data: Option<&[u8]>,
     ) -> Result<n::PipelineCache, d::OutOfMemory> {
-        Ok(n::PipelineCache {
-            #[cfg(feature = "cross")]
-            modules: FastStorageMap::default(),
-        })
+        let device = self.shared.device.lock();
+
+        let create_binary_archive = |data: &[u8]| {
+            let supports_binary_archive = device.supports_family(metal::MTLGPUFamily::Apple3)
```
kvark, line 1391: let's have this as a field in `PrivateCapabilities`

src/backend/metal/src/device.rs
```diff
@@ -1386,69 +1383,117 @@ impl hal::device::Device<Backend> for Device {
...
     unsafe fn get_pipeline_cache_data(
         &self,
-        _cache: &n::PipelineCache,
+        cache: &n::PipelineCache,
     ) -> Result<Vec<u8>, d::OutOfMemory> {
-        //empty
-        Ok(Vec::new())
+        let binary_archive = || {
+            let binary_archive = match &cache.binary_archive {
+                Some(binary_archive) => binary_archive,
```
kvark, line 1452: our clippy is configured to not like matching on references. Instead, let's be explicit here:
```rust
match cache.binary_archive {
  Some(ref archive) => archive,
   ...
}
```

src/backend/metal/src/device.rs
```diff
@@ -1769,7 +1820,20 @@ impl hal::device::Device<Backend> for Device {
             .map_err(|err| {
                 error!("PSO creation failed: {}", err);
                 pso::CreationError::Other
-            })
+            })?;
+
+        // We need to add the pipline descriptor to the binary archive after creating the
+        // pipeline, otherwise `new_render_pipeline_state_with_fail_on_binary_archive_miss`
+        // succeeds when it shouldn't.
+        if let Some(binary_archive) = cache.and_then(|cache| cache.binary_archive.as_ref()) {
```
kvark, line 1828: looks like this `cache.and_then(|cache| cache.binary_archive.as_ref())` could use a helper method

src/backend/metal/src/native.rs
```diff
@@ -204,10 +204,65 @@ pub struct ModuleInfo {
...
+
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub struct SerializableModuleInfo {
+    pub source: String,
+    pub entry_point_map: EntryPointMap,
+    pub rasterization_enabled: bool,
+}
+
+pub(crate) type SpvToMsl = FastStorageMap<
+    (
```
kvark, line 223: let's have a proper structure for this tuple

kvark (commented):
This is some top quality work, thank you for kicking this forward!

expenses:
I timed it again and having a binary archive pipeline cache definitely makes a difference when the system cache has been cleaned. 
```
wiped system cache, with binary archive: 293.24 288.50
wiped system cache, no binary archive: 892.77 889.07
```

src/backend/metal/Cargo.toml
```diff
@@ -15,7 +15,7 @@ edition = "2018"
 [features]
 default = []
 signpost = []
-cross = ["spirv_cross", "auxil", "naga/spv-out"]
+cross = ["spirv_cross", "auxil"]
```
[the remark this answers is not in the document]
expenses: Because we're now enabling it by default for `create_shader_module_from_naga`.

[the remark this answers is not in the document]
expenses: I've changed things around now and we only use `naga/spv-out` if we're using the `pipeline-cache` feature or the `cross` feature.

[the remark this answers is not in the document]
kvark: so for pipeline caching, we basically have to always go through SPIR-V? It's not the worst thing, I imagine wgpu will just not have this feature enabled for the near future.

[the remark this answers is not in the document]
expenses: Not necessarily, I could probably look into reworking things so we cache the hash of the WSL as well.

src/backend/metal/src/device.rs
```diff
@@ -1195,8 +1207,8 @@ impl hal::device::Device<Backend> for Device {
                                     .contains(n::DescriptorContent::IMMUTABLE_SAMPLER)
                                 {
                                     let immutable_sampler = &immutable_samplers[&layout.binding];
-                                    let handle =
-                                        inline_samplers.append(immutable_sampler.data.clone());
+                                    let handle = inline_samplers.len() as u8;
```
kvark, line 1210: nit: here and in the push constants, we should ideally use the types, like `naga::back::msl::InlineSamplerIndex` and `naga::back::msl::Slot` instead of hard-coding `u8`.
This is super minor and non-blocking.

src/backend/metal/src/device.rs
```diff
@@ -1384,71 +1407,134 @@ impl hal::device::Device<Backend> for Device {
...
+
+                // We need to keep the temp file alive so that it doesn't get deleted until after a
+                // binary archive has been created.
+                let _temp_file = if !data.is_empty() {
+                    // It would be nice to use a `data:text/plain;base64` url here and just pass in a
+                    // base64-encoded version of the data, but metal validation doesn't like that:
+                    // -[MTLDebugDevice newBinaryArchiveWithDescriptor:error:]:1046: failed assertion `url, if not nil, must be a file URL.'
+
+                    let temp_file = tempfile::NamedTempFile::new().unwrap();
+
```
kvark, line 1437: nit: is it necessary to put each command between empty lines?

src/backend/metal/src/native.rs
```diff
@@ -204,18 +210,92 @@ pub struct ModuleInfo {
     pub rasterization_enabled: bool,
 }
 
+#[cfg(feature = "pipeline-cache")]
```
kvark, line 213: would it help to put all of the `pipeline-cache`-releated values into a submodule? It would allow us to have a single `#[cfg(feature = "pipeline-cache")]` as well as group them together.

kvark (approved):
This looks great!
I noted a few ideas here, but overall happy with this landing as soon as naga's gfx-22 train is up.

kvark:
@expenses Naga's "gfx-22" is up, please update the PR to use it (and make sure every commit is testable).

kvark:
@expenses can we make it so every commit is testable? If not sure, let's squash everything.

expenses:
Okay, I'll do a squash then I think.

expenses:
I'm trying to find a good way to do with while having the same commit message that doing a squash merge on github does. Maybe just doing it through the site would be better?

PR #3721, opened 2021-04-10
Added the ability to create a surface directly from a display, allowing to render without a windowing system

Uniformbuffer2:
Hi, i'm writing a Wayland compositor for my master thesis and i would like to present images on screen using only the rendering api, without using directly drm/kms system to reduce the code i have to manage.
So i have added to some functions to HAL and to the Vulkan Backend to take advantage of the `VK_KHR_display` vulkan extension that allow to create a surface directly from a display plane. 

It is important to note that the starting function, the `Instance::enumerate_available_displays`, will return the list of the displays available to be managed by the application. Generally compositors will take the control of every display available on the system when they are actively running, so those monitors will not be returned in the list of the available monitors; if a compositor is currently running, it is very likely that the function will return an empty list.
Changing to an empty TTY on Linux will make all the monitors available again to be gathered, so it is suggested to run programs that take advantage of this feature directly from the TTY (including the `quad_windowless` example).

I have added the `quad_windowless` example, a slightly modified `quad` example where the surface initialization is done using the new functions instead of the Winit ones. Since Winit is not used, inputs cannot be gathered from the windowing system, so currently the example present the Gfx logo for 5 seconds and than terminate. To be noticed that, at least on my system, the terminal used to run the example does not return automatically to "text mode", so the monitor will remain black until a TTY change has been made.
I will realize a separate component to makes possible to gather inputs to make the `quad_windowess` example more comparable to the `quad` one.

Since these functions rely on a Vulkan specific extension, it is possible to take advantage of this ability only from the Vulkan backend currently. Unfortunately, as far as i know, no other apis support such feature, but i have tried to make the HAL interface as general as possible to allow future integrations.

So i would like to know what do you think. Obviously i'm ready to adapt and change stuff if you think there is need of.

Thanks for listening, have a good day!


src/backend/vulkan/src/lib.rs
```diff
@@ -647,6 +649,285 @@ impl hal::Instance<Backend> for Instance {
...
+                hal::display::SurfaceTransformation::HorizontalMirror=>builder.transform(vk::SurfaceTransformFlagsKHR::HORIZONTAL_MIRROR),
+                hal::display::SurfaceTransformation::HorizontalMirrorRotate90=>builder.transform(vk::SurfaceTransformFlagsKHR::HORIZONTAL_MIRROR_ROTATE_90),
+                hal::display::SurfaceTransformation::HorizontalMirrorRotate180=>builder.transform(vk::SurfaceTransformFlagsKHR::HORIZONTAL_MIRROR_ROTATE_180),
+                hal::display::SurfaceTransformation::HorizontalMirrorRotate270=>builder.transform(vk::SurfaceTransformFlagsKHR::HORIZONTAL_MIRROR_ROTATE_270),
+                hal::display::SurfaceTransformation::Inherit=>builder.transform(vk::SurfaceTransformFlagsKHR::INHERIT)
+            };
+
+            match alpha
+            {
+                hal::display::DisplayPlaneAlpha::Opaque=>builder.alpha_mode(vk::DisplayPlaneAlphaFlagsKHR::OPAQUE),
```
kvark, line 919: similarly here, if it's flags on Vulkan, any reason not to follow?

src/hal/src/display/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+pub struct DisplayPlane<'a,B: Backend>
+{
+    /// Display mode
+    pub display_mode: &'a DisplayMode<'a,B>,
+    /// Plane index
+    pub plane: &'a Plane<'a,B>,
+    /// Supported alpha capabilities
+    pub supported_alpha: Vec<DisplayPlaneAlpha>,
+    /// The minimum source rectangle offset supported by this plane using the specified mode.
+    pub min_src_position: (i32,i32),
```
kvark, line 174: looks like we need new types for the `Extent` and `Origin` here, or something like this

src/backend/vulkan/src/lib.rs
```diff
@@ -1070,4 +1351,8 @@ impl hal::Backend for Backend {
     type Semaphore = native::Semaphore;
     type Event = native::Event;
     type QueryPool = native::QueryPool;
+
+    type Display = native::Display;
+    type Plane = u32;
```
kvark, line 1356: why does this need to be abstracted away? Can't we just say at the gfx-hal level that `type Plane = u32`, and be done with it?

kvark (commented):
Noted a few things :)

examples/quad_windowless/main.rs
```diff
@@ -0,0 +1,1044 @@
...
+    //Get the first available display mode (generally the preferred one)
+    let display_mode = &display_modes[0];
+    println!("Display mode: {:#?}",&display_mode);
+
+    //Create display plane
+    let display_plane = instance.create_display_plane(&display_mode,&plane).expect("Failed to create display plane");
+    println!("Display plane: {:#?}",&display_plane);
+
+    //Create a surface from the display
+    let surface = instance.create_display_plane_surface(
```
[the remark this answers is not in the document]
Uniformbuffer2: Sure, no problem, i will do it.

src/backend/vulkan/src/lib.rs
```diff
@@ -647,6 +649,285 @@ impl hal::Instance<Backend> for Instance {
             .functor
             .destroy_surface(surface.raw.handle, None);
     }
+
+    fn enumerate_available_displays<'a>(&self,adapter: &'a adapter::Adapter<Backend>)->Result<Vec<hal::display::Display<'a,Backend>>,hal::device::OutOfMemory>{
+        let display_extension = ash::extensions::khr::Display::new(&self.entry,&self.raw.inner);
```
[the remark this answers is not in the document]
Uniformbuffer2: you are right! I don't know why i thought that functions were related to the `Instance`. Maybe because `create_surface` is a `Instance` function, i thought they could be similar. I will move all that functions to `PhysicalDevice`.

src/backend/vulkan/src/lib.rs
```diff
@@ -647,6 +649,285 @@ impl hal::Instance<Backend> for Instance {
...
+        };
+
+        let mut displays = Vec::new();
+        for display_property in display_properties
+        {
+            let display_handle = native::Display(display_property.display);
+
+            let supported_transforms = {
+                let mut transformations = Vec::new();
+                if display_property.supported_transforms.contains(vk::SurfaceTransformFlagsKHR::IDENTITY) {transformations.push(hal::display::SurfaceTransformation::Identity);}
```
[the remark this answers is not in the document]
Uniformbuffer2: That is because all the functions that accept as parameter `SurfaceTransformation` (effectively there is only one) assume that only one bit will be enabled (a surface cannot be at the same time setted as Indentity or Rotated90 for example). Since there is no related error to return on that function, passing a `SurfaceTransformation` with multiple bits enabled will fall into undefined behaviour (probably the driver will use the mode related to the first enabled bit, while ignoring the rest, but i have not tested this). The same apply for the `DisplayPlaneAlpha` struct. So i thought i could avoid this problem by defining that bitflags as enums: while making the function that gather info returns the available transformations as a `Vec<SurfaceTransformation>`, the function that use `SurfaceTransformation` simply accept a enum value that can be only one of the available transformations. This was made to make the function safe by api, so that there is no need to make additions checks into the function.
If you prefer to have bitflags instead, i can replace them without problems.

src/backend/vulkan/src/lib.rs
Uniformbuffer2: I thought that others future apis could implement this path. Drm/kms and vulkan use a u32 but maybe another api could use an opaque object as handle. I have simply thought that representing it in that way could be be more future proof, but currently there is no need to be abstracted, so i can remove it and simply use u32 if you desire.

src/backend/vulkan/src/lib.rs
```diff
@@ -647,6 +649,285 @@ impl hal::Instance<Backend> for Instance {
...
+                {
+                    ash::vk::Result::ERROR_OUT_OF_HOST_MEMORY=>return Err(hal::device::OutOfMemory::Host),
+                    ash::vk::Result::ERROR_OUT_OF_DEVICE_MEMORY=>return Err(hal::device::OutOfMemory::Device),
+                    _=>panic!("Unexpected error returned")
+                };
+            }
+        }.iter().map(|display_mode_properties|
+        {
+            hal::display::DisplayMode {
+                display: &display,
```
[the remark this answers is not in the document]
Uniformbuffer2: The `DIsplay` member is used mainly in the `create_display_plane_surface` to make some tests end eventually throw `DisplayPlaneSurfaceError::UnsupportedParameters`. If you prefer i could remove the tests and replace the `Display` member with `PhysicalDevice`: it is needed for the vulkan functions (like the one in `create_display_plane`) and it is useful to avoid physical device/display.

src/hal/src/display/mod.rs
Uniformbuffer2: Sure, but i have noticed that there is `hal::image::Extent` that is a combination of width, height and depth so there could be some name conflict. Maybe i could just follow the Vulkan name scheme like Extent2D,Extent3D,Offset2D and Offset3D and put everything in something like `hal::common` mod. What do you think?

src/backend/vulkan/src/lib.rs
kvark: We can cross this bridge when we get there. Unless we have exact knowledge of the future work, let's stick to the minimum of changes here.

src/hal/src/display/mod.rs
kvark: I'm not worried about the name conflict. We aren't re-exporting everything from the root, so it's up to users to decide how to resolve this. All the user code I write, at least, would not be affected.

> Maybe i could just follow the Vulkan name scheme like Extent2D,Extent3D,Offset2D and Offset3D and put everything in something like hal::common mod. What do you think?

Currently we have `hal::window::Extent2D`... How do you feel about putting your code into the `window` module instead of having a separate one? It has the stuff about windowing, surface, and presentation. If you feel that a better isolation of code would be needed, we can have submodules, like `src/window/display.rs`, that are potentially re-exported from the `window`.

Uniformbuffer2: > How do you feel about putting your code into the `window` module instead of having a separate one?

I didn't noticed there was `hal::window::Extent2D`, so i can simply add a `Offset2D` structure below.

kvark:
What is the status of this PR?

Uniformbuffer3:
This PR can be closed because the feature has been finalized and merged with #3759.

PR #3722, opened 2021-04-13
max_value() is being deprecated, use ::MAX instead

hgallagher1993:
https://doc.rust-lang.org/std/primitive.u16.html#method.max_value
PR checklist:
- [X] tested examples with the following backends: DX12


kvark:
Looks like our MSRV for Naga matches this, phew!
bors r+

PR #3724, opened 2021-04-14
implement debug markers for metal

adamnemecek:


src/backend/metal/src/command.rs
```diff
@@ -1927,6 +1927,15 @@ where
                 offset,
             );
         }
+        Cmd::InsertDebugMarker { ref name } => {
+            encoder.insert_debug_signpost(name);
+        }
+        Cmd::PushDebugMarker { ref name } => {
+            encoder.push_debug_group(name);
```
kvark, line 1934: There is one thing I wasn't sure about when thinking about this. Maybe you can clarify.
In Metal, is it valid to, say, push debug group inside a render pass, and pop it outside of the render pass?
In Vulkan, it seems totally valid.

src/backend/metal/src/soft.rs
```diff
@@ -116,6 +116,13 @@ pub enum RenderCommand<R: Resources> {
         buffer: BufferPtr,
         offset: hal::buffer::Offset,
     },
+    InsertDebugMarker {
```
kvark, line 119: We don't want to force owning of `String` here. This would mean an unconditional heap allocation.
Instead, let's extend `trait Resources` to work with strings, i.e:
```rust
trait Resources {
  type Data;
  type Marker;
  ...
}
impl Resources for Own {
    type Data = Vec<u32>;
    type Marker = String;
    ...
}
#[derive(Debug)]
pub struct Ref;
impl<'a> Resources for &'a Ref {
    type Data = &'a [u32];
    type Marker = &'a str;
    ...
}
```

This would allow us to pass in `&str`, and let it be converted to `String` only when necessary

kvark (changes requested):
Thank you for writing this!
A few small things here to address

src/backend/metal/src/command.rs
adamnemecek: I believe it is but I'll double check with @warrenm, my spiritual advisor on Metal matters.

src/backend/metal/src/soft.rs
adamnemecek: Yeah, the allocation was also making me uncomfortable but I didn't know any better :-).

kvark (approved):
Awesome, thank you!
bors r+

adamnemecek:
Thanks you!

PR #3725, opened 2021-04-15
Add image usage parameter to the view creation

### LINKED ISSUE #3659: Add Usage flags to the ImageView creation
Vulkan's maintenance2 has this structure - https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkImageViewUsageCreateInfo.html
I think it would work well for the backends: DX12 for example would know if RTV/DSV/UAV/SRV need to be associated with a view without guessing.
From the user perspective, it's a little more verbose, but it's the information they already have. In the worst case, they'll pass the same flags they provided to `create_image`.

kvark:
Closes #3659
PR checklist:

### DIFF
--- a/examples/colour-uniform/main.rs
+++ b/examples/colour-uniform/main.rs
@@ -1063,6 +1063,7 @@ impl<B: Backend> ImageState<B> {
                 i::ViewKind::D2,
                 ColorFormat::SELF,
                 f::Swizzle::NO,
+                i::Usage::SAMPLED,
                 i::SubresourceRange {
                     aspects: f::Aspects::COLOR,
                     ..Default::default()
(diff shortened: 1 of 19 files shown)


cwfitzgerald (approved):
Code looks good. Could you explain what the usefullness of this change is?

bors r+

kvark:
It's not useful for `wgpu` right now, clearly, since WebGPU's views don't know about the usage.
Other clients, including gfx-portability (via KHR_Maintenance2) can specify the usage per view, which allows both the Vulkan driver to make better decisions, and our DX11/DX12 implementations to avoid creating unnecessary SRV/UAV/RTV/DSV stuff.

cwfitzgerald:
Ah, sounds useful, thanks!

PR #3726, opened 2021-04-15
Sampler reduction modes exposed

### LINKED ISSUE #3523: Sampler reduction modes
Would be great to support samplers that compute min/max instead of the weighted average.
See also - https://github.com/gpuweb/gpuweb/pull/1283

kvark:
Fixes #3523
PR checklist:


cwfitzgerald (approved):
LGTM!

bors r+

PR #3728, opened 2021-04-18
[0.7/dx12] Reset command allocators periodically

grovesNL:
Backport #3680 to hal-0.7 to avoid excess memory usage in DX12.

Some Bevy users were hitting this through wgpu, so it might be nice to backport this fix this in 0.7 too (instead of waiting for the next wgpu release).

I tried hal-0.7 with and without the fix on quad by repeatedly resizing the window, and the memory grows without the fix but seems stable with it (as expected).

PR checklist:
- [X] tested examples with the following backends: DX12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+### backend-dx12-0.7.1 (17-04-2021)
+  - reduce memory usage by resetting command allocators periodically
+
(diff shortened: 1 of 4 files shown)


kvark (approved):
Thank you!
bors r+

kvark:
This is published in 0.7.1 now 🚀

grovesNL:
Thank you!

PR #3729, opened 2021-04-18
Rename blend color to blend constants

kvark:
"color" is an overloaded term. Constants is what Vulkan calls it, and it matches our blend factor variants.
PR checklist:


cwfitzgerald (approved):
bors r+

PR #3731, opened 2021-04-18
macos: Set layerContentsPlacement to topleft

scoopr:


scoopr:
One weird thing I noticed, that when dragging a window between hidpi and lodpi screens, it tries to animate the change in scale, which I find weird. And I'm not sure how to disable, as I don't know around what I could do the CATransation. But it is a really minor thing.

src/backend/metal/src/lib.rs
```diff
@@ -338,7 +338,17 @@ impl GfxManagedMetalLayerDelegate {
             }
             decl.register();
         });
-        GfxManagedMetalLayerDelegate(Class::get(CAML_DELEGATE_CLASS).unwrap())
+        let klass = Class::get(CAML_DELEGATE_CLASS).unwrap();
+        let obj = unsafe { msg_send![klass, new] };
```
kvark, line 342: nice catch!

src/backend/metal/src/lib.rs
```diff
@@ -338,7 +338,17 @@ impl GfxManagedMetalLayerDelegate {
...
+        GfxManagedMetalLayerDelegate(obj)
+    }
+}
+
+impl Drop for GfxManagedMetalLayerDelegate {
+    fn drop(&mut self) {
+        unsafe {
+            let () = msg_send![self.0, release];
+        };
```
kvark, line 351: extra ";"

src/backend/metal/src/lib.rs
```diff
@@ -418,6 +428,16 @@ impl Instance {
             layer
         };
 
+        // Set set top-left layer placement, so that resizing the window
+        // doesn't result in stretching, which combined with newly rendered
+        // frames results in jerky movement.
+        #[allow(non_upper_case_globals)]
+        const NSViewLayerContentsPlacementTopLeft: i32 = 11;
```
kvark, line 435: I'm a bit concerned about this path. The view is something controlled by the client, not us. We should try to do minimum changes, if any, to that view. Is it reasonable to say that this should be a `winit` problem? Or basically, whoever actually manages the `NSView`, while gfx-rs shouldn't be that person.

kvark (changes requested):
Thank you for making this!

src/backend/metal/src/lib.rs
scoopr: now, I agree that winit should be setting the placement, but last time I tried, it didn't work, so I figured it is something about the layer within the view, and not the view itself.

I can do more tests with winit setting it too, but it'll take some time until I can get to it

kvark: thanks! if there is something we are missing here, we'd need to know, what it is.
Note that this issue doesn't need to block gfx-0.8, since it's not an API breaking fix.

scoopr:
How about this?

Though I was thinking of making the layer delegate a proper instance anyway, with the idea that it would actually return NO, but set the contentScale directly itself and then call out for the layer/view `display` method, if that would route through the winit to actually draw the layer at that point. But it may be that this is such small corner-case that it doesn't matter.

src/backend/metal/src/lib.rs
```diff
@@ -421,6 +421,14 @@ impl Instance {
             layer
         };
 
+
+        extern "C" {
```
kvark, line 425: let's move this to the top level, perhaps? it's good to see all extern declarations in one place

kvark (commented):
Looks good! Just one note

kvark (approved):
Thank you!
bors r+

PR #3732, opened 2021-04-19
[metal] integrate profiling

kvark:
This should help us look at the performance as a whole, and also compare spirv-cross to Naga shader paths.


kvark:
This should be harmless.
bors r+

PR #3737, opened 2021-04-27
[vk]: implement Device::start_capture() and Device::stop_capture() on Vulkan using renderdoc (as optional feature/dependency)

manuel-woelker:
TODOs:
 - [x] ~~find a way to use the correct device/window pointers instead of NULL~~ Null pointers should be fine for the majority of use cases
 - [x] integrate into example code? -> Integrated into the compute example
 - [x] Consider moving the renderdoc struct into the raw device?
 - [x] ~~Implement DX11, DX12, GL backends?~~ TBD in a separate PR

This is a follow up on #3708

Fixes #issue
PR checklist:

### DIFF
--- a/src/backend/vulkan/Cargo.toml
+++ b/src/backend/vulkan/Cargo.toml
@@ -30,6 +30,8 @@ parking_lot = "0.11"
 smallvec = "1.0"
 raw-window-handle = "0.3"
 inplace_it = "0.3.3"
+renderdoc-sys = "0.7.1"
+libloading = "0.7"
 
 [dependencies.naga]
 git = "https://github.com/gfx-rs/naga"
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -15,6 +15,7 @@ use hal::{
 use std::{ffi::CString, marker::PhantomData, mem, ops::Range, ptr, sync::Arc};
 
 use crate::{command as cmd, conv, native as n, pool::RawCommandPool, window as w, Backend as B};
+use ash::vk::Handle;
 
 #[derive(Debug, Default)]
 struct GraphicsPipelineInfoBuf<'a> {
@@ -1891,11 +1892,31 @@ impl d::Device<B> for super::Device {
     }
 
     fn start_capture(&self) {
-        //TODO: RenderDoc
+        unsafe {
+            match self.shared.instance.render_doc_entry {
+                Ok(ref entry) => {
+                    entry.api.StartFrameCapture.unwrap()(
+                        self.shared.raw.handle().as_raw() as *mut _,
+                        ptr::null_mut(),
+                    );
+                }
+                Err(ref cause) => warn!("Could not start render doc frame capture: {}", cause),
+            };
+        }
     }
 
     fn stop_capture(&self) {
-        //TODO: RenderDoc
+        unsafe {
+            match self.shared.instance.render_doc_entry {
+                Ok(ref entry) => {
+                    entry.api.EndFrameCapture.unwrap()(
+                        self.shared.raw.handle().as_raw() as *mut _,
+                        ptr::null_mut(),
+                    );
+                }
+                Err(ref cause) => warn!("Could not stop render doc frame capture: {}", cause),
+            };
+        }
     }
 }
 
--- a/src/backend/vulkan/src/lib.rs
+++ b/src/backend/vulkan/src/lib.rs
@@ -74,6 +74,8 @@ pub struct RawInstance {
     inner: ash::Instance,
     debug_messenger: Option<DebugMessenger>,
     get_physical_device_properties: Option<vk::KhrGetPhysicalDeviceProperties2Fn>,
+
+    pub render_doc_entry: Result<RenderDocEntry, String>,
 }
 
 pub enum DebugMessenger {
@@ -144,6 +146,16 @@ impl Into<Version> for u32 {
         Version(self)
     }
 }
+
+#[repr(C)]
+pub struct RenderDocEntry {
+    api: renderdoc_sys::RENDERDOC_API_1_4_1,
+    lib: libloading::Library,
+}
+
+unsafe impl Send for RenderDocEntry {}
+
+unsafe impl Sync for RenderDocEntry {}
 
 pub struct Instance {
     pub raw: Arc<RawInstance>,
@@ -565,6 +577,7 @@ impl hal::Instance<Backend> for Instance {
                 inner: instance,
                 debug_messenger,
                 get_physical_device_properties,
+                render_doc_entry: unsafe { load_renderdoc_entrypoint() },
             }),
             extensions,
             entry,
@@ -647,6 +660,51 @@ impl hal::Instance<Backend> for Instance {
             .functor
             .destroy_surface(surface.raw.handle, None);
     }
+}
+
+unsafe fn load_renderdoc_entrypoint() -> Result<RenderDocEntry, String> {
+    if !cfg!(debug_assertions) {
+        return Err("RenderDoc support is only enabled with 'debug_assertions'".into());
+    }
+
+    type GetApiFn = unsafe extern "C" fn(version: u32, out: *mut *mut std::ffi::c_void) -> i32;
+
+    #[cfg(windows)]
+    let renderdoc_filename = "renderdoc.dll";
+    #[cfg(all(unix, not(target_os = "android")))]
+    let renderdoc_filename = "librenderdoc.so";
+    #[cfg(target_os = "android")]
+    let renderdoc_filename = "libVkLayer_GLES_RenderDoc.so";
+
+    let renderdoc_lib = libloading::Library::new(renderdoc_filename).map_err(|e| {
+        format!(
+            "Unable to load renderdoc library '{}': {:?}",
+            renderdoc_filename, e
+        )
+    })?;
+
+    let get_api: libloading::Symbol<GetApiFn> =
+        renderdoc_lib.get(b"RENDERDOC_GetAPI\0").map_err(|e| {
+            format!(
+                "Unable to get RENDERDOC_GetAPI from renderdoc library '{}': {:?}",
+                renderdoc_filename, e
+            )
+        })?;
+    let mut obj = std::ptr::null_mut();
+    match get_api(10401, &mut obj) {
+        1 => Ok(RenderDocEntry {
+            api: *(obj as *mut renderdoc_sys::RENDERDOC_API_1_4_1),
+            lib: renderdoc_lib,
+        }),
+        return_value => Err(format!(
+            "Unable to get API from renderdoc library '{}': {}",
+            renderdoc_filename, return_value
+        )),
+    }
+    .map_err(|error| {
+        warn!("Error loading renderdoc library: {}", error);
+        error
+    })
 }
 
 #[derive(Debug, Clone)]
--- a/examples/compute/main.rs
+++ b/examples/compute/main.rs
@@ -71,6 +71,8 @@ fn main() {
             .unwrap()
     };
     let device = &gpu.device;
+    // Start render doc frame capture if available
+    device.start_capture();
     let queue_group = gpu.queue_groups.first_mut().unwrap();
 
     let glsl = fs::read_to_string("compute/shader/collatz.comp").unwrap();
@@ -257,6 +259,7 @@ fn main() {
             "Times: {:?}",
             slice::from_raw_parts::<u32>(mapping as *const u8 as *const u32, numbers.len()),
         );
+        device.stop_capture();
         device.unmap_memory(&mut staging_memory);
     }
 


src/backend/vulkan/Cargo.toml
```diff
@@ -30,6 +30,7 @@ parking_lot = "0.11"
 smallvec = "1.0"
 raw-window-handle = "0.3"
 inplace_it = "0.3.3"
+renderdoc = { version = "0.10.1", optional = true }
```
kvark, line 33: can we depend on renderdoc-sys instead of renderdoc here?

manuel-woelker: Should be technically possible, but it would mean probably reimplementing a good bit of the unsafe code in the renderdoc crate, for loading the lib and getting the function pointers
cf. https://github.com/ebkalderon/renderdoc-rs/blob/master/src/version.rs#L79

What would the advantages be of using renderdoc-sys instead of renderdoc?

manuel-woelker:
Implementing this for the other backends (DX11, DX12, GL) should be fairly similar.

Any preferences on having one PR per backend, or all in one?

src/backend/vulkan/Cargo.toml
kvark: I think it would make sense for us to fully control how external libraries are loaded. For example `renderdoc` crate uses
```
static RD_LIB: OnceCell<Library> = OnceCell::new();
```
but here we'd want to just own the library inside our `struct Instance`, just how we own the Vulkan DLL and other DLLs.

Overall, we want minimal dependencies and maximum control. With renderdoc-sys, we may have it as non-optional, since we'd be in full control of how the API is used.

kvark:
All in one is fine, but let's first figure out the proper way to do this on one backend.

manuel-woelker:
Next iteration, using renderdoc-sys and some pointer wrangling. RenderDoc still seems to be happy :-)

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,27 @@ impl d::Device<B> for super::Device {
     }
 
     fn start_capture(&self) {
-        //TODO: RenderDoc
+        #[cfg(feature = "renderdoc")]
+        unsafe {
+            ((*self.shared.instance.render_doc_entry.lock().0)
+                .StartFrameCapture
+                .unwrap())(
+                std::ptr::null::<std::ffi::c_void>() as *mut _,
```
kvark, line 1899: let's use `ptr::null_mut` here instead of this (ptr is already in scope)

src/backend/vulkan/src/lib.rs
```diff
@@ -559,12 +569,51 @@ impl hal::Instance<Backend> for Instance {
                 None
             }
         };
+        #[cfg(feature = "renderdoc")]
+        let render_doc_entry = unsafe {
+            type GetApiFn =
+                unsafe extern "C" fn(version: u32, out: *mut *mut std::ffi::c_void) -> i32;
+
+            let renderdoc_filename = "renderdoc.dll";
+            let renderdorc_lib = libloading::Library::new(renderdoc_filename).map_err(|e| {
```
kvark, line 578: We  shouldn't error on this. If we remove the feature, this code should be expected to handle *any* connection errors.
Let's also make it so this work is only done `if cfg!(debug_assertions)`.

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,27 @@ impl d::Device<B> for super::Device {
     }
 
     fn start_capture(&self) {
-        //TODO: RenderDoc
+        #[cfg(feature = "renderdoc")]
+        unsafe {
+            ((*self.shared.instance.render_doc_entry.lock().0)
```
kvark, line 1896: do we actually need to lock it?. Is calling `StartFrameCapture` requiring `&mut self`?

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,27 @@ impl d::Device<B> for super::Device {
...
     }
 
     fn stop_capture(&self) {
-        //TODO: RenderDoc
+        #[cfg(feature = "renderdoc")]
+        unsafe {
+            ((*self.shared.instance.render_doc_entry.lock().0)
+                .EndFrameCapture
+                .unwrap())(
+                std::ptr::null::<std::ffi::c_void>() as *mut _,
```
kvark, line 1911: also `ptr::null_mut()`

kvark (commented):
Thank you!
We are almost there!

src/backend/vulkan/Cargo.toml
manuel-woelker: Good point!

src/backend/vulkan/src/lib.rs
manuel-woelker: Using an `Option::None` here now instead. I was wondering whether a `Result<_, String>` might make sense, in order to log a warning when frame capture functionality is used but not available because of some initialization issue

Regarding `if cfg!(debug_assertions)`: Wouldn't that complete disable this functionality in production builds? I was under the impression that frame debugging was fairly useful on production code as well, and the overhead should be relatively minor.

src/backend/vulkan/src/device.rs
manuel-woelker: 👍

src/backend/vulkan/src/device.rs
manuel-woelker: Nope, I don't think so. I had to unsafe impl Sync+Send on the `RenderDocEntry` though.

The [renderdoc.rs code](https://github.com/ebkalderon/renderdoc-rs/blob/560731b184b784725f6c00e32c13f46764a13ec7/src/version.rs#L62) mentions some thread safety issues when loading the entry point itself. I have scanned the renderdoc docs, but I haven't found anything obvious regarding thread safety issues.

kvark: ok, in this case, let's not wrap it into a mutex, even. Just store it plan on the instance

manuel-woelker:
1. Any thoughts on how to handle this in the example code? There is currently a crude solution that allows starting and stopping a frame capture using "S" and "E"respectively. Any recommended approach?

2. If the plan is to encapsulate and reuse this functionality in other backends, where would the best place to put it be? gfx-hal?

Thanks for the review and advice, much appreciated!

kvark:
> Any thoughts on how to handle this in the example code? There is currently a crude solution that allows starting and stopping a frame capture using "S" and "E"respectively. Any recommended approach?

I recommend just unconditionally starting the capture in the compute example at start, and finishing it at the end. This would mean you don't need to write any input handling code.

> If the plan is to encapsulate and reuse this functionality in other backends, where would the best place to put it be? gfx-hal?

It's already in gfx-hal. The start/stop capture API is on the `hal::Device`

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,31 @@ impl d::Device<B> for super::Device {
...
+        unsafe {
+            match self.shared.instance.render_doc_entry.as_ref() {
+                Ok(entry) => {
+                    ((*entry.api).StartFrameCapture.unwrap())(ptr::null_mut(), ptr::null_mut());
+                }
+                Err(cause) => warn!("Could not start render doc frame capture: {}", cause),
+            };
+        }
+        #[cfg(not(feature = "renderdoc"))]
+        debug!("RenderDoc frame capture only supported with feature 'renderdoc' and 'debug_assertions'");
```
kvark, line 1904: they are going to see the `warn!` in this case, so this extra message seems redundant

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,31 @@ impl d::Device<B> for super::Device {
...
+        unsafe {
+            match self.shared.instance.render_doc_entry.as_ref() {
+                Ok(entry) => {
+                    ((*entry.api).EndFrameCapture.unwrap())(ptr::null_mut(), ptr::null_mut());
+                }
+                Err(cause) => warn!("Could not stop render doc frame capture: {}", cause),
+            };
+        }
+        #[cfg(not(feature = "renderdoc"))]
+        debug!("RenderDoc frame capture only supported with feature 'renderdoc' and 'debug_assertions'");
```
kvark, line 1918: same here

src/backend/vulkan/Cargo.toml
```diff
@@ -15,6 +15,7 @@ edition = "2018"
 [features]
 default = []
 use-rtld-next = ["libc"]
+renderdoc = ["renderdoc-sys", "libloading"]
```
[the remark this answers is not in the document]
manuel-woelker: Ok, I removed the feature, which also cleaned up all the conditional cfg's

src/backend/vulkan/src/device.rs
manuel-woelker: removed by the feature removal as well

src/backend/vulkan/src/device.rs
manuel-woelker: ditto

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,31 @@ impl d::Device<B> for super::Device {
...
     }
 
     fn stop_capture(&self) {
-        //TODO: RenderDoc
+        #[cfg(feature = "renderdoc")]
+        unsafe {
+            match self.shared.instance.render_doc_entry.as_ref() {
```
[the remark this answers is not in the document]
manuel-woelker: ✔️

manuel-woelker:
Pushed another version, with removed feature flag and some cleanup :-)

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,25 @@ impl d::Device<B> for super::Device {
     }
 
     fn start_capture(&self) {
-        //TODO: RenderDoc
+        unsafe {
+            match &self.shared.instance.render_doc_entry {
```
kvark, line 1895: Instead of matching on a reference, let's match on the actual thing. We'd then have `Ok(ref entry)` and `Err(ref cause)` cases

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,25 @@ impl d::Device<B> for super::Device {
...
+                Ok(entry) => {
+                    ((*entry.api).StartFrameCapture.unwrap())(ptr::null_mut(), ptr::null_mut());
```
kvark, line 1897: we have the device handle, we should be passing it here to RenderDoc

src/backend/vulkan/src/device.rs
```diff
@@ -1891,11 +1891,25 @@ impl d::Device<B> for super::Device {
...
+            };
+        }
     }
 
     fn stop_capture(&self) {
-        //TODO: RenderDoc
+        unsafe {
+            match &self.shared.instance.render_doc_entry {
+                Ok(entry) => {
+                    ((*entry.api).EndFrameCapture.unwrap())(ptr::null_mut(), ptr::null_mut());
```
kvark, line 1908: same here

src/backend/vulkan/src/lib.rs
```diff
@@ -649,6 +662,54 @@ impl hal::Instance<Backend> for Instance {
     }
 }
 
+fn load_renderdoc_entrypoint() -> Result<RenderDocEntry, String> {
```
kvark, line 665: let's just make this an `unsafe fn`

kvark (changes requested):
Thank you!
Here is the last round of change requests :)

manuel-woelker:
Ready for another review, I cleaned up the match expressions and marked `load_renderdoc_entrypoint()` unsafe

src/backend/vulkan/src/lib.rs
```diff
@@ -649,6 +662,54 @@ impl hal::Instance<Backend> for Instance {
     }
 }
 
+unsafe fn load_renderdoc_entrypoint() -> Result<RenderDocEntry, String> {
+    if cfg!(debug_assertions) {
+        {
```
kvark, line 667: unnecessary nesting?

src/backend/vulkan/src/lib.rs
```diff
@@ -649,6 +662,54 @@ impl hal::Instance<Backend> for Instance {
...
+                    renderdoc_filename, return_value
+                )),
+            }
+            .map_err(|error| {
+                warn!("Error loading renderdoc library: {}", error);
+                error
+            })
+        }
+    } else {
+        Err("RenderDoc support is only enabled with 'debug_assertions'".into())
```
kvark, line 709: nit: could early exit here in the beginning of the function, to not indent the whole thing. We have early-outs anyway with "?" in the body.

kvark (commented):
This looks good! Just one nit

manuel-woelker:
Reordered the `cfg!(debug_assertions)` and cleaned up the nesting

src/backend/vulkan/src/lib.rs
manuel-woelker: yep, removed

src/backend/vulkan/src/lib.rs
manuel-woelker: sounds good

kvark:
Great!
bors r+

kvark (approved):

kvark:
Thank you for tirelessly addressing my concerns! I'm excited to finally remove all renderdoc-specific code from downstream crates :)

adamnemecek:
@manuel-woelker This is all you. If i were to make a suggestion (sorry for being absent) but I would move the start capture on the instance itself so that on the device, we can just call `instance.start_capture` but maybe this doesn't matter.

PR #3738, opened 2021-04-28
Make Dx11 fill_buffer use an extra buffer for non 16-byte aligned fills

### LINKED ISSUE #3718: fill_buffer start has to be 16-byte aligned on DX11
See https://github.com/gfx-rs/wgpu/issues/1303#issuecomment-813756734
We may need to expose a limit for this.
cc @cwfitzgerald

Imberflur:
Fixes #3718
PR checklist:

### DIFF
--- a/src/backend/dx11/src/lib.rs
+++ b/src/backend/dx11/src/lib.rs
@@ -2811,29 +2811,108 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
             assert_eq!(size % 4, 0, "Buffer sub range size must be multiple of 4");
         }
 
-        let mut desc: d3d11::D3D11_UNORDERED_ACCESS_VIEW_DESC = mem::zeroed();
-        desc.Format = dxgiformat::DXGI_FORMAT_R32_TYPELESS;
-        desc.ViewDimension = d3d11::D3D11_UAV_DIMENSION_BUFFER;
-        *desc.u.Buffer_mut() = d3d11::D3D11_BUFFER_UAV {
-            FirstElement: sub.offset as u32 / 4,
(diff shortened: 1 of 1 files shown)


src/backend/dx11/src/lib.rs
```diff
@@ -2811,29 +2811,110 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
...
+                back: 1,
+            };
+
+            self.context.CopySubresourceRegion(
+                buffer.internal.raw as _,
+                0,
+                sub.offset as _, // offset in bytes
+                0,
+                0,
+                // ID3D11Buffer cast to ID3D11Resource valid?
```
cwfitzgerald, line 2912: Yes

From: https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nn-d3d11-id3d11buffer

> The ID3D11Buffer interface inherits from ID3D11Resource. 

And because of how COM works, you can freely cast the pointers like this.

src/backend/dx11/src/lib.rs
```diff
@@ -2811,29 +2811,110 @@ impl command::CommandBuffer<Backend> for CommandBuffer {
...
+                )
+            );
+
+            let temp_buffer = ComPtr::from_raw(temp_buffer);
+
+            let src_box = d3d11::D3D11_BOX {
+                left: 0,
+                top: 0,
+                front: 0,
+                // TODO: is this supposed to be in bytes?
```
cwfitzgerald, line 2900: Yeah

cwfitzgerald (approved):
LGTM!

Just remove the two unneeded comments and lets ship this puppy

Imberflur:
Cleaned up those comments!

kvark:
bors r=cwfitzgerald

PR #3742, opened 2021-04-30
metal: Link to QuartzCore

scoopr:
Fixes linking issues when QuartzCore is not linked by some other means

Fixes #issue
PR checklist:


kvark (approved):
Feels a bit strange that we are adding this link just for one single thing.

PR #3748, opened 2021-05-07
renderdoc: backend-agnostic RenderDoc frame capture functionality

manuel-woelker:
This is a follow up to #3737 to bring RenderDoc frame capture functionality to more backends

This is a first implementation sketch to bring RenderDoc frame capture to more backends. My first goal was to get the internal API right, and figure out if gfx-hal is the right location for this code (maybe auxil would be better suited?)

Fixes #issue
PR checklist:
- [x] `make` succeeds ~~(on *nix)~~ on Windows 10
- [x] tested examples with the following backends: Vulkan, DX11, DX12

### DIFF
--- a/src/auxil/renderdoc/Cargo.toml
+++ b/src/auxil/renderdoc/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "gfx-renderdoc"
+version = "0.1.0"
+description = "Generic RenderDoc integration used by gfx-rs backends"
+homepage = "https://github.com/gfx-rs/gfx"
+repository = "https://github.com/gfx-rs/gfx"
+keywords = ["renderdoc"]
+license = "MIT OR Apache-2.0"
+authors = ["The Gfx-rs Developers"]
+documentation = "https://docs.rs/gfx-auxil-renderdoc"
+categories = ["graphics"]
+workspace = "../../../"
+edition = "2018"
+
+[lib]
+name = "gfx_renderdoc"
+
+[dependencies]
+renderdoc-sys = "0.7.1"
+libloading = "0.7"
+log = "0.4"
--- a/src/auxil/renderdoc/src/lib.rs
+++ b/src/auxil/renderdoc/src/lib.rs
@@ -1,0 +1,127 @@
+#![warn(
+    trivial_casts,
+    trivial_numeric_casts,
+    unused_extern_crates,
+    unused_import_braces,
+    unused_qualifications
+)]
+
+//! RenderDoc integration - https://renderdoc.org/
+
+/// The dynamically loaded RenderDoc API function table
+#[repr(C)]
+#[derive(Debug)]
+pub struct RenderDocApi {
+    api: renderdoc_sys::RENDERDOC_API_1_4_1,
+    lib: libloading::Library,
+}
+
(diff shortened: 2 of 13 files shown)


kvark:
This shouldn't be in gfx-hal because it's not a part of the API. It's just a shared implementation block.
If you think it's worth sharing, let's move it into a separate crate, like `src/auxil/renderdoc`?

manuel-woelker:
I moved the renderdoc code to a new crate `gfx-auxil-renderdoc` and used that in the Vulkan and DX12 backends.

The DX12 implementation _seems_ to work, but RenderDoc crashes when loading the capture, not sure what the issue is there, will investigate further.

src/auxil/renderdoc/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "gfx-auxil-renderdoc"
```
kvark, line 2: let's just call it `gfx-renderdoc` perhaps?

src/auxil/renderdoc/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.0.1"
```
kvark, line 3: let's make it `0.1.0`

kvark (approved):

kvark:
Looks like for D3D12 we'd need to try reproducing this in a smaller example. I.e. something using d3d12-rs and winapi directly.

manuel-woelker:
After some fruitless dead-ends I decided to try to use the quad example again to verify the renderdoc integration.
This fixed both of the issues I had. Captures now look good on Vulkan, DX11 and DX12.

kvark:
> After some fruitless dead-ends I decided to try to use the quad example again to verify the renderdoc integration.

So what was the problem then, originally?

src/auxil/renderdoc/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl Default for RenderDoc {
+    fn default() -> Self {
+        if !cfg!(debug_assertions) {
+            return RenderDoc::NotAvailable {
+                reason: "RenderDoc support is only enabled with 'debug_assertions'".into(),
+            };
+        }
+        unsafe {
```
kvark, line 45: can we move this into `pub unsafe fn new() -> Self` method on `RenderDoc`? This will allow backends to overrule the decision to check for debug assertions, if needed

src/backend/dx12/src/device.rs
```diff
@@ -3777,11 +3777,17 @@ impl d::Device<B> for Device {
     }
 
     fn start_capture(&self) {
```
kvark, line 3779: I think all of the `start_capture` and `end_capture` should themselves be `unsafe fn`, to match the other methods. But this is slightly above the scope of this PR. If you feel positive about doing this change, let's do it right here.

kvark (commented):
Looks great!

src/auxil/renderdoc/src/lib.rs
manuel-woelker: Sure thing!

kvark (approved):
shipit!
bors r+

PR #3751, opened 2021-05-09
Draft support for buffer sizes

lachlansneff:
In progress.

I'm not sure my approach is the best.

### DIFF
--- a/src/backend/metal/Cargo.toml
+++ b/src/backend/metal/Cargo.toml
@@ -57,7 +57,7 @@ features = ["msl"]
 optional = true
 
 [dependencies.naga]
-#git = "https://github.com/gfx-rs/naga"
+git = "https://github.com/gfx-rs/naga"
 #tag = "gfx-23"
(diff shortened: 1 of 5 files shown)


kvark:
Copying from Matrix discussion for posterity:
> We shouldn't be asking Metal about anything, since we should have the data (stuff passed into write_descriptor_sets). Also, this data may be more restrictive than what Metal tells us (e.g. the descriptor not covering the whole buffer, or Metal buffer being internally shared between gfx buffers, etc).
So basically, each descriptor set should carry a Vec<BufferAddress> - the sizes of the storage buffer bindings. The command buffer also would have Vec<BufferAddress>, composed of the join of the descriptor set's vectors.
when a descriptor set is bound, the command buffer would overwrite some of these sizes in its list.
(the range corresponding to this set)

Also, it appears that wgpu update is basically blocked on this being solved now - https://github.com/gfx-rs/wgpu/pull/1381

Issue #3420, opened 2020-10-18
Panic on reopen Android app. ERROR_SURFACE_LOST_KHR


enfipy:
Short info header:
- GFX version: gfx-backend-vulkan-0.6.1
- OS: Android, Miui Global 11.0.6
- GPU: Vulkan

I'm not sure that the actual error is with `gfx-backend-vulkan` but the error links to this crate.

When I build an Android app with Wgpu - everything works well until I hide and reopen the app several times. After some time it just crashes with the following error: `thread '<unnamed>' panicked at 'Unable to query surface capabilities: ERROR_SURFACE_LOST_KHR', /Users/enfipy/.cargo/registry/src/github.com-1ecc6299db9ec823/gfx-backend-vulkan-0.6.1/src/window.rs:343:10`.

I think this can be a `winit` error but I'm not sure about it. I was able to reproduce the error with the hello_world example in [android-ndk-rs](https://github.com/rust-windowing/android-ndk-rs/tree/master/ndk-examples).

kvark:
We should make the Vulkan backend to expect this error and not panic on it.

PR #3754, opened 2021-05-09
Fix panic on Android app reopen

zakorgy:
Fixes https://github.com/gfx-rs/gfx/issues/3420
I still have to test this on Android. Also I wasn't sure that changing `capabilities` to return with a `Result` thus adding an API change would be the proper way to do this, so I just put some `default`s there.
PR checklist:

### DIFF
--- a/src/backend/vulkan/src/window.rs
+++ b/src/backend/vulkan/src/window.rs
@@ -249,11 +249,16 @@ impl w::Surface<Backend> for Surface {
     fn capabilities(&self, physical_device: &PhysicalDevice) -> w::SurfaceCapabilities {
         // Capabilities
         let caps = unsafe {
-            self.raw
+            match self.raw
                 .functor
                 .get_physical_device_surface_capabilities(physical_device.handle, self.raw.handle)
-        }
-        .expect("Unable to query surface capabilities");
+                {
+                    Ok(caps) => caps,
+                    #[cfg(target_os = "android")]
+                    Err(vk::Result::ERROR_SURFACE_LOST_KHR) => vk::SurfaceCapabilitiesKHR::default(),
+                    Err(e) => panic!("Unable to query surface capabilities {:?}", e),
+                }
+        };
 
         // If image count is 0, the support number of images is unlimited.
         let max_images = if caps.max_image_count == 0 {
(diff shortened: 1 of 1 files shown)


src/backend/vulkan/src/window.rs
```diff
@@ -249,11 +249,16 @@ impl w::Surface<Backend> for Surface {
...
-            self.raw
+            match self.raw
                 .functor
                 .get_physical_device_surface_capabilities(physical_device.handle, self.raw.handle)
-        }
-        .expect("Unable to query surface capabilities");
+                {
+                    Ok(caps) => caps,
+                    #[cfg(target_os = "android")]
+                    Err(vk::Result::ERROR_SURFACE_LOST_KHR) => vk::SurfaceCapabilitiesKHR::default(),
```
Gordon-F, line 258: It would be great to log some warning here. And in other places with same logic.

src/backend/vulkan/src/window.rs
```diff
@@ -308,11 +318,15 @@ impl w::Surface<Backend> for Surface {
...
-            self.raw
+            match self.raw
                 .functor
-                .get_physical_device_surface_formats(physical_device.handle, self.raw.handle)
-        }
-        .expect("Unable to query surface formats");
+                .get_physical_device_surface_formats(physical_device.handle, self.raw.handle) {
+                    Ok(formats) => formats,
+                    #[cfg(target_os = "android")]
+                    Err(vk::Result::ERROR_SURFACE_LOST_KHR) => return None,
```
kvark, line 326: This is a bit more tricky, unfortunately. `None` result means that all formats are supported. Is this really what we want to report if the surface is lost? I think returning `Some(Vec::new())` would be a little better

src/backend/vulkan/src/window.rs
```diff
@@ -308,11 +318,15 @@ impl w::Surface<Backend> for Surface {
...
         let raw_formats = unsafe {
-            self.raw
+            match self.raw
                 .functor
-                .get_physical_device_surface_formats(physical_device.handle, self.raw.handle)
-        }
-        .expect("Unable to query surface formats");
+                .get_physical_device_surface_formats(physical_device.handle, self.raw.handle) {
+                    Ok(formats) => formats,
+                    #[cfg(target_os = "android")]
```
kvark, line 325: let's not have this on Android only, let's have this everywhere

src/backend/vulkan/src/window.rs
```diff
@@ -284,11 +289,16 @@ impl w::Surface<Backend> for Surface {
...
         let raw_present_modes = unsafe {
-            self.raw
+            match self.raw
                 .functor
                 .get_physical_device_surface_present_modes(physical_device.handle, self.raw.handle)
-        }
-        .expect("Unable to query present modes");
+                {
+                    Ok(present_modes) => present_modes,
+                    #[cfg(target_os = "android")]
```
kvark, line 297: same here, we can have this code on all platforms

src/backend/vulkan/src/window.rs
```diff
@@ -284,11 +289,16 @@ impl w::Surface<Backend> for Surface {
...
+                    Err(vk::Result::ERROR_SURFACE_LOST_KHR) => vec![vk::PresentModeKHR::default()],
```
kvark, line 298: would it make sense to return an empty vector here?

kvark (changes requested):
Thank you for addressing this!
We'll need one more iteration on the code :)

kvark (approved):

PR #3756, opened 2021-05-11
External memory feature for Vulkan backend

Uniformbuffer3:
Hi, i have implemented the external memory functions for the Vulkan backend. 
This is a work in progress and i would like to know your opinion about it, especially about the api.
I have also written a test program (that it is not in the pr), that i use to test these functions. If you want i could upload it somewhere or introduce it as a reproducible test for this feature. As the tests will succeed, i will update the list below. Unfortunately i can test it only from a linux prospective.

Thanks for the dedicated time,
Have a good day

Compile test:
- [x] Linux
- [x] Windows
- [x] Android
- [x] Wasm32
- [ ] MacOS
- [ ] iOS

Test application: https://github.com/Uniformbuffer3/gfx_external_memory_test
The test application will try to:
- Create external memory (will also be wrote some data inside)
- Export that memory
- Re-import that memory
- Check if the data inside the re-imported memory is the same as the one wrote in the first phase

The test application will try to use a subgroup of this function if the system does not support the whole iteration.

Legend:
:heavy_check_mark: : Passed
:x: : Failed
:fast_forward: : Skipped (due to previous failed test or because unsupported by the system)

**Linux Intel Haswell Vulkan driver:**
OPAQUE_FD
create_allocate_external_buffer:	:heavy_check_mark:
export_memory:				:heavy_check_mark:
import_external_buffer:			:heavy_check_mark:
data_check:					:heavy_check_mark:

DMA_BUF
create_allocate_external_buffer:	:heavy_check_mark:
export_memory:				:heavy_check_mark:
import_external_buffer:			:heavy_check_mark:
data_check:					:heavy_check_mark:

HOST_ALLOCATION
create_allocate_external_buffer::heavy_check_mark:
export_memory::heavy_check_mark:
import_external_buffer::heavy_check_mark:
data_check::heavy_check_mark:

HOST_MAPPED_FOREIGN_MEMORY
create_allocate_external_buffer:	:fast_forward:
export_memory:				:fast_forward:
import_external_buffer:			:fast_forward:
data_check:					:fast_forward:

**Linux Nvidia Maxwell (2° gen) Proprietary Vulkan driver:**
OPAQUE_FD
create_allocate_external_buffer::heavy_check_mark:
export_memory::heavy_check_mark:
import_external_buffer::heavy_check_mark:
data_check::heavy_check_mark:

DMA_BUF
create_allocate_external_buffer::fast_forward:
export_memory::fast_forward:
import_external_buffer::fast_forward:
data_check::fast_forward:

HOST_ALLOCATION
create_allocate_external_buffer::heavy_check_mark:
export_memory::heavy_check_mark:
import_external_buffer::heavy_check_mark:
data_check::heavy_check_mark:

HOST_MAPPED_FOREIGN_MEMORY
create_allocate_external_buffer::fast_forward:
export_memory::fast_forward:
import_external_buffer::fast_forward:
data_check::fast_forward:


src/hal/src/device.rs
```diff
@@ -714,6 +714,56 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    unsafe fn create_external_buffer(
+        &self,
+        external_memory_type_flags: external_memory::ExternalMemoryTypeFlags,
+        usage: buffer::Usage,
+        sparse: memory::SparseFlags,
+        size: u64,
+    ) -> Result<B::Buffer, external_memory::ExternalBufferCreateError>;
+
+    /// Allocate external memory
+    unsafe fn allocate_exportable_memory(
```
kvark, line 727: is this meant to be `exportable`? the comment says "external"

src/hal/src/device.rs
```diff
@@ -714,6 +714,56 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+        &self,
+        external_memory_types: external_memory::ExternalMemoryTypeFlags,
```
kvark, line 729: let's call things consistently, the other place calls this `external_memory_type_flags`

src/hal/src/external_memory/fd.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+#[cfg_attr(feature = "unstable", doc(cfg(unix)))]
+#[derive(Debug)]
+#[allow(non_camel_case_types)]
+/// External memory that rely on unix file descriptors
+pub enum ExternalMemoryFd {
+    /// Tmp
+    OPAQUE_FD(Fd, u64),
```
kvark, line 29: We should use the proper snake naming scheme

src/hal/src/external_memory/fd.rs
```diff
@@ -0,0 +1,144 @@
...
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+#[cfg_attr(feature = "unstable", doc(cfg(unix)))]
+#[derive(Debug)]
+#[allow(non_camel_case_types)]
+/// External memory that rely on unix file descriptors
+pub enum ExternalMemoryFd {
```
kvark, line 27: let's move the external memory definitions (not tied to gfx) into a separate crate, like `src/auxil/external-memory`

kvark, line 27: Does it make sense to separate the different external memory types? Can there instead be one `enum` with both Unix FD's, DXGI handles, and Apple's IO surfaces?

src/hal/src/device.rs
```diff
@@ -714,6 +714,56 @@ pub trait Device<B: Backend>: fmt::Debug + Any + Send + Sync {
...
+    unsafe fn export_memory_as_fd(
+        &self,
+        external_memory_type: external_memory::ExternalMemoryFdType,
+        memory: &B::Memory,
+    ) -> Result<std::os::unix::io::RawFd, external_memory::ExternalMemoryExportError>;
+
+    #[cfg(any(windows))]
+    #[cfg_attr(feature = "unstable", doc(cfg(windows)))]
+    /// Export memory as windows handle
+    unsafe fn export_memory_as_handle(
```
kvark, line 755: There is quite a few new entry points and types. Is all of this necessary? Can we find a way to reduce the API surface to the minimum that would be sufficient?

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+        external_memory_properties: ExternalMemoryProperties,
+    ) -> Self {
+        Self {
+            usage,
+            sparse,
+            external_memory_properties,
+        }
+    }
+    /// Is the queried configuration exportable
+    pub fn get_queried_buffer_usage(&self) -> crate::buffer::Usage {
```
kvark, line 53: Rust idiomatic style says we don't need `get_` prefix here (true for all accessors)

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,263 @@
...
+    memory_type: ExternalMemoryType,
+    compatible_memory_types: ExternalMemoryTypeFlags,
+    export_from_imported_memory_types: ExternalMemoryTypeFlags,
+}
+impl ExternalMemoryProperties {
+    /// Constructor
+    pub fn new(
+        exportable: bool,
+        importable: bool,
+        dedicated_allocation: bool,
```
kvark, line 92: As an example of trying to reduce the API surface here, would it make sense to just limit the API to dedicated allocations only? This should be simpler, and yet I think it would cover all the cases at least Gecko cares about.

kvark (commented):
This is some grand work! Let's try to see how we can reduce the complexity (and associated parameters, such as maintenance cost, compile times, etc) while still keeping the use cases covered.

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Assuming that every external memory would be a dedicated allocation would simplify a lot the API. Since they could not be suballocated, makes no sense to keep separated the functions for the creation of external buffer and external memory, so that the two functions that do this could be merged toghether while reducing the chance of wrong api utilization (so improving safety).

kvark: Wonderful. I hope you don't mind reducing the scope a bit here :)

Uniformbuffer3: No problem, i have already merged the functions and i'm working to fix all the other problems. I will push the new code soon

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Right, sometime the java ghost of the past come back. I will change that

src/hal/src/device.rs
Uniformbuffer3: Yes, my bad. `allocate_external_memory` would be more correct. Now that the allocate memory and create buffer functions have been merged, the new function is called `create_allocate_external_buffer`.

src/hal/src/device.rs
Uniformbuffer3: Right. Corrected.

src/hal/src/external_memory/fd.rs
Uniformbuffer3: I had chosen this name patter so that it was matching with the `ExternalMemoryTypeFlags`, but i had not thought that the user will anyway build that type with something "subtype".into(), so matching name was not so relevant. With the new code a lot of "subtypes" has removed and every enum is using snake naming scheme.

src/hal/src/external_memory/fd.rs
Uniformbuffer3: Sure, done!

Uniformbuffer3: Since, with the new code, all platform specific functions have been merged, most of those "subtypes" are no more needed and they have been removed. There are a couple of types that are used in the vulkan backend to make possible some "match arms" to share code path, so reducing the code size. I will try to remove them too if possible.

src/hal/src/device.rs
Uniformbuffer3: With the new code the patform specific functions have been merged into the `export_memory` function. Now the whole external memory api is composed by 1 physical device function and 3 device function (2 of them are `Buffer` specific, so there will be 2 more for the `Image` counterpart). Maybe could be possible to pass both buffer or image to the same function using an enum, so that the same function could be used for both buffer and image. In this way there will be only 3 device functions.

src/hal/src/external_memory/mod.rs
Uniformbuffer3: The new code has been pushed. I have also modified the test application to match the new api. Since the new api is more concentrated and makes more integrated checks, the test application code has been reduced to 1/3 of the previous version, so this means the effort to use this feature has been greatly reduced.

src/backend/vulkan/src/conv.rs
```diff
@@ -623,3 +623,103 @@ pub fn map_vk_memory_heap_flags(flags: vk::MemoryHeapFlags) -> hal::memory::Heap
 
     hal_flags
 }
+
+use hal::external_memory;
+
+pub fn map_external_memory_handle_types(
+    flags: external_memory::ExternalMemoryTypeFlags,
+) -> vk::ExternalMemoryHandleTypeFlags {
+    let mut vk_flags = vk::ExternalMemoryHandleTypeFlags::empty();
```
kvark, line 632: if hal flags are 1:1 with Vulkan flags, we can just do `from_bits` without checking each individual bit

src/backend/vulkan/src/conv.rs
```diff
@@ -623,3 +623,103 @@ pub fn map_vk_memory_heap_flags(flags: vk::MemoryHeapFlags) -> hal::memory::Heap
...
+    if flags.contains(external_memory::ExternalMemoryTypeFlags::HOST_ALLOCATION) {
+        vk_flags |= vk::ExternalMemoryHandleTypeFlags::HOST_ALLOCATION_EXT;
+    }
+    if flags.contains(external_memory::ExternalMemoryTypeFlags::HOST_MAPPED_FOREIGN_MEMORY) {
+        vk_flags |= vk::ExternalMemoryHandleTypeFlags::HOST_MAPPED_FOREIGN_MEMORY_EXT;
+    }
+    vk_flags
+}
+/*
+pub fn map_vk_external_memory_handle_type_flags(
```
kvark, line 678: leftover code

src/backend/vulkan/src/physical_device.rs
```diff
@@ -248,7 +248,8 @@ impl PhysicalDeviceFeatures {
             | Features::MUTABLE_COMPARISON_SAMPLER
             | Features::MUTABLE_UNNORMALIZED_SAMPLER
             | Features::TEXTURE_DESCRIPTOR_ARRAY
-            | Features::BUFFER_DESCRIPTOR_ARRAY;
+            | Features::BUFFER_DESCRIPTOR_ARRAY
+            | Features::EXTERNAL_MEMORY;
```
kvark, line 252: Why are we exposing this unconditionally? It should be based off the presense of the extensions "DedicatedAllocation" and friends (everything that is required for external memory).

src/backend/vulkan/src/physical_device.rs
```diff
@@ -582,6 +583,18 @@ impl PhysicalDeviceInfo {
             requested_extensions.push(vk::KhrGetDisplayProperties2Fn::name()); // TODO NOT NEEDED, RIGHT?
         }
 
+        if requested_features.contains(Features::EXTERNAL_MEMORY) {
```
kvark, line 586: need to also check for `self.api_version`. Some of these extensions were promoted into core at some point, which means they may no longer be advertised.

src/backend/vulkan/src/physical_device.rs
```diff
@@ -889,6 +902,57 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
             None
         };
 
+        let dedicated_allocation =
```
kvark, line 905: these all should be based on `requested_features` instead of `enabled_extensions`

src/backend/vulkan/src/physical_device.rs
```diff
@@ -1083,6 +1156,63 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+    fn query_external_buffer_properties(
+        &self,
+        usage: hal::buffer::Usage,
+        sparse: hal::memory::SparseFlags,
+        memory_type: hal::external_memory::ExternalMemoryType,
+    ) -> Result<
+        hal::external_memory::ExternalBufferProperties,
+        hal::external_memory::ExternalMemoryQueryError,
+    > {
+        let external_memory_capabilities_extension = match &self
```
kvark, line 1168: Let's not match on a reference. Instead, do ref binding, i.e. `Some(ref functor)`

src/backend/vulkan/src/physical_device.rs
```diff
@@ -1083,6 +1156,63 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+            .instance
+            .external_memory_capabilities
+        {
+            Some(functor) => functor,
+            _ => return Err(hal::external_memory::ExternalMemoryQueryError::UnsupportedFeature),
```
kvark, line 1173: I don't think this is necessary. We should just `unwrap()`.
This is a very low-level API. The only kind of errors we should be returning are ones that users can't expect.
In case of this, the user knows when they requested the `EXTERNAL_MEMORY` feature, hence they know when they can call this.

src/backend/vulkan/src/physical_device.rs
```diff
@@ -1083,6 +1156,63 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
...
+            vk_mem_properties
+                .external_memory_features
+                .contains(vk::ExternalMemoryFeatureFlags::IMPORTABLE),
+            vk_mem_properties
+                .export_from_imported_handle_types
+                .contains(vk_external_memory_type),
+            memory_type,
+        );
+
+        Ok(hal::external_memory::ExternalBufferProperties::new(
```
kvark, line 1209: I suppose this function could then just return the structure instead of being `Result`

src/backend/vulkan/src/physical_device.rs
```diff
@@ -1287,6 +1426,7 @@ impl adapter::PhysicalDevice<Backend> for PhysicalDevice {
             performance_caveats: Default::default(),
             dynamic_pipeline_states: DynamicStates::all(),
             downlevel: DownlevelProperties::all_enabled(),
+            external_memory_limits: external_memory_limits,
```
kvark, line 1429: nit: `external_memory_limits,`

src/backend/vulkan/src/conv.rs
Uniformbuffer3: Right, it is no more needed. Removed.

src/backend/vulkan/src/physical_device.rs
Uniformbuffer3: Done!

src/backend/vulkan/src/physical_device.rs
Uniformbuffer3: Ok, so i will apply the change for every function returning `UnsupportedFeature`.

Uniformbuffer3: Done.

src/backend/vulkan/src/physical_device.rs
Uniformbuffer3: Corrected.

src/backend/vulkan/src/physical_device.rs
Uniformbuffer3: Done. I have found other instances where i made the same error, corrected them too.

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+pub enum Resource<B: crate::Backend> {
+    /// Buffer
+    Buffer(B::Buffer),
+    /// Image
+    Image(B::Image),
+}
+
+/// External memory properties
+#[derive(Debug, PartialEq)]
+pub struct ExternalMemoryProperties {
```
kvark, line 22: should this be in bitflags? or are we anticipating other non-bit properties in the future?

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+    /// Constructor
+    pub fn new(exportable: bool, importable: bool, exportable_from_imported: bool) -> Self {
+        Self {
+            exportable,
+            importable,
+            exportable_from_imported,
+        }
+    }
+    /// Is the queried configuration exportable
+    pub fn is_exportable(&self) -> bool {
```
kvark, line 37: I don't think we should go out of our way to encapsulate the immutability of these fields. Just follow the existing pattern. We have memory requirements, and they are returned just by value, with all the fields public.

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+            ExternalMemory::HostMappedForeignMemory(ptr) => {
+                (ExternalMemoryType::HostMappedForeignMemory, ptr.into())
+            }
+        }
+    }
+}
+
+impl std::convert::TryFrom<(ExternalMemoryType, PlatformMemory)> for ExternalMemory {
+    type Error = &'static str;
+    fn try_from(tuple: (ExternalMemoryType, PlatformMemory)) -> Result<Self, Self::Error> {
```
kvark, line 386: Do we have to use `TryFrom` here? It seems to me that we are trying to fit into this trait for no apparent benefit.
Just having a method `fn from_platform(ty: ExternalMemoryType, pm: PlatformMemory) -> Result<Self, &'static str>` would be cleaner.

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+
+#[cfg(windows)]
+impl std::convert::TryFrom<(ExternalMemoryType, Handle)> for ExternalMemory {
+    type Error = &'static str;
+    fn try_from(tuple: (ExternalMemoryType, Handle)) -> Result<Self, Self::Error> {
+        ExternalMemory::try_from((tuple.0, PlatformMemory::from(tuple.1)))
+    }
+}
+
+impl std::convert::TryFrom<(ExternalMemoryType, Ptr)> for ExternalMemory {
```
kvark, line 449: We don't need to implement all of these. If the user has a `Handle`, let them call `PlatformMemory::from` and just use the single way.

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+    type Error = &'static str;
+    fn try_from(tuple: (ExternalMemoryType, Ptr)) -> Result<Self, Self::Error> {
+        ExternalMemory::try_from((tuple.0, PlatformMemory::from(tuple.1)))
+    }
+}
+
+bitflags::bitflags! {
+    /// Possible usages for an image.
+    pub struct ImageUsage: u32 {
```
kvark, line 458: can we just use the existing `hal::image::Usage` instead?

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+        const BLIT_DST = (1 << 11);
+        /// Format can be filtered with VK_FILTER_LINEAR when being sampled
+        const SAMPLED_IMAGE_FILTER_LINEAR = (1 << 12);
+    }
+}
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+#[derive(Debug)]
+/// Drm format properties
+pub struct DrmFormatProperties {
```
kvark, line 491: Do we need to have it here? If it's a Vulkan/Linux-only structure and an entry point, can we just make them public in `gfx-backend-vulkan` instead of having them in gfx-hal?

src/hal/src/external_memory/mod.rs
```diff
@@ -0,0 +1,498 @@
...
+            importable: false,
+            exportable_from_imported: false,
+        }
+    }
+}
+
+bitflags!(
+    /// External memory type flags.
+    #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+    pub struct ExternalMemoryTypeFlags: u32 {
```
kvark, line 62: Can we move this into the `external-memory` crate? We'd want the gfx-hal interface to be platform-abstract, so the least we can do here is to encapsulate the `cfg()` stuff into a sub-crate.

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Initially `ExternalMemoryProperties` got some other non boolean members, but due to the API shrink, they become useless and removed, but i hadn't noticed it can be turned into a bitflags. I will do that

src/hal/src/external_memory/mod.rs
Uniformbuffer3: I could move it to the Vulkan backend if you desire, i have wrote it in hal because OpenGL also got the possibility to use this feature, even if i'm not enough familiar to implement on it's backend. I know that this feature is used by current Wayland compositors to match the OpenGL image format with the drm/kms surfaces one. The Vulkan extension also states that it is equivalent to using 2 EGL extensions:
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_image_drm_format_modifier.html#_description

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Done!

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Since the structure has become a bitflag, that functions has been removed, they are no more needed.

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Effectively your approach is much better and cleaner. Done!

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Yes, they are not needed. Removed.

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Yes, but some usages are missing. Since `hal::image::Usage` is a biflags, i could just add the missing usages without breaking the API. I will do that

src/hal/src/external_memory/mod.rs
Uniformbuffer3: Yes, i can move it, or, if you prefer, i could just remove `cfg()`s and just expose every memory type to every platforms (and makes the functions panic if a memory type is used on the wrong platform). The initial idea was to just expose what the target platform can support, so that the user simply cannot use the wrong memory type, but i understant that having all these `cfg()`s could be undesirable.

src/backend/vulkan/src/physical_device.rs
Uniformbuffer3: I didn't know that. I have fixed the external memory extensions and adjusted functions accordingly.

src/hal/src/external_memory/mod.rs
kvark: Basically what we want is gfx-hal having no cfg() switches. If this is supported on Vk and Gl, then let's make it an optional feature that just happens to be linux-only. Would it work?

src/hal/src/external_memory/mod.rs
kvark: Moving seems fine

src/hal/src/external_memory/mod.rs
kvark: The other functions don't need to care about this. Since gfx-hal is unsafe, it has a set of expectations for the users behavior. Functions like `create_image` would just not expect anything unusual. In addition, the flags will be documented about what contexts are applicable.

src/backend/dx11/src/device.rs
```diff
@@ -2461,6 +2461,30 @@ impl device::Device<Backend> for Device {
         // TODO
     }
 
+    unsafe fn set_display_power_state(
```
kvark: weird to see these changes here. Wrong rebase?

Uniformbuffer3: There was some conflicts that makes the github merge check to fail, so i have merged it with the master branch and solved the conflicts. Shouldn't I have?

kvark: Please rebase instead of merging

src/backend/vulkan/src/conv.rs
Uniformbuffer3: Done!

src/backend/vulkan/src/physical_device.rs
Uniformbuffer3: Look to works correctly now. Please notify me if you still detect the problem and i will fix it as soon as possible.

src/hal/src/external_memory/mod.rs
Uniformbuffer3: I thought that `hal::image::Usage` was the equivalent of `ImageUsage` i was using, but instead the right equivalent was `hal::format::ImageFeature` that is already present and it is now used instead. So `ImageUsage` does not exist anymore because it is no more needed.

src/hal/src/external_memory/mod.rs
Uniformbuffer3: `DrmFormatProperties` is now returned as part of `format::Properties` as `Vec<DrmFormatProperties>`. Cfg flags as been removed and if the platform it is not supported, the field will simply be an empty vec. This change has allowed to merge the required function into the already existing `PhysicalDevice::format_properties` so no new function on hal is required.

src/backend/dx11/src/device.rs
Uniformbuffer3: I'm sorry for the inconvenience, i hope i made the rebase correctly. If not, please notify me and will fix it as soon as possible.

kvark:
What's the status of the PR now?

Uniformbuffer3:
I have to test the part of the functions related to exporting and importing an image with a drm format modifier and improve the documentation. Then it should be ready for review. To fit the drm format modifier feature i have slightly changed the api, but the general behaviour is the same.

Uniformbuffer3:
I have moved all the structures/enums that has some `#[cfg(...)]` fields but one: `ExternalImageMemory`. Unfortunately it has the `DmaBuf` variant containing the `DrmFormatImageProperties` structure. That structure is defined as:
```
/// Description of drm format properties used to create an image using drm format modifier.
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct DrmFormatImageProperties {
    /// Drm format modifier
    pub drm_modifier: format::DrmModifier,
    /// Plane subresource layouts
    pub plane_layouts: Vec<SubresourceFootprint>,
}
```
The problem is that it use the `SubresourceFootprint` structure that is an already existing one.
If i move both `ExternalImageMemory` and `DrmFormatImageProperties` to the auxiliary crate, `SubresourceFootprint` would not be available, so it does not compile.
The only solutions that come in my mind are:
- Move `SubresourceFootprint` too to an axiliary crate (not necessarily the external memory one)
- Made a copy of `SubresourceFootprint` with a different name, so that everything can be moved the auxiliary crate.
Can i ask you a suggestion about how could i solve this problem?

kvark:
Thank you for raising this up! It appears to me that making a copy of `SubresourceFootprint` is a fine way to proceed here.

kvark (approved):

PR #3758, opened 2021-05-12
[metal] provide size information to the shaders

### LINKED ISSUE #3757: Support dynamic array sizes on Metal
https://github.com/gfx-rs/wgpu/pull/1381 is blocked because gfx-backend-metal isn't providing a buffer for storage binding sizes. There is a draft PR #3751 that addresses this

kvark:
Fixes #3757
Closes #3751

~~The only thing left here is re-binding on descriptor changes. Also need to test to see if it actually works...~~

### DIFF
--- a/src/backend/metal/src/command.rs
+++ b/src/backend/metal/src/command.rs
@@ -274,11 +274,12 @@ pub struct CommandBuffer {
 unsafe impl Send for CommandBuffer {}
 unsafe impl Sync for CommandBuffer {}
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 struct Temp {
     clear_vertices: Vec<ClearVertex>,
     blit_vertices: FastHashMap<(Aspects, i::Level), Vec<BlitVertex>>,
     render_attachments: Vec<(metal::Texture, com::ClearValue)>,
+    binding_sizes: Vec<native::StorageBindingSize>,
 }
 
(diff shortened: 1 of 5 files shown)


kvark:
After a few more fixes, I confirmed that this works, at least on Boids example.

kvark:
With https://github.com/gfx-rs/naga/pull/858 and https://github.com/gfx-rs/naga/pull/850 (in an override), I can see the proper array length in the shadow example (to test graphics vs compute in boids). So the gfx-rs part works.

kvark:
(copying from Matrix for posterity)

I thought about simplifying this, and the only idea I got is to skip the pipeline layout stage, and store a hashmap(group, binding) to buffer size in the command encoder state. This would potentially be a tiny bit slower, since hashmap lookups are slower than areay indexing, but unclear how much.

Another approach - make the contents of the sizes buffer dependent on pipeline layout only, and nothing else. Thix would avoid the need to re-map it to the pipeline, but would potentially have larger buffers with mire stages than necessary...

Neither is a clear win.

kvark:
I tried the hashmap path, and it seems to work out pretty well, shaving off 50 LOC:
>  3 files changed, 47 insertions(+), 97 deletions(-)

kvark:
let's take the risk, need to unblock wgpu
post-landing review would be nice!
bors r+

Issue #3698, opened 2021-03-25
Is there a way to access raw handles for components?


jdu:
Short info header: 

- GFX version: 0.6.0+
- OS: All
- GPU: All

Is there way to expose the raw handle for things like the `Device` in order "shim" libraries into gfx-rs?

For example the skulpin project using rafx, can get a raw handle to instance, device, etc.. to pass into skia-safe to bind to a skia `BackendContext`. You can see how skulpin does this here https://github.com/aclysma/skulpin/blob/master/skulpin-renderer/src/skia_support.rs#L48 with rafx.

Is there a way to get a passable handle from gfx-rs to perform something similar in order to allow shimming of skia into the gfx-backend-* I've looked through the docs a fair bit and tried a few different things, but i'm coming up empty-handed to see how I could achieve this.

kvark:
There is no *portable* way to get the raw handles off gfx-rs objects, since there isn't always 1:1 mapping to anything under the API. Instead, each backend can expose its own raw APIs as needed. For example, Vulkan Instance could make the `inner` here public: https://github.com/gfx-rs/gfx/blob/2ac006f7fe86f32dd5b5e07f9daf44cd876a66f0/src/backend/vulkan/src/lib.rs#L74

If you make a list of things you'd need exposed, we can make sure they are available.

jdu:
I'll have a look at what exactly needs to be made public and see if I can come up with a concrete list based on the different backends. I can do PRs to the individual backends to mark the items public if that's easier than dumping a list here and you can approve/reject depending on whether it aligns with the project.

kvark:
The PRs are definitely preferred and much appreciated. Thank you!

zarik5:
I'm also interested in this. Personally I would use this to connect gfx-hal with OpenXR, OpenVR (driver API) and FFmpeg hardware acceleration API. Integrating these APIs directly into gfx-hal would not make much sense to me.

PR #3762, opened 2021-05-16
Low level interop with Vulkan backend

zarik5:
Related issues: #3698 #3761 https://github.com/blaind/xrbevy/issues/1

This PR exposes some Vulkan-backend-specific details needed for interop with other low level APIs. In particular this PR aims to implement a minimum viable API to allow OpenXR integration as a separate crate.

This PR is nowhere complete. I opened it to get some feedback; this is my first PR on a project the size of gfx-hal.

This is the first of multiple PRs needed for OpenXR support in wgpu.

PR checklist:
- [ ] tested examples with the following backends: Vulkan

EDIT: To make things clear, this is another take on #3761. Me and @blaind are working together on this route.

### DIFF
--- a/src/backend/vulkan/src/device.rs
+++ b/src/backend/vulkan/src/device.rs
@@ -1216,37 +1216,12 @@ impl d::Device<B> for super::Device {
         let is_cube = image
             .flags
             .intersects(vk::ImageCreateFlags::CUBE_COMPATIBLE);
-        let mut image_view_info;
-        let mut info = vk::ImageViewCreateInfo::builder()
-            .flags(vk::ImageViewCreateFlags::empty())
-            .image(image.raw)
-            .view_type(match conv::map_view_kind(kind, image.ty, is_cube) {
-                Some(ty) => ty,
-                None => return Err(image::ViewCreationError::BadKind(kind)),
-            })
-            .format(conv::map_format(format))
-            .components(conv::map_swizzle(swizzle))
-            .subresource_range(conv::map_subresource_range(&range));
+        let view_type = match conv::map_view_kind(kind, image.ty, is_cube) {
+            Some(ty) => ty,
+            None => return Err(image::ViewCreationError::BadKind(kind)),
+        };
 
-        if self.shared.image_view_usage {
-            image_view_info = vk::ImageViewUsageCreateInfo::builder()
-                .usage(conv::map_image_usage(usage))
-                .build();
-            info = info.push_next(&mut image_view_info);
-        }
-
-        let result = self.shared.raw.create_image_view(&info, None);
-
-        match result {
-            Ok(raw) => Ok(n::ImageView {
-                image: image.raw,
-                raw,
-                range,
-            }),
-            Err(vk::Result::ERROR_OUT_OF_HOST_MEMORY) => Err(d::OutOfMemory::Host.into()),
-            Err(vk::Result::ERROR_OUT_OF_DEVICE_MEMORY) => Err(d::OutOfMemory::Device.into()),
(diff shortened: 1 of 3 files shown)


zarik5:
I see some fields of the backend structs `Instance`, `PhysicalDevice`, `Device` etc. are `pub`. Shouldn't they be `pub(crate)`? If some fields need to be made public (for example as part of this PR) I think getter methods should be preferred.

zarik5:
For now I remove `::raw()` methods. They are not needed for the OpenXR integration and it is not clear what should be exposed for potential integration with other APIs (which might require Vulkan objects created by gfx-hal). This can be deferred to a future PR.

src/backend/vulkan/src/lib.rs
```diff
@@ -353,6 +354,205 @@ unsafe extern "system" fn debug_report_callback(
...
+
+    /// # Safety
+    /// `raw_instance` must be created using at least the extensions provided by `Instance::required_extensions()`
+    /// and the layers provided by `Instance::required_extensions()`.
+    /// `driver_api_version` must match the version used to create `raw_instance`.
+    /// `extensions` must match the extensions used to create `raw_instance`.
+    /// `raw_instance` must be manually destroyed *after* gfx-hal Instance has been dropped.
+    pub unsafe fn from_raw(
+        #[cfg(not(feature = "use-rtld-next"))] entry: Entry,
+        #[cfg(feature = "use-rtld-next")] entry: EntryCustom<()>,
```
kvark, line 462: we should probably use the fact that `type Entry = EntryCustom<Arc<Library>>`. So the only thing different between `use-rtld-next` and not is the generic parameter (which we can typedef somewhere).

kvark (approved):
I really like the way this is written. Top-notch work here!
Assuming this API works for you, let's make this mergeable.

zarik5:
@kvark Thank you! The work here is not finished, I'm developing this API while working on a minimum example for integration with OpenXR. I have to work around the missing support of multiview in gfx-hal, which should be handled in another PR.

src/backend/vulkan/src/lib.rs
zarik5: Great idea! Using cfg-gated parameters was also confusing rust-analyzer.

zarik5: I preferred using a typedef on `Entry`, typedef-ing only the type argument felt like splitting the type in half.

src/backend/vulkan/src/device.rs
```diff
@@ -2044,6 +2019,45 @@ impl super::Device {
...
+
+    pub unsafe fn image_view_from_raw(
+        &self,
+        raw_image: vk::Image,
+        view_type: vk::ImageViewType,
+        format: format::Format,
+        swizzle: format::Swizzle,
+        usage: image::Usage,
+        range: image::SubresourceRange,
+    ) -> Result<n::ImageView, image::ViewCreationError> {
```
zarik5, line 2031: I would have preferred adding a `Device::image_from_raw()` method, but in my case it is OpenXR that creates the Vulkan images and it hides some details needed to construct gfx images.

zarik5:
The PR is ready. I tested it with this [example](https://github.com/zarik5/openxrs/blob/master/openxr/examples/vulkan-gfx.rs). I took the vulkan.rs example from [openxrs](https://github.com/Ralith/openxrs) and converted most of vulkan code to gfx-hal. I had some problems with crashing on close, fixed by using DropGuard on vk_instance and vk_device.

kvark (approved):
bors r+

PR #3767, opened 2021-05-29
Make `enabled_extensions` public in Vulkan backend

LaylConway:
This is required for OpenXR support, when creating a custom device gfx-hal needs to be queried for what raw vulkan extensions to enable for the given gfx-hal features.

PR checklist:
 - gfx_backend_vulkan


kvark:
No problem exposing this, but seems rather suspicious :)
We'll see how you are using it.
bors r+

PR #3771, opened 2021-06-02
Fix WebGL backend for wgpu-rs cube example

Frizi:
Fixes "cube" example in wgpu-rs for webgl backend on Windows. Will also fix MacOS once https://github.com/gfx-rs/naga/pull/936 gets included.

`delete_shader` is delayed after linking the program. Not doing that seems to be randomly triggering linking issues. I'm not exactly sure under what circumstances it tends to happen, but the fix made it reliable.

The most important fix is about texture filtering for intager textures. ANGLE decides to ignore integer textures if their filtering is kept as default. Instead, it silently substitutes an empty 1px placeholder texture into the shader binding. Forcing the integer textures to NEAREST filtering makes that problem go away.

### DIFF
--- a/src/backend/gl/src/device.rs
+++ b/src/backend/gl/src/device.rs
@@ -145,6 +145,8 @@ impl Device {
             name_binding_map: &mut name_binding_map,
         };
 
+        let mut shaders_to_delete = arrayvec::ArrayVec::<[_; 3]>::new();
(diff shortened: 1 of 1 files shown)


kvark (approved):
Good stuff, thank you!
bors r+

PR #3772, opened 2021-06-06
Correction on the description of the direct display behaviour for the quad example

Uniformbuffer3:
Hi,
during the development of the direct display path i had removed the input acquisition so that no newer depencencies was needed. Unfortunately i have forgotten to update the example description accordingly, so this will fix it. The code works as intendend, just the description was not updated. Sorry for the incovenience.

Have a good day


kvark (approved):
thanks!
bors r+

PR #3773, opened 2021-06-11
Support wgpu skybox example on webgl

Frizi:
This changesets adds necessary features to GL backend in order to support skybox example from wgpu repository in webgl.

Most notable is adding support for cubemaps. Unfortunately due to the complexity of emulating image views in GL and the fact that you can't use a single gl texture in multiple ways, the support is very limited. Every 6-layer square image is assumed to be a cubemap, and creating other than cubemap views to it will fail. Further work on that should be done eventually, but I want to get the basic support first.

Another thing is support for compressed textures, specifically S3TC formats. This was added, because skybox example texture format fallback ise BGRA, which is not supported on webgl at all. Adding support for compressed format allowed me to run the example unmodified. Further work here is adding more formats and adding a downlevel flag for BGRA texture format, which is for some reason not supported on GLES.

### DIFF
--- a/src/backend/gl/src/conv.rs
+++ b/src/backend/gl/src/conv.rs
@@ -1,20 +1,6 @@
 use crate::native::VertexAttribFunction;
 use hal::{format::Format, image as i, pso};
 
-/*
-pub fn _image_kind_to_gl(kind: i::Kind) -> t::GLenum {
-    match kind {
-        i::Kind::D1(_) => glow::TEXTURE_1D,
-        i::Kind::D1Array(_, _) => glow::TEXTURE_1D_ARRAY,
-        i::Kind::D2(_, _, i::AaMode::Single) => glow::TEXTURE_2D,
-        i::Kind::D2(_, _, _) => glow::TEXTURE_2D_MULTISAMPLE,
-        i::Kind::D2Array(_, _, _, i::AaMode::Single) => glow::TEXTURE_2D_ARRAY,
-        i::Kind::D2Array(_, _, _, _) => glow::TEXTURE_2D_MULTISAMPLE_ARRAY,
-        i::Kind::D3(_, _, _) => glow::TEXTURE_3D,
-        i::Kind::Cube(_) => glow::TEXTURE_CUBE_MAP,
-        i::Kind::CubeArray(_, _) => glow::TEXTURE_CUBE_MAP_ARRAY,
-    }
-}*/
 
 pub fn filter_to_gl(mag: i::Filter, min: i::Filter, mip: i::Filter) -> (u32, u32) {
     use hal::image::Filter::*;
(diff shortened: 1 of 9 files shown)


src/backend/gl/src/info.rs
```diff
@@ -689,6 +690,13 @@ pub(crate) fn query_all(
     {
         features |= Features::INDEPENDENT_BLENDING;
     }
+    if info.is_supported(&[
+        Es(3, 0),
+        Ext("WEBGL_compressed_texture_s3tc"),
```
Gordon-F, line 695: Currently, extensions list is empty for `Webgl`.

https://github.com/gfx-rs/gfx/blob/0a0cdaa39a1a84873aa7b195bdf2ba7fd3df478e/src/backend/gl/src/info.rs#L302-L304

src/backend/gl/src/queue.rs
```diff
@@ -115,7 +115,7 @@ impl Queue {
                 (glow::BYTE, glow::SHORT, glow::INT),
             C::Uint | C::Unorm =>
                 (glow::UNSIGNED_BYTE, glow::UNSIGNED_SHORT, glow::UNSIGNED_INT),
-            C::Float => (glow::ZERO, glow::HALF_FLOAT, glow::FLOAT),
```
Gordon-F, line 118: Accidentally removed?

Frizi: Yes, though the commented out code shouldn't really be here in the first place. We have git for that.

Gordon-F: Maybe it would be better to remove all commented code?

src/backend/gl/src/info.rs
Frizi: Oh I see. It worked for me because the `is_supported` is an "or" of passed requirements, and `Es 3.0` is always satisfied. That means this code is indeed wrong. I need to populate the extension list.

src/backend/gl/src/conv.rs
```diff
@@ -87,6 +76,70 @@ impl FormatDescription {
     }
 }
 
+pub const COMPRESSED_RGB_S3TC_DXT1_EXT: u32 = 0x83F0;
```
Gordon-F, line 79: I think it would be better to expose this on `glow` side.

src/backend/gl/src/info.rs
Gordon-F: First we need to expose extension list from `glow` somehow. Or copy-paste extensions checking from `glow` codebase:
https://github.com/grovesNL/glow/blob/7775e74363877347b2210d91277646a09d70dd08/src/web_sys.rs#L105-L109

Frizi: I've exposed the extensions in https://github.com/grovesNL/glow/pull/165. Let's wait for it to be in.

src/backend/gl/src/conv.rs
Frizi: 👍 opened a PR https://github.com/grovesNL/glow/pull/166

PR #3780, opened 2021-07-14
Fix crash when GfxManagedMetalLayerDelegate's backing Objective-C class has already been registered

yamadapc:
Fixes a crash when GfxManagedMetalLayerDelegate objective-c class is already registered.

I'm sure there's a better solution, let me describe the issue:

* I'm writing an application which hosts audio plugins & uses `iced`, internally it's using `gfx`
* I've an audio plugin I've written which also uses `iced` & uses `gfx` internally too
* When the plugin is loaded into the host, this class name has already been registered and thus registration fails
* Unwrap on registration bits causes a crash

I really don't know what this code is trying to do, but as far as registering the objective-c class decl, I suppose there're a couple of options:

* Declare an unique name every time
* Don't declare the class if it's present
  - Issue: plugins may not have the same version of `gfx-backend-metal`, so the class decl needs to be binary compatible
* Declare a versioned class name

Let me know what you think & I'm happy to change the PR to go with another option. I suppose an unique class name is a good way to avoid problems, but that can be a bit messy if this class decl will be read somewhere else by name.

PR checklist:

### DIFF
--- a/src/backend/metal/src/lib.rs
+++ b/src/backend/metal/src/lib.rs
@@ -65,6 +65,8 @@ use hal::{
 use range_alloc::RangeAllocator;
 
 use cocoa_foundation::foundation::NSInteger;
+use core_graphics_types::base::CGFloat;
+use core_graphics_types::geometry::CGSize;
 #[cfg(feature = "dispatch")]
 use dispatch;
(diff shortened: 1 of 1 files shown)


kvark:
thank you!
bors r+

PR #3781, opened 2021-07-16
Fix broken links.

meowjesty:
- wgpu-rs was linking to the archived version;
- rust-tutorials/learn-gfx-hal doesn't exist anymore;

As there is only 1 tutorial link for gfx, I've changed plurals to singular as well.

This is a minor change, does it need an issue?


cwfitzgerald:
Thanks!

bors r+

